PR #1973, opened 2016-11-10
INT-4156: Merge HTTP Java DSL and some refinements

artembilan:
JIRA: https://jira.spring.io/browse/INT-4156

* Copy `Http` Java DSL from with some fixes to the `HttpMessageHandlerSpec` and Java 8 style
* Port `HttpDslTests` and "de-Boot" it with the Mock MVC
* Upgrade to Spring Security 4.2.0
* Make `spring-integration-security` as a dependency for HTTP module for better test coverage for Security from SI Web perspective

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/BaseHttpInboundEndpointSpec.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/BaseHttpInboundEndpointSpec.java
@@ -42,6 +42,7 @@ import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.multipart.MultipartResolver;
 
 /**
+ * A base {@link MessagingGatewaySpec} for the {@link HttpRequestHandlingEndpointSupport} implementations.
  *
  * @param <S> the target {@link BaseHttpInboundEndpointSpec} implementation type.
  * @param <E> the target {@link HttpRequestHandlingEndpointSupport} implementation type.
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/BaseHttpInboundEndpointSpec.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/BaseHttpInboundEndpointSpec.java
@@ -1,0 +1,268 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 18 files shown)


garyrussell:
Can you explain the second commit?

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/BaseHttpInboundEndpointSpec.java
```diff
@@ -0,0 +1,268 @@
...
+import org.springframework.integration.http.inbound.HttpRequestHandlingEndpointSupport;
+import org.springframework.integration.http.inbound.RequestMapping;
+import org.springframework.integration.http.support.DefaultHttpHeaderMapper;
+import org.springframework.integration.mapping.HeaderMapper;
+import org.springframework.util.Assert;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.multipart.MultipartResolver;
+
+/**
+ *
```
garyrussell: Javadoc

garyrussell (changes requested):
LGTM; same deal as file, though - inconsistent javadocs

artembilan:
> Can you explain the second commit?

I jsut wondered if that will work without exposing `AuthenticationManager` as a bean, since that isn't Spring Security requirements and we are in the exactly `WebSecurityConfigurerAdapter` implementation. Therefore can use super methods. :smile: 

Will address JavaDocs concern soon.

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	HttpMessageHandlerSpec(Expression uriExpression, RestTemplate restTemplate) {
+		this.target = new HttpRequestExecutingMessageHandler(uriExpression, restTemplate);
+		this.target.setUriVariableExpressions(this.uriVariableExpressions);
+		this.target.setHeaderMapper(this.headerMapper);
+		this.restTemplate = restTemplate;
+	}
+
+	HttpMessageHandlerSpec expectReply(boolean expectReply) {
```
garyrussell: Still some missing here (and in the `Http` factory).

artembilan: This is without JavaDocs because it isn't `public`.

Doh... Have missed `Http` factory. It is fully without JavaDocs. 
That let me run away somehow :smile:

garyrussell: >This is without JavaDocs because it isn't public.

oops

artembilan: It isn't `public` because there is no reason to expose this option to the API, since we can simply restrict an access via `outboundChannelAdapter()/outboundGateway()` factory methods.

Does it make sense for you?

garyrussell: Yes, the oops was to acknowledge my mistake - I missed the package scope.

garyrussell:
Merged as 6c4d9b2111221250e9d8e30f1be7ff794fe60636

PR #1976, opened 2016-11-14
INT-4158: Port Java DSL for JPA module

artembilan:
JIRA: https://jira.spring.io/browse/INT-4158

* Add JavaDocs
* De-Boot `JpaDslTests`
* Fix typos and generics inconsistency in the `JpaOutboundGatewayFactoryBean`

### DIFF
--- a/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/dsl/JpaBaseOutboundEndpointSpec.java
+++ b/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/dsl/JpaBaseOutboundEndpointSpec.java
@@ -21,13 +21,10 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 
-import org.springframework.beans.factory.BeanCreationException;
-import org.springframework.beans.factory.support.DefaultListableBeanFactory;
 import org.springframework.integration.dsl.ComponentsRegistration;
 import org.springframework.integration.dsl.MessageHandlerSpec;
 import org.springframework.integration.jpa.core.JpaExecutor;
 import org.springframework.integration.jpa.outbound.JpaOutboundGateway;
-import org.springframework.integration.jpa.outbound.JpaOutboundGatewayFactoryBean;
 import org.springframework.integration.jpa.support.JpaParameter;
 import org.springframework.integration.jpa.support.parametersource.BeanPropertyParameterSourceFactory;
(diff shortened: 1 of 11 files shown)


spring-integration-jpa/src/test/java/org/springframework/integration/jpa/dsl/JpaDslTests.java
```diff
@@ -0,0 +1,261 @@
...
+import org.springframework.transaction.PlatformTransactionManager;
+
+/**
+ * @author Artem Bilan
+ *
+ * @since 5.0
+ */
+@RunWith(SpringRunner.class)
+@DirtiesContext
+public class JpaDslTests {
```
garyrussell: `Dsl` is redundant - however, I can see it's useful to have it in the class name - I haven't been renaming these tests in the modules I ported; perhaps we should be consistent either way.

garyrussell (changes requested):
That's all; minor.

spring-integration-jpa/src/test/java/org/springframework/integration/jpa/dsl/JpaDslTests.java
artembilan: Yeah! Not a question. 
I just was a bit paranoiac that the name like `JpaTests`  doesn't say so much about the purpose of the test. Although maybe `.dsl` package does the trick...

Will rename for consistency.

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/dsl/JpaBaseOutboundEndpointSpec.java
```diff
@@ -0,0 +1,191 @@
...
+	@Override
+	protected JpaOutboundGateway doGet() {
+		if (!this.jpaParameters.isEmpty()) {
+			this.jpaExecutor.setJpaParameters(this.jpaParameters);
+		}
+
+		/*
+		We need this artificial BeanFactory to overcome JpaOutboundGatewayFactoryBean initialization.
+		The real BeanFactory will be applied later for the target JpaOutboundGateway instance.
+		*/
```
garyrussell: Got it - ok; let's leave it like this.

garyrussell (approved):

garyrussell:
Merged as 1bba73fc060faec6141e2edd957c98ad6a56615c

PR #1977, opened 2016-11-14
INT-4138: MQTT: Outbound Adapter Improvements

garyrussell:
JIRA:https://jira.spring.io/browse/INT-4138

Expressions for topic, qos, retained.

Also change inbound mapping to `RECEIVED_...` headers.

Fix some minor asciidoc problems in (s)ftp.


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/AbstractMqttMessageHandler.java
```diff
@@ -109,8 +198,49 @@ protected void incrementClientInstance() {
...
+						else {
+							Integer result = this.qosProcessor.processMessage(m);
+							if (result == null) {
+								return this.defaultQos;
+							}
+							else {
+								return result;
+							}
+						}
+					},
```
artembilan: I try to find the proper words to say that this design doesn't look good, but still failing.
Maybe the `this.qosProcessor` can be just of that `defaultQosFunction`:
```
private MessageProcessor<Integer> qosProcessor = MqttMessageConverter.defaultQosFunction(this.defaultQos)::apply;
```
?
Although from there it looks like that `defaultQosFunction()` doesn't need to have `this.defaultQos` fallback option, because we have it here in the current lambda. But looking further we don't need that  `defaultQosFunction()` , because we can simply have `qosProcessor` as a function of ` message.getHeaders().get(MqttHeaders.QOS)` and fallback to `this.defaultQos` here again.

I don't say that the logic is wrong, but I just see the design a bit differently...

Huh! That is my five fingers :smile:

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/AbstractMqttMessageHandler.java
garyrussell: Yes; I got a bit wrapped around the axle with this wrt tangling between the adapter and converter; I will rethink it a bit.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/AbstractMqttMessageHandler.java
```diff
@@ -109,8 +198,49 @@ protected void incrementClientInstance() {
...
+		}
+		if (this.qosProcessor instanceof BeanFactoryAware && getBeanFactory() != null) {
+			((BeanFactoryAware) this.qosProcessor).setBeanFactory(getBeanFactory());
+		}
+		if (this.retainedProcessor instanceof BeanFactoryAware && getBeanFactory() != null) {
+			((BeanFactoryAware) this.retainedProcessor).setBeanFactory(getBeanFactory());
+		}
 		if (this.converter == null) {
-			this.converter = new DefaultPahoMessageConverter(this.defaultQos, this.defaultRetained);
+			DefaultPahoMessageConverter mConverter = new DefaultPahoMessageConverter(
```
artembilan: Ah! Sorry. Didn't notice that `this.converter` isn't of `DefaultPahoMessageConverter` type.
Yep! Never mind then :smile: 
Maybe just name of the variable would be more informative ? :wink: 

Typically we avoid reduction since we are Open Source and our code should be as specific as possible.
Much easier to read it in the future. 

Right, first of all I was confused about variable name.

artembilan (changes requested):
The Travis problem isn't up to this PR, but looks like we need to increase the memory, but I don't know how...

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/3035bc716d6bc8199a5f917ceb7d46c3074a42e3

PR #1978, opened 2016-11-17
Remove DSL deprecated methods and others

artembilan:
Since the DSL code is new in this version there is no reason to keep deprecated method

* Remove deprecated `resequence()` and `aggregate()` in the `IntegrationFlowDefinition`
* Remove `InternalAggregatingMessageHandler` with the `MessageGroupProcessorWrapper` delegation logic in favor of newly introduced `AbstractCorrelatingMessageHandler.setOutputProcessor()`
* Rework `Promise` Gateway Docs to the `Mono`
* Rename `AsyncGatewayTests` "promise" words to "mono"


artembilan:
Do not merge yet.
I have something for the `RecipientListRouter` as well.

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMessageHandler.java
```diff
@@ -31,24 +29,24 @@
  *
  * @since 5.0
  */
-class GatewayMessageHandler extends AbstractReplyProducingMessageHandler implements Lifecycle {
+public class GatewayMessageHandler extends AbstractReplyProducingMessageHandler implements Lifecycle {
```
garyrussell: Needs Javadocs now it's public.

spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
```diff
@@ -40,7 +39,7 @@
  *
  * @since 5.0
  */
-class LambdaMessageProcessor implements MessageProcessor<Object>, BeanFactoryAware {
+public class LambdaMessageProcessor implements MessageProcessor<Object>, BeanFactoryAware {
```
garyrussell: Needs Javadocs

spring-integration-core/src/main/java/org/springframework/integration/router/RecipientListRouter.java
```diff
@@ -93,12 +93,15 @@ public void setChannels(List<MessageChannel> channels) {
 	 * @param recipients The recipients.
 	 */
 	public void setRecipients(List<Recipient> recipients) {
-		Assert.notEmpty(recipients, "recipients must not be empty");
+		Assert.notEmpty(recipients, "'recipients' must not be empty");
 		ConcurrentLinkedQueue<Recipient> originalRecipients = this.recipients;
 		this.recipients.clear();
 		this.recipients.addAll(recipients);
```
garyrussell: This should be atomic - looks like it was broken by INT-2856; although not 

There is a small timing hole where there's no recipients, or an incomplete list.

spring-integration-core/src/main/java/org/springframework/integration/router/RecipientListRouter.java
```diff
@@ -93,12 +93,15 @@ public void setChannels(List<MessageChannel> channels) {
...
+		if (getBeanFactory() != null) {
+			this.recipients.forEach(recipient -> recipient.setChannelResolver(getChannelResolver()));
+		}
```
garyrussell: This needs to be done before the atomic update above.

spring-integration-core/src/main/java/org/springframework/integration/dsl/RecipientListRouterSpec.java
```diff
@@ -89,8 +97,14 @@ public RecipientListRouterSpec recipientMessageSelector(String channelName, Mess
...
 	 */
 	public <P> RecipientListRouterSpec recipient(String channelName, GenericSelector<P> selector) {
-		Assert.hasText(channelName);
-		((DslRecipientListRouter) this.target).add(channelName, selector);
+		MessageSelector messageSelector;
+		if (selector instanceof MessageSelector) {
+			messageSelector = (MessageSelector) selector;
+		}
+		else {
+			messageSelector = new MethodInvokingSelector(new LambdaMessageProcessor(selector, null));
```
garyrussell: Is this strictly correct? Don't we have to check it's a Lambda first? Someone might implement his own `GenericSelector<P>` that is not a `MessageSelector`.

spring-integration-core/src/main/java/org/springframework/integration/dsl/AggregatorSpec.java
```diff
@@ -57,17 +52,16 @@ public AggregatorSpec processor(Object target) {
...
+	 * @param target the target object.
 	 * @param methodName The method name for the output processor (or 'null' in which case, the
-	 *                   target object must have an {@link org.springframework.integration.annotation.Aggregator}
-	 *                   annotation).
+	 * target object must have an {@link org.springframework.integration.annotation.Aggregator} annotation).
 	 * @return the handler spec.
 	 */
 	public AggregatorSpec processor(Object target, String methodName) {
 		super.processor(target);
 		return this.outputProcessor(methodName != null
```
garyrussell: Remove `this.`; also line 75.

garyrussell (changes requested):
Some minor issues.

spring-integration-core/src/main/java/org/springframework/integration/dsl/RecipientListRouterSpec.java
artembilan: It might be called `Lamba` by accident, but the purpose is still there single method invocation purpose for provided `payloadType`. My idea to rely on the  strict `GenericSelector` implementation and bypass `MessagingMethodInvokerHelper` which is based on the SpEL and a bit slower than direct method invocation.

garyrussell: I realize the `LMP` will work for any functional interface. My point is, this won't work...

	.recipient("foo", new Foo())

	public static class Foo implements GenericSelector<Message<?>> {

		@Override
		public boolean accept(Message<?> source) {
			return false;
		}

		public boolean foo(String bar) {
			return false;
		}

	}

...because the LMP will select the second method not the first.

garyrussell:
Merged as 092d876fae368ff34a1af79a58bf869e3b5de2a3 with checkstyle fixes.

PR #1984, opened 2016-11-28
INT-4174: Expose Channel Subscriber Count

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4174

I considered just using the channel counter but I suppose there is a (small) possibility
that someone might possibly use a custom dispatcher and subscribe to it directly.

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractSubscribableChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractSubscribableChannel.java
@@ -21,6 +21,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.springframework.integration.MessageDispatchingException;
 import org.springframework.integration.dispatcher.AbstractDispatcher;
 import org.springframework.integration.dispatcher.MessageDispatcher;
+import org.springframework.integration.support.management.SubscribableChannelManagement;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageDeliveryException;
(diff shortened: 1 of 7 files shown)


spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractSubscribableChannel.java
```diff
@@ -37,19 +38,30 @@
...
 
 	private final AtomicInteger handlerCounter = new AtomicInteger();
 
+	@Override
+	public int getSubscriberCount() {
+		MessageDispatcher dispatcher = getRequiredDispatcher();
+		if (dispatcher instanceof AbstractDispatcher) {
+			return ((AbstractDispatcher) dispatcher).getHandlerCount();
+		}
+		return this.handlerCounter.get();
```
garyrussell: Good idea - working on it now...

artembilan: Great! 
But I think we can't backport it to `4.3.x`. 
Therefore that work should be done in the separate PR.

What is in your mind ?

garyrussell: Just cherry-pick the first commit.

Let me know if you prefer a separate PR.

artembilan (approved):
Merging and cherry-picking to `4.3.x` only first commit...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/96ff547c8c06c9e465270485890974ccfbdfdebb and cherry-picked to `4.3.x` only first commit.

Also note the fix for JPA tests race condition.

PR #1990, opened 2016-11-30
INT-4176: Fix Gemfire tests

artembilan:
JIRA: https://jira.spring.io/browse/INT-4176

The `DelayerHandlerRescheduleIntegrationTests` doesn't close `context` in the end of test.
That causes clashes for `beanFactoryResolver` in other tests.
When those tests are fail, we come into condition when new processes can't be started/stopped because of effect of non-stopped context

* Fix all the `DelayerHandlerRescheduleIntegrationTests` to stop `context` in the end of test
* Remove unused files in the `gemfire/store` test package
* Add current date value to the `CacheServerProcess` name to avoid clashes with other ran on the same host
* Add NPE check for `region` in the `GemfireMetadataStoreTests`
* Extract `@BeforeClass/@AfterClass` for the `CacheWritingMessageHandlerTests` to start only one Gemfire cache per class

**Cherry-pick to 4.3.x & 4.2.x**


garyrussell:
Merged as 9078a74a19ea0dd7beba93ee7e71f1776093bf8a ; cherry-picking now...

PR #1991, opened 2016-12-01
INT-4178: Retry `CannotSerializeTransactionException` in `JdbcLockRegistry`

vpavic:
JIRA: https://jira.spring.io/browse/INT-4178

The `JdbcLockRegistry` does not retry `CannotSerializeTransactionException` thrown by `DefaultLockRepository.acquire` which in turn causes `LockRegistryLeaderInitiator`s thread to get stuck.

* catch `CannotSerializeTransactionException` on `DefaultLockRepository.acquire` and ignore it to ensure a retry

I'm not sure how to test this properly since it's environment specific (PostgreSQL) but I'll be happy to update the PR if you have an idea how to test it.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -26,6 +26,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.springframework.dao.CannotAcquireLockException;
+import org.springframework.dao.CannotSerializeTransactionException;
 import org.springframework.dao.DataAccessResourceFailureException;
 import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
@@ -120,6 +121,9 @@ public class JdbcLockRegistry implements ExpirableLockRegistry {
 					}
 					break;
 				}
+				catch (CannotSerializeTransactionException e) {
+					// try again
+				}
 				catch (TransactionTimedOutException e) {
 					// try again
 				}
@@ -153,6 +157,9 @@ public class JdbcLockRegistry implements ExpirableLockRegistry {
 						}
 					}
 					break;
+				}
+				catch (CannotSerializeTransactionException e) {
+					// try again
 				}
 				catch (TransactionTimedOutException e) {
 					// try again
@@ -197,6 +204,9 @@ public class JdbcLockRegistry implements ExpirableLockRegistry {
 						this.delegate.unlock();
 					}
 					return acquired;
+				}
+				catch (CannotSerializeTransactionException e) {
+					// try again
 				}
 				catch (TransactionTimedOutException e) {
 					// try again


artembilan:
@vpavic ,

Having your description in the sample project:
> After some time, usually within a minute from startup, the following exception is logged to console:

I think even if we write PostgreSQL-specific test-case, that won't be so valuable to wait a minute until a `CannotSerializeTransactionException` occurs.

Although I think mock test on the `JdbcTemplate` to throw `SQLException` with the `cannotSerializeTransactionCodes` would be enough for coverage.

OTOH looking to your change I don't see reason to worry about.
Since recommendation is like:

> Reason code: Canceled on identification as a pivot, during write.
> Hint: The transaction might succeed if retried.

We can simply live without test on the matter.

WDYT?

vpavic:
Hey @artembilan, thanks for the feedback!

Yes, I agree with your assessment that the test probably doesn't add much value and that the proposed change simply follows the general recommendation on handling concerns specific to this particular isolation level.

If you're OK with that, I am as well :smile:

artembilan (approved):
Merging and cherry-picking to `master`...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/8198ff18615df6161a621d54544af2ebd906692c and cherry-picked.

stefanvassilev:
Hi @artembilan, @vpavic, 

I don't know whether I should open an issue, or ask here, however I experienced the same problems that are addressed and fixed in this PR. 

The problem, however was that the exception produced by the `DefaultLockRepository` was one coming from hibernate and was not not `CannotSerializeTransactionException` which would be then the translated spring one.

It seems to me that if one defines a custom default repository e.g 
```java
	@Bean
	public DefaultLockRepository lockRepository(DataSource dataSource) {
                DefaultLockRepository defaultLockRepository  = new DefaultLockRepository(dataSource);
                defaultLockRepository.setPrefix("prefix");
		return defaultLockRepository;
	}
```
which is not annotated with `@Repository` it won't get processed by`PersistenceExceptionTranslationPostProcessor` and exception that it produces will be coming from the provider.

I was able to work around it by wrapping the defaultLockRepository with wrapper.
```java
@Repository 
public class DefaultLockRepositorywrapper {
}
```
which i then annotated with repository and I didn't get the issue.

I have `spring-boot-starter-data-jpa` in classpath as well, however I don't know if that would be relevant.

Am I missing something here? Is there a better way to do it?

artembilan:
Not sure what you are talking about.
The code looks like this though:
```
@Repository
@Transactional
public class DefaultLockRepository implements LockRepository, InitializingBean {
```

So, there is a `@Repository` on the class.

Would you mind to start a StackOverflow thread with much more info?
Then after discussion over there we may consider to raise an new issue over here.


Thanks for understanding!

stefanvassilev:
@artembilan, 

I saw the code for `DefaultLockRepository` prior to writing. My question whether the`@Repository` annotation is honored when we define a custom `DefaultLockRepository` e.g. like this: 

```java
	@Bean
	public DefaultLockRepository lockRepository(DataSource dataSource) {
                DefaultLockRepository defaultLockRepository  = new DefaultLockRepository(dataSource);
                defaultLockRepository.setPrefix("prefix");
		return defaultLockRepository;
	}
```

as I was getting hibernate exceptions and not spring's `DataAccessExeption`.

artembilan:
I think it does. Because `DefaultLockRepository` has that annotation on it.
Your one probably doesn't, that's why it is a problem.

Please, raise a StackOverflow thread with much more info. For example some simple project to let us to play with to understand better what is your concern.

Thanks

PR #1989, opened 2016-11-30
Prepare for release

artembilan:
Change all the `BUILD-SNAPSHOT`s to their latest Milestones
Fix compatibility with those Milestones

NOTE: Would be better to merge this already after releasing Spring AMPQ 2.0.0.M1

PR #1993, opened 2016-12-02
INT-4177: Use Shared DefaultConversionService

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4177

Use DefaultConversionService.getSharedInstance()

    $ find . | grep '\.java$' | xargs sed -i '' -e 's/new DefaultConversionService/DefaultConversionService.getSharedInstance/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -447,7 +447,15 @@ project('spring-integration-jpa') {
 		compile ("org.eclipse.persistence:javax.persistence:$jpaApiVersion", optional)
 
 
-		testCompile "org.springframework.data:spring-data-jpa:$springDataJpaVersion"
+		testCompile ("org.springframework.data:spring-data-jpa:$springDataJpaVersion") {
+			exclude group: 'org.springframework', module: 'spring-beans'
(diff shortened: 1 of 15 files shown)


artembilan:
Looks like you haven't tested the fix:
```
/MethodInvokingMessageGroupProcessorTests.java:269: error: incompatible types: ConversionService cannot be converted to GenericConversionService

		GenericConversionService conversionService = DefaultConversionService.getSharedInstance();
```
😄 
See Travis report.

Although I think we can still live with the `new DefaultConversionService()` in those test.
That might be critical for `main` code base, but that is fine with tests. At least with me.

Just let me know and I'll fix during merge!

artembilan (changes requested):
How about also to restore what I have changed here:
https://github.com/spring-projects/spring-integration/pull/1989/files#diff-d764837e96980762e77da52c5277974eL71
?

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/00796702c06cef52c5ec85f6c2318363957ccb72

PR #1995, opened 2016-12-03
INT-4180: RecipientListRouterSpec Fix

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4180

Use true `ValueExpression` for recipient() without selector.

Assert selector epresion string is not null or empty.

Add test.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/RecipientListRouterSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/RecipientListRouterSpec.java
@@ -20,6 +20,7 @@ import org.springframework.expression.Expression;
 import org.springframework.integration.channel.DirectChannel;
 import org.springframework.integration.core.GenericSelector;
 import org.springframework.integration.core.MessageSelector;
+import org.springframework.integration.expression.ValueExpression;
 import org.springframework.integration.filter.ExpressionEvaluatingSelector;
 import org.springframework.integration.filter.MethodInvokingSelector;
 import org.springframework.integration.handler.LambdaMessageProcessor;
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/RecipientListRouterSpec.java
```diff
@@ -122,7 +124,7 @@ private static boolean isLambda(Object o) {
 	 * @return the router spec.
 	 */
 	public RecipientListRouterSpec recipient(MessageChannel channel) {
-		return recipient(channel, (String) null);
+		return recipient(channel, new ValueExpression<Boolean>(Boolean.TRUE));
```
artembilan: I think would be better to rework the target method as:
```
public RecipientListRouterSpec recipient(MessageChannel channel, Expression expression) {
        if (expression != null) {
              ExpressionEvaluatingSelector selector = new ExpressionEvaluatingSelector(expression);
              this.target.addRecipient(channel, selector);
              this.componentsToRegister.add(selector);
        }
        else {
              this.target.addRecipient(channel);
        }
	return _this();
}
```

to save some memory.

Let me know and I'll fix on merge

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dsl/RecipientListRouterSpec.java
garyrussell: OK; fine with me; thanks.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/295d609eb3e792a50594fab9d5b374dacb0c8844

PR #1988, opened 2016-11-29
INT-4173: Pagination for mongodb inbound adapter

yaronyam:
JIRA: https://jira.spring.io/browse/INT-4173

Support Pagination for mongodb inbound adapter by accepting a SpEL expression that can be evaluated into a BasicQuery object , so `skip()` and `limit()` can be used.
Example:

`<int-mongodb:inbound-channel-adapter query-expression="new org.springframework.data.mongodb.core.query.BasicQuery('{''name'' : ''Bob''}').limit(2)" />`

PR #1996, opened 2016-12-07
INT-4173: Pagination for mongodb inbound adapter - v5.x

yaronyam:
JIRA: https://jira.spring.io/browse/INT-4173

Support Pagination for mongodb inbound adapter by accepting a SpEL expression that can be evaluated into a BasicQuery object , so `skip()` and `limit()` can be used.
Example:

`<int-mongodb:inbound-channel-adapter query-expression="new org.springframework.data.mongodb.core.query.BasicQuery('{''name'' : ''Bob''}').limit(2)" />` 

This PR targets 5.x version, original PR raised for 4.3.x:
https://github.com/spring-projects/spring-integration/pull/1988


artembilan:
Would you mind to explain why do we need this separate PR, if the functionality is fully similar to that one for `4.3.x`?

OTOH we really prefer PRs against the current `master` and only after that we cherry-pick changes to the previous versions.

So, maybe we can close one of them and finish the work in another and only after that decide if we need some critical backport changes?

Thanks

yaronyam:
@artembilan , sure https://github.com/spring-projects/spring-integration/pull/1988 can be closed. It was raised since I was under the assumption that it would be useful for merging to both branches.
I've addressed the latest CR comments on this PR, sorry for the mess with the duplicate PR's

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/inbound/MongoDbMessageSource.java
```diff
@@ -37,7 +38,7 @@
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 
-import com.mongodb.DBObject;
+import java.util.List;
```
artembilan: This is minor and I'll fix on merge, but the order of imports in Spring Integration is like: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-Framework-Code-Style#import-statements

artembilan (changes requested):

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/6a1a2ad3320c4d757bf7b5f5004405f641964fdc and cherry-picked.

@yaronyam , thank you for the contribution.
Looking forward for more!

PR #1997, opened 2016-12-08
Int 3335: Implement MongoDb Outbound Gateway

xpadro:
INT-3335: Implement MongoDb Outbound Gateway

JIRA: https://jira.spring.io/browse/INT-3335

### DIFF
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayParser.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayParser.java
@@ -50,7 +50,7 @@ public class MongoDbOutboundGatewayParser extends AbstractConsumerEndpointParser
 
 		BeanDefinition queryExpressionDef =
 				IntegrationNamespaceUtils.createExpressionDefinitionFromValueOrExpression("query", "query-expression",
-						parserContext, element, false);
+						parserContext, element, true);
 
 		if (queryExpressionDef != null) {
 			builder.addPropertyValue("queryExpression", queryExpressionDef);
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayParser.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayParser.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.mongodb.config;
+
+import org.w3c.dom.Element;
+
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.xml.ParserContext;
+import org.springframework.integration.config.xml.AbstractConsumerEndpointParser;
+import org.springframework.integration.config.xml.IntegrationNamespaceUtils;
+import org.springframework.integration.mongodb.outbound.MongoDbOutboundGateway;
+import org.springframework.util.StringUtils;
+
+/**
+ * Parser for MongoDb outbound gateways
+ *
+ * @author Xavier Padró
+ * @since 5.0
+ */
+public class MongoDbOutboundGatewayParser extends AbstractConsumerEndpointParser {
+
+	@Override
+	protected BeanDefinitionBuilder parseHandler(Element element, ParserContext parserContext) {
+		final BeanDefinitionBuilder builder =
+				BeanDefinitionBuilder.genericBeanDefinition(MongoDbOutboundGateway.class);
+
+		MongoParserUtils.processCommonAttributes(element, parserContext, builder);
+
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "reply-timeout");
+		String replyChannel = element.getAttribute("reply-channel");
+
+		if (StringUtils.hasText(replyChannel)) {
+			builder.addPropertyReference("outputChannel", replyChannel);
+		}
+
+		BeanDefinition queryExpressionDef =
+				IntegrationNamespaceUtils.createExpressionDefinitionFromValueOrExpression("query", "query-expression",
+						parserContext, element, false);
+
+		if (queryExpressionDef != null) {
+			builder.addPropertyValue("queryExpression", queryExpressionDef);
+		}
+
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "expect-single-result");
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "entity-class");
+
+		return builder;
+	}
+
+	@Override
+	protected String getInputChannelAttributeName() {
+		return "request-channel";
+	}
+}
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/dsl/MongoDbOutboundGatewaySpec.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/dsl/MongoDbOutboundGatewaySpec.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.mongodb.dsl;
+
+import java.util.function.Function;
+
+import org.springframework.data.mongodb.MongoDbFactory;
+import org.springframework.data.mongodb.core.MongoOperations;
+import org.springframework.data.mongodb.core.convert.MongoConverter;
+import org.springframework.data.mongodb.core.query.Query;
+import org.springframework.expression.Expression;
+import org.springframework.expression.common.LiteralExpression;
+import org.springframework.integration.dsl.MessageHandlerSpec;
+import org.springframework.integration.expression.FunctionExpression;
+import org.springframework.integration.mongodb.outbound.MongoDbOutboundGateway;
+import org.springframework.messaging.Message;
+
+/**
+ * A {@link MessageHandlerSpec} extension for the MongoDb Outbound endpoint {@link MongoDbOutboundGateway}
+ *
+ * @author Xavier Padró
+ * @since 5.0
+ */
+public class MongoDbOutboundGatewaySpec
+		extends MessageHandlerSpec<MongoDbOutboundGatewaySpec, MongoDbOutboundGateway> {
+
+	MongoDbOutboundGatewaySpec(MongoDbFactory mongoDbFactory, MongoConverter mongoConverter) {
+		this.target = new MongoDbOutboundGateway(mongoDbFactory, mongoConverter);
+		this.target.setRequiresReply(true);
+	}
+
+	MongoDbOutboundGatewaySpec(MongoOperations mongoTemplate) {
+		this.target = new MongoDbOutboundGateway(mongoTemplate);
+		this.target.setRequiresReply(true);
+	}
+
+	public MongoDbOutboundGatewaySpec expectSingleResult(boolean expectSingleResult) {
+		this.target.setExpectSingleResult(expectSingleResult);
+		return this;
+	}
+
+	public MongoDbOutboundGatewaySpec query(String query) {
+		this.target.setQueryExpression(new LiteralExpression(query));
+		return this;
+	}
+
+	public MongoDbOutboundGatewaySpec queryExpression(String queryExpression) {
+		this.target.setQueryExpression(PARSER.parseExpression(queryExpression));
+		return this;
+	}
+
+	public <P> MongoDbOutboundGatewaySpec queryFunction(Function<Message<P>, Query> queryFunction) {
+		this.target.setQueryExpression(new FunctionExpression<>(queryFunction));
+		return this;
+	}
+
+	public MongoDbOutboundGatewaySpec entityClass(Class<?> entityClass) {
+		this.target.setEntityClass(entityClass);
+		return this;
+	}
+
+	public MongoDbOutboundGatewaySpec collectionName(String collectionName) {
+		this.target.setCollectionNameExpression(new LiteralExpression(collectionName));
+		return this;
+	}
+
+	public MongoDbOutboundGatewaySpec collectionNameExpression(Expression collectionNameExpression) {
+		this.target.setCollectionNameExpression(collectionNameExpression);
+		return this;
+	}
+
+	public <P> MongoDbOutboundGatewaySpec collectionNameFunction(Function<Message<P>, String> collectionNameFunction) {
+		this.target.setCollectionNameExpression(new FunctionExpression<>(collectionNameFunction));
+		return this;
+	}
+}
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/dsl/MongoDbOutboundGatewaySpec.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/dsl/MongoDbOutboundGatewaySpec.java
@@ -58,7 +58,7 @@ public class MongoDbOutboundGatewaySpec
 	}
 
 	public MongoDbOutboundGatewaySpec queryExpression(String queryExpression) {
-		this.target.setQueryExpression(PARSER.parseExpression(queryExpression));
+		this.target.setQueryExpressionString(queryExpression);
 		return this;
 	}
 
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/dsl/MongoDbOutboundGatewaySpec.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/dsl/MongoDbOutboundGatewaySpec.java
@@ -22,7 +22,6 @@ import org.springframework.data.mongodb.MongoDbFactory;
 import org.springframework.data.mongodb.core.MongoOperations;
 import org.springframework.data.mongodb.core.convert.MongoConverter;
 import org.springframework.data.mongodb.core.query.Query;
-import org.springframework.expression.Expression;
 import org.springframework.expression.common.LiteralExpression;
 import org.springframework.integration.dsl.MessageHandlerSpec;
 import org.springframework.integration.expression.FunctionExpression;
@@ -78,8 +77,8 @@ public class MongoDbOutboundGatewaySpec
 		return this;
 	}
 
-	public MongoDbOutboundGatewaySpec collectionNameExpression(Expression collectionNameExpression) {
-		this.target.setCollectionNameExpression(collectionNameExpression);
+	public MongoDbOutboundGatewaySpec collectionNameExpression(String collectionNameExpression) {
+		this.target.setCollectionNameExpressionString(collectionNameExpression);
 		return this;
 	}
 
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
@@ -110,11 +110,11 @@ public class MongoDbOutboundGateway extends AbstractReplyProducingMessageHandler
 
 	@Override
 	protected void doInit() {
-		if (queryExpression == null) {
+		if (this.queryExpression == null) {
 			throw new IllegalStateException("no query specified");
 		}
 
-		if (collectionNameExpression == null) {
+		if (this.collectionNameExpression == null) {
 			throw new IllegalStateException("no collection name specified");
 		}
 
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.mongodb.outbound;
+
+import org.bson.Document;
+
+import org.springframework.data.mongodb.MongoDbFactory;
+import org.springframework.data.mongodb.core.MongoOperations;
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.convert.DefaultDbRefResolver;
+import org.springframework.data.mongodb.core.convert.MappingMongoConverter;
+import org.springframework.data.mongodb.core.convert.MongoConverter;
+import org.springframework.data.mongodb.core.mapping.MongoMappingContext;
+import org.springframework.data.mongodb.core.query.BasicQuery;
+import org.springframework.data.mongodb.core.query.Query;
+import org.springframework.expression.EvaluationContext;
+import org.springframework.expression.Expression;
+import org.springframework.expression.TypeLocator;
+import org.springframework.expression.spel.support.StandardTypeLocator;
+import org.springframework.integration.expression.ExpressionUtils;
+import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * Makes outbound operations to query a MongoDb database using a {@link MongoOperations}
+ *
+ * @author Xavier Padró
+ * @since 5.0
+ */
+public class MongoDbOutboundGateway extends AbstractReplyProducingMessageHandler {
+
+	private volatile EvaluationContext evaluationContext;
+
+	private volatile Expression queryExpression;
+
+	private volatile MongoOperations mongoTemplate;
+
+	private volatile MongoDbFactory mongoDbFactory;
+
+	private volatile MongoConverter mongoConverter;
+
+	private volatile boolean expectSingleResult = false;
+
+	private volatile Class<?> entityClass = Document.class;
+
+	private volatile Expression collectionNameExpression;
+
+	public MongoDbOutboundGateway(MongoDbFactory mongoDbFactory) {
+		this(mongoDbFactory, new MappingMongoConverter(new DefaultDbRefResolver(mongoDbFactory),
+				new MongoMappingContext()));
+	}
+
+	public MongoDbOutboundGateway(MongoDbFactory mongoDbFactory, MongoConverter mongoConverter) {
+		Assert.notNull(mongoDbFactory, "mongoDbFactory must not be null.");
+		Assert.notNull(mongoConverter, "mongoConverter must not be null.");
+		this.mongoDbFactory = mongoDbFactory;
+		this.mongoConverter = mongoConverter;
+	}
+
+	public MongoDbOutboundGateway(MongoOperations mongoTemplate) {
+		Assert.notNull(mongoTemplate, "mongoTemplate must not be null.");
+		this.mongoTemplate = mongoTemplate;
+	}
+
+	public void setQueryExpression(Expression queryExpression) {
+		this.queryExpression = queryExpression;
+	}
+
+	public void setExpectSingleResult(boolean expectSingleResult) {
+		this.expectSingleResult = expectSingleResult;
+	}
+
+	public void setEntityClass(Class<?> entityClass) {
+		Assert.notNull(entityClass, "entityClass must not be null.");
+		this.entityClass = entityClass;
+	}
+
+	public void setCollectionNameExpression(Expression collectionNameExpression) {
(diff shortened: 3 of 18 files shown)


artembilan:
Cool!

Will come back to you form Monday for review.

Thank you for contribution, @xpadro !

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/core/MongoDbExecutorTests.java
```diff
@@ -0,0 +1,254 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.integration.mongodb.core;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.springframework.integration.mongodb.test.MongoDbTestUtils.TEST_COLLECTION;
+import static org.springframework.integration.mongodb.test.MongoDbTestUtils.createPerson;
```
artembilan: No `static` imports, unless it is something like:
```
	<property name="excludes"
				value="org.assertj.core.api.Assertions.*,
				org.junit.Assert.*,
				org.junit.Assume.*,
				org.hamcrest.CoreMatchers.*,
				org.hamcrest.Matchers.*,
				org.mockito.Mockito.*,
				org.mockito.BDDMockito.*,
				org.mockito.Matchers.*,
				org.springframework.test.web.servlet.request.MockMvcRequestBuilders.*,
				org.springframework.test.web.servlet.result.MockMvcResultMatchers.*,
				org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.*,
				org.springframework.security.test.web.servlet.setup.SecurityMockMvcConfigurers.*" />
```

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/dsl/MongoDbTests.java
```diff
@@ -0,0 +1,261 @@
...
+import org.springframework.messaging.PollableChannel;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.springframework.integration.mongodb.test.MongoDbTestUtils.TEST_COLLECTION;
+import static org.springframework.integration.mongodb.test.MongoDbTestUtils.createPerson;
```
artembilan: `static` imports first.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/dsl/MongoDbTests.java
```diff
@@ -0,0 +1,261 @@
...
+
+		@Bean
+		public MongoOperations mongoTemplate() {
+			return new MongoTemplate(mongoDbFactory());
+		}
+
+		private MongoDbOutboundGatewaySpec queryOutboundGateway(String query, boolean expectSingleResult) {
+			return MongoDb.outboundGateway(mongoDbFactory())
+					.query(query)
+					.collectionNameExpression(new LiteralExpression(TEST_COLLECTION))
```
artembilan: We should introduce simple `.collectionName(String collectionName)` option.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/dsl/MongoDbTests.java
```diff
@@ -0,0 +1,261 @@
...
+					.entityClass(PersonDomain.class);
+		}
+
+		private MongoDbOutboundGatewaySpec queryExpressionOutboundGateway(boolean expectSingleResult, int maxResults) {
+			return MongoDb.outboundGateway(mongoDbFactory())
+					.queryExpression("payload.query")
+					.collectionNameExpression(new LiteralExpression(TEST_COLLECTION))
+					.expectSingleResult(expectSingleResult)
+					.entityClass(PersonDomain.class)
+					.maxResults(maxResults);
```
artembilan: When we introduce `BasicQuery` support for the `query-expression` (as well as separate `.query(Query)` option for DSL), we don't need this `maxResults()` any more.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/test/entity/QueryMessage.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.integration.mongodb.test.entity;
+
+/**
+ * @author Xavier Padró
+ * @since 5.0
+ */
+public class QueryMessage {
```
artembilan: M-m-m. Seems for me we can live with this class. 
If you would like to check if expression works well, maybe try to perform it against some header, e.g.:
```
.queryExpression("headers.query")
```

artembilan (changes requested):
Also some general suggestion.
I would like to see something like `setCollectionCallback(CollectionCallback)` to perform any arbitrary operation, not only `query`. I thought about `update` as well, like we have in `JdbcOutboundGateway`, but that leads to some other modification options like `FindAndModifyOptions` etc. So, I think a common `CollectionCallback` would be sufficient. Although we can address that in the separate PR.
Also there is need for XML configuration support for this new `<int-mongo:outbound-gateway>` component.

In the end we also need documentation on the matter.

Thank you for the contribution anyway, @xpadro !

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/dsl/MongoDbTests.java
```diff
@@ -0,0 +1,261 @@
...
+import static org.springframework.integration.mongodb.test.MongoDbTestUtils.TEST_COLLECTION;
+import static org.springframework.integration.mongodb.test.MongoDbTestUtils.createPerson;
+
+/**
+ * @author Xavier Padró
+ * @since 5.0
+ */
+@RunWith(SpringRunner.class)
+@DirtiesContext
+public class MongoDbTests {
```
artembilan: The problem with all of these real MongoDB tests that the target environment may not have `mongod` process. Therefore we always use `MongoDbAvailableTests` which includes `MongoDbAvailableRule` JUnit `@Rule`.

From there pay attention that we have `createPerson()` methods, so maybe your new `MongoDbTestUtils` and `PersonDomain` are redundant, too? I mean that when we have these new tests based on the `MongoDbAvailableTests`, we don't need any new testing entities.

xpadro:
Hi Artem, I pushed again with following updates:

- Added XML configuration support
- Fixed all requested changes

artembilan:
@xpadro , thank you for the update!

The TCP tests problem has been fixed here https://github.com/spring-projects/spring-integration/commit/44ad81114c6f3fd0ecdb8c73f1ed5b25e0479c41. So, you should `rebase` your PR to the latest `master`.

Will come back to you for review later.

xpadro:
I Merged with master since I read in the contribution guide that I shouldn't rebase on an already pushed branch. I hope this is right. 

I also updated the PR in order to include documentation for the outbound gateway.

artembilan:
@xpadro ,
Not sure what you mean about you shouldn't rebase, but here what is said in our Wiki: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc#keeping-your-local-code-in-sync:

> Rebase the topic branch to minimize the distance between it and your recently synched master branch: git rebase master 

So, it really would be better to `rebase` to minimize the count of commits in your branch.
We perform `merge` when move a top-level branches changes to the `master`.

Will come back to you soon after other critical bug fixes reviews.

Thank you anyway for update!

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayParserTests.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * @author Xavier Padró
+ * @since 5.0
+ */
+public class MongoDbOutboundGatewayParserTests {
+
+	@Test
+	public void minimalConfig() {
+		ClassPathXmlApplicationContext context =
+				new ClassPathXmlApplicationContext("outbound-gateway-parser-config.xml", this.getClass());
```
artembilan: I think we can just go with the 
```
@ContextConfiguration
@RunWith(SpringRunner.class)
@DirtiesContext
```
and avoid extra loads for the the same config in each test method.
I that case the `outbound-gateway-parser-config.xml` should be renamed to the `MongoDbOutboundGatewayParserTests-context.xml`

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayXmlTests.java
```diff
@@ -0,0 +1,182 @@
...
+
+		mongoTemplate.dropCollection(COLLECTION_NAME);
+	}
+
+
+	@Test
+	@MongoDbAvailable
+	public void testSingleQuery() throws Exception {
+		ClassPathXmlApplicationContext context =
+				new ClassPathXmlApplicationContext("outbound-gateway-config.xml", this.getClass());
```
artembilan: The same Spring Test Framework usage concern is here as well.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,201 @@
...
+ * Makes outbound operations to query a MongoDb database using a {@link MongoOperations}
+ *
+ * @author Xavier Padró
+ * @since 5.0
+ */
+public class MongoDbOutboundGateway extends AbstractReplyProducingMessageHandler {
+
+	private volatile boolean initialized = false;
+
+	private static final String DEFAULT_COLLECTION_NAME = "data";
```
artembilan: I think this kind of "magic" word for default does not make sense in the real word.
Seems for me would be much easier just to require `collectionName` from the end user.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,201 @@
...
+
+	private volatile MongoOperations mongoTemplate;
+
+	private volatile MongoDbFactory mongoDbFactory;
+
+	private volatile MongoConverter mongoConverter;
+
+	private volatile boolean expectSingleResult = false;
+
+	private volatile Class<?> entityClass = DBObject.class;
```
artembilan: I can't right now "why?" but looks like since the move to the MongoDB 3.x it would be better to make default to the `org.bson.Document` . Fix the same in the XSD attribute description.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayXmlTests.java
```diff
@@ -0,0 +1,182 @@
...
+
+		EventDrivenConsumer consumer = context.getBean("gatewaySingleQuery", EventDrivenConsumer.class);
+		PollableChannel outChannel = context.getBean("out", PollableChannel.class);
+
+		Message<String> message = MessageBuilder.withPayload("").build();
+		consumer.getHandler().handleMessage(message);
+
+		Message<?> result = outChannel.receive(10000);
```
artembilan: M-m-m. Having such a hard logic I'd say this test class belongs to the `outbound`, not `config`. 
Because here is no any config related tests.

src/reference/asciidoc/mongodb.adoc
```diff
@@ -290,3 +290,116 @@ and other attributes that are common across all other inbound adapters (e.g., 'c
...
+    request-channel="in"
+    reply-channel="out"
+    entity-class="org.springframework.integration.mongodb.test.entity.Person"/>
+----
+
+* `collection-name` or `collection-name-expression` - Identifies the name of the MongoDb collection to use.
+* `mongo-converter` - reference to an instance of `o.s.data.mongodb.core.convert.MongoConverter` to assist with converting a raw java object to a JSON document representation
+* `mongodb-factory` - reference to an instance of `o.s.data.mongodb.MongoDbFactory`
+* `mongo-template` - reference to an instance of `o.s.data.mongodb.core.MongoTemplate` (NOTE: you can not have both mongo-template and mongodb-factory set)
+* `entity-class` - The fully qualified name of the entity class to to be passed to find(..) or findOne(..) method in MongoTemplate. If this attribute is not provided the default value is com.mongodb.DBObject
```
artembilan: Double `to`. 

`find(..)` and so should be wrapped to code style.
And the same for the `DBObject` although we decided with you that `Document` should be there now.

Plus take a look into other chapters (e.g. `amqp.adoc`) how call-outs are done - to have cross-links from the code snippet to the description and back.

src/reference/asciidoc/mongodb.adoc
```diff
@@ -290,3 +290,116 @@ and other attributes that are common across all other inbound adapters (e.g., 'c
 The example above is relatively simple and static since it has a literal value for the `collection-name`.
 Sometimes you may need to change this value at runtime based on some condition.
 To do that, simply use `collection-name-expression` where the provided expression can be any valid SpEL expression.
+
+=== MongoDB Outbound Gateway
+
+The MongoDb Outbound Gateway allows you to query a database by sending a Message to its request channel.
```
artembilan: Would be better to say since which version it is available.

Plus the great addition would be in the  `whats-new.adoc` with the link to this new paragraph.

artembilan (changes requested):
Great work, @xpadro !

Even if I don't agree in some places that doesn't mean that you do wrong things 😄 

One more time about my concerns commented here: https://github.com/spring-projects/spring-integration/pull/1997#pullrequestreview-12515917

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,201 @@
...
+	public void setCollectionNameExpression(Expression collectionNameExpression) {
+		Assert.notNull(collectionNameExpression, "collectionNameExpression cannot be null");
+		this.collectionNameExpression = collectionNameExpression;
+	}
+
+	public void setMongoConverter(MongoConverter mongoConverter) {
+		Assert.notNull(mongoConverter, "mongoConverter cannot be null");
+		Assert.isNull(this.mongoTemplate,
+				"'mongoConverter' can not be set when instance was constructed with MongoTemplate");
+		this.mongoConverter = mongoConverter;
```
xpadro: I will change it. However the gateway xml parser `MongoDbOutboundGatewayParser` is invoking this method:

`MongoParserUtils.processCommonAttributes(element, parserContext, builder);`

which expects the `MongoConverter` to be in a setter:

`IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element, "mongo-converter");`

Should I change this generic class or use something like `MongoGatewayParserUtils` in my parser where I will expect `MongoConverter` as a constructor argument?

artembilan: OK. I got!
Sad, of course, but we can live with that.
Let's then leave this part as is - a setter for the `MongoConverter`!
We can reconsider refactor in the separate task. Luckily we are in the major version, so we can provide breaking changes. I mean move that `MongoConverter` to the ctor argument for all components and, therefore, change the common `MongoParserUtils.processCommonAttributes()` to populate ctor arg instead of setter's value.

Everything rest still looks like in a power. 
Correct me if you have some doubts.

Thanks

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/dsl/MongoDbOutboundGatewaySpec.java
```diff
@@ -0,0 +1,92 @@
...
+		this.mongoDbOutboundGateway.setExpectSingleResult(expectSingleResult);
+		return this;
+	}
+
+	public MongoDbOutboundGatewaySpec query(String query) {
+		this.mongoDbOutboundGateway.setMongoDbQuery(query);
+		return this;
+	}
+
+	public MongoDbOutboundGatewaySpec queryExpression(String queryExpression) {
```
xpadro: I implemented this for DSL configuration.

However, should we also provide XML support for query function? I made some tests and I don't seem to figure out how the `MongoDbOutboundGatewayParser` should parse the XML property. 

Something like this is not supported by the SpEL parser:

`query-function="msg -> new BasicQuery('{''name'' : msg.payload}')"`

This should be mapped to the `Function<Message<?>, Query> queryFunction` property in the outbound gateway.

xpadro:
Updated PR with the latest fixes. I also included `Function` for both `query` and `collectionName`.

The last pending change is `CollectionCallback`, but as you said, we can leave this for a different PR.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/config/MongoDbOutboundGatewayParser.java
```diff
@@ -0,0 +1,69 @@
...
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "reply-timeout");
+		String replyChannel = element.getAttribute("reply-channel");
+
+		if (StringUtils.hasText(replyChannel)) {
+			builder.addPropertyReference("outputChannel", replyChannel);
+		}
+
+		BeanDefinition queryExpressionDef =
+				IntegrationNamespaceUtils.createExpressionDefinitionFromValueOrExpression("query", "query-expression",
+						parserContext, element, false);
```
artembilan: @xpadro ,
Thank you for the update!

But the review is still here 😄 

I think this is not true.
We definitely need to have `queryExpression`. 
So, the last `oneRequired` should be `true` for this `createExpressionDefinitionFromValueOrExpression()` invocation.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	public MongoDbOutboundGateway(MongoOperations mongoTemplate) {
+		Assert.notNull(mongoTemplate, "mongoTemplate must not be null.");
+		this.mongoTemplate = mongoTemplate;
+	}
+
+	public void setQueryExpression(Expression queryExpression) {
+		this.queryExpression = queryExpression;
```
artembilan: `Assert.notNull()` - just for fail fast, since we really don't support `null` for this property.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,159 @@
...
+		this.expectSingleResult = expectSingleResult;
+	}
+
+	public void setEntityClass(Class<?> entityClass) {
+		Assert.notNull(entityClass, "entityClass must not be null.");
+		this.entityClass = entityClass;
+	}
+
+	public void setCollectionNameExpression(Expression collectionNameExpression) {
+		this.collectionNameExpression = collectionNameExpression;
```
artembilan: `Assert.notNull()` as well

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,159 @@
...
+		Assert.notNull(mongoConverter, "mongoConverter cannot be null");
+		Assert.isNull(this.mongoTemplate,
+				"'mongoConverter' can not be set when instance was constructed with MongoTemplate");
+		this.mongoConverter = mongoConverter;
+	}
+
+	@Override
+	protected void doInit() {
+		Assert.isTrue(this.queryExpression != null, "no query specified");
+		Assert.isTrue(this.collectionNameExpression != null, "no collection name specified");
```
artembilan: Well, not `isTrue()` - `state()`. We are here not in the argument any more. So, we are indicating an `IllegalState` really.
See their JavaDocs.
Right, nit-pick, but still a pursuit for clean perfect code 😄

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/dsl/MongoDbTests.java
```diff
@@ -0,0 +1,351 @@
...
+				.build());
+
+		Message<?> result = this.getResultChannel.receive(10_000);
+
+		assertNotNull(result);
+		Person retrievedPerson = (Person) result.getPayload();
+		assertEquals("Artem", retrievedPerson.getName());
+	}
+
+	@Test(expected = ReplyRequiredException.class)
```
artembilan: M-m-m, do we expose `requires-reply` attribute to the XML config?
```
<xsd:attribute name="requires-reply" type="xsd:string" use="optional" default="true">
	<xsd:annotation>
		<xsd:documentation>
			Specify whether this outbound gateway must return a non-null value. This value is
			'true' by default, and a ReplyRequiredException will be thrown when
			the underlying service returns a null value.
		</xsd:documentation>
	</xsd:annotation>
</xsd:attribute>
```
The Java DSL should be fine for that: `ConsumerEndpointSpec.requiresReply()`

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGatewayTests.java
```diff
@@ -0,0 +1,305 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+/**
+ * @author Xavier Padró
+ * @since 5.0
+ */
+@ContextConfiguration
+@RunWith(SpringJUnit4ClassRunner.class)
+public class MongoDbOutboundGatewayTests extends MongoDbAvailableTests {
```
artembilan: `@DirtiesContext`, too.
Just in case 😄

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGatewayTests.java
```diff
@@ -0,0 +1,305 @@
...
+
+	@Autowired
+	private MongoDbFactory mongoDbFactory;
+
+	@Before
+	public void setUp() {
+		this.mongoTemplate.save(this.createPerson("Artem"), COLLECTION_NAME);
+		this.mongoTemplate.save(this.createPerson("Gary"), COLLECTION_NAME);
+		this.mongoTemplate.save(this.createPerson("Oleg"), COLLECTION_NAME);
+		this.mongoTemplate.save(this.createPerson("Xavi"), COLLECTION_NAME);
```
artembilan: `BulkOperations.insert(List<? extends Object> documents)` ?
Just for the knowledges extension.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGatewayTests.java
```diff
@@ -0,0 +1,305 @@
...
+		gateway.afterPropertiesSet();
+
+		Object result = gateway.handleRequestMessage(message);
+
+		Person person = (Person) result;
+
+		assertEquals("Xavi", person.getName());
+	}
+
+	@Test(expected = IllegalArgumentException.class)
```
artembilan: I would like to know the reason of this exception.
Therefore like in other tests - `try...catch` or consider to use `ExpectedException` `@Rule`

src/reference/asciidoc/mongodb.adoc
```diff
@@ -290,3 +290,117 @@ and other attributes that are common across all other inbound adapters (e.g., 'c
 The example above is relatively simple and static since it has a literal value for the `collection-name`.
 Sometimes you may need to change this value at runtime based on some condition.
 To do that, simply use `collection-name-expression` where the provided expression can be any valid SpEL expression.
+
+[[mongodb-outbound-gateway]]
+=== MongoDB Outbound Gateway
+
+Starting with _version 5.0_, the MongoDb Outbound Gateway allows you to query a database by sending a Message to its request channel.
```
artembilan: Something wrong with this sentence.
Maybe like this?
```
Starting with _version 5.0_, the MongoDb Outbound Gateway is provided and it allows...
```

src/reference/asciidoc/mongodb.adoc
```diff
@@ -290,3 +290,117 @@ and other attributes that are common across all other inbound adapters (e.g., 'c
...
+The gateway will then send the response to the reply channel.
+The Message payload and headers can be used to specify the query.
```
artembilan: ... the query, as well as collection name.

src/reference/asciidoc/mongodb.adoc
```diff
@@ -290,3 +290,117 @@ and other attributes that are common across all other inbound adapters (e.g., 'c
...
+    @Bean
+    public MessageChannel replyChannel() {
+        return new QueueChannel(5);
+    }
+
+    @Bean
+    @ServiceActivator(inputChannel = "requestChannel")
+    public MessageHandler mongoDbOutboundGateway() {
+        MongoDbOutboundGateway gateway = new MongoDbOutboundGateway(this.mongoDbFactory);
+        gateway.setCollectionNameExpression(new LiteralExpression("foo"));
```
artembilan: Yeah... That's why in other places we have `set...ExpressionString(String expression)` variant and parse it directly in the component. For example `AbstractAmqpOutboundEndpoint.setExchangeNameExpressionString()`

src/reference/asciidoc/mongodb.adoc
```diff
@@ -290,3 +290,117 @@ and other attributes that are common across all other inbound adapters (e.g., 'c
...
+    public IntegrationFlow gatewaySingleQueryFlow() {
+        return f -> f
+                .handle(queryOutboundGateway())
+                .channel(c -> c.queue("retrieveResults"));
+    }
+
+    private MongoDbOutboundGatewaySpec queryOutboundGateway() {
+        return MongoDb.outboundGateway(this.mongoDbFactory, this.mongoConverter)
+                .query("{name : 'Bob'}")
+                .collectionNameExpression(new LiteralExpression("foo"))
```
artembilan: We talked before that Java DSL should be free from `Expression` API.
Therefore that `collectionNameExpression()` should be with the `String` instead.

artembilan (changes requested):
That's all for now.

Please, consider do not `squash` commits in the future.
That helps us to track changes.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/dsl/MongoDbTests.java
xpadro: Added attribute to the XML config

xpadro:
Updated PR with all fixes. It seems build failed on curl command. Could it be relaunched?

Sorry about the squash!

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,171 @@
...
+		Assert.notNull(mongoConverter, "mongoConverter cannot be null");
+		Assert.isNull(this.mongoTemplate,
+				"'mongoConverter' can not be set when instance was constructed with MongoTemplate");
+		this.mongoConverter = mongoConverter;
+	}
+
+	@Override
+	protected void doInit() {
+		if (this.queryExpression == null) {
+			throw new IllegalStateException("no query specified");
```
artembilan: @xpadro ,

Thank you for so prompt input! 
Well, looks like Travis is dead. Not our business. At least for now. 😄 

M-m-m.
No, I meant `Assert.state()`. Why didn't it work for you ?

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,171 @@
...
+				this.queryExpression.getValue(this.evaluationContext, requestMessage, Object.class);
+
+		if (expressionValue instanceof String) {
+			query = new BasicQuery((String) expressionValue);
+		}
+		else if (expressionValue instanceof Query) {
+			query = ((Query) expressionValue);
+		}
+		else {
+			throw new IllegalStateException("'queryExpression' must evaluate to String or org.springframework.data.mongodb.core.query.Query");
```
artembilan: Try to keep code line length as 120.
Yes, minor and I can fix on merge.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -0,0 +1,171 @@
...
+		Assert.notNull(entityClass, "entityClass must not be null.");
+		this.entityClass = entityClass;
+	}
+
+	public void setCollectionNameExpression(Expression collectionNameExpression) {
+		Assert.notNull(collectionNameExpression, "collectionNameExpression must not be null.");
+		this.collectionNameExpression = collectionNameExpression;
+	}
+
+	public void setCollectionNameExpressionString(String collectionNameExpressionString) {
```
artembilan: Would be great to have similar `String` setter for the `query` as well

artembilan (changes requested):

spring-integration-mongodb/src/main/resources/org/springframework/integration/mongodb/config/spring-integration-mongodb-5.0.xsd
```diff
@@ -102,6 +102,140 @@
...
+								The "reply-timeout" attribute maps to the "sendTimeout" property of the
+								underlying 'MessagingTemplate' instance (org.springframework.integration.core.MessagingTemplate).
+
+								The attribute will default, if not specified, to '-1', meaning that
+								by default, the Gateway will wait indefinitely. The value is
+								specified in milliseconds.
+							]]></xsd:documentation>
+						</xsd:annotation>
+					</xsd:attribute>
+					<xsd:attribute name="requires-reply" type="xsd:string" use="optional" default="true">
```
artembilan: Let's add `<xsd:union memberTypes="xsd:boolean xsd:string" />` to this one as well!

artembilan (changes requested):
@xpadro ,

you know, that's all.
There are some other minor issue in the Docs, but I think it fine to be merged and polished silently.

So, let me know if you are fine, that I will finish it for you.
And will start to think how to add `CollectionCallback<T>` support 😄

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
xpadro: Yes it works. I misunderstood your comment :)

xpadro:
Updated again the PR.

Yes sure, I am fine with that! :)

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/0fa8849f7f44f70ac84a36153351a7a45ccc6f72.

@xpadro ,
thank you very much for the contribution!

I really looking forward for evolution for this MongoDB Java DSL 😄 !
Plus don't forget our deal about `CollectionCollback<?>` 😉

xpadro:
Glad to contribute :)

Should I create a new JIRA issue for the `CollectionCallback` or just a new pull request?

artembilan:
Sure think, @xpadro .
Let it be from JIRA first!

Thanks you!

PR #2000, opened 2016-12-13
INT-4182: Use thread-safe `DateTimeFormatter`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4182

To avoid extra allocation for the `SimpleDateFormat` use a new Java 8 `DateTimeFormatter` based on new `Temporal` abstraction

Leave test-cases as is to save some time for other tasks


garyrussell:
Merged as 3d433ca3f8321ed112673d057292a4e812a7eff8

PR #2001, opened 2016-12-14
Modified IntegrationResourceHolder to support list of messages

bojanv55:
Since in IntegrationResourceHolder we could store only one message, this is back-compatible fix for support of collection of messages.


artembilan:
Sorry, we can't apply this change.

The `IntegrationResourceHolder` is aimed to be use from the `source` side, when we really produce only one `message` per transaction.

That is just our trick with you on SO (http://stackoverflow.com/questions/41104769/spring-integration-transaction-commit-before-rest-of-pipeline) to minimize custom code as possible.

If you talk about `splitter` and its list of messages within the same transaction, you can consider to use `IntegrationResourceHolder` per message anyway. Or you can switch to the `IntegrationResourceHolder.addAttribute()` for list of messages if that is your use-case.

But anyway there is no any common approach to be reused from the Framework perspective.
The main purpose of Spring Integration component is **loose coupling** when one message doesn't affect all others.

Of course, on the other hand, you can use this variant in your own application. But again: that doesn't fit for Framework.

artembilan:
Closed as `Invalid`.

Feel free to use the solution in your application anyway!

Merry Christmas and we are always welcome for contribution and discussions

PR #2003, opened 2016-12-14
Fix race conditions in TcpNioConnectionReadTests

artembilan:
https://build.spring.io/browse/INT-SI43X-60

After introduction the `sendErrorToListener()` function for the `TcpConnectionSupport`, all the tests in the `TcpNioConnectionReadTests` have flaw to be affected by the race condition to release semaphore in the `onMessage()` and don't get a connection as closed yet.

* Fix the tests adding `CountDownLatch` for the `ErrorMessage` and `AtomicReference` to assert an Exception message afterwards.

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionReadTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionReadTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionReadTests.java
```diff
@@ -411,8 +474,15 @@ private void testClosureMidMessageGuts(AbstractByteArraySerializer serializer, S
 		final Semaphore semaphore = new Semaphore(0);
 		final List<TcpConnection> added = new ArrayList<TcpConnection>();
 		final List<TcpConnection> removed = new ArrayList<TcpConnection>();
+
+		final CountDownLatch errorMessageLetch = new CountDownLatch(1);
+		final AtomicReference<Throwable> errorMessageRef = new AtomicReference<Throwable>();
+
 		AbstractServerConnectionFactory scf = getConnectionFactory(serializer, message -> {
-			responses.add(message);
```
garyrussell: Looks like we should have removed this variable - noticed during conflict resolution during cherry-pick.

garyrussell:
Merged as a0845b1caf892afd53c4419167539d1b007383e2 and cherry-picked.

PR #2004, opened 2016-12-15
INT-4185: Fix FileReadingMessageSource for Java 6

artembilan:
JIRA: https://jira.spring.io/browse/INT-4185

Even if `WatchEventType` doesn't expose Java 7 API, its instantiation leads to the `StandardWatchEventKinds` object which already isn't Java 6

* Move `this.watchEvents = new WatchEventType[] { WatchEventType.CREATE }` to the ctor and guard it with the `ClassUtils.isPresent("java.nio.file.WatchService")`

**Tested on Java 6 locally**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
@@ -270,13 +270,16 @@ public class FileReadingMessageSource extends IntegrationObjectSupport implement
 
 	/**
 	 * Switch this {@link FileReadingMessageSource} to use its internal
-	 * {@link FileReadingMessageSource.WatchServiceDirectoryScanner}.
+	 * {@link FileReadingMessageSource.WatchServiceDirectoryScanner} based on the Java 7 {@link WatchService}.
 	 * @param useWatchService the {@code boolean} flag to switch to
 	 * {@link FileReadingMessageSource.WatchServiceDirectoryScanner} on {@code true}.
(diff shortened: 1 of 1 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
```diff
@@ -161,6 +165,9 @@ public FileReadingMessageSource(int internalQueueCapacity) {
 	public FileReadingMessageSource(Comparator<File> receptionOrderComparator) {
 		this.toBeReceived = new PriorityBlockingQueue<File>(
 				DEFAULT_INTERNAL_QUEUE_CAPACITY, receptionOrderComparator);
+		if (WATCH_SERVICE_PRESENT) {
+			this.watchEvents = new WatchEventType[] { WatchEventType.CREATE };
+		}
```
artembilan: Yeah! Agree about `Assert.state()` in the `setUseWatchService()`, but not sure what that event should mean...

garyrussell: I mean add an assertion in `setWatchEvents()` too.

garyrussell:
Looks like you missed my request to force `useWatchService` to `false` for Java 6.

If you agree, I can fix on merge.

garyrussell:
Oops - sorry - it's false by default 😄

artembilan:
>  it's false by default 

Eh... That was my point 😉

garyrussell:
Merged as 6a51da65e696d102b3bb2e347362b1b01603a0c5

PR #2006, opened 2016-12-16
INT-4187 Add an option to skip StdErr process if it is not needed

shahbour:
based on INT-4187 and my question in [stack overflow ](http://stackoverflow.com/questions/41075112/file-tail-inbound-channel-adapter-stderr-and-stdout) . I did name the new field enableStatusReader per function name .

This is my first contribution so if i did any thing wrong please tell me

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileTailInboundChannelAdapterFactoryBean.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileTailInboundChannelAdapterFactoryBean.java
@@ -35,6 +35,7 @@ import org.springframework.util.StringUtils;
 /**
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Ali Shahbour
  * @since 3.0
  *
  */
@@ -42,6 +43,8 @@ public class FileTailInboundChannelAdapterFactoryBean extends AbstractFactoryBea
 		implements BeanNameAware, SmartLifecycle, ApplicationEventPublisherAware {
 
 	private volatile String nativeOptions;
+
+	private volatile boolean enableStatusReader = true;
 
 	private volatile File file;
 
@@ -75,6 +78,10 @@ public class FileTailInboundChannelAdapterFactoryBean extends AbstractFactoryBea
 		if (StringUtils.hasText(nativeOptions)) {
 			this.nativeOptions = nativeOptions;
 		}
+	}
+
+	public void setEnableStatusReader(boolean enableStatusReader) {
+		this.enableStatusReader = enableStatusReader;
 	}
 
 	public void setFile(File file) {
@@ -180,6 +187,7 @@ public class FileTailInboundChannelAdapterFactoryBean extends AbstractFactoryBea
 		FileTailingMessageProducerSupport adapter;
 		if (this.delay == null && this.end == null && this.reopen == null) {
 			adapter = new OSDelegatingFileTailingMessageProducer();
+			((OSDelegatingFileTailingMessageProducer) adapter).setEnableStatusReader(this.enableStatusReader);
 			if (this.nativeOptions != null) {
 				((OSDelegatingFileTailingMessageProducer) adapter).setOptions(this.nativeOptions);
 			}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileTailInboundChannelAdapterFactoryBean.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileTailInboundChannelAdapterFactoryBean.java
@@ -84,6 +84,7 @@ public class FileTailInboundChannelAdapterFactoryBean extends AbstractFactoryBea
 	 * If false, thread for capturing stderr will not be started
 	 * and stderr output will be ignored
 	 * @param enableStatusReader true or false
+	 * @since 4.3.6
 	 */
 	public void setEnableStatusReader(boolean enableStatusReader) {
 		this.enableStatusReader = enableStatusReader;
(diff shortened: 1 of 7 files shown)


shahbour:
i did not change any thing in spring-integration-scripting but it failed

> Execution failed for task ':spring-integration-scripting:test'.

garyrussell:
@shahbour Looks good; thanks for the contribution!

We should add javadocs to the new setter methods; we also need to confirm the parser change with an update to `FileTailInboundChannelAdapterParserTests` to verify the boolean is reset when so configured.

If you don't have time to address these; let us know and we'll take of it during the merge.

shahbour:
It's ok I'll do that

Thanks

shahbour:
@garyrussell i did the updated , but i am not sure this is what you mean in tests , if you don't mind elaborate more

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileTailInboundChannelAdapterFactoryBean.java
```diff
@@ -77,6 +80,15 @@ public void setNativeOptions(String nativeOptions) {
 		}
 	}
 
+	/**
+	 * If false, thread for capturing stderr will not be started
+	 * and stderr output will be ignored
+	 * @param enableStatusReader true or false
```
artembilan: Add `@since 5.0`.
Although if you decided with @garyrussell to backport the fix to `4.3.x`, the version must be `4.3.6`

garyrussell:
Thanks, it's close, but that's just testing the default value (true) - i.e. tests that the parser doesn't change the default when not specified, which is a good test; but we also need a test for explicitly setting it to false in the test case's XML config and then verify it's false in the bean.

spring-integration-file/src/test/java/org/springframework/integration/file/config/FileTailInboundChannelAdapterParserTests.java
```diff
@@ -95,6 +95,7 @@ public void testNative() {
 		assertSame(exec, TestUtils.getPropertyValue(nativeAdapter, "taskExecutor"));
 		assertSame(sched, TestUtils.getPropertyValue(nativeAdapter, "taskScheduler"));
 		assertTrue(TestUtils.getPropertyValue(nativeAdapter, "autoStartup", Boolean.class));
+		assertTrue(TestUtils.getPropertyValue(nativeAdapter, "enableStatusReader" ,Boolean.class));
```
artembilan: OK. 
Now let's make make a test for `assertFalse()`.
I mean we should be sure that changes to the parser reflects to the target `OSDelegatingFileTailingMessageProducer`.

What I want to say that it isn't possible now.
You should add `enable-status-reader` XSD attribute and modify `FileTailInboundChannelAdapterParserTests-context.xml` with `enable-status-reader="false"`

artembilan (changes requested):
Would be great to have some note in the Reference Manual as well.
Thank you for the prompt contribution!

artembilan:
Also pay attention to the CheckStyle failures:
```
FileTailInboundChannelAdapterParserTests.java:98:90: ',' is preceded with whitespace. [NoWhitespaceBefore]

[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/test/java/org/springframework/integration/file/config/FileTailInboundChannelAdapterParserTests.java:98:92: ',' is not followed by whitespace. [WhitespaceAfter]
```
For this purpose you can locally run `check` Gradle command instead of just `test`.

garyrussell:
I think this can be backported so `@since 4.3.6` is correct.

Also note the checkstyle failure in travis.

You can run checkstyle locally with 

    ./gradlew :spring-integration-file:check

shahbour:
Thanks guys for the support, as I am new to tests it might take a while
with me , but I am looking forward to do it.

artembilan:
@shahbour , please, don't hesitate to ask questions on the matter.

Also, please, add `@DirtiesContex` for the `FileTailInboundChannelAdapterParserTests`.
There are several adapters with the `auto-startup="true"` meaning that they start with the context and spawn threads. By default `@RunWith(SpringJUnit4ClassRunner.class)` doesn't close application context for caching reasons in between test classes. In Spring Integration we don't need that feature, but we should be sure that all processes are destroyed after test class. Therefore `@DirtiesContex`. See it's JavaDoc.

shahbour:
how can i add note to the Reference Manual

src/reference/asciidoc/file.adoc
```diff
@@ -448,6 +448,18 @@ If the tail command fails (on some platforms, a missing file causes the `tail` t
...
 ----
+<int-file:tail-inbound-channel-adapter id="native"
+	channel="input"
+	enable-status-reader="false"
+	task-executor="exec"
+	file="/tmp/foo"/>
+----
+
+By default native adapter capture from standard output and send them as messages and from standard error to raise events.
+You can discard the standard error events by setting enable-status-reader to false.
```
artembilan: Need to wrap attribute to the code formatting, but that is minor. Will fix on merge.
LGTM

Also need `TailAdapterSpec` modification for Java DSL.
Plus I don't see any sample in the doc.
All this work is for another pull request.

Feel free to consider that part as a subsequent contribution as well, @shahbour 😄

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/16c7535577a1cd38681188b0194ad2d8c8cbb684 and cherry-picked.

@shahbour , 
thank you for contribution.
Looking forward for more!

PR #2008, opened 2016-12-22
INT-4188: Add Idle Event Interval Support

shahbour:
JIRA: https://jira.springsource.org/browse/INT-4188

    * add `<idle-event-interval>` XSD element
    * add `FileTailingIdleEvent`
    * move `TaskScheduler` and `getRequiredTaskScheduler` to FileTailingMessageProducerSupport
    * add `setIdleEventInterval`

I only exposed one attribute and it is used for both scheduling the check and checking the idle time , as a result it may take up to 2X the idle time to be fired .

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
@@ -169,7 +169,7 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 
 	private void publishIdleEvent(long idleTime) {
 
-		if (this.eventPublisher != null) {
+		if (this.eventPublisher != null && getFile().exists()) {
 			FileTailingIdleEvent event = new FileTailingIdleEvent(this, this.file, idleTime);
 			this.eventPublisher.publishEvent(event);
 		}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
@@ -28,7 +28,6 @@ import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.file.event.FileIntegrationEvent;
 import org.springframework.messaging.Message;
-import org.springframework.scheduling.TaskScheduler;
 import org.springframework.util.Assert;
 
 /**
@@ -108,11 +107,6 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 	}
 
 	@Override
-	public void setTaskScheduler(TaskScheduler taskScheduler) {
-		super.setTaskScheduler(taskScheduler);
-	}
-
-	@Override
 	public String getComponentType() {
 		return "file:tail-inbound-channel-adapter";
 	}
@@ -144,11 +138,11 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 		if (this.idleEventInterval > 0) {
 			this.idleEventScheduledFuture = getTaskScheduler().scheduleWithFixedDelay(() -> {
 				long now = System.currentTimeMillis();
-				long lastAlertAt = FileTailingMessageProducerSupport.this.lastNoMessageAlert.get();
-				long lastReceive = FileTailingMessageProducerSupport.this.lastReceive;
-				if (now > lastReceive + FileTailingMessageProducerSupport.this.idleEventInterval
-						&& now > lastAlertAt + FileTailingMessageProducerSupport.this.idleEventInterval
-						&& FileTailingMessageProducerSupport.this.lastNoMessageAlert.compareAndSet(lastAlertAt, now)) {
+				long lastAlertAt = this.lastNoMessageAlert.get();
+				long lastReceive = this.lastReceive;
+				if (now > lastReceive + this.idleEventInterval
+						&& now > lastAlertAt + this.idleEventInterval
+						&& this.lastNoMessageAlert.compareAndSet(lastAlertAt, now)) {
 					publishIdleEvent(now - lastReceive);
 				}
 			}, this.idleEventInterval);
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.file.tail;
 
 import java.io.File;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.springframework.context.ApplicationEventPublisher;
@@ -28,7 +29,6 @@ import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.file.event.FileIntegrationEvent;
 import org.springframework.messaging.Message;
 import org.springframework.scheduling.TaskScheduler;
-import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
 import org.springframework.util.Assert;
 
 /**
@@ -57,7 +57,7 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 
 	private volatile long lastReceive = System.currentTimeMillis();
 
-	private volatile TaskScheduler scheduler;
+	private ScheduledFuture<?> idleEventScheduledFuture;
 
 	@Override
 	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
@@ -107,17 +107,9 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 		return this.taskExecutor;
 	}
 
-	protected TaskScheduler getRequiredTaskScheduler() {
-		if (this.scheduler == null) {
-			TaskScheduler taskScheduler = super.getTaskScheduler();
-			if (taskScheduler == null) {
-				ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
-				scheduler.initialize();
-				taskScheduler = scheduler;
-			}
-			this.scheduler = taskScheduler;
-		}
-		return this.scheduler;
+	@Override
+	public void setTaskScheduler(TaskScheduler taskScheduler) {
+		super.setTaskScheduler(taskScheduler);
 	}
 
 	@Override
@@ -131,7 +123,7 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 				.setHeader(FileHeaders.ORIGINAL_FILE, this.file)
 				.build();
 		super.sendMessage(message);
-		this.updateLastReceive();
+		updateLastReceive();
 	}
 
 	protected void publish(String message) {
@@ -144,11 +136,13 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 		}
 	}
 
+
+
 	@Override
 	protected void doStart() {
 		super.doStart();
 		if (this.idleEventInterval > 0) {
-			this.getRequiredTaskScheduler().scheduleWithFixedDelay(() -> {
+			this.idleEventScheduledFuture = getTaskScheduler().scheduleWithFixedDelay(() -> {
 				long now = System.currentTimeMillis();
 				long lastAlertAt = FileTailingMessageProducerSupport.this.lastNoMessageAlert.get();
 				long lastReceive = FileTailingMessageProducerSupport.this.lastReceive;
@@ -158,6 +152,14 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 					publishIdleEvent(now - lastReceive);
 				}
 			}, this.idleEventInterval);
+		}
+	}
+
+	@Override
+	protected void doStop() {
+		super.doStop();
+		if (this.idleEventScheduledFuture != null) {
+			this.idleEventScheduledFuture.cancel(true);
 		}
 	}
 
@@ -178,7 +180,7 @@ public abstract class FileTailingMessageProducerSupport extends MessageProducerS
 			this.eventPublisher.publishEvent(event);
 		}
 		else {
-			logger.info("No publisher for idle event:");
+			logger.info("No publisher for idle event");
 		}
 	}
 
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.file.tail;
 
 import java.io.File;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
@@ -26,6 +27,8 @@ import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.file.event.FileIntegrationEvent;
 import org.springframework.messaging.Message;
+import org.springframework.scheduling.TaskScheduler;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
 import org.springframework.util.Assert;
 
 /**
@@ -33,6 +36,7 @@ import org.springframework.util.Assert;
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Ali Shahbour
  * @since 3.0
  *
  */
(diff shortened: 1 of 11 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
```diff
@@ -118,6 +144,68 @@ protected void publish(String message) {
 		}
 	}
 
+	@Override
+	protected void doStart() {
+		super.doStart();
+		if (this.idleEventInterval > 0) {
+			this.getRequiredTaskScheduler().scheduleWithFixedDelay(() -> {
```
artembilan: There must be a `Future` property to `cancel(true)` in the `doStop()`

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
```diff
@@ -118,6 +144,68 @@ protected void publish(String message) {
...
+	}
+
+	private void publishIdleEvent(long idleTime) {
+
+		if (this.eventPublisher != null) {
+			FileTailingIdleEvent event = new FileTailingIdleEvent(this, this.file, idleTime);
+			this.eventPublisher.publishEvent(event);
+		}
+		else {
+			logger.info("No publisher for idle event:");
```
artembilan: Typo. No need colon in the end.
Copy/paste artifact 😄

spring-integration-file/src/main/java/org/springframework/integration/file/tail/OSDelegatingFileTailingMessageProducer.java
```diff
@@ -138,18 +134,7 @@ private void runExec() {
...
-		if (this.scheduler == null) {
-			TaskScheduler taskScheduler = super.getTaskScheduler();
-			if (taskScheduler == null) {
-				ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
-				scheduler.initialize();
-				taskScheduler = scheduler;
-			}
-			this.scheduler = taskScheduler;
-		}
-		return this.scheduler;
```
artembilan: OK. I see where is the root of `getRequiredTaskScheduler()`. 
But a decision is the same: we always reuse `TaskScheduler` from the application context.

spring-integration-file/src/main/java/org/springframework/integration/file/tail/OSDelegatingFileTailingMessageProducer.java
```diff
@@ -185,7 +170,7 @@ private void startProcessMonitor() {
 				if (logger.isInfoEnabled()) {
 					logger.info("Restarting tail process in " + getMissingFileDelay() + " milliseconds");
 				}
-				getRequiredTaskScheduler()
+				this.getRequiredTaskScheduler()
```
artembilan: No `this.` on methods.

artembilan (changes requested):
And general comment:

We really need some real test-case for event emission. 
At least for coverage.
Plus documentation would be great addition as well.

Thank you one more time for cotribution!

shahbour:
code updated, i will work on the test now as i need to read on how to do that

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
```diff
@@ -96,6 +108,11 @@ protected TaskExecutor getTaskExecutor() {
 	}
 
 	@Override
+	public void setTaskScheduler(TaskScheduler taskScheduler) {
```
artembilan: M-m-m. We don't need this setter - `AbstractEndpoint`:
```
@Override
public void setTaskScheduler(TaskScheduler taskScheduler) {
	super.setTaskScheduler(taskScheduler);
}
```
Am I missing anything?

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
```diff
@@ -118,6 +136,78 @@ protected void publish(String message) {
...
 
+
+
+	@Override
+	protected void doStart() {
+		super.doStart();
+		if (this.idleEventInterval > 0) {
+			this.idleEventScheduledFuture = getTaskScheduler().scheduleWithFixedDelay(() -> {
+				long now = System.currentTimeMillis();
+				long lastAlertAt = FileTailingMessageProducerSupport.this.lastNoMessageAlert.get();
```
artembilan: Having this `Runnable` as a Lambda we don't need class prefix for properties any more. 😄

artembilan (changes requested):
That's all.
Very close to merge, but I still need tests and docs.

Thank you!

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
shahbour: i thought that also , but i saw it in aggregation so i though i should add it. i will remove it

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
shahbour: got you , Copy/paste artifact :smile:

shahbour:
Test at my machine are failing today , i don't know how it passed yesterday . `getTaskScheduler()` is returning null
```
protected TaskScheduler getTaskScheduler() {
		if (this.taskScheduler == null && this.beanFactory != null) {
			this.taskScheduler = IntegrationContextUtils.getTaskScheduler(this.beanFactory);
		}
		return this.taskScheduler;
	}
```

Am not sure why this.beanFactory is null .

shahbour:
> org.springframework.integration.file.tail.FileTailingMessageProducerTests > testOS FAILED
>     java.lang.AssertionError at FileTailingMessageProducerTests.java:156

Confirmed , the problem is from `getTaskScheduler`  , in the above test , the file doesn't exist and it has to reschedule to start fetching that is not happening.

Should i revert back to the initial task-scheduler implementation. i could not figure out why this.beanFactory is null even though `FileTailingMessageProducerSupport` do extent `IntegrationObjectSupport`

artembilan:
@shahbour ,

Well, since we have changed logic to rely on the provided `taskScheduler` or that one from the application context, we have to fix that `FileTailingMessageProducerTests` via `setTaskScheduler()`.

It is definitely a fact that `this.beanFactory is null` when you use component out side of Dependency Injection container.
Looks we have a code like:
```
@Test
@TailAvailable
public void testOS() throws Exception {
	OSDelegatingFileTailingMessageProducer adapter = new OSDelegatingFileTailingMessageProducer();
	adapter.setOptions(TAIL_OPTIONS_FOLLOW_NAME_ALL_LINES);
	testGuts(adapter, "reader");
}
```
As you see we don't have here any Spring related environment.
There is nothing to call ours `BeanFactoryAware` callback.
Therefore we have to provide its hooks manually.

spring-integration-file/src/test/java/org/springframework/integration/file/tail/FileTailingMessageProducerTests.java
```diff
@@ -124,9 +127,47 @@ public void canRecalculateCommandWhenFileOrOptionsChanged() throws IOException {
...
+		OSDelegatingFileTailingMessageProducer adapter = new OSDelegatingFileTailingMessageProducer();
+		adapter.setOptions(TAIL_OPTIONS_FOLLOW_NAME_ALL_LINES);
+		ThreadPoolTaskScheduler taskScheduler = new ThreadPoolTaskScheduler();
+		taskScheduler.afterPropertiesSet();
+		adapter.setTaskScheduler(taskScheduler);
+		final List<FileTailingIdleEvent> events = new ArrayList<>();
+		adapter.setApplicationEventPublisher(event -> {
+			if (event instanceof FileTailingIdleEvent) {
+				FileTailingIdleEvent tailEvent = (FileTailingIdleEvent) event;
+				logger.warn(event);
```
artembilan: I'm not sure in the WARN.
 Let remove it or at least convert it into DEBUG. 
I see the place you borrowed an approach and that doesn't look correct for me, too.
Let's change that to DEBUG as well!
My concern to avoid extra logging message, when we can live without them.
If there is some problem with test, we always can switch on DEBUG level for the category in the logging config.

spring-integration-file/src/test/java/org/springframework/integration/file/tail/FileTailingMessageProducerTests.java
```diff
@@ -124,9 +127,47 @@ public void canRecalculateCommandWhenFileOrOptionsChanged() throws IOException {
...
+			}
+		});
+		File file = new File(testDir, "foo");
+		file.delete();
+		FileOutputStream foo = new FileOutputStream(file);
+		for (int i = 0; i < 50; i++) {
+			foo.write(("hello" + i + "\n").getBytes());
+		}
+		foo.flush();
+		foo.close();
```
artembilan: I think if we expect `FileTailingIdleEvent`, we shouldn't expect any data from the file.
I find this writing to file as a overhead for this test case.

spring-integration-file/src/test/java/org/springframework/integration/file/tail/FileTailingMessageProducerTests.java
```diff
@@ -124,9 +127,47 @@ public void canRecalculateCommandWhenFileOrOptionsChanged() throws IOException {
...
+		adapter.setFile(file);
+		QueueChannel outputChannel = new QueueChannel();
+		adapter.setOutputChannel(outputChannel);
+		adapter.setIdleEventInterval(1000);
```
artembilan: Since we would like to be sure that we emit `FileTailingIdleEvent` anyway, that would be better to have `idleEventInterval` as small as possible to make the test case reasonably quick.

spring-integration-file/src/test/java/org/springframework/integration/file/tail/FileTailingMessageProducerTests.java
```diff
@@ -124,9 +127,47 @@ public void canRecalculateCommandWhenFileOrOptionsChanged() throws IOException {
...
+		adapter.afterPropertiesSet();
+		adapter.start();
+		Thread.sleep(4000);
```
artembilan: The `Thread.sleep()` is always evil for test-case.
First of all you may not need to sleep so much for condition to appear.
But on the other hand our long-term experience demonstrates that we can't trust `Thread.sleep()` because there is no guarantee that condition happens during that time on slow or busy machine like our CI server.

To to reach a goal you should use `CountDownLatch` and wait for its result here instead of `Thread.sleep()` .
You than `countDown()` it in that `event ->` Lamba.
In the end I'd really would like to see assert for the `FileTailingIdleEvent` type.

Hope that is clear.

src/reference/asciidoc/file.adoc
```diff
@@ -411,6 +411,10 @@ Examples of such events are:
 
 This sequence of events might occur, for example, when a file is rotated.
 
+`[message=Idle timeout, file=/tmp/foo] [idle time=5438]`
```
artembilan: This something unrelated.
Would be better to tell:
> Starting with _version 5.0_, a `FileTailingIdleEvent` is emitted when there is no data in the file during `idleEventInterval`.

As a general note. And only after that a sample.

This addition should be mentioned in the `What's New`, too.

artembilan (changes requested):
I'm pretty sure this is the last our review cycle.

shahbour:
No problem, I will fix all.

Thanks for the review and support.

shahbour:
i am not sure if this is the behavior we want , but currently if the file got deleted or doesn't exist in the first place we will  emit the idle events.

artembilan:
> but currently if the file got deleted or doesn't exist in the first place we will emit the idle events

Yeah... Good catch, @shahbour !

I thought about that before, but I didn't find a hook how to catch the case if file exists and tailing process has been started.

Perhaps a middle ground would be like schedule idle detection process only after first read line.
I mean in the `send(String line)` method.
Not sure what is the behavior if we don't have any lines just from the beginning. Exactly what I asked you to do in your new test-case.
Would you mind to investigate and provide an appropriate improvement?

Thank you very much for the effort and Merry Christmas finally to you!

shahbour:
May be listening to events can be a solution , i will figure out some thing solid re commit
Merry Christmas for you too

spring-integration-file/src/test/java/org/springframework/integration/file/tail/FileTailingMessageProducerTests.java
```diff
@@ -124,13 +131,48 @@ public void canRecalculateCommandWhenFileOrOptionsChanged() throws IOException {
...
+		});
+		File file = new File(testDir, "foo");
+		file.delete();
+		adapter.setFile(file);
+		QueueChannel outputChannel = new QueueChannel();
+		adapter.setOutputChannel(outputChannel);
+		adapter.setIdleEventInterval(100);
+		adapter.afterPropertiesSet();
+		adapter.start();
+		boolean noEvent = countDownLatch.await(500, TimeUnit.MILLISECONDS);
```
artembilan: M-m-m. Sorry, I'm very sensitive to an extra sleep for any our tests.
I would like to finish them as fast as possible 
Imagine like I run tests for the project at about one hundred times as day 😄 .

I guess there should be some special even when file doesn't exist. I'm pretty sure tailers should throw something on the matter.
So, it would be much better to catch that event as well via its particular `CountDownLatch`. And we have to ensure that there hasn't been any other events before. Like in our case we don't expect `FileTailingIdleEvent` if there is no file.

I don't say that your solution is wrong. Definitely it looks like the best compromise.
I just mean that the test would be better 😉 

That's all. 
Only what I need right now is a proper test-case.
Everything rest I will polish on merge and share here as a link to the commit.

So, I'm waiting from your for the last commit.

shahbour: i am not sure we can skip this wait, because even if we catch the no file exist event without any other event that doesn't mean `FileTailingIdleEvent` is not going to be emitted later on , we have to wait till `IdleEventInterval`  pass and then check no event is emitted .

Unless i am missing some thing , i think we have to wait at least the `IdleEventInterval`

artembilan: Yeah, what you say sounds reasonable.
Let's take a look to variant of spying `File` and mock the call to its `exists()` with one more latch and verify that we don't have idle event afterwards, just only that one about file doesn't exist.

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
```diff
@@ -118,6 +140,68 @@ protected void publish(String message) {
...
+
+
+	@Override
+	protected void doStart() {
+		super.doStart();
+		if (this.idleEventInterval > 0) {
+			this.idleEventScheduledFuture = getTaskScheduler().scheduleWithFixedDelay(() -> {
+				long now = System.currentTimeMillis();
+				long lastAlertAt = this.lastNoMessageAlert.get();
+				long lastReceive = this.lastProduce;
```
artembilan: The wrong local variable name. As well as for the `updateLast...` method

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
```diff
@@ -118,6 +140,68 @@ protected void publish(String message) {
...
+	}
+
+	private void publishIdleEvent(long idleTime) {
+
+		if (this.eventPublisher != null && getFile().exists()) {
+			FileTailingIdleEvent event = new FileTailingIdleEvent(this, this.file, idleTime);
+			this.eventPublisher.publishEvent(event);
+		}
+		else {
+			logger.info("No publisher for idle event");
```
artembilan: This logger doesn't reflect reality any more. We come here even when file doesn't exist. So, we should move `exists` condition inside

shahbour:
Sorry I did not understand what is required , can you please explain more.

src/reference/asciidoc/file.adoc
```diff
@@ -411,6 +411,10 @@ Examples of such events are:
 
 This sequence of events might occur, for example, when a file is rotated.
 
+NOTE: Starting with version 5.0, a `FileTailingIdleEvent` is emitted when there is no data in the file during `idleEventInterval`.
```
artembilan: I don't think that it is a NOTE. It should be just regular narrative. Just imaging the next version and some new option. So, all of them should be in notes? No, just simple expression

src/reference/asciidoc/whats-new.adoc
```diff
@@ -51,6 +51,8 @@ See <<feed>> for more information.
 The new `FileHeaders.RELATIVE_PATH` Message header has been introduced to represent relative path in the `FileReadingMessageSource`.
 See <<file-reading>> for more information.
 
+The tail adapter now support `IdelEventInterval` to emit events when idle.
+
```
artembilan: Please, add a link to the tail adapter.

artembilan (changes requested):

artembilan:
Take a look into Mockito Framework: http://site.mockito.org/. Also we have planty of tests using that. My idea is to avoid that wait time, but just react to the particular event ( file exists in our cases) and be sure in the proper state of the adapter.
If that is still complicated for you , I'll come back with the solution on Tuesday. It's holidays here in US these days 😀

shahbour:
No worries you don't have to bother your self in vacation by the way same here vacation till next year  :grinning: 
Have a nice vacation and Merry Christmas and Happy New Year .

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f7b1ec88851219d666d4c6046c29661bc5717682.

@shahbour ,
Looking forward for more!

PR #2009, opened 2016-12-22
INT-4193: Large Group Aggregation Performance

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4193

When using the internal `SequenceAwareMessageGroup` within an correlating
message handler, the messages were copied to a new collection before
checking for duplicate sequences in `canAdd()`.

This was unnecessary since we never add anything to this group, if `canAdd()`
returns true, the message is added to the store; this group is discarded.

Instead, use the message collection from the original group; although it is
not modifiable, this is not an issue because we don't need to modify it.

__cherry-pick to master__ (expect minor conflicts in docs).


artembilan:
... and cherry-picked to `master` as https://github.com/spring-projects/spring-integration/commit/318bb4c4b7d087c079284aed3a0a10e02eabed6f.

That was awesome fix for so silly issue, @garyrussell .
In your short-as-wise style 😄

PR #2010, opened 2016-12-22
INT-4195: More Aggregator Performance Improvements

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4195

Avoid the N^2 search in `SequenceAwareMessageGroup` when the group is a
`SimpleMessageGroup`.
Avoid the use of the header accessor if the N^2 search is needed (SPR-15045).

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -752,6 +752,8 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	protected static class SequenceAwareMessageGroup extends SimpleMessageGroup {
 
+		private final SimpleMessageGroup sourceGroup;
+
 		public SequenceAwareMessageGroup(MessageGroup messageGroup) {
 			/*
 			 * Since this group is temporary, and never added to, we simply use the
@@ -760,6 +762,12 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 			 */
 			super(messageGroup.getMessages(), null, messageGroup.getGroupId(), messageGroup.getTimestamp(),
 					messageGroup.isComplete(), true);
+			if (messageGroup instanceof SimpleMessageGroup) {
+				this.sourceGroup = (SimpleMessageGroup) messageGroup;
+			}
+			else {
+				this.sourceGroup = null;
+			}
 		}
 
 		/**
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -777,17 +785,25 @@ public boolean canAdd(Message<?> message) {
...
 			return true;
 		}
 
 		private boolean containsSequenceNumber(Collection<Message<?>> messages, Integer messageSequenceNumber) {
 			for (Message<?> member : messages) {
-				Integer memberSequenceNumber = new IntegrationMessageHeaderAccessor(member).getSequenceNumber();
-				if (messageSequenceNumber.equals(memberSequenceNumber)) {
-					return true;
+				Object sequenceNumber = member.getHeaders().get(
+						IntegrationMessageHeaderAccessor.SEQUENCE_NUMBER);
```
artembilan: Yes, I thought not one time that use extra object to just get friendly accessor to the value isn't good idea at all...
Let's don't use it in the `canAdd()` as well !

Anyway why `member.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_NUMBER, Integer.class)` doesn't work for you here?
Looks like pretty parallel with what you do manually with this `if..else`

spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageGroup.java
```diff
@@ -123,6 +128,10 @@ public int getLastReleasedMessageSequenceNumber() {
 	}
 
 	private boolean addMessage(Message<?> message) {
+		Object sequence = message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_NUMBER);
+		if (sequence instanceof Integer) {
```
artembilan: M-m-m. I think that should be fine just to request only `Integer` type for the header, like it is done right now via `messageHeaderAccessor.getSequenceNumber()`.
I mean we should just do this directly without any `if`:
```
Integer sequence = message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_NUMBER, Integer.class);
this.sequences.add(sequence != null ? sequence : 0);
```
and that's all

spring-integration-core/src/test/java/org/springframework/integration/aggregator/AggregatorTests.java
```diff
@@ -130,6 +131,55 @@ public void testAggPerf() throws InterruptedException, ExecutionException, Timeo
...
+					.build());
+		}
+		stopwatch.stop();
+		logger.warn("Sent " + 120000 + " in " + stopwatch.getTotalTimeSeconds() +
+				" (10k in " + stopwatch.getLastTaskTimeMillis() + "ms)");
+
+		Collection<?> result = resultFuture.get(10, TimeUnit.SECONDS);
+		assertNotNull(result);
+		assertEquals(120000, result.size());
+		assertThat(stopwatch.getTotalTimeSeconds(), lessThan(60.0)); // actually < 2.0, was many minutes
```
artembilan: Well, I have given up already after:
```
2016-12-22 18:18:43,198 WARN AggregatorTests [main] : Sent 10000 in 26.661 (10k in 26661ms)
2016-12-22 18:20:01,117 WARN AggregatorTests [main] : Sent 20000 in 104.58 (10k in 77919ms)
```
😄

artembilan (changes requested):
All minor.
I can address on merge.
Jut let me know!

artembilan:
H-m.
That fully unclear to me what's wrong with that MongoDB test on Travis.
There is no any race condition. Only I can guess that didn't have connect to the MongoDB during TX commit.
Different story though...

artembilan:
LGTM.
Although WDYT about many other `new IntegrationMessageHeaderAccessor()` usage?
Wouldn't it be better to get rid of them as well?
Thanks

garyrussell:
I don't think it's such a big issue once Spring is fixed.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/a466579ee1e9b3a846373a53a9b5ce20a921294d and cherry-picked.

PR #2016, opened 2017-01-03
INT-4201: Fix Gateway Param Annotation Aliases

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4201

name<->value alias was not processed correctly so using
`@Header(name = "baz") String baz` did not work.

Use `SynthesizingMethodParameter`.

__cherry-pick to 4.3.x__


artembilan (approved):

artembilan:
... and cherry-picked

PR #2017, opened 2017-01-03
INT-4202: Fix StoredProcOutboundGateway NPE

artembilan:
JIRA: https://jira.spring.io/browse/INT-4202

The `StoredProcOutboundGateway` uses `MessageBuilder` directly for procedure result without any conditions.
If procedure result is `null`, `MessageBuilder.withPayload` throws `java.lang.IllegalArgumentException: payload must not be null`.

* Since the super `AbstractReplyProducingMessageHandler` class takes care about `null` reply properly via its `requiresReply` property and really uses `MessageBuilder`
 for reply, just fix `StoredProcOutboundGateway` to return procedure result as is from the `handleRequestMessage()` implementation
* Fix some typos and code style
* Increase some JDBC tests performance changing `Thread.sleep()` solution to proper `PollableChannel.receive()` or iterations over expected result

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/StoredProcOutboundGateway.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/StoredProcOutboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/StoredProcOutboundGateway.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/StoredProcOutboundGateway.java
@@ -45,10 +45,9 @@ public class StoredProcOutboundGateway extends AbstractReplyProducingMessageHand
 	 * @param storedProcExecutor Must not be null.
 	 */
 	public StoredProcOutboundGateway(StoredProcExecutor storedProcExecutor) {
-
 		Assert.notNull(storedProcExecutor, "storedProcExecutor must not be null.");
 		this.executor = storedProcExecutor;
-
+		setRequiresReply(true); // NOSONAR It's fine to be modified from outside
 	}
 
 	@Override
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/StoredProcOutboundGateway.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/StoredProcOutboundGateway.java
@@ -40,6 +40,8 @@ public class StoredProcOutboundGateway extends AbstractReplyProducingMessageHand
 
 	private volatile boolean expectSingleResult = false;
 
+	private boolean requiresReplyExplicitlySet;
+
 	/**
 	 * Constructor taking {@link StoredProcExecutor}.
 	 * @param storedProcExecutor Must not be null.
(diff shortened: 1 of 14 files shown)


spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/StoredProcOutboundGatewayWithSpelIntegrationTests-context.xml
```diff
@@ -57,6 +57,7 @@
 										   stored-procedure-name="GET_MESSAGE"
 										   ignore-column-meta-data="true"
 										   expect-single-result="true"
+										   requires-reply="true"
```
garyrussell: Perhaps this should be true by default? That way, there would be no behavior change - exception when the SP returns `null`.

artembilan: True. More over this is about `gateway` anyway.

artembilan:
You know, I've changed my mind.
Now I remember why I didn't make this gateway as a `requiresReply = true` by default.
Since this is a filter-aware component (a-la `SELECT ... WHERE`), it isn't surprise that we may receive a `null` (or empty list) from procedure on regular basis and it is still unfortunate that we don't have `NullPayload` abstraction...
But the story isn't about that...

My main concern that making it as a `requiresReply = true` by default may provide a breaking change for those who periodically gets empty lists from the procedure. 

I think that we should restore `requiresReply = true` not only for that reason, but for the logical purpose of the previous `IllegalStateException` and current `ReplyRequiredException`. They are really not orthogonal to replace each other. 

So, the original fix is still in its power, because that was a bug.
But `requiresReply = true`  should be revised somewhere [else](https://jira.spring.io/browse/INT-3333) and not as a breaking change to the maintain version...

WDYT?

garyrussell:
I don't follow your logic - today, if the SP returns `null`, the subject exception is thrown. Silently dropping the `null` would be a behavior change, with `requires-reply="true"`, yes, we'll get a different exception, but it still will fail rather than silently drop.

So that will be "less" of a breaking change.

artembilan:
The subject exception is thrown when `expectSingleResult = true`.
Changing to the `requires-reply="true"` will cause empty list case to fail with the `ReplyRequiredException`.

Remember, I don't like your tendency to keep functionality "replying on bug" :wink 
?
That is my point for this fix: by design we have to drop `null` by default, but right now we fail with the `IllegalState`, which can't be fixed by end-user.

garyrussell:
OK; now that I've actually looked at the code, see what you mean; however, I think we should set `requiresReply = this.expectSingleResult` so that if they expect a single result and null is returned, it should be an error.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/StoredProcOutboundGateway.java
```diff
@@ -56,6 +87,14 @@ public String getComponentType() {
 	}
 
 	@Override
+	protected void doInit() {
+		super.doInit();
+		if (!this.requiresReplyExplicitlySet) {
+			setRequiresReply(this.expectSingleResult);
+		}
```
garyrussell: Why not just do this in the setter instead of overriding this method?

artembilan: Well, there is no guaranty in the order how setters are called via XML parser.
So, we may end up with undesired affect, when `setRequiresReply()` is called later with `false`, although we would like true by the `expectSingleResult`.

A similar issue with non-idempotent behavior with raw java, when we change the order how we call setters.

With a `requiresReplyExplicitlySet` flag and modification in the `doInit()` we reach a robust state.

Am I missing something ?

garyrussell: I was suggesting keeping the flag and using it in the setter; that way, it doesn't matter which order the setters are called. But, maybe that's too much logic in a setter and, in any case, this component won't work properly if it's configured outside of Spring; so I am ok with this. Will merge soon.

artembilan: Great! Thank you!

garyrussell:
Merged as b7d7d767899de933b4be6937286fd2bfe46ee280 and cherry-picked as 0c2ba3f7616cdc22ac839875f95d8fb1e2ac0d15

PR #2020, opened 2017-01-03
INT-4197: Fix annotation case when non `messages`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4197

Since `Collection<Message<?>>` can be possible only if really deal with messages as a group (`MethodInvokingMessageListProcessor`), the case with the `List<?>` param for Messaging Annotation method should not be treated as candidate for `messages` collection.

* Add `this.canProcessMessageList` condition to avoid `messages` SpEL expression when we are not in the `MethodInvokingMessageListProcessor` environment

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -391,7 +391,7 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 				return;
 			}
 			if (methodName == null
-					&& ObjectUtils.containsElement(new String[] {"start", "stop", "isRunning"}, method1.getName())) {
+					&& ObjectUtils.containsElement(new String[] { "start", "stop", "isRunning" }, method1.getName())) {
 				return;
 			}
 			HandlerMethod handlerMethod1 = null;
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -391,7 +391,7 @@ else if (!Modifier.isPublic(method1.getModifiers())) {
 				return;
 			}
 			if (methodName == null
-					&& ObjectUtils.containsElement(new String[] { "start", "stop", "isRunning" }, method1.getName())) {
+					&& ObjectUtils.containsElement(new String[] {"start", "stop", "isRunning"}, method1.getName())) {
```
garyrussell: Hmmm - my STS config adds these spaces - I guess it's not a checkstyle rule though.

spring-integration-core/src/test/java/org/springframework/integration/config/annotation/RouterAnnotationPostProcessorTests.java
```diff
@@ -63,6 +75,24 @@ public void testRouter() {
...
 
+	@Test
+	public void testRouterWithListParam() {
+		MessagingAnnotationPostProcessor postProcessor = new MessagingAnnotationPostProcessor();
+		postProcessor.setBeanFactory(context.getBeanFactory());
+		postProcessor.afterPropertiesSet();
+		TestRouter testRouter = new TestRouter();
+		postProcessor.postProcessAfterInitialization(testRouter, "test");
+		context.refresh();
+		routingChannel.send(new GenericMessage<String>("foo"));
```
garyrussell: Although we are relying on the `ReflectiveMethodExecutor` to convert this to a list of `String`, that would not be obvious to a casual reader who might be trying to understand how things work by reading test cases.

We should either send a `Collections.singletonList("foo")` or make a comment here to help such people.

garyrussell (changes requested):

garyrussell:
Merged as dd34f3de2aaba039a7e099b767266e1f554162f7 and cherry-picked.

PR #2021, opened 2017-01-03
INT-4203: Expression RH Advice Improvements

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4203

Add channel names for easier use in DSL.

Add documentation (boot) example.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ExpressionEvaluatingRequestHandlerAdvice.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ExpressionEvaluatingRequestHandlerAdvice.java
@@ -28,7 +28,6 @@ import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.messaging.MessagingException;
 import org.springframework.messaging.support.ErrorMessage;
-import org.springframework.util.Assert;
 
 /**
  * Used to advise {@link MessageHandler}s.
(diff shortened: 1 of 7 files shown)


artembilan (changes requested):
LGTM, of course, but I wonder if we can provide some changes to those expression setters ,e.g. rename `setOnSuccessExpression(String)` to the `setOnSuccessExpressionString(String)`. And `setExpressionOnSuccess(Expression)` to `setOnSuccessExpression(Expression)`

garyrussell:
@artembilan We could do that but we'd have to deprecate them in 4.3 and I don't think we can do that because it will cause ambiguity for XML configuration.

artembilan:
Hm. So, we have to implement the `ExpressionEvaluatingRequestHandlerAdvice` Namespace support first of all: https://jira.spring.io/browse/INT-3596 😄 .

Well, we can do via iterative steps: 
* Introduce `setOnSuccessExpressionString(String)` and deprecate `setOnSuccessExpression(String)`  in 4.3.x
* Remove  `setOnSuccessExpression(String)`, introduce `setOnSuccessExpression(Expression)` and deprecate `setExpressionOnSuccess(Expression)` in 5.0

WDYT ?

garyrussell:
Now should be __cherry-picked to 4.3.x__ but restore (and deprecate) the deleted messages and remove the new `set...Expression(Expression)` variants.

Let me know if you want a separate PR.

artembilan:
No, I'm fine.
Thank you!
Just wonder what "the deleted messages" means. 
Setters ?

artembilan:
```
org.springframework.integration.file.config.FileOutboundChannelAdapterParserTests > adapterUsageWithFailMode FAILED

    java.lang.IllegalStateException

        Caused by: org.springframework.beans.factory.BeanCreationException

            Caused by: org.springframework.beans.factory.BeanCreationException

                Caused by: org.springframework.beans.ConversionNotSupportedException

                    Caused by: java.lang.IllegalStateException
```
See Travis report.
Yeah... it's really unfortunate, but that is breaking change, right.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/77fd8a4684df7f7e54b1ff5ba86c19a55f75bafb
and cherry-picked as https://github.com/spring-projects/spring-integration/commit/3301da4157123264df1e0c7db2d4b3095edd7193

PR #2022, opened 2017-01-04
INT-4134: Add `IntegrationFlows.from(Class<?>)`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4134

* Introduce `AnnotationGatewayProxyFactoryBean` to parse `@MessagingGateway` annotation directly in the bean.
Useful for Java (`@Bean`) configuration variant
* Use `AnnotationGatewayProxyFactoryBean` for the newly introduced `IntegrationFlows.from(Class<?>)`
* Override `requestChannel` in the result gateways to the auto-created `DirectChannel` in the current `IntegrationFlowDefinition`
* Register `AnnotationGatewayProxyFactoryBean` from the `IntegrationFlows.from(Class<?>)` as a bean with the name like: `IntegrationFlow` bean name plus `.gateway` suffix
* Merge `GatewayCompletableFutureProxyFactoryBean` to `GatewayProxyFactoryBean`, since we are based on the Java 8 already anyway


garyrussell:
LGTM; I wonder if we should also consider:

```java
public interface Foo {

    void bar(String s);

    void baz(Integer i);

}
```

and

    IntegrationFlows.fromGateway("my.Foo.bar")...

and/or

```java
Method m = Foo.class.getDeclaredMethod(...);
...
IntegrationFlows.from(m)...
```

garyrussell:
Of course, it would be a bit more complex because we'd have to merge flows into a single GPFB, so maybe not worth it (at least for now).

artembilan:
Well, I'm not sure in the first case.
Each `IntegrationFlows.from(Class<?>)` registers a new proxy bean. That's why I have introduced `flowBeanName.gateway` bean name for qualification. And at the same time you still can have a regular top-level bean via `@IntegrationComponentScan`.
So, I don't see reason to distinguish methods. Just need to document properly that in case of  `IntegrationFlows.from(Class<?>)` all interface methods will send message to the same channel. That is a purpose of such an addition - to start a flow from the Messaging Gateway.

The method case looks really pretty complex and reminds me `@Publisher` and doesn't looks related to `IntegrationFlow`. That is really a method level responsibility.

garyrussell:
Right, I was just trying to mirror the XML and `@ IntegrationComponentScan ` mechanism where each method can go to a different flow (based on `requestChannel`). My initial thought was to register some global GPFB per interface and hook the individual methods up to separate flows.

But, after more thought, concluded it's not worth the effort that would be required (handle missing methods, duplicates etc, etc).

So merged.

PR #2029, opened 2017-01-12
INT-4214: Migrate to `InvocableHandlerMethod`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4186

* Rework `MessagingMethodInvokerHelper` to delegate to the `InvocableHandlerMethod` with its `HandlerMethodArgumentResolver` infrastructure instead of SpEL
* Introduce several `HandlerMethodArgumentResolver` to address some SI use-cases like `@Payloads`, `@Payload(expression = "")` and `Collection` as argument
* Initialize `DefaultMessageHandlerMethodFactory` in the `MessagingMethodInvokerHelper.start()`.
With that I observed several `Lifecycle` problem when we don't have proper delegate from the top.
* Fix `AbstractCorrelatingMessageHandler` and similar to delegate `Lifecycle` properly
* Fix `ReactiveConsumer` to delegate `Lifecycle` to the `MessageHandler`
* Fix `MutableMessage` do not `generateId()` and set `timestamp` headers if we already have them in the provided headers
* With all that `Lifecycle` many tests must be fixed to call `start()`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.handler.support;
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+import org.springframework.core.MethodParameter;
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.integration.util.AbstractExpressionEvaluator;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
+import org.springframework.util.Assert;
+
+/**
+ * @author Artem Bilan
+ *
+ * @since 5.0
+ */
+public class CollectionArgumentResolver extends AbstractExpressionEvaluator
+		implements HandlerMethodArgumentResolver {
+
+	private final boolean canProcessMessageList;
+
+	public CollectionArgumentResolver(boolean canProcessMessageList) {
+		this.canProcessMessageList = canProcessMessageList;
+	}
+
+	@Override
+	public boolean supportsParameter(MethodParameter parameter) {
+		Class<?> parameterType = parameter.getParameterType();
+		return Collection.class.isAssignableFrom(parameterType) || parameterType.isArray();
+	}
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public Object resolveArgument(MethodParameter parameter, Message<?> message) throws Exception {
+		Object value = message.getPayload();
+
+		if (this.canProcessMessageList) {
+			Assert.state(value instanceof Collection,
+					"This Argument Resolver support only messages with payload as Collection<Message<?>>");
+			Collection<Message<?>> messages = (Collection<Message<?>>) value;
+
+			parameter.increaseNestingLevel();
+			if (Message.class.isAssignableFrom(parameter.getNestedParameterType())) {
+				value = messages;
+			}
+			else {
+				value = messages.stream()
+						.map(Message::getPayload)
+						.collect(Collectors.toList());
+			}
+			parameter.decreaseNestingLevel();
+		}
+		return getEvaluationContext()
+				.getTypeConverter()
+				.convertValue(value,
+						TypeDescriptor.forObject(value),
+						TypeDescriptor.valueOf(parameter.getParameterType()));
+	}
+
+}
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
@@ -17,6 +17,8 @@
 package org.springframework.integration.handler.support;
 
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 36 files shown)


artembilan:
@garyrussell ,
Take a look, please.
I have spent on this already so much time and still don't see the finish.

* Need to address `Iterator` and `Map` conversion cases.
* Still need to revise the new logic for method analyzing. I feel like we don't need some of the branches.
* I can move init logic outside of `start()` (lazy init in the `process()`) to make test support easier, but i still think that those `Lifecycle`  delegation issues are still valid.
* Would be interested to hear from you if you would like to see some performance comparison test, but unfortunately I dropped SpEL altogether. Hope we still can address conversion concerns via `HandlerMethodArgumentResolver`s 😄

garyrussell:
Thanks; I'll take a look tomorrow; my intention was always that we keep SpEL as a fallback option, just in case there is some corner case in user-land that we didn't think of; I still think we will need that; perhaps MIMH2 ? (if it's too complex to keep both).

In any case, I think we can do stand-alone performance analysis.

artembilan:
See https://jira.spring.io/browse/SPR-15139 for more info.

artembilan:
Since Juergen fixed already https://jira.spring.io/browse/SPR-15139, I think we can go ahead with the solution like falback to SpEL, if `InvocableHandlerMethod` has failed. Just because there is now `MethodArgumentResolutionException` thrown from the `invoke()` when we can't resolve argument for the provided data and method signature.
Any other thoughts?

garyrussell:
Sounds good; I haven't had a chance to look at your PR yet.

artembilan:
This isn't ready to merge yet. 
Just to share a progress.
Also pay attention to the: https://jira.spring.io/browse/SPR-15151

artembilan:
Provide some further progress after fix https://jira.spring.io/browse/SPR-15151.

There is now performance test-case thanks to new `useSpelInvoker` option.
The result looks like:
```
StopWatch 'SpEL vs Invocable Performance': running time (millis) = 1238
-----------------------------------------
ms     %     Task name
-----------------------------------------
01174  095%  SpEL
00064  005%  Invocable
```

Looks unbelievable 😄 !

So, I might have missed something...

garyrussell:
Interesting - would be cool to compare with compiled SpEL too 😄

artembilan:
Looks like this:
```
-----------------------------------------
ms     %     Task name
-----------------------------------------
01217  065%  SpEL
00122  007%  Invocable
00535  029%  Compiled SpEL
```

I think we still good to go ahead with all that refactoring hell.
It is worth that 😄 !

If I don't miss something yet...

garyrussell:
Cool - I'll start reviewing after the SCSt calls.

garyrussell:
It's not yet clear to me why you need all that lifecycle stuff.

>Initialize DefaultMessageHandlerMethodFactory in the MessagingMethodInvokerHelper.start().

I see that was removed in a later commit.

artembilan:
Right, I moved it to the lazy-init part. See my comment https://github.com/spring-projects/spring-integration/pull/2029#issuecomment-272297813.
But we still have to delegate lifecycle control to the target. No?

I maybe would agree that it isn't related to the method invocation. Especially if we use the same `target` for different method invocations. So, `target` lifecycle control should be external.
But! We implemented that part because of https://jira.spring.io/browse/INT-3571.

So, to simplify the review I really can revert that " lifecycle stuff".
But we definitely should revise it afterwards...

artembilan:
The work is done.
Ready for review and merging.

spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
```diff
@@ -0,0 +1,91 @@
...
+
+import org.springframework.core.MethodParameter;
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.integration.util.AbstractExpressionEvaluator;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
+import org.springframework.util.Assert;
+
+/**
+ * @author Artem Bilan
```
garyrussell: Javadocs

spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public Object resolveArgument(MethodParameter parameter, Message<?> message) throws Exception {
+		Object value = message.getPayload();
+
+		if (this.canProcessMessageList) {
+			Assert.state(value instanceof Collection,
+					"This Argument Resolver support only messages with payload as Collection<Message<?>>");
```
garyrussell: `s/support only/only supports/`

`s/with payload as/with a payload of/`

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MapArgumentResolver.java
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.core.convert.ConversionService;
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.integration.util.BeanFactoryTypeConverter;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.handler.annotation.Headers;
+import org.springframework.messaging.handler.annotation.Payload;
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
+
+/**
+ * @author Artem Bilan
```
garyrussell: Javadoc

garyrussell (changes requested):
That's all for now - will do a more thorough review tomorrow.

garyrussell:
Excellent work!!

One thing that's not clear is how the user practically can call `MIMP.setUseSpelInvoker(true)`.

I am not sure we want to go to the trouble of adding an attribute to every annotation, but on the other hand, setting global option would not be good because it might be just one corner case where they need to force SpEL.

Maybe some kind of registry.

But we can leave this for another PR. Will merge after javadocs are addressed.

artembilan:
Yes, I thought about that, Gary, as well.
And my only conclusion to have something like global registry which accepts `MessageHandler` ids to switch to the `useSpelInvoker` in the `afterSingletonsInstantiated()`.
Anyway I think there is still needed some work to pull that `MIMP.setUseSpelInvoker(true)` to the `MH` for proper delegation.
Would be glad to come back to this subject in the separate PR.

Finishing with JavaDocs now...

Meanwhile, please, share your thoughts about `Lifecycle` propagation concern: https://github.com/spring-projects/spring-integration/pull/2029#issuecomment-273321135

garyrussell:
Merged as b3db6a97a25f397bc25c4868422c902035b540fb

PR #2033, opened 2017-01-16
INT-4204: Introduce `ExpressionFileListFilter`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4204

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/ExpressionFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/ExpressionFileListFilter.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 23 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/filters/ExpressionFileListFilter.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A SpEL expression based {@link AbstractFileListFilter} implementation.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.0
+ */
+public class ExpressionFileListFilter<F> extends AbstractFileListFilter<F>
```
garyrussell: Given that SpEL is not type safe, and we don't need subclasses, should this be generic?

    public class ExpressionFileListFilter extends AbstractFileListFilter<Object>

?

garyrussell (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/filters/ExpressionFileListFilter.java
artembilan: Well, we could, indeed but then we can't use `AbstractInboundFileSynchronizer.setFilter(FileListFilter<F> filter)` as is.
We have to go through all the casting hell.

garyrussell: 👍

garyrussell (approved):

garyrussell:
Merged as 9836828f54def6e3276b3370d8e2dab9f9788020

PR #2034, opened 2017-01-19
INT-4210: DSL: Make a final `.log()` Terminal

artembilan:
JIRA: https://jira.spring.io/browse/INT-4210

Previously use-case like:
```
.transform()
.log()
.get();
```
failed with the `Dispatcher has no subscribers` for an implicit channel populated by the `.wireTap()` because actually `LoggingHandler` is placed on the wire-tapped channel

Since logically it looks like end-user tries to have `LoggingHandler` as a terminal of the flow, we add `nullChannel` if `WireTapSpec` is in the end of the flow definition.

If user would like to continue flow (e.g. return reply to the `replyChannel`), he must end more EIP-methods after `.log()` (like before, of course)

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Optional;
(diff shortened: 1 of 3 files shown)


garyrussell:
Merged s 36f3c533aa6768b9ed8d47fda37ccba7b9982d2d

garyrussell:
Oops - I didn't squash.

artembilan:
> Oops - I didn't squash.

He-he! So, one more commit for me in `master` 😄 
At least the commit flow is still linear.

PR #2035, opened 2017-01-20
INT-4212: FileWritingMessageHandler.flushWhenIdle

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4212

Add an option to flush after the `flushInterval`, regardless of intermediate writes.

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -142,6 +142,8 @@ public class FileWritingMessageHandler extends AbstractReplyProducingMessageHand
 	private volatile int bufferSize = DEFAULT_BUFFER_SIZE;
 
 	private volatile long flushInterval = DEFAULT_FLUSH_INTERVAL;
+
+	private volatile boolean flushWhenIdle = true;
 
 	private volatile ScheduledFuture<?> flushTask;
 
@@ -303,10 +305,25 @@ public class FileWritingMessageHandler extends AbstractReplyProducingMessageHand
 	 * {@code flushInterval} and {@code flushInterval * 1.33} with an average of
 	 * {@code flushInterval * 1.167}.
 	 * @param flushInterval the interval.
+	 * @see #setFlushWhenIdle(boolean)
 	 * @since 4.3
 	 */
 	public void setFlushInterval(long flushInterval) {
 		this.flushInterval = flushInterval;
+	}
+
+	/**
+	 * Determine whether the {@link #setFlushInterval(long) flushInterval} applies only
+	 * to idle files (default) or whether to flush on that interval after the first
+	 * write to a previously flushed or new file.
+	 * @param flushWhenIdle false to flush on the interval after the first write
+	 * to a closed file.
+	 * @see #setFlushInterval(long)
+	 * @see #setBufferSize(int)
+	 * @since 4.3.7
+	 */
+	public void setFlushWhenIdle(boolean flushWhenIdle) {
+		this.flushWhenIdle = flushWhenIdle;
 	}
 
 	@Override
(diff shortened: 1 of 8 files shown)


src/reference/asciidoc/file.adoc
```diff
@@ -661,6 +661,9 @@ or `FileWritingMessageHandler.MessageFlushPredicate` implementation.
 The predicates are called for each open file.
 See the java docs for these interfaces for more information.
 
+When using `flushInterval`, the interval starts at the last write - the file is flushed only if it is idle for the interval.
+Starting with _version 4.3.7_, and additional property `flushWhenIdle` can be set to `false`, meaning that the interval starts with the first write to a previously flushed (or new) file.
```
artembilan: I have only one question: why just don't do flushing all the time independently of the first/last write ?

My use-case is to write to file in arbitrary time frame, but really flush periodically according to the provided `flushInterval`.
Looks like such a behavior covers both these options in your change...

What is your explanation?
Thanks

garyrussell: Because the flush actually closes the file and removes the state; to do what you suggest would mean separating flushing from closing - we'd need more configuration (e.g. `idleIntervalForClose`) or more logic - flush but don't remove state (unless the flush is because of an idle interval).

I just thought it would just add more complexity for no real reason.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -878,9 +895,12 @@ private static void rename(File source, File target) throws IOException {
 
 		private volatile long lastWrite;
 
+		private volatile long lastFlush;
+
 		FileState(BufferedWriter writer) {
 			this.writer = writer;
 			this.stream = null;
+			this.lastFlush = System.currentTimeMillis();
```
artembilan: `lastFlush` is never updated.
Am I missing something?

Also I wonder if we should update `MessageFlushPredicate.shouldFlush()` for this new `lastFlush` argument as well.

src/reference/asciidoc/file.adoc
artembilan: Thank you! 
I didn't realize that call `close()` .
Although shouldn't we say something in "What's new"?

garyrussell: I don't think it's a big enough feature for a what's new entry; the existing doc already suggests looking at the javadocs.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -878,9 +895,12 @@ private static void rename(File source, File target) throws IOException {
 
 		private volatile long lastWrite;
 
+		private volatile long firstWrite;
```
artembilan: Since this property is set only from ctor and is never update, I guess it should be `final`.
And from here we realize that `FileState(BufferedOutputStream)` ctor should be supplied with population of this property as well.

No? Am I missing anything else?

src/reference/asciidoc/file.adoc
artembilan: Ok. I think `MessageTriggerAction` is an answer to my periodic flush use-case.

At the same time the question about a new `firstWrite` argument for the `shouldFlush()` method remains.

Plus I guess we should supply a new `flushWhenIdle()` option for the `FileWritingMessageHandlerSpec`.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
garyrussell: That was sloppy of me; good catch; thanks.

artembilan:
OK. I think now it is good for merging.
Those my two concerns about DSL and `FlushPredicate` changes should be addressed in the separate PR.
Looks like this one is intended for backporting to `4.3.x`.

Any objections ?

garyrussell:
Sounds good; perhaps in 5.0 we can add an overloaded `default` method to the `FlushPredicate` with the additional parameter.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/5f450ed9596a8e6741a251d6139eb686e2ce1fd6 and cherry-picked.

Now the `master` should be supplied with Java DSL addition and `FlushPredicate` should be reconsidered for one more `firstWrite` argument.

garyrussell:
Thanks; will do.

PR #2036, opened 2017-01-23
INT-4212: FWMH Predicates and DSL: flushWhenIdle

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4212

Add `firstWrite` to predicate methods.
Add DSL support for `flushWhenIdle`

While these are breaking API changes on the interfaces, I feel this is
likely not going to affect many users and this is a major release after
all.

I have added a note to the 5.0 migration guide.


artembilan (approved):
LGTM
Merging.
And fixed camel case for method param in the Migration Guide note

PR #2037, opened 2017-01-24
Fix Compiled SpEL Test

garyrussell:
Several problems:

- Setting the system property is not enough since the `static` parser already has its configuration
- Changing the ParametersWrapper to `private` made property accessing not compilable (the method and class must be public).
- The expression...

```
 #target.messageAndHeader(message, headers['number'] != null ? headers['number'] : T(org.springframework.util.Assert).isTrue(false, 'required header not available: number'))
```

...is not compilable anyway because else part of the ternery is not compilable because the `TypeReference.exitTypeDescriptor` is `null` (assert call).

(My mistake - I originally thought it was the `NullLiteral`.

So, in effect, the compiledSpEL test is invalid.

This commit addresses the first two bullets.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -979,7 +979,7 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 		}
 	}
 
-	private static class ParametersWrapper {
+	public static class ParametersWrapper {
 
 		private final Object payload;
 
--- a/spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
@@ -576,17 +576,17 @@ public class MethodInvokingMessageProcessorTests {
 	@Test
 	public void testPerformanceSpelVersusInvocable() throws Exception {
 		AnnotatedTestService service = new AnnotatedTestService();
-		Method method = service.getClass().getMethod("messageAndHeader", Message.class, Integer.class);
+		Method method = service.getClass().getMethod("integerMethod", Integer.class);
 
 		MethodInvokingMessageProcessor processor = new MethodInvokingMessageProcessor(service, method);
 		processor.setUseSpelInvoker(true);
 
-		Message<String> message = MessageBuilder.withPayload("foo").setHeader("number", 42).build();
+		Message<Integer> message = MessageBuilder.withPayload(42).build();
 
 		StopWatch stopWatch = new StopWatch("SpEL vs Invocable Performance");
 
 		stopWatch.start("SpEL");
-		int count = 10_000;
+		int count = 20_000;
 		for (int i = 0; i < count; i++) {
 			processor.processMessage(message);
 		}
@@ -613,8 +613,6 @@ public class MethodInvokingMessageProcessorTests {
 			processor.processMessage(message);
 		}
 		stopWatch.stop();
-
-		System.clearProperty("spring.expression.compiler.mode");
 
 		logger.warn(stopWatch.prettyPrint());
 	}
@@ -727,8 +725,7 @@ public class MethodInvokingMessageProcessorTests {
 
 	}
 
-	@SuppressWarnings("unused")
-	private static class AnnotatedTestService {
+	public static class AnnotatedTestService {
 
 		AnnotatedTestService() {
 			super();
--- a/spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
@@ -40,13 +40,16 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import org.springframework.beans.DirectFieldAccessor;
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.expression.spel.SpelCompilerMode;
 import org.springframework.expression.spel.SpelEvaluationException;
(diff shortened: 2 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -979,7 +979,7 @@ private void setExclusiveTargetParameterType(TypeDescriptor targetParameterType,
 		}
 	}
 
-	private static class ParametersWrapper {
+	public static class ParametersWrapper {
```
artembilan: Thank you for such a good catch, @garyrussell ! 😄 

Really appreciate.

Previously this class wasn't `static`. 
But there is really nothing to make it tied with an outer class and the `static` is better for instance.

So, right now making it as a `public` (because we need it any way) makes me think that we should poll this class to the package level and bring it better public name. Just in case some one some where will be interested in similar abstraction.

WDYT?

Otherwise yes, LGTM.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
garyrussell: I am not so sure we should make it top level; it's pretty specialized.

Also, I just noticed that `HandlerMethod.getExpression()` is no longer needed since you changed the code to use the field directly.

Doesn't IDEA warn you of unused methods - time to switch to STS ? 😄

artembilan: OK. I can live with that.

Ha-ha. Yes, it really doesn't point to such a violation. Only to the unused fields.

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
```diff
@@ -583,26 +586,30 @@ public void testPerformanceSpelVersusInvocable() throws Exception {
...
-		for (int i = 0; i < 10000; i++) {
+		for (int i = 0; i < count; i++) {
 			processor.processMessage(message);
 		}
 		stopWatch.stop();
 
+		// Update the parser configuration compiler mode
+		SpelParserConfiguration config = TestUtils.getPropertyValue(processor,
+				"delegate.handlerMethod.EXPRESSION_PARSER.configuration", SpelParserConfiguration.class);
+		new DirectFieldAccessor(config).setPropertyValue("compilerMode", SpelCompilerMode.IMMEDIATE);
```
artembilan: Well, since according to your comment the SpEL of this test isn't compilable, we might should consider to switch to something really compilable.
When I use `messageOnly()` method I have this result:
```
00202  037%  SpEL
00187  034%  Invocable
00164  030%  Compiled SpEL
```
So, somehow for the whole `Message<>` SpEL is faster 😄 

When I use `integerMethod()` I get this:
```
00179  051%  SpEL
00068  020%  Invocable
00101  029%  Compiled SpEL
```

So, looks like there is some performance bottleneck when we really use  `Message` parameter...

The `System.clearProperty("spring.expression.compiler.mode");` must be removed from this test now. Redundant.

garyrussell: Interesting; thankfully it's not a large difference so I don't think we need to get into the business of auto-selecting based on arg types.

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
garyrussell: One of these days, both IDEs will catch up with each other and we won't have to taunt each other with these comments 😄

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
artembilan: OK. Merging with the hope that @aclement will relax `exitTypeDescriptor` for `null` type.

Although I debug locally what may be a problem in the `ArgumentResolver`s for the `Message` parameter.

garyrussell: `integerMethod` is not compilable either, because `AnnotatedTestService` is not `public`.

garyrussell: Yikes!

```
-----------------------------------------
ms     %     Task name
-----------------------------------------
00180  082%  SpEL
00030  014%  Invocable
00010  005%  Compiled SpEL
```

artembilan: I found one of the performance bottleneck - `InvocableHandlerMethod` : 
```
private Object[] getMethodArgumentValues(Message<?> message, Object... providedArgs) throws Exception {
	MethodParameter[] parameters = getMethodParameters();
        ...
	for (int i = 0; i < parameters.length; i++) {
		MethodParameter parameter = parameters[i];
	...
		GenericTypeResolver.resolveParameterType(parameter, getBean().getClass());

```
The `GenericTypeResolver.resolveParameterType()` always does the same for the same `MethodParameter` object for nothing on each method invocation!

I did some local hacking moving such a initialization to the ctor and increase performance for `1_000_000` from:
```
02284  085%  SpEL
00339  013%  Invocable
00056  002%  Compiled SpEL
```
to
```
02243  087%  SpEL
00262  010%  Invocable
00071  003%  Compiled SpEL
```
Yeah... not a big deal but still.
I guess the rest of time we spend on the reflection method invocation instead of direct call from compiled SpEL...

Merging and opening JIRA against SF at least to see what guys think

artembilan:
For back tracking: https://jira.spring.io/browse/SPR-15186

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/eaaa21ef5d769452587d739d99c81d8a9f8103fd

PR #2038, opened 2017-01-24
INT-4215: CompositeFileListFilter runs all files through all filters

afgrant:
Files filtered out by earlier filters should not be passed to later filters.

https://jira.spring.io/browse/INT-4215


artembilan:
Superseded with #2039 

Thank you very much, @afgrant, for the contribution such a nice scenario to invesitgate!

Issue #1998, opened 2016-12-09
how to use .done to read file in diractory?


opticaline:
i have test.txt in diractory,
only when test.txt.done or test.done file in diractory start reading test.txt.

artembilan:
I'm not sure in what context you mean.

Would you mind to shad more light what's going on? 

OTOH the StackOverflow is a place for similar questions. 

GH issues and JIRA are for bug tickets or new requests.

garyrussell:
You need to use a custom `FileListFilter` that filters files that don't have the appropriate `.done` file present.

opticaline:
@garyrussell 
Hi, this is my filter.
`
private final AntPathMatcher matcher = new AntPathMatcher();

    private String doneFile = ".done";
    private String pattern;

    public DoneFileFilter(String pattern, String doneFile) {
        this.pattern = pattern;
        this.doneFile = doneFile.startsWith(".") ? doneFile : "." + doneFile;
    }

    @Override
    public List<File> filterFiles(File[] files) {
        List<File> fileList = new ArrayList<>();
        for (File file : files) {
            System.out.println(file);
            String name = file.getName();
            if (matcher.match(pattern, name)) {
                File f = new File(file.getParent(), name + doneFile);
                if (f.exists()) {
                    fileList.add(f);
                }
            }
        }
        System.out.println(fileList);
        return null;
    }
`

but is now working.
Is there anything I do wrong?

garyrussell:
I assume you mean __not__ not __now__.

On the face of it, it looks ok to me; you need to debug it. If you can't figure it out, you need to show all your configuration.

opticaline:
@garyrussell 
Hi, this is my full configuration.

```
@SpringBootApplication
public class FileReadingJavaApplication {

    public static void main(String[] args) {
        new SpringApplicationBuilder(FileReadingJavaApplication.class)
                .web(false)
                .run(args);
    }

    @Bean
    public IntegrationFlow fileReadingFlow() {
        return IntegrationFlows
                .from(s -> s.file(new File("D:\\data\\input"))
                                .useWatchService(true).watchEvents(FileReadingMessageSource.WatchEventType.CREATE)
                                .filter(new DoneFileFilter("*.txt", "ok")),
                        e -> e.poller(Pollers.fixedDelay(1000)))
                .split(new FileSplitter(true))
                .handle(new GenericHandler<Object>() {
                    @Override
[...]

PR #2039, opened 2017-01-25
INT-4215: Add `ChainFileListFilter`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4215
Fixes GH-1998 (https://github.com/spring-projects/spring-integration/issues/1998)

Just make an `CompositeFileListFilter` extension which chains result of the previous filter to the next

**Cherry-pick 4.3.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
```diff
@@ -42,7 +42,7 @@
  */
 public class CompositeFileListFilter<F> implements ReversibleFileListFilter<F>, ResettableFileListFilter<F>, Closeable {
 
-	private final Set<FileListFilter<F>> fileFilters;
+	protected final Set<FileListFilter<F>> fileFilters;
```
garyrussell: // NOSONAR

will add on merge.

artembilan: Thanks.
I have never understand that rule.
How can we mutate the property if it is `final`.
That's why I made it `protected` instead of getter.

garyrussell: Right it's bogus but I don't see a way on Sonar to suppress only for final fields.

garyrussell:
Fix with doc polish here: https://github.com/garyrussell/spring-integration/commit/6f8cdc17cac27ae1a61273a5f45d892b7518ee5d

Please check.

garyrussell:
Merged as d47bca48cc50dfa79735839d6ea51a31840cf977 cherry-picked as 35a633d3049ec1667069efea9671686677f951b9 and fix for empty list pushed as 1674cd0f85d44ec688f675f22276dc04c97c283d

PR #2040, opened 2017-01-26
INT-4216: Fix occurrence of ClassCastException in ChainFileListFilter

afgrant:
https://jira.spring.io/browse/INT-4216

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/ChainFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/ChainFileListFilter.java
@@ -45,7 +45,7 @@ public class ChainFileListFilter<F> extends CompositeFileListFilter<F> {
 				break;
 			}
 			@SuppressWarnings("unchecked")
-			F[] fileArray = leftOver.toArray((F[]) Array.newInstance(files[0].getClass(), leftOver.size()));
+			F[] fileArray = leftOver.toArray((F[]) Array.newInstance(leftOver.get(0).getClass(), leftOver.size()));
 			leftOver = fileFilter.filterFiles(fileArray);
 		}
 		return leftOver;
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/ChainFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/ChainFileListFilter.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.file.filters;
 
+import java.lang.reflect.Array;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


artembilan:
Why do we need this?
Where is a description?
If it is failing it doesn't deserve PullRequest.
The Gist with a description is enough.

Thanks

afgrant:
Sorry for not adding a description, my impression was that the JIRA issue number in the title was the crucial part.  The discussion is here, or I could create a new issue if that's preferable: https://jira.spring.io/browse/INT-4215

I was doing my best to follow `CONTRIBUTING` which didn't mention a Gist.  My intention was that additional commits would bring the status back to green, then all could be squashed and rebased into whichever branch you need.

artembilan:
Well, OK. 
If that isn't clear in the CONTRIBUTING, please, consider PullRequest as some proposal for changes back to the core code. 
If the test just claims the problem it doesn't mean to be a part of the `master` code.
Yes, I see the problem with generics erasure during subsequent cast, but I still don't see the solution.

If you can share it here in the PR, I'll accept it via merge.
Otherwise we will be forced to closed it in favor of something else like that was with previous your one.

Right we spend time right now for nothing, because we really have a problem, but that really would be better do not PR just failing tests. Although, of course, there might be arguments to merge other passing tests because we are missing them in code base any way 😄 

Nevertheless, thank you very much for such a thorough investigation around feature.
And we'll very appreciate if you share a proper solution here or in that JIRA.

artembilan:
@afgrant , 

OK, thank you for the PR!

Please, follow with the CONTRIBUTION and provide the proper solution and commit message mentioning JIRA https://jira.spring.io/browse/INT-4216.

You have to add `@Copyright` and `@since 4.3.8`  to a new class etc.

artembilan:
BTW, we can do that ourselves on merge.
Sorry to nit-picking.

Please, just fix `ChainFileListFilter` like @garyrussell describes in the JIRA  - using `leftOver` instead, - and this PR will be ready for merge.

afgrant:
OK, hopefully this meets your needs pretty closely and will only need minor edits after squash.  Travis-CI should be caught up soon.

Let me know if anything else is needed and thanks for your help.

artembilan (approved):

artembilan:
Merged and cherry-picked as https://github.com/spring-projects/spring-integration/commit/1f9d07ae2414e9df49a993651b247c4707cc2162 

Pay attention how I have moved `ChainFileListFilterIntegrationTests` and `CompositeFileListFilterTests` to the `filters` package. Just for consistency.

@afgrant , thank you for contribution!

Looking forward for more!

PR #2041, opened 2017-01-27
INT-4217: SpEL Compilable for Required Header Args

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4217

Previously, SpEL method calls with required header parameters had the following form:

    #target.messageAndHeader(message, headers['number'] != null ? headers['number'] : T(org.springframework.util.Assert).isTrue(false, 'required header not available: number'))

The SpEL compiler cannot compile this because the else clause of the ternary has no `exitDescriptor` to indicate the type.

Change the expression to use a function for required headers.
Also use an Elvis operator when possible.

Some examples of new expressions:

    #target.optionalAndRequiredHeader(headers['prop'] ?: null, #requiredHeader(headers, 'num'))

    #target.optionalAndRequiredDottedHeader(headers['dot1'] != null ? headers['dot1'].foo : null,
         #requiredHeader(headers, 'dot2').baz

In the second case, we can't use an Elvis because we're accessing a `foo` property of the header.

__cherry-pick to 4.3.x__

- make `ParametersWrapper` `static`
- minor conflicts in imports
- remove the perf test

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -31,6 +31,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
@@ -314,7 +315,7 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 		this.displayString = sb.toString() + "]";
 	}
 
-	private void prepareEvaluationContext() {
+	private void prepareEvaluationContext() throws Exception {
 		StandardEvaluationContext context = getEvaluationContext(false);
 		Class<?> targetType = AopUtils.getTargetClass(this.targetObject);
 		if (this.method != null) {
@@ -333,6 +334,8 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 			context.registerMethodFilter(targetType, filter);
 		}
 		context.setVariable("target", this.targetObject);
+		context.registerFunction("requiredHeader", ParametersWrapper.class.getDeclaredMethod("getHeader",
+				Map.class, String.class));
 	}
 
 	private boolean canReturnExpectedType(AnnotatedMethodFilter filter, Class<?> targetType,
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -949,12 +952,16 @@ else if (valueAttribute.indexOf('.') != -1) {
...
 			String headerRetrievalExpression = "headers['" + headerName + "']";
 			String fullHeaderExpression = headerRetrievalExpression + relativeExpression;
-			String fallbackExpression = (annotationAttributes.getBoolean("required")
-					&& !methodParameter.getParameterType().getName().equals("java.util.Optional"))
-					? "T(org.springframework.util.Assert).isTrue(false, 'required header not available: "
-					+ headerName + "')"
-					: "null";
-			return headerRetrievalExpression + " != null ? " + fullHeaderExpression + " : " + fallbackExpression;
+			if (annotationAttributes.getBoolean("required")
+					&& !methodParameter.getParameterType().equals(Optional.class)) {
```
artembilan: Seems for me we can't backport this to `4.3.x`.
I see your comment in the PR description, but it doesn't include this change

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
```diff
@@ -348,6 +349,110 @@ public void multipleHeadersWithAnnotatedMethod() throws Exception {
...
+	@Test
+	public void compiledOptionalAndRequiredWithAnnotatedMethod() throws Exception {
+		AnnotatedTestService service = new AnnotatedTestService();
+		Method method = service.getClass().getMethod("optionalAndRequiredHeader", String.class, Integer.class);
+		MethodInvokingMessageProcessor processor = new MethodInvokingMessageProcessor(service, method);
+		processor.setUseSpelInvoker(true);
+		DirectFieldAccessor compilerConfigAccessor = compileImmediate(processor);
+		optionalAndRequiredWithAnnotatedMethodGuts(processor, true);
+		assertNotNull(TestUtils.getPropertyValue(processor, "delegate.handlerMethod.expression.compiledAst"));
+		optionalAndRequiredWithAnnotatedMethodGuts(processor, true);
```
artembilan: Why do we have this call again?

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
```diff
@@ -757,6 +869,11 @@ public String optionalAndRequiredHeader(@Header(required = false) String prop,
 			return prop + num;
 		}
 
+		public String optionalAndRequiredDottedHeader(@Header(name = "dot1.foo", required = false) String prop,
+				@Header(name = "dot2.baz", required = true) Integer num) {
```
artembilan: redundant

And as you see IDEA shows unused methods.
So, that was strange with that getter before...

artembilan (changes requested):
LGTM generally.

Just a couple minor concerns.

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
garyrussell: Sorry - I meant to mention that in the backport notes - revert `Optional` to compare the class name.

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
artembilan: OK. Now I see complex logic in the target `Guts` method.

Thanks.
So, merging...

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
garyrussell: I copied this from another test; feel free to remove if it annoys you 😄

garyrussell: Oh, this __is__ the original test.

artembilan: He-he!
Just wanted to continue our best IDE holly-war 😉

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/c5c874f5635cb68383e3b84fac97e8f2a69e1423 and cherry-picked as https://github.com/spring-projects/spring-integration/commit/d9c6ffee746c1899ef321eaa8a2d2d79d1d66251 after resolving some conflicts.
See commit message

PR #2048, opened 2017-02-02
INT-4218: MessagingMethodInvokerHelp improvements

artembilan:
JIRA: https://jira.spring.io/browse/INT-4218

* Add `InvocableHandlerMethod` invocation threshold logic to give up eventually in favor of permanent expression evaluation
* Add `ParametersWrapper.toString()` for better logging messages
* Log `InvocableHandlerMethod` invocation failure message in favor of SpEL only once when `failedAttempts` is exceeded already
* Switch to `spelOnly` mode after that
* Refactor `MessagingMethodInvokerHelper.processInternal()` to separate SpEL or Ivocable logic via their own invocation methods
* Move `MessagingMethodInvokerHelper.HandlerMethod` static fields to the `MessagingMethodInvokerHelper` level since it doesn't matter from this class perspective but can be reused in other places from top level of the `MessagingMethodInvokerHelper` class
* Catch `IllegalArgumentException` with the `java.lang.ClassCastException@...` message.
See http://stackoverflow.com/questions/16042591/reflections-illegalargumentexception-causes for more info

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -440,17 +440,18 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 				}
 			}
 			else if (e instanceof IllegalStateException) {
-				if (e.getCause() instanceof IllegalArgumentException
-						&& (!"argument type mismatch".equals(e.getCause().getMessage())
+				if (!(e.getCause() instanceof IllegalArgumentException) ||
+						!e.getStackTrace()[0].getClassName().equals(InvocableHandlerMethod.class.getName()) ||
+						(!"argument type mismatch".equals(e.getCause().getMessage()) &&
 						// JVM generates GeneratedMethodAccessor### after several calls with less error checking
-						&& !e.getCause().getMessage().startsWith("java.lang.ClassCastException@"))) {
+						!e.getCause().getMessage().startsWith("java.lang.ClassCastException@"))) {
 					throw e;
 				}
 			}
 
 			Expression expression = handlerMethod.expression;
 
-			if (++handlerMethod.failedAttempts == FAILED_ATTEMPTS_THRESHOLD) {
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -408,27 +441,26 @@ private T processInternal(ParametersWrapper parameters) throws Exception {
...
+			Expression expression = handlerMethod.expression;
 
-		if (result != null && this.expectedType != null) {
-			return (T) getEvaluationContext(true)
-					.getTypeConverter()
-					.convertValue(result, TypeDescriptor.forObject(result), this.expectedType);
-		}
-		else {
-			return result;
+			if (++handlerMethod.failedAttempts == FAILED_ATTEMPTS_THRESHOLD) {
```
garyrussell: Must be `>=` (race condition)

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -408,27 +441,26 @@ private T processInternal(ParametersWrapper parameters) throws Exception {
 			}
 			else if (e instanceof IllegalStateException) {
 				if (e.getCause() instanceof IllegalArgumentException
-						&& !"argument type mismatch".equals(e.getCause().getMessage())) {
+						&& (!"argument type mismatch".equals(e.getCause().getMessage())
+						// JVM generates GeneratedMethodAccessor### after several calls with less error checking
+						&& !e.getCause().getMessage().startsWith("java.lang.ClassCastException@"))) {
 					throw e;
 				}
 			}
```
garyrussell: I think there are still some issues with this exception logic and falling back to SpEL.

If the method was invoked correctly but the user code threw the exception (e.g `IllegalStateException`) we'd retry with SpEL (and we shouldn't).

We need to (somehow) more reliably detect that the exception was thrown by the `InvocableHandlerMethod` itself.

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
Merged as 9839b5fcaac43aadf8c87fa8723ec13c5e7346a1

PR #2049, opened 2017-02-06
INT-4115: (S)FtpPersistentFileFilter by default

artembilan:
JIRA: https://jira.spring.io/browse/INT-4115

Apply `(S)FtpPersistentAcceptOnceFileListFilter` for the `(S)FtpInboundFileSynchronizer` by default to avoid cases to sync the same remote files to the local directory again.
Especially when `localFileName` strategy is applied and we end up with new local files, but with the same remote content

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
@@ -51,8 +51,7 @@ public abstract class AbstractRemoteFileInboundChannelAdapterParser extends Abst
 
 		// configure the InboundFileSynchronizer properties
 		BeanDefinition expressionDef = IntegrationNamespaceUtils.createExpressionDefinitionFromValueOrExpression(
-				"remote-directory", "remote-directory-expression", parserContext,
-				element, false);
+				"remote-directory", "remote-directory-expression", parserContext, element, false);
 		if (expressionDef != null) {
 			synchronizerBuilder.addPropertyValue("remoteDirectoryExpression", expressionDef);
 		}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
@@ -25,6 +25,7 @@ import org.springframework.beans.factory.xml.ParserContext;
 import org.springframework.integration.config.ExpressionFactoryBean;
 import org.springframework.integration.config.xml.AbstractPollingInboundChannelAdapterParser;
 import org.springframework.integration.config.xml.IntegrationNamespaceUtils;
+import org.springframework.integration.file.filters.AbstractPersistentAcceptOnceFileListFilter;
 import org.springframework.integration.file.filters.FileListFilter;
 import org.springframework.integration.file.remote.synchronizer.InboundFileSynchronizer;
 import org.springframework.util.StringUtils;
@@ -36,6 +37,7 @@ import org.springframework.util.StringUtils;
  * @author Mark Fisher
  * @author Gary Russell
  * @author Artem Bilan
+ *
  * @since 2.0
  */
 public abstract class AbstractRemoteFileInboundChannelAdapterParser extends AbstractPollingInboundChannelAdapterParser {
@@ -49,7 +51,8 @@ public abstract class AbstractRemoteFileInboundChannelAdapterParser extends Abst
 
 		// configure the InboundFileSynchronizer properties
 		BeanDefinition expressionDef = IntegrationNamespaceUtils.createExpressionDefinitionFromValueOrExpression(
-				"remote-directory", "remote-directory-expression", parserContext, element, false);
+				"remote-directory", "remote-directory-expression", parserContext,
+				element, false);
 		if (expressionDef != null) {
 			synchronizerBuilder.addPropertyValue("remoteDirectoryExpression", expressionDef);
 		}
@@ -59,8 +62,10 @@ public abstract class AbstractRemoteFileInboundChannelAdapterParser extends Abst
 		String remoteFileSeparator = element.getAttribute("remote-file-separator");
 		synchronizerBuilder.addPropertyValue("remoteFileSeparator", remoteFileSeparator);
 		IntegrationNamespaceUtils.setValueIfAttributeDefined(synchronizerBuilder, element, "temporary-file-suffix");
+
 		FileParserUtils.configureFilter(synchronizerBuilder, element, parserContext,
-				getSimplePatternFileListFilterClass(), getRegexPatternFileListFilterClass());
+				getSimplePatternFileListFilterClass(), getRegexPatternFileListFilterClass(),
+				getPersistentAcceptOnceFileListFilterClass());
 
 		// build the MessageSource
 		BeanDefinitionBuilder messageSourceBuilder =
@@ -93,5 +98,7 @@ public abstract class AbstractRemoteFileInboundChannelAdapterParser extends Abst
 	protected abstract Class<? extends FileListFilter<?>> getSimplePatternFileListFilterClass();
 
 	protected abstract Class<? extends FileListFilter<?>> getRegexPatternFileListFilterClass();
+
+	protected abstract Class<? extends AbstractPersistentAcceptOnceFileListFilter<?>> getPersistentAcceptOnceFileListFilterClass();
 
 }
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
@@ -20,12 +20,16 @@ import org.w3c.dom.Element;
 
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.support.ManagedList;
 import org.springframework.beans.factory.xml.ParserContext;
 import org.springframework.integration.config.xml.IntegrationNamespaceUtils;
 import org.springframework.integration.file.DefaultFileNameGenerator;
+import org.springframework.integration.file.filters.AbstractPersistentAcceptOnceFileListFilter;
+import org.springframework.integration.file.filters.CompositeFileListFilter;
 import org.springframework.integration.file.filters.ExpressionFileListFilter;
 import org.springframework.integration.file.filters.FileListFilter;
 import org.springframework.integration.file.remote.RemoteFileOperations;
+import org.springframework.integration.metadata.SimpleMetadataStore;
 import org.springframework.util.StringUtils;
 
 /**
@@ -34,6 +38,7 @@ import org.springframework.util.StringUtils;
  * @author David Turanski
  * @author Gary Russell
  * @author Artem Bilan
+ *
  * @since 3.0
  *
  */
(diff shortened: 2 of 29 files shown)


garyrussell:
I am a bit confused. INT-4115 is about the `AbstractRemoteFile__Streaming__MessageSource` and this PR does not address that.

Further, when configuring with XML, we have pattern (and filter) attributes.

Don't we need to combine this into a composite (or chain) filter when configured with XML?

artembilan:
Oh! Really?
I have missed that 😄 

But is it still applied for the `AbstractInboundFileSynchronizer` impl? Isn't it?

Re, XML parsing we have this:
```
parserContext.getReaderContext().error("at most one of 'filename-pattern', " +
		"'filename-regex', 'filter' or 'filter-expression' is allowed on remote file inbound adapter",
						element);
```
So, I think no, we shouldn't add more complexity there.
More over the local part is only like this:
```
IntegrationNamespaceUtils.setReferenceIfAttributeDefined(messageSourceBuilder, element, "local-filter");
IntegrationNamespaceUtils.setValueIfAttributeDefined(messageSourceBuilder, element, "local-directory");
```
So, we still have to build composition manually here.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -165,6 +168,10 @@ public void setRemoteDirectoryExpression(Expression remoteDirectoryExpression) {
 	 * @param filter the file list filter.
 	 */
 	public void setFilter(FileListFilter<F> filter) {
+		doSetFilter(filter);
+	}
+
+	public void doSetFilter(FileListFilter<F> filter) {
```
garyrussell: Did you intend to make this `final` ?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -109,6 +111,10 @@ public void setRemoteFileSeparator(String remoteFileSeparator) {
 	 * @param filter the file list filter.
 	 */
 	public void setFilter(FileListFilter<F> filter) {
+		doSetFilter(filter);
+	}
+
+	public void doSetFilter(FileListFilter<F> filter) {
```
garyrussell: `final` ?

garyrussell (changes requested):
I am still a bit concerned by this; I would guess that most people apply a pattern filter of some kind, which effectively removes this new filter. I would say we should always combine it in a composite filter. Otherwise, this change won't be applied in most cases.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
artembilan: Hm... Seems for me I was going to make them `private`.
Does it make sense?

garyrussell: but these are in the abstract class and called from the subclasses.

`protected final ...` ??

artembilan: Doh. Yeah, really `protected final`.
That was already some time ago when this PR has been fresh in my mind. 😄

artembilan:
> I am still a bit concerned by this; I would guess that most people apply a pattern filter of some kind, which effectively removes this new filter. I would say we should always combine it in a composite filter. Otherwise, this change won't be applied in most cases.

OK. I think I see your point. 
And the idea is to use something like `FileListFilterFactoryBean` what we have from the `FileInboundChannelAdapterParser`.

Not sure that we always have to add it into the composite, but at least we really can do that when `regexp` or `pattern` is in use. For any other custom case there is always `CompositeFileListFilter` to configure externally and inject.

That is already like an improvement, next steps, to the current fix.
I think it is still valid to have it like thing by default.

garyrussell:
>I think it is still valid to have it like thing by default.

I agree; my concern is that it won't actually be a default in most cases, because people generally use pattern or regex.

spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
```diff
@@ -49,7 +51,8 @@ protected final BeanMetadataElement parseSource(Element element, ParserContext p
 
 		// configure the InboundFileSynchronizer properties
 		BeanDefinition expressionDef = IntegrationNamespaceUtils.createExpressionDefinitionFromValueOrExpression(
-				"remote-directory", "remote-directory-expression", parserContext, element, false);
+				"remote-directory", "remote-directory-expression", parserContext,
+				element, false);
```
garyrussell: I am just wondering what drove this change - unnecessary reformatting causes loss of blame.

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
```diff
@@ -118,7 +124,7 @@ static void configureFilter(BeanDefinitionBuilder synchronizerBuilder, Element e
 			}
 			if (count != 1) {
 				parserContext.getReaderContext().error("at most one of 'filename-pattern', " +
-						"'filename-regex', 'filter' or 'filter-expression' is allowed on remote file inbound adapter",
+								"'filename-regex', 'filter' or 'filter-expression' is allowed on remote file inbound adapter",
```
garyrussell: Bad indent - pushed past 120.

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
```diff
@@ -132,16 +138,44 @@ else if (hasFilterExpression) {
 				synchronizerBuilder.addPropertyValue("filter", expressionFilterBeanDefinition);
 			}
 			else if (hasFileNamePattern) {
-				BeanDefinitionBuilder filterBuilder = BeanDefinitionBuilder.genericBeanDefinition(patternClass);
-				filterBuilder.addConstructorArgValue(fileNamePattern);
-				synchronizerBuilder.addPropertyValue("filter", filterBuilder.getBeanDefinition());
+				BeanDefinition patterFilter =
```
garyrussell: `s/patter/pattern/`

spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
```diff
@@ -94,4 +99,6 @@ protected final BeanMetadataElement parseSource(Element element, ParserContext p
 
 	protected abstract Class<? extends FileListFilter<?>> getRegexPatternFileListFilterClass();
 
+	protected abstract Class<? extends AbstractPersistentAcceptOnceFileListFilter<?>> getPersistentAcceptOnceFileListFilterClass();
```
garyrussell: Perhaps return null by default? (Allow subclasses, e.g. AWS to not provide a filter).

Maybe not - since this is a major release, I guess it's ok.

garyrussell (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
artembilan: The line exceeds 120 symbols.

I feel like there might be a lot of code which we won't touch in the future at all.
But its problem that it might be inconsistent with our code style.
So, if we request 120 line length we should prove that with all our code.
Otherwise we may meet some nasty commentators that we don't do that in our own code. 😄 

I know that you don't like such a code style reformatting for nothing, but sometimes doing that via IDEA tool I meet some interesting violations, like extra whitespaces in the code line, missed blank lines around class members etc...

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
artembilan: I wish my IDEA has the same formatting settings as your STS, but I really don't know how to configure this indent settings....

spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
garyrussell: ?? The previous line length was 98

I have no issue when the reformat is to comply with our style but this doesn't fit that rule.

Hence my question.

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
artembilan: @snicoll ,

any advice for IntelliJ IDEA?
Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
garyrussell: And the line above is clearly longer (113).

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
garyrussell: But the extra indent pushed it past 120 - why did you even touch this line? - do you have some "auto-format on save"

spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileInboundChannelAdapterParser.java
artembilan: Doh...
In new IDEA it looks like:
idea_120

The visual perception deceived me... 😢 

My bad, sorry.
Will be much careful in the future.
Reverting...

garyrussell: Sometimes these IDEs try to be too smart 😄 - if it adds metadata, it should push the margin out appropriately for those lines.

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileParserUtils.java
artembilan: I used the reformatting tool manually (Ctrl+Alt+L) and it is applied for the whole code.

Again: I don't know how to make it only for the touch code and don't know how to not miss violations in other lines.

I guess if I have the proper setting for these kind of indents, even if I perform that reformatting, this line of code wouldn't be touch and we wouldn't have such a long discussion....

garyrussell: In STS, Ctrl-Shift-F (with nothing selected) reformats the whole file, but if you just select the lines you are interested in it only reformats the selected lines. I try to avoid Ctrl-Shift-F without making a selection first.

I wouldn't mind (although as you know I dislike unnecessary formatting), but in this case, it pushed it past 120, which is bad.

snicoll: we have the same issue in Spring Boot. @philwebb was working on a generic formatter solution but this had to be put on hold to work on other things. It's current ad-hoc at the moment and I don't think you can't configure STS and IJ to match perfectly (especially with indent, the algorithm doesn't match for everything).

artembilan: Thank you @snicoll for sharing your pain as  well 😄 

I think we with @garyrussell will work out some formal rules to follow with formatting.

garyrussell:
Merged as 10ce68d3e33e38ecaf109b866fedb2fd0408ab3b

PR #2050, opened 2017-02-07
MMIH: Use original method for InvocableHandMethod

artembilan:
The `MessagingMethodInvokerHelper` extracts `targetClass` to process methods and avoid extra methods when class is Proxy

The `InvocableHandlerMethod` is based on the reflection method invocation and we should use runtime methods on the proxy to invoke.
We don't see such a problem in SpEL because it is parsed from the plain String to the real code later


artembilan:
Tested against `spring-integration-flow` after `gradlew install` from SI and `mavenLocal()` on SIF

garyrussell:
Merged as 8276bf8eae6d189527f7127fa1af9100d56a0faa after adding an `Advice` to the test.

PR #2051, opened 2017-02-11
INT-4225: Add MessageConverters infrastructure

artembilan:
JIRA: https://jira.spring.io/browse/INT-4225

* To allow flexible way to convert incoming messages for the target arguments in method invocation
add `ConfigurableCompositeMessageConverter` global bean under `ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME`
* Retrieve that bean for the `MessageHandlerMethodFactor` in the `MessagingMethodInvokerHelper` to enrich afterwards target `HandlerMethodArgumentResolver`s
* The `ContentTypeConversionTests` demonstrates conversion incoming JOSN `string` into the target POJO argument in the service method

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -43,7 +43,6 @@ import org.springframework.aop.framework.Advised;
 import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.context.Lifecycle;
-import org.springframework.context.annotation.Primary;
 import org.springframework.core.LocalVariableTableParameterNameDiscoverer;
 import org.springframework.core.MethodParameter;
 import org.springframework.core.ParameterNameDiscoverer;
@@ -58,6 +57,7 @@ import org.springframework.expression.Expression;
 import org.springframework.expression.TypeConverter;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
+import org.springframework.integration.annotation.Default;
 import org.springframework.integration.annotation.Payloads;
 import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.integration.context.IntegrationContextUtils;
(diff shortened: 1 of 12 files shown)


artembilan:
If it's OK with you, I can go ahead with the Docs on the matter.

OTOH the main idea that it can be overridden by the target application.
And one of the tricks is to add `MarshallingMessageConverter` from the SI XML module. So, when SI-XML is in CP, we have one more converter.

At the same time SCSt would be able to override it its own way with all those Cloud-aware converters, e.g. `TupleJsonMessageConverter` as it is now there.

garyrussell:
Looks good, although we should probably clearly document that conversion can only occur when there is only one candidate method.

Perhaps we could support `@Primary` and, if we find no matching methods for the payload type we invoke one marked with that annotation?

artembilan:
I like the idea with the `@Primary`! 😄 

Your thoughts about my proposal for automatic `MarshallingMessageConverter`?

garyrussell:
Makes sense (marshalling converter).

garyrussell:
I wonder if it's ok to overload Spring's `@Primary` - the javadocs are wrong for this use case; OTOH  I don't think we want a second annotation - too confusing.

I wonder why travis keeps failing on SFTP.

garyrussell:
Doc polish here - https://github.com/garyrussell/spring-integration/commit/99e029a3adb7b9ae6fc28d3442b4d6f48c4fb547

artembilan:
>  the javadocs are wrong for this use case;

Well, not necessarily. We create a method invocation bean anyway, so that might be still fine in this case to expect `@Primary`  together with the `@ServiceActivator`.

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -545,6 +564,11 @@ else if (!Modifier.isPublic(method1.getModifiers())) {
 				}
 				return;
 			}
+			if (AnnotationUtils.getAnnotation(method1, Default.class) != null) {
```
garyrussell: Should we use `findAnnotation()` here, or do you think one level of meta-annotation is enough for this purpose?

I am leaning that way - having `@Default` on some obscure meta-meta-meta... could be confusing.

garyrussell (commented):
Just that one comment/question and a couple of minor doc polishes which I'll do during merge after you answer.

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
artembilan: Well, my idea to always have it explicitly. If you think the method must be default, so make it like that with declaring `@Default` exactly on place!

garyrussell: Agreed, but get __will__ handle one level of meta; merging...

garyrussell:
Merged as 28e8f23fd0ac04b538aeacdbccd3fbbf90517e91

PR #2052, opened 2017-02-13
INT-4227: Allow custom Messaging Annotations

artembilan:
JIRA: https://jira.spring.io/browse/INT-4227

Expose `MessagingAnnotationPostProcessor` for inheritors.
The `CustomMessagingAnnotationTests` demonstrates how custom Messaging Annotation can be registered and used

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -74,12 +74,12 @@ import org.springframework.util.StringUtils;
 public class MessagingAnnotationPostProcessor implements BeanPostProcessor, BeanFactoryAware,
 		InitializingBean, SmartInitializingSingleton {
 
-	private final Log logger = LogFactory.getLog(this.getClass());
+	protected final Log logger = LogFactory.getLog(this.getClass());
 
-	private final Map<Class<? extends Annotation>, MethodAnnotationPostProcessor<?>> postProcessors =
+	protected final Map<Class<? extends Annotation>, MethodAnnotationPostProcessor<?>> postProcessors =
 			new HashMap<Class<? extends Annotation>, MethodAnnotationPostProcessor<?>>();
 
-	private final MultiValueMap<String, String> lazyLifecycleRoles = new LinkedMultiValueMap<String, String>();
+	protected final MultiValueMap<String, String> lazyLifecycleRoles = new LinkedMultiValueMap<String, String>();
 
 	private ConfigurableListableBeanFactory beanFactory;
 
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java
```diff
@@ -74,12 +74,12 @@
...
 
-	private final Log logger = LogFactory.getLog(this.getClass());
+	protected final Log logger = LogFactory.getLog(this.getClass());
 
-	private final Map<Class<? extends Annotation>, MethodAnnotationPostProcessor<?>> postProcessors =
+	protected final Map<Class<? extends Annotation>, MethodAnnotationPostProcessor<?>> postProcessors =
 			new HashMap<Class<? extends Annotation>, MethodAnnotationPostProcessor<?>>();
 
-	private final MultiValueMap<String, String> lazyLifecycleRoles = new LinkedMultiValueMap<String, String>();
+	protected final MultiValueMap<String, String> lazyLifecycleRoles = new LinkedMultiValueMap<String, String>();
```
garyrussell: `// NOSONAR` on each.

garyrussell: However, I think I'd prefer `protected void addPostProcessor(...)` etc, rather than exposing the internals.

spring-integration-core/src/test/java/org/springframework/integration/config/annotation/CustomMessagingAnnotationTests.java
```diff
@@ -0,0 +1,159 @@
...
+	private LoggingHandler loggingHandler;
+
+	@Autowired
+	private MessageChannel loggingChannel;
+
+	@Test
+	public void testLogAnnotation() {
+		assertNotNull(this.loggingHandler);
+
+		org.apache.commons.logging.Log log =
```
garyrussell: Why not change the interface name to avoid this FQCN noise?

garyrussell (changes requested):

artembilan:
How about this?

There is another idea like request all those `MethodAnnotationPostProcessor` from `BeanFactory` as beans and don't mess with the `MessagingAnnotationPostProcessor` override or explicit declaration at all.

But that requires some breaking changes for the `AbstractMethodAnnotationPostProcessor` design.

OTOH the `MessagingAnnotationPostProcessor` is `BPP` and it might really be better do not interfere `BeanFactory` from that level and just let end-user to override the bean on its own purpose...

garyrussell:
I think this technique is fine.

However, perhaps we should break `pPAI` into smaller chunks and make most methods protected?

artembilan:
> perhaps we should break pPAI into smaller chunks and make most methods protected?

M-m-m. Would you mind to elaborate more? Some simple sample what to change and how.
Thanks

garyrussell:
I am not exactly sure but, let's say someone wants to "enhance" (e.g. synthesize) an annotation.

e.g. Annotation `@Foo` meta-annotated with `@ServiceActivator` and they want to set some attributes on SA.

Currently there's no way to do that.

If we make `getAnnotationChain()` protected, the user could do it there.

Also, as we iterate over the `annotationChains.keySet()`, if the body of the for loop was a protected method, the subclass could take some different action in some cases, and call `super` in others.

artembilan:
OK!

I'll take a look into that.
Thanks

garyrussell (approved):

garyrussell:
Merged as 693adfd774a2c81734573681b620845a96a7696f 

Cherry-pick conflicts with https://github.com/spring-projects/spring-integration/commit/868004e9e4194f59820427fe1db3d44fbb67596c

It's probably easiest for you to fix and create a new PR for the back port. thx

artembilan:
Backport PR #2069

PR #2053, opened 2017-02-13
Introduce MockIntegration Framework

artembilan:


artembilan:
Start looking from the `MockMessageSourceTests` of course.

Any other ideas are welcome!

artembilan:
Interesting fact: when we would like to test some flow when the entry-point is some gateway, which depends on some extra resource or configuration (e.g. AMQP or WS), we can just use `MessagingTemplate.sendAndReceive()` for the same `requestChannel` and assert result for expectations.
I think it worth to mention when we document the testing for Spring Integration apps.

Initially I thought about something like injection Messaging Gateway instead of those protocol-specific inbound endpoints, but then I decided it would be much easier to perform direct action for the channel and rely on the `replyChannel` header for the reply from the flow.

Toparvion:
@artembilan, thank you for initiating this PoC!
It would be great to see how to test particular component(s) of integration flows. For instance, an aggregator may become quite sophisticated component of a flow because of its configuration: correlation & release strategies, output processor, expiration policies, etc. Therefore it can be considered individual SUT.

The problem with this now is in the following. Normally such an aggregator is configured somewhere within a flow (assuming it is described in DSL) with the help of appropriate specification (`AggregatorSpec`). The simplest approach to testing might be to extract the configuration into separate class and test it independently by providing a test instance of the `AggregatorSpec`. But this class has package-local constructor only (for feasible reasons) and thus can't be instantiated from outside of its package. Moreover, this approach requires quite a lot of code that heavily depends on Spring Integration internal structure. 

Much better solution, probably, would be to have a way to "extract" certain component(s) from the flow for testing. The identification of the components might be of 2 kinds: by the components' names and by surrounding channels' names (though it can be cumbersome in case of default ones). To achieve this, the testing framework should be able to build (possibly on mocks) the integration flows described in the application context but not run them. Then, by request, find and return specified components to a test.

To be more specific, here's the use case I encountered recently. Pay attention to commented lines only.

```java
    return IntegrationFlows
        .from(tailAdapter((new File(logPath))))
        .enrichHeaders(e -> e.headerFunction(LOG_TIMESTAMP_VALUE__HEADER, timestampExtractor::extractTimestamp))
        .enrichHeaders(e -> e.headerFunction(CORRELATION_ID, correlationProvider::obtainCorrelationId))
        .channel(preAggregatorQueueChannel)                  // (1)
        .aggregate(recordAggregatorConfigurer::configure)    // (2)
        .enrichHeaders(e -> e.headerFunction(RECORD_LEVEL__HEADER, this::detectRecordLevel))
        .channel(channels -> channels.publishSubscribe(logPath))
        .get();
```
`(1)` and `(2)` are flow components that work in pair and need to be tested together. `recordAggregatorConfigurer::configure` is a consumer that accepts `AggregatorSpec` and configures the aggregator with it. `preAggregatorQueueChannel` is a queue channel instance created early to be reused in other places. All that is needed here is to test these components separately from the entire flow.

Sorry if my suggestions would seem naive to you as I propose them from my only point of view. Hope you'll develop them into something decent for the community.

artembilan:
@Toparvion ,

Thank you very much for the feedback and sharing your thoughts on the matter.

Well, let me to answer to your concerns ASAP, first of all!
Then I'll study your comment thoroughly for further ideas.

So, see my previous comment about `MessagingTemplate`. You can get an appropriate channel from the `ctx`. In your case it is that `preAggregatorQueueChannel` as an input to the next aggregator. And use `MessagingTemplate.send()` to really send a test message especially for that aggregator.

You don't need to specify strict message channel in the flow. There will be just enough to follow with naming convention for [implicit channels in between](https://github.com/spring-projects/spring-integration-java-dsl/wiki/Spring-Integration-Java-DSL-Reference#message-channels) - `[IntegrationFlow.beanName].channel#[channelNameIndex]`.

This part is already there out-of-the-box. I think we should just document it properly 😄 

Of course, I understand that you would like to *verify* the result after an aggregator.
For this purpose you again can extract the message channel, but this time from the aggregator. And add a `ChannelInterceptor` with the `preSend()` implementation to catch a message and verify it after `MessagingTemplate.send()`.

This way you will have a code to test a component in the flow individually.

artembilan:
The `VerifyingChannelInterceptor` is one of the ideas to add to these testing tools.
It's purpose is really to be injected to the channel from the real application context, to be configured for some expectations and maybe verification in the end of the test, just because our flow might be async and even if we fail fast in the interceptor, we might not get an exception in the main test code by reason of thread shifting in the flow.

Toparvion:
Thank you Artem for the quick and detailed answers! The first one has shed the light on the way I can test my components.
I'd like to add a couple of comments on them. 

- The naming convention for implicit channels is described well in documentation. The only thing to improve here, possibly, is to stress the point that these names may be used to address individual channels from tests (as it doesn't seem obvious for the first sight).
However this approach seems to be a little bit crocky in the sense that if some time later someone inserts a new component into the flow somewhere _before_ the component being tested, he/she'd unintentionally make all the successive channels' indexes shift. While this would certainly lead to corresponding test's fail, the exact kind of the fail (its exception or stacktrace) is unpredictable ahead of time (as we don't know which channel happened to be shifted to the test). This would make us investigate every such situation individually. To avoid this, the tests should be targeted to explicit channel names, not the implicit ones. Ultimately this leads to a thought which might deserve special focus in the documentation:

> If your SUT components are located in the integration flow which is likely to change in future, it's worth to surround these components with explicitly named channels that will be addressed from tests.

- Another thought possibly deserving attention is dynamic nature of some integration flows (concerns Spring Integration 5.0+ only). The flow I've shown in the previous comment is not registered as usual bean at the start of the application. Instead, it is created and registered on demand in runtime by means of `IntegrationFlowContext`. So the testing tool should be able either to work with flows created at test's runtime or to provide its own `IntegrationFlowContext` able to register new flows in current test application context by demand. The latter seems more preferable as it allows to mock the flow's dependencies out in-place and gives a little more flexibility over test objects' life cycle. May be the current `IntegrationFlowContext` is already capable of that?

artembilan:
@Toparvion ,

Thank you very much for your deep thoughts on the matter!

You know, we really need such a feedback from the fields for everything we do here, just because the tool is for people like you, who use Spring to build the target application and face real challenges. So, the tool from us should really be as useful as possible, even if the solution has seemed nice from the first glance.

Well, I think, if you change something withing the flow, it really should lead to re-testing it and definitely test modification because the main flow behavior has been changed.

Any way I agree that implicit channel naming isn't so useful for testing, but at the same time there are many other feature which can't be tested in a forehead because they represent black-box, e.g. `<chain>` or nested bean. Or even no reason to go so far - `private` methods!

OTOH I have an other idea for this "test-in-isolation" concern. Instead of dealing with channels, I suggests something like `mockIntegrationContext.messagingTemplateFor(String endpointId)`, which will wrap the target `MessageHandler` and its endpoint to the `TestMessagingTemplate` where we can `send()`/`sendAndReceive()`/`receive()` around that component without care what are those channels near by the component. In your case for aggregator you can send several messages for testing and in the end call `receive()` to get a single result for the bunch of messages. But yes, you still have to specify an `id` for the endpoint to be able to mock it. 😄 

Regarding your:
>  register new flows in current test application context by demand. 

Yes, it is possible even right now with the `IntegrationFlowContext`. See the [commi](https://github.com/spring-projects/spring-integration/pull/2053/commits/3b4aec8411e39c4890c33d2f303e2a8d0f4a16a6) to this PR about `testMockMessageSourceDynamicFlow()`. So, when we will have enough mocking components (e.g. `mockMessageHandler()`) you really will have everything in hands to even build "dynamic mock flows" 😄 

Your turn! 😉

Toparvion:
@artembilan,
I'm really glad to have a chance to apply my humble experience for making such an excellent framework as Spring Integration even better!

First of all let me reply to your statement that
> if you change something withing the flow, it really should lead to re-testing it and definitely test modification because the main flow behavior has been changed

I absolutely agree with this. You might have thought that I want to avoid re-testing in case of further modification of a flow. Not exactly. By saying "corresponding test's fail" I meant a _unit_ test for an _individual_ component of the flow (the aggregator in my case). I believe such test should not fail if some _other single_ component of the flow changes because their tests are (or at least must be) independent. Unfortunately, in case of implicit channels addressing these tests may have spurious coupling in the form of channels' indexes. AFAIU, by saying "should lead to re-testing it" you meant testing of the _entire_ flow, i.e. _integration_ test. If so, there is no doubt that this test must be performed over again after any change to any flow component. Hope we both have gained understanding in this concern :blush:

Though this mess barely matters thanks to "test-in-isolation" concern. I really like it! This feature seems to become a swiss-knife in testing integration flows. Regarding your:
> you still have to specify an id for the endpoint to be able to mock it.

This doesn't look difficult thanks to concise syntax of endpoints configuration in Java DSL:
```java
    .handle(outboundAdapter(...)
            conf -> conf.id("handlerId"))
```
Furthermore, in fact I already used to assign explicit names to all the main endpoints in my production integration flows. It became a kind of best practice due to monitoring facilities - having named them as `1_ftpSource`, `2_unzipper`, `3_splitter` and so on, I've got the appropriate flow representation in my JMX console for performance monitoring (as it sorts the beans alphabetically by default). Hence addressing the endpoints from the tests shouldn't be a challenge for me :wink:

Thanks a lot for the sample test of dynamic flow registration! I've checked out PR's source branch and familiarized with the new subproject in general. I find it really handy (though it failed to build separately from others due to `Unable to find: src/checkstyle/checkstyle-suppressions.xml`). By the moment, the only thing staying unclear to me is how to deal with this:
```java
	@Bean(name = PollerMetadata.DEFAULT_POLLER)
	public PollerSpec defaultPoller() {
		return Pollers.fixedDelay(10);
	}
```
Is there a way to mock or re-configure the default poller for test if such poller exists in my production `@Configuration` class that is imported into the test context?

Toparvion:
BTW, for the sake of better understanding, I've built and installed current version of `MockIntegration` into my local Maven repository and tried to use it for testing my private (hobby) project. I've started with the (dynamic) flow I mentioned above. Instead of (unit-)testing aggregator individually, I've decided to write a comprehensive integration test for the whole flow. Because it starts with file tailing message source, I'd like to replace the source with mocked one:

```java
    IntegrationFlow tailingFlow = IntegrationFlows
        .from(tailAdapter((new File(logPath)))
            .id("tailSource"))
        .enrichHeaders(e -> e.headerFunction(LOG_TIMESTAMP_VALUE__HEADER, timestampExtractor::extractTimestamp))
        .enrichHeaders(e -> e.headerFunction(CORRELATION_ID, correlationProvider::obtainCorrelationId))
        .channel(preAggregatorQueueChannel)
        .aggregate(recordAggregatorConfigurer::configure)
        .enrichHeaders(e -> e.headerFunction(RECORD_LEVEL__HEADER, this::detectRecordLevel))
        .channel(channels -> channels.publishSubscribe(logPath))
        .get();
    IntegrationFlowRegistration tailingFlowRegistration = integrationFlowContext
        .registration(tailingFlow)
        .autoStartup(false)
        .register();

    mockIntegrationContext.instead("tailSource", mockMessageSource(LOG_LINES), true);
    // LOG_LINES is just an array of Strings
```

This looks much like the use of `instead` method in `MockMessageSourceTests` but ends up with:
```
org.springframework.beans.NotReadablePropertyException: Invalid property 'source' of bean class [org.springframework.integration.file.tail.OSDelegatingFileTailingMessageProducer]: Bean property 'source' is not readable or has an invalid getter method: Does the return type of the getter match the parameter type of the setter?

	at org.springframework.beans.AbstractNestablePropertyAccessor.getPropertyValue(AbstractNestablePropertyAccessor.java:615)
	at org.springframework.beans.AbstractNestablePropertyAccessor.getPropertyValue(AbstractNestablePropertyAccessor.java:606)
	at org.springframework.integration.mock.MockIntegration$Context.instead(MockIntegration.java:127)
	at org.springframework.integration.mock.MockIntegration$Context.instead(MockIntegration.java:121)
	at com.github.toparvion.testing.TailingFlowProviderTest.provideTailingFlow(TailingFlowProviderTest.java:93)

```

, though I clearly see the `Object source` field (and its getter) in `EventObject` class which is an ancestor for `OSDelegatingFileTailingMessageProducer`. Have you any idea of what is the reason? Do I use it incorrectly?

artembilan:
Hello, @Toparvion !

Sorry for the delay. 
A quick answer to your issue.

You can't use `mockMessageSource` instead of `OSDelegatingFileTailingMessageProducer` because the last one is a `MessageProducerSupport`, not `MessageSource`. 
See here for more information about different Spring Integration components: http://docs.spring.io/spring-integration/reference/html/overview.html#programming-tips

We definitely should provide the mocking part there as well. Although one more idea like 
````java
mockIntegrationContext.forChannel(String messageChannelId, 
                    MessageSource<?> mockMessageSource, 
                    Consumer<SourcePollingChannelAdapterSpec> endpointConfigurer)
````

To let to attache a new testing source endpoint to the existing channel for producing test messages.

**Comment by artembilan** [comment]
Regarding default poller...
Well, to be honest it isn't possible to replace it when context is started already.
OTOH it is even dangerous because `default` also means **shared**, therefore replacing it you may end up with unexpected behavior in other places of your flow.

That's why I suggest that `mockIntegrationContext.forChannel()` variant when you can introduce your mocking `MessageSource` with the desired polling options. That won't be **instead** of default one, or an existing endpoint, but will provide you similar way to emit messages to the same channel. Any way the subscriber on that channel doesn't care who is producer of messages for the channel.

This way you can mock and then verify in the some some part of your flow sending mock message to the channel even in the middle of  the flow.

Just an idea from the mind though...
Not sure that final solution will be like that.

BTW even right now I think that `mockIntegrationContext` is wrong name. Just because we might deal not only wit mocks. The `IntegrationTestContext` would be much better - we can create dynamic flows and inject them into real flow during testing. That doesn't insist to provide only mocks on the matter.
Although we still **mock** from the real context. 😄

**Comment by Toparvion** [comment]
Hello, Artem!
I didn't expect any answer during the weekend, it's ok :blush:

Yes, the exception aforementioned above is my stupid mistake (it was bad idea to code after midnight :smile:). Nevertheless it seems not very obvious for the end user to associate parameter name with component class which should be used in the replacing. I realize that it's just a draft and final solution may differ significantly. But still it looks like a point to improve, possibly, by checking the provided component's class against expected interfaces (to warn the user proactively).

The `forChannel()` method you've suggested above looks like a complement to `VerifyingChannelInterceptor` suggested several messages ago. AFAIU, together they would allow to logically "extract" arbitrary subflows from the main flow and test them in isolation. Although, to make it a comprehensive test equipment they should be supplemented with a tool for suppressing side effects arising from inside a flow, e.g. undesired behavior of some its intermediate components. As of now, the only universal solution for this seems to me a properly "charged" `instead()` method. But it would require a developer to use it with caution as it becomes extremely easy to break the flow altogether... Have you any ideas on how to deal with it?
This also makes me refer to the default poller again. I absolutely agree that its shared role may cause problems in case of reckless mocking. But still there may be situations when the default poller becomes an initiator of undesired behavior for the tests. Say, I have a flow that originates from a pollable message source which relies on the default poller and has `autoStartup(true)` flag set. Considering:
> ... you can introduce your mocking MessageSource with the desired polling options. That won't be **instead** of default one...

, if I understand you right, both the default and the specific poller would be active, wouldn't they? If they would, it may be worth to give the user (on their responsibility of course) a facility to stop the default poller, e.g. by means of an annotation or a kind of shortcut to `Lifecycle#stop()`. This approach does not seem dangerous as it's unlikely to have a test that covers a lot of flows all relying on the default poller. 

Sure, `IntegrationTestContext` suites better regarding to its meaning. However, `mockIntegrationContext` doesn't sound like something restricting to mocks only. Contrarily, the `mock` prefix makes it advisably differ from many other Spring terms that are hard to remember due to their similarity. So, if you eventually decide to leave current name as is there would at least one person who appreciate that.

**Comment by artembilan** [comment]
@Toparvion ,

One more time thank you for the review, testing and commenting!
All good feedback!

We really appreciate your time with us 😄 

I think I will pull the story to the team and we will make some decision how to be with this PR.
We might merge it as is and any other features can be addressed in the separate tasks.
/CC @garyrussell 

Now about your latest reply.

> But still it looks like a point to improve, possibly, by checking the provided component's class against expected interfaces (to warn the user proactively).

Sold: https://github.com/spring-projects/spring-integration/pull/2053/commits/9e9e8fbb7b7d44a93a72e9f71d0e0a8982b15200
Pay attention to the assert for exception message. That is really very helpful to determine where end-user is wrong.

> The `forChannel()` method you've suggested above looks like a complement to `VerifyingChannelInterceptor` suggested several messages ago. 

Indeed, the `VerifyingChannelInterceptor` can be used as hook to verify in the end of test what we have sent with that `forChannel()`. That is like one of the variants for verification. Another one, flowing in the air for a while already, is a `mockMessageHandler()` with `expect()` and `produce()` functions to mock some existing `MessageHandler` in the real context. Another one might be the same `TestMessagingTemplate.receive()`. Now I think that it is not necessarily to always send messages before starting to receive results from the component...

>  But it would require a developer to use it with caution as it becomes extremely easy to break the flow altogether.

Well, that is what we have even when we use Mockito Framework directly. I think no need extra comments: you always should be careful because even I can write some test which proves that the proper target solution is wrong, although meanwhile exactly my test is wrong because I don't provide proper mock or do that wrong way. 😄 

>  both the default and the specific poller would be active, wouldn't they?

Yes, gotcha! And yes I agree that it would be better to have some hook to stop undesired components, especially when we would like to "insert" a mock one for the same channel. So, how about an attribute for that annotation:
```
@RunWith(SpringRunner.class)
@MockIntegrationTest(stopEndpoints = {"foo", "bar", "baz"})
```
Because I guess you might not be interested to have all endpoints stopped otherwise you will be forced to start them manually when you would like to test a sub-flow. I think mostly what we should to stop are inbound endpoints, e.g. JMS Inbound Gateway, JDBC Source Polling Channel Adapter, WebSocket Inbound Channel Adapter and similar. The intermediate endpoints might be still ran, because they are not active and react only for messages in their input channels. 

> However, mockIntegrationContext doesn't sound like something restricting to mocks only. 

We'll see with the team what we can do on the matter 😄

**Comment by garyrussell** [comment]
Yes; I agree; we can merge this (when it has docs and javadocs) and add more features as separate stories.

**Comment by Toparvion** [comment]
Hello, @artembilan!
Thanks for the detailed answer and the commit you added!
I took a look on the latter and found the new behavior much more user-friendly than before :blush:

Talking of your suggestions for additional verification approaches, they all sound good and I think they should be actively promoted by means of documentation, otherwise the good-old `QueueChannel#receive()` verification style would stay among favorites for a long time. The detailed discussion of them would be much more effective if we have some sample implementations. 

> that is what we have even when we use Mockito Framework directly.

No doubt. I pointed to that caution just because in case of mocking arbitrary integration flow components the solution seemed to me especially fragile. Though, you're right, it barely differs from other mocking use cases.

> how about an attribute for that annotation: `@MockIntegrationTest(stopEndpoints = {"foo", "bar", "baz"})`

That's great! While this would be irreplaceable for simple tests, I think it is still worth for more complicated tests to add analogous capability in the form of a method which would call `Lifecycle#stop()` on the given component. This might be useful for two aims:

1.  To address test cases where the specified endpoint should be stopped **not** at the start of the test but later (though I'm not sure how frequent such cases are);
2. To improve test readability as it would clearly state that the stopping is called right here, at the place where it is expected to happen. Of course, this matters only for relatively big tests (which shouldn't exist ideally).

And yes, I agree that stopping should be applied mostly to inbound endpoints. It will be rarely needed (if ever) to stop intermediate components. Possibly, this fact might be used to make the testing API more concrete.

Well, it seems this story contains ideas enough for a separate framework :smile: I appreciate the decision to merge the PR as it is because it will expose an MVP which can be tried out by the community to gain the first real feedback. 

As I said before, I'm very glad to have a chance to take part in Spring Integration improvement. But being a developer I'd prefer to do it by code, not by words only. During this conversation I realized that the subject of integration testing is interesting, allied and useful for me. Therefore, if you don't mind I'd like to join your project as a contributor (according to [Contributing Guidelines](https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc) of course). 
As you might have noticed I'm not an expert in Spring Integration (yet). I've been using Spring Framework on and off for several years (since its 3+ version), but started to use Spring Integration just several months ago. I see there will be quite a lot of different sized tasks in Spring Integration Mock testing tool. So may be some of the least critical of them might be entrusted to me to start. What do you think about it?

**Comment by artembilan** [comment]
Pushed JavaDocs for existing classes and `stopEndpoints` feature as we discussion above.

Meanwhile you are reviewing, I'm working under some doc for the Reference Manual as suggested by @garyrussell .

@Toparvion ,
all your comments are good and looks like that we are finally on the same page 😄 

Well, of course, contribution are welcome!
And actually you're doing well so far!
Such a discussion is really valuable for us.
Thank you for your mind and time!

Since we don't have this testing code in the master yet, it isn't possible to improve it from your side clearly. Meanwhile you can consider to do some other contribution task to get used to our JIRA/PR/review/feedback/merge process. Feel free to pick up any task from the back log which feels easy for you: https://jira.spring.io/browse/INT/fixforversion/15433!

**Comment by Toparvion** [comment]
@artembilan, 
I'm glad to see you appreciate my intention!
Yes, I think it would be better for me to get familiar with your processes by doing some simple task like subtle improving of the documentation. 
Unfortunately I won't be able to review the latest changes today, but I'll try to do it ASAP.

File: spring-integration-mock/src/main/java/org/springframework/integration/mock/IntegrationEndpointsInitializer.java [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.beans.factory.BeanFactoryAware;
+import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.integration.endpoint.AbstractEndpoint;
+import org.springframework.util.Assert;
+import org.springframework.util.PatternMatchUtils;
+
+/**
+ * A component to customize {@link AbstractEndpoint} beans according
+ * provided options in the {@link SpringIntegrationTest} annotation after all beans
```
garyrussell: according to the provided

spring-integration-mock/src/main/java/org/springframework/integration/mock/MockIntegrationContext.java
```diff
@@ -0,0 +1,110 @@
...
+import org.springframework.beans.factory.BeanFactoryAware;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.context.Lifecycle;
+import org.springframework.integration.core.MessageSource;
+import org.springframework.integration.endpoint.SourcePollingChannelAdapter;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link BeanFactoryAware} component with an API to customize real beans
+ * in the application context from tests code.
```
garyrussell: test code

spring-integration-mock/src/main/java/org/springframework/integration/mock/MockIntegrationContext.java
```diff
@@ -0,0 +1,110 @@
...
+ * <p>
+ * The bean for this class is registered automatically via {@link SpringIntegrationTest}
+ * annotation and can be autowired into test class.
```
garyrussell: via the

spring-integration-mock/src/main/java/org/springframework/integration/mock/MockIntegrationContextCustomizerFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+
+import org.springframework.core.annotation.AnnotatedElementUtils;
+import org.springframework.test.context.ContextConfigurationAttributes;
+import org.springframework.test.context.ContextCustomizer;
+import org.springframework.test.context.ContextCustomizerFactory;
+
+/**
+ * The {@link ContextCustomizerFactory} implementation to produce
+ * {@link MockIntegrationContextCustomizer} if {@link SpringIntegrationTest} annotation
```
garyrussell: to produce a

if a

spring-integration-mock/src/main/java/org/springframework/integration/mock/SpringIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
...
+public @interface SpringIntegrationTest {
+
+	/**
+	 * Specify a simple matching patterns for
+	 * {@link org.springframework.integration.endpoint.AbstractEndpoint}
+	 * bean names to mark them as {@code autoStartup = false}
+	 * during context initialization.
+	 * @return the endpoints name patterns to stop during context initialization
+	 */
+	String[] stopEndpoints() default {};
```
garyrussell: `dontStartEndpoints()` ?

`overrideAutoStartFalseEndpoints()` ?

spring-integration-mock/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,3 @@
+# Spring Test ContextCustomizerFactories
+org.springframework.test.context.ContextCustomizerFactory=\
+org.springframework.integration.mock.MockIntegrationContextCustomizerFactory
```
garyrussell: newline at EOF

garyrussell (changes requested):
That's all for this round ... looking good!!

build.gradle
```diff
@@ -500,6 +500,14 @@ project('spring-integration-mail') {
 	[compileJava,compileTestJava]*.options*.compilerArgs = ["${xLintArg},-path"]
 }
 
+project('spring-integration-mock') {
```
garyrussell: I wonder if we should rename the current `spring-integration-test` to `spring-integration-test-support` and call this one `spring-integration-test` - after all, it's not only about mocks.

artembilan: Ha-ha! That's true!
The end-user will still use `spring-integration-test` dependency and existing utilities will be pulled transitively from that new artifact 😄

garyrussell: True, but we should mention it in the migration guide.

spring-integration-mock/src/main/java/org/springframework/integration/mock/SpringIntegrationTest.java
artembilan: Well, "don't start" is fully synonym to the "stop".
The other variant is so long.

The good JavaDocs (existing?) should lead the light what is the purpose of the attribute.

garyrussell: Not really - `stop()` implies they are already started.

How about `noStart()` or `dontStart()` or `noAutoStart()` or `autoStartOverride()`.

We don't really need `Endpoints`

artembilan: OK! Sold. How about `noAutoStartup()`  as a sibling to the `SmartLifecycle.isAutoStartup()`? 😄

garyrussell: 👍 

Maybe Regex would be more flexible than a simple pattern - but I don't think we need both.

build.gradle
garyrussell: We'll need to think about the package name too.

`org.springframework.integration.testing`, `org.springframework.integration.testframework` ?

artembilan: Well, I decided to leave `SI-test-support` package as is.
And in a new module I just did a couple sub-packages.
So, now we have there `o.s.i.test.context` and `o.s.i.test.mock`. With the the new module is just like a sub-module of the existing `support` structure.

Does it make sense?

garyrussell: Yes, that's fine, we just cannot have the same package in 2 jars but, yes, I agree it's less invasive to leave the "old" classes in their original packages.

src/reference/asciidoc/testing.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+Spring Integration provides a number of utilities and annotations to help when testing your application.
+Test support is provided by two modules: `spring-integration-test-support` contains core items and shared utilities; and `spring-integration-test` which provides mocking and context configuration components for integration tests.
+
+Meanwhile `spring-integration-test-support` provides basic, independent utilities, rules and matchers for unit testing (also doesn't depend on Spring Integration code base and is used internally in the Framework), the `spring-integration-test` is aimed to break the integration testing challenge wall and provides comprehensive high level API to mock integration components and verify behaviour individual components, as well as whole integration flows or their parts.
+(__A thorough treatment of testing in the enterprise is beyond the scope of this reference manual.__)
+
+[[testing-intro]]
+=== Introduction
+
```
garyrussell: Is this all the docs you are planning for this PR? Or should I wait for more?

artembilan: No, please, wait! 😄

src/reference/asciidoc/testing.adoc
```diff
@@ -2,12 +2,278 @@
...
+    private MockIntegrationContext mockIntegrationContext;
+
+}
+----
+
+The `@SpringIntegrationTest` populates `MockIntegrationContext` bean which can be autowired to the test class for further usage.
+With the provided `noAutoStartup` option, the Spring Integration Test Framework switches off endpoints, which are matched to the provided patterns, from the auto-startup during application context refreshing.
+This is useful, when we would like do not have real connections to the target systems from Inbound Channel Adapters, for example AMQP Inbound Gateway, JDBC Polling Channel Adapter, WebSocket Message Producer in client mode etc.
+
+The `MockIntegrationContext` is aimed to be used in the target test-cases for modifications beans in the real application context, for example those stopped endpoints can be replaced with mocks:
```
Toparvion: > those stopped endpoints

What if those endpoints were not stopped by `noAutoStartup`? (because for example the user forgot to include them into the annotation parameter or mistaken while doing it) AFAIU, the endpoint will be relocated into `MockIntegrationContext#beans` but will continue to act as before.
Maybe we should explicitly stop the endpoint being replaced? Or at least warn the users here in documentation that replacing an endpoint doesn't stop it?

src/reference/asciidoc/testing.adoc
```diff
@@ -2,12 +2,278 @@
...
+}
+----
+
+See their JavaDocs for more information.
+
+[[testing-mocks]]
+=== Integration Mocks
+
+The `org.springframework.integration.test.mock` package offers tools and utilities for mocking, stubbing and verifications of Spring Integration components.
+The mocking functionality is fully based and compatible with well know Mockito Framework.
```
Toparvion: well know**n**

Toparvion: Probably it's worth to explicitly denote the version of Mockito (2) because there are a lot of Mockito 1 users (for example those who have to use its PowerMockito extension which is not yet fully compatible with version 2).

src/reference/asciidoc/testing.adoc
```diff
@@ -2,12 +2,278 @@
...
+
+==== MockIntegration
+
+The `MockIntegration` factory provides an API to build mocks for Spring Integration beans which are parts of the integration flow - `MessageSource`, `MessageProducer`, `MessageHandler`, `MessageChannel`.
```
Toparvion: AFAIU, the rest of corresponding methods (beside one for `MessageSource`) are still in development, aren't they?

Toparvion (commented):
I always liked Spring Framework for its excellent documentation and JavaDocs and now I'm seeing how this magic happens :blush:

spring-integration-mock/src/main/java/org/springframework/integration/mock/MockIntegration.java
```diff
@@ -19,48 +19,41 @@
...
-import org.springframework.util.Assert;
 
 /**
+ * The factory for integration specific mock components.
+ *
  * @author Artem Bilan
  *
  * @since 5.0
  */
 public final class MockIntegration {
```
artembilan: Well, that is a parallel with an existing `MockMvc`.
The name has been chosen by 	intend.
Actually my idea have more then `factory` purpose on this class, but I don't see them yet... 😄

Toparvion: I just wondered if this might be SRP violation. But as judged by `MockMvc ` it's OK.

src/reference/asciidoc/testing.adoc
artembilan: Yes, that's true. An endpoint will act as it is with regular application. But no one endpoint moves to that beans map, unless an explicit `instead()` call. If you don't want to have endpoint started after replacement, use an overloaded `instead` version. Otherwise the replaced part in that endpoint will start act immediately after replacement. But yeah, I think I'll move replacement to the very end of `instead` to avoid some race conditions.

artembilan: I've changed my mind. I'm stopping endpoint before replacing.
The subsequent start is depended on the provided `autoStartup` option of the `instead()`

src/reference/asciidoc/testing.adoc
```diff
@@ -1,11 +1,11 @@
 [[testing]]
-=== Tests support
+== Tests support
 
 Spring Integration provides a number of utilities and annotations to help when testing your application.
 Test support is provided by two modules: `spring-integration-test-support` contains core items and shared utilities; and `spring-integration-test` which provides mocking and context configuration components for integration tests.
 
 Meanwhile `spring-integration-test-support` provides basic, independent utilities, rules and matchers for unit testing (also doesn't depend on Spring Integration code base and is used internally in the Framework), the `spring-integration-test` is aimed to break the integration testing challenge wall and provides comprehensive high level API to mock integration components and verify behaviour individual components, as well as whole integration flows or their parts.
-(A thorough treatment of testing in the enterprise is beyond the scope of this reference manual.)
+(__A thorough treatment of testing in the enterprise is beyond the scope of this reference manual.__)
```
artembilan: Doh. I wasn't aware of that document.

Will investigate it for some ideas in our testing framework! 😄 

Feel free to share your ideas already. But please, in the separate thread in JIRA under ` Testing` component.

artembilan:
@Toparvion ,

I hope with the latest commit I have addressed all your concerns.
And if you noticed, when I'm not agreed with you, I add a comment. 
Otherwise just silent fix.

We believe with @garyrussell that this communication and PR progress way helps to save some time for all of us.

BTW, thank you for such a great review. And it is really such a shame that I'm not familiar with that Gregor's doc 😢 .

Will look eventually what can borrow to our "canonical EIP" implementation. I mean Spring Integration at all. 😄

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/TestUtils.java
```diff
@@ -56,6 +56,15 @@
  */
 public abstract class TestUtils {
 
+	/**
+	 * Obtain a value for the propty from the provide object.
```
Toparvion: ...for the prop**er**ty...

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/TestUtils.java
```diff
@@ -85,6 +105,11 @@ else if (i == tokens.length - 1) {
 		return (T) value;
 	}
 
+	/**
+	 * Create a {@link TestApplicationContext} instance
+	 * supplied with the besic Spring Integration infrastructure.
```
Toparvion: > supplied with the besic

artembilan: Welcome back, @Toparvion !

M-m-m. I think you mean the typo: s/besic/basic.

Otherwise you have just copied/pasted text 

Maybe that was intention to pull my attention, but since I did a mistake it isn't obvious for me where I'm wrong 😄

Toparvion:
@artembilan,
Thank you for considering my comments! I believe they would make the product a little bit more convenient for the end-users. And yes, all the of them have been addressed. I've just added a couple of trivial points to the latest commit. 

To be honest, I also wasn't familiar with that paper before start to search for an example of the source on the matter. Well, it turned out to be a nice chance to eventually get familiar with it :smiley:

It's a quite busy time for me these days but upon availability I'll continue to try the tool out on my personal project. Probably, this will be happening after this PR is merged. Until then, I would be glad to help you as usual - by means of discussion or participating in review.

src/reference/asciidoc/testing.adoc
Toparvion: Ok, thanks! If I find something potentially interesting, I'll share it with you through JIRA.

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/TestUtils.java
Toparvion: Yes, that was an unlucky attempt to point to the typo (should be `basic`). I had been hesitating over a choice of expressing form for a while and at last chose the worst possible variant :smiley:

artembilan:
@garyrussell ,

any issues do not merge this yet?
I'm a bit shattered to always refresh project in IDEA to switch between different code based for this branch and `master` 😄 

Thank you!

artembilan:
Also see: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.3-to-5.0-Migration-Guide#testing-support

garyrussell:
I haven't had a chance to review the docs (fighting with Kafka) - will do after lunch.

garyrussell:
Doc polishing here: https://github.com/garyrussell/spring-integration/commit/6d6c00b334ef70bc2ff1ce1af2cbb895b50289af for your review.

I also opened INT-4521 since this is an important new feature that should show up in the JIRA release notes.

artembilan:
Great!
Thank you, @garyrussell !

Reviewed your polishing. I will never write clean English text 😄 

Also linked PR to the JIRA.

garyrussell:
>I will never write clean English text

Your English is 1,000,000 times better than any foreign language ability that I have.

Merging...

garyrussell (approved):

garyrussell:
Merged as 2862c449c512665326d5821fb8f0b779cf57b723

PR #2055, opened 2017-02-14
[doc] Fix incorrect package name of PropertiesPersistingMetadataStore

dalbani:


artembilan:
Thank you very much for the fix, @dalbani 

Looking forward for more!

PR #2057, opened 2017-02-14
INT-3561: WS: add `web-service-template` attr

artembilan:
JIRA: https://jira.spring.io/browse/INT-3561

To allow any possible custom properties on the `WebServiceTemplate` it would be better just to have an injection for the rather than try to pursue exposing for all `WebServiceTemplate` setters


garyrussell:
Merged as 1a892ab1c9814bf70bc5a9b9881e023d10e0099e with minor doc polish and fix for inconsistent indentation

PR #2058, opened 2017-02-15
INT-4089 Add Source support for WS header mapping

artembilan:
JIRA: https://jira.spring.io/browse/INT-4089

Right now `DefaultSoapHeaderMapper` can map only `String` user-defined headers and only to the `SAOP:Header` attributes.

* Add functionality to map `Source` user-defined headers as sub-elements of the `SAOP:Header`

### DIFF
--- a/spring-integration-ws/src/test/java/org/springframework/integration/ws/DefaultSoapHeaderMapperTests.java
+++ b/spring-integration-ws/src/test/java/org/springframework/integration/ws/DefaultSoapHeaderMapperTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-integration-ws/src/test/java/org/springframework/integration/ws/DefaultSoapHeaderMapperTests.java
```diff
@@ -93,20 +100,20 @@ public void testCustomSoapHeader() {
...
-			+ "</soapenv:Envelope>";
+						+ "</auth>"
+						+ "<bar>BAR</bar>"
+						+ "<baz>BAZ</baz>"
+						+ "<qux>qux</qux>"
+						+ "</soapenv:Header>"
+						+ "<soapenv:Body>"
+						+ "<foo>foo</foo>"
+						+ "</soapenv:Body>"
+						+ "</soapenv:Envelope>";
```
garyrussell: Yikes! You removed all my pretty formatting 👎 

I wish your IDEA didn't mess with stuff you didn't change 😄

garyrussell (commented):
Will fix on merge.

spring-integration-ws/src/test/java/org/springframework/integration/ws/DefaultSoapHeaderMapperTests.java
artembilan: Ha! that wasn't intentional, really 😄 

Will take a look what IDEA can do on the matter.

Do you tell the same to to SCSt guys when they reformat imports order or anything similar with code style?

garyrussell:
Merged as b0094b32b72920755722287483acb15687f9a61b with minor doc polishing

PR #2059, opened 2017-02-16
INT-2655: (un)marshal only via WebServiceTemplate

artembilan:
JIRA: https://jira.spring.io/browse/INT-2655

There might be some use-cases when end-user expect more control over sending/receiving messages via `WebServiceTemplate`.
It would be better to delegate (un)marshalling logic to the `WebServiceTemplate.marshalSendAndReceive()` directly.

### DIFF
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
@@ -66,7 +66,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 
 	private final Map<String, Expression> uriVariableExpressions = new HashMap<String, Expression>();
 
-	private volatile  StandardEvaluationContext evaluationContext;
+	private volatile StandardEvaluationContext evaluationContext;
 
 	private volatile WebServiceMessageCallback requestCallback;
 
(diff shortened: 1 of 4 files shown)


spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
```diff
@@ -66,7 +66,7 @@
 
 	private final Map<String, Expression> uriVariableExpressions = new HashMap<String, Expression>();
```
garyrussell: copyright 2017

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
```diff
@@ -82,23 +82,32 @@ public MarshallingWebServiceOutboundGateway(String uri, Marshaller marshaller) {
 		this(uri, marshaller, (WebServiceMessageFactory) null);
 	}
```
garyrussell: remove `this.` from `this.configureMarshallers(...)`

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
```diff
@@ -107,39 +116,39 @@ private void configureMarshallers(Marshaller marshaller, Unmarshaller unmarshall
...
+		getWebServiceTemplate().setUnmarshaller(unmarshaller);
+	}
 
-		MarshallingRequestMessageCallback(WebServiceMessageCallback requestCallback,
-				Message<?> requestMessage) {
-			super(requestCallback, requestMessage);
-		}
+	@Override
+	public String getComponentType() {
+		return "ws:outbound-gateway(marshaling)";
```
garyrussell: `s/l/ll/`

garyrussell (commented):
That's all; minor; will fix on merge.

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
artembilan: He-he! Existing code for a while already 😄 

But thanks for the care on merge!

garyrussell:
Merged as dda2fea60eb172e9108e263bc786318d3f5b7089

PR #2061, opened 2017-02-19
INT-4230: Fix (S)FTP Preserve Timestamp on MGET

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4230

The `get()` method only preserves the timestamp when an LS operation is done within it.
For MGET ops, we don't perform another LS, so the timestamp was not updated.

Add code to the recursive and non-recursive MGET methods to preserve the timestamp if so configured.

__cherry-pick to 4.3.x__


artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-integration/commit/9f7e10f927312981409788699bba90389d049014

PR #2064, opened 2017-02-22
INT-4226: Fix PollableAmqpChannel.receive(timeout)

artembilan:
JIRA: https://jira.spring.io/browse/INT-4226

Since `rabbitTemplate` properly implement `receive()` with `timeout` it looks organic to delegate a `PollableChannel.receive(timeout)` there.

* Refactor `PollableAmqpChannel` to properly delegate `PollableChannel.receive(timeout)` implementation
* Refactor `ChannelTests` to perform `PollableAmqpChannel.receive(timeout)` directly instead of `Thread.sleep()` loop
* Get rid of `BrokerRunning` in favor of the same rule from the `spring-rabbit-junit` dependency


garyrussell:
LGTM; but needs docs - https://github.com/garyrussell/spring-integration/commit/4c77fdd13b9b10afc08191c4bd951a427ad93d4b for your review.

I also added a migration guide entry: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.3-to-5.0-Migration-Guide#amqp-backed-pollable-channel

garyrussell:
Regarding your comment on my commit...

>LGTM of course, but I just don't see reason in such an apology about blocking since it is really an expected behavior for the `receive(timeout)`. In other words we have just aligned this AMQP-backed channel with the contract.

Well, I think this is a special case because it's more expensive than a simple `basicGet()` - we have to create and cancel a consumer for each message.

garyrussell:
Merged as 257a2d3dab7235e95acc5b84d959a4f21581acfa - I added another sentence.

PR #2065, opened 2017-02-22
INT-4233: Add (S)FTP FileInfo Header (Streaming)

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4233

Add the complete file info as JSON (when Jackson or Boon available) to the message headers
when streaming inbound.

Provide a mechanism to configure Boon to provide similar output to Jackson.

Also allow subclasses to provide their own object mapper.

Also clean up after the AMQP DSL tests, and don't use a queue `foo`. I often have such a queue
with content; this caused tests to fail.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,13 @@ package org.springframework.integration.support.json;
 
 
 import org.springframework.util.ClassUtils;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * Simple factory to provide {@linkplain JsonObjectMapper}
@@ -61,12 +68,103 @@ public final class JsonObjectMapperProvider {
 	}
 
 	/**
+	 * Return an object mapper builder if available.
+	 * @param preferBoon true to prefer boon if available.
+	 * @return the mapper builder.
+	 * @throws IllegalStateException if an implementation is not available.
+	 * @since 5.0
+	 */
+	public static JsonObjectMapperBuilder<?> newInstanceBuilder(boolean preferBoon) {
+		if (JacksonJsonUtils.isJackson2Present() && (!preferBoon || !boonPresent)) {
+			return new JacksonJsonObjectMapperBuilder();
+		}
+		else if (boonPresent) {
+			return new BoonJsonObjectMapperBuilder();
+		}
+		else {
+			throw new IllegalStateException("Neither jackson-databind.jar, nor boon.jar is present in the classpath.");
+		}
+	}
+
+	/**
 	 * Returns true if a supported JSON implementation is on the class path.
 	 * @return true if {@link #newInstance()} will return a mapper.
 	 * @since 4.2.7
 	 */
 	public static boolean jsonAvailable() {
 		return JacksonJsonUtils.isJackson2Present() || boonPresent;
+	}
+
+	public static abstract class JsonObjectMapperBuilder<B extends JsonObjectMapperBuilder<B>> {
+
+		protected boolean usePropertyOnly = true; // NOSONAR
+
+		protected boolean includeAllValues = true; // NOSONAR
+
+		public B usePropertyOnly(boolean use) {
+			this.usePropertyOnly = use;
+			return _this();
+		}
+
+		public B includeAllValues(boolean include) {
+			this.includeAllValues = include;
+			return _this();
+		}
+
+		@SuppressWarnings("unchecked")
+		protected final B _this() {
+			return (B) this;
+		}
+
+		public abstract JsonObjectMapper<?, ?> build();
+
+	}
+
+	private static class JacksonJsonObjectMapperBuilder
+			extends JsonObjectMapperBuilder<JacksonJsonObjectMapperBuilder> {
+
+		JacksonJsonObjectMapperBuilder() {
+			super();
+		}
+
+		@Override
+		public JsonObjectMapper<?, ?> build() {
+			ObjectMapper objectMapper = new ObjectMapper();
+			objectMapper.configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false);
+			objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+			if (this.usePropertyOnly) {
+				objectMapper.setVisibility(PropertyAccessor.GETTER, Visibility.PUBLIC_ONLY);
+				objectMapper.setVisibility(PropertyAccessor.FIELD, Visibility.NONE);
+			}
+			if (this.includeAllValues) {
+				objectMapper.setSerializationInclusion(JsonInclude.Include.ALWAYS);
+			}
+			return new Jackson2JsonObjectMapper(objectMapper);
+		}
+
+	}
+
+	private static class BoonJsonObjectMapperBuilder extends JsonObjectMapperBuilder<BoonJsonObjectMapperBuilder> {
+
+		BoonJsonObjectMapperBuilder() {
+			super();
+		}
+
+		@Override
+		public JsonObjectMapper<?, ?> build() {
+			return new BoonJsonObjectMapper(null, f -> {
+				f.useAnnotations();
+				if (this.usePropertyOnly) {
+					f.usePropertyOnly();
+				}
+				if (this.includeAllValues) {
+					f.includeDefaultValues()
+						.includeNulls()
+						.includeEmpty();
+				}
+			});
+		}
+
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,13 +18,6 @@ package org.springframework.integration.support.json;
 
 
 import org.springframework.util.ClassUtils;
-
-import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.PropertyAccessor;
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.MapperFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * Simple factory to provide {@linkplain JsonObjectMapper}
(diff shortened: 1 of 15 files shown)


spring-integration-amqp/src/test/java/org/springframework/integration/amqp/rule/BrokerRunning.java
```diff
@@ -132,12 +132,15 @@ public Statement apply(Statement base, Description description) {
 		return super.apply(base, description);
 	}
 
-	public void removeTestQueues() {
+	public void removeTestQueues(String... additionalQueues) {
```
artembilan: How about to merge https://github.com/spring-projects/spring-integration/pull/2064 first?
We won't need this change because there won't be this local `BrokerRunning` any more.
The `spring-rabbit-junit` `BrokerRunning` has this method already.

Feel free to apply those `AmqpTests` changes during merge that PR

garyrussell: Agreed

garyrussell: Moved AMQP test fix to https://github.com/spring-projects/spring-integration/pull/2064

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -136,11 +157,20 @@ protected Object doReceive() {
...
+				AbstractIntegrationMessageBuilder<InputStream> builder = getMessageBuilderFactory()
+						.withPayload(session.readRaw(remotePath))
 						.setHeader(IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE, session)
 						.setHeader(FileHeaders.REMOTE_DIRECTORY, file.getRemoteDirectory())
-						.setHeader(FileHeaders.REMOTE_FILE, file.getFilename())
-						.build();
+						.setHeader(FileHeaders.REMOTE_FILE, file.getFilename());
+				if (getObjectMapper() != null) {
+					try {
+						builder.setHeader(FileHeaders.REMOTE_FILE_INFO, getObjectMapper().toJson(file));
```
artembilan: Why can't we transfer `FileHeaders.REMOTE_FILE_INFO` as a regular `FileInfo` instance if we don't want JSON and everything downstream is in memory?

artembilan (changes requested):
That's the first round.
I will think about the rest meanwhile...

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -74,6 +79,12 @@ protected AbstractRemoteFileStreamingMessageSource(RemoteFileTemplate<F> templat
 			Comparator<AbstractFileInfo<F>> comparator) {
 		this.remoteFileTemplate = template;
 		this.comparator = comparator;
+		if (JacksonJsonUtils.isJackson2Present()) {
+			this.objectMapper = JsonObjectMapperProvider.newInstanceBuilder(false).build();
```
artembilan: So, if there is no `jackson`, but is `boon` we fail with the:

> Neither jackson-databind.jar, nor boon.jar 

Because of:
```
else if (boonPresent) {
	return new BoonJsonObjectMapperBuilder();
}
else {
	throw new IllegalStateException("Neither jackson-databind.jar, nor boon.jar is present in the classpath.");
}
```

Why an approach from the `FileSplitter` doesn't work here?
```
JsonObjectMapperProvider.jsonAvailable() ? JsonObjectMapperProvider.newInstance() : null;
```
In other words fallback to `boon`  if present and no `jackson`.

spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
```diff
@@ -61,6 +68,25 @@ else if (boonPresent) {
...
 	/**
+	 * Return an object mapper builder if available.
+	 * @param preferBoon true to prefer boon if available.
+	 * @return the mapper builder.
+	 * @throws IllegalStateException if an implementation is not available.
+	 * @since 5.0
+	 */
+	public static JsonObjectMapperBuilder<?> newInstanceBuilder(boolean preferBoon) {
+		if (JacksonJsonUtils.isJackson2Present() && (!preferBoon || !boonPresent)) {
+			return new JacksonJsonObjectMapperBuilder();
```
artembilan: I think existing `newInstance()` ca use these builders with all those preferences.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
artembilan: I wonder why can't we build JSON manually from such a simple `FileInfo` object?..
To be honest I'm not sure that we should serialize to the JSON a target implementation returned by the `getFileInfo()`.

src/reference/asciidoc/ftp.adoc
```diff
@@ -534,7 +534,13 @@ If you don't actually want to persist the state, an in-memory `SimpleMetadataSto
 If you wish to use a filename pattern (or regex) as well, use a `CompositeFileListFilter`.
 The java configuration below shows one technique to remove the remote file after processing.
 
-Use the `max-fetch-size` attribute to limit the number of files fetched on each poll when a fetch is necessary; set to 1 and use a persistent filter when running in a clustered environment;
+Use the `max-fetch-size` attribute to limit the number of files fetched on each poll when a fetch is necessary; set to 1 and use a persistent filter when running in a clustered environment.
+
+The adapter puts the remote directory and file name in headers `FileHeaders.REMOTE_DIRECTORY` and `FileHeaders.REMOTE_FILE` respectively.
+Starting with _version 5.0_, complete remote file information, in JSON, is provided in the `FileHeaders.REMOTE_FILE_INFO` header, if a supported JSON object mapper on the classpath - currently Jackson 2 and Boon.
+
+Common properties are available as top level map entries, but the complete `FTPFile` object is included, under key `fileInfo`, providing all properties supported by the underlying library.
```
artembilan: OK, I see that you explain my concerns here, but still would like to share what is in mind respectively

artembilan (changes requested):

garyrussell:
OK; let's take that simpler approach - we can just add `toJson` on the abstract class.

I wasted a lot of time getting Boon to work something like Jackson's OOTB mapper and it was still not very reliable.

garyrussell:
Rejected.

artembilan:
> ...  getting Boon to work something like Jackson's OOTB mapper and it was still not very reliable.

Yeah... That might was wrong choice in the beginning. And that's maybe why SF doesn't provide its support.

Now I'm looking into `FileSplitter` and not sure why we use `objectMapper` there. The `FileMarker` is so simple type to represent it as a JSON manually...

garyrussell:
Have written a simple serializer; testing now.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -136,10 +136,12 @@ protected Object doReceive() {
 			String remotePath = remotePath(file);
 			Session<?> session = this.remoteFileTemplate.getSession();
 			try {
-				return getMessageBuilderFactory().withPayload(session.readRaw(remotePath))
+				return getMessageBuilderFactory()
+						.withPayload(session.readRaw(remotePath))
 						.setHeader(IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE, session)
 						.setHeader(FileHeaders.REMOTE_DIRECTORY, file.getRemoteDirectory())
 						.setHeader(FileHeaders.REMOTE_FILE, file.getFilename())
+						.setHeader(FileHeaders.REMOTE_FILE_INFO, (file.toJson()))
```
artembilan: So, you still aren't going to provide alternative with direct object, not JSON. Are you?

garyrussell: I thought about also posting the object but decided not to. Do you think we should?

garyrussell: I guess something like `FileHeaders.RAW_FILE_INFO`.

artembilan: M-m-m. No, I thought about something like `boolean fileInfoJson`.
The same meaning like we have in `FileSplitter.markersJson`.
I understand that it is a new feature here, unless an existing the already `markers`.
But seems for me for consistency it won't hurt to support raw object here as well.

spring-integration-core/src/test/java/org/springframework/integration/json/SimpleJsonSerializerTests.java
```diff
@@ -0,0 +1,81 @@
...
+ *
+ */
+public class SimpleJsonSerializerTests {
+
+	@Test
+	public void test() throws Exception {
+		Foo foo = new Foo();
+		String json = SimpleJsonSerializer.toJson(foo, "fileInfo");
+		@SuppressWarnings("unchecked")
+		Map<String, Object> fromJson = JsonObjectMapperProvider.newInstance().fromJson(json, Map.class);
```
artembilan: Does it work for the `Foo.class`?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -136,10 +150,13 @@ protected Object doReceive() {
...
 			Session<?> session = this.remoteFileTemplate.getSession();
 			try {
-				return getMessageBuilderFactory().withPayload(session.readRaw(remotePath))
+				return getMessageBuilderFactory()
+						.withPayload(session.readRaw(remotePath))
 						.setHeader(IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE, session)
 						.setHeader(FileHeaders.REMOTE_DIRECTORY, file.getRemoteDirectory())
 						.setHeader(FileHeaders.REMOTE_FILE, file.getFilename())
+						.setHeader(FileHeaders.REMOTE_FILE_INFO,
+										this.fileInfoJson ? (file.toJson()) : file.getFileInfo())
```
artembilan: Why `file.getFileInfo()`?
Isn't consistent with `file.toJson()`.
So, we are expecting the `FileInfo` object there, not low-level entity.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/StreamingInboundTests.java
```diff
@@ -81,6 +91,14 @@ public void testAllData() throws Exception {
 		assertEquals("baz\nqux", new String(received.getPayload()));
 		assertEquals("/foo", received.getHeaders().get(FileHeaders.REMOTE_DIRECTORY));
 		assertEquals("bar", received.getHeaders().get(FileHeaders.REMOTE_FILE));
+		fileInfo = (String) received.getHeaders().get(FileHeaders.REMOTE_FILE_INFO);
+		assertThat(fileInfo, containsString("remoteDirectory\":\"/foo"));
+		assertThat(fileInfo, containsString("permissions\":\"-rw-rw-rw"));
+		assertThat(fileInfo, containsString("size\":42"));
+		assertThat(fileInfo, containsString("directory\":false")); // Boon doesn't emit default values (false)
```
artembilan: I think we can not talk about `Boon` here any more

src/reference/asciidoc/sftp.adoc
```diff
@@ -573,7 +573,12 @@ If you don't actually want to persist the state, an in-memory `SimpleMetadataSto
 If you wish to use a filename pattern (or regex) as well, use a `CompositeFileListFilter`.
 The java configuration below shows one technique to remove the remote file after processing.
 
-Use the `max-fetch-size` attribute to limit the number of files fetched on each poll when a fetch is necessary; set to 1 and use a persistent filter when running in a clustered environment;
+Use the `max-fetch-size` attribute to limit the number of files fetched on each poll when a fetch is necessary; set to 1 and use a persistent filter when running in a clustered environment.
+
+The adapter puts the remote directory and file name in headers `FileHeaders.REMOTE_DIRECTORY` and `FileHeaders.REMOTE_FILE` respectively.
+Starting with _version 5.0_, additional remote file information, in JSON, is provided in the `FileHeaders.REMOTE_FILE_INFO` header.
+If you set the `fileInfoJson` property on the `SftpStreamingMessageSource` to `false`, the `LsEntry` object provided by the underlying JSch library is used instead.
```
artembilan: Well, I see you document your solution.
But why this way?
Why that `FileInfo.getFileInfo()` isn't enough to get access to the underlying entity?

Seems for me our `FileInfo` is more convenient in most cases, especially when we have a common process after FTP as well as SFTP.
And as I said above: it isn't consistent with JSON representation.

Any additional argument to go your way?
Thanks

artembilan (changes requested):
Very close to merge IMO...

spring-integration-core/src/test/java/org/springframework/integration/json/SimpleJsonSerializerTests.java
garyrussell: No, the `SimpleJsonSerializer` is object to JSON only.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
garyrussell: Doh.

spring-integration-core/src/test/java/org/springframework/integration/json/SimpleJsonSerializerTests.java
garyrussell: Doh - ignore that 😄

artembilan: That's true, but if it is JSON String, the Jackson must be able to deserialize it into `Foo`.
I mean `.fromJson(json, Foo.class)`

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/inbound/FtpStreamingMessageSourceTests.java
```diff
@@ -95,7 +125,6 @@ public PollerMetadata defaultPoller() {
 		public MessageSource<InputStream> ftpMessageSource() {
 			FtpStreamingMessageSource messageSource = new FtpStreamingMessageSource(template(), null);
 			messageSource.setRemoteDirectory("ftpSource/");
-			messageSource.setFilter(new FtpPersistentAcceptOnceFileListFilter(new SimpleMetadataStore(), "streaming"));
```
artembilan: What is the purpose of this removal?

garyrussell: So the new part of the test (with FileInfo) has some results.

src/reference/asciidoc/sftp.adoc
garyrussell: Good point - changing.

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/inbound/FtpStreamingMessageSourceTests.java
garyrussell: We'll have to fix this when your PR is merged.

artembilan: Got it! Let's merge your one then I'll rebase my and we shell see how is it going!

spring-integration-file/src/main/java/org/springframework/integration/file/FileHeaders.java
```diff
@@ -47,4 +47,10 @@
 	 */
 	public static final String MARKER = PREFIX + "marker";
 
+	/**
+	 * JSON representation of remote file information (if a JSON object mapper is
+	 * available).
```
artembilan: Out-dated JavaDoc.
Will fix on merge

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/66b53e749fee06413860ad98924ac1357e276f54

PR #2066, opened 2017-02-23
INT-4236: Status for SmartLifecycleRoleControlller

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4236

Add methods to obtain status from the `SmartLifecycleRoleController`.

__backport will require removal of diamonds and stream operations - I will issue a separate PR after review/merge__

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,14 @@
 package org.springframework.integration.support;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -34,6 +38,7 @@ import org.springframework.context.SmartLifecycle;
 import org.springframework.integration.leader.event.AbstractLeaderEvent;
 import org.springframework.integration.leader.event.OnGrantedEvent;
 import org.springframework.integration.leader.event.OnRevokedEvent;
+import org.springframework.integration.support.context.NamedComponent;
 import org.springframework.util.Assert;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
```diff
@@ -185,7 +190,76 @@ public void stopLifecyclesInRole(String role) {
...
+	 * @since 4.3.8
+	 */
+	public boolean allEndpointsRunning(String role) {
+		Map<String, Boolean> status = getEndpointsRunningStatus(role);
+		if (status.isEmpty()) {
+			return false;
+		}
+		else {
+			return status.values().stream().allMatch(b -> b);
+		}
```
artembilan: Can be simplified:
```
return !status.isEmpty() && status.values().stream().allMatch(b -> b);
```

spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
```diff
@@ -185,7 +190,76 @@ public void stopLifecyclesInRole(String role) {
...
+	 * @return true if there are no endpoints or none are running.
+	 * @since 4.3.8
+	 */
+	public boolean noEndpointsRunning(String role) {
+		Map<String, Boolean> status = getEndpointsRunningStatus(role);
+		if (status.isEmpty()) {
+			return true;
+		}
+		else {
+			return status.values().stream().allMatch(b -> !b);
```
artembilan: Negative ``allMatch` can be replaced with the `noneMatch`.
Plus `if...else` can be simplified too:
```
return status.isEmpty() || status.values().stream().noneMatch(b -> b);
```
That is what IDEA does for me! 😄

spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
```diff
@@ -185,7 +190,76 @@ public void stopLifecyclesInRole(String role) {
...
+			addLazyLifecycles();
+		}
+		if (!this.lifecycles.containsKey(role)) {
+			return Collections.emptyMap();
+		}
+		Map<String, Boolean> runningStatus = new HashMap<>();
+		AtomicInteger index = new AtomicInteger();
+		this.lifecycles.get(role)
+				.stream()
+				.forEach(e -> {
```
artembilan: Just in case:
redundant_stream

artembilan: OTOH, there is something like this:
````java
return this.lifecycles.get(role)
		.stream()
		.collect(Collectors.toMap(e ->
						(e instanceof NamedComponent)
								? ((NamedComponent) e).getComponentName()
								: (e.getClass().getSimpleName() + "#" + index.getAndIncrement()),
				SmartLifecycle::isRunning));
````
http://stackoverflow.com/questions/20363719/java-8-listv-into-mapk-v

File: src/reference/asciidoc/endpoint.adoc [hunk 4]
```diff
@@ -633,6 +633,27 @@ start/stop its configured `SmartLifecycle` objects when leadership is granted/re
...
+public Map<String, Boolean> getEndpointsRunningStatus(String role) <4>
+----
+
+<1> Returns a list of the roles being managed.
+
+<2> Returns true if all endpoints in the role are running.
+
+<3> Returns true if none of the endpoints in the role are running.
+
+<4> Returns a map of `compoonent name : running status` - the component name is usually the bean name.
```
artembilan: Typo: comp~~o~~onent

artembilan (changes requested):
That's all
I have all those changes locally.
Can merge if you agree.

garyrussell:
Good; thanks!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f23360d53955e2e1ca13c8f462680e5562cd2a1a.
Pay attention how I have fixed log message. Might be useful for backport PR.

PR #2068, opened 2017-02-23
INT-4233: Prework - Simple JSON Serializer

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4233

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/json/SimpleJsonSerializer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/json/SimpleJsonSerializer.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.json;
+
+import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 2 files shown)


garyrussell:
I am going to change it to a utility and make it stateless.

garyrussell:
Should be good to go now.

spring-integration-core/src/main/java/org/springframework/integration/json/SimpleJsonSerializer.java
```diff
@@ -0,0 +1,95 @@
...
+			catch (InvocationTargetException e) {
+				if (logger.isDebugEnabled()) {
+					logger.debug("Failed to serialize property " + method.getName(), e);
+				}
+				result = e.getMessage();
+			}
+			stringBuilder.append(toElement(result)).append(",");
+		}, method -> {
+			String name = method.getName();
+			return (name.length() > 3 && name.startsWith("get") && propertyMap.keySet().contains(name.substring(3)))
```
artembilan: Can't we just use `BeanUtils.getPropertyDescriptors()`?

OTOH wouldn't it be better to introspect all the properties unless `ignoreProperties` provided instead?
For example the `FileInfo` should be serialized fully except `getFileInfo()` result. 
Isn't it?

spring-integration-core/src/test/java/org/springframework/integration/json/SimpleJsonSerializerTests.java
```diff
@@ -0,0 +1,76 @@
...
+		Foo foo = new Foo();
+		String json = SimpleJsonSerializer.toJson(foo, "foo", "bar", "dub", "bool");
+		assertThat(json, startsWith("{"));
+		assertThat(json, containsString("\"bool\":true"));
+		assertThat(json, containsString("\"bar\":42"));
+		assertThat(json, containsString("\"foo\":\"bar\""));
+		assertThat(json, containsString("\"dub\":1.6"));
+		assertThat(json, endsWith("}"));
+		String[] split = json.split(",");
+		assertThat(split, arrayWithSize(4));
```
artembilan: ??? Why don't test that it is fully compatible with JSON expected by the Jackson for example?
I mean serialize our way, but deserialize to `Foo` with Jackson.

PR #2070, opened 2017-02-24
INT-4224: Improve Direct Handler Binding Docs

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4224

Explain when `ref` can can be used on certain endpoints.


artembilan:
... and cherry-picked.

PR #2075, opened 2017-03-01
INT-4237: FWMH: Acquire Lock Before Flushing

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4237

It was possible to flush (close) the file while a write was in process; more likely when
`flushWhenIdle` is false.

This probably would not occur in the real world, just tests with short flush intervals,
but certainly possible.

There is already a lock used to prevent concurrent writes while appending; use the same
lock when flushing.


artembilan:
And cherry-picked as https://github.com/spring-projects/spring-integration/commit/e872791a69ddc2b615050b9edc437d1267e53e80 after resolving conflicts.

PR #2076, opened 2017-03-01
INT-4219: Add @UseSpelInvoker

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4219

Needs (a lot) more work, tests, docs.

Just want to get buy-in for the approach.

Also see note about resolving placeholders for compiler mode.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/annotation/UseSpelInvoker.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/annotation/UseSpelInvoker.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.core.annotation.AliasFor;
+
+/**
+ * Indicates that a POJO handler method ({@code @ServiceActivator, @Transformer, } etc.,
+ * or such methods invoked from XML definitions) should be invoked using SpEL.
+ * <p>In versions prior to 5.0, such methods were always invoked using SpEL. In 5.0, the
+ * framework switched to using
+ * {@link org.springframework.messaging.handler.invocation.InvocableHandlerMethod} instead
+ * which is generally more efficient than (interpreted) SpEL.
+ * <p>There may be some unanticipated corner case where it is necessary to revert to using
+ * SpEL. Also, for very high performance requirements, you may wish to consider using
+ * compiled SpEL which is often the fastest solution (when the expression is compilable).
+ * <p>Applying this annotation to those methods will cause SpEL to be used for the
+ * invocation. An optional compilerMode property (aliased to value) is also provided.
+ *
+ * @author Gary Russell
+ */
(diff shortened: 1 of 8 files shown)


artembilan (changes requested):
I we're good to go with this approach and that doesn't look so ugly if you don't like to use `InvocableMessageHandler`.

I don't see how you apply that `compilerMode`, but I guess that willl in upcoming commits.

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -243,6 +243,10 @@ protected String getInputChannelAttribute() {
 		return INPUT_CHANNEL_ATTRIBUTE;
 	}
 
+	protected ConfigurableListableBeanFactory getBeanFactory() {
+		return this.beanFactory;
```
garyrussell: Yeah - this is some left-over dross from when I was trying to use the BF too early.

spring-integration-core/src/main/java/org/springframework/integration/annotation/UseSpelInvoker.java
```diff
@@ -0,0 +1,71 @@
...
+ * <p>Applying this annotation to those methods will cause SpEL to be used for the
+ * invocation. An optional compilerMode property (aliased to value) is also provided.
+ *
+ * @author Gary Russell
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Documented
+public @interface UseSpelInvoker {
```
garyrussell: I don't think so; we will need that for XML config (either a new attribute or they can autowire the handler and set the flag).

artembilan: ??? Even for XML config the POJO method can be marked with the `@UseSpelInvoker`.
We don't prohibit to use `@ServiceActivator` and similar when we parse XML config.
Finally everything ends up in the `MessagingMethodInvokerHelper` which just reads a `Method` for particular options and generates SpEL and `InvocableHandlerMethod` for runtime.

garyrussell: The problem is they might want to invoke a class/method they don't own and can't annotate.

garyrussell:
Next iteration.

artembilan:
Well, I think the progress is good here.
You can go ahead what you want to do here else.

And I see your point about an external class to use from XML.
There is always a workaround like local wrapper around that remote method.
My point just don't spend time for XML effort.
At least until strong requirements.

garyrussell:
>My point just don't spend time for XML effort.

Agreed.

artembilan:
FYI: there are some failures according Travis report.

artembilan:
Well, looks like we need only docs on the matter already!
And resolving conflicts after rebase.
Do you have anything else in mind to do for this task ?

garyrussell:
PR soon (tests and docs) - doing a `testAll` right now 😄

spring-integration-core/src/main/java/org/springframework/integration/annotation/UseSpelInvoker.java
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * Specify that the annotated method (or methods in the annotated class) will be
+	 * invoked using SpEL instead of an
+	 * {@link org.springframework.messaging.handler.invocation.InvocableHandlerMethod}
+	 * with the specified compilerMode. If left empty, the default runtime compiler
+	 * mode will be used. Must evaluate to a String containing a valide compiler mode.
```
artembilan: Typo: valid

Will fix on merge

spring-integration-core/src/main/java/org/springframework/integration/annotation/UseSpelInvoker.java
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	@AliasFor("compilerMode")
+	String value() default "";
+
+	/**
+	 * Specify that the annotated method (or methods in the annotated class) will be
+	 * invoked using SpEL instead of an
+	 * {@link org.springframework.messaging.handler.invocation.InvocableHandlerMethod}
+	 * with the specified compilerMode. If left empty, the default runtime compiler
+	 * mode will be used. Must evaluate to a String containing a valide compiler mode.
```
artembilan: DITO.
Copy/paste artefact 😄

artembilan (approved):
A couple (in fact one) typos.
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/a50d866f81888f27c56f7298d946b786c52e9281

PR #2077, opened 2017-03-02
INT-4100: Improve `RequestReplyExchanger`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4100

* Add `throws MessagingException` to the `RequestReplyExchanger` to make it fully messaging contract based
* Add test to the `ServiceActivatorDefaultFrameworkMethodTests` to demonstrate how many info is in the stack trace after `RequestReplyExchanger`
* Document how to be with the `throws` clause in custom gateways
* Fix `boolean` attributes in the `spring-integration-5.0.xsd`

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/handler/ServiceActivatorDefaultFrameworkMethodTests-context.xml
+++ b/spring-integration-core/src/test/java/org/springframework/integration/handler/ServiceActivatorDefaultFrameworkMethodTests-context.xml
@@ -42,9 +42,9 @@
 
 	<gateway id="gateway" default-request-channel="requestChannel" default-reply-channel="replyChannel"/>
 
-	<channel id="requestChannel"/>
-
-	<bridge id="bridge" input-channel="requestChannel" output-channel="replyChannel"/>
+	<chain id="bridge" input-channel="requestChannel" output-channel="replyChannel">
+		<transformer expression="payload == 'foo' ? T(org.springframework.integration.handler.ServiceActivatorDefaultFrameworkMethodTests).throwIllegalStateException('Wrong payload') : payload"/>
+	</chain>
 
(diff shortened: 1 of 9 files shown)


spring-integration-core/src/test/java/org/springframework/integration/handler/ServiceActivatorDefaultFrameworkMethodTests-context.xml
```diff
@@ -42,9 +42,9 @@
 
 	<gateway id="gateway" default-request-channel="requestChannel" default-reply-channel="replyChannel"/>
 
-	<channel id="requestChannel"/>
-
-	<bridge id="bridge" input-channel="requestChannel" output-channel="replyChannel"/>
+	<chain id="bridge" input-channel="requestChannel" output-channel="replyChannel">
```
garyrussell: We should change the id - it's no longer a bridge.

garyrussell (commented):
That's all, but we need a what's new for the `RequestReplyExchanger` behavior change (for people who are using it and expect the exception to be unwrapped) - advise them to use their own interface. 

Also a migration guide entry.

garyrussell:
Hmmm - you didn't address... `... but we need a what's new ...`

artembilan:
Also see https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.3-to-5.0-Migration-Guide#requestreplyexchanger-contract

garyrussell:
Doc polishing here for your review https://github.com/garyrussell/spring-integration/commit/a7feb979ab46311ddd712756a5f18b2fa944c5a8

garyrussell:
Merged as 8da89c7cce0cdd20407f5a9ae352d0c9923c16dc

PR #2079, opened 2017-03-05
INT-4238: MQTT Doesn't Detect a Subscription Error

defshine:
INT-4238: MQTT Doesn't Detect a Subscription Error

    JIRA: https://jira.spring.io/browse/INT-4238

    * add `doSubscribe`  method to check SUBACK

Check subscribe ack after connect and subscribe
If return qos is 0x80, that means failure according to [mqtt protocol](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/errata01/os/mqtt-v3.1.1-errata01-os-complete.html#_Toc442180884)

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
@@ -22,6 +22,7 @@ import java.util.concurrent.ScheduledFuture;
 
 import org.eclipse.paho.client.mqttv3.IMqttAsyncClient;
 import org.eclipse.paho.client.mqttv3.IMqttDeliveryToken;
+import org.eclipse.paho.client.mqttv3.IMqttToken;
 import org.eclipse.paho.client.mqttv3.MqttCallback;
 import org.eclipse.paho.client.mqttv3.MqttConnectOptions;
 import org.eclipse.paho.client.mqttv3.MqttException;
@@ -237,8 +238,7 @@ public class MqttPahoMessageDrivenChannelAdapter extends AbstractMqttMessageDriv
 		try {
 			this.client.connect(connectionOptions)
 					.waitForCompletion(this.completionTimeout);
-			this.client.subscribe(getTopic(), getQos())
-					.waitForCompletion(this.completionTimeout);
+			doSubscribe(getTopic(), getQos());
 		}
 		catch (MqttException e) {
 			if (this.applicationEventPublisher != null) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@defshine Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2079)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2079) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@defshine Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2079)!

garyrussell:
@defshine Thanks for the contribution !

Please note that we track bugs using JIRA - [INT-4238](https://jira.spring.io/browse/INT-4238).

We also require tests.

Also, `./gradlew :spring-integration-mqtt:check` would have told you we don't allow wildcard imports.

Please follow the [Contribution Guidelines](https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc). For example, commit comment headlines should be limited to 50 characters.

I am providing this for your future information; we can take care of polishing this PR if you like.

Thanks again for your contribution.

defshine:
@garyrussell 
Thank you!
I have change the commit.

artembilan:
Interesting...
Exactly this logic is implemented in the `MqttClient`:
````java
public void subscribe(String[] topicFilters, int[] qos) throws MqttException {
	IMqttToken tok = aClient.subscribe(topicFilters, qos, null, null);
	tok.waitForCompletion(getTimeToWait());
	int[] grantedQos = tok.getGrantedQos();
	for (int i = 0; i < grantedQos.length; ++i) {
		qos[i] = grantedQos[i];
	}
	if (grantedQos.length == 1 && qos[0] == 0x80) {
		throw new MqttException(MqttException.REASON_CODE_SUBSCRIBE_FAILED);
	}
}
````
And pay attention how this `sync` client delegates to the internal `async` one.
I wonder if we really need to worry about the client type and can just come back to the `sync` client without any impacts, but including already this `0x80` requirement as an out-of-the-box feature. 

@garyrussell , your thoughts?

@defshine ,

Thank you, of course, for the contribution but as Gary mentioned above we don't accepts PRs without tests.

**Comment by garyrussell** [comment]
@artembilan From the commit when we switched to using the Async client (primarily for the outbound adapter)...

```
Also use the async client for the inbound adapter; while
it doesn't make any performance difference, it does
allow us to timeout the disconnect, which we have seen
to cause hangs on the CI servers.
```

So it seems we have no choice but to replicate the code here.

**Comment by artembilan** [comment]
That's good I don't mind!

Really isn't clear why they do like this in the `MqttClient`:
```
public void disconnect() throws MqttException {
	aClient.disconnect().waitForCompletion();
}
```
without consulting `getTimeToWait()` like in other places...

Although why we can't consider to use:
```
public void disconnectForcibly(long disconnectTimeout) throws MqttException {
	aClient.disconnectForcibly(disconnectTimeout);
}
```
somewheren deep in the code it does:
```
// Wait util the disconnect packet sent with timeout
token.waitForCompletion(disconnectTimeout);
```
Looks like fully parallel what we have now in the `MqttPahoMessageDrivenChannelAdapter`.

Another observation. The `MqttClient` has `protected MqttAsyncClient aClient`, so we can extend it internally and get access to the `MqttAsyncClient` if we still need it.

And again we will have code as much cleaner as possible using the maximum from out-of-the-box.

Just IMHO to share with you, guys...

I won't mind to accepts the PR and merge with minimal changes.
Just wonder what we have missed else to fix similar way in the future...

garyrussell:
@artembilan Thanks for digging into the paho client - yes, I agree we can switch back to using the sync client here and use `disconnectForcibly` with the timeout.

@defshine Many thanks for pointing out the issue; I am closing this PR and will replace it with the above logic.

Looking forward to future contributions!!

Thanks again.

PR #2080, opened 2017-03-05
INT-4101 MessagingException to include failedMessage

m-terra:
JIRA: https://jira.spring.io/browse/INT-4101

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -33,15 +33,15 @@ import org.springframework.integration.support.channel.BeanFactoryChannelResolve
 import org.springframework.integration.transaction.ExpressionEvaluatingTransactionSynchronizationProcessor;
 import org.springframework.integration.transaction.IntegrationResourceHolder;
 import org.springframework.integration.transaction.IntegrationResourceHolderSynchronization;
+import org.springframework.integration.transaction.PassthroughTransactionSynchronizationFactory;
 import org.springframework.integration.transaction.TransactionSynchronizationFactory;
 import org.springframework.integration.util.ErrorHandlingTaskExecutor;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
-import org.springframework.messaging.MessageHandlingException;
 import org.springframework.messaging.MessagingException;
-import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.scheduling.Trigger;
 import org.springframework.scheduling.support.PeriodicTrigger;
+import org.springframework.transaction.interceptor.TransactionInterceptor;
 import org.springframework.transaction.support.TransactionSynchronization;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
@@ -54,6 +54,7 @@ import org.springframework.util.ErrorHandler;
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Andreas Baer
  */
 public abstract class AbstractPollingEndpoint extends AbstractEndpoint implements BeanClassLoaderAware {
 
@@ -110,7 +111,8 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 		this.beanClassLoader = classLoader;
 	}
 
-	public void setTransactionSynchronizationFactory(TransactionSynchronizationFactory transactionSynchronizationFactory) {
+	public void setTransactionSynchronizationFactory(TransactionSynchronizationFactory
+			transactionSynchronizationFactory) {
 		this.transactionSynchronizationFactory = transactionSynchronizationFactory;
 	}
 
@@ -166,6 +168,11 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 						this.errorHandlerIsDefault = true;
 					}
 					this.taskExecutor = new ErrorHandlingTaskExecutor(this.taskExecutor, this.errorHandler);
+				}
+			}
+			if (this.transactionSynchronizationFactory == null && this.adviceChain != null) {
+				if (this.adviceChain.stream().anyMatch(TransactionInterceptor.class::isInstance)) {
+					this.transactionSynchronizationFactory = new PassthroughTransactionSynchronizationFactory();
 				}
 			}
 			this.initialized = true;
@@ -300,19 +307,19 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 	}
 
 	private IntegrationResourceHolder bindResourceHolderIfNecessary(String key, Object resource) {
-
-		if (this.transactionSynchronizationFactory != null && resource != null) {
-			if (TransactionSynchronizationManager.isActualTransactionActive()) {
-				TransactionSynchronization synchronization = this.transactionSynchronizationFactory.create(resource);
-				TransactionSynchronizationManager.registerSynchronization(synchronization);
-				if (synchronization instanceof IntegrationResourceHolderSynchronization) {
-					IntegrationResourceHolder holder =
-							((IntegrationResourceHolderSynchronization) synchronization).getResourceHolder();
-					if (key != null) {
-						holder.addAttribute(key, resource);
-					}
-					return holder;
+		if (this.transactionSynchronizationFactory != null && resource != null &&
+				TransactionSynchronizationManager.isActualTransactionActive()) {
+			TransactionSynchronization synchronization = this.transactionSynchronizationFactory.create(resource);
+			TransactionSynchronizationManager.registerSynchronization(synchronization);
+			if (synchronization instanceof IntegrationResourceHolderSynchronization) {
+				IntegrationResourceHolderSynchronization integrationSynchronization =
+						((IntegrationResourceHolderSynchronization) synchronization);
+				integrationSynchronization.setShouldUnbindAtCompletion(false);
+				IntegrationResourceHolder resourceHolder = integrationSynchronization.getResourceHolder();
+				if (key != null) {
+					resourceHolder.addAttribute(key, resource);
 				}
+				return resourceHolder;
 			}
 		}
 		return null;
@@ -343,11 +350,26 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 						count++;
 					}
 					catch (Exception e) {
-						if (e instanceof RuntimeException) {
-							throw (RuntimeException) e;
+						if (e instanceof MessagingException) {
+							throw (MessagingException) e;
 						}
 						else {
-							throw new MessageHandlingException(new ErrorMessage(e), e);
+							Message<?> failedMessage = null;
+							if (AbstractPollingEndpoint.this.transactionSynchronizationFactory != null) {
+								Object resource = TransactionSynchronizationManager.getResource(getResourceToBind());
+								if (resource instanceof IntegrationResourceHolder) {
+									failedMessage = ((IntegrationResourceHolder) resource).getMessage();
+								}
+							}
+							throw new MessagingException(failedMessage, e);
+						}
+					}
+					finally {
+						if (AbstractPollingEndpoint.this.transactionSynchronizationFactory != null) {
+							Object resource = getResourceToBind();
+							if (TransactionSynchronizationManager.hasResource(resource)) {
+								TransactionSynchronizationManager.unbindResource(resource);
+							}
 						}
 					}
 				}
--- a/spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2014-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import org.springframework.transaction.support.ResourceHolderSynchronization;
  * The base {@link ResourceHolderSynchronization} for {@link IntegrationResourceHolder}.
  *
  * @author Artem Bilan
+ * @author Andreas Baer
  * @since 4.0
  */
 public class IntegrationResourceHolderSynchronization
--- a/spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
@@ -24,10 +24,12 @@ import org.springframework.transaction.support.ResourceHolderSynchronization;
  * @author Artem Bilan
  * @since 4.0
  */
-public abstract class IntegrationResourceHolderSynchronization
+public class IntegrationResourceHolderSynchronization
 		extends ResourceHolderSynchronization<IntegrationResourceHolder, Object> {
 
 	protected final IntegrationResourceHolder resourceHolder;
+
+	private boolean shouldUnbindAtCompletion = true;
 
 	public IntegrationResourceHolderSynchronization(IntegrationResourceHolder resourceHolder,
 			Object resourceKey) {
@@ -39,4 +41,12 @@ public abstract class IntegrationResourceHolderSynchronization
 		return this.resourceHolder;
 	}
 
+	public void setShouldUnbindAtCompletion(boolean shouldUnbindAtCompletion) {
+		this.shouldUnbindAtCompletion = shouldUnbindAtCompletion;
+	}
+
+	@Override
+	protected boolean shouldUnbindAtCompletion() {
+		return this.shouldUnbindAtCompletion;
+	}
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
@@ -25,7 +25,7 @@ import org.springframework.util.Assert;
  * {@link TransactionSynchronization} using {{@link #create(Object)} and registers the
  * resource in the {@link TransactionSynchronizationManager}
  *
- * @author Andreas Bär
+ * @author Andreas Baer
  * @since 5.0
  */
 public class PassthroughTransactionSynchronizationFactory implements TransactionSynchronizationFactory {
--- a/spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
@@ -22,8 +22,8 @@ import org.springframework.util.Assert;
 
 /**
  * A simple implementation of {@link TransactionSynchronizationFactory} which creates a
- * {@link TransactionSynchronization} using {{@link #create(Object)} and registers the
- * resource in the {@link TransactionSynchronizationManager}
+ * {@link IntegrationResourceHolder} under provided {@code key} and registers the
+ * resource in the {@link TransactionSynchronizationManager}.
  *
  * @author Andreas Baer
  * @since 5.0
--- a/spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.transaction;
+
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+import org.springframework.util.Assert;
+
+/**
+ * A simple implementation of {@link TransactionSynchronizationFactory} which creates a
+ * {@link TransactionSynchronization} using {{@link #create(Object)} and registers the
+ * resource in the {@link TransactionSynchronizationManager}
+ *
+ * @author Andreas Bär
+ * @since 5.0
+ */
+public class PassthroughTransactionSynchronizationFactory implements TransactionSynchronizationFactory {
+
+
+	@Override
+	public TransactionSynchronization create(Object key) {
+		Assert.notNull(key, "'key' must not be null");
+		IntegrationResourceHolderSynchronization synchronization = new IntegrationResourceHolderSynchronization(new
+				IntegrationResourceHolder(), key);
+		TransactionSynchronizationManager.bindResource(key, synchronization.getResourceHolder());
+		return synchronization;
+	}
+
+}
--- a/spring-integration-core/src/test/java/org/springframework/integration/dispatcher/PollingTransactionTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/dispatcher/PollingTransactionTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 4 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -343,13 +345,16 @@ public void run() {
...
 					}
 					catch (Exception e) {
-						if (e instanceof RuntimeException) {
-							throw (RuntimeException) e;
+						if (e instanceof MessagingException) {
+							throw (MessagingException) e;
 						}
 						else {
-							throw new MessageHandlingException(new ErrorMessage(e), e);
+							throw new MessagingException(messageHolder.get(), e);
```
artembilan: I think, I'm fine with the solution.
I thought that we don't that `ThreadLocal` when we have a TX, but indeed it is already finished and cleaned on the `Poller.this.pollingTask.call()`. And if TX throws exception during commit, we really don't have a TX-synched resource to consult for the `Message`.

So, LGTM here, although I still think that it should be `MessageHandlingException`. Just because it isn't so generic case, but really the `handleMessage(Message)` problem.

spring-integration-core/src/test/java/org/springframework/integration/endpoint/PollingEndpointErrorHandlingTests.java
```diff
@@ -41,6 +42,10 @@ public void checkExceptionPlacedOnErrorChannel() {
 		Message errorMessage = errorChannel.receive(5000);
 		assertNotNull("No error message received", errorMessage);
 		assertEquals("Message received was not an ErrorMessage", ErrorMessage.class, errorMessage.getClass());
+		assertEquals("ErrorMessage payload was not a MessagingException", MessagingException.class, errorMessage.getPayload().getClass());
```
artembilan: Would be better do not exceed `120` line length.

artembilan (changes requested):
A couple simple concerns.

@garyrussell , the solution looks so simple, that I worry I'm missing something.

LGTM, but would like to see your opinion.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
garyrussell: Actually, I don't think `MessageHandlingException` is correct here; if the exception occurred on a downstream handler, it will already be an `MHE`; it not, then it's not really an MHE it's an "unknown" type of ME.

I wonder if we should take the opportunity we discussed last week - i.e. an enhanced messaging exception where the payload contains the failedMessage as well as this (`inputMessage`).

garyrussell:
Overall, it looks ok, but I don't like the overhead of maintaining a `ThreadLocal` on every poll, for every poller since it's only when a non-messaging-aware advice (such as TxAdvice) throws an exception that we need to create the enhanced exception. That's why I originally suggested adding an advice to the advice chain in the JIRA.

This branch has a non-linear commit log; it needs to be rebased to master (and conflicts resolved).

andybaer:
I don't like to use ThreadLocal's as well but we needed a fix and this one did it. The advice approach I out of my league ;)
I did not notice the conflicts and I can change the formatting of course. Please let me know if I should proceed.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
artembilan: Well, that is fully different story (enhanced exception) and let's consider it in the separate task!

And well, let it be just general `ME`!

artembilan:
Well another suggestion is something like end-user choice - a `MethodInterceptor` implementation which will be applied around `TxAdvice` in the `AbstractPollingEndpoint` and wrap an exception from there to the `MessagingException`. But still... how to obtain a `Message` there?

If we apply an advice just to the `handleMessage()`, we will be already within `TxAdvice`, therefore won't be able to interfere to the TX exception around.

I'm confused how you are going to proceed without `ThreadLocal`...

garyrussell:
>Well, that is fully different story (enhanced exception) and let's consider it in the separate task!

Well, if we agree that will be useful (I agree it's a separate issue), then we will need the `ThreadLocal` anyway so let's go ahead and polish and merge this - @artembilan you or me?

Thanks @andybaer !!

andybaer:
I rebased and reformatted. Thanks to both of you for considering this fix!

garyrussell:
>I'm confused how you are going to proceed without ThreadLocal...

The TL would be in the advice instead, and therefore only used when the advice is configured - we can still consider some optimization later, perhaps before GA; let's go with it as-is for now and revisit if/when we consider the enhanced ME.

garyrussell:
>I'm confused how you are going to proceed without ThreadLocal...

The message would be a local variable in the advice instead (captured before the `proceed()`) - we can still consider some optimization later, perhaps before GA; let's go with it as-is for now and revisit if/when we consider the enhanced ME.

andybaer:
Had some trouble rebasing but now the commit history is linear

artembilan:
Guys, looking to the problem one more time, I don't see (yet) some generic solution.
Let's fix the current problem and come back if we will figure out what is needed.

So, as we know the `handleMessage(Message)` throws `MessagingException` with the "guilty" message.
Therefore we're good in the `errorChannel` flow and have some message context.

When we come to the TX and error during commit, we are already far away from the `handleMessage(Message)` and just lose the "guilty" message context.
Fortunately we bind to the TX there `IntegrationResourceHolder` via `IntegrationResourceHolderSynchronization`, which has an option like `shouldUnbindAtCompletion()` with `true` by default.
We can switch it into `false` and unbind manually in the same place like @andybaer does it now.
WDYT?

andybaer:
Sound like a nice solution. This implies a public setter for `shouldUnbindAtCompletion()`. I can change it to this implementation if you agree.

garyrussell:
Sounds good to me, but it will require a `TransactionSynchronizationFactory` configured into the poller (that's when we save off the message).

Perhaps we could always configure a pass-through one on all transactional pollers?

artembilan:
Well, that's true.

So, the way to go:
1. Add setter to the `IntegrationResourceHolderSynchronization`  to let to configure `shouldUnbindAtCompletion` and set it in the `AbstractPollingEndpoint` to `false`.
2. Use a new `catch` logic in the `Poller` class to extract `IntegrationResourceHolder` from the `TransactionSynchronizationManager` and `unbind` it in the `finally`
3. Introduce new `PassThroughTransactionSynchronizationFactory` to use when there is no provided `transactionSynchronizationFactory` and `TransactionSynchronizationManager.isActualTransactionActive()`.

This way we nail the premise of the problem after TX commit exception and we won't have an extra `ThreadLocal` variable.

The long way solution might be like `exposeInputMessage = true/false` on the `AbstractPollingEndpoint` and introduce something like `InputMessageContext` with the `ThreadLocal` already.
That `InputMessageContext` with `public static` methods can be used in any downstream component. 
Exactly similar way how we use these days `SecurityContextHolder`.

Any other thoughts?
Thanks

andybaer:
This sounds very interesting. Let me give it a shot and I will come back on the same pull request.

Thanks a lot @artembilan & @garyrussell !

andybaer:
Hi @artembilan 

I gave it a try but `TransactionSynchronizationManager.clearSynchronization()` and `TransactionSynchronizationManager.clear()` both are invoked before the exception arrives in the poller's catch block. So the `shouldUnbindAtCompletion` setting is not preventing the advice from clearing the tx state. Code is on this branch: https://github.com/andybaer/spring-integration/tree/INT-4101-NEW

Any advice how to proceed?

Thanks!

artembilan:
@andybaer ,

thank you for the effort!

Well, even if `TransactionSynchronizationManager` is cleared that way in the end of transaction, pay attention that `resources` isn't cleared. So, we still have there `IntegrationResourceHolder` with a desired message. We should do `TransactionSynchronizationManager.unbindResource(this.resourceKey);` manually any way. Just because it is correct - we don't have TX any more, but we still pollute `ThreadLocal` .

artembilan:
OK. I see your problem.
You should use `TransactionSynchronizationManager.getResource()` instead of iteration over `TransactionSynchronizationManager.getSynchronizations()`

andybaer:
Yes thanks your comment was clear. Just running all the tests and rebase and then I will push to this pull request

andybaer:
@artembilan 

The new implementation using `TransactionSynchronizationManager` is finished. Please have a look.

Thanks!

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -168,6 +169,9 @@ protected void onInit() {
 					this.taskExecutor = new ErrorHandlingTaskExecutor(this.taskExecutor, this.errorHandler);
 				}
 			}
+			if (this.transactionSynchronizationFactory == null) {
+				this.transactionSynchronizationFactory = new PassthroughTransactionSynchronizationFactory();
+			}
```
garyrussell: We only want to take the hit of saving the message in a TL __if the poller is transactional__ - we can easily tell that in the XML parser and via the DSL spec, but not so easily via the `@Poller` annotation; we may need to add a `transactional` property to `@Poller` for that.

On the other hand, any "around" advice that can throw an exception may need access to the message if it throws the exception after the proceed.

So, I don't mind if we always apply it if the advice chain is > 0 length - that's probably the simplest solution that would work in all cases while only taking the extra unneeded hit if someone configures an advice chain.

artembilan: Well, my idea was to fix exactly the TX case. Therefore this `PassthroughTransactionSynchronizationFactory`.
We don't care about transactional poller or not but just use `TransactionSynchronizationManager.isActualTransactionActive()`.
So, that still should be enough for us here as well.
Any other "exotic" advices can access to the message in their after `proceed()` block via the same `ThreadLocal` solution but already as a population in the downstream flow.

Let me know if I have missed something or you need some other explanation.

spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
```diff
@@ -24,11 +24,13 @@
  * @author Artem Bilan
  * @since 4.0
  */
-public abstract class IntegrationResourceHolderSynchronization
+public class IntegrationResourceHolderSynchronization
 		extends ResourceHolderSynchronization<IntegrationResourceHolder, Object> {
 
 	protected final IntegrationResourceHolder resourceHolder;
 
+	private boolean shouldUnbindAtCompletion;
```
artembilan: `true` by default.
That is only `AbstractPollingEndpoint` to take care about the `TL` by its own.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -343,11 +346,21 @@ public void run() {
...
+							Object resource = TransactionSynchronizationManager.getResource(getResourceToBind());
+							if (resource instanceof IntegrationResourceHolder) {
+								failedMessage = ((IntegrationResourceHolder) resource).getMessage();
+							}
+							throw new MessagingException(failedMessage, e);
+						}
+					}
+					finally {
+						if (TransactionSynchronizationManager.hasResource(getResourceToBind())) {
+							TransactionSynchronizationManager.unbindResource(getResourceToBind());
```
artembilan: Minor, but we can extract local variable for the `getResourceToBind()`.
Not sure if compiler does an optimization.
Although it can't because of possibility to have a different result from each method invocation.

spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
```
artembilan: New classes only with the current year.

spring-integration-core/src/test/java/org/springframework/integration/dispatcher/FailingCommitTransactionManager.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.integration.util.TestTransactionManager;
+import org.springframework.transaction.IllegalTransactionStateException;
+import org.springframework.transaction.TransactionException;
+import org.springframework.transaction.support.DefaultTransactionStatus;
+
+/**
+ * @author Andreas Baer
+ */
+@SuppressWarnings("serial")
+public class FailingCommitTransactionManager extends TestTransactionManager {
```
artembilan: M-m-m. This looks like a test-case scope.
Let's move it into the test class - `PollingTransactionTests`!

spring-integration-core/src/test/java/org/springframework/integration/dispatcher/PollingTransactionTests.java
```diff
@@ -196,7 +200,53 @@ public void propagationMandatory() throws Throwable {
...
+		assertNotNull(messagingException.getFailedMessage());
+		assertNotNull(output.receive(0));
+		assertEquals(0, txManager.getCommitCount());
+		context.close();
+	}
+
+	@Test
+	public void handlerFalilureTest() throws Throwable {
+		ClassPathXmlApplicationContext context = new ClassPathXmlApplicationContext(
+				"transactionFailureTests.xml", this.getClass());
```
artembilan: Can't we combine two these new tests?
I'm very sensible to any build time increases, like the same app ctx start and close. 😄 

Well, I see that is a style of this test class at all.
Would be better to revise it to a managed test with a single ctx for better performance.
But different story...
OK, let's leave your code as is for that future refactoring.

spring-integration-core/src/test/java/org/springframework/integration/endpoint/PseudoTransactionalMessageSourceTests.java
```diff
@@ -342,18 +342,25 @@ public void testRollbackWithManagerUsingStatus() {
...
 		doPoll(adapter);
 		TransactionSynchronizationManager.setActualTransactionActive(false);
+		TransactionSynchronizationManager.unbindResource(source);
 
 		// Before INT-2777 this test was failed here
 		TransactionSynchronizationManager.setActualTransactionActive(true);
 		doPoll(adapter);
 		TransactionSynchronizationManager.setActualTransactionActive(false);
+		TransactionSynchronizationManager.clearSynchronization();
+		TransactionSynchronizationManager.unbindResource(source);
```
artembilan: M-m-m. What made you to do these changes?
Thanks

spring-integration-core/src/test/java/org/springframework/integration/endpoint/PseudoTransactionalMessageSourceTests.java
```diff
@@ -342,18 +342,25 @@ public void testRollbackWithManagerUsingStatus() {
 
 	@Test
 	public void testInt2777UnboundResourceAfterTransactionComplete() {
+		MessageSource<?> source = () -> null;
 		SourcePollingChannelAdapter adapter = new SourcePollingChannelAdapter();
+		adapter.setSource(source);
+		adapter.setOutputChannelName("bar");
+		adapter.setBeanFactory(new TestUtils.TestApplicationContext());
```
artembilan: `TestUtils.createTestApplicationContext()` is there for you.
OTOH you always can do `mock(BeanFactory.class)` if that satisfies a requirement.

spring-integration-test/src/main/java/org/springframework/integration/test/util/TestUtils.java
```diff
@@ -132,7 +132,7 @@ else if (beanFactory instanceof GenericApplicationContext) {
 
 	public static class TestApplicationContext extends GenericApplicationContext {
 
-		TestApplicationContext() {
+		public TestApplicationContext() {
```
artembilan: As I said: `TestUtils.createTestApplicationContext()` so, no changes here, please.

spring-integration-core/src/test/java/org/springframework/integration/dispatcher/PollingTransactionTests.java
```diff
@@ -196,7 +200,53 @@ public void propagationMandatory() throws Throwable {
...
+		TestTransactionManager txManager = (TestTransactionManager) context.getBean("txManagerGood");
+		PollableChannel input = (PollableChannel) context.getBean("inputHandlerFailure");
+		PollableChannel output = (PollableChannel) context.getBean("output");
+		PollableChannel errorChannel = (PollableChannel) context.getBean("errorChannel");
+		assertEquals(0, txManager.getCommitCount());
+		input.send(new GenericMessage<String>("commitFalilureTest"));
+		Message<?> errorMessage = errorChannel.receive(3000);
```
artembilan: Usually our CI machine is pretty slow, so 3 secs might not be enough to wait for the answer from another thread.
Right, I see that it is a practice of this test class, but let's move on to the more robust solution.
So, 10 sec is our typical time.

spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.springframework.integration.transaction;
+
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+import org.springframework.util.Assert;
+
+/**
+ * Stub implementation of {@link TransactionSynchronizationFactory} which allows you to create a
```
artembilan: M-m-m. I don't like "stub" word. Let's say "A simple TransactionSynchronizationFactory implementation which creates blah-blah and register blah-blah"

"Register" is very important word here because we do `TransactionSynchronizationManager.bindResource()` unconditionally.

artembilan (changes requested):
That's all for this round.
And, to be honest, it is so organic solution that I don't see reason to document it somehow...

spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
andybaer: oops

spring-integration-core/src/test/java/org/springframework/integration/endpoint/PseudoTransactionalMessageSourceTests.java
andybaer: This test did not bind resources to the `TransactionSynchronizationManager` before my changes because the `TransactionSynchronizationFactory` was null. Now that it does bind resources they must be cleared otherwise subsequent tests fail.

spring-integration-core/src/test/java/org/springframework/integration/endpoint/PseudoTransactionalMessageSourceTests.java
andybaer: Personal preference, I try to avoid mocking ;)

spring-integration-core/src/test/java/org/springframework/integration/dispatcher/PollingTransactionTests.java
andybaer: I agree, no big change

andybaer:
All checked in, thanks for the review!

Not sure about the consent on when the presence of `PassthroughTransactionSynchronizationFactory`

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
garyrussell: `TSM.iATA()` accesses a thread local - since 99.99% of pollers are __not transactional__ I'd prefer to not take two TL hits (here and in the poller's finally block) on every poll - a simple `null` check on the TSF field will be much more efficient.

Something less than 99.99%, but probably > 98% of pollers don't have an advice chain at all, so I can live with that (presence of an advice chain) being the condition we use to decide whether or not to store the message in a TL.

@andybaer You don't need to take any action until @artembilan and I have resolved our differences 😄

artembilan: OK, Gary! 

That is fully different story. Sorry, that wasn't obvious by your comment that you wouldn't like to hit `ThreadLocal` 😄 

Does that mean that we have to revise all our `TSM.iATA()` usage
For example `ConnectionFactoryUtils`:
```
public static RabbitResourceHolder bindResourceToTransaction(RabbitResourceHolder resourceHolder,
		ConnectionFactory connectionFactory, boolean synched) {
	if (TransactionSynchronizationManager.hasResource(connectionFactory)
			|| !TransactionSynchronizationManager.isActualTransactionActive() || !synched) {
		return (RabbitResourceHolder) TransactionSynchronizationManager.getResource(connectionFactory);
	}
```
And `ImapIdleChannelAdapter`:
```
if (TransactionSynchronizationManager.isActualTransactionActive()) {
		if (ImapIdleChannelAdapter.this.transactionSynchronizationFactory != null) {
```
?

I don't mind about optimization. 
So, another trick is check if our `adviceChain` contains `TransactionInterceptor`.
If it is there and `TSF == null` we populate this our new `PassthroughTransactionSynchronizationFactory`

Does that work for you?

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -343,11 +346,22 @@ public void run() {
...
+							Message<?> failedMessage = null;
+							Object resource = TransactionSynchronizationManager.getResource(getResourceToBind());
+							if (resource instanceof IntegrationResourceHolder) {
+								failedMessage = ((IntegrationResourceHolder) resource).getMessage();
+							}
+							throw new MessagingException(failedMessage, e);
+						}
+					}
+					finally {
+						Object resource = getResourceToBind();
```
artembilan: There is one more in this method - `TransactionSynchronizationManager.getResource(getResourceToBind())`

spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
```diff
@@ -0,0 +1,43 @@
...
+package org.springframework.integration.transaction;
+
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+import org.springframework.util.Assert;
+
+/**
+ * A simple implementation of {@link TransactionSynchronizationFactory} which creates a
+ * {@link TransactionSynchronization} using {{@link #create(Object)} and registers the
+ * resource in the {@link TransactionSynchronizationManager}
```
artembilan: Would be better to say `{@link IntegrationResourceHolder} under provided {@code key}` instead.
Plus the general class JavaDoc must end with a period.

spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
```diff
@@ -24,11 +24,13 @@
  * @author Artem Bilan
  * @since 4.0
  */
-public abstract class IntegrationResourceHolderSynchronization
+public class IntegrationResourceHolderSynchronization
```
artembilan: Since you did some changes in this class, add your name to the `@author` and update Copyright to the actual year. In all affected classes

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
garyrussell: We should probably take a look at other uses but I am not too concerned about, say, ImapIdle but, in this case, we would have added unnecessary overhead to a zillion pollers worldwide.

As I said, I don't mind if the test is for a non-empty advice chain, but if you want to further optimize the test to look for a tx interceptor, I am ok with that too. The only problem with that would be if someone wraps the tx interceptor for some reason.

spring-integration-core/src/test/java/org/springframework/integration/endpoint/PseudoTransactionalMessageSourceTests.java
```diff
@@ -402,6 +408,7 @@ protected void doPoll(SourcePollingChannelAdapter adapter) {
 			method.invoke(adapter);
 		}
 		catch (Exception e) {
+			e.printStackTrace();
```
artembilan: No `e.printStackTrace()`, please, at all.
It doesn't provide much info and just pollutes logs on CI builds

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
artembilan: Well, for that purpose we can document:
"We have `PassthroughTransactionSynchronizationFactory` if you would like to have `TL` value for message in case of custom TX Advice. It is applied by default if TX Advice is exactly `TransactionInterceptor`."

artembilan:
@andybaer ,

Please, stop to `squash` commits. It is much easier for us to review just small peaces of the changes.
Plus now we have lost all the history of the changes.

andybaer:
Sorry @artembilan

I thought that was prefered since there were many usless commits. But I see your point. No more squashing! :)

artembilan:
OK, @andybaer , now we are very close to the merge.
Please, try to address the rest couple my concerns and add a not into `transactions.adoc` about this our new `PassthroughTransactionSynchronizationFactory`.

@garyrussell , any other concerns from your side?

Thanks both for so comprehensive team work! 😄

andybaer:
@artembilan 

I addressed everything else so I will put my hands on the adoc

garyrussell:
LGTM; thanks both!

spring-integration-core/src/main/java/org/springframework/integration/transaction/IntegrationResourceHolderSynchronization.java
artembilan: Isn't addressed.

spring-integration-core/src/main/java/org/springframework/integration/transaction/PassthroughTransactionSynchronizationFactory.java
artembilan: DITTO.

I can do that on merge any way, but since we are expecting one more commit from you about docs, it would be better to have this there as well.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
```
artembilan: No, you have to leave the first year and fix the last to the current.
The first year is about the birth date, the last about maturity.

src/reference/asciidoc/transactions.adoc
```diff
@@ -164,6 +164,10 @@ In each case, the _#root_ variable is the original `Message`; in some cases, oth
 For example, the `MongoDbMessageSource` provides the _#mongoTemplate_ variable which references the message source's `MongoTemplate`; the `RedisStoreMessageSource` provides the _#store_ variable which references the `RedisStore` created by the poll.
 
 To enable the feature for a particular poller, you provide a reference to the `TransactionSynchronizationFactory` on the poller's <transactional/> element using the _synchronization-factory_ attribute.
+Spring Integration 5.0 introduces `PassthroughTransactionSynchronizationFactory` which is applied for pollers if there
+is no `TransactionSynchronizationFactory` available and the TX Advice is an instance of `TransactionInterceptor`.
+A `PassthroughTransactionSynchronizationFactory` can also be explicitly configured if you are using a a custom TX
```
artembilan: Double `a` and it really must be a separate paragraph.

Plus it would be helpful to say that now `AbstractPollingEndpoint` correctly throws `MessagingException` with the polled message as `failedMessage` property. And say that it is picked up from the `TransactionSynchronizationManager`.
Having a note about `PassthroughTransactionSynchronizationFactory` confuses me a bit now. 
The question is: what is the purpose of this factory used by default?

Let me know if I'm not clear yet and I'll come back to you with the polishing as I see for review and other comments.

andybaer: I'm pretty sure I understand but let's see...

artembilan (approved):

artembilan:
Awesome, @andybaer !

Please, review the polishing before I proceed with the merge: https://github.com/artembilan/spring-integration/commit/20b0888d2226059733609258113d4f8d4cae405b

garyrussell:
I made some comments on the polish.

Many thanks, once again @andybaer and @artembilan

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/3aa830b7c0c8cde3eba735cbd33563e180bf334d

@andybaer ,

thank you for your patience for so complex problem and contribution, of course.
Looking forward for more!

andybaer:
Thanks @artembilan and @garyrussell !!!

It's a great honour as well as free lessons from real professionals ;)

I'd be up for more, guess I will browse the Jira then?!

artembilan:
Yep! Sure thing! There is always something open to do 😄 

Plus feel free to come to us with any other ideas for improvements or just bugs.
Through the same JIRA, of course! 😉

PR #2081, opened 2017-03-06
INT-4219: Test Polishing

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4219

- Use `AdditionalAnswers.returnFirstArgument` instead of lambda to return first arg
- Enable CheckStyle `AvoidStaticImport` for tests
- Fix static import violations in tests


artembilan:
I can apply this PR, but would like to understand how it is related to the https://jira.spring.io/browse/INT-4219.
Thanks

garyrussell:
Well, it started out as a one-line change to the new test in the PR for that issue...

```java
 -		willAnswer(i -> i.getArgument(0)).given(bf).resolveEmbeddedValue(anyString());
 +		willAnswer(returnsFirstArg()).given(bf).resolveEmbeddedValue(anyString());
```

...when I discovered `AdditionalAnswers.returnsFirstArg()`.

However, I expected to get a CheckStyle failure for the new static import.

Then I discovered we were suppressing that CheckStyle for all tests.

Then I discovered I broke the rule by recently using a static import for `....regex.Matcher` (and a bunch of other - old - violations).

So... that's the history.

I am fine if you want to detach it from that JIRA during merge, if you feel strongly that it should be.

artembilan:
OK. Thank you for explanation!

Well, since it had been connected to JIRA I expected to see something to that test about compiled class name failure, because I'm still not sure why it passes here on Travis and there for me locally on Windows.
And I guess for you on Mac.

But then I see nothing related to JIRA and just curious what am I missing.

That is some side-effect of the confusing...

Sorry if I'm look very nit-pick 😄 

Merging...

PR #2082, opened 2017-03-06
INT-4086: (S)FTP: Add local file tree support

artembilan:
JIRA: https://jira.spring.io/browse/INT-4086

The remote files may be carried with sub-path, not just file name.
The same sub-dirs logic can be achieve with the `localFilenameGeneratorExpression`

* Add the logic into `AbstractInboundFileSynchronizer` to build sub-dirs for local files
* Switch on `WatchService` for the internal `FileReadingMessageSource` in the `AbstractInboundFileSynchronizingMessageSource` to react for all the changes in the `localDirectory`

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.file;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.springframework.integration.file.filters.AbstractFileListFilter;
+import org.springframework.integration.file.filters.FileListFilter;
+
+/**
+ * The {@link DefaultDirectoryScanner} extension which walks through the directory tree
+ * using {@link Files#walk}.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.0
+ *
+ * @see Files#walk
+ */
+public class RecursiveDirectoryScanner extends DefaultDirectoryScanner {
+
+	@Override
+	public List<File> listFiles(File directory) throws IllegalArgumentException {
+		FileListFilter<File> filter = getFilter();
+		boolean supportAcceptFilter = filter instanceof AbstractFileListFilter;
+		try {
+			Stream<File> fileStream = Files.walk(directory.toPath())
+					.skip(1)
+					.map(Path::toFile)
+					.filter(file -> !supportAcceptFilter
+							|| ((AbstractFileListFilter<File>) filter).accept(file));
+
+			if (supportAcceptFilter) {
+				return fileStream.collect(Collectors.toList());
+			}
+			else {
+				return filter.filterFiles(fileStream.toArray(File[]::new));
+			}
+		}
+		catch (IOException e) {
+			throw new IllegalArgumentException(e);
+		}
+	}
+
+}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
@@ -18,6 +18,7 @@ package org.springframework.integration.file;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
@@ -26,10 +27,13 @@ import java.util.stream.Stream;
 
 import org.springframework.integration.file.filters.AbstractFileListFilter;
 import org.springframework.integration.file.filters.FileListFilter;
+import org.springframework.util.Assert;
 
 /**
  * The {@link DefaultDirectoryScanner} extension which walks through the directory tree
- * using {@link Files#walk}.
+ * using {@link Files#walk(Path, int, FileVisitOption...)}.
+ * <p>
+ * By default this class visits all levels of the file tree without any {@link FileVisitOption}s.
  *
  * @author Artem Bilan
  *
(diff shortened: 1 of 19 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
```diff
@@ -144,6 +144,10 @@ public void afterPropertiesSet() throws Exception {
 					throw new FileNotFoundException(this.localDirectory.getName());
 				}
 			}
+			this.fileSource.setUseWatchService(true);
```
garyrussell: Should this be an option?

Watch events are slow on OS X - this means using the `delete()` to reset the filter might not be fast enough.

We would also need to keep the current documentation for people who can't rely on the watch service (manual removal from filter on rollback).

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/inbound/RollbackLocalFilterTests-context.xml
```diff
@@ -31,7 +30,7 @@
 	<bean id="acceptOnceFilter" class="org.springframework.integration.file.filters.AcceptOnceFileListFilter" />
 
 	<int:transaction-synchronization-factory id="syncFactory">
-		<int:after-rollback expression="@acceptOnceFilter.remove(payload)" />
+		<int:after-rollback expression="payload.delete()" />
```
garyrussell: Yes, this test fails on OS X even with a 60 second delay - I am not sure why the DELETE event is not received.

I am not sure if the watch service is reliable on any platform - we've introduced a race condition in that the filter removal is now done on a different thread - previously it was done on the poller thread.

In any case, how can this test work now anyway? Given we now populate the remote filter with a persistent accept once filter, the fact that we have deleted the file means the file is removed from the local filter, but it's still in the remote filter.

garyrussell (changes requested):
See comments.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
artembilan: So, let's reinstate our old `RecursiveLeafOnlyDirectoryScanner` and provide an option to the remote `AbstractInboundFileSynchronizingMessageSource` to let to inject `DirectoryScanner` impl?!

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/inbound/RollbackLocalFilterTests-context.xml
artembilan: The remote persistence file list filter is something new from the `master` now.
Didn't think about that yet 😄 

All other your concerns to this test unrelated.
That was a behavior to remove from the filter, now I want to check if file will come again with the watchService and when I delete it.

Since it isn't stable on all platforms we shouldn't rely on the `WatchService` by default.
Therefore this test must be reinstated with its original purpose. 
That's it.

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
```diff
@@ -0,0 +1,65 @@
...
+ * @see Files#walk
+ */
+public class RecursiveDirectoryScanner extends DefaultDirectoryScanner {
+
+	@Override
+	public List<File> listFiles(File directory) throws IllegalArgumentException {
+		FileListFilter<File> filter = getFilter();
+		boolean supportAcceptFilter = filter instanceof AbstractFileListFilter;
+		try {
+			Stream<File> fileStream = Files.walk(directory.toPath())
```
garyrussell: We should support depth and `FileVisitOption`s.

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
```diff
@@ -0,0 +1,65 @@
...
+					.skip(1)
+					.map(Path::toFile)
+					.filter(file -> !supportAcceptFilter
+							|| ((AbstractFileListFilter<File>) filter).accept(file));
+
```
garyrussell: We should perhaps not call the filter for directories; perhaps we should always filter them?

garyrussell (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
artembilan: Well, looks like we have introduced `AbstractDirectoryAwareFileListFilter` especially for this responsibility...

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
artembilan: OK. We can do that, but what should be by default?
You see, I use this in the `AbstractInboundFileSynchronizingMessageSource`

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
garyrussell: True, but it seems a bit weird filtering a directory, but not the files beneath it.

Consider this change to your test...

```
-               this.subSubLevelFile = new File(this.subSubFolder, "file2");
+               this.subSubLevelFile = new File(this.subSubFolder, "file3");
                this.subSubLevelFile.createNewFile();
        }
 
        @Test
-       public void shouldReturnAllFilesIncludingDirs() {
-               List<File> files = new RecursiveDirectoryScanner()
-                               .listFiles(this.recursivePath.getRoot());
+       public void shouldReturnAllFilesIncludingDirs() throws IOException {
+               RecursiveDirectoryScanner scanner = new RecursiveDirectoryScanner();
+               scanner.setFilter(new AcceptOnceFileListFilter<>());
+               List<File> files = scanner.listFiles(this.recursivePath.getRoot());
                assertEquals(5, files.size());
                assertThat(files, hasItem(this.topLevelFile));
                assertThat(files, hasItem(this.subLevelFile));
                assertThat(files, hasItem(this.subSubLevelFile));
                assertThat(files, hasItem(this.subFolder));
                assertThat(files, hasItem(this.subSubFolder));
+               File file = new File(this.subSubFolder, "file4");
+               file.createNewFile();
+               files = scanner.listFiles(this.recursivePath.getRoot());
+               assertEquals(1, files.size());
+               assertThat(files, hasItem(file));
```

It just seems a bit strange that we get the directories on the first pass but not the second.

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
garyrussell: The default should be what you have now...

    public static Stream<Path> walk(Path start, FileVisitOption... options) throws IOException {
        return walk(start, Integer.MAX_VALUE, options);
    }

i.e. max depth and no options 😄

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
garyrussell: Maybe another option `includeDirectories` - false by default - when `true` returns the directories as well. I suppose someone might want to examine the directory attributes (when using outside of framework classes).

artembilan: I still think that it is a filter responsibility. 

Maybe other option to skip directories from our out-of-the-box filters?

garyrussell: ok; leave as is.

garyrussell: That explains why the old scanner was `RecursiveLeafOnlyDirectoryScanner` 😄

artembilan:
@garyrussell ,
is there anything else not addressed do not let us to merge this?
Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/RecursiveDirectoryScanner.java
```diff
@@ -39,12 +43,33 @@
...
+		this.maxDepth = maxDepth;
+	}
+
+	/**
+	 * The options to configure the traversal.
+	 * @param fileVisitOptions options to configure the traversal
+	 */
+	public void setFileVisitOptions(FileVisitOption... fileVisitOptions) {
+		Assert.notNull(fileVisitOptions, "'fileVisitOptions' must not be null");
+		this.fileVisitOptions = fileVisitOptions;
```
garyrussell: `Arrays.copyOf()` - will fix on merge.

garyrussell (approved):

garyrussell:
Merged as 96ca9647a4f2b8bcd4bdfd47b8420652826f9486

garyrussell:
Merge as 96ca9647a4f2b8bcd4bdfd47b8420652826f9486 and 80d22cb46defceff61fe25d7bff4963880f9e2db

PR #2083, opened 2017-03-06
INT-4238: Detect Subscription Failures and QOS

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4238

Revert to using the sync client in the message-driven adapter so we can detect
subscription failures (the sync client throws an exception).

The only reason to use the async client was to timeout disconnects; this can
be achieved with the sync client and `disconnectForcibly`.

Also, the subscribe method updates the qos argument with the granted QOS values.

Detect and log if any QOS does not match the request.

__cherry-pick to 4.3.x, 4.2.x (change 4.2.x to use Java 8 for tests?)__

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ScheduledFuture;
 import org.eclipse.paho.client.mqttv3.IMqttClient;
 import org.eclipse.paho.client.mqttv3.IMqttDeliveryToken;
 import org.eclipse.paho.client.mqttv3.MqttCallback;
+import org.eclipse.paho.client.mqttv3.MqttClient;
 import org.eclipse.paho.client.mqttv3.MqttConnectOptions;
 import org.eclipse.paho.client.mqttv3.MqttException;
 import org.eclipse.paho.client.mqttv3.MqttMessage;
@@ -110,7 +111,7 @@ public class MqttPahoMessageDrivenChannelAdapter extends AbstractMqttMessageDriv
 	}
 
 	/**
-	 * Set the completion timeout for async operations. Not settable using the namespace.
+	 * Set the completion timeout for operations. Not settable using the namespace.
 	 * Default 30000 milliseconds.
 	 * @param completionTimeout The timeout.
 	 * @since 4.1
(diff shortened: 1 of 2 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -159,16 +159,14 @@ protected synchronized void doStop() {
...
+					this.client.unsubscribe(getTopic());
 				}
 			}
 			catch (MqttException e) {
 				logger.error("Exception while unsubscribing", e);
 			}
 			try {
-				this.client.disconnect()
-						.waitForCompletion(this.completionTimeout);
+				this.client.disconnectForcibly(this.completionTimeout);
```
artembilan: You don't propagate `completionTimeout` to the `MqttClient.setTimeToWait()`.
Is that intentional?
I think so, because of that your commit message explanation.
Any way just want to be sure that we are on the same page.
Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -230,31 +226,39 @@ private synchronized void connectAndSubscribe() throws MqttException {
...
-					.waitForCompletion(this.completionTimeout);
-			this.client.subscribe(getTopic(), getQos())
-					.waitForCompletion(this.completionTimeout);
+			this.client.connect(connectionOptions);
+			int[] requestedQos = getQos();
+			int[] grantedQos = Arrays.copyOf(requestedQos, requestedQos.length);
+			this.client.subscribe(topics, grantedQos);
+			for (int i = 0; i < requestedQos.length; i++) {
+				if (grantedQos[i] != requestedQos[i]) {
+					logger.info("Granted QOS different to Requested QOS; topics: " + Arrays.toString(topics)
```
artembilan: Maybe `warn()`?
Otherwise `if (logger.isInfoEnabled()) {`

artembilan (changes requested):
A couple concerns.

I'll try to switch to Java 8 in tests for `4.2.x`, although not sure that @wilkinsona doesn't have an IO CI plan against Java 7 on the matter...

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/d8dc9d427e85e56612278e1249fdf186c83805c1 and cherry-picked to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/789ed9d1b809ca82bbce08b12c307f8436ae1f6d 

Can't get it working for `4.2.x` yet... No one integration test works. 😢

artembilan:
Backported to `4.2.x` as https://github.com/spring-projects/spring-integration/commit/6550cda78962d9ca9ad89f95ce6ea9acb8239a5c

PR #2086, opened 2017-03-07
INT-4192: Improve ExprEvaluatingReleaseStrategy

artembilan:
JIRA: https://jira.spring.io/browse/INT-4192

To avoid implicit overhead for loading messages from persistent message store, use entire `MessageGroup` as root object for expression evaluation context

Also see https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.3-to-5.0-Migration-Guide


garyrussell:
Merged as d3fb8b8f9ef463e130986a07501ba69572ed4ca6 with minor doc polishing.

PR #2090, opened 2017-03-10
INT-4243: Upgrade to sshd 1.4

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4243

Remove newline char from key exchange.

__we could consider just backporting the newline stripping if updating the test server version will cause issues with IO__

### DIFF
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/RemoteFileTestSupport.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/RemoteFileTestSupport.java
@@ -19,6 +19,7 @@ package org.springframework.integration.file.remote;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.util.Arrays;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -171,6 +172,7 @@ public abstract class RemoteFileTestSupport {
 			File[] files = file.listFiles();
 			if (files != null) {
 				for (File fyle : files) {
+					logger.info("Deleting: " + fyle + " in " + testName.getMethodName());
 					if (fyle.isDirectory()) {
 						recursiveDelete(fyle);
 					}
@@ -181,7 +183,13 @@ public abstract class RemoteFileTestSupport {
 					}
 				}
 			}
-			file.delete();
+			logger.info("Deleting: " + file + " in " + testName.getMethodName());
+			if (!file.delete()) {
+				logger.error("Couldn't delete: " + file + " in " + testName.getMethodName());
+				if (file.isDirectory()) {
+					logger.error("Contents: " + Arrays.toString(file.listFiles()));
+				}
+			}
 		}
 	}
 
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/RemoteFileTestSupport.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/RemoteFileTestSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,9 +20,13 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.junit.Before;
 import org.junit.ClassRule;
+import org.junit.Rule;
 import org.junit.rules.TemporaryFolder;
+import org.junit.rules.TestName;
 
 /**
  * Abstract base class for tests requiring remote file servers, e.g. (S)FTP.
@@ -33,6 +37,8 @@ import org.junit.rules.TemporaryFolder;
  */
 public abstract class RemoteFileTestSupport {
 
+	protected final Log logger = LogFactory.getLog(getClass());
+
 	protected static int port;
 
 	@ClassRule
@@ -40,6 +46,9 @@ public abstract class RemoteFileTestSupport {
 
 	@ClassRule
 	public static final TemporaryFolder localTemporaryFolder = new TemporaryFolder();
+
+	@Rule
+	public final TestName testName = new TestName();
 
 	protected volatile File sourceRemoteDirectory;
 
@@ -166,7 +175,9 @@ public abstract class RemoteFileTestSupport {
 						recursiveDelete(fyle);
 					}
 					else {
-						fyle.delete();
+						if (!fyle.delete()) {
+							logger.error("Couldn't delete: " + fyle + " in " + testName.getMethodName());
+						}
 					}
 				}
 			}
--- a/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/SftpTestSupport.java
+++ b/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/SftpTestSupport.java
@@ -17,7 +17,6 @@
 package org.springframework.integration.sftp;
 
 import java.io.File;
-import java.nio.file.Paths;
 import java.util.Collections;
 
 import org.apache.sshd.common.NamedFactory;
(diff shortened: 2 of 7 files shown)


spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpServerTests.java
```diff
@@ -125,7 +127,12 @@ private void testKeyExchange(String pubKey, String privKey, String passphrase)
 
 	private PublicKey decodePublicKey(String key) throws Exception {
 		InputStream stream = new ClassPathResource(key).getInputStream();
-		byte[] decodeBuffer = Base64.decodeBase64(StreamUtils.copyToByteArray(stream));
+		byte[] keyBytes = StreamUtils.copyToByteArray(stream);
+		// strip any newline chars
+		while (keyBytes[keyBytes.length - 1] == 0x0a || keyBytes[keyBytes.length - 1] == 0x0c) {
```
artembilan: Any ideas why that failed sporadically?
Thanks

Merging BTW

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/SftpTestSupport.java
```diff
@@ -60,9 +62,10 @@ public static void createServer() throws Exception {
 		server = SshServer.setUpDefaultServer();
 		server.setPasswordAuthenticator((username, password, session) -> true);
 		server.setPort(0);
-		server.setKeyPairProvider(new SimpleGeneratorHostKeyProvider("hostkey.ser"));
-		server.setSubsystemFactories(Collections.<NamedFactory<Command>>singletonList(new SftpSubsystem.Factory()));
-		server.setFileSystemFactory(new VirtualFileSystemFactory(remoteTemporaryFolder.getRoot().getAbsolutePath()));
+		server.setKeyPairProvider(new SimpleGeneratorHostKeyProvider(new File("hostkey.ser")));
+		server.setSubsystemFactories(Collections.<NamedFactory<Command>>singletonList(new SftpSubsystemFactory()));
+		server.setFileSystemFactory(
+				new VirtualFileSystemFactory(Paths.get(remoteTemporaryFolder.getRoot().getAbsolutePath())));
```
artembilan: Why `remoteTemporaryFolder.getRoot().toPath()` doesn't work for you?

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpServerTests.java
```diff
@@ -125,7 +127,12 @@ private void testKeyExchange(String pubKey, String privKey, String passphrase)
...
+			keyBytes = Arrays.copyOf(keyBytes, keyBytes.length - 1);
+		}
+		byte[] decodeBuffer = Base64Utils.decode(keyBytes);
```
artembilan: Hm... Fails for me on Windows like this:
```
java.lang.IllegalArgumentException: Illegal base64 character d
	at java.util.Base64$Decoder.decode0(Base64.java:714)
	at java.util.Base64$Decoder.decode(Base64.java:526)
	at org.springframework.util.Base64Utils.decode(Base64Utils.java:63)
	at org.springframework.integration.sftp.session.SftpServerTests.decodePublicKey(SftpServerTests.java:135)
```

artembilan (changes requested):

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpServerTests.java
garyrussell: Oops, my mistake should be 0x0d not ox0c

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/SftpTestSupport.java
garyrussell: ok; fine

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpServerTests.java
garyrussell: See my comment on the JIRA - problem is unrelated but we should do the upgrade anyway.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/dsl/SftpTests.java
```diff
@@ -24,11 +24,13 @@
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.InputStream;
 import java.util.List;
 import java.util.regex.Matcher;
 
+import org.apache.log4j.lf5.util.StreamUtils;
```
artembilan: ??? Let me guess you were going to use `org.springframework.util.StreamUtils`

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/dsl/SftpTests.java
```diff
@@ -117,12 +119,14 @@ public void testSftpInboundStreamFlow() throws Exception {
 		assertNotNull(message);
 		assertThat(message.getPayload(), instanceOf(InputStream.class));
 		assertThat(message.getHeaders().get(FileHeaders.REMOTE_FILE), isOneOf(" sftpSource1.txt", "sftpSource2.txt"));
+		StreamUtils.copy((InputStream) message.getPayload(), new ByteArrayOutputStream());
```
artembilan: Well, my solution is actually like this:
```
((InputStream) message.getPayload()).close();
```
Because the problem with the file on Windows that we have an open stream to it. And, yeah..., exactly payload is `InputStream` in case of streaming inbound channel adapter.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/RemoteFileTestSupport.java
```diff
@@ -162,15 +172,24 @@ public void recursiveDelete(File file) {
...
 					else {
-						fyle.delete();
+						if (!fyle.delete()) {
+							logger.error("Couldn't delete: " + fyle + " in " + testName.getMethodName());
+						}
 					}
 				}
 			}
-			file.delete();
+			logger.info("Deleting: " + file + " in " + testName.getMethodName());
```
artembilan: I think we have figure out the problem with files and we don't need these diagnostics any more.
Let me know and I'll polish on merge respectively.

Thank you for the fix any way!

garyrussell: I figured it won't hurt leaving these at INFO level - in case we have problems in future.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/dsl/SftpTests.java
garyrussell: ok; feel free to fix that way.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/dsl/SftpTests.java
garyrussell: hehe

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/dsl/SftpTests.java
garyrussell: Interestingly, windows did not say that the file deletion failed, just silently left the file in place, preventing the deletion of the directory.

artembilan: Well, actually it does, if we check the result of the `File.delete()`.
And that is exactly what we are doing now with all those your new log diagnostics.

OK, merging having your agreement...
I also will increase the timeout in that Redis test

garyrussell: Right that's what I am saying; on my Windows VM, the file.delete() was successful (I didn't get the "couldn't delete" message - that's why I added the additional "Contents" log when the directory delete fails.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4bb3f5041f0a84cd736399fc1a0da96b7aeab742 and cherry-picked to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/96331d5415dd4b6b04bd5bd683884d57aa5848fe

artembilan:
Back-ported to `4.2.x` as https://github.com/spring-projects/spring-integration/commit/8036a4bab14082a56b6e0b24a195165a15659140

PR #2091, opened 2017-03-10
INT-4257: Introduce ErrorMessagePublishingRecoveryCallback

artembilan:
JIRA: https://jira.spring.io/browse/INT-4257

To make a target `ErrorMessage` customizable introduce
`ErrorMessagePublishingRecoveryCallback` and
`ErrorMessageStrategy` to inject

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -173,12 +173,22 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 		catch (RuntimeException e) {
 			MessageChannel errorChannel = getErrorChannel();
 			if (errorChannel != null) {
-				this.messagingTemplate.send(errorChannel, new ErrorMessage(e));
+				this.messagingTemplate.send(errorChannel, buildErrorMessage(e));
 			}
-			else  {
+			else {
 				throw e;
 			}
 		}
+	}
+
+	/**
+	 * Build an {@link ErrorMessage} for the exception.
+	 * @param exception the exception.
+	 * @return the error message.
+	 * @since 4.3.9
+	 */
+	protected ErrorMessage buildErrorMessage(RuntimeException exception) {
+		return new ErrorMessage(exception);
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -20,11 +20,13 @@ import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.core.MessagingTemplate;
 import org.springframework.integration.history.MessageHistory;
+import org.springframework.integration.support.DefaultErrorMessageStrategy;
+import org.springframework.integration.support.ErrorMessageStrategy;
+import org.springframework.integration.support.ErrorMessageUtils;
 import org.springframework.integration.support.management.TrackableComponent;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessagingException;
-import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
@@ -40,6 +42,8 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 		SmartInitializingSingleton {
 
 	private final MessagingTemplate messagingTemplate = new MessagingTemplate();
+
+	private ErrorMessageStrategy errorMessageStrategy = new DefaultErrorMessageStrategy();
 
 	private volatile MessageChannel outputChannel;
 
@@ -127,6 +131,17 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 		this.shouldTrack = shouldTrack;
 	}
 
+	/**
+	 * Set an {@link ErrorMessageStrategy} to use to build an error message when a exception occurs.
+	 * Default is the {@link DefaultErrorMessageStrategy}.
+	 * @param errorMessageStrategy the {@link ErrorMessageStrategy}.
+	 * @since 4.3.10
+	 */
+	public void setErrorMessageStrategy(ErrorMessageStrategy errorMessageStrategy) {
+		Assert.notNull(errorMessageStrategy, "'errorMessageStrategy' cannot be null");
+		this.errorMessageStrategy = errorMessageStrategy;
+	}
+
 	protected MessagingTemplate getMessagingTemplate() {
 		return this.messagingTemplate;
 	}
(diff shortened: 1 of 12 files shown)


garyrussell:
Suggested polish here: https://github.com/garyrussell/spring-integration/commit/f1929f590d08401c4ddb7ec381b1f12e5b7f440a

garyrussell:
More polish - add `EnhancedErrorMessage` https://github.com/garyrussell/spring-integration/commit/63cb9b84945a04bca0c1eba004aa01452ea792b1

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ErrorMessageSendingRecoverer.java
```diff
@@ -40,46 +32,37 @@
...
+	public ErrorMessageSendingRecoverer() {
+		this(null);
+	}
 
 	public ErrorMessageSendingRecoverer(MessageChannel channel) {
-		Assert.notNull(channel, "channel cannot be null");
-		this.messagingTemplate.setDefaultDestination(channel);
+		setRecoveryChannel(channel);
+		setErrorMessageStrategy(new RecovererErrorMessageStrategy());
 	}
```
garyrussell: Needs a 3rd ctor so we can pass in a strategy.

spring-integration-core/src/main/java/org/springframework/integration/support/ErrorMessagePublishingRecoveryCallback.java
```diff
@@ -0,0 +1,134 @@
...
+ * after retry exhaustion.
+ * <p>
+ * A {@link ErrorMessageStrategy} can be used to provide customization for the
+ * target {@link ErrorMessage} based on the {@link RetryContext}.
+ *
+ * @author Artem Bilan
+ *
+ * @since 4.3.9
+ */
+public class ErrorMessagePublishingRecoveryCallback implements RecoveryCallback<Object>, BeanFactoryAware {
```
garyrussell: `s/Object/Void/`

garyrussell:
Final Polishing: https://github.com/garyrussell/spring-integration/commit/ffff4c229ea19cc810a5388c44ecb3191ff19243

spring-integration-core/src/main/java/org/springframework/integration/message/EnhancedErrorMessage.java
```diff
@@ -0,0 +1,61 @@
...
+ * stack frame where the error message is generated. Typically this will be a
+ * message that begins a subflow, whereas if the {@link Throwable} payload is
+ * a {@link org.springframework.messaging.MessagingException}, it's failedMessage
+ * property will contain the message at the point where the exception occurred.
+ *
+ * @author Gary Russell
+ * @since 4.3.9
+ *
+ */
+public class EnhancedErrorMessage extends ErrorMessage {
```
garyrussell: I don't like it either; I started with `Composite` but it's not a composite.

IAEM is ok, I guess, but I'm not thrilled with that either.

artembilan: OK. Since we can't come up with something what sounds better and still provides the proper context, I won't mind to live with this name.

@mbogoevici , maybe you have some ideas? Thanks

garyrussell:
Decouple `ErrorMessagePublisher` from `spring-retry`... https://github.com/garyrussell/spring-integration/commit/2667b739b007a785dea22eda7271d1200579cb3b

artembilan:
> Decouple `ErrorMessagePublisher`

OK. Will come back to you with some polishing soon.

Thank you for looking forward to this!
I really is distracted with other stuff to think more here 😢

artembilan:
Pushed some further refactoring and polishing.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -173,12 +173,22 @@ protected void sendMessage(Message<?> message) {
 		catch (RuntimeException e) {
 			MessageChannel errorChannel = getErrorChannel();
 			if (errorChannel != null) {
-				this.messagingTemplate.send(errorChannel, new ErrorMessage(e));
+				this.messagingTemplate.send(errorChannel, buildErrorMessage(e));
```
mbogoevici: Wouldn't it be preferable to delegate this to an injected `ErrorMessageStrategy` rather than invoking the overridable method below ?

artembilan: The question relates to the change here: https://github.com/spring-projects/spring-integration-kafka/pull/163/files#diff-b49c842da794eb3622242c8acb03e1c1R290

I don't know the story yet, but let @garyrussell to answer!

garyrussell: Yes, that is my plan; I meant to add that change in my latest polishing commit.

artembilan: I got it! Let me fix this concern 😄

garyrussell: Oops - here's my version: https://github.com/garyrussell/spring-integration/commit/6a4ed7604a6570e3981779d0381df533e5961254

artembilan: Ha-ha! OK. I give up on this stuff.
Is there some work which you really can't do and I am still would be useful?  😉

garyrussell: Sorry - I have sort-of hijacked your PR ☹️

artembilan: Well, that doesn't look so bad (hijacking) because my mind doesn't go somehow so far 😢

spring-integration-core/src/main/java/org/springframework/integration/message/EnhancedErrorMessage.java
mbogoevici: One possible idea (for better or worse) - `RequestAwareErrorMessage` and rename `inputMessage` to `requestMessage` ?

artembilan:
Applied your changes Gary.
Awesome work as usual!
I have just added JavaDocs to that `DefaultErrorMessageStrategy`.

My idea meanwhile was to use that new `ErrorMessagePublisher` as an internal instance for the provided `ErrorMessageStrategy` and don't expose the `DefaultErrorMessageStrategy` to the top level class.
But your approach is fine as well.
Anyway we have to keep the behavior like re-throw exception when the `errorChannel` isn't provided.

@mbogoevici , I think we can live with that `EnhancedErrorMessage` class until `5.0`. The JIRA https://jira.spring.io/browse/SPR-15459 has been just fixed and `ErrorMessage` will have `originalMessage` property since there.

I'm adding a deprecation note there though...

garyrussell:
Perhaps for 5.0 only, we should consider reworking the MPEH to use this strategy from pollers and executors too.

artembilan:
> we should consider reworking the MPEH to use this strategy

Let's do that in separate PRs, @garyrussell !

garyrussell:
agreed

spring-integration-core/src/main/java/org/springframework/integration/support/ErrorMessageUtils.java
```diff
@@ -0,0 +1,72 @@
...
+			attributes.setAttribute(INPUT_MESSAGE_CONTEXT_KEY, inputMessage);
+		}
+		if (failedMessage != null) {
+			attributes.setAttribute(FAILED_MESSAGE_CONTEXT_KEY, failedMessage);
+		}
+		return attributes;
+	}
+
+	@SuppressWarnings("serial")
+	private static class ErrorMessageAttributes extends AttributeAccessorSupport {
```
garyrussell: Do you mind if I make this public? I need it in S-I-K. Or, we could make it a top level class.

garyrussell: Actually, no; forget that.

garyrussell:
Minor polish needed for S-I-K: https://github.com/garyrussell/spring-integration/commit/1797e95dc4aceff2256128d11044fde12899309e

artembilan:
The latest polishing is fine with me.
Let me know if there is need to be done with this PR yet!

garyrussell:
If you are happy; I will merge and cherry-pick, and we can move over to S-I-K 😄

artembilan:
Ship it!
We always can add subsequent commits if that 😄

garyrussell:
Merged as 28058884f4c9ec1c99e5d89f7360d08ce0d1e41e cherry-picked as f7adfc06286fc5171592839fd78dae81a030e248

PR #2092, opened 2017-03-16
INT-4244: Optimize JpaOutboundGateway for message

artembilan:
JIRA: https://jira.spring.io/browse/INT-4244

The `JpaOutboundGateway` returns a whole `Message` for reply copying request headers.
Building the reply a `AbstractMessageProducingHandler` uses `MessageBuilder` to create a new `Message` with all request headers because of `shouldCopyRequestHeaders() == true`

* To avoid some overhead with new `Message` and `MessageBuilder` for nothing, just return a JPA result as is and let super class to build the Message

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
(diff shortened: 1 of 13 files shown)


artembilan:
Looks like we have many other gateways with similar problem. For example `AbstractAmqpOutboundEndpoint`, `AbstractRemoteFileOutboundGateway` and others.

If you are fine with the fix, I can go ahead and fix those here as well.
I even won't mind do not back port the fix since it doesn't look so critical.

garyrussell:
Looks fine to me - I was going to merge but do you want to add the others to this PR?

garyrussell:
(I agree about not backporting)

artembilan:
I will add others.
Thanks

garyrussell:
Merged as df9593a528b0317e46c8c9081f4719c38e1437ba

PR #2094, opened 2017-03-17
JavaConfig/DSL Docs for Routers (Part 1)

garyrussell:
Also fix DSL messing `sendTimeout` property for routers.
Since other message handlers have this property on the `ConsumerEndpointSpec`, add support there
instead of `RouterSpec`, for consistency.
Also fix `ConsumerEndpointSpec` `sendTimeout` - can be applied to any
`AbstractMessageProducingHandler`, not just `AbstractReplyProducingMessageHandler`.
Ditto for `async`.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/ConsumerEndpointSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/ConsumerEndpointSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,7 +24,9 @@ import org.aopalliance.aop.Advice;
 
 import org.springframework.integration.config.ConsumerEndpointFactoryBean;
 import org.springframework.integration.handler.AbstractMessageHandler;
+import org.springframework.integration.handler.AbstractMessageProducingHandler;
 import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.integration.router.AbstractMessageRouter;
 import org.springframework.integration.scheduling.PollerMetadata;
 import org.springframework.integration.transaction.TransactionInterceptorBuilder;
 import org.springframework.messaging.MessageHandler;
@@ -39,6 +41,7 @@ import org.springframework.transaction.interceptor.TransactionInterceptor;
  * @param <H> the target {@link MessageHandler} implementation type.
  *
  * @author Artem Bilan
+ * @author Gary Russell
  *
  * @since 5.0
  */
@@ -170,15 +173,19 @@ public abstract class ConsumerEndpointSpec<S extends ConsumerEndpointSpec<S, H>,
 	/**
 	 * @param sendTimeout the send timeout.
 	 * @return the endpoint spec.
-	 * @see AbstractReplyProducingMessageHandler#setSendTimeout(long)
+	 * @see AbstractMessageProducingHandler#setSendTimeout(long)
 	 */
 	public S sendTimeout(long sendTimeout) {
 		assertHandler();
-		if (this.handler instanceof AbstractReplyProducingMessageHandler) {
-			((AbstractReplyProducingMessageHandler) this.handler).setSendTimeout(sendTimeout);
+		if (this.handler instanceof AbstractMessageProducingHandler) {
+			((AbstractMessageProducingHandler) this.handler).setSendTimeout(sendTimeout);
+		}
+		else if (this.handler instanceof AbstractMessageRouter) {
+			// This should probably go on the RouterSpec, but we put it here for consistency
+			((AbstractMessageRouter) this.handler).setSendTimeout(sendTimeout);
 		}
 		else {
-			logger.warn("'sendTimeout' can be applied only for AbstractReplyProducingMessageHandler");
+			logger.warn("'sendTimeout' can be applied only for AbstractMessageProducingHandler");
 		}
 		return _this();
 	}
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/ConsumerEndpointSpec.java
```diff
@@ -170,15 +173,19 @@ public S requiresReply(boolean requiresReply) {
...
 	 */
 	public S sendTimeout(long sendTimeout) {
 		assertHandler();
-		if (this.handler instanceof AbstractReplyProducingMessageHandler) {
-			((AbstractReplyProducingMessageHandler) this.handler).setSendTimeout(sendTimeout);
+		if (this.handler instanceof AbstractMessageProducingHandler) {
+			((AbstractMessageProducingHandler) this.handler).setSendTimeout(sendTimeout);
+		}
+		else if (this.handler instanceof AbstractMessageRouter) {
+			// This should probably go on the RouterSpec, but we put it here for consistency
```
artembilan: I see your concern for discussion, but IMO you have done the proper decision.
This option must be here (it is here already anyway) just because it really will be a mess to have it on the `RouterSpec` as well:
```
.route("true", r -> r.sendTimeout(100), e -> e.sendTimeout(100))
```

src/reference/asciidoc/router.adoc
```diff
@@ -589,6 +589,35 @@ The example below demonstrates a `PayloadTypeRouter` configuration which is equi
...
+----
+
+and using the Java DSL:
+
+[source, java]
+----
+@Bean
+public IntegrationFlow routerFlow() {
+    return IntegrationFlows.from("routingChannel")
+            .route(new PayloadTypeRouter(), m -> m
```
artembilan: There is nice solution for raw Java on the matter:
```
.<Object, Class<?>>route(Object::getClass, m -> m
                            .channelMapping(String.class, "stringsChannel")
                            .channelMapping(Integer.class, "integersChannel"))
```
On the other hand your sample doesn't work.
Because of another method selection:
wrong_route

src/reference/asciidoc/router.adoc
```diff
@@ -627,6 +656,35 @@ However, in cases where the header value is mapped to a channel name but the cha
...
+----
+
+and using the Java DSL:
+
+[source, java]
+----
+@Bean
+public IntegrationFlow routerFlow() {
+    return IntegrationFlows.from("routingChannel")
+            .route(new HeaderValueRouter("testHeader"), m -> m
```
artembilan: This must be like this:
```
.<Message<?>, String>route(m -> m.getHeaders().get("testHeader", String.class),
```
And that `channelMapping()` will suffer from the generic argument.

artembilan (changes requested):
Overall it is good.
Just a couple concerns in those samples.

If you agree with me, I'll fix one merge.

spring-integration-core/src/main/java/org/springframework/integration/dsl/ConsumerEndpointSpec.java
garyrussell: Yes, it's the right decision, my question was really why we decided to put it on the consumer spec in the first place, given that it's a handler property. But, I agree, it's too late to change now.

src/reference/asciidoc/router.adoc
garyrussell: Yes, I saw that in the test but we'll have to explain, somehow, why we don't use the PTR with the DSL (I did have to add the c.id though). Perhaps we should show both options?

src/reference/asciidoc/router.adoc
garyrussell: Again, we would have to explain why we don't use the HVR here.

spring-integration-core/src/main/java/org/springframework/integration/dsl/ConsumerEndpointSpec.java
artembilan: Well, my point was about simple configurations based on Lambdas and expressions or method invocation in most cases. All those options (`sendTimeout`, `async`, `requestHandlerAdviceChain` etc.) are really optional and, on the other hand, they are the same for any `MessageHandler` impl. So, I decided to keep them close to the `endpoint` even if we allow to configure `MH` directly, like in your `route(new PayloadTypeRouter())` sample.

Although with that your sample we really come back to the duplicated and confused options issue. Since we can configure `sendTimeout()` on the `PayloadTypeRouter()` directly as well as on the `EndpointSpec`.

There is a solution like merge `RouterSpec` with the `ConsumerEndpointSpec`, like we did for the `BarrierSpec`, `CorrelationHandlerSpec`. That will eat a couple overloaded methods and should kill this our discussion.
I think the `ConsumerEndpointSpec` for the `AbstractMessageRouter` variant  should still be there any way.

src/reference/asciidoc/router.adoc
artembilan: Yes, I think show both options would be great.
That will be some kind of sample which end users are always seeking.
So, let's move on with the `new PayloadTypeRouter()` (but proper), and regular, based on the Lambda and generics resolution during typing.

src/reference/asciidoc/router.adoc
artembilan: We can use, but that is different method based on the `AbstractMessageRouter` argument.

We provide choices for preferences:
1. To use fluent method chain API, which let you type flow quickly and straightforward.
2. To declare separate definition of the canonical component via its setters and then lightweight flow.

artembilan (approved):

artembilan:
I think the goal of this PR has been reached.
Merging...
Meanwhile let me know if we really go they way to reconsider that `.route()` configuration mess just in a favor of absorbing `ConsumerEndpointSpec` via `AbstractRouterSpec`.

garyrussell:
>Meanwhile let me know if we really go they way to reconsider that .route() configuration mess just in a favor of absorbing ConsumerEndpointSpec via AbstractRouterSpec.

Do you mean prohibiting `.channelMapping()` when an external `router()`  is provided?

I guess that would make sense, but it really doesn't hurt anything if they do that.

If you mean something else; please elaborate.

artembilan:
> Do you mean prohibiting .channelMapping() when an external router() is provided?

Yes, that one as well.
And with that combination we really will nail your original concern how to be with the `sendTimeout()`.
Having only one `RouterSpec` there will eliminate the confusion.
Plus in case of `AbstractMessageRouter` we won't provide `RouterSpec`, just only `ConsumerEndpointSpec`. Also the issue with `setSendTimeoute()` and its `sendTimeout()` will be still there. I guess similar concern we have with many other properties...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/70652739b862c443be2d7ba963c022b335af90fc

garyrussell:
I am not overly concerned about it; it would be disruptive to existing apps to change it, but we do have only one opportunity for such refactoring.

artembilan:
I think I'll give it a shot if you don't mind.
That should not take a long time, but result might be more robust with less method choice in the `IntegrationFlowDefinition`.

garyrussell:
👍

garyrussell:
Ah, sorry - I didn't get a notification about your comment for some reason.

>Well, my point ...

Yes; all that makes sense 😄

PR #2096, opened 2017-03-21
Java DSL polishing around `HeaderEnricherSpec`

artembilan:
* Make `HeaderEnricherSpec extends ConsumerEndpointSpec` to avoid extra
EIP-method with two `Consumer`s, when only one for
the `HeaderEnricherSpec` can address all the options - header enricher,
as well as target endpoint
* Move `this.endpointFactoryBean.setHandler(this.handler)` into
the `ConsumerEndpointSpec#doGet()` instead of ctor to avoid
duplicate code from the target implementations (e.g. `BarrierSpec`)


artembilan:
Also see: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.3-to-5.0-Migration-Guide#java-dsl-breaking-changes

PR #2097, opened 2017-03-21
Support for sub flows in defining content enricher

chiknrice:
The current enricher allows for specifying a payload, request and response channel where one can request additional data for enriching content or header. This change would allow one to specify an inline call to a handler instead of a request/response channel to grab the additional data.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/EnricherSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/EnricherSpec.java
@@ -142,14 +142,14 @@ public class EnricherSpec extends ConsumerEndpointSpec<EnricherSpec, ContentEnri
 	 * @param subFlow the subFlowDefinition
 	 * @return the enricher spec
 	 */
-	public EnricherSpec subFlow(IntegrationFlow subFlow) {
+	public EnricherSpec requestSubFlow(IntegrationFlow subFlow) {
 		Assert.notNull(subFlow, "'subFlow' must not be null");
 
 		DirectChannel requestChannel = new DirectChannel();
 		IntegrationFlowBuilder flowBuilder = IntegrationFlows.from(requestChannel);
 		subFlow.configure(flowBuilder);
 
-		this.componentsToRegister.add(flowBuilder);
+		this.componentsToRegister.add(flowBuilder.get());
 
 		return requestChannel(requestChannel);
 	}
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/EnricherSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/EnricherSpec.java
@@ -41,6 +41,7 @@ import reactor.util.function.Tuple2;
  *
  * @author Artem Bilan
  * @author Tim Ysewyn
+ * @author Ian Bondoc
  *
  * @since 5.0
  */
@@ -150,7 +151,7 @@ public class EnricherSpec extends ConsumerEndpointSpec<EnricherSpec, ContentEnri
 
 		this.componentsToRegister.add(flowBuilder);
 
-		return _this().requestChannel(requestChannel);
+		return requestChannel(requestChannel);
 	}
 
 	/**
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/EnricherSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/EnricherSpec.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.function.Function;
 
 import org.springframework.expression.Expression;
+import org.springframework.integration.channel.DirectChannel;
 import org.springframework.integration.config.ConsumerEndpointFactoryBean;
 import org.springframework.integration.expression.FunctionExpression;
 import org.springframework.integration.expression.ValueExpression;
@@ -134,6 +135,22 @@ public class EnricherSpec extends ConsumerEndpointSpec<EnricherSpec, ContentEnri
 	public <P> EnricherSpec requestPayload(Function<Message<P>, ?> requestPayloadFunction) {
 		this.handler.setRequestPayloadExpression(new FunctionExpression<>(requestPayloadFunction));
 		return _this();
+	}
+
+	/**
+	 * @param subFlow the subFlowDefinition
+	 * @return the enricher spec
+	 */
+	public EnricherSpec subFlow(IntegrationFlow subFlow) {
+		Assert.notNull(subFlow, "'subFlow' must not be null");
+
+		DirectChannel requestChannel = new DirectChannel();
+		IntegrationFlowBuilder flowBuilder = IntegrationFlows.from(requestChannel);
+		subFlow.configure(flowBuilder);
+
+		this.componentsToRegister.add(flowBuilder);
+
+		return _this().requestChannel(requestChannel);
 	}
 
 	/**
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
@@ -108,7 +108,6 @@ import reactor.util.function.Tuple2;
  * @author Artem Bilan
  * @author Gary Russell
  * @author Gabriele Del Prete
- * @author Ian Bondoc
  *
  * @since 5.0
  *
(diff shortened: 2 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
```diff
@@ -1172,7 +1172,28 @@ public B delay(String groupId, Consumer<DelayerEndpointSpec> endpointConfigurer)
...
+		if (!CollectionUtils.isEmpty(componentsToRegister)) {
+			for (Object component : componentsToRegister) {
+				// Currently EnricherSpec only has IntegrationFlowBuilder as a sub component
+				if (component instanceof IntegrationFlowDefinition) {
+					IntegrationFlowDefinition<?> flowBuilder = (IntegrationFlowDefinition<?>) component;
+					if (flowBuilder.isOutputChannelRequired()) {
+						addComponent(flowBuilder.get());
+					}
+					else {
+						throw new BeanCreationException("Enricher subFlow must require an output channel");
```
artembilan: To be honest I don't see reason in this restriction.
There is no guarantee that flow ends with just a reply producer. It may sends to the channel and a flow behind it may return the result to the `replyChannel`.

Yes, that is the point to wait for the reply from the downstream flow for the `enricher`, but similar concern with unpredictable sub-flow behavior we have even with an existing XML definition.
So, my point to relax this assertion and really just configure `requestChannel` as an input to the that sub-flow and 
```
this.componentsToRegister.add(flowBuilder.get());
```
in the `EnricherSpec`.

That way this `IntegrationFlowDefinition` won't be affected at all.

The problem with "no reply from sub-flow" remains to the end-user. 
No choice!

spring-integration-core/src/main/java/org/springframework/integration/dsl/EnricherSpec.java
```diff
@@ -137,6 +138,22 @@ public EnricherSpec requestPayloadExpression(String requestPayloadExpression) {
...
+	public EnricherSpec subFlow(IntegrationFlow subFlow) {
+		Assert.notNull(subFlow, "'subFlow' must not be null");
+
+		DirectChannel requestChannel = new DirectChannel();
+		IntegrationFlowBuilder flowBuilder = IntegrationFlows.from(requestChannel);
+		subFlow.configure(flowBuilder);
+
+		this.componentsToRegister.add(flowBuilder);
+
+		return _this().requestChannel(requestChannel);
```
artembilan: We don't need to call `_this()` here

artembilan (changes requested):
@chiknrice ,

See comments bellow.
Only several simple concerns.
No reason to comment if you agree with me.
Just push the fix and that's all.
I'll merge it!

One more comment: please, add your name to the `@author` of affected classes.

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
chiknrice: Agreed, I didn't thought of that :+1: also added a test of a terminating sub-flow

spring-integration-core/src/test/java/org/springframework/integration/dsl/transformers/TransformerTests.java
```diff
@@ -287,6 +342,71 @@ public IdempotentReceiverInterceptor idempotentReceiverInterceptor() {
...
+							.subFlow(sf -> sf
+									.<String>handle((p, h) -> someServiceGateway.deriveValueSync(p)))
+							.<String>headerFunction("foo", m -> m.getPayload())
+							.propertyFunction("name", m -> m.getPayload()))
+					.channel("subFlowTestReplyChannel");
+		}
+
+		@Bean
+		@Autowired
+		public IntegrationFlow asyncSubFlowEnricher(SomeServiceGateway someServiceGateway) {
```
artembilan: Not sure why do you worry about `async` behavior in the sub-flow, since `ContentEnricher` uses blocking `sendAndReceiveMessage()`. So, it doesn't matter how the reply is going to appear. But yeah, it won't hurt to have that test as well. 
M-m-m. I believe that `@Autowired` is redundant on the `@Bean` method. Any its arguments are really considered for autowiring.

spring-integration-core/src/test/java/org/springframework/integration/dsl/transformers/TransformerTests.java
```diff
@@ -287,6 +342,71 @@ public IdempotentReceiverInterceptor idempotentReceiverInterceptor() {
...
+		}
+
+		@Bean
+		public SomeService someService() {
+			return new SomeService();
+		}
+
+		@Bean
+		public AsyncTaskExecutor exec() {
+			SimpleAsyncTaskExecutor simpleAsyncTaskExecutor = new SimpleAsyncTaskExecutor();
```
artembilan: Let's remove this bean altogether!
The `GatewayProxyFactoryBean` uses its instance by default anyway:
```
private volatile AsyncTaskExecutor asyncExecutor = new SimpleAsyncTaskExecutor();
```
This test isn't about Messaging Gateway, so, would be better to make it as simple as possible.

spring-integration-core/src/test/java/org/springframework/integration/dsl/transformers/TransformerTests.java
```diff
@@ -287,6 +342,71 @@ public IdempotentReceiverInterceptor idempotentReceiverInterceptor() {
...
+							.subFlow(sf -> sf
+									.<String>handle(m -> someService.aTerminatingServiceMethod(m)))
+							.replyChannel("enricherReplyChannel")
+							.<String>headerFunction("foo", m -> m.getPayload())
+							.propertyFunction("name", m -> m.getPayload()))
+					.channel("subFlowTestReplyChannel");
+		}
+
+		@Bean
+		public GatewayProxyFactoryBean someServiceGateway() {
```
artembilan: I think you are aware of the `@MessagingGateway`  and `@IntegrationComponentScan`.
Maybe we really don't need this bean definition?

spring-integration-core/src/test/java/org/springframework/integration/dsl/transformers/TransformerTests.java
```diff
@@ -287,6 +342,71 @@ public IdempotentReceiverInterceptor idempotentReceiverInterceptor() {
...
+			GatewayProxyFactoryBean gateway = new GatewayProxyFactoryBean(SomeServiceGateway.class);
+			gateway.setAsyncExecutor(exec());
+			return gateway;
+		}
+
+		@Bean
+		public SomeService someService() {
```
artembilan: I think this `SomeService` can be marked with the `@Service` as well.
Yes, `@ComponentScan` must be there on the `@Configuration`.

That feels simpler.
Just IMO any way.
We can live with the current state of the test.
Looks like it demonstrates the premise of the feature and that is enough.

artembilan (changes requested):
@chiknrice ,

thank you for the update!

Please, meet the next review round.
I don't see other issues any more and after your next push this might be merged.

spring-integration-core/src/test/java/org/springframework/integration/dsl/transformers/TransformerTests.java
chiknrice: I initially had the requirement of enriching a message with information coming from a web resource and was thinking it was a good idea to make it an async call.  I realise now it doesn't makes sense since the whole enrich call is blocking - maybe i'm approaching the problem incorrectly.  Anyway, will still submit this change as it could be helpful to some.  I've applied the necessary changes to the test related to this.

spring-integration-core/src/test/java/org/springframework/integration/dsl/transformers/TransformerTests.java
chiknrice: I removed the beans relating to flow with async handler and I think the test was designed to not do component scanning.

spring-integration-core/src/test/java/org/springframework/integration/dsl/transformers/TransformerTests.java
chiknrice: Done.

artembilan (approved):
LGTM
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/6616242831a9634fd8f9e260ccc448406e98522b.

Looking forward for more!

PR #2098, opened 2017-03-22
INT-4246: Revoke leadership when `LockRegistryLeaderInitiator` is unable to acquire lock

vpavic:
JIRA: https://jira.spring.io/browse/INT-4246

The `LockRegistryLeaderInitiator` currently does not revoke leadership when leading instance is unable to acquire lock from the underlying lock. This can result in multiple `LockRegistryLeaderInitiator` instances becoming leaders in the exceptional situations such as lock timeouts.

* handle leadership revoking when leading `LockRegistryLeaderInitiator` is unable to acquire lock

I've added a test that simulates described scenario using mocks. The test fails with current `4.3.x` branch on two assertions.

Few other notes:

- I've ran the test to verify the proposed solution successfully from IntelliJ, however while building the project locally it complained about illegal Mockito usage - looking forward to see how it fares out on Travis CI
- I've also made some minor changes to `LockRegistryLeaderInitiator` to remove code duplication and improve readability.
- this PR also includes separate commit to polish javadoc from my previous contribution in #1991 (hope this is OK, I prefer to use ASCII chars only for my name in author tag :smile:)

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2016 the original author or authors.
+ * Copyright 2012-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,9 +18,16 @@ package org.springframework.integration.support.leader;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.mock;
 
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
 
 import org.apache.log4j.Level;
 import org.junit.Before;
@@ -37,6 +44,7 @@ import org.springframework.integration.test.rule.Log4jLevelAdjuster;
 /**
  * @author Dave Syer
  * @author Artem Bilan
+ * @author Vedran Pavic
  *
  * @since 4.3.1
  */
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
```diff
@@ -100,6 +107,58 @@ public void competing() throws Exception {
...
+		given(firstLock.tryLock(anyLong(), any(TimeUnit.class))).willReturn(true);
+
+		// set up first initiator instance using first LockRegistry
+		LockRegistryLeaderInitiator first =
+				new LockRegistryLeaderInitiator(firstRegistry, new DefaultCandidate());
+		CountDownLatch firstGranted = new CountDownLatch(1);
+		CountDownLatch firstRevoked = new CountDownLatch(1);
+		first.setLeaderEventPublisher(new CountingPublisher(firstGranted, firstRevoked));
+
+		// set up first registry instance - this one will NOT be able to obtain lock initially
```
artembilan: s/first/second

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
```diff
@@ -100,6 +107,58 @@ public void competing() throws Exception {
...
+				new LockRegistryLeaderInitiator(secondRegistry, new DefaultCandidate());
+		CountDownLatch secondGranted = new CountDownLatch(1);
+		CountDownLatch secondRevoked = new CountDownLatch(1);
+		second.setLeaderEventPublisher(new CountingPublisher(secondGranted, secondRevoked));
+
+		// start initiators
+		first.start();
+		second.start();
+
+		Thread.sleep(1000);
```
artembilan: Let's make this timeout smaller!
Since we deal only with mocks, we can just decrease those `heartBeatMillis` and `busyWaitMillis` to small milliseconds and sleep here only for `100`.
WDYT?

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
```diff
@@ -100,6 +107,58 @@ public void competing() throws Exception {
...
+
+		// first initiator should lead and publish granted event
+		assertThat(first.getContext().isLeader(), is(true));
+		assertThat(second.getContext().isLeader(), is(false));
+		assertThat(firstGranted.await(10, TimeUnit.SECONDS), is(true));
+
+		// simulate first registry instance unable to obtain lock, for example due to lock timeout
+		given(firstLock.tryLock(anyLong(), any(TimeUnit.class))).willReturn(false);
```
artembilan: This is the problem with Mockito.
You can't re-stub the same method on mock.

I guess `Mockito.reset(firstLock)` might help us, but it isn't an advice from the Mockito itself. See `reset()` JavaDocs.

I think you should consider to implement an `Answer` for the first `given()` which will return `false` according some external condition, e.g. some `AtomicBoolean` variable change here. For example:
````java
AtomicBoolean firstLocked = new AtomicBoolean(true);
...
given(firstLock.tryLock(anyLong(), any(TimeUnit.class))).willAnswer(i -> firstLocked.get());
...
given(secondLock.tryLock(anyLong(), any(TimeUnit.class))).willAnswer(i -> !firstLocked.get());
...
// simulate first registry instance unable to obtain lock, for example due to lock timeout
firstLocked.set(false);
````

I also do this in your test:
```
first.setBusyWaitMillis(1);
first.setHeartBeatMillis(10);
...
Thread.sleep(100);
```
For better performance.

artembilan (changes requested):
@vpavic ,

Thank you very much for the fix!
Overall I find this as a good improvement.
Only a couple concerns to the test-case.

vpavic:
Hi @artembilan, thanks for the prompt feedback on this!

I fully agree with the tweaks you proposed and have updated the PR accordingly.

artembilan (approved):
LGTM
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/d9340030bbdb5ca6e17643ce48c03927aaf88c1a and cherry-picked to `master` as https://github.com/spring-projects/spring-integration/commit/fb7e4d68691adf869fc5884ec925a99646b77906.

Please, be aware that we have to stop `initiator` in the end of test. 
Otherwise it spins for a while causing hangs and OutOfMemory 😄 

Will fix `4.3.x` soon

@vpavic ,
thank you very much for one more valuable contribution!
As usual looking forward for more.

Pay attention, please, how I have modified commit message.
The main concern to have head line no longer then `50` symbols.
More info is in the legendary Chris Beams' [article](https://chris.beams.io/posts/git-commit/).

vpavic:
> Please, be aware that we have to stop initiator in the end of test.
Otherwise it spins for a while causing hangs and OutOfMemory :smile:

Whoops! Sorry about that, I guess I was too focused on the problematic assertions so stopping the initiators completely slipped my mind :smile:

Do you have any rough estimate of the `4.3.9.RELEASE` release date?

garyrussell:
We would normally release with alongside 5.0.0.M3 which is scheduled for March 28 (next week) but we can do it earlier if urgent.

vpavic:
March 28th is OK for us, thanks @garyrussell!

PR #2108, opened 2017-04-03
Upgrade to Reactor-3.0.6

artembilan:
According to some rework and deprecations in the latest Reactor,
fix Reactor-based components to follow with the latest paradigmas

* Use `Flux` wrapper in the `ReactiveChannel` to track cancellation
instead of deprecated `Operators.SubscriberAdapter`
* Do similar `Flux.from(publisher)` in the `doSubscribeTo()`
to avoid subscribers delegation and, therefore, unnecessary call stack
* Rework `ReactiveConsumer` to wrap to a new `BaseSubscriber`
on each `ReactiveConsumer.start()`.
We need this wrapping to prevent subscriber cancellation from the
upstream `Publisher` on error.
We can't rely here on the `Flux.retry()` because upstream
`DirectProcessor` is marked as terminated on any error

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,7 +125,7 @@ subprojects { subproject ->
 		pahoMqttClientVersion = '1.0.2'
 		postgresVersion = '9.1-901-1.jdbc4'
 		reactorNettyVersion = '0.6.2.RELEASE'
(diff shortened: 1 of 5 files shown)


garyrussell:
Merged as 499e1167d1619c2d93b4cf07e61e4fa0ed47ec18

PR #2114, opened 2017-04-06
INT-4248: Refactor `RedisLockRegistry`

vpavic:
JIRA: https://jira.spring.io/browse/INT-4248

This is still WIP, please review @artembilan. The changes to `RedisLockRegistry` are quite significant and the implementation is now very similar to one in `JdbcLockRegistry`.

Tests are a mess ATM, I'll address them after you confirm the approach taken here is OK with you. I did however include a test that demonstrate the issue originally described in Jira issue - take a look at `RedisLockRegistryTests#testExpireTwoRegistries`.

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -16,36 +16,28 @@
 
 package org.springframework.integration.redis.util;
 
-import java.net.InetAddress;
-import java.net.UnknownHostException;
-import java.nio.ByteBuffer;
-import java.time.Instant;
-import java.time.ZoneId;
-import java.time.format.DateTimeFormatter;
-import java.util.Arrays;
-import java.util.Collection;
+import java.text.SimpleDateFormat;
 import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
-import java.util.stream.Collectors;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
-import org.springframework.data.redis.core.RedisCallback;
-import org.springframework.data.redis.core.RedisTemplate;
-import org.springframework.data.redis.core.TimeoutUtils;
-import org.springframework.data.redis.serializer.RedisSerializer;
-import org.springframework.data.redis.serializer.SerializationException;
-import org.springframework.data.redis.serializer.StringRedisSerializer;
-import org.springframework.integration.support.locks.DefaultLockRegistry;
+import org.springframework.data.redis.core.StringRedisTemplate;
+import org.springframework.data.redis.core.script.DefaultRedisScript;
+import org.springframework.data.redis.core.script.RedisScript;
+import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.integration.support.locks.LockRegistry;
 import org.springframework.util.Assert;
 
@@ -66,61 +58,47 @@ import org.springframework.util.Assert;
  * <p>
  * <b>Note: This is not intended for low latency applications.</b> It is intended
  * for resource locking across multiple JVMs.
- * When a lock is released by a remote system, waiting threads may take up to 100ms
- * to acquire the lock.
- * A more performant version would need to get notifications from the Redis stores
- * of key changes. This is currently only available using the SYNC command.
- * <p>
- * This limitation will usually not apply when a lock is released within this registry,
- * unless another system takes the lock after the local lock is acquired here.
- * A {@link DefaultLockRegistry} is used internally to achieve this optimization.
  * <p>
  * {@link Condition}s are not supported.
  *
  * @author Gary Russell
  * @author Konstantin Yakimov
  * @author Artem Bilan
+ * @author Vedran Pavic
  * @since 4.0
  *
  */
-public final class RedisLockRegistry implements LockRegistry {
+public final class RedisLockRegistry implements ExpirableLockRegistry {
 
 	private static final Log logger = LogFactory.getLog(RedisLockRegistry.class);
 
-	private static final byte[] hostName;
-
 	private static final long DEFAULT_EXPIRE_AFTER = 60000;
+
+	private static final String OBTAIN_LOCK_SCRIPT =
+			"local lockClientId = redis.call('GET', KEYS[1])\n" +
+			"if lockClientId == ARGV[1] then\n" +
+			"  redis.call('PEXPIRE', KEYS[1], ARGV[2])\n" +
+			"  return true\n" +
+			"elseif not lockClientId then\n" +
+			"  redis.call('SET', KEYS[1], ARGV[1], 'PX', ARGV[2])\n" +
+			"  return true\n" +
+			"end\n" +
+			"return false";
+
+	private final ConcurrentMap<String, RedisLock> locks = new ConcurrentHashMap<>();
+
+	private final String clientId = UUID.randomUUID().toString();
 
 	private final String registryKey;
 
-	private final RedisTemplate<String, RedisLock> redisTemplate;
+	private final StringRedisTemplate redisTemplate;
 
-	private final ThreadLocal<Set<RedisLock>> weakThreadLocks = new ThreadLocal<Set<RedisLock>>();
-
-	private final ThreadLocal<List<RedisLock>> hardThreadLocks = new ThreadLocal<List<RedisLock>>();
+	private final RedisScript<Boolean> obtainLockScript;
 
 	private final long expireAfter;
 
-	private final LockRegistry localRegistry;
-
-	private final LockSerializer lockSerializer = new LockSerializer();
-
-	private boolean useWeakReferences = false;
-
-	static {
-		String host;
-		try {
-			host = InetAddress.getLocalHost().getHostName();
-		}
-		catch (UnknownHostException e) {
-			host = "unknownHost";
-		}
-		hostName = host.getBytes();
-	}
-
 	/**
-	 * Constructs a lock registry with the default (60 second) lock expiration and a default
-	 * local {@link DefaultLockRegistry}.
+	 * Constructs a lock registry with the default (60 second) lock expiration.
 	 * @param connectionFactory The connection factory.
 	 * @param registryKey The key prefix for locks.
 	 */
@@ -129,208 +107,65 @@ public final class RedisLockRegistry implements LockRegistry {
 	}
 
 	/**
-	 * Constructs a lock registry with the supplied lock expiration and a default
-	 * local {@link DefaultLockRegistry}.
+	 * Constructs a lock registry with the supplied lock expiration.
 	 * @param connectionFactory The connection factory.
 	 * @param registryKey The key prefix for locks.
 	 * @param expireAfter The expiration in milliseconds.
 	 */
 	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, long expireAfter) {
-		this(connectionFactory, registryKey, expireAfter, new DefaultLockRegistry());
-	}
-
-	/**
-	 * Constructs a lock registry with the supplied lock expiration and a custom local {@link LockRegistry}.
-	 * @param connectionFactory The connection factory.
-	 * @param registryKey The key prefix for locks.
-	 * @param expireAfter The expiration in milliseconds.
-	 * @param localRegistry The local registry used to reduce wait time,
-	 * {@link DefaultLockRegistry} is used by default.
-	 */
-	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey,
-			long expireAfter, LockRegistry localRegistry) {
 		Assert.notNull(connectionFactory, "'connectionFactory' cannot be null");
 		Assert.notNull(registryKey, "'registryKey' cannot be null");
-		Assert.notNull(localRegistry, "'localRegistry' cannot be null");
-		this.redisTemplate = new RedisTemplate<>();
-		this.redisTemplate.setConnectionFactory(connectionFactory);
-		this.redisTemplate.setKeySerializer(new StringRedisSerializer());
-		this.redisTemplate.setValueSerializer(this.lockSerializer);
-		this.redisTemplate.afterPropertiesSet();
+		this.redisTemplate = new StringRedisTemplate(connectionFactory);
+		this.obtainLockScript = new DefaultRedisScript<>(OBTAIN_LOCK_SCRIPT, Boolean.class);
 		this.registryKey = registryKey;
 		this.expireAfter = expireAfter;
-		this.localRegistry = localRegistry;
-	}
-
-	/**
-	 * Change the state of thread local weak references storage for unlocked locks.
-	 * Thread local weak references are used for lock obtaining optimization -
-	 * thread will get same {@link RedisLock} object for certain key before actual
-	 * locking and after unlocking (if variable still exists).
-	 * <p>While is switched off (by default) every {@link RedisLockRegistry#obtain} call will provide
-	 * different {@link RedisLock} objects for same unlocked key.
-	 * @param useWeakReferences set to true for switch thread local weak references storage on, false by default
-	 * @since 4.0.7
-	 */
-	public void setUseWeakReferences(boolean useWeakReferences) {
-		this.useWeakReferences = useWeakReferences;
-	}
-
-	/**
-	 * Weak referenced locks, lock is kept here when actual lock is NOT gained.
-	 * Used for obtaining same lock object within same thread and key.
-	 * To avoid memory leaks lock objects without actual lock are kept as weak references.
-	 * After gaining the actual lock, lock object moves from weak reference to hard reference and vise a versa.
-	 */
-	private Collection<RedisLock> getWeakThreadLocks() {
-		Set<RedisLock> locks = this.weakThreadLocks.get();
-		if (locks == null) {
-			locks = Collections.newSetFromMap(new WeakHashMap<RedisLock, Boolean>());
-			this.weakThreadLocks.set(locks);
-		}
-		return locks;
-	}
-
-	/**
-	 * Hard referenced locks, lock is kept here when actual lock is gained.
-	 */
-	private Collection<RedisLock> getHardThreadLocks() {
-		List<RedisLock> locks = this.hardThreadLocks.get();
-		if (locks == null) {
-			locks = new LinkedList<>();
-			this.hardThreadLocks.set(locks);
-		}
-		return locks;
-	}
-
-	private RedisLock findLock(Collection<RedisLock> locks, Object key) {
-		if (locks != null) {
-			for (RedisLock lock : locks) {
-				if (lock.getLockKey().equals(key)) {
-					return lock;
-				}
-			}
-		}
-		return null;
-	}
-
-	private void toHardThreadStorage(RedisLock lock) {
-		if (this.weakThreadLocks.get() != null) {
-			this.weakThreadLocks.get().remove(lock);
-		}
-
-		getHardThreadLocks().add(lock);
-
-		//clean up
-		if (this.weakThreadLocks.get() != null && this.weakThreadLocks.get().isEmpty()) {
-			this.weakThreadLocks.remove();
-		}
-	}
-
-	private void toWeakThreadStorage(RedisLock lock) {
-		//to avoid collection creation on existence check use direct fields
-		if (this.hardThreadLocks.get() != null) {
-			getHardThreadLocks().remove(lock);
-		}
-
-		if (this.useWeakReferences) {
-			getWeakThreadLocks().add(lock);
-		}
-
-		//clean up
-		if (this.hardThreadLocks.get() != null && this.hardThreadLocks.get().isEmpty()) {
-			this.hardThreadLocks.remove();
-		}
 	}
 
 	@Override
 	public Lock obtain(Object lockKey) {
 		Assert.isInstanceOf(String.class, lockKey);
-
-		//try to find the lock within hard references
-		RedisLock lock = findLock(this.hardThreadLocks.get(), lockKey);
-
-		/*
-		 * If the lock is locked, check that it matches what's in the store.
-		 * If it doesn't, the lock must have expired.
-		 */
-		if (lock != null && lock.thread != null) {
-			RedisLock lockInStore = this.redisTemplate.boundValueOps(this.registryKey + ":" + lockKey).get();
-			if (lockInStore == null || !lock.equals(lockInStore)) {
-				try {
-					lock.unlock();
-				}
-				catch (Exception e) {
-					if (logger.isWarnEnabled()) {
-						logger.warn("Lock was released due to expiration. A new one will be obtained...", e);
-					}
-				}
-				if (this.hardThreadLocks.get() != null) {
-					this.hardThreadLocks.get().remove(lock);
-				}
-				if (this.weakThreadLocks.get() != null) {
-					this.weakThreadLocks.get().remove(lock);
-				}
-				lock = null;
-			}
-		}
-
-		if (lock == null) {
-			//try to find the lock within weak references
-			lock = findLock(this.weakThreadLocks.get(), lockKey);
-
-			if (lock == null) {
-				lock = new RedisLock((String) lockKey);
-
-				if (this.useWeakReferences) {
-					getWeakThreadLocks().add(lock);
-				}
-			}
-		}
-
-		return lock;
+		String path = (String) lockKey;
+		return this.locks.computeIfAbsent(path, RedisLock::new);
 	}
 
-	public Collection<Lock> listLocks() {
-		return this.redisTemplate.execute((RedisCallback<Collection<Lock>>) connection -> {
-			Set<byte[]> keys = connection.keys((RedisLockRegistry.this.registryKey + ":*").getBytes());
-			if (keys.size() > 0) {
-				List<byte[]> locks = connection.mGet(keys.toArray(new byte[keys.size()][]));
-				return locks.stream()
-						.map(RedisLockRegistry.this.lockSerializer::deserialize)
-						.collect(Collectors.toList());
+	@Override
+	public void expireUnusedOlderThan(long age) {
+		synchronized (this.locks) {
+			Iterator<Map.Entry<String, RedisLock>> iterator = this.locks.entrySet().iterator();
+			long now = System.currentTimeMillis();
+			while (iterator.hasNext()) {
+				Map.Entry<String, RedisLock> entry = iterator.next();
+				RedisLock lock = entry.getValue();
+				if (now - lock.getLockedAt() > age && !lock.isAcquiredInThisProcess()) {
+					iterator.remove();
+				}
 			}
-			return Collections.emptyList();
-		});
+		}
 	}
 
 	private final class RedisLock implements Lock {
 
 		private final String lockKey;
 
-		private long lockedAt;
+		private volatile long lockedAt;
 
-		private Thread thread;
+		private ReentrantLock localLock = new ReentrantLock();
 
-		private String threadName;
-
-		private byte[] lockHost;
-
-		private int reLock;
-
-		RedisLock(String lockKey) {
-			this.lockKey = lockKey;
-			this.lockHost = RedisLockRegistry.hostName;
+		private RedisLock(String path) {
+			this.lockKey = constructLockKey(path);
 		}
 
-		private String getLockKey() {
-			return this.lockKey;
+		private String constructLockKey(String path) {
+			return RedisLockRegistry.this.registryKey + ":" + path;
+		}
+
+		public long getLockedAt() {
+			return this.lockedAt;
 		}
 
 		@Override
 		public void lock() {
-			Lock localLock = RedisLockRegistry.this.localRegistry.obtain(this.lockKey);
-			localLock.lock();
+			this.localLock.lock();
 			while (true) {
 				try {
 					while (!this.obtainLock()) {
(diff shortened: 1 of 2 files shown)


artembilan:
Looks awesome at a glance, @vpavic !

I'm pretty sure you can go ahead with tests polishing.
I'll review properly tomorrow.

Only concern for now that we can't just remove public API in the point version. It must be `@Deprecated` with an appropriate JavaDocs.
Also I don't think that we need to keep that script in the separate file. Unlike SQL DDL scripts, this one isn't for installation in the end-user environment - just for invocation, like `SELECT`s.

Thank you very much!

vpavic:
Thanks @artembilan!

> Only concern for now that we can't just remove public API in the point version. It must be `@Deprecated` with an appropriate JavaDocs.
Also I don't think that we need to keep that script in the separate file. Unlike SQL DDL scripts, this one isn't for installation in the end-user environment - just for invocation, like `SELECT`s.

I've updated the PR to address this. Will revisit the tests later today.

artembilan:
> I've updated the PR to address this.

Please, don't squash commits for PR.
It is much easier to trace changed during by the commit history.
We do squash during merge into `upstream`.

vpavic:
> Please, don't squash commits for PR.
It is much easier to trace changes during review by the commit history.

:+1: I'll keep that in mind for future Spring Integration contributions.

I've extracted the `RedisLockRegistry` diff applied in the last PR update from my local history:

```patch
--- spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java	(revision )
+++ spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java	(revision )
@@ -17,6 +17,7 @@
 package org.springframework.integration.redis.util;
 
 import java.text.SimpleDateFormat;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
@@ -31,16 +32,15 @@
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import org.springframework.core.io.ClassPathResource;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.core.RedisTemplate;
 import org.springframework.data.redis.core.script.DefaultRedisScript;
 import org.springframework.data.redis.core.script.RedisScript;
 import org.springframework.data.redis.serializer.StringRedisSerializer;
+import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.integration.support.locks.LockRegistry;
-import org.springframework.scripting.support.ResourceScriptSource;
 
 /**
@@ -78,6 +78,18 @@
 
 	private static final String CLIENT_ID_HASH_KEY = "clientId";
 
+	private static final String OBTAIN_LOCK_SCRIPT =
+			"local lockClientId = redis.call('HGET', KEYS[1], '" + CLIENT_ID_HASH_KEY + "')\n" +
+			"if lockClientId == ARGV[1] then\n" +
+			"  redis.call('PEXPIRE', KEYS[1], ARGV[2])\n" +
+			"  return true\n" +
+			"elseif not lockClientId then\n" +
+			"  redis.call('HMSET', KEYS[1], '" + CLIENT_ID_HASH_KEY + "', ARGV[1])\n" +
+			"  redis.call('PEXPIRE', KEYS[1], ARGV[2])\n" +
+			"  return true\n" +
+			"end\n" +
+			"return false";
+
 	private final Map<String, RedisLock> locks = new HashMap<String, RedisLock>();
 
 	private final String clientId = UUID.randomUUID().toString();
@@ -109,11 +121,25 @@
 		Assert.notNull(connectionFactory, "'connectionFactory' cannot be null");
 		Assert.notNull(registryKey, "'registryKey' cannot be null");
 		this.redisTemplate = createRedisTemplate(connectionFactory);
-		this.obtainLockScript = createObtainLockScript();
+		this.obtainLockScript = new DefaultRedisScript<Boolean>(OBTAIN_LOCK_SCRIPT, Boolean.class);
 		this.registryKey = registryKey;
 		this.expireAfter = expireAfter;
 	}
 
+	/**
+	 * Constructs a lock registry with the supplied lock expiration and a custom local {@link LockRegistry}.
+	 * @param expireAfter The expiration in milliseconds.
+	 * @param localRegistry The local registry used to reduce wait time,
+	 * {@link DefaultLockRegistry} is used by default.
+	 * @deprecated since 4.3.9
+	 */
+	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey,
+			long expireAfter, LockRegistry localRegistry) {
+		this(connectionFactory, registryKey, expireAfter);
+	}
+
 	private static RedisTemplate<Object, Object> createRedisTemplate(RedisConnectionFactory connectionFactory) {
 		RedisTemplate<Object, Object> template = new RedisTemplate<Object, Object>();
 		template.setConnectionFactory(connectionFactory);
@@ -122,12 +148,18 @@
 		return template;
 	}
 
-	private static RedisScript<Boolean> createObtainLockScript() {
-		DefaultRedisScript<Boolean> script = new DefaultRedisScript<Boolean>();
-		script.setScriptSource(new ResourceScriptSource(
-				new ClassPathResource("org/springframework/integration/redis/util/obtainlock.lua")));
-		script.setResultType(Boolean.class);
-		return script;
+	/**
+	 * Change the state of thread local weak references storage for unlocked locks.
+	 * Thread local weak references are used for lock obtaining optimization -
+	 * thread will get same {@link RedisLock} object for certain key before actual
+	 * locking and after unlocking (if variable still exists).
+	 * <p>While is switched off (by default) every {@link RedisLockRegistry#obtain} call will provide
+	 * different {@link RedisLock} objects for same unlocked key.
+	 * @param useWeakReferences set to true for switch thread local weak references storage on, false by default
+	 * @since 4.0.7
+	 * @deprecated since 4.3.9
+	 */
+	public void setUseWeakReferences(boolean useWeakReferences) {
 	}
 
 	@Override
@@ -160,6 +192,10 @@
 		}
 	}
 
+	public Collection<Lock> listLocks() {
+		return Collections.<Lock>unmodifiableCollection(this.locks.values());
+	}
+
 	private final class RedisLock implements Lock {
 
 		private final String lockKey;
@@ -260,8 +296,8 @@
 
 		private boolean obtainLock() {
 			boolean success = RedisLockRegistry.this.redisTemplate.execute(RedisLockRegistry.this.obtainLockScript,
-					Collections.<Object>singletonList(constructLockKey()), CLIENT_ID_HASH_KEY,
-					RedisLockRegistry.this.clientId, String.valueOf(RedisLockRegistry.this.expireAfter));
+					Collections.<Object>singletonList(constructLockKey()), RedisLockRegistry.this.clientId,
+					String.valueOf(RedisLockRegistry.this.expireAfter));
 			if (success) {
 				this.lockedAt = System.currentTimeMillis();
 			}
```

artembilan:
@kiakimov, hello!

You might be interested in this change. 
At least you did the most hard work for the `RedisLockRegistry`, so maybe you can provide your review as well.
Thank you!

kiakimov:
Hello @artembilan and @vpavic 
I investigated an issue with the sample @vpavic kindly provided, so here is some thoughts:

- RedisLockRegistry does check actual TTL but on obtaining not on actual locking, so from the first sight I would say that it is improper LockRegistry usage - better to obtain the lock in the place it is going to be used. LockRegistryLeaderInitiator obtains it once - in constructor. If we update LockRegistryLeaderInitiator with proper obtaining then sample will work but will produce a warning each time it catches the expire (I checked it).

- From another hand we have a gap between obtaining and actual locking so we have obvious inconsistency - Redis lock can expire right after the obtaining so we will face the same issue @vpavic captured. As for me, I'd rather say that actual Redis lock expire time is some kind of fallback - to be not be blocked on lock holder crash, in regular scenario unlock will be called before expire and locked routine has to be much faster comparing to Redis TTL.

- Purposed solution removes local optimizations and if we do so then current implementation could be simplified to just `SET NX <key> <expire>` (for each `obtain`) and local keys cache. Anyway without Redis events we cant relay on such distributed lock - we still have inconsistency. The only thing we have is polling (LockRegistryLeaderInitiator does it) and post-mortem exception on obtain/unlock. Basically we need to choose what do we want more: local optimizations with some usage tradeoffs (focus on frequent lock usages) or straight Redis pass thru with bigger latencies  (focus on distributed competitions). Anyway both solutions have same inconsistency. As for LockRegistryLeaderInitiator issue - moving `obtain` closer to the `tryLock` and increased TTL will solve the problem without RedisLockRegistry changes - this is the case of distributed competition when local optimizations bring no profit.

P.S. btw looks like LockRegistryLeaderInitiator has one more issue - in the last `finally` around its polling routine it tries to unlock in any case - even if it had no lock - exception is guaranteed.

vpavic:
@artembilan I've updated the tests.

Hi @kiakimov, thanks for sharing.

> RedisLockRegistry does check actual TTL but on obtaining not on actual locking, so from the first sight I would say that it is improper LockRegistry usage - better to obtain the lock in the place it is going to be used.

Can you explain why do you think this isn't a proper usage of `LockRegistry`? I didn't find any javadoc and/or documentation suggesting this.

> I'd rather say that actual Redis lock expire time is some kind of fallback - to be not be blocked on lock holder crash, in regular scenario unlock will be called before expire and locked routine has to be much faster comparing to Redis TTL.

> Purposed solution removes local optimizations and if we do so then current implementation could be simplified to just SET NX <key> <expire> (for each obtain) and local keys cache.

IMO you need to update the lock state in the underlying data store each time it is re-acquired, otherwise the TTL will kick-in to delete the lock which can, and often times will, result in change of leading `LockRegistryLeaderInitiator` instance.

In a system with quite a lot of `SmartLifecycle` components that depend on `LockRegistryLeaderInitiator` to manage them this results in needless stopping and starting of those components, which can be quite costly operation. Once a `LockRegistryLeaderInitiator` instance is granted leadership I wouldn't expect it to lose leadership other than in some sort of exceptional situation (shutdown, crash, etc.).

artembilan:
@kiakimov , 

thank you very much for your valuable feedback!

@vpavic ,

thanks to you too for update and following discussion!

My opinion is:

The `LockRegistry.obtain()` must just return a `Lock` instance, nothing more. Like it is with the `DefaultLockRegistry`, `GemfireLockRegistry`, `ZookeeperLockRegistry`. I don't talk here about `JdbcLockRegistry` because this PR is based on that one and I don't want the "holly war" that the fix conflicts with the previous purpose. 

It's really already the `lock()` responsibility to track expired or not. And I think that really would be better to follow with all other distributed locks solutions: update repository on each `lock()` to actualize the `expire` property. Seems for me that is exactly the purpose of distributed locks without any local optimizations do not involving repository state.

Sorry, @kiakimov , but @vpavic 's fix is more closer to me then the previous (yours) local optimization 😉 

Anyway, would you mind, @kiakimov , to raise a JIRA for that `LockRegistryLeaderInitiator` concern about its `finally` logic? Thanks

kiakimov:
@artembilan mb I was not clear but actually I tried to raise the question do we really need that optimizations and I didnt implement it, I just fixed memory leak there ;)
So if we really dont need this optimization then Im absolutely fine with this decision :)

@vpavic yes, u're right - there is no such suggestions, I use redis locks for a couple of years and I used to obtain it near the actual lock just because of relaying on current implementation, so yes, there is no such requirements

Guys, if you dont mind I'll review the purposed PR in a few days, thank you both for the feedback and discussion

@artembilan for the mentioned issue - https://jira.spring.io/browse/INT-4254

artembilan:
Ah! That was the point, @kiakimov ! 😄 All you do is adapt your application for an existing implementation.
So, that is big "Huh!.." that we all think the same way.

That would be great to have your review, @kiakimov , and maybe some tests against your real applications to be sure before merging. I'd appreciate to have it until this Wednesdays. Otherwise I'm going to vacation since Thurthday and you both will have to explain everything one more time to @garyrussell to enter him into the subject. (If he doesn't follow us silently already, of course 😉 )

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -147,20 +133,20 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
 	 * @param expireAfter The expiration in milliseconds.
 	 * @param localRegistry The local registry used to reduce wait time,
 	 * {@link DefaultLockRegistry} is used by default.
+	 * @deprecated since 4.3.9
```
artembilan: `- the local {@link LockRegistry} isn't used any more`

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -172,171 +158,65 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
 	 * different {@link RedisLock} objects for same unlocked key.
 	 * @param useWeakReferences set to true for switch thread local weak references storage on, false by default
 	 * @since 4.0.7
+	 * @deprecated since 4.3.9
```
artembilan: `- the local optimization isn't used any more because of cluster-wide inconsistency`

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -364,168 +244,85 @@ private void rethrowAsLockException(Exception e) {
...
-				localLock.unlock();
+				this.localLock.unlock();
 				rethrowAsLockException(e);
 			}
 			return false;
 		}
 
+		private boolean obtainLock() {
+			boolean success = RedisLockRegistry.this.redisTemplate.execute(RedisLockRegistry.this.obtainLockScript,
+					Collections.<Object>singletonList(constructLockKey()), RedisLockRegistry.this.clientId,
```
artembilan: I think we have to call `constructLockKey()` only once in the ctor. That is really logical that the `RedisLock` instance is tied with the `Registry` and must not be equal to any other instance in different registry even if their keys are the same.
I mean the `equals()` impl here.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -364,168 +244,85 @@ private void rethrowAsLockException(Exception e) {
...
+					String.valueOf(RedisLockRegistry.this.expireAfter));
```
artembilan: Why it must be `String`?
Doesn't Redis driver convert it automatically if needed?

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -97,13 +100,15 @@ public void testLock() throws Exception {
...
-				assertNotNull(TestUtils.getPropertyValue(registry, "hardThreadLocks", ThreadLocal.class).get());
+				assertEquals(1, TestUtils.getPropertyValue(registry, "locks", Map.class).size());
 			}
 			finally {
 				lock.unlock();
 			}
 		}
-		assertNull(TestUtils.getPropertyValue(registry, "hardThreadLocks", ThreadLocal.class).get());
+		Thread.sleep(10);
+		registry.expireUnusedOlderThan(0);
```
artembilan: `registry.expireUnusedOlderThan(-1000);` instead of `Thread.sleep()` 😉 
I don't like any blocking in tests for nothing.

artembilan (changes requested):
That's all from my side.
Excellent work, @vpavic !

Only several minor concerns which I even can address on merge if that.

@kiakimov , I'm sure we will wait for your feedback anyway.

Thank you both!

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
vpavic: Nice - I guess the same optimization could be done in other places as well (for example `JdbcLockRegistryTests`).

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
vpavic: Well, it's important to consider the execution context of Lua script - I've only been able to make things work this way. Revisiting `RedisTemplate` setup, I guess we could switch to `StringRedisTemplate` since that's what we effectively use. It would save a few lines of code as well. WDYT?

vpavic:
@artembilan I've addressed most of your comments, just see [this comment](https://github.com/spring-projects/spring-integration/pull/2114#discussion_r110976706) on use of `RedisTemplate`.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
artembilan: 👍 Yes, that sounds reasonable

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
artembilan: Well, that's true. Although I see only one `Thread.sleep()` there 😄

artembilan:
So far so good, @vpavic .

I have answered to your `StringRedisTemplate` remark.
That is the last thing I could ask you to do.

Only what we need before merging is @kiakimov 's 👍 

I'll do some minor Migration Guide note on the matter. Just in case we are protected from unexpected questions 😄

artembilan:
Here we go: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.2-to-4.3-Migration-Guide#redislockregistry !

Doing that I realized that we are missing some important part like lock object in the store.
I see that you removed `LockSerializer`. But what is the alternative now @vpavic ?
Thanks

vpavic:
Well, we don't need a `LockSerializer` since with this new implementation we're dealing with simple `String` values (we're writing a hash using [`HMSET`](https://redis.io/commands/hmset)) hence the `StringRedisSerializer` is now used.

When writing lock to the store, all we need now is lock key, client id and TTL. The first two are natural `String` values while TTL is sent to the wire stringified since [`PEXPIRE`](https://redis.io/commands/pexpire) command that sets it is executed by Lua script handles obtain lock "transaction".

Regarding reading operations, all we actually ever read is client id, again in Lua script and in `RedisLockRegistry.RedisLock#isAcquiredInThisProcess`.

Hope this explains the storage/serialization concerns. This approach in completely analogous to one from `JdbcLockRegistry`.

kiakimov (commented):
Hi @artembilan, @vpavic!

Here is my concerns summarizing my review:

1) Instead of GC duty to clear unused locks we move this responsibility to the user, without calling `expireUnusedOlderThan` it is easy to get into troubles (especially if user missed migration guide).

2) We break previous behaviour - it is impossible to get all cluster wide locks with suggested implementation.

3) No guarantee to get same Lock object from the same thread even in sequential calls - coz of `expireUnusedOlderThan` work. It is debatable coz Im not sure is it a big deal or not, but definitely if someone relays on such behaviour he/she wont be happy with new version.

So in scope of raised problem I'd rather suggest leave most of the code and fix just one problem - sudden expire of the remote key protected by local optimization - just fix RedisLock.obtainLock method and do additional Redis read there with skipping entrance local optimization check.

Ofcouse if we would talk about local project my concerns would cost nothing so it is up to you guys to choose the way - backward compatibility or code simplification.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -172,171 +149,69 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
...
-				if (this.hardThreadLocks.get() != null) {
-					this.hardThreadLocks.get().remove(lock);
-				}
-				if (this.weakThreadLocks.get() != null) {
-					this.weakThreadLocks.get().remove(lock);
-				}
-				lock = null;
+		String path = (String) lockKey;
+		RedisLock lock;
+		synchronized (this.locks) {
```
artembilan: 👍 without comments.
We also can consider to address the same in the `JdbcLockRegistry` and `ZookeeperLockRegistry`

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -172,171 +149,69 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
...
-
-			if (lock == null) {
-				lock = new RedisLock((String) lockKey);
-
-				if (this.useWeakReferences) {
-					getWeakThreadLocks().add(lock);
+	@Override
+	public void expireUnusedOlderThan(long age) {
+		synchronized (this.locks) {
+			Iterator<Map.Entry<String, RedisLock>> iterator = this.locks.entrySet().iterator();
```
artembilan: `ExpirableLockRegistry`:
```
* A {@link LockRegistry} implementing this interface supports the removal of aged locks
* that are not currently locked.
```
Since we can't assume if the lock in the Redis is locked in some other process, we can remove only local instances.
And from the [Reference Manual](http://docs.spring.io/spring-integration/reference/html/zookeeper.html#zk-lock-registry):

> For unbounded strategies (such as the default) you will need to invoke expireUnusedOlderThan(long age) from time to time, to remove old unused locks from memory.

So, we are really talking here only about **in-memory**.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -172,171 +149,69 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
...
-				if (keys.size() > 0) {
-					List<byte[]> locks = connection.mGet(keys.toArray(new byte[keys.size()][]));
-					for (byte[] lock : locks) {
-						list.add(RedisLockRegistry.this.lockSerializer.deserialize(lock));
-					}
-				}
-				return list;
-			}
-		});
+		return Collections.<Lock>unmodifiableCollection(this.locks.values());
```
artembilan: I think this is good catch, @kiakimov !
And it reflects my question about `Lock` serialization, @vpavic .

Although this API is pretty arbitrary and similar in the `JdbcLockRegisrty` returns only local instances, and even any other impls don't provide such an API, I think that should not be so critical to be consistent with existing functionality and make this method to return all the locks from the Redis.

I don't know why why needs it, because there is no use for a `Lock` object obtained in the different process, but that is what we had before.

So, would you mind, @vpavic , to restore `Lock` serialization in some form for this solution? E.g. some simple JSON representation?

Thanks!

artembilan (changes requested):
> without calling expireUnusedOlderThan

Well, we may consider to use `LinkedHashMap.removeEldestEntry()` to protect those people who don't look into migration guide.

>  it is impossible to get all cluster wide locks with suggested implementation.

As I said in the line comment: we should to restore previous behavior in this part

> No guarantee to get same Lock object from the same thread even in sequential calls

Good point! I guess we don't have choice unless document that properly. And I guess the best place is exactly JavaDoc of the `expireUnusedOlderThan()`

> So in scope of raised problem I'd rather suggest leave most of the code and fix just one problem 

That's the good point, too. I think we will come to something closer to the "minimal fix" when all these latest concerns will be addressed.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
vpavic: We should store additional fields in the hash that represents the lock (currently there's only `clientId`), and use [`HGETALL`](https://redis.io/commands/hgetall) to load the entire hash and reconstruct the `RedisLock`.

JSON serialization isn't suitable here as we need to consider the Lua script for obtaining lock - we don't want to parse `clientId` from a JSON in there :smile:

> Although this API is pretty arbitrary and similar in the `JdbcLockRegisrty` returns only local instances, and even any other impls don't provide such an API, I think that should not be so critical to be consistent with existing functionality and make this method to return all the locks from the Redis.

Can you clarify the _similar in the `JdbcLockRegisrty` returns only local instances_ part a bit? AFAICT `JdbcLockRegistry` doesn't offer functionality similar to `RedisLockRegistry#listLocks`.

Also could we consider providing this in `4.3.x`, to maintain backwards compatibility, and deprecate `RedisLockRegistry#listLocks` and remove it in `5.0`? It's not a part of `LockRegistry` contract and as you say _the API is pretty arbitrary_. IMO it's unfortunate that such method forces us to make our storage more complicated than they would've been otherwise.

artembilan:
Well, if understand HMSET properly, we can store several field under that key. So, my point was to leave `clientId` as separate property, but add one more for lock JSON representation, which should be read only and only for that list method. 

On the other hand I really can't imagine the use-case for that list method and how the phantom locks (deserialized instances from othe cluster members) can be useful . More over that can be dangerous because they will be returned with their own remote `clientId`. And what is bad we don't have control over that in the current impl. We might add something like read only RedisLock version if client is different from the Registry deserialing lock .

Any other ideas?

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
vpavic: I don't quite agree that `ConcurrentMap#putIfAbsent` does the same since it returns previous value from the map if it puts, rather than the one we've just put into the map. It also forces us to eagerly create `RedisLock` instance meaning with `ConcurrentMap` we have to do something like this:

```java
RedisLock newLock = new RedisLock(path);
RedisLock oldLock = this.locks.putIfAbsent(path, newLock);
return (oldLock == null) ? newLock : oldLock;
```

I'd personally leave it as it is. Thoughts @artembilan?

vpavic:
I've given this some thought over the past 2 days and I didn't like any of the options of implementing this. What also bothers me is that if we implement this as suggested, is that we'd return a different object for the same lock if it was fetched via `RedisLockRegistry#obtain` vs `RedisLockRegistry#listLocks`.

TBH from the first moment `RedisLockRegistry#listLocks` looked like a method for testing purposes that was accidentally made `public` instead of `package-private`. After all, I'd expect users to interact with the lock registry via `LockRegistry`/`ExpirableLockRegistry` and `Lock` APIs.

Like mentioned in one of the [previous comments](https://github.com/spring-projects/spring-integration/pull/2114#discussion_r111316748), I'd prefer to deprecate this in `4.3.x` and remove in `5.0` as we did in two other occasions in this PR.

artembilan:
@vpavic ,

I сompletely share your concerns about `listLocks()` API. 
Yes, we can deprecate it in `4.3.x` and remove in `5.0` as something redundant and confusing.
But! We should have some strong argument do not allow that API.

The `ReentrantLock` is `Serializable`, for example. Therefore can be deserialized in other process and used for its locking purpose. So, the same may be applied for the `RedisLock`, but yeah, we definitely should not let to  use the `RedisLock` created in other client.

I might definitely won't mind to remove that API in `5.0` and simplify the store representation and network communication as you are doing right now, but I insist that we have to provide some compatible fix for the `4.3.x`. 

The  `@Deprecated` ctor and setter do nothing because they don't affect our new internals.
But this `listLocks()` must not change behavior and really return expected list of locks from the store.

That is one vision.

@garyrussell , 

we would like to hear your opinion as well, please.

In two words the problem is like.

When we obtain the lock, each its subsequent `lock()` doesn't update entity in the store and eventually we lose the connection with the `expired` property. The other registry instance decides that the lock is old and locks its local version, too. We see the problem with the `LockRegistryLeaderInitiator`.

The suggested fix is fully similar to the `JdbcLockRegistry` and introduce `clientId` - who is locker.
And also remove some "useless", but introduce breaking changes to the `listLocks()`.

Since the concern is valid, but unfortunately it is so breaking change that the other way to go for `4.3.x` might be really like @kiakimov suggests: 

>  If we update `LockRegistryLeaderInitiator` with proper obtaining then sample will work but will produce a warning each time it catches the expire (I checked it).

All the rest thorough work leave for the `5.0` version.

Thanks

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -172,171 +149,69 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
...
 	}
 
 	private final class RedisLock implements Lock {
 
 		private final String lockKey;
 
-		private long lockedAt;
-
-		private Thread thread;
+		private volatile long lockedAt = System.currentTimeMillis();
```
vpavic: Yes, this seems unnecessary.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -364,187 +239,104 @@ private void rethrowAsLockException(Exception e) {
...
+				this.localLock.unlock();
 				rethrowAsLockException(e);
 			}
 			return false;
 		}
 
+		private boolean obtainLock() {
+			boolean success = RedisLockRegistry.this.redisTemplate.execute(RedisLockRegistry.this.obtainLockScript,
+					Collections.singletonList(this.lockKey), RedisLockRegistry.this.clientId,
+					String.valueOf(RedisLockRegistry.this.expireAfter));
```
vpavic: Now that you mention it I also think we're good with a simple `SET` command. I guess I started with `HMSET` since I expected we'll be storing more attributes and never revisited that.

vpavic:
@artembilan I've addressed your latest concerns.

garyrussell:
I haven't really been following this dicussion, but I am a bit wary of putting such a big change into a point release (4.3.9). I wonder if we should first put it in our upcoming 5.0.0.M4 milestone (in a couple of weeks) and let people kick its tires there and then put it in a 4.3.x release after it's been shaken out.

Or am I being too paranoid?

artembilan:
Let me provide some fix for the `4.3.x` based on the @kiakimov 's suggestion!

Will come back to you soon with an update...

artembilan:
Guys, this is what I suggest as a fix for `4.3.x`: #2121

vpavic:
@artembilan Good to see `4.3.x` has been taken care of with a simpler fix! :+1:

So I guess should I rebase this PR against `master` now? With one additional question - how do we now handle methods that were planned to be deprecate in `4.3.x` and removed in `5.0`?

artembilan:
Sorry, @vpavic , missed that part in my first round. 
Here we go: #2122!

And yes, your branch must be rebased against `master` now and you feel free to go ahead with discussed solution.

vpavic:
@artembilan I've rebased the PR against the `master` branch.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f7f7bdd067428ae570c6bc0cfe489e1698c0a203.

Thank you @vpavic for crazy work!
And especial thanks to @kiakimov for thorough review and ideas for fixes.

Looking forward for more to work with your guys, even if I'm not so useful in the subject 😄

PR #2115, opened 2017-04-07
Upgrade to the SF and SA BUILD-SNAPSHOTs

artembilan:
* Upgrade to Jackson-2.9.0 to align with SF
* Fix `ReactiveHttpRequestExecutingMessageHandler` to use latest `WebClient`
* Introduce `Supplier<URI> uriSupplier` to evaluate the `realUri` on demand.
In case of `WebClient` use `.uri(Function)` to perform `uriSupplier`
* Move `doNext()` for the `WebClientException` out of `one-way` block.
According the `RestTemplate` logic the erroneous response is handled
before returning from the `exchange()`.
Therefore both request/reply and one-way scenarios must have the same
error handling logic
* Some refactoring to the `AbstractHttpRequestExecutingMessageHandler`
and `AbstractHttpRequestExecutingMessageHandler` according a new `uriSupplier`


artembilan:
This should be reviewed and merged ASAP - the nightly IO plan fails because of no SF 5.0 B-S compatibility.

PR #2116, opened 2017-04-11
INT-4254: LockRegLInit: Fix unconditional unlock

artembilan:
JIRA: https://jira.spring.io/browse/INT-4254

The `LockRegistryLeaderInitiator#LeaderSelector` unconditional calls
`unlock()` in the `finally` block when the `Lock` might not be locked.

Another problem that `running = true` is set after submitting `LeaderSelector` task.
That might bring the problem that `LeaderSelector` won't be selected because of
`this.running` race condition

* Move `this.running = true` before submitting `LeaderSelector` task
* Move `unlock()` into the `if (this.locked)` condition

**Cherry-pick to 4.3.x**


garyrussell:
Cherry-picked as 2dc21e5d406957534edbe8dec060c9b5361835bc

PR #2117, opened 2017-04-12
INT-4050: Refactoring for JDBC schema scripts

artembilan:
JIRA: https://jira.spring.io/browse/INT-2625,
https://jira.spring.io/browse/INT-4050,
https://jira.spring.io/browse/INT-4052

Make some refactoring for consistency:

- Move `JdbcMessageStore` to the `store` package
- Make `JdbcMessageStore` only ctor-based configuration for the `JdbcOperations`
- Remove `/channel/` SQL scripts in favor of combined in the `/jdbc/` directory
- Make scripts for the `INT_CHANNEL_MESSAGE` table and indexes generated via VPP


artembilan:
I will add Migration Guide note after review.

spring-integration-jdbc/src/main/sql/destroy.sql.vpp
```diff
@@ -1,5 +1,9 @@
-DROP TABLE $!{IFEXISTSBEFORE} INT_MESSAGE $!{IFEXISTS};
-DROP TABLE $!{IFEXISTSBEFORE} INT_MESSAGE_GROUP $!{IFEXISTS};
-DROP TABLE $!{IFEXISTSBEFORE} INT_GROUP_TO_MESSAGE $!{IFEXISTS};
-DROP TABLE $!{IFEXISTSBEFORE} INT_LOCK $!{IFEXISTS};
-DROP INDEX $!{IFEXISTSBEFORE} INT_MESSAGE_IX1 $!{IFEXISTS};
+DROP INDEX$!{IFEXISTSBEFORE}INT_MESSAGE_IX1$!{IFEXISTS};
```
vpavic: Space missing before `$!{IFEXISTSBEFORE}` causes invalid syntax in generated scripts.

vpavic (commented):
Looks good with the exception of invalid drop scripts. Looking forward to this change :smile:

artembilan:
Thank you, @vpavic , for quick review!
Fixed your concern.

I wonder if we should still support the old MySQL. Quoting Reference Manual:

> If you plan on using *MySQL*, please use MySQL version _5.6.4_ or higher, if possible.
Prior versions do not support _fractional seconds_ for temporal data types.
Because of that, messages may not arrive in the precise FIFO order when polling from such a MySQL Message Store.

So, let's encourage users to go ahead and upgrade their MySQL! 😄 

@garyrussell ?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
```diff
@@ -204,20 +200,31 @@ public String getSql() {
...
 	/**
 	 * Convenient constructor for configuration use.
+	 * @deprecated since 5.0 in favor of {@link #JdbcMessageStore(DataSource)}
 	 */
+	@Deprecated
 	public JdbcMessageStore() {
-		this.deserializer = new DeserializingConverter();
-		this.serializer = new SerializingConverter();
+		throw new UnsupportedOperationException("Please, use DataSource-based constructor");
 	}
```
garyrussell: Usually deprecated methods still work, we just want to discourage use; since this is a major release, and given this exception, I see no point in deprecating - just get rid of it.

We could/should deprecate in 4.3, though.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
```diff
@@ -242,23 +249,23 @@ public void setRegion(String region) {
...
 
 	/**
-	 * The JDBC {@link DataSource} to use when interacting with the database. Either this property can be set or the
-	 * {@link #setJdbcTemplate(JdbcOperations) jdbcTemplate}.
-	 *
+	 * The JDBC {@link DataSource} to use when interacting with the database.
 	 * @param dataSource a {@link DataSource}
+	 * @deprecated in favor of {@link #JdbcMessageStore(DataSource)}
 	 */
+	@Deprecated
```
garyrussell: ditto; remove.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
```diff
@@ -242,23 +249,23 @@ public void setRegion(String region) {
...
 	/**
-	 * The {@link JdbcOperations} to use when interacting with the database. Either this property can be set or the
-	 * {@link #setDataSource(DataSource) dataSource}.
-	 *
+	 * The {@link JdbcOperations} to use when interacting with the database.
 	 * @param jdbcTemplate a {@link JdbcOperations}
+	 * @deprecated in favor of {@link #JdbcMessageStore(JdbcOperations)}
 	 */
+	@Deprecated
 	public void setJdbcTemplate(JdbcOperations jdbcTemplate) {
```
garyrussell: ditto

garyrussell (changes requested):

garyrussell:
I am not sure that consolidating the scripts is a good idea.

For users that only use the channel-optimized store and not the general store (and vice versa) will get tables that they never use. Some DBAs will throw a tantrum over that.

What is the motivation?

garyrussell:
I see no compelling reason to drop support for older MySQLs.

OTOH, these DDL are samples only so I guess I don't mind either way as long as the note remains in the reference.

vpavic:
Hi @garyrussell - regarding the motivation, spring-projects/spring-boot#8881 aims to provide convenient database initializer for Spring Integration users. Equivalent functionality already exists for other Spring projects such as Batch and Session.

Such database initializers use stock scripts from corresponding project but can be configured to use user provided custom script as well. From Boot's perspective, ideally there would be a single script provided with Spring Integration that initializes schema.

Other than that, current situation is inconsistent anyway, since scripts in `org.springframework.integration.jdbc` initializes both `JdbcMessageStore` and `DefaultLockRepository` tables while scripts in `org.springframework.integration.jdbc.store.channel` initializes `JdbcChannelMessageStore` tables. Splitting tables for each use case only makes things less convenient. Also consider that there might be new use cases/tables introduced in the future.

As you said yourself these scripts are samples and certainly won't be able to cover everyone's requirements. IMO splitting tables to dedicated scripts won't help much there.

artembilan:
+1 for what @vpavic says. Really was going to answer something similar. And sorry that I didn't add such a description into PR. And that Boot stuff is really the reason why I suggest to drop an old MySQL script. The version number does provides some confusion since it collides with the latest MySQL version. Plus having an argument as "this scripts are samples" that really does not make sense to distinguish by the version. And that's why we decided to drop version in the Oracle script.

Thank you both for the feedback! Will update PR the next week after PTO.

artembilan:
The refactoring and deprecation for `4.3.x`: #2120

garyrussell:
Merged as 5224779a04cf2417828482cc08e277e225dba10e

PR #2118, opened 2017-04-14
[Discussion Only] Upgrade to Reactor 3.1

garyrussell:
__Discussion only__


garyrussell:
I went ahead and merged this as effca79f3f9c83070cbc49cffd8d24997a8422d0 because master was broken, but we should still discuss when @artembilan gets back from PTO.

PR #2119, opened 2017-04-16
Reactor 3.1 Polishing

garyrussell:
Use FluxProcessor.wrap rather than change the argument to FluxProcessor.


garyrussell:
Per discussion, `FluxProcessor<>` is correct.

PR #2124, opened 2017-04-20
INT-4260: MessagePublishingErrorHandler Orig. Msg.

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4260

MPEH: Populate the `ErrorMessage.originalMessage`, if available and not the same as the `failedMessage`.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
@@ -23,7 +23,7 @@ import java.util.List;
 import java.util.concurrent.Executor;
 
 import org.springframework.integration.dispatcher.AbstractDispatcher;
-import org.springframework.integration.support.OriginalMessageContainingMessagingException;
+import org.springframework.integration.support.MessagingExceptionWrapper;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHandler;
@@ -156,7 +156,7 @@ public abstract class AbstractExecutorChannel extends AbstractSubscribableChanne
 					triggerAfterMessageHandled(message, ex, interceptorStack);
 				}
 				if (ex instanceof MessagingException) {
-					throw new OriginalMessageContainingMessagingException(message, (MessagingException) ex);
+					throw new MessagingExceptionWrapper(message, (MessagingException) ex);
 				}
 				String description = "Failed to handle " + message + " to " + this + " in " + messageHandler;
 				throw new MessageDeliveryException(message, description, ex);
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.concurrent.Executor;
 
 import org.springframework.integration.dispatcher.AbstractDispatcher;
+import org.springframework.integration.support.OriginalMessageContainingMessagingException;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHandler;
@@ -43,6 +44,7 @@ import org.springframework.util.CollectionUtils;
  * is handed to the {@link Executor#execute(Runnable)}.
  *
  * @author Artem Bilan
+ * @author Gary Russell
  * @see ExecutorChannel
  * @see PublishSubscribeChannel
  * @since 4.2
@@ -154,7 +156,7 @@ public abstract class AbstractExecutorChannel extends AbstractSubscribableChanne
 					triggerAfterMessageHandled(message, ex, interceptorStack);
 				}
 				if (ex instanceof MessagingException) {
-					throw (MessagingException) ex;
+					throw new OriginalMessageContainingMessagingException(message, (MessagingException) ex);
 				}
 				String description = "Failed to handle " + message + " to " + this + " in " + messageHandler;
 				throw new MessageDeliveryException(message, description, ex);
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
@@ -16,10 +16,6 @@
 
 package org.springframework.integration.channel;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
-import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.support.ErrorMessagePublisher;
 import org.springframework.integration.support.ErrorMessageStrategy;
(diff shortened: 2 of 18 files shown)


spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
```diff
@@ -96,6 +108,16 @@ public void setSendTimeout(long sendTimeout) {
 		this.sendTimeout = sendTimeout;
 	}
 
+	/**
+	 * Set an {@link ErrorMessageStrategy} to use.
+	 * @param errorMessageStrategy the strategy.
+	 * @since 5.0
+	 */
+	public void setErrorMessageStrategy(ErrorMessageStrategy errorMessageStrategy) {
```
artembilan: Why don't we `extends ErrorMessagePublisher` ?

spring-integration-core/src/main/java/org/springframework/integration/support/OriginalMessageContainingMessagingException.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A {@link MessagingException} used to convey the cause and original message to
+ * a {@link org.springframework.integration.channel.MessagePublishingErrorHandler}.
+ *
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public class OriginalMessageContainingMessagingException extends MessagingException {
```
artembilan: I don't like this name - too long.
` OriginalMessageException` ?
That really isn't necessary to fully mention the super class.

spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
```diff
@@ -140,8 +162,12 @@ public final void handleError(Throwable t) {
 	}
 
 	private MessageChannel resolveErrorChannel(Throwable t) {
-		Message<?> failedMessage = (t instanceof MessagingException) ?
-				((MessagingException) t).getFailedMessage() : null;
+		Throwable actualThrowable = t;
+		if (t instanceof OriginalMessageContainingMessagingException) {
+			actualThrowable = t.getCause();
+		}
+		Message<?> failedMessage = (actualThrowable instanceof MessagingException) ?
```
artembilan: According to the `OriginalMessageContainingMessagingException` ctor the cause is always `MessagingException`.

Do you have anything else in mind?

artembilan (changes requested):

artembilan:
I wonder if you are going to address my comments somehow...
At least a short answer to explain would be great.
Thanks

garyrussell:
Sorry - missed these comments somehow.

spring-integration-core/src/main/java/org/springframework/integration/support/OriginalMessageContainingMessagingException.java
olegz: Hmm, looking at the class and what it does it actually seems a bit backwards, since this class represents a MessagingException containing original message (not the other way around). I personally don't mind long name so may be MessagingExceptionContainingOriginalMessage?

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -266,7 +267,17 @@ private boolean doPoll() {
...
+			try {
+				this.handleMessage(message);
+			}
+			catch (Exception e) {
+				if (e instanceof MessagingException) {
+					throw new OriginalMessageContainingMessagingException(message, (MessagingException) e);
+				}
+				else {
+					throw new MessagingException(message, e);
+				}
```
olegz: I am slightly confused. Why do we need a new class if existing one (`MessagingException`) provides everything you need to hold and access both the _cause_ and the _message_.
Is it just a Marker interface? If so you could still simplify its implementation since you can still rely on `getFailedMessage()` and other provided functionality 
What am I missing?

garyrussell: The intention is to enhance the error message so that it has the `failedMessage` (the message at the point of failure) and the `originalMessage` (the message that started the flow) - it could have undergone many transformations before it failed.

artembilan: Looking to this one more time, I'd say it would be much better to have `originalMessage` on the `MessagingException` alongside with the `failedMessage`.
From here, of course, we don't need the same `originalMessage` on the `ErrorMessage` at all...

I think Oleg's idea is like wrap caught `MessagingException` it into a new instance but use `message` as a `failedMessage` already. 

On the client side we would just scan the cause for the targeted `failedMessage`.
That way it would be much confusing, of course, but may work 😉

garyrussell: I think that would be difficult for people to understand and it would be a big breaking change.

With this solution, people who don't care about the original message will be oblivious to it.

garyrussell: >Is it just a Marker interface? 

Kind of; it's so we know to unwrap the original exception and populate the enhanced error message. But I see you're point, it could be a simple subclass of ME and use the failedMessage property (otherwise it doesn't need to be an ME).

Fixing...

spring-integration-core/src/main/java/org/springframework/integration/support/OriginalMessageContainingMessagingException.java
garyrussell: I went with `MessagingExceptionWrapper` since that really describes what it does, and simply use its `failedMessage` property as discussed below.

spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
garyrussell: Yes, but `t` might be any throwable (not in our code but we can't control user code).

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
```diff
@@ -154,7 +156,7 @@ public void run() {
 					triggerAfterMessageHandled(message, ex, interceptorStack);
 				}
 				if (ex instanceof MessagingException) {
-					throw (MessagingException) ex;
+					throw new MessagingExceptionWrapper(message, (MessagingException) ex);
```
artembilan: Well, technically it doesn't have much difference with the previous implementation.
If we consider this only as for an internal use, I'm fine with any name for the exception class.
Although we might mention that in its JavaDocs.

spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
```diff
@@ -41,30 +42,34 @@
  * @author Gary Russell
  * @author Artem Bilan
  */
-public class MessagePublishingErrorHandler implements ErrorHandler, BeanFactoryAware {
+public class MessagePublishingErrorHandler extends ErrorMessagePublisher implements ErrorHandler, BeanFactoryAware {
```
artembilan: `BeanFactoryAware`is already present on the `ErrorMessagePublisher`.

spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
```diff
@@ -41,30 +42,34 @@
...
+	private static final ErrorMessageStrategy DEFAULT_ERROR_MESSAGE_STRATEGY = (t, a) -> {
+		if (t instanceof MessagingExceptionWrapper) {
+			return new ErrorMessage(t.getCause(), ((MessagingExceptionWrapper) t).getFailedMessage());
+		}
+		else {
+			return new ErrorMessage(t);
+		}
+	};
 
+	private final Log logger = LogFactory.getLog(this.getClass());
```
artembilan: There is a `protected logger`  in super class already.

artembilan (changes requested):
Just a couple concerns.

@olegz , 

if this is fine with you, let's ask @garyrussell to add some docs on the matter and merge!

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/572467b1c49719786bcf117090e398c9bfcc790b

PR #2125, opened 2017-04-20
INT-4261: Do not register resource to TX twice

artembilan:
JIRA: https://jira.spring.io/browse/INT-4261

When we have some `Advice` withing TX Advice which may perform `doPoll()`
several times, we unconditionally call
`transactionSynchronizationFactory.create(resource)`.
With the out-of-the-box implementations
`DefaultTransactionSynchronizationFactory` and
`PassThroughTransactionSynchronizationFactory`
we preform `TransactionSynchronizationManager.bindResource()`.
If resource is already there, an `IllegalStateException` is thrown

* Check that resource isn't bound already to the TX and don't create a new
`TransactionalResourceSynchronization` - just return `null`
* Check in the target users for the `null` before registering synchronization

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -320,7 +320,7 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 					integrationSynchronization.setShouldUnbindAtCompletion(false);
 
 					if (!TransactionSynchronizationManager.hasResource(resource)) {
-						TransactionSynchronizationManager.bindResource(key,
+						TransactionSynchronizationManager.bindResource(resource,
 								integrationSynchronization.getResourceHolder());
 					}
 				}
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -310,18 +310,21 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 		if (this.transactionSynchronizationFactory != null && resource != null &&
 				TransactionSynchronizationManager.isActualTransactionActive()) {
 			TransactionSynchronization synchronization = this.transactionSynchronizationFactory.create(resource);
-			TransactionSynchronizationManager.registerSynchronization(synchronization);
-			if (synchronization instanceof IntegrationResourceHolderSynchronization) {
-				IntegrationResourceHolderSynchronization integrationSynchronization =
-						((IntegrationResourceHolderSynchronization) synchronization);
-				integrationSynchronization.setShouldUnbindAtCompletion(false);
-				IntegrationResourceHolder resourceHolder = integrationSynchronization.getResourceHolder();
-				if (key != null) {
-					resourceHolder.addAttribute(key, resource);
+			if (synchronization != null) {
+				TransactionSynchronizationManager.registerSynchronization(synchronization);
+				if (synchronization instanceof IntegrationResourceHolderSynchronization) {
+					IntegrationResourceHolderSynchronization integrationSynchronization =
+							((IntegrationResourceHolderSynchronization) synchronization);
+					integrationSynchronization.setShouldUnbindAtCompletion(false);
+					IntegrationResourceHolder resourceHolder = integrationSynchronization.getResourceHolder();
+					if (key != null) {
+						resourceHolder.addAttribute(key, resource);
+					}
+					return resourceHolder;
 				}
-				return resourceHolder;
 			}
 		}
+
 		return null;
 	}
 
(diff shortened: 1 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -310,18 +310,21 @@ private IntegrationResourceHolder bindResourceHolderIfNecessary(String key, Obje
 		if (this.transactionSynchronizationFactory != null && resource != null &&
 				TransactionSynchronizationManager.isActualTransactionActive()) {
 			TransactionSynchronization synchronization = this.transactionSynchronizationFactory.create(resource);
-			TransactionSynchronizationManager.registerSynchronization(synchronization);
```
garyrussell: You can simply move this inside the if statement. The instance of test will return false for null.

spring-integration-core/src/main/java/org/springframework/integration/transaction/DefaultTransactionSynchronizationFactory.java
```diff
@@ -26,10 +26,13 @@
  * Default implementation of {@link TransactionSynchronizationFactory} which takes an instance of
  * {@link TransactionSynchronizationProcessor} allowing you to create a {@link TransactionSynchronization}
  * using {{@link #create(Object)} method.
+ * <p>
+ * If resource under the provided {@code key} is already registered, the factory returns {@code null}.
  *
```
garyrussell: This is a bit smelly.

garyrussell: Maybe add an exists () method?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transaction/DefaultTransactionSynchronizationFactory.java
artembilan: We can, but with that the `TransactionSynchronizationFactory` stops to be `@FunctionalInterface`. And there is no guarantee that custom implementation won't return `null` by some reason.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
artembilan: With your suggestion we change behavior when the `TransactionSynchronization` is not `null` and not `IntegrationResourceHolderSynchronization`. Right now we register it independently of its type.

garyrussell: Note to self: Don't try to review PRs on your 'phone

spring-integration-core/src/main/java/org/springframework/integration/transaction/DefaultTransactionSynchronizationFactory.java
```diff
@@ -46,13 +49,16 @@ public DefaultTransactionSynchronizationFactory(TransactionSynchronizationProces
 	@Override
 	public TransactionSynchronization create(Object key) {
 		Assert.notNull(key, "'key' must not be null");
-		DefaultTransactionalResourceSynchronization synchronization = new DefaultTransactionalResourceSynchronization(key);
-		TransactionSynchronizationManager.bindResource(key, synchronization.getResourceHolder());
-		return synchronization;
+		if (!TransactionSynchronizationManager.hasResource(key)) {
```
garyrussell: I think it would be cleaner to move this test to the callers rather than this strange return `null` if something's already bound with the key.

artembilan: OK. So, let's move `bindResource()` to caller as well!

artembilan: Well, no. I'm not agreed.

That is really the factory responsibility to bind resource to the TX or not.
And since there is no guarantee that factory will always return the same `Synchronization` instance for the same `key`, we can't unequivocally make a decision in the caller what to do.

I only see the `null` way do not create extra `Synchronization` from our default factories and do nothing afterwards.

garyrussell: I can live with it if we deprecate and rename the method to something like `createAndBindSynchronizationIfNeeded`.

artembilan: ??? That isn't a `TransactionSynchronizationFactory` contract. That is only our impls side effect that we register something to the TX.
The custom one can return the same `Sync` for the same `key`, or always new, or always `null`.
That is only our design that we bind resource to the TX. 
In fact (according this JIRA) we can't register into the same resource and according our logic we even don't need to create a new `Sync`. The custom `factory` can do what ever it needs: creates and returns a new instance, but bind if it is necessary for their purpose.

On the other hand renaming method (contract in the interface) doesn't protect us from the unpredictable custom implementations.
We may just say on the `create()` JavaDocs that implementation may return `null`.

garyrussell: We seem to have broken this "factory" in INT-3408; it's no longer a simple factory that creates a sync. It's hard to see how a user TSF that does __not__ bind the resource will work properly - the recent code to provide the message after a commit failure will not work if it doesn't do that. Hence I believe the caller (APE etc) __must__ do the binding.

artembilan: OK. Let me play with that a bit!
Thank you for your patience! 😄

garyrussell (approved):

garyrussell:
Merged as b8696660217c87740aa09f1ade598fe0634c23c4

PR #2131, opened 2017-04-24
INT-4257 Polishing - Subclasses: buildErrorMessage

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4257

So that subclasses can also generate error message for conditions such
as message conversion errors.

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -28,6 +28,7 @@ import org.springframework.integration.support.management.TrackableComponent;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessagingException;
+import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -187,18 +188,42 @@ protected void sendMessage(Message<?> message) {
...
+		if (errorChannel != null) {
+			this.messagingTemplate.send(errorChannel, buildErrorMessage(message, exception));
+			return true;
+		}
+		return false;
+	}
+
+	/**
+	 * Build an error message for the exception and message using the configured
+	 * {@link ErrorMessageStrategy}.
```
artembilan: Or let's make this method `final` or mentioning of the `ErrorMessageStrategy` does not make sense.
The implementor can simply override this method on his own.
And `ErrorMessageStrategy` premise may be eliminated.

artembilan (changes requested):

artembilan (approved):

garyrussell:
oops (travis)

artembilan:
Yeah... I see that locally.
To have `buildErrorMessage()` overridable is fine with me.

garyrussell:
This is better, I think, since we don't have a message in this case.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/c66540593c6bb18fd56564d2a0f0ab056c402475 and cherry-picked.

PR #2132, opened 2017-04-24
INT-4257: Polishing

garyrussell:
Make `EnhancedErrorMessage` ctors consistent with the new 5.0 ErrorMessage

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/message/EnhancedErrorMessage.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/message/EnhancedErrorMessage.java
@@ -42,17 +42,17 @@ public class EnhancedErrorMessage extends ErrorMessage {
 
 	private final Message<?> originalMessage;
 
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/message/EnhancedErrorMessage.java
```diff
@@ -42,17 +42,17 @@
 
 	private final Message<?> originalMessage;
 
-	public EnhancedErrorMessage(Message<?> originalMessage, Throwable payload) {
+	public EnhancedErrorMessage(Throwable payload, Message<?> originalMessage) {
```
artembilan: ??? Don't we need usages of this class changes?

garyrussell: Doh - I am not having a good day today ☹️

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b5c61581177e0582c089d0b6ca64692e17b520d0

PR #2133, opened 2017-04-25
INT-4263 Added MessagingAnnotationUtils.getValue() operation

olegz:
Added MessagingAnnotationUtils.getValue() operation that treats InvocationTargetException in special way allowing issue with synthesized annotation to be propagated
Added test

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingAnnotationUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingAnnotationUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2016 the original author or authors.
+ * Copyright 2014-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


olegz:
NOTE: This PR may end up to be obsolete before it is merged due to https://jira.spring.io/browse/SPR-15481. In other words the core issue is in the spring core and I had a quick chat with Juergen about it. If SPR-15481 gets into 4.3.9, then we may be good without this PR. 
Stay tuned

artembilan:
Good catch, @olegz !
Great fix and distribution of responsibility.

I wonder how did we live without you 😉 

I'd prefer to have the fix on the top level in the SF and apply your fix in the test-case part.
That really so sad that we didn't have an appropriate test before to fix such an issue from day first.

So, my opinion wait for the fix in SF and redo this PR to provide only test-case.

The Travis failure isn't related to your changes.
Looks like WebFlux has done one more breaking change. Sigh...

olegz:
Closing as it is no longer required as the bug was fixed in Spring Core and I've verified that SI is now acting appropriately

PR #2134, opened 2017-04-25
INT-4264: `.publishSubscribeChannel()`: fix order

artembilan:
JIRA: https://jira.spring.io/browse/INT-4264

When we use Java DSL `.publishSubscribeChannel()`, we expect that subscribers
will be called in the order they are defined in the flow.
But actually everything is depend on the populated `MessageHandler` implementation.
The `AbstractMessageHandler` implements `Ordered` and according the
`OrderedAwareCopyOnWriteArraySet` logic they are placed in the beginning of the
handlers set, meanwhile all others (un-ordered) are moved to the end.

* Add implicit `.bridge()` to all the subFlows added to the `PublishSubscribeSpec`.
This way all the subFlow subscribers will be ordered the same way (because of
`BridgeHandler`) and, therefore, will be preformed in the expected (as declared) order

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/dsl/publishsubscribe/PublishSubscribeTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/dsl/publishsubscribe/PublishSubscribeTests.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/test/java/org/springframework/integration/dsl/publishsubscribe/PublishSubscribeTests.java
```diff
@@ -0,0 +1,111 @@
...
+
+	@Test
+	public void executeFirstFlow() {
+		this.inputChannel.send(new GenericMessage<>("Test"));
+
+
+		assertEquals(6, this.subscribersOrderedCall.size());
+		for (int i = 0; i < 6; i++) {
+			assertEquals(new Integer(i), this.subscribersOrderedCall.get(i));
+		}
```
garyrussell: `assertThat(this.subscribersOrderedCall, contains(0, 1, 2, 3, 4, 5));` ?

Can change on merge if you agree.

artembilan: Doh...
Yes, it nails both my asserts:
```
  * Creates a matcher for {@link Iterable}s that matches when a single pass over the
   * examined {@link Iterable} yields a series of items, each logically equal to the
   * corresponding item in the specified items.  For a positive match, the examined iterable
   * must be of the same length as the number of specified items.
```
😄 

Yes, please, fix on merge.
Thank you!

garyrussell:
Merged as 3c047c5076450c8568cf0cab533754460573f454

PR #2136, opened 2017-04-26
INT-4265: Add Java DSL .routeByError() EIP-method

artembilan:
JIRA: https://jira.spring.io/browse/INT-4265

* Add `.routeByError()` EIP-method to the `IntegrationFlowDefinition`
based on the `ErrorMessageExceptionTypeRouter`
* Add missed `dynamicChannelLimit()` option to the `RouterSpec`
* Rework `RouterSpec#RouterMappingProvider` into the `ContextRefreshedEvent`
phase initialization to be sure that dependent `MappingMessageRouterManagement`
is fully initialized before applying mapping conversions

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
@@ -2034,7 +2034,7 @@ public abstract class IntegrationFlowDefinition<B extends IntegrationFlowDefinit
 	 * Typically used with a Java 8 Lambda expression:
 	 * <pre class="code">
 	 * {@code
-	 *  .routeByError(r -> r
+	 *  .routeByException(r -> r
 	 *      .channelMapping(IllegalArgumentException.class, "illegalArgumentChannel")
 	 *      .subFlowMapping(MessageHandlingException.class, sf ->
 	 *                                sf.handle(...))
@@ -2043,8 +2043,9 @@ public abstract class IntegrationFlowDefinition<B extends IntegrationFlowDefinit
 	 * </pre>
 	 * @param routerConfigurer the {@link Consumer} to provide {@link ErrorMessageExceptionTypeRouter} options.
 	 * @return the current {@link IntegrationFlowDefinition}.
+	 * @see ErrorMessageExceptionTypeRouter
 	 */
-	public B routeByError(
+	public B routeByException(
 			Consumer<RouterSpec<Class<? extends Throwable>, ErrorMessageExceptionTypeRouter>> routerConfigurer) {
 		return route(new RouterSpec<>(new ErrorMessageExceptionTypeRouter()), routerConfigurer);
 	}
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
```diff
@@ -178,28 +198,31 @@ void addMapping(Object key, NamedComponent channel) {
...
-			}
-			for (Map.Entry<Object, NamedComponent> entry : this.mapping.entrySet()) {
-				Object key = entry.getKey();
-				String channelKey;
-				if (key instanceof String) {
-					channelKey = (String) key;
-				}
-				else if (key instanceof Class) {
-					channelKey = ((Class<?>) key).getName();
+		public void onApplicationEvent(ContextRefreshedEvent event) {
```
garyrussell: Shouldn't we confirm that the event context is this bean's declaring context?

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
```diff
@@ -2029,6 +2030,26 @@ public B routeToRecipients(Consumer<RecipientListRouterSpec> routerConfigurer) {
...
+	 *      .channelMapping(IllegalArgumentException.class, "illegalArgumentChannel")
+	 *      .subFlowMapping(MessageHandlingException.class, sf ->
+	 *                                sf.handle(...))
+	 *    )
+	 * }
+	 * </pre>
+	 * @param routerConfigurer the {@link Consumer} to provide {@link ErrorMessageExceptionTypeRouter} options.
+	 * @return the current {@link IntegrationFlowDefinition}.
+	 */
+	public B routeByError(
```
garyrussell: Perhaps `routeByThrowable` would be more correct?

The javadocs for `ErrorMessageExceptionTypeRouter` are incorrect too (`Exception` Vs. `Throwable`).

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
artembilan: Well, `ErrorMessage` accepts `Throwable`, too...
We don't pursue here the strong alignment with types we use. Just human-friendly is enough, IMO...

spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
artembilan: I think so. Will fix.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
garyrussell: I agree, but specifically, here (and in the EMTR), we are routing on a `Throwable` payload, not an `ErrorMessage`.

garyrussell: `Error` has a meaning in Java and we shouldn't mislead.

artembilan:
Pushed. But I renamed it into `routeByException` . Somehow I don't feel comfort with all those `.able()` in the public method API names...
If you insist, I'll rename to the `routeByThrowable()`. That is just IMHO...

garyrussell:
Merged as f03356ccbcdf4394e7ab3daeeb3a7362a856e5f8

Issue #2110, opened 2017-04-04
Invalid information in RedisMessageStore documentation


urgelacko:
According to the documentation's last sentence below it should be possible to configure the RedisMessageStore to serialize messages in Jackson: http://docs.spring.io/spring-integration/reference/html/redis.html#redis-message-store

I couldn't manage to do it and I think it is impossible right now because of the following reasons:

- I should provide a valueSerializer which uses Jackson to serialize objects
- *RedisMessageStore* extends *AbstractKeyValueMessageStore *
- *AbstractKeyValueMessageStore * uses redisTemplate and the previously configured serializers for example for the following call: https://github.com/spring-projects/spring-integration/blob/v4.3.8.RELEASE/spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java#L114
- This means the Jackson serializer should deserialize to a MessageGroupMetadata but it is impossible since the MessageGroupMetadata class has no default constructor and the one accepts an object which shouldn't be serialized at all (MessageGroup). Additionally that argument is not needed in case of deserialization at all.
- Since MessageGroupMetadata is not an interface I can't provide my own implementation as a result of deserialization

I think if this is a known limitation or it is intended the documentation should be more precise about these limitations.

artembilan:
Well, I agreed and don't have a quick answer how to proceed.
On the one hand that document doesn't claim to use Jackson for (de)serialization.
But on the other one, even if we add default ctor to the `MessageGroupMetadata`, the `Message<?>` implementations are still stay non-compatible for JSON deserializaiton, because of several ctors there.

I don't want to drop that functionality since it promises some flexibility, but that really becomes useless in case of Jackson...

Any ideas besides dropping `setValueSerializer()` and polishing Docs for limitations?

Thanks

urgelacko:
Thank you for your quick response @artembilan 
As I see the biggest problem with serializing `MessageGroupMetadata `is the `groupId `field. However according to a call hierarchy check the `getGroupId()` method is never used in the framework nor in the class itself. So it seems a bit useless for me currently but maybe I do not understand something happening under the hood.

Actually if we skip the `groupId `field problem above the `MessageGroupMetadata `class needs minor modifications to make it serializable with out of the box *`GenericJackson2JsonRedisSerializer`* implementation.
I have created a JUnit test and it seems the followings are do the work:
- remove `final `modifiers from the fields
- introduce a private default constructor
- add a public `getMessageIds()` getter

[...]

PR #2138, opened 2017-05-02
INT-4267: Add JSON (De)Serializers for Messaging

artembilan:
JIRA: https://jira.spring.io/browse/INT-4267
Fixes: spring-projects/spring-integration#2110

The documentation clearly point that we can simply use JSON (de)serialization
with the `RedisMessageStore`, but actually it isn't so easy

* Fix `MessageGroupMetadata`, `MessageHolder`, `MessageMetadata` for Jackson
deserialization compatibility
* Add `MessageHeaders`-based ctor to the `AdviceMessage`
* Add `MessageHeadersJacksonSerializer` to serialize `MessageHeaders`
to the `HashMap` for easier deserialization afterwards
* Add deserializer implementations for all `Message` types
* Add convenient `JsonObjectMapperProvider#jacksonMessageAwareMapper()`
factory method to build `ObjectMapper` supplied with mentioned above
(de)serializers

**Cherry-pick to 4.3.10 without `MessageHolder` and `MessageMetadata`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupMetadata.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupMetadata.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.store;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -50,7 +51,7 @@ public class MessageGroupMetadata implements Serializable {
 	private volatile int lastReleasedMessageSequenceNumber;
 
 	private MessageGroupMetadata() {
-		//For Jackson
+		//For Jackson deserialization
 	}
 
 	public MessageGroupMetadata(MessageGroup messageGroup) {
@@ -92,7 +93,7 @@ public class MessageGroupMetadata implements Serializable {
 	}
 
 	public List<UUID> getMessageIds() {
-		return this.messageIds;
+		return Collections.unmodifiableList(this.messageIds);
 	}
 
 	void complete() {
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupMetadata.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupMetadata.java
@@ -31,17 +31,17 @@ import org.springframework.util.Assert;
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Laszlo Szabo
+ *
  * @since 2.1
  */
 public class MessageGroupMetadata implements Serializable {
 
 	private static final long serialVersionUID = 1L;
 
-	private final Object groupId;
+	private List<UUID> messageIds = new LinkedList<UUID>();
 
-	private final List<UUID> messageIds = new LinkedList<UUID>();
-
-	private final long timestamp;
+	private long timestamp;
 
 	private volatile boolean complete;
 
@@ -49,9 +49,12 @@ public class MessageGroupMetadata implements Serializable {
 
 	private volatile int lastReleasedMessageSequenceNumber;
 
+	private MessageGroupMetadata() {
+		//For Jackson
+	}
+
 	public MessageGroupMetadata(MessageGroup messageGroup) {
 		Assert.notNull(messageGroup, "'messageGroup' must not be null");
-		this.groupId = messageGroup.getGroupId();
 		for (Message<?> message : messageGroup.getMessages()) {
 			this.messageIds.add(message.getHeaders().getId());
 		}
@@ -73,10 +76,6 @@ public class MessageGroupMetadata implements Serializable {
 		this.lastModified = lastModified;
 	}
 
-	public Object getGroupId() {
-		return this.groupId;
-	}
-
 	public Iterator<UUID> messageIdIterator() {
 		return this.messageIds.iterator();
 	}
@@ -90,6 +89,10 @@ public class MessageGroupMetadata implements Serializable {
 			return this.messageIds.get(0);
 		}
 		return null;
+	}
+
+	public List<UUID> getMessageIds() {
+		return this.messageIds;
 	}
 
 	void complete() {
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/MessageMetadata.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/MessageMetadata.java
@@ -29,9 +29,13 @@ public class MessageMetadata implements Serializable {
 
 	private static final long serialVersionUID = 1L;
 
-	private final UUID messageId;
+	private UUID messageId;
 
 	private long timestamp;
+
+	private MessageMetadata() {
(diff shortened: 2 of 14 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/json/AdviceMessageJacksonDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * The {@link MessageJacksonDeserializer} implementation for the {@link AdviceMessage}.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.0
```
garyrussell: 4.3.10? (A bunch of these; I can fix on merge).

garyrussell (commented):
That's all.

spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupMetadata.java
```diff
@@ -92,6 +91,10 @@ public UUID firstId() {
 		return null;
 	}
 
+	public List<UUID> getMessageIds() {
+		return this.messageIds;
```
olegz: Should we return immutable list?

spring-integration-core/src/main/java/org/springframework/integration/store/MessageMetadata.java
```diff
@@ -29,10 +29,14 @@
 
 	private static final long serialVersionUID = 1L;
 
-	private final UUID messageId;
+	private UUID messageId;
```
olegz: Just curious why was `final` removed?

spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupMetadata.java
artembilan: Makes sense.
Thank you!

spring-integration-core/src/main/java/org/springframework/integration/store/MessageMetadata.java
artembilan: See the `private` empty ctor for Jackson deserialization.
Not sure if it will be able to inject value to the `final` property if we assign it to the `null` from that ctor...
I can take a look if that worry you a lot.

olegz: No, i was just curious. . . primarily due to _if not final should it be volatile_ kind of thing. . .

artembilan: It should be `volatile` if it potentially can be modified concurrently.

Frankly we have stopped to add `volatile` to those properties which mostly are modified only during initialization for a while already.

There is really some performance impact from the `volatile`. Can't find the link to discussion, but I recall that really was from you do not abuse `volatile`...

spring-integration-core/src/main/java/org/springframework/integration/support/json/MutableMessageJacksonDeserializer.java
```diff
@@ -0,0 +1,50 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * The {@link MessageJacksonDeserializer} implementation for the {@link MutableMessage}.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.0
+ */
+public class MutableMessageJacksonDeserializer extends MessageJacksonDeserializer<MutableMessage<?>> {
```
olegz: I was just chatting with @garyrussell and am wondering if this one is really needed given that MutableMessage _is not such a good thing to have_ and perhaps we should limit the support for it until we realize that we may get rid of it all together. Again, i may not know the full context, but figured I'll ask

artembilan: Well, the cat's outta the bag, and if " not such a good thing" as `MutableMessage` will arrive to your JSON-based `MessageStore` you will get some unexpected behavior. And there will be a feedback from the community why we don't support that one too?

If you think we can get rid of `MutableMessage` in the future, let's then deprecate this one there as well!

olegz: As I said, I don't know the full context yet, so I'll fall back on you two, but "mutable' in front of Message scares me a bit ;)
Your call

olegz: . . .actually scares me a lot! ;)

garyrussell: One of the reasons we introduced this was because there were several places in the framework where a message was rebuilt multiple times before sending out (header copies etc). This was intended for those uses. However, we now pass the message builder around instead, so that use case is no longer an issue.

olegz: So, then may be it's time to consider deprecating `MutableMessage` and gradually remove all references to it internally. . . (as a separate issue)

artembilan: Well, from big height that `MutableMessage` object just represents a wrapper around `MutableMessageHeaders`, since the message isn't responsible in `payload` mutability.
Also pay attention to the `MutableMessageBuilder`. The main point for this feature was like do not create new messages all the time in one single process. RTI team call.
Of course, you have to be careful with such an object type and maybe since it scares us, too, we have never mentioned it even in the documentation 😉

garyrussell: @artembilan I didn't notice at the time, but you use it in the new MIMH code @line 940.

Looks like that's the only place the framework uses it.

I remember now it was driven by RTI, not XD.

artembilan: I use `MutableMessage` there in the  code like:
```
if (this.canProcessMessageList) {
}
```
because `MutableMessageHeaders` doesn't generate `ID` and `TIMESTAMP`.
I need the message object there anyway because of `CollectionArgumentResolver` and as you see restricted contract of the `HandlerMethodArgumentResolver.resolveArgument()` where we can have only one `message` to consult.

garyrussell:
Merged as 7e919a4017c2760fa53c742ad91e529ee8c8e219 with minor doc polishing,

artembilan:
Cool! However this must be cherry-picked to `4.3.x` as well.
Thanks

garyrussell:
It's not a simple cherry-pick but I am working on it...

garyrussell:
Backported as 376405c39b1d96779940f83cdfc061431dd277b8 with fixes for `ErrorMessage` and changes to the test for branch differences.

PR #2139, opened 2017-05-02
INT-4259: ACorrelationMH: WARN group reschedule

artembilan:
JIRA: https://jira.spring.io/browse/INT-4259

Since DEBUG can switched off for the application, it becomes difficult
to determine the wrong configuration for the discard flow from
the scheduler thread.

* Change logging level to the `WARN` and log the full stack-trace as well
to give more information about the reason of rescheduling group for
`forceComplete`

**Cherry-pick to 4.3.x**


garyrussell:
And cherry-picked as e8396c5dbf7b0c0a85feb6b7918b4b36d511bb96 after resolving conflicts.

PR #2140, opened 2017-05-03
INT-4258: Redis: Make ZADD INCR false by default

artembilan:
JIRA: https://jira.spring.io/browse/INT-4258

Since Redis `ZADD` command doesn't perform incrementation by default,
it should be consistent in the `RedisStoreWritingMessageHandler` as well

* Make a default behavior of the `RedisStoreWritingMessageHandler`
regarding `INCR` option as `false`
* Provide some performance refactoring for the
`RedisStoreWritingMessageHandler`  moving the code to Java 8 style
and using `FunctionExpression`
* Fix tests according to the changed default behavior

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStoreWritingMessageHandler.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStoreWritingMessageHandler.java
@@ -75,11 +75,9 @@ import org.springframework.util.NumberUtils;
  */
 public class RedisStoreWritingMessageHandler extends AbstractMessageHandler {
 
-	private final Expression zsetIncrementScoreExpression =
+	private Expression zsetIncrementScoreExpression =
 			new FunctionExpression<Message<?>>(m ->
 					m.getHeaders().get(RedisHeaders.ZSET_INCREMENT_SCORE));
-
-	private volatile StandardEvaluationContext evaluationContext;
 
 	private volatile Expression keyExpression =
 			new FunctionExpression<Message<?>>(m ->
@@ -90,6 +88,8 @@ public class RedisStoreWritingMessageHandler extends AbstractMessageHandler {
 					m.getHeaders().get(RedisHeaders.MAP_KEY));
 
 	private volatile boolean mapKeyExpressionExplicitlySet;
+
+	private volatile StandardEvaluationContext evaluationContext;
 
 	private volatile RedisTemplate<String, ?> redisTemplate = new StringRedisTemplate();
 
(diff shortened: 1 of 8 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStoreWritingMessageHandler.java
```diff
@@ -72,22 +69,25 @@
...
  * @since 2.2
  */
 public class RedisStoreWritingMessageHandler extends AbstractMessageHandler {
 
-	private final Log logger = LogFactory.getLog(this.getClass());
-
 	private final Expression zsetIncrementScoreExpression =
-			new SpelExpressionParser().parseExpression("headers." + RedisHeaders.ZSET_INCREMENT_SCORE);
+			new FunctionExpression<Message<?>>(m ->
+					m.getHeaders().get(RedisHeaders.ZSET_INCREMENT_SCORE));
```
garyrussell: Not a part of this PR but why is this even an `Expression`? It's not configurable by the user at all.

... time passes ...

>* Fixed logic in RedisCollectionPopulatingMessageHandler to ensure that RedisHeaders.ZSET_INCREMENT_SCORE can be passed as Boolean true/false as well as String true/false. The current solution is based on relyiong oin ConversionService registered with SpEL Context (see RedisCollectionPopulatingMessageHandler.extractZsetIncrementHeader method). This way other types could be supported without introducing any code changes.

Interesting way to achieve that result.

Maybe we should just add a setter to allow the user to specify his own expression.

spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStoreWritingMessageHandler.java
```diff
@@ -284,49 +277,48 @@ private void writeToZset(RedisZSet<Object> zset, final Message<?> message) throw
...
 			}
 		}
 		else {
-			this.incrementOrOverwrite(ops, payload, this.determineScore(message), zsetIncrementHeader);
+			incrementOrOverwrite(ops, payload, this.determineScore(message), zsetIncrementHeader);
 		}
 	}
 
 	private boolean extractZsetIncrementHeader(Message<?> message) {
 		if (message.getHeaders().containsKey(RedisHeaders.ZSET_INCREMENT_SCORE)) {
```
garyrussell: If we open up the expression for user specification, this if test would have to go inside the default `FunctionExpression`.

garyrussell (commented):
That's all; WDYT about allowing user expressions?

garyrussell:
Needs a what's new/migration update.

artembilan:
I wonder what happened to GitHub: I don't have notification about your review 😕

artembilan:
Also see Migration Guilde: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.3-to-5.0-Migration-Guide#redisstorewritingmessagehandler

garyrussell:
Merged as 47433754b8acf5086a18f8658346465171cfb9cc

PR #2143, opened 2017-05-04
INT-4270: Make MessageChannels CGLib-Compatible

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4270

Remove `final` modifiers from channel methods and make `private` methods `protected`.

Since Spring Boot 2.0 now uses `proxyTargetClass=true` by default (and has done that for
transactional proxies since 1.4), we must relax these restrictions so that CGLib can
proxy channels (which is often done to make subflows run in a transaction).

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/channel/PublishSubscribeAmqpChannel.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/channel/PublishSubscribeAmqpChannel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


artembilan:
I can merge, of course, but shouldn't we have some tests to prove that we haven't missed something to open else?
Thanks

garyrussell:
Actually, the `private` is ok since it's only called by the delegate; it's just the `final` that's the problem; reverting those.

I am not sure what you mean about tests.

However, I do think we need to look at endpoints, especially `MessageHandler`s too.

artembilan:
I'm fine with relaxing our `final` policy.
Although we should keep in mind those called from ctors...

I mean tests with the CGLib mode. Since we haven't caught this before I guess we don't have any `proxy-target-class= true`  tests on the matter...

garyrussell:
Good point; ok; will rework and create tests - I am not sure if we can satisfy both requirements (only final setters from ctors and CGLib-compatible) - I somehow doubt it.

garyrussell:
Closing for now

PR #2146, opened 2017-05-08
Fix race condition in the AMQP ChannelTests

artembilan:
https://build.spring.io/browse/INT-AT42SIO-502

There is a small time window when we remove the current consumer
from the local store, but there is no a new one yet.
So, we have to check the `Set`(`Map`) size before calling its `iterator`

**Cherry-pick to 4.3.x**


artembilan:
If this is fine, I'll backport the fix to the `4.2.x` - that is based on the `Map consumers`, so cherry-pick might not be easy.

garyrussell:
And cherry-picked as 2c96f6a9582c1dc6aaba3ec46b5fe04918be8861

artembilan:
Cherry-picked to `4.2.x` as https://github.com/spring-projects/spring-integration/commit/0a7dfa96e7417b59e67cdb167b042d342d6be9c6

PR #2147, opened 2017-05-08
Fix Race condition in the TxSyncQueueChannelTests

artembilan:
https://build.spring.io/browse/INT-MASTER-653

The test-case uses only 1 second to wait for the message
in the `QueueChannel`.
That isn't enough on slow environment like CI causing failure for
the current test and unexpected value in the queue for the subsequent
tests

* Fix timeouts
* Get rid of `CountDownLatch` - the same is done by the wait on queues
* Purge queue in between tests.
This way we may not have failed subsequent test, thus failure analyze
will be much easier
* Rework `testRollback()` to wait for good result after one retry over
rollback
* Remove duplicate bean definition from XML config

**Cherry-pick to 4.3.x and 4.2.x**

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/channel/TransactionSynchronizationQueueChannelTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/channel/TransactionSynchronizationQueueChannelTests.java
@@ -18,20 +18,16 @@ package org.springframework.integration.channel;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertTrue;
 
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/test/java/org/springframework/integration/channel/TransactionSynchronizationQueueChannelTests.java
```diff
@@ -49,63 +45,59 @@
...
 	}
 
 	@Test
 	public void testIncludeChannelName() throws Exception {
-		service.latch = new CountDownLatch(1);
 		Message<String> sentMessage = MessageBuilder.withPayload("hello")
 				.setHeader("foo", "bar").build();
 		queueChannel2.send(sentMessage);
-		assertTrue(service.latch.await(10, TimeUnit.SECONDS));
 		Message<?> message = good.receive(1000);
```
garyrussell: Why not increase this one too? I can fix on merge.

artembilan: Huh! Missed.
Sorry and thank you!

garyrussell:
Merged as aca97e702547a493f25fd811563c2060af11f811
cherry-picked as 194d710c3b5d4fac339e15dc0786530b580bd18d to 4.3.x and c4bbc3b35601714fb689d89faa496ea0ad5c10f7 to 4.2.x.

PR #2148, opened 2017-05-08
Add `MockMessageHandler` for testing framework

artembilan:


artembilan:
@Toparvion, @tdalsing-pivotal, for your review.

The usage is like:
```
MockIntegration.mockMessageHandler(hasHeader("bar", "BAR"))
	.andReply()
	.expect(hasHeader("baz", "BAZ"))
	.expect(hasPayload("foo"))
	.andReply(m -> m.getPayload("X"));
```
Where we iterate over expectations as messages are arriving to the `MH`. The last one is used for the rest subsequent requests. Not all messages may produce reply. That's why you see that one more `.expect()` after previous one.

The `MockIntegrationContext` provides `instead()` utility to replace real `MH` with our mock.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -0,0 +1,103 @@
...
+		Matcher<Message<?>> matcher = this.lastKey;
+		Function<Message<?>, ?> replyFunction = this.lastReplyFunction;
+
+		synchronized (this) {
+			Iterator<Map.Entry<Matcher<Message<?>>, Function<Message<?>, ?>>> entryIterator = matchers.entrySet().iterator();
+			if (entryIterator.hasNext()) {
+				Map.Entry<Matcher<Message<?>>, Function<Message<?>, ?>> matcherFunctionEntry = entryIterator.next();
+				matcher = matcherFunctionEntry.getKey();
+				replyFunction = matcherFunctionEntry.getValue();
+				entryIterator.remove();
```
Toparvion: AFAIU, this makes the (mock) handler stateful. While it would be very useful for roughly defined test scenarios, there may be test cases where the handler mock is excepted to behave exactly the same way independently of invocations number and order. Is it possible to satisfy such requirement with current implementation?

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.messaging.Message;
+
+/**
+ * @author Artem Bilan
+ *
+ * @since 5.0
+ */
+public class MockMessageHandler extends AbstractReplyProducingMessageHandler {
```
Toparvion: The idea behind this component is excellent but the API is quite hard to understand (for me of course). We know that the readability is one of the first requirements to test code so I'd like to pay your attention to a couple of points to improve. In particular:

- it is not clear at the first glance that the `expect`/[`andReply`] methods chain composes a kind of one-time-use scenario for the handler mock. It may seem that they just describe different variations of conditions for unlimited matchings against incoming messages. To avoid misuse the purpose of methods should be clearly stated in JavaDoc at least;
- `andReply` methods name may confuse end user in case of trying to read the chain in a straightforward manner, for instance, the fragment `.expect(hasHeader("baz", "BAZ")).andReply()` may seem like expecting a reply (like in `FileWritingMessageHandler#setExpectReply` method). I think it would be better to rename it to `thenReply` with some additions (see further);
- no-arg version of `andReply()` method is not self-describable as it is not obvious that the method involves the `Function#identity` function. Perhaps, it would be better to rename to something like `thenReplyWithRequest`.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
Toparvion: Answering my own question I guess that it can be done with handler mock with single (and only) `andReply` method in the chain. But it still doesn't shade a light to specifying such behavior to a custom condition.

Toparvion (commented):
Thank you very much Artem for keeping me in touch!
It seems I don't yet understand the idea behind the feature as thoroughly as you do but still I've left a couple of thought on the matter. Hope they'll be helpful.

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -112,15 +120,31 @@ public void instead(String pollingAdapterId, MessageSource<?> mockMessageSource,
...
 	}
 
-	private void instead(String endpointId, Object mock, Class<?> endpointClass, String property,
+	public void instead(String consumerEndpointId, MessageHandler mockMessageHandler) {
+		instead(consumerEndpointId, mockMessageHandler, true);
+	}
+
+	public void instead(String consumerEndpointId, MessageHandler mockMessageHandler, boolean autoStartup) {
+		instead(consumerEndpointId, mockMessageHandler, IntegrationConsumer.class, "handler", autoStartup);
+		if (mockMessageHandler instanceof MessageProducer) {
```
artembilan: Thank you, @Toparvion, for coming  back for review!

Let me try to answer to your concerns.

This PR is exactly about `MessageHandler` part which is possible in the middle/end of flow.
The Inbound Gateway is fully different story and can be covered via different approach.

So, here we deal with `MessageHandler` which is also a `MessageProducer`.
In Framework they are all of the `AbstractMessageProducingHandler` extensions.

So, this new `MockMessageHandler` is exactly for them, nothing more.

Since we **mock** here anyway that should be expectable that there is something mocking of our real components. There it should not be surprise when we tell `isntead("foo", mock)`, nothing is broken in the flow logic, but only we have a verification `MH` in the required place. So, I don't see reason (yet) to document the feature.

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -112,15 +120,31 @@ public void instead(String pollingAdapterId, MessageSource<?> mockMessageSource,
...
+			Object targetMessageHandler = this.beans.get(consumerEndpointId);
+			if (targetMessageHandler instanceof MessageProducer) {
+				MessageChannel outputChannel = TestUtils.getPropertyValue(targetMessageHandler, "outputChannel",
+						MessageChannel.class);
+				((MessageProducer) mockMessageHandler).setOutputChannel(outputChannel);
+			}
```
artembilan: Well, in most cases it really doesn't matter.
But, yeah... I'll consider to warn that it does not make sense to use `andReply()` if your target `MH` is fully unaware about returns at all. For example `FileTransferringMessageHandler`. But we can't change the `MockMessageHandler` type. For convenience and fluent API. Although we might can just wrap it into raw `MH` for this non-`MessageProducer` use-case! 😄

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
artembilan: All good feedback.
Will take into account.

Thank you very much for sharing vision!

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
artembilan: I made it such a stateful to align with standard Mockito API:

```
 /**
     * Use it for stubbing consecutive calls in {@link Mockito#doReturn(Object)} style.
     * <p>
     * See javadoc for {@link Mockito#doReturn(Object, Object...)}
     *
     * @param toBeReturned to be returned when the stubbed method is called
     * @param nextToBeReturned to be returned in consecutive calls when the stubbed method is called
     * @return stubber - to select a method for stubbing
     */
    @SuppressWarnings({"unchecked", "varargs"})
    Stubber doReturn(Object toBeReturned, Object... nextToBeReturned);
```

So, I don't try to make up something other since people are already familiar with that behavior.
As I said: the last `expect/andReply` is used for all the rest requests. That is exactly what Mockito does.

But yeah... I agree that renaming them to the `expectNext()` and `thenReply()` should be much cleaner API for purpose of this class.

artembilan:
@Toparvion ,

the next iteration is ready for review! 😄 

Very appreciate your feedback

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
Toparvion: That would be quite surprising for the user to find his non-`MessageProducer` wrapped into `MH` :smile: Plain warning looks more safe.

artembilan: Please, check the latest commit for this part.
Nothing to discuss on this line any more.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
Toparvion: Well, now that you've described the idea behind the solution from Mockito API's perspective, it looks much more logical to me. Yes, I definitely support your intention to make new API similar to Mockito's one as it would really ease its adoption. Probably it's my personal fall that I didn't catch that idea. Though I might be not alone.

artembilan: Well, I'm pretty sure that you'd be better with understanding if I added some JavaDocs at least.
But since this is still a PoC and ideas sharing I didn't waste time on the documentation part.
Just because that may be just dropped by reason of full rework of the solution. 😄 

Although it looks like with that in hands we might avoid such a discussion...

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
Toparvion: Well, yes, probably I've put the cart before the horse supposing that this solution should be somehow capable of dealing with other types of components such as Inbound Gateway. I think mocking of `MessageHandler`s will be more in-demand than other components including Inbound Gateways.

artembilan: Well, you always can `stop()` any Inbound Gateways or `MessageProducer`s.
Remember `@SpringIntegrationTest(noAutoStartup)`?
And provide an `@InboundChannelAdapter` for the same `requestChannel` and use `MockIntegration.mockMessageSource()` to produce messages for testing.

So, I don't feel like we should right now spend time thinking how useful to mock Inbound Gateways.

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
Toparvion: Yes, I've seen the changes, thanks.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
Toparvion: Never mind; it's absolutely OK to have no JavaDocs at the PoC stage. On the other hand, it's a good chance to examine the code for self-description.

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -125,14 +126,41 @@ public void instead(String consumerEndpointId, MessageHandler mockMessageHandler
...
-			Object targetMessageHandler = this.beans.get(consumerEndpointId);
 			if (targetMessageHandler instanceof MessageProducer) {
 				MessageChannel outputChannel = TestUtils.getPropertyValue(targetMessageHandler, "outputChannel",
 						MessageChannel.class);
 				((MessageProducer) mockMessageHandler).setOutputChannel(outputChannel);
 			}
+			else {
+				if (mockMessageHandler instanceof MockMessageHandler) {
+					if (!TestUtils.getPropertyValue(mockMessageHandler, "hasReplies", Boolean.class)) {
+						mockMessageHandler = mockMessageHandler::handleMessage;
```
Toparvion: Can't catch the idea behind this line - the method reference points to the method belonging to the same instance's functional interface. Usually this means that the assignment of the method reference can be replaced with a plain object reference (qualifier). But it this case that would turn into variable self-assignment. Can you please reveal the intention?

spring-integration-test/src/test/java/org/springframework/integration/test/mock/MockMessageHandlerTests.java
```diff
@@ -145,18 +198,29 @@ public PollableChannel results() {
...
-					.andReply()
-					.expect(hasHeader("baz", "BAZ"))
-					.andReply()
-					.expect(hasPayload("foo"))
-					.andReply();
+					.thenReply()
+					.assertNext(hasHeader("baz", "BAZ"))
+					.thenReply()
+					.assertNext(hasPayload("foo"))
+					.thenReply("foo");
```
Toparvion: Being expressed this way the chain looks much more readable. Thank you!

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
artembilan: 😄 Yep! I expected such a question. 
I guess what confuses here is that we assign lambda do the same variable.

The idea is here like do not replace raw `MH` with our full reply supporting `MockMessageHandler` do not not produce unexpected behavior in the target endpoint.

In other words: 
- Our `MockMessageHandler` is `MessageProducer`, too;
- The `mockMessageHandler::handleMessage` produces just `MessageHandler`;
- I reuse replace a variable with this new instance.

Without a lambda we have an inline `MessageHandler` imple which just delegates to the provided `mockMessageHandler`.

Although I think that is overhead. we really don't need to worry about the target type from the endpoint perspective - produce or not reply is not its responsibility.
If our `MockMessageHandler` is without replies we are good to replace the raw `MH` with it and don't worry about aftermath.

Will remove this expression in the next commit.

Toparvion: Well, you've just widened my Java perception horizons. :smile: I've checked the code in IDEA and it highlighted it as a warning. Your intention appears to be more thoughtful than Jetbrains guys might suppose.
Agree with you on the matter.

spring-integration-test-support/src/main/java/org/springframework/integration/test/matcher/HeaderMatcher.java
```diff
@@ -68,103 +68,99 @@
...
 		super();
 		this.matcher = matcher;
 	}
 
-	/**
-	 * {@inheritDoc}
-	 */
 	@Override
-	public boolean matchesSafely(Message<?> item) {
+	public boolean matchesSafely(Message item) {
```
garyrussell: Why `rawTypes` ??

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+import org.hamcrest.Matcher;
+
+import org.springframework.integration.handler.AbstractMessageProducingHandler;
+import org.springframework.messaging.Message;
+
+/**
+ * @author Artem Bilan
```
garyrussell: Javadocs

garyrussell (changes requested):

spring-integration-test-support/src/main/java/org/springframework/integration/test/matcher/HeaderMatcher.java
artembilan: Otherwise it doesn't compile with the `MockMessageHandler.assertNext()`.
I tried to fix `PayloadMatcher` to be generic-aware in the one of previous commits, but that fell to the issue with undesired casts in the target test code when we would like just deal with `Matcher`s.

Since I would like to make `MockMessageHandler.assertNext()` compatible with both `HeaderMatcher` and `PayloadMatcher`, plus `PayloadAndHeaderMatcher`, there is no way unless make everything generic-aware or raw types.

Any other ideas?

Thanks for pointing to this anyway!

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Override
+	protected void handleMessageInternal(Message<?> message) throws Exception {
+		Matcher<Message> matcher = this.lastKey;
+		Function<Message<?>, ?> replyFunction = this.lastReplyFunction;
+
+		synchronized (this) {
+			Iterator<Map.Entry<Matcher<Message>, Function<Message<?>, ?>>> entryIterator =
+					this.matchers.entrySet().iterator();
+			if (entryIterator.hasNext()) {
```
garyrussell: I wonder if it would be better to have the iterator in a field and not remove the matchers.

That way we could have a `reset()` method and the same config could be used multiple times in a test.

artembilan: Such an approach doesn't follow with Mockito way.
Even if they have `resent(mock)`, it isn't recommend for use and would be better to create a new mock object  for new test-case, even in the same test method:
```
* Instead of <code>#reset()</code> please consider writing simple, small and focused test methods over lengthy, over-specified tests.
* <b>First potential code smell is <code>reset()</code> in the middle of the test method.</b> This probably means you're testing too much.
* Follow the whisper of your test methods: "Please keep us small & focused on single behavior".
```
I don't want to encourage user to follow bad habits...

garyrussell: ok

garyrussell:
ok; just the javadocs then and some asciidoctor discussion.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockIntegration.java
```diff
@@ -85,12 +87,11 @@
 	 * @return the mocked {@link MessageSource}
 	 * @see Mockito#mock(Class)
 	 */
-	@SuppressWarnings("rawtypes")
 	public static MessageSource<?> mockMessageSource(Message<?> message) {
 		MessageSource messageSource = Mockito.mock(MessageSource.class);
 
-		given(messageSource.receive())
-				.<Message<?>>willReturn(message);
+		BDDMockito.given(messageSource.receive())
```
garyrussell: Why? We usually prefer static imports.

garyrussell:
We still need some reference manual updates.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockIntegration.java
artembilan: Right, we prefer but in `test` code.
This one is in `main`, so production code. Therefore no static imports.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockIntegration.java
```diff
@@ -105,18 +105,64 @@
...
 
 		return messageSource;
 	}
 
-	@SuppressWarnings("rawtypes")
-	public static MockMessageHandler.MockMessageHandlerWithReply mockMessageHandler(Matcher<Message> matcher) {
+	/**
+	 * Build a Mockito spy over the {@link MockMessageHandler} instance
+	 * based on the provided payload to assert incoming messages.
+	 * @param payload the to assert incoming messages
```
Toparvion: Missing `payload` word?

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockIntegration.java
```diff
@@ -105,18 +105,64 @@
...
+	 * @see MockMessageHandler
+	 */
+	public static MockMessageHandler.MockMessageHandlerWithReply mockMessageHandler(Object payload) {
+		return mockMessageHandler(PayloadMatcher.hasPayload(payload));
+	}
+
+	/**
+	 * Build a Mockito spy over the {@link MockMessageHandler} instance
+	 * based on the provided message to assert incoming messages.
+	 * @param message the to assert incoming messages
```
Toparvion: Missing `message` word?

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -26,19 +24,39 @@
...
+ * the similar behavior exists in the
+ * {@code Mockito.doReturn(Object toBeReturned, Object... toBeReturnedNext)}.
+ * <p>
+ * Typically is used as a chain of assertions and optional replies for them:
+ * <pre class="code">
+ * {@code
+ *      MockIntegration.mockMessageHandler(hasHeader("bar", "BAR"))
+ *               .thenReply()
+ *               .assertNext(new GenericMessage<>("foo", Collections.singletonMap("key", "value")))
+ *               .assertNext(hasPayload("foo"))
```
Toparvion: The fact that absence of `thenReply()` after `assertNext()` supposes producing no reply (and not proceeding to next `assertNext()`) may be not self-evident and hence probably deserves explicit statement.

artembilan: I think you (and all other) will get used to this kind of API.
Maybe we should come back to the `andReply()` ?

It becomes much obvious that *and* is tied with the current *next*.

If we don't have *and* the means that we just go to the *next*. And of course when that next happens.

artembilan: I think *then* has a comprehensive meaning if we have a good outcome from the assertion.

I don't know what better can we have here, but I feel comfortable with the current API...

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockIntegration.java
```diff
@@ -105,18 +105,64 @@
 	 * @return the mocked {@link MessageSource}
 	 * @see Mockito#mock(Class)
 	 */
-	@SuppressWarnings("rawtypes")
 	public static MessageSource<?> mockMessageSource(Message<?> message, Message<?>... messages) {
 		MessageSource messageSource = Mockito.mock(MessageSource.class);
```
garyrussell: Needs `<?>`  or suppression `rawtypes`

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -79,18 +135,34 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
...
+		MockMessageHandlerWithReply(Matcher<Message<?>> nextMessageMatcher) {
 			assertNext(nextMessageMatcher);
 		}
 
+		/**
+		 * Add the {@link Function#identity()} reply to the current assertion.
+		 * Produces the {@code requestMessage} as a reply.
+		 * @return this
+		 */
 		public MockMessageHandler thenReply() {
```
garyrussell: Maybe `thenReplySame()` ?

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
garyrussell: I am a little torn here; I tend to agree to @Toparvion that it's a bit confusing but I can it's useful, especially with `thenReply(m -> ...)` since you will likely generate the replies based on the request.

I am not sure there's much value in `thenReply()` echoing back the same message so I wouldn't mind it being removed altogether - it's name is a bit blah too (see comment elsewhere in this review).

I also don't really like the first matcher being the factory method argument.

How about...

    MockIntegration.mockMessageHandler()
             .replyFunction(m -> ...)
             .assertNext(...)
             .assertNext(...)
             .assertNext(...).withReply(m -> ...)
             .assertNext(...)

or...

    MockIntegration.mockMessageHandler()
             .replyFunction(m -> ...)
             .assertNext(...)
             .assertNext(...)
             .replyFunction(m -> ...)
             .assertNext(...)
             .assertNext(...)

garyrussell (changes requested):

artembilan:
OK, guys!

Since you both feel a bit uncomfortable with the API, I decided to re-think the approach at all.

The main reason here is like this `MockMessageHandler` becomes a part of the original flow instead of some real component (e.g. HTTP or JDBC gateway), so actually we might like do not just assert incoming messages, but do some other logic. For example in case of mocking aggregator we would like to collect messages until decide to release the group or not.

So, my current idea is like:

```
MockIntegration.mockMessageHandler()
             .handleNext(Consumer<Message<?>>)
             .handleNext(Consumer<Message<?>>)
             .handleNextAndReply(Function<Message<?>, Object>)
             .handleNext(Consumer<Message<?>>)
             .handleNextAndReply(Function<Message<?>, Object>)
             .handleNextAndReply(Function<Message<?>, Object>);             
```

If you wish to assert those messages that may be as a part of that `Consumer/Function` implementation.

Only the problem we should keep in mind that our flow can be *async* and started from the Message Driver/ Polling Consumer Channel Adapter. Therefore all our assertion errors would be a noise in those threads and we would bring some hooks to catch them in the rest of our real flow.

For this purpose I suggest to path to the `mockMessageHandler()` regular Mockito `ArgumentCaptor<Message<?>>`:
```
* Use it to capture argument values for further assertions.
```

With its `getAllValues()` we get access to all the passed messages and can assert them any possible way.

Need to come up with the async await though, but that might be different story. I'm afraid in most cases we would like to pass messages not only through this `MockMessageHandler` but to some other downstream components. And come to the verification only after some particular signal from the flow. Or... file on the file system via `WatchService` if our flow ends up with `FileWritingMessageHandler`.

WDYT now?

Thank you for the feedback though! Very useful to work in a team with thorough minds 😄

Toparvion:
This approach makes the API a bit more complicated from the verification perspective but OTOH it effectively covers both assertion and replying sides of mock handler in the same manner. I like it 👍 
I think using `ArgumentCaptor` within `Consumer` or `Function` definitely deserves explicit example in JavaDoc as it is not the most wide-spread feature of Mockito and often causes misunderstanding. 

The only thing keeping me doubt is that with this approach we deprive the user from that simplicity of `assertThat()` method where he might include assertion arguments in-place. This would be very handy for simple synchronous `MessageHandler`s which are (I think) majority. It would be great if we could save a couple of methods in this form though it logically conflicts with the new approach.

I also would like to support Garry's suggestion to assign special name to no-arg `...Reply()` method (for example `...ReplySame()`) as its name is not self-descriptive.

artembilan:
Thank you, @Toparvion , for sharing your thoughts as usual!

One more time: there is no guarantee that all the flows will be *sync*. So, we can't make an assumption that our simple `Matcher`-based API won't trap end-user in the unhandled and unthrown exception from the flow. That's why I retract from the previous approach in favor of callbacks with much flexible signature.
The current solution you reach pretty easy:
```
handleNext(m -> assertThat(m, hasHeader("baz", "BAZ"))
``` 
But at the same time you have a flexibility to apply as much assertion to the same message as you want or need. The current one will require from you come up with the custom complex `Matcher` impl.

With this idea you can do like this in the context of the current message:
```
handleNext(m -> {
     assertThat(m, hasHeader("baz", "BAZ");
     assertThat(m, hasPayload("foo");
})
``` 
which looks more natural for unit tests coders.

Anyway I'd recommend to go to the `ArgumentCaptor` phase for proper test failure from verification.

Of course with the new approach we don't have a  no-arg `...Reply()` discussion any more 😉

Toparvion:
Thank you, @artembilan, for clarification! And sorry for my misleading proposals; they came from inadequate understanding. Now with these examples on hand it became much more clear to me.

I think this new approach deserves to be implemented and included into Spring Integration's Testing Framework.

artembilan:
Some external use-case on the matter: http://stackoverflow.com/questions/44035327/unit-test-spring-integration-flow-dsl

Looks like our solution with the `instead()` fully fits that requirements.

Toparvion:
@artembilan, yes, looks like that.
> With the new `MockIntegrationContext` features the Framework will take care about that stuff for you

Maybe it's worth to denote some date and/or version when the feature is expected to be released for the user?

artembilan:
>  when the feature is expected to be released for the user?

That isn't related to the question. And user does his current task and needs solution right now.
If he asks about the date, we will get him pointers. 

The current situation that we are going to release `5.0 GA` somewhere just after Spring Framework `5.0 GA`.
Currently it is somewhere in the end of June this year.

garyrussell (approved):

garyrussell:
Minor doc polishing for your approval: https://github.com/garyrussell/spring-integration/commit/e4bc5412d36fc883741d12bd215e675f78be1d79

garyrussell:
Merged as 079ccb84e21deabeb3867e97ed062ea243123458

PR #2150, opened 2017-05-12
INT-4272: Read Only Header Improvements

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4272

- Honor read only header configuration when building from an unchanged message
- Support adding read only headers to the message builder factory

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationMessageHeaderAccessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationMessageHeaderAccessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2016 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


artembilan:
LGTM
Merging...

Meanwhile I wonder why we don't honor read only headers dealing with `MutableMessage`.
WDYT?

spring-integration-core/src/main/java/org/springframework/integration/IntegrationMessageHeaderAccessor.java
```diff
@@ -157,4 +157,29 @@ protected boolean isReadOnly(String headerName) {
 		return super.isReadOnly(headerName) || this.readOnlyHeaders.contains(headerName);
 	}
 
+	public boolean containsReadOnly(MessageHeaders headers) {
```
artembilan: M-m-m. Looking to the usage of this method in the `MessageBuilder` I think we don't need it.
It is a bit confusing as public API.
The `IntegrationMessageHeaderAccessor` represents some volatile object around particular headers.
That smells a bit to check the state of this object against arbitrary external `MessageHeaders` instance.

WDYT?

I can fix on merge

garyrussell: How do you propose accessing the readOnlyHeaders from the builder?

artembilan:
```
public MessageBuilder<T> readOnlyHeaders(String... readOnlyHeaders) {
	this.headerAccessor.setReadOnlyHeaders(readOnlyHeaders);
	return this;
}
```

garyrussell: Ah - you mean capture them in the builder? OK.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f4a17886be2db62b323ea802f02fadb26c0443af and cherry-picked

PR #2151, opened 2017-05-13
Test Mail Server Add CC: and BCC: Headers

garyrussell:
Allow assertions of proper population in tests.

Also set Github tab size to 4.


artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4bf266be50dd4119b01da9834a86900c140a4820

PR #2152, opened 2017-05-16
INT-4274: Improve IntegrationFlow Lifecycle

artembilan:
JIRA: https://jira.spring.io/browse/INT-4274

Since any component from the `StandardIntegrationFlow` can be started
independently, the `StandardIntegrationFlow.stop()` should propagate
the stop control to the target components independently of its previous
`running` state

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +21,7 @@ package org.springframework.integration.dsl;
  * <p>
  * The {@link StandardIntegrationFlow} implementation (produced by {@link IntegrationFlowBuilder})
  * represents a container for the integration components, which will be registered
- * in the application context. Typically is used as {@code &#64;Bean} definition:
+ * in the application context. Typically is used as a {@code @Bean} definition:
  * <pre class="code">
  *  &#64;Bean
  *  public IntegrationFlow fileReadingFlow() {
@@ -73,6 +73,11 @@ package org.springframework.integration.dsl;
 @FunctionalInterface
 public interface IntegrationFlow {
 
+	/**
+	 * The callback-based function to declare the chain of EIP-methods to
+	 * configure an integration flow with the provided {@link IntegrationFlowDefinition}.
+	 * @param flow the {@link IntegrationFlowDefinition} to configure
+	 */
 	void configure(IntegrationFlowDefinition<?> flow);
 
 }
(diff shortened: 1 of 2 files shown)


artembilan:
@garyrussell ,

not sure what you mean with the request in JIRA, but this is the best we can do to still have it as a `@FunctionalInterface`.

Going farther I'd say that we have too many similar places (`AbstractEndpoint` ?) where we definitely are `Lifecycle` as well. And that was always fine to cast whenever it is necessary. 

Why this `IntegrationFlow` use-case is so specific that we should provide some ugly solution just to let end-user to avoid to cast?

Thanks

garyrussell:
When a JIRA title contains "Consider" then rejection is an acceptable answer.

However, in this case, I think it's important because the flow can control the lifecycle of the component beans.

In fact, when using dynamic registration I think, perhaps, that if `autoStartup=false` on the registration, it should be propagated to the component beans.

We really should document that the flow should be considered a single lifecyle and manipulating the lifecycles of the component beans manually may cause unexpected results (such as encountered in the referenced SO question).

garyrussell:
Polishing: https://github.com/garyrussell/spring-integration/commit/f26ada6a787798dda08e12bde7a66133b03ad86c for your review.

artembilan:
What makes me nervous about this discussion, that when we have a Lambda `IntegrationFlow` it is replaced by the `StandardIntegrationFlow` in the `IntegrationFlowBeanPostProcessor`. So, you original `IntegrationFlow` instance isn't relevant and you really should deal with lifecycle only via `beanFactory.getBean(flowId)` or via `IntegrationFlowContext.getRegistrationById()`.

I think exactly that we should document in that JavaDoc and that's all. Nothing else should be exposed on the `IntegrationFlow` interface.

The expression " the flow will not be aware that they are running and they won't be stopped if/when the registration is removed from the flow context." is wrong. See `StandardIntegrationFlow.stop()`. So, all related components for the flow are stopped as well.

I think only the problem with that SO question that `destroy()` isn't implemented in that version yet.

And I don't think that we encourage people to always treat the `IntegrationFlow` as a single component. We don't do that for regular (without DSL) flow definition and we don't warn that there might be something unexpected if a middle component is started manually. 

We really should treat `IntegrationFlow` as a logical component with some conveniences over its `IntegrationFlowRegistration` when we decided to control everything related to this `IntegrationFlow` in one place.

IMO I'd do nothing about this story since everything seems working when we have `destroy()` management on board already...

garyrussell:
>See StandardIntegrationFlow.stop()

	public void stop() {
		if (this.running) {
                     ...

The problem is the poster didn't start the flow, he started the adapter. When he unregistered the flow, the adapter continued running.

If you have other suggestions, I am all ears.

Perhaps we could remove that "isRunning"? - unconditionally stop components on `remove()`.

artembilan:
> Perhaps we could remove that test on running?

I think that really is the way to go. Since we can't be sure in all those related component we should just call their `stop()` unconditionally.

garyrussell:
Works for me; please revert this and push a new commit (I will update the JIRA).

artembilan:
Pushed the way we discussed.

garyrussell:
A bit of polishing for your review.

garyrussell:
https://github.com/garyrussell/spring-integration/commit/82c2981d5dc13fd2634086f23e0a3917af1bd1dd

garyrussell:
Merged as be98a6c2d0a89b57d9a328b0a3cbb722152cc485

PR #2154, opened 2017-05-16
INT-4252 IntegrationFlow: allow custom bean names

artembilan:
JIRA: https://jira.spring.io/browse/INT-4252

To allow to provide arbitrary bean names for the intermediate components
in the `IntegrationFlow` change `ComponentsRegistration` to return
a `Map<Object, String>` instead of raw `Collection<Object>`

* Use the value from that map in the `IntegrationFlowBeanPostProcessor` as a
fallback option before walking into bean name generation
* Provide `containerId` option for the AMQP DSL components

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpInboundGatewaySpec.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpInboundGatewaySpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2015 the original author or authors.
+ * Copyright 2014-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,8 +16,9 @@
 
 package org.springframework.integration.amqp.dsl;
 
-import java.util.Collection;
 import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.Executor;
 
 import org.springframework.amqp.core.AmqpTemplate;
 import org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer;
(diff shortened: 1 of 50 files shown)


spring-integration-amqp/src/test/java/org/springframework/integration/amqp/dsl/AmqpTests.java
```diff
@@ -82,11 +85,9 @@
 	@Autowired
 	private AmqpInboundGateway amqpInboundGateway;
 
-	@AfterClass
-	public static void tearDown() {
-		brokerRunning.removeTestQueues("amqpOutboundInput", "amqpReplyChannel", "asyncReplies", "defaultReplyTo",
```
garyrussell: We still need this (but with no `additionalQueues` args).

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpInboundGatewaySpec.java
```diff
@@ -270,10 +272,19 @@ public AmqpInboundGatewaySpec defaultRequeueRejected(boolean defaultRequeueRejec
 		return this;
 	}
 
-	@Override
-	public Collection<Object> getComponentsToRegister() {
-		return Collections.<Object>singleton(this.listenerContainer);
+	/**
+	 * @param containerId the bean name for internal listener container instance.
```
garyrussell: /* Set the bean name of the listener container if it is created internally or a non-bean
    /* object was supplied. Ignored if the container is already a bean.

garyrussell: I think we need this on the Jms specs that generate a container too.

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
Merged as ea89682368a3b7946f5378fbbfba82e692894021

PR #2157, opened 2017-05-17
INT-4060: FTP Gateway: Add `NLST` and `workDir`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4060

* Add `NLST` command to the `AbstractRemoteFileOutboundGateway` to perform
`listNames` on the target session.
Useful in case of server doesn't allow to perform `LS` or the names set is
sufficient for application requirements
* Add `workingDirExpression` to the `FtpOutboundGateway` to allow to perform
`FtpClient.changeWorkingDirectory()` based on the current request message
* Change `slf4j-log4j12` to the `testCompile` -
the FTP tests fail in the IDE with `ClassNotFoundException`

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
@@ -23,6 +23,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -55,6 +56,7 @@ import org.springframework.util.StringUtils;
  * @author David Turanski
  * @author Gary Russell
  * @author Artem Bilan
+ *
  * @since 3.0
  *
  */
@@ -66,6 +68,13 @@ public class RemoteFileTemplate<F> implements RemoteFileOperations<F>, Initializ
 	 * the {@link SessionFactory} for acquiring remote file Sessions.
 	 */
 	protected final SessionFactory<F> sessionFactory;
+
+	/*
+	 * Not static as normal since we want this TL to be scoped within the template instance.
+	 */
+	private final ThreadLocal<Session<F>> contextSessions = new ThreadLocal<>();
+
+	private final AtomicInteger activeTemplateCallbacks = new AtomicInteger();
 
 	private volatile String temporaryFileSuffix = ".writing";
 
@@ -398,15 +407,25 @@ public class RemoteFileTemplate<F> implements RemoteFileOperations<F>, Initializ
 
 	@Override
 	public Session<F> getSession() {
-		return this.sessionFactory.getSession();
+		if (this.activeTemplateCallbacks.get() > 0) {
+			return this.contextSessions.get();
+		}
+		else {
+			return this.sessionFactory.getSession();
+		}
 	}
 
 	@SuppressWarnings("rawtypes")
 	@Override
 	public <T> T execute(SessionCallback<F, T> callback) {
 		Session<F> session = null;
+		if (this.activeTemplateCallbacks.get() > 0) {
+			session = this.contextSessions.get();
+		}
 		try {
-			session = this.sessionFactory.getSession();
+			if (session == null) {
+				session = this.sessionFactory.getSession();
+			}
 			Assert.notNull(session, "failed to acquire a Session");
 			return callback.doInSession(session);
 		}
@@ -420,7 +439,7 @@ public class RemoteFileTemplate<F> implements RemoteFileOperations<F>, Initializ
 			throw new MessagingException("Failed to execute on session", e);
 		}
 		finally {
-			if (session != null) {
+			if (this.activeTemplateCallbacks.get() == 0 && session != null) {
 				try {
 					session.close();
 				}
@@ -429,6 +448,28 @@ public class RemoteFileTemplate<F> implements RemoteFileOperations<F>, Initializ
 						this.logger.debug("failed to close Session", ignored);
 					}
 				}
+			}
+		}
+	}
+
+	@Override
+	public <T> T invoke(OperationsCallback<F, T> action) {
+		Session<F> contextSession = this.contextSessions.get();
+		if (contextSession == null) {
+			this.contextSessions.set(this.sessionFactory.getSession());
+		}
+		this.activeTemplateCallbacks.incrementAndGet();
+		try {
+			return action.doInOperations(this);
+		}
+		finally {
+			this.activeTemplateCallbacks.decrementAndGet();
+			if (contextSession == null) {
+				Session<F> session = this.contextSessions.get();
+				if (session != null) {
+					session.close();
+				}
+				this.contextSessions.remove();
 			}
 		}
 	}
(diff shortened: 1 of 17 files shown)


artembilan:
If you are fine with the solution, TODO:
* add `nlst` to the XSD
* wrap more commands to the `doInWorkingDirectory()` in the `FtpOutboundGateway`
* Documentation

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/outbound/FtpServerOutboundTests-context.xml
```diff
@@ -178,8 +178,8 @@
 
 	<int-ftp:outbound-gateway session-factory="ftpSessionFactory"
 							  request-channel="inboundLs"
-							  command="ls"
-							  command-options="-1"
+							  command="nlst"
+							  working-dir-expression="'ftpSource'"
 							  reply-channel="output"/>
```
garyrussell: We should really test both forms.

garyrussell (changes requested):

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/gateway/FtpOutboundGateway.java
```diff
@@ -151,9 +185,75 @@ protected long getModified(FTPFile file) {
...
+	private <V> V doInWorkingDirectory(Message<?> message, Session<FTPFile> session, Callable<V> task)
+			throws IOException {
+		Expression workingDirExpression = this.workingDirExpression;
+		FTPClient ftpClient = (FTPClient) session.getClientInstance();
+		String currentWorkingDirectory = null;
+		try {
+			if (workingDirExpression != null) {
+				currentWorkingDirectory = ftpClient.printWorkingDirectory();
+				String workingDir = workingDirExpression.getValue(this.evaluationContext, message, String.class);
+				ftpClient.changeWorkingDirectory(workingDir);
```
artembilan: `null` is fine for FTP: http://docs.spring.io/spring-integration/reference/html/ftp.html#ftp-inbound :
> Starting with version 4.3, the `remote-directory/remote-directory-expression` attributes can be omitted assuming `null`. In this case, according to the FTP protocol, the Client working directory is used as a default remote directory.

artembilan:
Still need documentation.

There is a question, though, regarding `PUT` and `MPUT`.
They are based on the hard work in the `RemoteFileTemplate.send()`.
What is bad there that each iteration in the `doMput()` may perform that `send` on a new `Session` (and new Client respectively).
How about to add `ThreadLocal<Session<F>>`  trick to the `RemoteFileTemplate` let it to reuse an instance for the current invocation? Plus with that we will be able to insert `workingDirectory`.
WDYT?

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/gateway/FtpOutboundGateway.java
garyrussell: True, but if `null`, we can avoid 2 round-trips to the server (change and change back noops).

garyrussell:
I assume you mean similar to the `RabbitOperations.invoke()` - if so, yes, I agree that would be a good pattern for those methods.

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/gateway/FtpOutboundGateway.java
artembilan: Well, we can't assume that user uses some other working dir by default and decided for some messages to switch to his/her root. And, of course, then come back after the operation. Remember `AbstractFtpSessionFactory.postProcessClientBeforeConnect/postProcessClientAfterConnect` ?

> The working directory can be set via the FTPClient.changeWorkingDirectory() function when you extend the DefaultFtpSessionFactory and implement postProcessClientAfterConnect() callback.

artembilan:
M-m-m. No, `MGET` is fine. It is already based on the dedicated `session`:
```
List<File> payload = this.remoteFileTemplate.execute(session ->
				mGet(requestMessage, session, remoteDir, remoteFilename));
```

artembilan:
Still need general doc for the new API, but there is already a plenty of changes to review.
Will document tomorrow.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
```diff
@@ -398,15 +407,25 @@ public boolean get(final String remotePath, final InputStreamCallback callback)
 
 	@Override
 	public Session<F> getSession() {
-		return this.sessionFactory.getSession();
+		if (this.activeTemplateCallbacks.get() > 0) {
+			return this.contextSessions.get();
```
garyrussell: Need to check for null.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
```diff
@@ -420,7 +439,7 @@ public boolean get(final String remotePath, final InputStreamCallback callback)
 			throw new MessagingException("Failed to execute on session", e);
 		}
 		finally {
-			if (session != null) {
+			if (this.activeTemplateCallbacks.get() == 0 && session != null) {
```
garyrussell: This isn't right - it might be another thread that's running in an invoke.

See `invokeScope` variable in rabbit template's doExecute

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
```diff
@@ -434,6 +453,28 @@ public boolean get(final String remotePath, final InputStreamCallback callback)
 	}
 
 	@Override
+	public <T> T invoke(OperationsCallback<F, T> action) {
+		Session<F> contextSession = this.contextSessions.get();
+		if (contextSession == null) {
+			this.contextSessions.set(this.sessionFactory.getSession());
+		}
+		this.activeTemplateCallbacks.incrementAndGet();
```
garyrussell: We should disallow nested invokes - see `RabbitTemplate`.

garyrussell (changes requested):
That's all for this round.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
artembilan: Why? I don't see any problems here with reusing already stored in the `ThreadLocal` session. 
I might agree about your comment in the `finally` block: have had similar doubts. But then decided to leave it somehow... Let me recall what was my conclusion.

We might consider to use additional `ThreadLocal` counter to track the recursion though...

garyrussell: I just think it adds unnecessary complications (recursive calls to invoke).

garyrussell: The `activeTemplateCallbacks` is simply an optimization to avoid accessing the thread local if no invokes are in process.

artembilan: I leave the logic in this method as is. That isn't so hard to track recursion here. Plus check our `MPUT` impl, especially when it is recursive as well. In the end it call `doPut()` for each file, which is based on the `invoke()` as well, since we have there not only `send()` operation, but `doChmod()`, too.

garyrussell (approved):

garyrussell:
Doc polishing for your approval https://github.com/garyrussell/spring-integration/commit/cd1f6809f59dbf46b46e7897ccbfaaded476cf02

garyrussell:
Merged as 3daeaab1b4b20d868a8b07d6b0813023a492e56c

PR #2158, opened 2017-05-18
INT-4278: DSL: AMQP DirectMessageListenerContainer

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4278

Support configuration of inbound endpoints using `DirectMessageListenerContainer`.

**Needs migration guide update**

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AbstractMessageListenerContainerSpec.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AbstractMessageListenerContainerSpec.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.amqp.dsl;
+
+import java.util.concurrent.Executor;
+
+import org.aopalliance.aop.Advice;
+
+import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.amqp.core.Queue;
+import org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.util.ErrorHandler;
+
+/**
+ * Base class for container specs.
+ *
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public abstract class AbstractMessageListenerContainerSpec<S extends AbstractMessageListenerContainerSpec<S, C>,
+		C extends AbstractMessageListenerContainer> {
+
+	private final C listenerContainer;
+
+	public AbstractMessageListenerContainerSpec(C listenerContainer) {
+		this.listenerContainer = listenerContainer;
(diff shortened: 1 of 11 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AbstractMessageListenerContainerSpec.java
```diff
@@ -0,0 +1,181 @@
...
+import org.springframework.util.ErrorHandler;
+
+/**
+ * Base class for container specs.
+ *
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public abstract class AbstractMessageListenerContainerSpec<S extends AbstractMessageListenerContainerSpec<S, C>,
```
artembilan: How about to make a `public id(String)` something similar what I did for JMS: https://github.com/spring-projects/spring-integration/pull/2154/files#diff-255bb1523904c44132f606cd30058739R48

Why this isn't `IntegrationComponentSpec` though?..

artembilan (changes requested):
More review tomorrow.
If we go your solution, we won't need that my odd `contianerId()`: https://github.com/spring-projects/spring-integration/pull/2154/files#diff-5fb038628f2989d0001cee1fde558370R268

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AbstractMessageListenerContainerSpec.java
```diff
@@ -0,0 +1,181 @@
...
+		C extends AbstractMessageListenerContainer> {
+
+	private final C listenerContainer;
```
artembilan: I see there are more `AbstractMessageListenerContainer` setters not exposed by the Spec: `setDeBatchingEnabled`, `setAfterReceivePostProcessors`, `setMessageConverter`, `setExposeListenerChannel`.
Or those don't make sense for SI?

WDYT?

artembilan (changes requested):
That's all.
I can fix on merge if that.
... or share the polishing commit if you are not going to fix yourself.
Thanks

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AbstractMessageListenerContainerSpec.java
garyrussell: Perhaps these were added after you build the DSL? Although not `setExposeListenerChannel` - that's been there a long time. So I am not sure why it's not there.

I suppose we could add them (not `setMessageConverter` though - that's just an (ugly) mechanism for getting the converter configured for `@RabbitListener`.

Please do the polishing - it was not obvious to me this should be an ICS - in fact, now I look at it, I am not sure why that's a factory bean, since the specs aren't registered as beans (unless I am missing something).

artembilan: Here is a polishing we discussed: https://github.com/artembilan/spring-integration/commit/38d84c355d1f8c720f332909b6d2fde25a42cafc

Yes, they can be registered as beans:
```
@Bean
MessageChannelSpec myChannel() {
     return MessageChannels.direct();
}
The most popular coding.

With the `FactoryBean` on board we can inject them regular way:
```
@Bean
IntegrationFlow myFlow(MessageChannel myChannel)
```
See `IntegrationFlowBeanPostProcessor.processIntegrationComponentSpec()` for more info.

garyrussell: Ah; yes of course; cool.

garyrussell: WOW - a lot of missing properties - good stuff - thanks for the polish.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/3cf85afe1f9f11e56310871da564d3b78e1eee42

PR #2159, opened 2017-05-21
INT-4255: JSON Embedded Headers Message Mapper

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4255

- Support embedding headers for transports that don't support headers (TCP, Kafka, etc)
- Use the new message-aware Jackson ObjectMapper
- Provide a mechanism to more efficiently support byte[] payloads (avoid Base64 encoding)
- Support decoding "legacy" SCSt embedded headers

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
@@ -1,0 +1,255 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.json;
+
+import java.io.UnsupportedEncodingException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.regex.Pattern;
+
+import org.springframework.integration.mapping.BytesMessageMapper;
+import org.springframework.integration.support.MutableMessage;
+import org.springframework.integration.support.MutableMessageHeaders;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.support.GenericMessage;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * For outbound messages, uses a message-aware Jackson object mapper to render the message
+ * as JSON. For messages with {@code byte[]} payloads, if rendered as JSON, Jackson
+ * performs Base64 conversion on the bytes. If the {@link #setRawBytes(boolean) rawBytes}
+ * property is true (default), the result has the form
+ * &lt;headersLen&gt;&lt;headers&gt;&lt;payloadLen&gt;&lt;payload&gt;; with the headers
+ * rendered in JSON and the payload unchanged.
+ * <p>
+ * By default, all headers are included; you can provide regular expressions to specify a
+ * subset of headers.
+ * <p>
+ * For inbound messages, in addition to pure JSON and the above format, the Spring Cloud
+ * Stream {@code EmbeddedHeadersMessageConverter} format is also supported:
+ * {@code 0xff, n(1), [ [lenHdr(1), hdr, lenValue(4), value] ... ]}. The 0xff indicates
+ * this format; n is the number of headers (max 255); for each header, the name length (1
+ * byte) is followed by the name, followed by the value length (int) followed by the value
+ * (json).
+ * <p>
+ * If neither special format is detected, or an error occurs during conversion, the
+ * payload of the message is the original {@code byte[]}.
+ * <p>
+ * <b>IMPORTANT</b>
+ * <p>
+ * The default object mapper uses default typing. From the Jacskson Javadocs:
+ * <p>
+ * NOTE: use of Default Typing can be a potential security risk if incoming content comes
+ * from untrusted sources, and it is recommended that this is either not done, or, if
+ * enabled, use {@link ObjectMapper#setDefaultTyping} passing a custom
+ * {@link com.fasterxml.jackson.databind.jsontype.TypeResolverBuilder} implementation that
+ * white-lists legal types to use.
+ * <p>
+ * A constructor is provided allowing the provision of such a configured object mapper.
+ *
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public class EmbeddedJsonHeadersMessageMapper implements BytesMessageMapper {
+
+	private final ObjectMapper objectMapper;
+
+	private final List<Pattern> headerPatterns;
+
+	private final boolean allHeaders;
+
+	private boolean rawBytes = true;
+
+	/**
+	 * Construct an instance that embeds all headers, using the default
+	 * JSON Object mapper.
+	 */
+	public EmbeddedJsonHeadersMessageMapper() {
+		this(Collections.singleton(Pattern.compile(".*")));
+	}
+
+	/**
+	 * Construct an instance that embeds headers matching the supplied patterns, using
+	 * the default JSON object mapper.
+	 * @param headerPatterns the patterns.
+	 */
+	public EmbeddedJsonHeadersMessageMapper(Collection<Pattern> headerPatterns) {
+		this(JacksonJsonUtils.messagingAwareMapper(), headerPatterns);
+	}
+
+	/**
+	 * Construct an instance that embeds headers matching the supplied patterns using the
+	 * supplied JSON object mapper.
+	 * @param objectMapper the object mapper.
+	 * @param headerPatterns the patterns.
+	 */
+	public EmbeddedJsonHeadersMessageMapper(ObjectMapper objectMapper, Collection<Pattern> headerPatterns) {
+		this.objectMapper = objectMapper;
+		this.headerPatterns = new ArrayList<>(headerPatterns);
+		this.allHeaders = this.headerPatterns.size() == 1 && this.headerPatterns.get(0).pattern().equals(".*");
+	}
+
+	/**
+	 * Set to false to encode byte[] payloads as base64 in JSON. When true,
+	 * byte[] payloads are rendered as discussed in the class description.
+	 * @param rawBytes false to encode as base64.
+	 */
+	public void setRawBytes(boolean rawBytes) {
+		this.rawBytes = rawBytes;
+	}
+
+	public Collection<Pattern> getHeaderPatterns() {
+		return Collections.unmodifiableCollection(this.headerPatterns);
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public byte[] fromMessage(Message<?> message) throws Exception {
+		Message<?> messageToEncode = this.allHeaders ? message : pruneHeaders(message);
+		if (this.rawBytes && message.getPayload() instanceof byte[]) {
+			return fromBytesPayload((Message<byte[]>) messageToEncode);
+		}
+		else {
+			return this.objectMapper.writeValueAsBytes(messageToEncode);
+		}
+	}
+
+	private Message<?> pruneHeaders(Message<?> message) {
+		Map<String, Object> headersToEmbed = new HashMap<>();
+		message.getHeaders().forEach((k, v) -> {
+			if (this.headerPatterns.stream().anyMatch(p -> p.matcher(k).matches())) {
+				headersToEmbed.put(k, v);
+			}
+		});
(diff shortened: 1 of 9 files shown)


garyrussell:
cc/ @mbogoevici 

FYI - it supports decoding "legacy" SCSt embedded headers - do you think we need an option to encode in that format too, for interop?

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -0,0 +1,255 @@
...
+		Map<String, Object> headersToEmbed = new HashMap<>();
+		message.getHeaders().forEach((k, v) -> {
+			if (this.headerPatterns.stream().anyMatch(p -> p.matcher(k).matches())) {
+				headersToEmbed.put(k, v);
+			}
+		});
+		return new MutableMessage<>(message.getPayload(), headersToEmbed);
+	}
+
+	private byte[] fromBytesPayload(Message<byte[]> message) throws UnsupportedEncodingException, Exception {
```
artembilan: Redundant `UnsupportedEncodingException`

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -0,0 +1,255 @@
...
+ * <p>
+ * The default object mapper uses default typing. From the Jacskson Javadocs:
+ * <p>
+ * NOTE: use of Default Typing can be a potential security risk if incoming content comes
+ * from untrusted sources, and it is recommended that this is either not done, or, if
+ * enabled, use {@link ObjectMapper#setDefaultTyping} passing a custom
+ * {@link com.fasterxml.jackson.databind.jsontype.TypeResolverBuilder} implementation that
+ * white-lists legal types to use.
+ * <p>
+ * A constructor is provided allowing the provision of such a configured object mapper.
```
artembilan: I wonder if we should go ahead and allow white-listing somehow for our `JacksonJsonUtils.messagingAwareMapper()`.

But I can't find the code in Jackson where we can inject the class filter...

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -0,0 +1,255 @@
...
+			return fromBytesPayload((Message<byte[]>) messageToEncode);
+		}
+		else {
+			return this.objectMapper.writeValueAsBytes(messageToEncode);
+		}
+	}
+
+	private Message<?> pruneHeaders(Message<?> message) {
+		Map<String, Object> headersToEmbed = new HashMap<>();
+		message.getHeaders().forEach((k, v) -> {
```
artembilan:
````java
Stream<Pattern> patternStream = this.headerPatterns.stream();
		
		Map<String, Object> headersToEmbed =
				message.getHeaders()
						.entrySet()
						.stream()
						.filter(e -> patternStream.anyMatch(p -> p.matcher(e.getKey()).matches()))
						.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
````
?

Also I would say the `this.headerPatterns.stream()` should be as a property, too. To avoid extra `Stream` object for each message.

**Review by artembilan — CHANGES_REQUESTED** [review]

**Comment by mbogoevici** [comment]
@garyrussell I believe we can drop interop support at this point. It was legacy even for Spring XD.

File: spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java [hunk 4]
```diff
@@ -0,0 +1,255 @@
...
+ * performs Base64 conversion on the bytes. If the {@link #setRawBytes(boolean) rawBytes}
+ * property is true (default), the result has the form
+ * &lt;headersLen&gt;&lt;headers&gt;&lt;payloadLen&gt;&lt;payload&gt;; with the headers
+ * rendered in JSON and the payload unchanged.
+ * <p>
+ * By default, all headers are included; you can provide regular expressions to specify a
+ * subset of headers.
+ * <p>
+ * For inbound messages, in addition to pure JSON and the above format, the Spring Cloud
+ * Stream {@code EmbeddedHeadersMessageConverter} format is also supported:
```
mbogoevici: We have moved the code from `EmbeddedHeadersMessageConverter` so that other binder implementations can use it. Also, you shouldn't need an instance. A strategy would be a good idea (I can see that being used in SCSt as well), but then the strategy shouldn't depend on an `ObjectMapper` (can be injected in the strategy object)

artembilan: Well, then it's not strategy, but delegate. The idea is to reuse this mapper as much as possible meanwhile allow to provide some custom aspect for legacy data. 
If we drop that support all together, there is no need in strategy discussion

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -0,0 +1,255 @@
...
+		}
+		if (message != null) {
+			return message;
+		}
+		else {
+			return new GenericMessage<>(bytes);
+		}
+	}
+
+	private Message<?> decodeNativeFormat(byte[] bytes) throws Exception {
```
mbogoevici: What is the native format?

artembilan: Means that one which belongs to SI per se.
This is SI precious 😄 

Well, in other places we usually use *default*, but Gary decided to go a bit further and just buy a patent to this format I guess 😉 

Seriously I don't know, but I'm fine with this because SI deals with many other formats and protocols, so that it pretty fine to introduce our own and name it **native**  from the SI perspective.

See description in the JavaDocs of this class BTW

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
garyrussell: My reading of the javadoc is you have to implement a `TypeResolverBuilder` - they don't seem to provide a white list filter out of the box, just the `DefaultTypeResolverBuilder`.

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -0,0 +1,255 @@
...
+	public EmbeddedJsonHeadersMessageMapper() {
+		this(Collections.singleton(Pattern.compile(".*")));
+	}
+
+	/**
+	 * Construct an instance that embeds headers matching the supplied patterns, using
+	 * the default JSON object mapper.
+	 * @param headerPatterns the patterns.
+	 */
+	public EmbeddedJsonHeadersMessageMapper(Collection<Pattern> headerPatterns) {
```
garyrussell: Yeah - I guess that's easier for java config.

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
garyrussell: I thought it would be important for SCSt to not force Base64 encoding of `byte[]` so I made this the default format for `byte[]` payloads.

I guess I could call it `decodeDefaultBytesFormat`.

mbogoevici: Ok, thanks - it surely makes sense for SI to have its own format. In fact, once we move to Kafka 0.11 we might phase out header embedding for the various varieties of Kafka binding (SI, reactive, KStream) - and any new binders can use the native SI format.

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
garyrussell: [INT-4290](https://jira.spring.io/browse/INT-4290).

garyrussell:
Closing for now; priority is reduced now that Kafka supports headers.

artembilan:
Well, this still have power for AWS Kinesis which doesn't have (yet) `headers` notation.

So, it definitely should be implemented on the core level, even if with just simple mapping pattern, without any strategies.

Thanks

garyrussell:
Agreed; also applies to tcp, MQTT etc. I just moved it to M6.

artembilan:
@garyrussell ,

maybe it's time to reopen this PR to be merged into the core finally?
Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
garyrussell: You can't reuse the patternStream like that; it will already be partially (or totally) consumed.

artembilan:
Gary, I don't tell that you've done something wrong, just wonder what is your opinion so far about existing `JsonOutboundMessageMapper` and `JsonInboundMessageMapper`?
Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -0,0 +1,218 @@
...
+	 */
+	public EmbeddedJsonHeadersMessageMapper(ObjectMapper objectMapper, Pattern... headerPatterns) {
+		this.objectMapper = objectMapper;
+		this.headerPatterns = Arrays.asList(headerPatterns);
+		this.allHeaders = this.headerPatterns.size() == 1 && this.headerPatterns.get(0).pattern().equals(".*");
+	}
+
+	/**
+	 * Set to false to encode byte[] payloads as base64 in JSON. When true,
+	 * byte[] payloads are rendered as discussed in the class description.
```
artembilan: IMO that's not JavaDoc language. I'm fine to have the copy of the sentence from the class level here.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpMessageMapper.java
```diff
@@ -208,6 +228,9 @@ protected final void addCustomHeaders(TcpConnection connection,
 
 	@Override
 	public Object fromMessage(Message<?> message) throws Exception {
+		if (this.bytesMessageMapper != null) {
+			return this.bytesMessageMapper.fromMessage(message);
+		}
 		if (this.stringToBytes) {
```
artembilan: We have to document somewhere that `stringToBytes` is ignored if `bytesMessageMapper` is provided. Or reconsider logic and don't use `bytesMessageMapper ` unconditionally...

artembilan (changes requested):

artembilan (changes requested):

artembilan:
Gary, I think I'm fully fine with the solution. Only concern is about friendly patterns configuration and I guess already Docs on the matter. 

Looking forward to get this into the account for Kinesis in SI-AWS-2.0 😄 !

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -0,0 +1,228 @@
...
+	public EmbeddedJsonHeadersMessageMapper() {
+		this(Pattern.compile(".*"));
+	}
+
+	/**
+	 * Construct an instance that embeds headers matching the supplied patterns, using
+	 * the default JSON object mapper.
+	 * @param headerPatterns the patterns.
+	 */
+	public EmbeddedJsonHeadersMessageMapper(Pattern... headerPatterns) {
```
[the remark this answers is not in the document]
garyrussell: Well; people might want to customize the compiled pattern, for example

    Pattern.compile("foo", Pattern.CASE_INSENSITIVE);

I suppose we could add 2 more CTORs with the `String... ` variant, but we can't avoid the `Pattern...` variant.

Feel free to add them when merging (or I can push another commit if you prefer).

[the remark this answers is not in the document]
artembilan: OK. Good catch about pattern customization!

WDYT about this variant otherwise:

How about to go the `PatternMatchUtils.simpleMatch()` instead of `Pattern` ?

I will merge, but when I know anything about Docs for this change.
Thanks

garyrussell:
Artem; I'll do the docs over the weekend, or Monday if we have time; otherwise they can wait until M8.

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonHeadersMessageMapper.java
```diff
@@ -135,7 +138,16 @@ public void setRawBytes(boolean rawBytes) {
 		this.rawBytes = rawBytes;
 	}
 
-	public Collection<Pattern> getHeaderPatterns() {
+	/**
+	 * Set to true to make the header name pattern match case sensitive.
+	 * Default false.
+	 * @param caseInsensitive true to make case sensitive.
```
artembilan: Merging after fixing this typo.
We definitely can address docs in the separate PR and even during RC-GA phase.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/73d9c6b6c5392b83ec70b73cfbe0bc0ecd53e7bf

PR #2162, opened 2017-05-22
INT-4256: AMQP: Conversion Errors to ErrorChannel

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4256

Also fix the JMS endpoint to use the `MessagingTemplate` instead of sending to the
error channel directly (ignored the send result).

__I suggest reviewing with `?w=1`__

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
```diff
@@ -518,6 +519,10 @@ public ErrorMessage buildErrorMessage(Throwable throwable) {
 			return super.buildErrorMessage(null, throwable);
 		}
 
+		MessagingTemplate getMessagingTemplate() {
```
artembilan: Just curious: what made you to add this method?

And if it is here already any way how about to make it `protected` ?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
```diff
@@ -78,15 +80,31 @@ public String getComponentType() {
 	@Override
 	protected void onInit() {
 		this.messageListenerContainer.setMessageListener((ChannelAwareMessageListener) (message, channel) -> {
```
artembilan: Not related to this change, but how about to extract the real class for this Lambda?
Something similar what we do in other places like Kafka or Kinesis. 

I feel like it would be much easier to consult logs in case of StackTraces.

artembilan (changes requested):

spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
garyrussell: Because the outer class cannot access the delegate's protected `messagingTemplate field.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b5bbb93fe8ead02d1e49401336b86ce961364527

PR #2163, opened 2017-05-22
INT-4166: ThreadAffinityClientConnectionFactory

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4166

Binds connections to threads.

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/ThreadAffinityClientConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/ThreadAffinityClientConnectionFactory.java
@@ -1,0 +1,379 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.ip.tcp.connection;
+
+import java.util.List;
+import java.util.concurrent.Executor;
+
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.core.convert.ConversionService;
+import org.springframework.core.serializer.Deserializer;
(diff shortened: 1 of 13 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/ThreadAffinityClientConnectionFactory.java
```diff
@@ -0,0 +1,379 @@
...
+import org.springframework.expression.Expression;
+import org.springframework.integration.support.MessageBuilderFactory;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.core.DestinationResolver;
+import org.springframework.util.Assert;
+
+/**
+ * A client connection factory that binds a connection to a thread. Close operations
+ * are ignored; to physically close a connection and release the thread local, invoke
+ * {@link #releaseConnection()).
```
artembilan:
```
/ThreadAffinityClientConnectionFactory.java:38: error: unterminated signature

 * {@link #releaseConnection()).
```
You can see that in the current Travis report.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/ThreadAffinityClientConnectionFactory.java
```diff
@@ -0,0 +1,379 @@
...
+
+	@Override
+	public TcpConnectionSupport getConnection() throws Exception {
+		TcpConnectionSupport connection = this.connections.get();
+		if (connection == null || !connection.isOpen()) {
+			connection = this.connectionFactory.getConnection();
+			connection.disableClose();
+			this.connections.set(connection);
+		}
+		return connection;
```
artembilan: Can't we instead of that peculiar `disable/enableClose` return here a wrapper with the empty `close()` impl and invoke `delegate.close()` in that `releaseConnection()` above?

Something similar we have in the Spring Kafka: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L108

src/reference/asciidoc/ip.adoc
```diff
@@ -364,6 +364,41 @@ Initially, the first factory in the configured list is used; if a connection sub
 
 NOTE: When using the failover connection factory, the singleUse property must be consistent between the factory itself and the list of factories it is configured to use.
 
+[[affinity-cf]]
```
artembilan: `tcp-affinity-cf` then?

Who knows what else we will introduce in the future...

artembilan (changes requested):
Not so much to worry about, but some comments any way

src/reference/asciidoc/ip.adoc
artembilan: Looks like you have missed this change in the `ip.adoc`, since I see one in the `whats-new`.
Will fix on merge though.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b84a33437975562c031ed26c69e817a8e42fb815

PR #2164, opened 2017-05-22
INT-4164: TCP Support Pushback and Subclassing

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4164

Previously, it was not possible to create a subclass of `TcpNetConnection`.
This was possible for `TcpNioConnection` via `TcpNioConnectionSupport`.

Add `TcpNetConnectionSupport` and a default implementation.

Create `AbstractTcpConnectionSupport` and add support for all implementation (Net, NIO, NIO with SSL)
to support wrapping the connection's `InputStream` in a `PushBackInputStream`.

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpNetConnectionSupport.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpNetConnectionSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpNetConnectionSupport.java
```diff
@@ -0,0 +1,79 @@
...
+
+	public static class PushBackTcpNetConnection extends TcpNetConnection {
+
+		private final int pushbackBufferSize;
+
+		private final String connectionId;
+
+		private volatile PushbackInputStream pushbackStream;
+
+		private volatile InputStream original;
```
artembilan: never_assigned

Something has been missed in the logic of this class.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpNetConnectionSupport.java
```diff
@@ -0,0 +1,79 @@
...
+		if (isPushbackCapable()) {
+			return new PushBackTcpNetConnection(socket, server, lookupHost, applicationEventPublisher,
+					connectionFactoryName, getPushbackBufferSize());
+		}
+		else {
+			return new TcpNetConnection(socket, server, lookupHost, applicationEventPublisher, connectionFactoryName);
+		}
+	}
+
+	public static class PushBackTcpNetConnection extends TcpNetConnection {
```
artembilan: Should this class really be `public`?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpNetConnectionSupport.java
```diff
@@ -0,0 +1,79 @@
...
+				ApplicationEventPublisher applicationEventPublisher, String connectionFactoryName, int bufferSize) {
+			super(socket, server, lookupHost, applicationEventPublisher, connectionFactoryName);
+			this.pushbackBufferSize = bufferSize;
+			this.connectionId = "pushback:" + super.getConnectionId();
+		}
+
+		@Override
+		protected InputStream inputStream() throws IOException {
+			if (this.pushbackStream == null || super.inputStream() != this.original) {
+				this.pushbackStream = new PushbackInputStream(super.inputStream(), this.pushbackBufferSize);
```
artembilan: Aren't there any chance that `super.inputStream()` **is** already `PushbackInputStream`, so, we don't need to wrap at all?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNetConnectionSupport.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.context.ApplicationEventPublisher;
+
+
+/**
+ * Used by NET connection factories to instantiate a {@link TcpNetConnection} object.
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public interface TcpNetConnectionSupport {
```
artembilan: `@FunctionalInterface` ?

artembilan (changes requested):
That's all from my side.

Cool stuff though!

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpNetConnectionSupport.java
garyrussell: No; it's not possible - either `socket.getInputStream()` or NIO `ChannelInputStream`.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4991d963e8b22887a9f7ade4bf0fdea278e0c969

PR #2165, opened 2017-05-23
INT-4117: File Outbound Set File Permissions

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4117

Add an option to set file permissions on the `FileWritingMessageHandler` when the file system supports it.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
@@ -29,10 +29,14 @@ import java.io.OutputStreamWriter;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.BitSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.locks.Lock;
 import java.util.regex.Matcher;
@@ -151,6 +155,8 @@ public class FileWritingMessageHandler extends AbstractReplyProducingMessageHand
 	private volatile MessageFlushPredicate flushPredicate = new DefaultFlushPredicate();
 
 	private volatile boolean preserveTimestamp;
+
+	private Set<PosixFilePermission> permissions;
 
 	/**
 	 * Constructor which sets the {@link #destinationDirectoryExpression} using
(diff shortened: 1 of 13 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -745,6 +807,20 @@ private void cleanUpAfterCopy(File fileToWriteTo, File resultFile, File original
...
+
+	/**
+	 * Set permissions on newly written files.
+	 * @param resultFile the file.
+	 * @throws IOException any exception.
+	 * @since 5.0
+	 */
+	protected void setPermissions(File resultFile) throws IOException {
+		if (this.permissions != null) {
+			Files.setPosixFilePermissions(resultFile.toPath(), this.permissions);
```
artembilan: If there's an option to check that this operation isn't supported on the target OS and `WARN` (or even `ERROR`) end-user on the matter, but not fail anyway?

Looking to the JavaDocs we really may fail and throw an exception to the caller:
```
  * @throws  UnsupportedOperationException
     *          if the associated file system does not support the {@code
     *          PosixFileAttributeView}
     * @throws  ClassCastException
     *          if the sets contains elements that are not of type {@code
     *          PosixFilePermission}
     * @throws  IOException
     *          if an I/O error occurs
     * @throws  SecurityException
     *          In the case of the default provider, and a security manager is
     *          installed, it denies {@link RuntimePermission}<tt>("accessUserInformation")</tt>
     *          or its {@link SecurityManager#checkWrite(String) checkWrite}
     *          method denies write access to the file.
     */
    public static Path setPosixFilePermissions(Path path,
                                               Set<PosixFilePermission> perms)
```

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -367,6 +373,62 @@ public void setPreserveTimestamp(boolean preserveTimestamp) {
...
+	/**
+	 * Set the file permissions after uploading, e.g. 0600 for
+	 * owner read/write. Only applies to file systems that support posix
+	 * file permissions.
+	 * @param chmod the permissions.
+	 * @since 5.0
+	 */
+	public void setChmod(int chmod) {
+		Assert.isTrue(chmod <= 0777, "'chmod' must be between 0 and 0777 (octal)");
+		BitSet bits = BitSet.valueOf(new byte[] { (byte) chmod, (byte) (chmod >> 8) });
```
artembilan: Looks like nothing to complain about, unless I'm ignorance in bit operation and this part (and below `switch`) isn't clear to me 😕 .

Would you mind to explain a bit how it works? Or at least share some doc to investigate

Thanks

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
garyrussell: Yes; it does fail; I tested on Windows - comment out the if statements in the test 😄

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
garyrussell: I will add a check for the supported filesystem and add some notes about this...

artembilan: Great! OK. 

Seems for me I see what you do here.
The `byte[]` has only 8 bits, and with the `chmod >> 8` you get access to the 9'th bit of the `chmod` variable. Where `Integer.toBinaryString(Integer.valueOf("0777", 8)) == "111111111"`

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -367,6 +374,75 @@ public void setPreserveTimestamp(boolean preserveTimestamp) {
...
+	/**
+	 * Set the file permissions after uploading, e.g. 0600 for
+	 * owner read/write. Only applies to file systems that support posix
+	 * file permissions.
+	 * @param chmod the permissions.
+	 * @throws IllegalArgumentException if the value is higher than 0777.
+	 * @since 5.0
+	 */
+	public void setChmod(int chmod) {
+		Assert.isTrue(chmod <= 0777, "'chmod' must be between 0 and 0777 (octal)");
```
artembilan: Wouldn't `0x777` look better in the code from IDE?
Also we need this `chmod` for the Java DSL

Let me know and I'll fix on merge!

artembilan (changes requested):
One simple comment yet

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
garyrussell: 0x777 is hexadecimal not octal = 1911 decimal

0777 is octal (3 bits per symbol) = 511 decimal

Octal is the normal representation of chmod.

Yes; we need it on the DSL - I'll push another commit.

artembilan: Doh! `0x` is *hexdecimal* notation for `FF` numbers...
Never mind that part.
So, only simple Java DSL support.

artembilan: Panic-stricken IDE:

panic

garyrussell: You can change it to 511 on merge, if you prefer 😄 (but leave the text as-is).

garyrussell: Please also add a test for `>=0` during merge; thx.

artembilan: Sure thing! I also going to merge those two warns and move null check to the existing warn message.
Will also run all the tests on Windows to be sure that we haven't missed POSIX hooks.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/2b99c191f6a02c44eebf8b7ba35a1ccc20c34ea1

PR #2170, opened 2017-05-30
INT-4283: MarkerFilePresentFileListFilters

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4283


spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractMarkerFilePresentFileListFilter.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public abstract class AbstractMarkerFilePresentFileListFilter<F> implements FileListFilter<F> {
+
+	private final Map<FileNameFileListFilter<F>, Function<String, String>> filtersAndFunctions = new HashMap<>();
+
+	/**
+	 * Construct an instance with a single {@link FileNameFileListFilter} and a suffix that will
+	 * be used to replace the file suffix of matched files when looking for the marker file.
+	 * i.e. if a file {@code foo.txt} is matched by the filter and the suffix is "complete",
+	 * this filter will only pass "foo.txt" if "foo.complete" is present.
```
artembilan: M-m-m. Why do you select the extension replacing approach as default one?
Seems for me `foo.txt.complete` is much robust solution, especially when we may have `foo.exe`, `foo.jpg`, `foo.java` at the same folder.

artembilan (changes requested):
That's for this round

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractMarkerFilePresentFileListFilter.java
garyrussell: :+1:

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractMarkerFilePresentFileListFilter.java
```diff
@@ -0,0 +1,141 @@
...
+		return s -> s + suffix;
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public List<F> filterFiles(F[] files) {
+		if (files.length < 2) {
+			return Collections.emptyList();
+		}
+		final Set<String> candidates = Arrays.asList(files).stream()
```
artembilan: image

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractMarkerFilePresentFileListFilter.java
```diff
@@ -0,0 +1,141 @@
...
+				.map(f -> getFilename(f))
```
artembilan: image

spring-integration-file/src/main/java/org/springframework/integration/file/filters/FileNameFileListFilter.java
```diff
@@ -0,0 +1,31 @@
...
+ * A marker interface for {@link FileListFilter}s that indicates the filter
+ * examines file names.
+ *
+ * @param <F> the type of the object being filtered.
+ *
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public interface FileNameFileListFilter<F> extends FileListFilter<F> {
```
artembilan: Gary, any strong requirements to have this extension?
Looks like it doesn't affect `AbstractMarkerFilePresentFileListFilter` unless there might be some stranger to inject an `AbstractMarkerFilePresentFileListFilter` into the `AbstractMarkerFilePresentFileListFilter`.
Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractMarkerFilePresentFileListFilter.java
```diff
@@ -0,0 +1,141 @@
...
+		for (F file : files) {
+			if (this.filtersAndFunctions.entrySet().stream().anyMatch(entry -> {
+				F[] fileToCheck = (F[]) Array.newInstance(file.getClass(), 1);
+				fileToCheck[0] = file;
+				if (entry.getKey().filterFiles(fileToCheck).size() > 0) {
+					String markerName = entry.getValue().apply(getFilename(file));
+					return markerName != null && candidates.contains(markerName);
+				}
+				return false;
+			})) {
```
artembilan: Just nit-picking from my head. Would look better with a local variable for the `anyMatch()` result.

artembilan (approved):
That's all.
A couple minor concern and simple question.

I can merge with small polishing after your answer.

artembilan:
@garyrussell ,

Any problems do not answer to my concerns and let me to merge then?

Thanks

garyrussell:
I will do the requested polishing

PR #2171, opened 2017-06-02
INT-4284: INFO about overriding readOnly headers

artembilan:
JIRA: https://jira.spring.io/browse/INT-4284

Add `INFO` into the `MessageBuilder#copyHeadersIfAbsent()` when end-user
tries to populate headers which are `readOnly`

We can't throw exception on the matter since can modify `readOnlyHeaders`
and that would force end-user to add `header-filter` logic to the application.

**Cherry-pick 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/MessageBuilder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/MessageBuilder.java
@@ -20,6 +20,9 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.integration.IntegrationMessageHeaderAccessor;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
@@ -42,6 +45,8 @@ import org.springframework.util.ObjectUtils;
  * @author Artem Bilan
  */
 public final class MessageBuilder<T> extends AbstractIntegrationMessageBuilder<T> {
+
+	private static final Log logger = LogFactory.getLog(MessageBuilder.class);
 
 	private final T payload;
 
@@ -176,7 +181,16 @@ public final class MessageBuilder<T> extends AbstractIntegrationMessageBuilder<T
 	 */
 	@Override
 	public MessageBuilder<T> copyHeadersIfAbsent(Map<String, ?> headersToCopy) {
-		this.headerAccessor.copyHeadersIfAbsent(headersToCopy);
+		if (headersToCopy != null) {
+			for (Map.Entry<String, ?> entry : headersToCopy.entrySet()) {
+				if (!this.headerAccessor.isReadOnly(entry.getKey())) {
+					this.headerAccessor.setHeaderIfAbsent(entry.getKey(), entry.getValue());
+				}
+				else if (logger.isInfoEnabled()) {
+					logger.info("The header [" + entry + "] is ignored for population because it is is readOnly.");
+				}
+			}
+		}
 		return this;
 	}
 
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/MessageBuilder.java
```diff
@@ -176,7 +181,16 @@ public T getPayload() {
...
 	@Override
 	public MessageBuilder<T> copyHeadersIfAbsent(Map<String, ?> headersToCopy) {
-		this.headerAccessor.copyHeadersIfAbsent(headersToCopy);
+		if (headersToCopy != null) {
+			for (Map.Entry<String, ?> entry : headersToCopy.entrySet()) {
+				if (!this.headerAccessor.isReadOnly(entry.getKey())) {
+					this.headerAccessor.setHeaderIfAbsent(entry.getKey(), entry.getValue());
+				}
+				else if (logger.isWarnEnabled()) {
+					logger.warn("The header [" + entry + "] is ignored for population because it is is readOnly.");
```
garyrussell: I don't think we want a WARN log for user-specified readOnlyHeaders on every build of a message.

Perhaps it would be better to put the WARN log in the for loop of `GatewayMethodInboundMessageMapper.DefaultMethodArgsMessageMapper.toMessage` Line 309 and don't invoke `headers.put(headerName, argumentValue);`

artembilan: Why do you talk only about `GatewayMethodInboundMessageMapper`?
It is fundamental restriction of any `MessageBuilder` usage.

Any `MessageBuilder.copyHeadersIfAbsent()` leads right now to the silent headers dropping...

garyrussell:
And we **can** throw an exception there (in 5.0).

spring-integration-core/src/main/java/org/springframework/integration/support/MessageBuilder.java
garyrussell: That's true; I was thinking we'd get a lot of log noise, but that would only happen if the user keeps trying to copy that header. OK; I am happy with this, but I still think we can throw an exception while building the gateway (in 5.0).

artembilan: M-m-m. Do you mean parse supplied header params and compare their names with the `readoOnlyHeaders` during gateway initialization?
Then it's OK with me.

garyrussell: Yes, just check `headerName` for read only before adding to the map (line 309) and throw an exception. We can back-port just the logging commit to 4.3.x.

artembilan:
OK. I have changed my mind and now it is an `INFO` instead of `WARN`, since that is really expectable to have incoming headers which may try to override these `readOnly`. 

The global `spring.integration.readOnly.headers` property also adds some level of complexity where end-user may expect to ignore some headers in one environment knowingly.

I would like to have this merged and have exception protection in favor of only `ID` and `TIMESTAMP` header in `gateway`, `header-enricher`, `enricher`, `header-filter`, `PublisherMetadataSource` etc.

Thanks

garyrussell:
>I don't think we want a WARN log for ... readOnlyHeaders on every build of a message.

Looks like my concern was warranted...

https://stackoverflow.com/questions/44474160/spring-integration-display-warning-that-header-is-ignored-for-population-because

Even at info, we are creating a lot of log noise because we try to copy the ID and timestamp on every mutation.

Perhaps we should drop this to debug, or remove it altogether, and just stick with https://github.com/spring-projects/spring-integration/pull/2176 in 5.0.

artembilan:
Yeah, I thought about that already, too. Since we have a note in the reference manual already: http://docs.spring.io/spring-integration/docs/4.3.10.RELEASE/reference/html/messaging-construction-chapter.html#read-only-headers, I think we are protected even without any log messages. Perhaps only improvement to make that note as important

PR #2172, opened 2017-06-02
INT-4033: Support SpEL in Gateway Timeouts

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4033

Provide a mechanism to support dynamic timeouts when invoking gateway methods.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/annotation/Gateway.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/annotation/Gateway.java
@@ -87,11 +87,12 @@ public @interface Gateway {
 	long requestTimeout() default Long.MIN_VALUE;
 
 	/**
-	 * Specify the timeout (ms) when sending to the request channel - only applies if the
-	 * send might block (such as a bounded {@code QueueChannel} that is currently full.
-	 * Overrides the encompassing gateway's default request timeout.
-	 * Overrides {@link #requestTimeout()}.
+	 * Specify a SpEL Expression to determine the timeout (ms) when sending to the request
+	 * channel - only applies if the send might block (such as a bounded
+	 * {@code QueueChannel} that is currently full. Overrides the encompassing gateway's
+	 * default request timeout. Overrides {@link #requestTimeout()}.
 	 * @return the timeout.
+	 * @since 5.0
 	 */
 	String requestTimeoutExpression() default "";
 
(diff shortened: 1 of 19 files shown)


spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodInboundMessageMapper.java
```diff
@@ -335,13 +348,25 @@ else if (Map.class.isAssignableFrom(methodParameter.getParameterType())) {
...
+					GatewayMethodInboundMessageMapper.this
+							.throwExceptionForMultipleMessageOrPayloadParameters(methodParameter);
 				}
 			}
-			Assert.isTrue(messageOrPayload != null,
-					"unable to determine a Message or payload parameter on method [" + GatewayMethodInboundMessageMapper.this.method + "]");
+			Assert.isTrue(messageOrPayload != null, "unable to determine a Message or payload parameter on method ["
+					+ GatewayMethodInboundMessageMapper.this.method + "]");
+			if (GatewayMethodInboundMessageMapper.this.sendTimeoutExpression != null) {
+				headers.putIfAbsent(GenericMessagingTemplate.DEFAULT_SEND_TIMEOUT_HEADER,
```
artembilan: `computeIfAbsent()`? to avoid unnecessary expression evaluation.

artembilan (changes requested):
I'm fine with the fix.
Just some minor concerns are requiring an explanation.

spring-integration-core/src/main/java/org/springframework/integration/annotation/Gateway.java
```diff
@@ -80,20 +80,40 @@
...
 	 * send might block (such as a bounded {@code QueueChannel} that is currently full.
-	 * Overrides the encompassing gatewsy's default request timeout.
+	 * Overrides the encompassing gateway's default request timeout.
 	 * @return the timeout.
+	 * @see #requestTimeoutExpression()
 	 */
 	long requestTimeout() default Long.MIN_VALUE;
 
 	/**
+	 * Specify the timeout (ms) when sending to the request channel - only applies if the
```
garyrussell: It's explained below...

>it can be a simple long value or a SpEL expression

spring-integration-core/src/main/java/org/springframework/integration/core/MessagingTemplate.java
```diff
@@ -97,4 +97,18 @@ public void setDefaultChannel(MessageChannel channel) {
 		return super.sendAndReceive(destination, requestMessage);
 	}
 
+	public Object receiveAndConvert(MessageChannel destination, long timeout) {
+		Message<?> message = doReceive(destination, timeout);
+		if (message != null) {
+			return doConvert(message, null);
```
garyrussell: I am not sure what you mean; see `receive()` in `MessagingGatewaySupport` - he passes in `null` too.

artembilan: Yeah! I see. Never mind

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f58106ec3cda2b1b6c828d3892ba683989d07eee

PR #2175, opened 2017-06-03
AMQP: Use ErrorMessageStrategy

garyrussell:
Fix a couple of places where `ErrorMessage` was created directly instead
of delegating to the configured `ErrorMessageStrategy`.


artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b1f74577db90a9dc1dfb89fc6d292e4ef20409f4

PR #2176, opened 2017-06-06
INT-4284: Exception to overwrite id or timestamp

artembilan:
JIRA: https://jira.spring.io/browse/INT-4284

To inform end-user that he/she can't override `id` and `timestamp` headers
throw a `BeanInitializationException` from the `gateway`, `header-enricher`,
`enricher` and `header-filter`  configuration when `id` and `timestamp` are
explicitly provided

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/MessageBuilder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/MessageBuilder.java
@@ -152,7 +152,12 @@ public final class MessageBuilder<T> extends AbstractIntegrationMessageBuilder<T
 	 */
 	@Override
 	public MessageBuilder<T> removeHeader(String headerName) {
-		this.headerAccessor.removeHeader(headerName);
+		if (!this.headerAccessor.isReadOnly(headerName)) {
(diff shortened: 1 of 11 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/MessageBuilder.java
```diff
@@ -152,7 +152,12 @@ public T getPayload() {
...
 	@Override
 	public MessageBuilder<T> removeHeader(String headerName) {
-		this.headerAccessor.removeHeader(headerName);
+		if (!this.headerAccessor.isReadOnly(headerName)) {
+			this.headerAccessor.removeHeader(headerName);
+		}
+		else if (logger.isInfoEnabled()) {
+			logger.info("The header [" + headerName + "] is ignored for removal because it is is readOnly.");
+		}
 		return this;
```
garyrussell: Need to remove logging from `copyHeadersIfAbsent`.

garyrussell:
Rebased, merged as 4a47a7ce4554e9a8feb69d3fa780bb5f57f16ad5

PR #2177, opened 2017-06-06
SCSt-913: Fix AMQP Inbound Retry logic

artembilan:
Relates to spring-cloud/spring-cloud-stream#913
and https://github.com/spring-projects/spring-integration/commit/914094e51d1c70567b524bf415d26e4f736a5f0b

Also see https://github.com/spring-projects/spring-integration-kafka/pull/163

The `RetryTemplate` and `errorChannel` are mutually exclusive options:
```
Assert.state(getErrorChannel() == null, "Cannot have an 'errorChannel' property when a 'RetryTemplate' is "
	+ "provided; use an 'ErrorMessageSendingRecoverer' in the 'recoveryCallback' property to "
	+ "send an error message when retries are exhausted");
```
Therefore we don't have any error handling functionality if there is a retry but no `RecoveryCallback`.
So, don't add `RetryContext` to `ThreadLocal<AttributeAccessor>` if we don't have `RecoveryCallback` provided,
because that value is out of use then.

To make code more logical perform `attributesHolder.remove()` only for the non-retryable branch.
With that refactoring the missed `attributesHolder.remove()` has been observed in the `AmqpInboundGateway`
Add `attributesHolder.remove()` to the `RetryListener.close()` for retryable branch

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
@@ -199,7 +199,12 @@ public class AmqpInboundChannelAdapter extends MessageProducerSupport implements
 		public void onMessage(final Message message, final Channel channel) throws Exception {
 			try {
 				if (AmqpInboundChannelAdapter.this.retryTemplate == null) {
-					processMessage(message, channel);
+					try {
+						processMessage(message, channel);
+					}
+					finally {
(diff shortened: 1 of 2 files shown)


artembilan:
@garyrussell , this should be considered for merging for the current `4.3.10` release.
Thanks

garyrussell:
Good catch; will cherry-pick soon.

garyrussell:
Cherry-picked as 4f83a953bfe4a14e0dadda01c93b055c5bc1bf61

PR #2179, opened 2017-06-08
INT-4290: JacksonJsonUtils: Add trusted packages

artembilan:
JIRA: https://jira.spring.io/browse/INT-4290

See CVE-2017-4995

To disallow deserialization of unknown classes,
the `JacksonJsonUtils#messagingAwareMapper()` can now be supplied
with the `trustedPackages`.
The default list is:
```
java.util
java.lang
org.springframework.messaging.support
org.springframework.integration.support
org.springframework.integration.message
org.springframework.integration.store
```
Can be configured with `*` (asterisk) with meaning trust all

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonUtils.java
@@ -38,7 +38,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.cfg.MapperConfig;
 import com.fasterxml.jackson.databind.jsontype.NamedType;
 import com.fasterxml.jackson.databind.jsontype.TypeIdResolver;
-import com.fasterxml.jackson.databind.jsontype.TypeResolverBuilder;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 
 /**
@@ -71,7 +70,7 @@ public final class JacksonJsonUtils {
 			mapper.configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false);
 			mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
-			mapper.setDefaultTyping(createWhitelistedDefaultTyping(trustedPackages));
+			mapper.setDefaultTyping(new WhitelistTypeResolverBuilder(trustedPackages));
 
 			GenericMessageJacksonDeserializer genericMessageDeserializer = new GenericMessageJacksonDeserializer();
 			genericMessageDeserializer.setMapper(mapper);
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonUtils.java
```diff
@@ -83,4 +99,146 @@ private JacksonJsonUtils() {
 		}
 	}
 
+	private static TypeResolverBuilder<?> createWhitelistedDefaultTyping(String... trustedPackages) {
+		TypeResolverBuilder<?> typeResolverBuilder = new WhitelistTypeResolverBuilder(trustedPackages);
+		typeResolverBuilder = typeResolverBuilder.init(JsonTypeInfo.Id.CLASS, null);
+		typeResolverBuilder = typeResolverBuilder.inclusion(JsonTypeInfo.As.PROPERTY);
```
garyrussell: Any reason for not using the builder pattern?

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageGroupStoreTests.java
```diff
@@ -453,6 +456,74 @@ public void testJsonSerialization() throws Exception {
...
+		}
+
+		public void setFoo(String foo) {
+			this.foo = foo;
+		}
+
+		@Override
+		public boolean equals(Object o) {
+			if (this == o) return true;
+			if (o == null || getClass() != o.getClass()) return false;
```
garyrussell: checkstyle

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageGroupStoreTests.java
```diff
@@ -453,6 +456,74 @@ public void testJsonSerialization() throws Exception {
...
+		}
+		catch (Exception e) {
+			assertThat(e.getCause().getCause(), instanceOf(IllegalArgumentException.class));
+			assertThat(e.getMessage(),
+					containsString("The class with " +
+							"org.springframework.integration.redis.store.RedisMessageGroupStoreTests$Foo and name of " +
+							"org.springframework.integration.redis.store.RedisMessageGroupStoreTests$Foo " +
+							"is not in the trusted packages:"));
+		}
+
```
garyrussell: Hmmm - I don't think serialization is a problem; it's deserialization that's critical.

Or, are you trying to preempt someone from storing a message that can't be deserialized?

artembilan: Right, we take care here exactly about *deserialization*. 
The problem that `addMessagesToGroup()` calls `addMessage(message)` which has `getMessage(messageId)`, therefore performs deserialization as well.

Having that I think we should go ahead and optimize `addMessage(message)` from the `addMessagesToGroup()` to something like `void doAddMessage(message)`. And we should not have extra call to the Redis/Gemfire.

garyrussell:
Merged as 99224f54c0676dadf8797e53f3c87d5d41662fcf cherry-picked as de00e72e5e03fa12d4b8bd682933d956f09bcd34

PR #2181, opened 2017-06-19
INT-4296: Fix APPEND_NO_FLUSH Timing Problem

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4296

The `FileState.close()` method correctly acquires the lock to prevent closing while
writing. However, with a short flush interval, the close could occur between state
creation and the write operation acquiring the lock.

Move the state creation to within the scope of the lock acquired by the writer.

__cherry-pick to 4.3.x__


artembilan:
... and cherry-picked.

@garyrussell , WDYT about https://jira.spring.io/browse/INT-4296?
Does this fix that issue as well?
Thanks

garyrussell:
You mean INT-4271, I think; no this doesn't fix that; see https://github.com/spring-projects/spring-integration/pull/2182

PR #2182, opened 2017-06-19
INT-4271: FWMH: Fix Race Condition with stop()

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4271

stop() is synchronized on the FWMH; a stop() while it's running will interrupt the task
which could cause a state to be removed without a flush actually happening (interrupt
on lock acquisition).

- synchronize `stop()` so it won't interrupt a running flusher
- detect an interrupt in the flusher and stop flusing

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
@@ -474,7 +474,7 @@ public class FileWritingMessageHandler extends AbstractReplyProducingMessageHand
 	}
 
 	@Override
-	public void stop() {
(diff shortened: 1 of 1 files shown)


artembilan:
Gary, I think you meant in the commit message:

> **The `Flusher`** is  synchronized on the `FWMH`

The `stop()` is `synchronized` only after this fix.

Otherwise LGTM.
Let me know and I'll polish that on merge!

garyrussell:
Yup; that's what I meant.

garyrussell:
Also typo `s/flusing/flushing/`

Thanks

artembilan:
... and cherry-picked

PR #2184, opened 2017-06-19
INT-4298: Add Interface HeaderPropagationAware

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4298

Used to indicate components that can propagate headers; allows suppression.

__cherry-pick to 4.3.x__

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageHandlerChain.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageHandlerChain.java
@@ -64,7 +64,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  */
 public class MessageHandlerChain extends AbstractMessageProducingHandler implements MessageProducer,
-		HeaderPropagationAware, CompositeMessageHandler, Lifecycle {
+		CompositeMessageHandler, Lifecycle {
 
 	private volatile List<MessageHandler> handlers;
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageHandlerChain.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageHandlerChain.java
@@ -64,7 +64,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  */
 public class MessageHandlerChain extends AbstractMessageProducingHandler implements MessageProducer,
-		CompositeMessageHandler, Lifecycle {
+		HeaderPropagationAware, CompositeMessageHandler, Lifecycle {
 
 	private volatile List<MessageHandler> handlers;
 
--- a/src/reference/asciidoc/message.adoc
+++ b/src/reference/asciidoc/message.adoc
@@ -384,6 +384,21 @@ When you try to build a new message using `MessageBuilder`, this kind of headers
 
 Starting with _version 5.0_, <<gateway,Messaging Gateway>>, <<header-enricher,Header Enricher>>, <<payload-enricher,Content Enricher>> and <<header-filter, Header Filter>> don't allow to configure `MessageHeaders.ID` and `MessageHeaders.TIMESTAMP` header names when `DefaultMessageBuilderFactory` is used and they throw `BeanInitializationException`.
 
+[[header-propagation]]
+===== Header Propagation
+
+When messages are processed (and modified) by message-producing endpoints (such as a <<service-activator, service activator>>), in general, inbound headers are propagated to the outbound message.
+One exception to this is a <<transformer, transformer>>, when a complete message is returned to the framework; in that case, the user code is responsible for the entire outbound message.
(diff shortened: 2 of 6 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/MessageHandlerChain.java
```diff
@@ -64,7 +64,7 @@
  * @author Artem Bilan
  */
 public class MessageHandlerChain extends AbstractMessageProducingHandler implements MessageProducer,
-		CompositeMessageHandler, Lifecycle {
+		HeaderPropagationAware, CompositeMessageHandler, Lifecycle {
```
artembilan: That's not true.
We don't propagate headers only in this case:
```
if (this.shouldCopyRequestHeaders()) {
		if (this.selectiveHeaderPropagation) {
			Map<String, Object> headersToCopy = new HashMap<String, Object>(requestHeaders);
			for (String header : this.notPropagatedHeaders) {
				headersToCopy.remove(header);
			}
			builder.copyHeadersIfAbsent(headersToCopy);
		}
```
So, or we shouldn't care about `MessageHandlerChain`  for this task (including any other should makes `shouldCopyRequestHeaders()` as `false`), or we should have some more sophisticated logic and clean the reply message from those headers independently of the component type. Pay attention to the `BridgeHandler` BTW...

artembilan (changes requested):

artembilan:
Yeah... I understand that my comment has disappeared after your new commit, but anyway it more and more looks for me as a dead end and only the proper solution is usage of `HeaderFilter` downstream any producing component. Otherwise `Chain`, `Resequencer`, `Bridge`, `Router` and many similar who doesn't modify request message are not affected by this logic at all and in the SCSt world we always will send an output message with mentioned headers any way.
WDYT?

garyrussell:
Good point - @mbogoevici ??

garyrussell:
BTW, your comment hasn't "disappeared" you can click "show outdated" above to see it.

artembilan:
>  "show outdated" 

Great! Thank you!

Superfluous fear 😄 

I wanted to show our with @mbogoevici discussion about proxy issue in his PR in SCSt, but wasn't able to find it. Looks like some side effect of squashing or something else...

mbogoevici:
@garyrussell @artembilan 

>Otherwise Chain, Resequencer, Bridge, Router and many similar who doesn't modify request message are not affected by this logic at all and in the SCSt world we always will send an output message with mentioned headers any way.

I see the point in the general case, but from my point of view if the request message is not modified, then we don't need to worry about header propagation either. They've been in the original message, and it's OK to keep them in the reply message (which is the same as the original message). Where we need this in particular is the case when a method returns the payload only, and the handler will blindly copy headers over, resulting in inconsistent stuff - like a POJO payload with an `application/json` header which ends up throwing off the binder. 

A `HeaderFilter` downstream of a `ServiceActivator` that only filters headers that the latter shouldn't copy at all may be a workaround but is IMO cumbersome.

artembilan:
I agree with @mbogoevici . His point of view is a good middle ground. I think we good to go ahead with the solution. Only what would be great to have is some JavaDoc on the matter that `selectiveHeaderPropagation` makes effect only if `shouldCopyRequestHeaders() == true` . Some Reference Manual would be good, too.

garyrussell:
When pulling those methods to the interface, I changed the javadocs from

    if {@link #shouldCopyRequestHeaders() shouldCopyRequestHeaaders} is true.

to

    if the handler is configured to copy headers.

Since that field is not relevant on the interface.

What would you suggest instead?

I will push some docs updates.

artembilan:
> When pulling those methods up to the interface,

Doh. I see now. Sorry.
LGTM then!

A couple words in the Docs and that's it.
Thanks

src/reference/asciidoc/message.adoc
```diff
@@ -384,6 +384,21 @@ When you try to build a new message using `MessageBuilder`, this kind of headers
...
 
+[[header-propagation]]
+===== Header Propagation
+
+When messages are processed (and modified) by message-producing endpoints (such as a <<service-activator, service activator>>), in general, inbound headers are propagated to the outbound message.
+One exception to this is a <<transformer, transformer>>, when a complete message is returned to the framework; in that case, the user code is responsible for the entire outbound message.
+When a transformer just returns the payload; the inbound headers **are** propagated.
+Also, a header is only propagated if it does not already exist in the outbound message, allowing user code to change header values as needed.
+
+Starting with _version 4.3.10_, you can configure message handlers (that modify messages and produce output) to suppress the propagation of specific headers
```
artembilan: Missed period. But merging anyway...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/42dd05d655d7e134242d40e5b12837f381c55de9 and cherry-picked.

PR #2185, opened 2017-06-23
INT-4303: (S)FTP MGET with FileExistsMode.IGNORE

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4303

Don't emit ignored files in the output message payload.

Also, always preserve the timestamp when using `FileExistsMode.REPLACE_IF_MODIFIED`.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
@@ -1039,12 +1039,16 @@ public abstract class AbstractRemoteFileOutboundGateway<F> extends AbstractReply
 			if (!appending && !tempFile.renameTo(localFile)) {
 				throw new MessagingException("Failed to rename local file");
 			}
-			if (this.options.contains(Option.PRESERVE_TIMESTAMP)) {
+			if (this.options.contains(Option.PRESERVE_TIMESTAMP)
+					|| fileExistsMode.equals(FileExistsMode.REPLACE_IF_MODIFIED)) {
 				localFile.setLastModified(getModified(fileInfo));
 			}
 		}
 		else if (FileExistsMode.REPLACE_IF_MODIFIED.equals(fileExistsMode)) {
 			logger.debug("Local file '" + localFile + "' has the same modified timestamp, ignored");
+			if (this.command.equals(Command.MGET)) {
+				localFile = null;
+			}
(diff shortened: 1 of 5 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
```diff
@@ -1039,12 +1039,16 @@ protected File get(Message<?> message, Session<F> session, String remoteDir, Str
...
 				throw new MessagingException("Failed to rename local file");
 			}
-			if (this.options.contains(Option.PRESERVE_TIMESTAMP)) {
+			if (this.options.contains(Option.PRESERVE_TIMESTAMP)
+					|| FileExistsMode.REPLACE_IF_MODIFIED.equals(fileExistsMode)) {
 				localFile.setLastModified(getModified(fileInfo));
 			}
 		}
 		else if (FileExistsMode.REPLACE_IF_MODIFIED.equals(fileExistsMode)) {
 			logger.debug("Local file '" + localFile + "' has the same modified timestamp, ignored");
```
artembilan: This DEBUG message should be wrapped with the `if (logger.isDebugEnabled()) {`
OTOH with the new logic via `if (this.command.equals(Command.MGET)) `, I'd say that this message should go to the else side.
WDYT?

I can fix on merge.

I guess since it is a logic change we can't backport it into `4.3.x`

garyrussell: I don't know what you mean by "move to the else side". The MGET logic has no bearing; it's just a signal to the caller to not include the file in the output message. The log should be emitted, regardless.

Right; no backport.

artembilan: OK. Merging with the simple polishing about `if (logger.isDebugEnabled()) {`...

garyrussell: I don't know what you mean by "move to the else side". The MGET logic has no bearing; it's just a signal to the caller to not include the file in the output message. The log should be emitted, regardless.

Right; no backport.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/0903df967680a7be8ed80d4d2edb78ceae25c459

PR #2187, opened 2017-06-23
INT-4304: (S)FTP Outbound Gateway Delete After GET

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4304

Add an option to delete remote file(s) after GET/MGET.

### DIFF
--- a/spring-integration-ftp/src/test/java/org/springframework/integration/ftp/outbound/FtpServerOutboundTests.java
+++ b/spring-integration-ftp/src/test/java/org/springframework/integration/ftp/outbound/FtpServerOutboundTests.java
@@ -60,6 +60,7 @@ import org.mockito.Mockito;
 
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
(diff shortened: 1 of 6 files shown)


spring-integration-ftp/src/test/java/org/springframework/integration/ftp/outbound/FtpServerOutboundTests.java
```diff
@@ -182,6 +189,19 @@ public void testInt2866LocalDirectoryExpressionGET() {
 	}
 
 	@Test
+	public void testGetWithRemove() {
+		String dir = "ftpSource/";
+		this.getGw.setOption(Option.DELETE);
```
artembilan: Since you modify and existing `MH` and don't reinstate its behavior after this test, I think this is dangerous and may lead to an unexpected behavior, especially in those JVMs when method declaration order isn't supported.

artembilan (approved):
Only single concern in test-case.
Will fix on merge.

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/outbound/FtpServerOutboundTests.java
artembilan: Ah! `AFTER_EACH_TEST_METHOD` does the trick. 😄 

Never mind. Sorry for the noise. 
Merging as is...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/5216dc2b023279f1e18aa070804deebb79263d1f

PR #2188, opened 2017-06-24
Added support for JdbcMetadataStore

bojanv55:


bojanv55:
Seems like derby sql generating script adds [] around Auto increment... Don't know if this was here before...

artembilan:
@bojanv55 ,

Thank you very much for so amazing contribution!
Will review it soon or since the next Monday.

Meanwhile...

> Don't know if this was here before...

Yes, I have the same problem since reworking scripts to the single files #2117.
 Looks like VPP tool somehow wrap that literal after column definition to the `[]` somehow.
I'm starting to think to get rid off of this VPP in favor of just manual sql script modifications.
Much easier to copy/paste and modify respectively, than spend time to determine why tool doesn't work as we expect...

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,198 @@
...
+	public JdbcMetadataStore(DataSource dataSource) {
+		this(new JdbcTemplate(dataSource));
+	}
+
+	/**
+	 * Instantiate a {@link JdbcMetadataStore} using provided jdbcOperations {@link JdbcOperations}.
+	 * @param jdbcOperations a {@link JdbcOperations}
+	 */
+	public JdbcMetadataStore(JdbcOperations jdbcOperations) {
+		Assert.notNull(jdbcOperations, "'dataSource' must not be null");
```
artembilan: Typo: **jdbcOperations** must not be null

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,198 @@
...
+		int updated = this.jdbcTemplate.update(getQuery(Query.REMOVE_VALUE), key);
+		if (updated != 0) {
+			return oldValue;
+		}
+		return null;
+	}
+
+	private String getQuery(Query base) {
+		return this.queryCache.computeIfAbsent(base, b ->
+				StringUtils.replace(b.getSql(), "%PREFIX%", DEFAULT_TABLE_PREFIX));
```
artembilan: If there is a `DEFAULT_TABLE_PREFIX`, then there should be a configurable property on the matter:
```
/**
	 * Specify the prefix for target data base table used from queries.
	 * @param prefix the prefix to set (default INT_).
	 */
	public void setPrefix(String prefix) {
		this.prefix = prefix;
	}
```
Also, I think the `format` solution in the `DefaultLockRepository` is the best from the performance perspective.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,198 @@
...
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Jdbc-based implementation of {@link MetadataStore}.
+ *
+ * @author Bojan Vukasovic
+ * @since 5.0
+ */
+public class JdbcMetadataStore implements ConcurrentMetadataStore {
```
artembilan: Do you consider also to add `region` support here:
```
/**
	 * A unique grouping identifier for all locks persisted with this store. Using
	 * multiple regions allows the store to be partitioned (if necessary) for different
	 * purposes. Defaults to <code>DEFAULT</code>.
	 * @param region the region name to set
	 */
	public void setRegion(String region) {
```
At least that is what we have in other places.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,198 @@
...
+				}
+				catch (EmptyResultDataAccessException e) {
+					//somebody deleted it between calls. try to insert again (go to beginning of while loop)
+				}
+			}
+		}
+	}
+
+	@Override
+	public boolean replace(String key, String oldValue, String newValue) {
```
artembilan: Why this isn't:
```
@Transactional(propagation = Propagation.REQUIRES_NEW)
```
?
Thanks

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,198 @@
...
+	private volatile Map<Query, String> queryCache = new HashMap<>();
+
+	private enum Query {
+		GET_VALUE("SELECT METADATA_VALUE FROM %PREFIX%METADATA_STORE WHERE METADATA_KEY=?"),
+		GET_VALUE_FOR_UPDATE("SELECT METADATA_VALUE FROM %PREFIX%METADATA_STORE WHERE METADATA_KEY=? FOR UPDATE"),
+		REPLACE_VALUE("UPDATE %PREFIX%METADATA_STORE SET METADATA_VALUE=? WHERE METADATA_KEY=? AND METADATA_VALUE=?"),
+		REPLACE_VALUE_BY_KEY("UPDATE %PREFIX%METADATA_STORE SET METADATA_VALUE=? WHERE METADATA_KEY=?"),
+		REMOVE_VALUE("DELETE FROM %PREFIX%METADATA_STORE WHERE METADATA_KEY=?"),
+		PUT_IF_ABSENT_VALUE("INSERT INTO INT_METADATA_STORE(METADATA_KEY, METADATA_VALUE) "
+				+ "SELECT ?, ? FROM INT_METADATA_STORE WHERE METADATA_KEY=? HAVING COUNT(*)=0");
```
artembilan: This query doesn't contain `%PREFIX%` placeholder.
I guess we need test-case for the custom prefix.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,198 @@
...
+		Assert.notNull(value, "'value' cannot be null");
+		while (true) {
+			//try to insert if does not exists
+			int affectedRows = this.jdbcTemplate.update(getQuery(Query.PUT_IF_ABSENT_VALUE), ps -> {
+				ps.setString(1, key);
+				ps.setString(2, value);
+				ps.setString(3, key);
+			});
+			if (affectedRows > 0) {
+				//it was not in the table, so we just inserted it
```
artembilan: We **have** just inserted

artembilan (changes requested):
That's all for now.
Will come back with more review a bit later.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
bojanv55: This was a copy-paste from JdbcMessageStore

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
bojanv55: Since this single query does not need transnational behavior? All other queries are transnational since we need to lock a row for editing and atomic operations. Here it is only replace and we don't care if it is not successful (affectedRows== 0).

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
bojanv55: Nope. Btw, cannot find it on ZookeeperMetadataStore nor on SimpleMetadataStore. And even with this description, still don't know what is the purpose. Any additional docs?

artembilan: Here you are: http://docs.spring.io/spring-integration/reference/html/jdbc.html#_partitioning_a_message_store.

I agree that many other impl may have similar functionality, but that is now specific only for JDBC components. So, for consistency I would like to see `region` here as well.
Thanks

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
artembilan: Cool! Would you mind to fix there with this PR as well?
Thanks

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,210 @@
...
+	private volatile Map<Query, String> queryCache = new HashMap<>();
+
+	private enum Query {
+		GET_VALUE("SELECT METADATA_VALUE FROM %PREFIX%METADATA_STORE WHERE METADATA_KEY=?"),
+		GET_VALUE_FOR_UPDATE("SELECT METADATA_VALUE FROM %PREFIX%METADATA_STORE WHERE METADATA_KEY=? FOR UPDATE"),
+		REPLACE_VALUE("UPDATE %PREFIX%METADATA_STORE SET METADATA_VALUE=? WHERE METADATA_KEY=? AND METADATA_VALUE=?"),
+		REPLACE_VALUE_BY_KEY("UPDATE %PREFIX%METADATA_STORE SET METADATA_VALUE=? WHERE METADATA_KEY=?"),
+		REMOVE_VALUE("DELETE FROM %PREFIX%METADATA_STORE WHERE METADATA_KEY=?"),
+		PUT_IF_ABSENT_VALUE("INSERT INTO %PREFIX%METADATA_STORE(METADATA_KEY, METADATA_VALUE) "
+				+ "SELECT ?, ? FROM INT_METADATA_STORE WHERE METADATA_KEY=? HAVING COUNT(*)=0");
```
artembilan: One more hard coded `INT_` has been missed

spring-integration-jdbc/src/main/sql/destroy.sql.vpp
```diff
@@ -12,7 +12,7 @@ DROP TABLE $!{IFEXISTSBEFORE} INT_MESSAGE_GROUP $!{IFEXISTS};
 DROP TABLE $!{IFEXISTSBEFORE} INT_GROUP_TO_MESSAGE $!{IFEXISTS};
 DROP TABLE $!{IFEXISTSBEFORE} INT_LOCK $!{IFEXISTS};
 DROP TABLE $!{IFEXISTSBEFORE} INT_CHANNEL_MESSAGE $!{IFEXISTS};
+DROP TABLE $!{IFEXISTSBEFORE} INT_METADATA_STORE $!{IFEXISTS};
 #if(${INT_MESSAGE_SEQ})
 DROP SEQUENCE INT_MESSAGE_SEQ $!{IFEXISTS};
-#end
-
+#end
```
artembilan: Please, consider to switch on the option in your IDE to add new line in the end of files on save.

spring-integration-jdbc/src/main/sql/schema.sql.vpp
```diff
@@ -52,3 +52,10 @@ CREATE TABLE INT_CHANNEL_MESSAGE (
 
 CREATE INDEX INT_CHANNEL_MSG_DATE_IDX ON INT_CHANNEL_MESSAGE (CREATED_DATE, MESSAGE_SEQUENCE);
 CREATE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (MESSAGE_PRIORITY DESC, CREATED_DATE, MESSAGE_SEQUENCE);
+
+
+CREATE TABLE INT_METADATA_STORE  (
+	METADATA_KEY ${VARCHAR}(255),
+	METADATA_VALUE ${VARCHAR}(255),
```
artembilan: I'm not sure in so short values for both fields. The typical use-case with the `FileListFilter` might end up with pretty long path for the file.
I agree that 4000 isn't good for the primary key anyway.
But what is the choice do we have?...
Let's make 4000 for `value` at least!

The bigger length for `key` can be done in the target application during they schema contribution.

artembilan (changes requested):
In general the solution is really cool.
I accept it.
There is need several improvements (not addressed comments). Docs on the matter. And I would like to see some integration test on the matter. Something like we have in the `PersistentAcceptOnceFileListFilterExternalStoreTests` for Redis and Gemfire.

Let us know if you are file to proceed with those requirements or we will do that for you in the separate JIRA otherwise.
BTW: https://jira.spring.io/browse/INT-4306

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,210 @@
...
+	 * executed. Defaults to {@link #DEFAULT_TABLE_PREFIX}.
+	 *
+	 * @param tablePrefix the tablePrefix to set
+	 */
+	public void setTablePrefix(String tablePrefix) {
+		this.tablePrefix = tablePrefix;
+	}
+
+	@Override
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
```
artembilan: Looking to your PR #2191 , I'd say that it isn't good idea to start a new separate isolated TX to make the record to be appeared in the store independently of the state the top level business logic.

That is exactly what I suggested you in your SO question: https://stackoverflow.com/questions/44700241/ftpstreamingmessagesource-retry-on-failure - to make the `FileListFilter` logic as a part of the whole TX and, therefore, the real record will appear in the store only after proper business TX commit.

I won't mind about `@Transactional` annotation here to enforce end-user to support `<transactional>` on the `<poller>` for FTP Inbound Channel Adapter, but that's all - no any `REQUIRES_NEW` to make this TX-based store working as all others.

Do you have anything else in mind?

Thanks

bojanv55: In JdbcMetadataStore I used transactions in order to make 2 or more SQL statements atomic (wherever I used FOR UPDATE - in putIfAbsent(), put() and remove()). This way inside transaction I'm sure that row that is selected FOR UPDATE is not modified. I used _NEW just to narrow time that lock is held on db row in this way (FOR UPDATE).

Regarding my other PR, I suppose that if I use same transaction (and DB server) inside MetadataStore, Poller and my class that saves to DB it would work probably. I would't have retry policies and ttl as in my other PR but it would hopefully work.

bojanv55: Also I think that when files are accepted in filters it is all or nothing. So if I have 1000 files, and file 1000 has processing error (or something else happens), on roll-back no files are saved to db. This way I have to re-process everything again. And if I had this workflow [start transaction] -> [read from ftp] -> [send to rabbitmq] -> [save to db] -> [commit], on rollback, I would still have 999 files sent to rabbitmq - meaning I would on next run re-send all again.

artembilan: Well, as I said: with the ` _NEW` this `MetadataStore` starts to behave as all other non-TX impls. 
Even if outer TX fails, this one will save data into DB. That is not we expect from the TX-aware `MS`, I think...
See my comment in that other your PR.

bojanv55: Yes, I agree regarding that. Like I said, just wanted to narrow db lock for performance. But in order to support my usecase (without commit filter), it would have to be without _NEW.

artembilan: For that purpose we have introduced recently `MessageSourceManagement` with its:
```
/**
	 * Set the maximum number of objects the source should fetch if it is necessary to
	 * fetch objects. Setting the
	 * maxFetchSize to 0 disables remote fetching, a negative value indicates no limit.
	 * @param maxFetchSize the max fetch size; a negative value means unlimited.
	 */
	@ManagedAttribute(description = "Maximum objects to fetch")
	void setMaxFetchSize(int maxFetchSize);
```
So, you should configure the FTP `MessageSource` to poll only one file remote file during TX.
(The `TransactionInterceptor` advice is applied to the `doPoll()` method). I advice to go this way to be sure that you fetch only one file during TX. 

Anyway that doesn't work like you explain: 
- on `poll()` the TX is started;
- the `poll()` fetch remote files if there is no local entities for them;
- then this `poll()` retrieves a single entity from local store and send a message for that;
- when the message is sent successfully, the TX is committed.

So, there is no 1000 files in one TX, that is just not possible by the implementation...

bojanv55: I'm using `FtpStreamingMessageSource source = new FtpStreamingMessageSource(new FtpRemoteFileTemplate(ftpSessionFactory()));` so I suppose there is no file download here to file system (only memory)? Also in this case I noticed that files are after accepting in filters saved immediately to metadatastore. so when you say local entities - i assume local records in metadata store? Also if I use direct channel to connect handler method to poller isn't it same thread then? if it is same thread I assume tx is still not committed?

artembilan: You understand the process correctly (almost).

Right, there is no local dir, but just only memory:
```
protected AbstractFileInfo<F> poll() {
	if (this.toBeReceived.size() == 0) {
		listFiles();
	}
	return this.toBeReceived.poll();
}
```
The `FileListFilter` is involved in that `listFiles()`, therefore in the scope of global TX.
Since your current impl `putIfAbsent()` is marked with `_NEW`, we start the new TX on that method and commit when we leave it. Right, the global TX is still active but that already isn't related to the separate transactions for each `putIfAbsent()`.

My idea really don't save in the store until after successful global TX after the process on that Direct Channel.

Or... Go the way I have described in that your other PR #2191  - remove failed files from the filter during app start according some external state store.

bojanv55: I understood what you are saying for _NEW. I'll correct that to @Transactional REQUIRED. This way I suppose that I don't need my PR anymore. But maybe I will use something like this in my project because of retry policy.

artembilan: Well, you always can add `RetryInterceptor` withing TX advice. 
See Spring Retry on the matter: https://github.com/spring-projects/spring-retry

spring-integration-jdbc/src/main/sql/schema.sql.vpp
```diff
@@ -56,6 +56,7 @@ CREATE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (MESSAGE_PRIORI
 
 CREATE TABLE INT_METADATA_STORE  (
 	METADATA_KEY ${VARCHAR}(255),
-	METADATA_VALUE ${VARCHAR}(255),
+	METADATA_VALUE ${VARCHAR}(4000),
+	REGION ${VARCHAR}(100) NOT NULL,
 	constraint METADATA_STORE primary key (METADATA_KEY)
```
artembilan: After introducing `REGION` it must become as a part of PK. 
See similar state on the `INT_LOCK` table.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,228 @@
+/*
+ * Copyright 2002-2017 the original author or authors.
```
artembilan: Not a big deal from your side, just in case you will follow with the rule in the future to simplify our job:
We use only the current year in Copyright for the new classes.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,228 @@
...
+import org.springframework.integration.metadata.ConcurrentMetadataStore;
+import org.springframework.integration.metadata.MetadataStore;
+import org.springframework.jdbc.core.JdbcOperations;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Jdbc-based implementation of {@link MetadataStore}.
```
artembilan: I'd like to see more explanation how it works.
Thanks

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,228 @@
...
+
+	private volatile String tablePrefix = DEFAULT_TABLE_PREFIX;
+
+	private volatile String region = "DEFAULT";
+
+	private final JdbcOperations jdbcTemplate;
+
+	private volatile Map<Query, String> queryCache = new HashMap<>();
+
+	private enum Query {
```
artembilan: One more time (maybe you have just missed my comment):

The approach in the `DefaultLockRepository` looks better for me:
```
	this.deleteQuery = String.format(this.deleteQuery, this.prefix);
	this.deleteExpiredQuery = String.format(this.deleteExpiredQuery, this.prefix);
	this.deleteAllQuery = String.format(this.deleteAllQuery, this.prefix);
	this.updateQuery = String.format(this.updateQuery, this.prefix);
	this.insertQuery = String.format(this.insertQuery, this.prefix);
	this.countQuery = String.format(this.countQuery, this.prefix);
}
```
There is no reason to keep lazy-load cache because we have only several entities. And only with variable usage that should be much faster, than map access on the matter.
WDYT?

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @author Bojan Vukasovic
+ * @since 5.0
+ */
+@ContextConfiguration
+@RunWith(SpringJUnit4ClassRunner.class)
+@DirtiesContext // close at the end after class
+@Transactional
+public class JdbcMetadataStoreTests {
```
artembilan: I don't see the test-case for `replace()`.
Also you didn't answered to me about integration test in the `PersistentAcceptOnceFileListFilterExternalStoreTests`.

artembilan (changes requested):
I addition I asked about Docs.
Are you going to address them yourself or should we take care about them?

In general good work, of course. Just a couple concerns and some questions seeking answers.

Thanks

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
bojanv55: Should this be in this same test class or separate one?

bojanv55:
Should this be in this same test class or separate one (as `PersistentAcceptOnceFileListFilterExternalStoreTests`)? Regarding documentation, I hope you can update that one?

artembilan:
@bojanv55 ,

Thank you for the update! 
Looks great.

Yes, please, place the integration test in the `PersistentAcceptOnceFileListFilterExternalStoreTests`. This way we have all the implementations picture in the same place. Can compare and be sure that all of them works the same way.

Yes, we will take care about Docs for you after merging this PR in the separate JIRA.

artembilan:
I added sub-task for Docs: https://jira.spring.io/browse/INT-4308

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -0,0 +1,223 @@
...
+import org.springframework.integration.metadata.MetadataStore;
+import org.springframework.jdbc.core.JdbcOperations;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.util.Assert;
+
+/**
+ * Implementation of {@link MetadataStore} using a relational database via JDBC. SQL scripts to create the necessary
+ * tables are packaged as <code>org/springframework/integration/jdbc/schema-*.sql</code>, where <code>*</code> is the
+ * target database type.
```
artembilan: Not a big deal now, just for info: we prefer `90` symbols for JavaDocs.

spring-integration-jdbc/src/main/sql/schema.sql.vpp
```diff
@@ -52,3 +52,11 @@ CREATE TABLE INT_CHANNEL_MESSAGE (
 
 CREATE INDEX INT_CHANNEL_MSG_DATE_IDX ON INT_CHANNEL_MESSAGE (CREATED_DATE, MESSAGE_SEQUENCE);
 CREATE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (MESSAGE_PRIORITY DESC, CREATED_DATE, MESSAGE_SEQUENCE);
+
+
+CREATE TABLE INT_METADATA_STORE  (
+	METADATA_KEY ${VARCHAR}(255),
+	METADATA_VALUE ${VARCHAR}(4000),
+	REGION ${VARCHAR}(100) NOT NULL,
+	constraint METADATA_STORE primary key (METADATA_KEY)
```
artembilan: Not sure how my comment has been lost, but I'll do that one more time.

After introducing `REGION` we really should have composite PK, because different clients may use the same `METADATA_KEY`. See other tables definitions.
Also `METADATA_KEY` must not be `NULL` as well.

artembilan (changes requested):
Very close to the merge, but still one big bug about `PK`.

spring-integration-file/src/test/java/org/springframework/integration/file/filters/PersistentAcceptOnceFileListFilterExternalStoreTests.java
```diff
@@ -26,26 +26,39 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 
 /**
  * @author Gary Russell
  * @author Artem Bilan
  * @since 4.0
  *
  */
+@ContextConfiguration
```
artembilan: We can do testing without ApplicationContext and without that `int-derby.properties` here in the File module.
See `org.springframework.integration.jdbc.DelayerHandlerRescheduleIntegrationTests` how it initializes an `EmbeddedDatabase`.

I'll do that polishing in the separate commit after testing locally your solution and merging.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/47003077cc96a0798be5c309e1ac4e9f03a34082.

I'll polish that part together with Docs task and ping you from the PR for review or just FYI.

Thank you very much for the valuable contribution.

Looking forward for more!

PR #2189, opened 2017-06-26
composite filter bugfix

bojanv55:
If we have multiple filters chained, previous filter is actually passing files that should be filtered out to the next filter in iteration. Don't know if this was actually wanted behavior?


artembilan:
Right, that behavior is intentional.
See `ChainFileListFilter`: http://docs.spring.io/spring-integration/docs/4.3.10.RELEASE/reference/html/whats-new.html#_file_filters

bojanv55:
Closing then. Thanks.

PR #2191, opened 2017-06-26
Added transactional support in FileList filter

bojanv55:
Before I complete this, just want to check if this is needed at all. Situation is:

- Connect to FTP and pull some files
- Files are marked as PROCESSING
- If at any point in time JVM is crashed, files will stay in processing and be reprocessed
- When files are processed, they are flagged as PROCESSED
- In case there is some error in file processing, we can set max number of retries and when reached flag file as REJECTED

I needed this since currently all files are market up-front as processed. Then if JVM fails those files will be lost basically. The use-case is common one -  I need to pull files from ftp server and put them to DB. If there is an error in processing or JVM is killed (server loses power etc.) I need to be able to re-process that file on next run. Currently all filters write to log, and then try to remove the file if it is not processed sucessfully. This does not handle properly situtation with JVM crash or hardware error.

In this case, I would just need to add this:

                   @Bean
		   public FtpPersistentCommittableAcceptOnceFileListFilter acceptOnceFileListFilter(){
			FtpPersistentCommittableAcceptOnceFileListFilter filter = new FtpPersistentCommittableAcceptOnceFileListFilter(metadataStore(), "remote");
			filter.setFlushOnUpdate(true);
			filter.setMaxRetries(3);
			filter.setProcessingTTL(5, TimeUnit.MINUTES);
			return filter;
		}

and this triggers for current transaction

                    @Bean
		public TransactionSynchronizationProcessor synchronizationProcessor(){
			return new TransactionSynchronizationProcessor() {

				@Autowired
				FtpPersistentCommittableAcceptOnceFileListFilter filter;

				@Override
				public void processBeforeCommit(IntegrationResourceHolder holder) {

				}

				@Override
				public void processAfterCommit(IntegrationResourceHolder holder) {
					Message<?> message = holder.getMessage();
					if(message!=null) {	//can happen if on pull we have nothing
						FTPFile file = new FTPFile();
						file.setName(String.valueOf(message.getHeaders().get(FileHeaders.REMOTE_FILE)));
						filter.commit(file);
					}
				}

				@Override
				public void processAfterRollback(IntegrationResourceHolder holder) {
					Message<?> message = holder.getMessage();
					FTPFile file = new FTPFile();
					file.setName(String.valueOf(message.getHeaders().get(FileHeaders.REMOTE_FILE)));
					filter.remove(file);
				}
			};
		}

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractPersistentCommittableAcceptOnceFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractPersistentCommittableAcceptOnceFileListFilter.java
@@ -1,0 +1,250 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.file.filters;
+
+import java.io.Closeable;
+import java.io.Flushable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import org.springframework.integration.file.support.FileProcessingRecord;
+import org.springframework.integration.file.support.FileProcessingRecordSerializer;
+import org.springframework.integration.file.support.FileProcessingStatus;
+import org.springframework.integration.metadata.ConcurrentMetadataStore;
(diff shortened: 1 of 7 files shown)


artembilan:
I wonder why your new `JdbcMetadataStore` and TX manager for the `<poller>` is not enough?
If we have a TX around JDBC operations, the result will appeare in the DB only when TX is commited successfully. No? What am I missing?
Thanks

artembilan:
Another thought on the matter.

The one of main Spring Integration (and EIP at all) principle is loosely-coupling, so, if FTP Inbound Channel Adapter has downloaded and sent the file to the message channel properly that's already not its responsibility to worry about downstream processing. 

With that state tracing you try just overcomplicate the purpose of the `FileListFilter`, but actually that is already your application logic and not FTP Inbound Channel Adapter. And if you need to reprocess remote file one more time because of some business requirements, you don't have choice unless perform manually `ResettableFileListFilter.remove()` when your application starts. 

In forehead only an aggregator is on my tongue to track those file states. And something like `MessageGroupStoreReaper.run()` in the begging of application to `discard` unreleased messages with an appropriate  `ResettableFileListFilter.remove()` function to call.

I agree that it feels like "single unit of work" to download file and process it, but that doesn't feet to EIP approach.

Hope I am clear...

Thank you for your effort anyway, but I don't feel that this artificial `CommittableFilter` will make it into the Framework. Sorry...

PR #2192, opened 2017-06-26
INT-4297: Add FileSplitter.setFirstLineAsHeader()

artembilan:
JIRA: https://jira.spring.io/browse/INT-4297

* Add `FileSplitter.setFirstLineAsHeader()` and extract the first line from
the context before `iterator`.
* Populate the extracted header to each subsequent line

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
@@ -150,13 +150,13 @@ public class FileSplitter extends AbstractMessageSplitter {
 	}
 
 	/**
-	 * Specify the header name for the first line to be carry with the rest lines.
-	 * @param firstLieHeaderName the header name to carry first line.
+	 * Specify the header name for the first line to be carried with the messages for the rest of lines.
+	 * @param firstLineHeaderName the header name to carry first line.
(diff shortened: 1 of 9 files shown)


artembilan:
If it is fine, I'll go ahead with Docs on the matter.
Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
```diff
@@ -208,6 +225,11 @@ public void close() throws IOException {
 
 		};
 
+		String firstLineAsHeader =
+				this.firstLieHeaderName != null ?
+						bufferedReader.lines().findFirst().orElse(null)
```
garyrussell: Need to be careful here; I think you'll find that the bufferedReader.lines() stream returns an empty string if the file contains just one newline.

Probably easier to avoid the stream here.

garyrussell (changes requested):
That's all.

garyrussell:
Doc polish and field name fix here https://github.com/garyrussell/spring-integration/commit/afaf42544817ce4e15f888f171b66084f3193821 for your approval.

garyrussell:
Merged as d8382e0bd04b4a650213012e5ff0e1900f2cbde0

PR #2193, opened 2017-06-27
INT-4305: (S)FTP: remove localFile before rename

artembilan:
JIRA: https://jira.spring.io/browse/INT-4305

The `File.renameTo()` operation may fail, therefore the content of the
local file isn't changed, but since we change `setLastModified()` anyway,
this file might be eligible for local polling.
So, we end up with the same content from local file in a new message,
meanwhile we expect a new content from the remote file

* Check the `File.renameTo()` result and attempt to `delete()`
for existing local file
* When file isn't renames remove the remote file from the `filter` to let
it be transferred one more time on the next poll.
The local file might be opened for processing, so this way we postpone a fresh
remote file for the future poll rounds
* Modify `copyFileToLocalDirectory()` to return `boolean` to reflect the fact of copy.
This way we check the real number of transferred files

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
@@ -39,6 +39,7 @@ import org.springframework.expression.Expression;
 import org.springframework.expression.common.LiteralExpression;
 import org.springframework.integration.expression.ExpressionUtils;
 import org.springframework.integration.file.filters.FileListFilter;
+import org.springframework.integration.file.filters.ResettableFileListFilter;
 import org.springframework.integration.file.filters.ReversibleFileListFilter;
 import org.springframework.integration.file.remote.RemoteFileTemplate;
 import org.springframework.integration.file.remote.session.Session;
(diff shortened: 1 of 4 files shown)


artembilan:
This is slightly breaking change, so I won't mind if you reject it for backport into `4.3.x`.
We always can recommend to remove local file after processing.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -342,18 +343,42 @@ protected void copyFileToLocalDirectory(String remoteDirectoryPath, F remoteFile
...
 			}
 
-			if (tempFile.renameTo(localFile)) {
+			boolean renamed = tempFile.renameTo(localFile);
+
+			if (!renamed) {
+				if (localFile.delete()) {
+					renamed = tempFile.renameTo(localFile);
+				}
+			}
```
garyrussell: We should log an ERROR is `localFile.delete` fails or the second `renameTo` fails the logging below does not provide enough information as to the root cause.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -342,18 +343,42 @@ protected void copyFileToLocalDirectory(String remoteDirectoryPath, F remoteFile
...
+							"Revert the remote file '" + remoteFile + "' from filter if possible...");
+				}
+				if (this.filter instanceof ResettableFileListFilter) {
+					((ResettableFileListFilter<F>) this.filter).remove(remoteFile);
+				}
 			}
 		}
+		else if (this.logger.isWarnEnabled()) {
+			this.logger.warn("The remote file '" + remoteFile + "' hasn't been transferred " +
+					"to the existing local file '" + localFile + "'. Consider to remove local file.");
```
garyrussell: `s/Consider to remove/Consider removing/`

garyrussell (changes requested):
Since this is a bug, I have no problems cherry-picking.

garyrussell:
A [little polishing for your review](https://github.com/garyrussell/spring-integration/commit/8725813a0c825cec1fb3ef53130fafcde3f0904d).

garyrussell (approved):

garyrussell:
Merged as 72bd4e5bcf4b40bd4537092b7fb026e6751a1aae

Unfortunately, cherry-pick needs a backport.

artembilan:
>  cherry-pick needs a backport. 

Will do soon.
Thanks

PR #2194, opened 2017-06-27
INT-4280: Add `taskScheduler()` to Java DSL

artembilan:
JIRA: https://jira.spring.io/browse/INT-4280

To let end-user easily to inject custom `TaskScheduler` to the endpoint
add `ConsumerEndpointSpec.taskScheduler()` option

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
@@ -164,7 +164,6 @@ public class ConsumerEndpointFactoryBean
 	}
 
 	public void setTaskScheduler(TaskScheduler taskScheduler) {
-		Assert.notNull(taskScheduler, "taskScheduler must not be null");
 		this.taskScheduler = taskScheduler;
 	}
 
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
```diff
@@ -160,6 +163,11 @@ public void setAdviceChain(List<Advice> adviceChain) {
 		this.adviceChain = adviceChain;
 	}
 
+	public void setTaskScheduler(TaskScheduler taskScheduler) {
+		Assert.notNull(taskScheduler, "taskScheduler must not be null");
```
garyrussell: It doesn't appear that this assertion is necessary.

spring-integration-core/src/main/java/org/springframework/integration/dsl/CorrelationHandlerSpec.java
```diff
@@ -135,7 +135,7 @@ public S groupTimeout(Function<MessageGroup, Long> groupTimeoutFunction) {
 	 * @see AbstractCorrelatingMessageHandler#setTaskScheduler(TaskScheduler)
 	 */
 	public S taskScheduler(TaskScheduler taskScheduler) {
-		Assert.notNull(taskScheduler, "'taskScheduler' must not be null.");
```
garyrussell: I think the assertion is better here (see my previous comment) - someone might delete that one without knowledge of this super() call.

garyrussell (changes requested):
That's all.

garyrussell:
Merged as 54b0800d08613aa844fb9997397949e2ebc5e933

PR #2198, opened 2017-07-01
INT-4300: Add WebFlux server support

artembilan:
JIRA: https://jira.spring.io/browse/INT-4300

* Add `ReactiveHttpInboundEndpoint` based on the WebFlux foundation
* Extract `BaseHttpInboundEndpoint` for common
HTTP Inbound Channel Adapters options
* Make `spring-webmvc` and `spring-webflux` as `optional` dependencies
to let end-user to choose
* Refactor `HttpContextUtils` to include constants
for newly added WebFlux support
* Introduce `BaseHttpInboundEndpoint.setRequestPayloadTypeClass()`
for raw `Class<?>` and modify existing `setRequestPayloadType()`
for the `ResolvableType`
* Refactor existing MVC tests and XML components parsers to use
new `setRequestPayloadTypeClass()`
* Add `MessagingGatewaySupport.sendAndReceiveMessageReactive()` to get
a reply from downstream flow reactive back-pressure manner
* Add `IntegrationHandlerResultHandler` implementation to let WebFlux
infrastructure to handle the `Mono<Void>` from the `ReactiveHttpInboundEndpoint`
properly
* Fix `JdbcLockRegistryLeaderInitiatorTests` race condition to assert
the `initiator1` is elected eventually after yielding when the `initiator2`
is stopped


artembilan:
Really need tests and afterwards Docs.
But ready for feedback

Thanks

/CC @smaldini , @rstoyanchev

spring-integration-http/src/main/java/org/springframework/integration/http/inbound/ReactiveHttpInboundEndpoint.java
```diff
@@ -0,0 +1,382 @@
...
+			}
+		});
+	}
+
+	@SuppressWarnings("unchecked")
+	private Mono<Void> doHandle(ServerWebExchange exchange) {
+		return Mono.just(exchange)
+				.doOnSubscribe(s -> this.activeCount.incrementAndGet())
+				.flatMap(this::extractRequestBody)
+				.switchIfEmpty(Mono.just(exchange.getRequest().getQueryParams()))
```
rstoyanchev: Why `Mono.just` + `flatMap` vs just `extractRequestBody()`?

rstoyanchev: Also the `switchIfEmpty` is probably better handled as part of `extractRequestBody` where there are some checks already for what kind of request it is. For example an HTTP POST with form data should always have a body and hence a fallback on query params doesn't make much sense. And do you really want query params as the body any time there is no body? I would think that only makes sense with something like a search form, i.e. HTTP GET with query params.

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -533,6 +543,127 @@ private Object doSendAndReceive(Object object, boolean shouldConvert) {
...
+
+			Message<?> theRequestMessage = requestMessage;
+
+			return Mono.just(theRequestMessage)
+					.doOnSubscribe(s -> {
+						if (!error && this.countsEnabled) {
+							this.messageCount.incrementAndGet();
+						}
+					})
+					.flatMap(message -> {
```
rstoyanchev: Same here. The `Mono.just` + `flatMap` seems unnecessary. @smaldini ?

spring-integration-http/src/main/java/org/springframework/integration/http/inbound/ReactiveHttpInboundEndpoint.java
artembilan: Thank you, @rstoyanchev , for feedback.
Yep! 
```
return extractRequestBody(exchange)
				.doOnSubscribe(s -> this.activeCount.incrementAndGet())
				.map(body -> buildMessage(new HttpEntity<>(body, exchange.getRequest().getHeaders()), exchange))
```
does exactly the same and looks better.
I have moved `switchIfEmpty()` logic to the `extractRequestBody()`.
We just use that query params fallback in the standard, blocking implementation:
```
if (payload == null) {
	if (requestBody != null) {
		payload = requestBody;
	}
	else {
		payload = requestParams;
	}
}
```

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
artembilan: OK. I'm changing this combination to the `Mono.defer(() -> {`.
Thank you!

artembilan:
Ready for merge.

I don't see reason to add XML configuration support.
I'll add Docs eventually after some review feedback.

Also, I guess, we need Migration Guide note that MVC is already `optional`, because you may select WebFlux instead.

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -560,42 +700,56 @@ protected AttributeAccessor getErrorMessageAttributes(Message<?> message) {
...
-			else {
-				throw new MessagingException("Unsupported 'replyChannel' type [" + replyChannel.getClass() + "]."
-						+ "SubscribableChannel or PollableChannel type are supported.");
-			}
-			if (this.isRunning()) {
-				correlator.start();
+	protected void registerReplyMessageCorrelatorIfNecessary() {
+		MessageChannel replyChannel = getReplyChannel();
+		if (replyChannel != null && this.replyMessageCorrelator == null) {
+			synchronized (this.replyMessageCorrelatorMonitor) {
```
smaldini: Review synchronized use:
when some code is synchronized, make sure is only doing local state mutation, then do call  user provided code outside, e.g. 
```boolean shouldStart = false;
    synchronized { //.. 
          shouldStart = isRunning();
     }
     if(shouldStart) {
        correlator.start();
     }
```

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -533,6 +546,135 @@ private Object doSendAndReceive(Object object, boolean shouldConvert) {
...
+		registerReplyMessageCorrelatorIfNecessary();
+
+		return doSendAndReceiveMessageReactive(requestChannel, object, false);
+	}
+
+	@SuppressWarnings("unchecked")
+	private Mono<Message<?>> doSendAndReceiveMessageReactive(MessageChannel requestChannel, Object object,
+			boolean error) {
+
+		AtomicReference<Message<?>> requestMessageReference = new AtomicReference<>();
```
smaldini: No need for this.
Move lines 623 `doOnSubscribe` and 628 `onErrorResume` inside the `defer`, just after line 621 `map`

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -533,6 +546,135 @@ private Object doSendAndReceive(Object object, boolean shouldConvert) {
...
+					FutureReplyChannel replyChannel = new FutureReplyChannel();
+
+					requestMessage = MutableMessageBuilder.fromMessage(requestMessage)
+							.setReplyChannel(replyChannel)
+							.setHeader(this.messagingTemplate.getSendTimeoutHeader(), null)
+							.setHeader(this.messagingTemplate.getReceiveTimeoutHeader(), null)
+							.setErrorChannel(replyChannel)
+							.build();
+
+					requestMessageReference.set(requestMessage);
```
smaldini: No need if comment line 566 is implemented

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -533,6 +546,135 @@ private Object doSendAndReceive(Object object, boolean shouldConvert) {
...
+											.setHeader(MessageHeaders.REPLY_CHANNEL, originalReplyChannelHeader)
+											.setHeader(MessageHeaders.ERROR_CHANNEL, originalErrorChannelHeader)
+											.build());
+				})
+				.doOnSubscribe(s -> {
+					if (!error && this.countsEnabled) {
+						this.messageCount.incrementAndGet();
+					}
+				})
+				.onErrorResume(t -> handleSendError(requestMessageReference.get(), t));
```
smaldini: if moved inside defer, can directly use message created in it.

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -533,6 +546,135 @@ private Object doSendAndReceive(Object object, boolean shouldConvert) {
...
+
+	private Mono<Message<?>> handleSendError(Message<?> requestMessage, Throwable exception) {
+		if (logger.isDebugEnabled()) {
+			logger.debug("failure occurred in gateway sendAndReceiveReactive: " + exception.getMessage());
+		}
+		MessageChannel errorChannel = getErrorChannel();
+		if (errorChannel != null) {
+			ErrorMessage errorMessage = buildErrorMessage(requestMessage, exception);
+			try {
+				return doSendAndReceiveMessageReactive(errorChannel, errorMessage, true);
```
smaldini: Does it resume the response from error channel then ? If so the message will be different, is it expected ?

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
artembilan: Ah! Cool! 
Thanks 😄

smaldini:
LGTM besides minor tweaks I commented. I'd need to diagnose an end to end flow when time permits to see where further support should be made separately 👍

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
artembilan: That's correct. If error handling flow returns good result we sent it back to the subscriber.
If that fails with an exception, we just propagate it into `onError`.

This behavior is fully similar to what we have with standard, blocking `sendAndReceive()`

artembilan: Thank you for the pointer, @smaldini !

After farther review I noticed that we really don't stop to call `handleSendError()` recursively.
That is exactly for what I have that `boolean error` argument.
So, now it is like this:
```
.onErrorResume(t -> error ? Mono.error(t) : handleSendError(requestMessage, t));
```

garyrussell:
>I'll add Docs eventually after some review feedback.

Docs?

artembilan:
> Docs?

Will do today.
Thanks

garyrussell:
Minor doc polish here: https://github.com/garyrussell/spring-integration/commit/90cad0dcafbf7d8581237c7bbf9dfb46cad0f53b

build.gradle
```diff
@@ -391,8 +391,9 @@ project('spring-integration-http') {
 	description = 'Spring Integration HTTP Support'
 	dependencies {
 		compile project(":spring-integration-core")
-		compile "org.springframework:spring-webmvc:$springVersion"
-		compile "org.springframework:spring-webflux:$springVersion"
+		compile "org.springframework:spring-web:$springVersion"
+		compile ("org.springframework:spring-webmvc:$springVersion", optional)
+		compile ("org.springframework:spring-webflux:$springVersion", optional)
```
garyrussell: Perhaps we should split into 3 modules; `http-common`, `http-mvc` (or leave it as `http`) and `http-webflux`, to make the transitive dependencies work ok for users.

garyrussell: It can be in the next milestone, however.

artembilan: Well, actually I'm starting to think that we can make them all non-optional.
As I see by our tests it doesn't hurt to have MVC and WebFlux in the same CP.
That is in the end only end-user choice which engine to choose: `@EnableWebMvc` or `@EnableWebFlux` and configure particular server  - Servlet Container or Reactor.

garyrussell: I don't think we want add webflux and reactor onto the CP if they're only using MVC.

artembilan: OK. So be it as `spring-integration-webflux` then!
We can make it depend on SI-http and exclude MVC.

Next milestone ?

garyrussell: OK; please open a JIRA so we don't forget. 

I don't think exclude is right though; it is possible they want both.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/d4a99919ed78c011dc8d69bae745eedd425eb672

build.gradle
artembilan: https://jira.spring.io/browse/INT-4315

Issue #2197, opened 2017-07-01
DefaultXmlPayloadConverter improvement


englishman:
The problem: `DefaultXmlPayloadConverter#convertToSource` method doesn't use `DocumentBuilderFactory` instance having it inside.

Proposed solution: 

> Seeing the fact that we can't affect internals of the javax.xml.validation.Validator, I think that would be good compromise to call convertToDocument() from the validator and wrap it into DOMSource. That way, indeed, your DefaultXmlPayloadConverter would have an effect.

Details: https://stackoverflow.com/q/44812496/2313177

PR #2201, opened 2017-07-05
GH-2197: Rely on the Document creation for Source

artembilan:
Resolves https://github.com/spring-projects/spring-integration/issues/2197

Sometime we would like to avoid some XML definition features (e.g. loading of the DTD).
The `XmlValidatingMessageSelector` doesn't allow to do that easily because it doesn't
rely on the `Document` creation via `DocumentBuilderFactory`, but directly
manipulates with `Source`

* Change `DefaultXmlPayloadConverter.convertToSource()` to fallback to the `convertToDocument()`
for non-Document payloads.
This way we can configure `DefaultXmlPayloadConverter` with custom `DocumentBuilderFactory` and
bypass DTD loading, for example
* Expose `xml-converter` option for the `<int-xml:validating-filter>`  component

### DIFF
--- a/src/reference/asciidoc/xml.adoc
+++ b/src/reference/asciidoc/xml.adoc
@@ -1033,8 +1033,9 @@ Please see below for an overview of all available configuration parameters:
                            schema-location=""                    <5>
                            schema-type="xml-schema"              <6>
                            throw-exception-on-rejection="false"  <7>
-                           xml-validator="">                     <8>
-    <int:poller .../>                                            <9>
+                           xml-converter=""                      <8>
+                           xml-validator="">                     <9>
+    <int:poller .../>                                            <10>
 </int-xml:validating-filter>
 ----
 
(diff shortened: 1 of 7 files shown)


src/reference/asciidoc/xml.adoc
```diff
@@ -1069,10 +1070,11 @@ If not set it defaults to `xml-schema` which internally translates to `org.sprin
 <7> If `true` a `MessageRejectedException` is thrown in case validation fails for the provided Message's payload._Optional_.
 Defaults to `false` if not set.
 
+<8> Reference to a custom `org.springframework.integration.xml.XmlPayloadConverter` strategy.
+_Optional_.
 
 <8> Reference to a custom `sorg.springframework.xml.validation.XmlValidator` strategy.
```
garyrussell: 2x `<8>` - will fix on merge.

garyrussell:
Merged as d79e70c6df6ff974f0d03d7f713a62a798be6e57

PR #2202, opened 2017-07-10
Update index on gh-pages

kazuki43zoo:
I've updated index page on gh-pages.


artembilan:
Thank you for the contribution, @kazuki43zoo !

PR #2203, opened 2017-07-11
INT-4311: JdbcLockRegistry retry on QueryTimeoutException

kaizimmerm:
Jira: https://jira.spring.io/browse/INT-4311

* Added `QueryTimeoutException` to retry handling

Note: I was not able to find any tests on the existing `TransactionTimedOutException` handling so I did not add one for query timeout either. Please let me know if I missed something.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -28,6 +28,7 @@ import java.util.concurrent.locks.ReentrantLock;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.dao.CannotSerializeTransactionException;
 import org.springframework.dao.DataAccessResourceFailureException;
+import org.springframework.dao.QueryTimeoutException;
 import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.integration.support.locks.LockRegistry;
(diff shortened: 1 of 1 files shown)


artembilan:
Yeah... We might need someday to come up with tests on the matter, but I guess it is a bit complicated to build an appropriate scenario.
Although as you see we have missed this `QueryTimeoutException` anyway.

LGTM, merging...

I'll add your name to the class during merge.
Since it is a bug we have to backport it into 4.3.x which is still Java 6 based, so multi catch isn't an option there. Will be on cherry-pick, too.

Thank you for contribution and looking forward for more!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -114,7 +115,7 @@ public void lock() {
 				catch (CannotSerializeTransactionException e) {
 					// try again
 				}
-				catch (TransactionTimedOutException e) {
+				catch (TransactionTimedOutException | QueryTimeoutException e) {
```
artembilan: Also you missed similar fix in the `lockInterruptibly()` and `tryLock()`

kaizimmerm:
Hope thats better.

kaizimmerm:
..and thanks for the backport. We are actually on 4.3 😄

kaizimmerm:
I was struggling a bit with your style (tabs...). Should be fine now.

artembilan:
Cherry-picked into `4.3.x` as https://github.com/spring-projects/spring-integration/commit/f50a2947ac3d9a667f2f177cb44bd150986d883a and merged to the `master` as https://github.com/spring-projects/spring-integration/commit/c0d4375b1e4d8eada2894e40f1806c5fc5efd3ff

PR #2204, opened 2017-07-12
INT-4301: Make CircuitBreakerOpenException as ME

artembilan:
JIRA: https://jira.spring.io/browse/INT-4301

To keep the context of the `failedMessage` and avoid extra wrapping
to the `MessagingException` make
`CircuitBreakerOpenException extends MessagingException` directly

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerCircuitBreakerAdvice.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerCircuitBreakerAdvice.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerCircuitBreakerAdvice.java
```diff
@@ -99,12 +102,12 @@ private AtomicInteger getFailures() {
...
 	 * An exception thrown when the circuit breaker is in an open state.
 	 */
-	public static final class CircuitBreakerOpenException extends RuntimeException {
+	public static final class CircuitBreakerOpenException extends MessagingException {
 
 		private static final long serialVersionUID = 1L;
 
-		CircuitBreakerOpenException(String message) {
-			super(message);
+		public CircuitBreakerOpenException(Message<?> message, String description) {
```
garyrussell: Why the change to public? I wouldn't expect a user to need to instantiate one.

artembilan: Doh. That is artifact of the automatic IDEA refactoring. Won't mind if you fix on merge.

Thanks

garyrussell:
Merged as 6829da995f164b6481cbf7c5a691cb9508c6d87a

PR #2205, opened 2017-07-22
Int 4316

viiicky:
This PR is for the issue https://jira.spring.io/browse/INT-4316

The PR provides improvements/features inspired from the discussion at https://stackoverflow.com/questions/44991488/how-to-set-date-format-for-jsonobjectmapper-in-spring-integration

The improvements/features added are:

- It enables the default Jackson `ObjectMapper` instance being used by `ObjectToMapTransformer` to identify Java 8 `Instant` fields. As of now `Instant` fields are not treated correctly, resulting two entries per `Instant` field in the map. This issue is fixed by calling `findAndRegisterModules()` on the default mapper and adding the `jackson-datatype-jsr310` dependency, thus enabling it to handle `Instant` fields

- It adds overloaded constructor to `ObjectToMapTransformer` to enable the caller to provide their own customized mapper to be used for the conversion. This will be useful when the caller needs to make changes to a lot of configurations for conversion. The caller can instantiate `ObjectToMapTransformer` using `new ObjectToMapTransformer(customizedMapper)`, where `customizedMapper` will be an instance of `com.fasterxml.jackson.databind.ObjectMapper` with all the custom configuration set by the caller. One such example is provided in JUnit test case. If the caller needs to use the default mapper the caller can always use the default constructor `new ObjectToMapTransformer()`

All tests have been run with success using the command `gradlew build` before creating this PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -143,6 +143,7 @@ subprojects { subproject ->
 		tomcatVersion = "8.5.14"
 		xmlUnitVersion = '1.6'
 		xstreamVersion = '1.4.7'
+		jsr310Version = '2.8.9'
 	}
 
 	eclipse {
@@ -314,6 +315,7 @@ project('spring-integration-core') {
 		}
 		compile("io.fastjson:boon:$boonVersion", optional)
 		compile("com.esotericsoftware:kryo-shaded:$kryoShadedVersion", optional)
+		compile("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jsr310Version")
 
 		testCompile ("org.aspectj:aspectjweaver:$aspectjVersion")
 		testCompile "io.projectreactor:reactor-test:$reactorVersion"
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonObjectMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonObjectMapper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2016 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -47,6 +47,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * </ul>
  *
  * @author Artem Bilan
+ * @author Vikas Prasad
  * @since 3.0
  */
 public class Jackson2JsonObjectMapper extends AbstractJacksonJsonObjectMapper<JsonNode, JsonParser, JavaType> {
@@ -57,6 +58,7 @@ public class Jackson2JsonObjectMapper extends AbstractJacksonJsonObjectMapper<Js
 		this.objectMapper = new ObjectMapper();
 		this.objectMapper.configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false);
 		this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+		this.objectMapper.findAndRegisterModules();	// handle JSR310
 	}
 
 	public Jackson2JsonObjectMapper(ObjectMapper objectMapper) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonObjectMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonObjectMapper.java
@@ -58,7 +58,7 @@ public class Jackson2JsonObjectMapper extends AbstractJacksonJsonObjectMapper<Js
 		this.objectMapper = new ObjectMapper();
 		this.objectMapper.configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false);
 		this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-		this.objectMapper.findAndRegisterModules();	// handle JSR310
+		this.objectMapper.findAndRegisterModules();    // handle JSR310
 	}
 
 	public Jackson2JsonObjectMapper(ObjectMapper objectMapper) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
@@ -19,6 +19,8 @@ package org.springframework.integration.support.json;
 
 import org.springframework.util.ClassUtils;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 /**
  * Simple factory to provide {@linkplain JsonObjectMapper}
  * instances dependently of jackson-databind or boon libs in the classpath.
@@ -27,6 +29,7 @@ import org.springframework.util.ClassUtils;
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Vikas Prasad
  * @since 3.0
  *
  * @see Jackson2JsonObjectMapper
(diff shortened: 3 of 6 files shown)


build.gradle
```diff
@@ -314,6 +315,7 @@ project('spring-integration-core') {
 		}
 		compile("io.fastjson:boon:$boonVersion", optional)
 		compile("com.esotericsoftware:kryo-shaded:$kryoShadedVersion", optional)
+		compile("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jsr310Version")
```
artembilan: I don't mind to test if `findAndRegisterModules()` works well, but I really mind to have this dependency as `compile`.
`testCompile` is fully enough for this task.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonObjectMapper.java
```diff
@@ -57,6 +58,7 @@ public Jackson2JsonObjectMapper() {
 		this.objectMapper = new ObjectMapper();
 		this.objectMapper.configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false);
 		this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+		this.objectMapper.findAndRegisterModules();    // handle JSR310
```
artembilan: This must be mentioned in the class JavaDoc.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
```diff
@@ -61,6 +64,17 @@ else if (boonPresent) {
...
 
 	/**
+	 * Return an instance of {@link Jackson2JsonObjectMapper} with the {@link Jackson2JsonObjectMapper#objectMapper}
+	 * field set as the mapper passed by the caller.
+	 * @param mapper the customized mapper that the caller wants to use
+	 * @return Jackson2JsonObjectMapper object that wraps within the customized mapper provided
+	 * @since 5.0.0
+	 */
+	public static JsonObjectMapper<?, ?> getCustomizedJacksonMapper(ObjectMapper mapper) {
+		return new Jackson2JsonObjectMapper(mapper);
```
artembilan: I don't see benefits of this factory method.
You always can call ` new Jackson2JsonObjectMapper(mapper)` manually with the same result.
Also this factory class is JSON-engine agnostic, so it can't advertise factories which depend on the particular library.

I mean let's remove it altogether!

spring-integration-core/src/main/java/org/springframework/integration/transformer/ObjectToMapTransformer.java
```diff
@@ -52,14 +54,23 @@
...
-	private final JsonObjectMapper<?, ?> jsonObjectMapper = JsonObjectMapperProvider.newInstance();
+	private final JsonObjectMapper<?, ?> jsonObjectMapper;
 
 	private volatile boolean shouldFlattenKeys = true;
 
+	public ObjectToMapTransformer() {
+		this.jsonObjectMapper = JsonObjectMapperProvider.newInstance();
+	}
+
+	public ObjectToMapTransformer(ObjectMapper objectMapper) {
```
artembilan: That's not good. We advertise this class as JSON-engine independent. So, if we let to inject something that really should be `JsonObjectMapper`, but not target Jackson `ObjectMapper`.

Also the `Assert.notNull()` must be here do not allow to inject `null` for potential `NPE`s in the target code.

spring-integration-core/src/test/java/org/springframework/integration/support/json/JsonObjectMapperProviderTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.rules.ExpectedException;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ *
+ * @author Vikas Prasad
+ * @since 5.0.0
+ */
+public class JsonObjectMapperProviderTest {
```
artembilan: Please, name test classes as `Tests` - plural.

spring-integration-core/src/test/java/org/springframework/integration/support/json/JsonObjectMapperProviderTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+	@Rule
+	public ExpectedException expectedException = ExpectedException.none();
+
+	@Test
+	public void getCustomizedJacksonMapper_PassObjectMapper_ReturnsJackson2JsonObjectMapperObject() {
+		// arrange + act
+		JsonObjectMapper<?, ?> mapper = JsonObjectMapperProvider.getCustomizedJacksonMapper(new ObjectMapper());
```
artembilan: Well, when we remove this factory method, we won't need this test class at all.

spring-integration-core/src/test/java/org/springframework/integration/transformer/ObjectToMapTransformerTests.java
```diff
@@ -154,6 +161,55 @@ public void testObjectToSpelMapTransformerWithCycle() {
...
+
+		Employee employee = new Employee();
+		employee.setPerson(person);
+
+		// act
+		Map<String, Object> transformedMap = new ObjectToMapTransformer().transformPayload(employee);
+
+		// assert
+		assertEquals("If JSR310 support is enabled by calling findAndRegisterModules() on the Jackson mapper, " +
+						"Instant field should not be broken. Thus the count should exactly be 1 here.",
```
artembilan: The assert messages are not for explaining features.
Please, move it to the comment.
If you want to say here something that really should be a warn why the assert hasn't passed.

spring-integration-core/src/test/java/org/springframework/integration/transformer/ObjectToMapTransformerTests.java
```diff
@@ -283,23 +339,21 @@ public void setDepartments(List<String> departments) {
 		private List<Map<String, Object>> remarks;
 		private Child child;
 		private BigDecimal age;
+		private Date birthDate;
+		public Instant deathDate;
+		private Address address;
 		public BigDecimal getAge() {
 			return age;
 		}
-
```
artembilan: Please, be sure to keep blank lines between method definitions.
I also prefer to have them as well in between field declarations.
The properties should be accessed via `this.` prefix.

Not critical and can be polished during merge, but for letting you to know what code style we pursue.

artembilan (changes requested):
When you fix the `ObjectToMapTransformer` into the `JsonObjectMapper` ctor arg, consider, please, contribute an XSD attribute on the matter for the `<object-to-map-transformer>`.
Also there is `Transformers.toMap()` factories. So, this new option should be exposed there as well.

In the end that would be great to have a note in the `whats-new.adoc` under the `Core Changes` section.
And `transformer.adoc` (`Object-to-Map and Map-to-Object Transformers` section) should mention this new option as well.

Thanks

artembilan:
OK.
Since you don't answer, @viiicky , and we have to release today, I'm taking your work as is and polish myself to make it into the current Milestone.

Cheers!

viiicky:
@artembilan have you started to work on it? I can spend around couple of hours from my side before sleeping. And what UTC time is release/code freeze?

artembilan:
No worries! Feel free to go to sleep!

I have almost finished polishing.
Will start release process just after merging this work to the `master`.

You always can contribute any other way! 

Thank you!

viiicky:
Ok, great!!

Apart from this if I have another idea/feature in my mind(not for this release), what is the best place to discuss about it? I mean, I do not want to create a JIRA ticket without discussing whether it needs to be done.

artembilan:
Well, you can start from StackOverflow then.
But if you really have an idea, then JIRA is a good place to go.
Even if I don't agree with you (for example), someone from our team may have different opinion.
Plus JIRA is open for everyone, so that isn't a surprise if any Community member will expression his own vision.

That is how Open Source works! 
So, no worries: we always comments in any JIRAs for different reasons.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/74b59ef8d3d157c7d19e7937314812fc5a9c15e6

Thank you again for the contribution.
Looking forward for more!

PR #2210, opened 2017-08-11
INT-4327: Performance optimization for SI Bean Post Processors

TexanHogman:
https://jira.spring.io/browse/INT-4327


artembilan (approved):
LGTM.

Only one concern: do we really need to use `targetClass.getName()`? Why just `targetClass` is not enough?

Let me know and I'll fix on merge!
Meanwhile I'm looking into other `BPP`s we provide for similar issue...

artembilan:
OK. Nothing in other `BPP`s.
You did good job investigating this!

So, give me, please an answer to my simple question and I'll merge.

TexanHogman:
Wish I had a data driven answer for you, but since I don't I will offer the thought process.

First, inspiration was taken from https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java#L241

The twist is, I thought it best to cache against the class over the bean name since I thought you lose some optimization because many beans may be of the same type.

So class object versus String from class.getName() -- both will work and the question I have is the memory cost of a Class versus a String object stored in the cache.  I didn't do the analysis on this so perhaps you have some thoughts to help make the best choice here.   Calling getName() may have some overhead since it eventually goes to vm, so I think you are right that changing from Set<String> to Set<Class> and then removing the getName() call is probably the best approach,

I did just verify that making the change does preserve the functionality and I can submit this change as part of the PR if you'd like.

artembilan:
I like your point of view and how you come to the issue!

Well, I think all the `Class` objects are the same and once it is loaded by class loaded we always refer to the same memory. So, that even might be overhead to have their String representation without reason. Because this way we introduce more objects to the memory. 😄 
At least it's IMO...

I will merge your PR when you come back with that polishing.

Thank you!

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/9f4f07bcfd87d830318f1ead10ae27dbcc149615 and cherry-picked to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/a459bb55469e687a93d83ad661b43cceda38163f.

@TexanHogman , 
thank you for contribution.
Looking forward for more!

Please, consider in the future do not merge `master` to your working branch. It is much better to `rebase`.
Otherwise we have some changes ordering problems during merging back to the `master`.

PR #2211, opened 2017-08-11
INT-4315: Add WebFlux module

artembilan:
JIRA: https://jira.spring.io/browse/INT-4315

Move Reactive components outside of HTTP module to the new WebFlux one,
including XSD, tests and documentation
Make an appropriate polishing for the `http.adoc` with cross-link
to the `webflux.adoc`


garyrussell:
As I said [here](https://github.com/spring-projects/spring-integration/pull/2198#discussion_r128546135) - I think we need to move the common components to `http-core` or `http-common`. Otherwise, with this depending on si-http, reactive users who don't want MVC will still get the mvc jar on the CP when they don't need it.

artembilan:
I don't like that `common` name. We don't have it anywhere yet...

That's not a big deal to have `spring-mvc` alongside with `spring-webflux`.

We can exclude `spring-mvc` here though.
If you need them both, you should include it explicitly.
Seems for me that is the way how it works now in Spring Boot 2.0.

garyrussell:
>I don't like that common name...

That's why I said `http-core`.

But, yes, I suppose it's ok to exclude `spring-mvc` here; if they want both stacks they will need to add `spring-http` as a hard dependency rather than relying on getting it transitively.

garyrussell:
Merged as 84d60f4ab49892aac4f3a9195ea67ed8689746f5

PR #2212, opened 2017-08-11
INT-4326: Add endpoints.noAutoStartup property

artembilan:
JIRA: https://jira.spring.io/browse/INT-4326

To let to disable autoStartup for particular endpoint bean globally,
in one place, add `spring.integration.endpoints.noAutoStartup`
to the `spring.integration.properties`

So, now all the `AbstractEndpoint` checks that property for matching its
bean name to configure `autoStartup` property to `false`

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2014 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +21,9 @@ import java.util.concurrent.locks.ReentrantLock;
 
 import org.springframework.context.SmartLifecycle;
 import org.springframework.integration.context.IntegrationObjectSupport;
+import org.springframework.integration.context.IntegrationProperties;
 import org.springframework.scheduling.TaskScheduler;
+import org.springframework.util.PatternMatchUtils;
 
 /**
  * The base class for Message Endpoint implementations.
@@ -36,8 +38,11 @@ import org.springframework.scheduling.TaskScheduler;
  * @author Mark Fisher
  * @author Kris Jacyna
  * @author Gary Russell
+ * @author Artem Bilan
  */
 public abstract class AbstractEndpoint extends IntegrationObjectSupport implements SmartLifecycle {
+
+	private boolean autoStartupSetExplicitly;
 
 	private volatile boolean autoStartup = true;
 
@@ -52,6 +57,7 @@ public abstract class AbstractEndpoint extends IntegrationObjectSupport implemen
 
 	public void setAutoStartup(boolean autoStartup) {
 		this.autoStartup = autoStartup;
+		this.autoStartupSetExplicitly = true;
 	}
 
 	public void setPhase(int phase) {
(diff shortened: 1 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
```diff
@@ -63,6 +69,23 @@ public void setTaskScheduler(TaskScheduler taskScheduler) {
...
+		if (!this.autoStartupSetExplicitly) {
+			String[] endpointNamePatterns =
+					getIntegrationProperty(IntegrationProperties.ENDPOINTS_NO_AUTO_STARTUP, String[].class);
+
+			for (String pattern : endpointNamePatterns) {
+				if (PatternMatchUtils.simpleMatch(pattern, getComponentName())) {
+					this.autoStartup = false;
+					break;
+				}
+			}
```
olegz: Wondering if we should start embracing functional style more
```
Stream.of(getIntegrationProperty(IntegrationProperties.ENDPOINTS_NO_AUTO_STARTUP, String[].class))
				.filter(pattern -> PatternMatchUtils.simpleMatch(pattern, getComponentName()))
				.findFirst()
				.ifPresent(s -> autoStartup = false);
```

artembilan: That's good idea, @olegz , but it works only for the `master`, where we already have Java 8.
Since we have to backport the feature to the `4.3.x`, still based on Java 7, we don't have choice unless continue to live with old iteration style here.

OTOH I'm always a bit suspicious for Java Stream API usage  since it can't leave without objects creating and garbage collection, meanwhile the simple iteration with the condition statements are just logical levers without any allocations.

spring-integration-core/src/main/resources/META-INF/spring.integration.default.properties
```diff
@@ -4,4 +4,5 @@ spring.integration.channels.maxBroadcastSubscribers=0x7fffffff
 spring.integration.taskScheduler.poolSize=10
 spring.integration.messagingTemplate.throwExceptionOnLateReply=false
 # Defaults to MessageHeaders.ID and MessageHeaders.TIMESTAMP
-spring.integration.readOnly.headers=
\ No newline at end of file
+spring.integration.readOnly.headers=
+spring.integration.endpoints.noAutoStartup=
```
garyrussell: Missing newline.

spring-integration-core/src/test/resources/META-INF/spring.integration.properties
```diff
@@ -3,3 +3,4 @@
 #spring.integration.channels.maxBroadcastSubscribers=1
 spring.integration.taskScheduler.poolSize=20
 spring.integration.messagingTemplate.throwExceptionOnLateReply=true
+spring.integration.endpoints.noAutoStartup=fooService*
```
garyrussell: newline

garyrussell (changes requested):

garyrussell:
What is driving this feature? I see nothing in the JIRA.

Wouldn't it be easier to have a simple property `defaultAutoStartup` which would apply to all endpoints that don't have it explicity set?

artembilan:
Mark Pollack requested it from the Skipper perspective. I guess as a level do not start endpoints until some deployment consensus.

Yes, the option you suggest is easier, but at the same time we can reach this behavior with my solution using `*` for pattern to match.

garyrussell:
Merged as 6c18cbcf57902e995dd8bcfc63f5578d75540736 and cherry-picked as d8ecdce83d732bfbbc61ede7bc4aa0f15bce9923

PR #2214, opened 2017-08-12
Polished ErrorMessageSendingRecoverer

olegz:
- delegated creation of 'DefaultErrorMessageStrategy' to another constructor in cases where user provided null
- see https://github.com/spring-cloud/spring-cloud-stream/pull/1038#discussion_r132707905 for more details and justification


artembilan (approved):
LGTM
Merging...

I think we should get this pattern into the service and use it everywhere the default strategy exists. This way the end-user really should not know what it is exactly and his/her configuration may end up with the `null` to let to fall back to the default strategy in the target component.

olegz:
@artembilan +1 for _everywhere the default strategy exists_

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/d41d70735722cc738f6683aebc66091f106015fc and cherry-picked to `4.3.x`.

PR #2215, opened 2017-08-15
INT-4328: AMQP: Returns/Nacks: Create ErrorMessage

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4328

Add support for sending `ErrorMessage`s to the return and nack channels.

__cherry-pick to 4.3.x, but change default EMS to `null` (will require minor adjustment to test - set the EMS in `adapterWithReturnsAndErrorMessageStrategy`)__


artembilan (approved):
I think I got it.
Merging...

artembilan:
... and cherry-picked to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/9f79bc9927dd7890494fb4681f69e75c82b01280 after resolving conflicts and reverting to the previous behavior by default via `EMC = null`.

PR #2216, opened 2017-08-16
INT-4330 added support for Supplier-based MessageSource

olegz:
- updated documentation
 - updated test


artembilan (approved):
LGTM.

Can we also add this functionality to the `IntegrationFlows`, alongside with the `from(MessageSource<?> messageSource)`?

olegz:
@artembilan I was thinking about that as well and then some more, but need to prototype it first. Do you want to wait or do it as a separate PR?

artembilan:
Works for me either way.
We have next Milestone release plan on September 5, but if you need to have this feature immediately, we can merge it as is and you'll come back to us with the Java DSL solution lately.

Thanks!

olegz:
I'd do it as separate commit, so your final call ;)

PR #2220, opened 2017-08-19
INT-4332 updated DSL with Supplier support

olegz:
- added support for Supplier as sources of messages to the DSL
- added tests and javadocs

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlows.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlows.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.dsl;
 
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import org.reactivestreams.Publisher;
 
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlows.java
```diff
@@ -147,6 +150,30 @@ public static IntegrationFlowBuilder from(Object service, String methodName) {
 	}
 
 	/**
+	 * Provides {@link Supplier} as source of messages to the integration flow
+	 * which will be triggered by a <b>global</b> {@link SourcePollingChannelAdapter}.
```
garyrussell: I think you mean "triggered by the application context's default poller (which must be present)".

There's no such thing as a global SPCA - each source gets its own SPCA.

garyrussell (commented):
That's all; I can fix on merge if you agree.

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlows.java
olegz: Yes, that is what I meant ;)
Thank you

garyrussell:
Merged as 318741d1fd2bb185e67f02b8ffe430c5414a9fea

@olegz Pls. leave a blank line after the commit headline (which must be <= 50 chars) - otherwise Git considers it all headline...

```
* d2b43c0962a4767e5f5eba643ce9594a8e616d3b INT-4332 updated DSL with Supplier support - added support for Supplier as sources of messages to the DSL - added tests and javadocs
* 946cc229ebe11e2d45f61365903f5925e97132dc INT-4330: Add support for Supplier MessageSource
* ab9e843cfaf1972622960dc69328eed223ba2440 Fix Checkstyle violations
* a1abae03a414b31402b2bbb5f6cdc2c33f6189da INT-4329: Fix JDBC [IMPORTANT] Block
```

PR #2224, opened 2017-09-01
INT-4333: Don't Override Explicit Management Props

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4333

If the user has explicitly set management properties in a bean definition,
the configurer should not override those settings with defaults.


artembilan (approved):
LGTM
Merging...
After `@Ignore` one more Reactor-based test: 
https://build.spring.io/browse/INT-MASTER-781

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/cfab1fb1a47697e25a04010b5c5493317f01993d

PR #2225, opened 2017-09-01
INT-4336: Fix NPE in IntegrationGraphServer

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4336

Don't add "hidden" unreachable methods to the map in the `GatewayProxyFactoryBean`.
The methods can never be executed and may not have all expected properties; they
should not appear in the graph at all since they are unreachable.

Also add defensive coding to the `IntegrationGraphServer` to avoid NPEs when unexpected
conditions arise.

__cherry-pick to 4.3.x__


artembilan:
... and chery-picked as https://github.com/spring-projects/spring-integration/commit/488ec15907c53aab8730cacad073377fb4a759bf

PR #2226, opened 2017-09-01
Honor the FluxSink state in the PollChPublisherAd

artembilan:
The `PollableChannelPublisherAdapter` is based on the poll model of the
`FluxSink` and iterate and poll downstream `PollableChannel` until
there is an item or `n > 0`.
Having `take(6)` we end up with the cancel from the downstream
`Subscriber` after the batch is filled, but at the same time we continue
to poll the upstream source because `n` is like `Long.MAX_VALUE`.

* The proper way to interact is check for the `!sink.isCancelled()`
as well.
This way cancelled `sink` won't "steal" data from other subscribers


spring-integration-core/src/test/java/org/springframework/integration/dsl/reactivestreams/ReactiveStreamsTests.java
```diff
@@ -124,7 +122,7 @@ public void testPollableReactiveFlow() throws Exception {
 								.map(v -> v.split(","))
 								.flatMapIterable(Arrays::asList)
 								.map(Integer::parseInt)
-								.<Message<Integer>>map(GenericMessage<Integer>::new)
+								.<Message<Integer>>map(GenericMessage::new)
 								.concatWith(this.pollablePublisher)
```
garyrussell: Why remove the generic type here (and below). Works for me with it restored (and removes the warning from my IDE).

artembilan: OK. Will restore.
My picture in the IDEA looks like:
image

So, that made me to go ahead and remove that gray code 😄

garyrussell: I guess eclipse is not so smart 👎 😄

garyrussell:
Merged as 46de69dbf2c11186ab250eee11f808b3e0581c15

PR #2229, opened 2017-09-06
RedisQueue* classes can now have a receive timeout of 0 and block indefinitely

matthjes:
INT-4341: Allow RedisQueue* classes to have 0 receive timeout.

    JIRA: https://jira.spring.io/browse/INT-4341

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisQueueInboundGateway.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisQueueInboundGateway.java
@@ -126,7 +126,7 @@ public class RedisQueueInboundGateway extends MessagingGatewaySupport implements
 	 * @param receiveTimeout Must be non-negative. Specified in milliseconds.
 	 */
 	public void setReceiveTimeout(long receiveTimeout) {
-		Assert.isTrue(receiveTimeout > 0, "'receiveTimeout' must be > 0.");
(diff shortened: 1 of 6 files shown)


artembilan (changes requested):
Would you mind adding some simple test-case, at least for the coverage?

Also you can add your name to the `@author` list of those classes. 

That's all.

Thank you!

matthjes:
Hm, I'm not sure how to properly test this trivial change. First, the method is just a setter, second I don't know how to test whether the thread is blocked indefinitely.

Also, adding my name because of this trivial changes seems too much.

artembilan:
True. The change is trivial, so test can be, too. There is really just enough to set it to `0` in the `RedisQueueInboundChannelAdapterParserTests-context.xml` and check in the `RedisQueueInboundChannelAdapterParserTests.testInt3017CustomConfig()` that the value is applied. 
As we know previously that wasn't possible.

artembilan (approved):
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b9c8a8edf89fd127f98ef1538aa2ffa35c894fd6 and cherry-picked to `4.3.x`.

@matthjes ,
Looking forward for more!

artembilan:
Cherry-picked to `4.2.x`  as https://github.com/spring-projects/spring-integration/commit/684dd5a60972f7b4c6f00e68246c622995d533d2 after resolving conflicts.

PR #2230, opened 2017-09-06
INT-4340: Roster is still not fully disabled if subscription mode is…

etschelp:
… set to null

JIRA: https://jira.spring.io/browse/INT-4340

* Roster is now disabled on login as well

### DIFF
--- a/spring-integration-test-support/src/main/java/org/springframework/integration/test/util/TestUtils.java
+++ b/spring-integration-test-support/src/main/java/org/springframework/integration/test/util/TestUtils.java
@@ -53,6 +53,7 @@ import org.springframework.util.StringUtils;
  * @author Oleg Zhurakousky
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Philipp Etschel
  */
 public abstract class TestUtils {
 
@@ -104,6 +105,19 @@ public abstract class TestUtils {
 			Assert.isAssignable(type, value.getClass());
 		}
 		return (T) value;
+	}
+
+	/**
+	 * Sets a value for the property from the provided object.
+	 * Supports nested properties via period delimiter.
+	 * @param root the object to set the property value
+	 * @param propertyPath the property name to set a value.
+	 * Can be nested path defined by the period.
+	 * @see DirectFieldAccessor
+	 */
+	public static void setPropertyValue(Object root, String propertyPath, Object value) {
+		DirectFieldAccessor accessor = new DirectFieldAccessor(root);
+		accessor.setPropertyValue(propertyPath, value);
 	}
 
 	/**
--- a/spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
+++ b/spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
@@ -22,6 +22,7 @@ import org.jivesoftware.smack.roster.Roster;
 import org.jivesoftware.smack.tcp.XMPPTCPConnection;
 import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
 import org.jxmpp.util.XmppStringUtils;
+
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.config.AbstractFactoryBean;
 import org.springframework.context.SmartLifecycle;
--- a/spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
+++ b/spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,6 +36,7 @@ import org.springframework.util.StringUtils;
  * @author Oleg Zhurakousky
  * @author Florian Schmaus
  * @author Artem Bilan
+ * @author Philipp Etschel
  *
  * @see XMPPTCPConnection
  * @since 2.0
@@ -66,7 +67,7 @@ public class XmppConnectionFactoryBean extends AbstractFactoryBean<XMPPConnectio
 
 	private volatile boolean running;
 
-	protected volatile XMPPTCPConnection connection;
+	private volatile XMPPTCPConnection connection;
 
 
 	public XmppConnectionFactoryBean() {
@@ -166,7 +167,8 @@ public class XmppConnectionFactoryBean extends AbstractFactoryBean<XMPPConnectio
 				Roster roster = Roster.getInstanceFor(this.connection);
 				if (this.subscriptionMode != null) {
 					roster.setSubscriptionMode(this.subscriptionMode);
-				} else {
+				}
+				else {
 					roster.setRosterLoadedAtLogin(false);
 				}
 				this.connection.login();
--- a/spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
+++ b/spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
@@ -22,7 +22,6 @@ import org.jivesoftware.smack.roster.Roster;
 import org.jivesoftware.smack.tcp.XMPPTCPConnection;
 import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
 import org.jxmpp.util.XmppStringUtils;
-
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.config.AbstractFactoryBean;
 import org.springframework.context.SmartLifecycle;
@@ -67,7 +66,7 @@ public class XmppConnectionFactoryBean extends AbstractFactoryBean<XMPPConnectio
 
 	private volatile boolean running;
 
-	private volatile XMPPTCPConnection connection;
+	protected volatile XMPPTCPConnection connection;
 
 
 	public XmppConnectionFactoryBean() {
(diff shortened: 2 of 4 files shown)


spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
```diff
@@ -67,7 +66,7 @@
 
 	private volatile boolean running;
 
-	private volatile XMPPTCPConnection connection;
+	protected volatile XMPPTCPConnection connection;
```
artembilan: That isn't good to allow to mutate the property from the iheritors. Not safe.
If you still think we need to provide access to this property, then add `protected` getter.

spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
```diff
@@ -164,11 +163,13 @@ public void start() {
 			try {
 				this.connection.connect();
 				this.connection.addConnectionListener(new LoggingConnectionListener());
-				this.connection.login();
+				Roster roster = Roster.getInstanceFor(this.connection);
 				if (this.subscriptionMode != null) {
-					Roster.getInstanceFor(this.connection)
-							.setSubscriptionMode(this.subscriptionMode);
+					roster.setSubscriptionMode(this.subscriptionMode);
+				} else {
```
artembilan: Our code style is like `else` from a new line after `}`.

You can catch it pretty easy with the `./gradlew check` command.

spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
```diff
@@ -164,11 +163,13 @@ public void start() {
...
+					roster.setRosterLoadedAtLogin(false);
```
artembilan: Would you mind to mention this command in the Reference Manual we talked about in the JIRA?

artembilan (changes requested):
Also add, please, your name to the `@author` for all the affected classes and consider to update the Copyright year if needed. 

Mostly nothing critical and, of course, we can fix that ourselves during merge, however we think that it would be better if people meet all the contribution requirements to have more smooth collaboration in the future contributions.

Thank you!

spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java
etschelp: I know. The issue is that the start() method of the factory bean can not be tested at all in a unit test in its current state, because there is no way to inject a mock for the connection. The only way to test it is by either injecting a mock by reflection, adding a setter, changing the accessor (all dirty hacks), or by introducing a mock XMPP server. Is there something like the MockRestServiceServer (spring-test) for XMPP? Then we could test the interaction, this would be the preferred and best option. If there is none I guess I have to revert all changes except for the changes in the start method. And we go from there. The other option would be to add a story to add a Mock XMPP server, and after that one is done start work on this ticket again.

artembilan: Well, you are almost there! 😄 
We really use a compromise in our tests via reflection.
For this purpose we provide `TestUtils.getPropertyValue()` utility:
```
TestUtils.getPropertyValue(xmppConnectionFactoryBean, "connection", XMPPTCPConnection.class);
```
So, you `createInstance()` becomes much simpler:
```
@Override
protected XMPPConnection createInstance() throws Exception {
			return mock(XMPPTCPConnection.class);
}
```

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/TestUtils.java
```diff
@@ -107,6 +108,19 @@ else if (i == tokens.length - 1) {
...
 
 	/**
+	 * Sets a value for the property from the provided object.
+	 * Supports nested properties via period delimiter.
+	 * @param root the object to set the property value
+	 * @param propertyPath the property name to set a value.
+	 * Can be nested path defined by the period.
+	 * @see DirectFieldAccessor
+	 */
+	public static void setPropertyValue(Object root, String propertyPath, Object value) {
```
artembilan: Well, no. Please, without new suspicious API.

spring-integration-xmpp/src/test/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBeanTests.java
```diff
@@ -46,4 +51,38 @@ public void testXmppConnectionFactoryBeanViaConfig() throws Exception {
...
+		XmppConnectionFactoryBean xmppConnectionFactoryBean = new XmppConnectionFactoryBean() {
+			@Override
+			protected XMPPConnection createInstance() throws Exception {
+				return mock(XMPPTCPConnection.class);
+			}
+		};
+		xmppConnectionFactoryBean.setSubscriptionMode(null);
+		XMPPConnection connection = xmppConnectionFactoryBean.createInstance();
+		assertNotNull(connection);
+		TestUtils.setPropertyValue(xmppConnectionFactoryBean, "connection", connection);
```
artembilan: I'm pretty sure we don't need this functionality.
There is just enough to check the `connection` after `start()`.
The second your test without `setSubscriptionMode(null)` just covers the case when we don't modify `RosterLoadedAtLogin`.
That's all.

I still don't see any documentation changes though...

artembilan (changes requested):

artembilan:
Also there is some Checkstyle violation:
```
:spring-integration-xmpp:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-xmpp/src/main/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBean.java:25: 'org.springframework.beans.factory.BeanInitializationException' should be separated from previous imports. [ImportOrder]
```

You can be sure in the proper code before pushing via `./gradlew clean :spring-integration-xmpp:check`

Thanks

spring-integration-xmpp/src/test/java/org/springframework/integration/xmpp/config/XmppConnectionFactoryBeanTests.java
etschelp: Now I'm confused. I cannot call start like this, the configuration will be null because I extended the factory bean and as the configuration property is private in the super class it will not be set. The test will never run that far without injecting the property first by some means, I will get a NPE. Maybe we are talking about two different things here?

etschelp: Concerning the documentation can you point me to the documentation that describes how you organise it? I could not find it in the github pages. Also I'm not sure what to change as we do not change the documented behaviour. Maybe the only thing I would change is: 

"To switch off the Roster listener for subscriptions you should configure it with..."

to 

"To turn off Roster you should configure it with..."

artembilan: In the root of project there is `/src/reference/asciidoc/xmpp.adoc`. You can find required sentence there. Since we change the behavior a bit (`roster.setRosterLoadedAtLogin(false);`) I really would to mention it in the Doc, even if that was a bug before. Just only like:
```
In addition to the subscription the roster is also disable at login phase. See `Roster.setRosterLoadedAtLogin(Boolean)` for more information.
```

artembilan: Oh! I see what you mean about `connection` inject. The fact that we don't rely there on the `super.getObject()`, but use the local `connection` variable. 

I reworked `XmppConnectionFactoryBean` to use:
```
protected XMPPTCPConnection getConnection() {
		try {
			return (XMPPTCPConnection) getObject();
		}
		catch (Exception e) {
			throw new IllegalStateException("Cannot obtain connection instance", e);
		}
	}
```
instead of local `connection` property and now test-case looks like:
```
public void testXmppConnectionFactoryBeanNoRoster() throws Exception {
		XmppConnectionFactoryBean xmppConnectionFactoryBean = new XmppConnectionFactoryBean() {
			@Override
			protected XMPPConnection createInstance() throws Exception {
				return mock(XMPPTCPConnection.class);
			}
		};

		xmppConnectionFactoryBean.setSubscriptionMode(null);
		xmppConnectionFactoryBean.afterPropertiesSet();
		xmppConnectionFactoryBean.start();
		XMPPConnection connection = xmppConnectionFactoryBean.getObject();

		assertFalse(Roster.getInstanceFor(connection).isRosterLoadedAtLogin());
	}
```
WDYT, please?

etschelp: OK I see where you are heading at. But then we need to replace all references to the configuration property with this new getter, because otherwise it will still be null in the test. But what you risk now is that clients get a bean that has a configuration proxy instead of a properly configured bean. Because now this will work xmppConnectionFactoryBean.setConnectionConfiguration(null).  So I think it is less evil to set the configuration property by reflection.

artembilan: > But what you risk now is that clients get a bean that has a configuration proxy instead of a properly configured bean.

Not sure what you mean here. If you call `afterPropertiesSet()` (and what is definitely done by Spring), you get the properly configured `connection` instance.

etschelp: Maybe I still don't get it, but how I understand you changes we need to do the following as well:
```
	@Override
	public void start() {
           // this.connection.connect(); -> will be null so replace with getter
	    this.getConnection().connect();
```
And getObject() works in a way that you always get a proxy no matter what your bean lifecycle is, because now we are calling it outside the lifecycle anyway. So with these changes in place nothing is preventing a user to insert null instead of a configuration object and the context will start no matter what.

You can test this by adding the following lines to the tests:

```
xmppConnectionFactoryBean.setSubscriptionMode(null);
xmppConnectionFactoryBean.setConnectionConfiguration(null);
xmppConnectionFactoryBean.afterPropertiesSet();
xmppConnectionFactoryBean.setConnectionConfiguration(null);
xmppConnectionFactoryBean.start();
```

artembilan:
Polishing for confirmation: https://github.com/artembilan/spring-integration/commit/bbeae70d7c9fd716f23945eae0a9956518185648.

@etschelp ,

I also not sure what is your concern and decided just to share the solution I've thought about.
Feel free to pull my change locally and play any way.

I won't merge until your confirmation.

Everything rest LGTM, BTW!

artembilan (approved):

etschelp:
Hey, your changes look good so please go ahead and merge your changes with mine. Adding the getConnection() method is actually the better way than injecting the connection via reflection. Also after these changes the factory becomes easier to test.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/1f220e0e66cadda4c2ad79f543aea8b6f75d6e86 and cherry-pick to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/138e406ba05109ab63706834add6d287f91e7f36

@etschelp ,
thank you for the contribution.
Looking forward for more!

PR #2237, opened 2017-09-15
Add option for MutableMB do not generate headers

artembilan:
Since `MutableMessage` can be mutated, the `ID` and `TIMESTAMP` headers
can be added lately.
Or there might be some scenarios which don't care about those headers at
all.
Plus we get a performance gain when we skip their generation


garyrussell:
Merged as 008a740fb0bf1dc3f83646368f3fe689af6c6202

PR #2241, opened 2017-09-18
INT-4334 - Adds a DirectoryScanner setter in AbstractInboundFileSynchronizingMessageSource

venilnoronha:
This PR adds a setter named `setScanner` in `AbstractInboundFileSynchronizingMessageSource` to specify a custom `DirectoryScanner` to be used by the `FileReadingMessageSource`.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
@@ -158,7 +158,6 @@ public abstract class AbstractInboundFileSynchronizingMessageSource<F>
 	 * @since 5.0
 	 */
 	public void setScanner(DirectoryScanner scanner) {
-		Assert.notNull(scanner, "scanner must not be null");
 		this.fileSource.setScanner(scanner);
 		this.scannerExplicitlySet = true;
 	}
@@ -186,9 +185,8 @@ public abstract class AbstractInboundFileSynchronizingMessageSource<F>
 			}
 			FileListFilter<File> filter = buildFilter();
 			if (this.scannerExplicitlySet) {
-				if (this.fileSource.isUseWatchService()) {
-					throw new IllegalStateException("Cannot set useWatchService and scanner at the same time");
-				}
+				Assert.state(!this.fileSource.isUseWatchService(),
+						"'useWatchService' and 'scanner' are mutually exclusive.");
 				this.fileSource.getScanner().setFilter(filter);
 			}
 			else if (!this.fileSource.isUseWatchService()) {
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
@@ -185,8 +185,11 @@ public abstract class AbstractInboundFileSynchronizingMessageSource<F>
 						new SimpleMetadataStore(), getComponentName());
 			}
 			FileListFilter<File> filter = buildFilter();
-			if (this.fileSource.isUseWatchService() && this.scannerExplicitlySet) {
-				throw new IllegalStateException("Cannot set useWatchService and scanner at the same time");
+			if (this.scannerExplicitlySet) {
+				if (this.fileSource.isUseWatchService()) {
+					throw new IllegalStateException("Cannot set useWatchService and scanner at the same time");
+				}
+				this.fileSource.getScanner().setFilter(filter);
 			}
 			else if (!this.fileSource.isUseWatchService()) {
 				RecursiveDirectoryScanner directoryScanner = new RecursiveDirectoryScanner();
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
@@ -26,6 +26,7 @@ import java.util.regex.Pattern;
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.context.Lifecycle;
 import org.springframework.integration.endpoint.AbstractFetchLimitingMessageSource;
+import org.springframework.integration.file.DirectoryScanner;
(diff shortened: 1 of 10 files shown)


artembilan (changes requested):
1. Need XML support on the matter. 
2. Tests.
3. Docs.
4. Proper commit message (optional, but preferable for the future contributions; at least to avoid similar comment).

And mostly important pay attention to the logic in the `afterPropertiesSet()` around:
```
if (!this.fileSource.isUseWatchService()) {
	RecursiveDirectoryScanner directoryScanner = new RecursiveDirectoryScanner();
	directoryScanner.setFilter(filter);
	this.fileSource.setScanner(directoryScanner);
}
```

So, that looks like `useWatchService` and custom `DirectoryScanner` are mutually exclusive and we have to have some logic to let end-user to know that his/her configuration is confusing and reject the attempt to configure custom scanner and `setUseWatchService` together.

venilnoronha:
I've made the logical changes in ced2dae. However, it seems like `spring-integration-file-5.0.xsd` already contains an entry for the `scanner` attribute. Do I still need to add it somewhere else?

artembilan (changes requested):
For the `scanner` attribute you should look into the `spring-integration-ftp-5.0.xsd` and `spring-integration-sftp-5.0.xsd`

Also, I'd be glad to see an appropriate mentioning for this new attribute in the `whats-new.adoc` and `ftp.adoc` & `sftp.adoc`.

Thank you so far!

venilnoronha:
Will do. Thanks!

venilnoronha:
I've updated the `xsd`s and tests in 7bc4957.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
```diff
@@ -145,6 +151,18 @@ public void setUseWatchService(boolean useWatchService) {
...
 
+	/**
+	 * Switch the local {@link FileReadingMessageSource} to use a custom
+	 * {@link DirectoryScanner}.
+	 * @param scanner the {@link DirectoryScanner} to use.
+	 * @since 5.0
+	 */
+	public void setScanner(DirectoryScanner scanner) {
+		Assert.notNull(scanner, "scanner must not be null");
+		this.fileSource.setScanner(scanner);
```
artembilan: The `this.fileSource.setScanner(scanner);` takes care about assert. 
We don't need this extra.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
```diff
@@ -167,7 +185,13 @@ public void afterPropertiesSet() throws Exception {
 						new SimpleMetadataStore(), getComponentName());
 			}
 			FileListFilter<File> filter = buildFilter();
-			if (!this.fileSource.isUseWatchService()) {
+			if (this.scannerExplicitlySet) {
+				if (this.fileSource.isUseWatchService()) {
+					throw new IllegalStateException("Cannot set useWatchService and scanner at the same time");
```
artembilan: I would prefer `Assert.state()`

spring-integration-ftp/src/main/resources/org/springframework/integration/ftp/config/spring-integration-ftp-5.0.xsd
```diff
@@ -577,6 +577,19 @@
 					]]></xsd:documentation>
 					</xsd:annotation>
 				</xsd:attribute>
+				<xsd:attribute name="scanner" type="xsd:string">
```
artembilan: Well, that's wrong place for this attribute.
This `base-inbound-adapter-type` is used from the `inbound-streaming-channel-adapter` as well, which doesn't have any local work by nature.

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/config/FtpInboundChannelAdapterParserTests.java
```diff
@@ -100,6 +102,9 @@ public void testFtpInboundChannelAdapterComplete() throws Exception {
 		FtpInboundFileSynchronizingMessageSource inbound =
 			(FtpInboundFileSynchronizingMessageSource) TestUtils.getPropertyValue(ftpInbound, "source");
 
+		FileReadingMessageSource fileSource = (FileReadingMessageSource) TestUtils.getPropertyValue(inbound, "fileSource");
+		assertEquals(context.getBean("dirScanner"), fileSource.getScanner());
```
artembilan: That's definitely about `assertSame()` because we use dependency injection, therefore our `scanner` is indeed same as a bean.

spring-integration-sftp/src/main/resources/org/springframework/integration/sftp/config/spring-integration-sftp-5.0.xsd
```diff
@@ -571,6 +571,19 @@
 					]]></xsd:documentation>
 					</xsd:annotation>
 				</xsd:attribute>
+				<xsd:attribute name="scanner" type="xsd:string">
```
artembilan: I think the same problem with wrong place for attribute as well.

src/reference/asciidoc/ftp.adoc
```diff
@@ -259,6 +260,7 @@ The inbound channel adapter first retrieves the file to a local directory and th
 Starting with _version 5.0_, you can now limit the number of files fetched from the FTP server when new file retrievals are needed.
 This can be beneficial when the target files are very large and/or when running in a clustered system with a persistent file list filter discussed below.
 Use `max-fetch-size` for this purpose; a negative value (default) means no limit and all matching files will be retrieved; see <<ftp-max-fetch>> for more information.
+You can also provide a custom `DirectoryScanner` implementation to the `inbound-channel-adapter` via the `scanner` attribute.
```
artembilan: Would be better to mention the version:
`Since _version 5.0_, you can also...`

artembilan (changes requested):
That's all.
Nothing critical.
I'm pretty sure this is the last review round and the next push will be exactly for merge.

venilnoronha:
Addressed your comments in 0ae5858.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/797e5045bf2ef0563446deb4eb8a1915ec40ea8e.

Great stuff! Thank you!

venilnoronha:
Thanks for merging!

PR #2242, opened 2017-09-18
INT-4313 - Adds ListenableMetadataStore implementation in GemfireMetadataStore

venilnoronha:
This PR implements the `ListenableMetadataStore` interface by introducing `GemfireCacheListener` to listen to events in `GemfireMetadataStore`.

### DIFF
--- a/spring-integration-gemfire/src/main/java/org/springframework/integration/gemfire/metadata/GemfireMetadataStore.java
+++ b/spring-integration-gemfire/src/main/java/org/springframework/integration/gemfire/metadata/GemfireMetadataStore.java
@@ -16,11 +16,18 @@
 
 package org.springframework.integration.gemfire.metadata;
 
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
 import org.apache.geode.cache.Cache;
+import org.apache.geode.cache.EntryEvent;
 import org.apache.geode.cache.Region;
 import org.apache.geode.cache.Scope;
+import org.apache.geode.cache.util.CacheListenerAdapter;
 
 import org.springframework.integration.metadata.ConcurrentMetadataStore;
+import org.springframework.integration.metadata.ListenableMetadataStore;
+import org.springframework.integration.metadata.MetadataStoreListener;
 import org.springframework.util.Assert;
 
 /**
@@ -30,22 +37,27 @@ import org.springframework.util.Assert;
  * restarts.
  *
  * @author Artem Bilan
+ * @author Venil Noronha
  * @since 4.0
  */
-public class GemfireMetadataStore implements ConcurrentMetadataStore {
+public class GemfireMetadataStore implements ConcurrentMetadataStore, ListenableMetadataStore {
 
 	public static final String KEY = "MetaData";
+
+	private GemfireCacheListener cacheListener = new GemfireCacheListener();
 
 	private final Region<String, String> region;
 
 	public GemfireMetadataStore(Cache cache) {
 		Assert.notNull(cache, "'cache' must not be null");
 		this.region = cache.<String, String>createRegionFactory().setScope(Scope.LOCAL).create(KEY);
+		this.region.getAttributesMutator().addCacheListener(this.cacheListener);
 	}
 
 	public GemfireMetadataStore(Region<String, String> region) {
 		Assert.notNull(region, "'region' must not be null");
 		this.region = region;
+		this.region.getAttributesMutator().addCacheListener(this.cacheListener);
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


spring-integration-gemfire/src/main/java/org/springframework/integration/gemfire/metadata/GemfireMetadataStore.java
```diff
@@ -82,4 +94,47 @@ public String remove(String key) {
...
+		Assert.notNull(listener, "'listener' must not be null");
+		this.listeners.add(listener);
+	}
+
+	@Override
+	public void removeListener(MetadataStoreListener listener) {
+		this.listeners.remove(listener);
+	}
+
+	private class GemfireCacheListener extends CacheListenerAdapter<String, String> {
```
artembilan: Since we don't have any references to the outer class unless `listeners` I would prefer to make it as `static` and propagate the `GemfireMetadataStore.this.listeners` to the local property of this class.
In addition, please, add package protected constructor to avoid some extra generation around default `private` one. See `MetadataStoreListenerInvokingPathChildrenCacheListener` for example.

artembilan (changes requested):
1. Need tests.
2. Docs.
3. Proper commit message.

Let me know if you are in stack with something!

That's great to have a contribution from the community, but anyway the code has to be ideal 😄

spring-integration-gemfire/src/main/java/org/springframework/integration/gemfire/metadata/GemfireMetadataStore.java
venilnoronha: Addressed in 0b3481c.

artembilan:
> Merge branch 'master' into INT-4313 

We prefer `rebase`.
During merge back to the `master` we do rebase and squash. In this case there is often some conflicts because `master` code-based is colliding with its own state integration to this  branch.
The idea is to always have changes in the branch on top of `master`'s code.
Thanks

venilnoronha:
I've undone the merge commit and rebased now.

spring-integration-gemfire/src/main/java/org/springframework/integration/gemfire/metadata/GemfireMetadataStore.java
```diff
@@ -30,22 +37,27 @@
...
  * @author Artem Bilan
+ * @author Venil Noronha
  * @since 4.0
  */
-public class GemfireMetadataStore implements ConcurrentMetadataStore {
+public class GemfireMetadataStore implements ConcurrentMetadataStore, ListenableMetadataStore {
 
 	public static final String KEY = "MetaData";
 
+	private GemfireCacheListener cacheListener = new GemfireCacheListener();
```
artembilan: This must be `final` for some performance optimization.

spring-integration-gemfire/src/main/java/org/springframework/integration/gemfire/metadata/GemfireMetadataStore.java
```diff
@@ -82,4 +94,53 @@ public String remove(String key) {
...
+
+		@Override
+		public void afterUpdate(EntryEvent<String, String> event) {
+			for (MetadataStoreListener listener : this.listeners) {
+				listener.onUpdate(event.getKey(), event.getNewValue());
+			}
+		}
+
+		@Override
+		public void afterInvalidate(EntryEvent<String, String> event) {
```
artembilan: I think we shouldn't consider this operation  because `GemfireMetadataStore` doesn't support `invalidate`:
```
/**
   * Invalidates the entry with the specified key. Invalidate only removes the value from the entry,
   * the key is kept intact. To completely remove the entry, destroy should be used. The invalidate
   * will be distributed to other caches if the scope is not Scope.LOCAL.
...
   */
  public void invalidate(Object key) throws TimeoutException, **EntryNotFoundException;**
```

src/reference/asciidoc/gemfire.adoc
```diff
@@ -210,3 +210,5 @@ The _Twitter Inbound Channel Adapter_ and the _Feed Inbound Channel Adapter_ wil
 NOTE: The `GemfireMetadataStore` also implements `ConcurrentMetadataStore`, allowing it to be reliably shared across multiple application instances where only one instance will be allowed to store or modify a key's value.
 These methods give various levels of concurrency guarantees based on the scope and data policy of the region.
 They are implemented in the peer cache and client/server cache but are disallowed in peer Regions having NORMAL or EMPTY data policies.
+
+NOTE: Since _Spring Integration 5.0_, the `GemfireMetadataStore` also implements `ListenableMetadataStore`, allowing users to listen to cache events by providing `MetadataStoreListener` instances to the store.
```
artembilan: Well, we definitely in Spring Integration. To say `_version 5.0_` is enough.
If that is not about Spring Integration we would say that, surely 😄

spring-integration-gemfire/src/test/java/org/springframework/integration/gemfire/metadata/GemfireMetadataStoreCacheListenerTests.java
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+
+	@Test
+	public void testAdd() throws InterruptedException {
+		String testKey = "key";
+		String testValue = "value";
+
+		CountDownLatch latch = new CountDownLatch(1);
+		StringBuilder actualKey = new StringBuilder();
```
artembilan: That's a bit unusual approach for testing data to use `StringBuilder`. Typically we rely on the `AtomicReference`.
But let it be.

artembilan (approved):
A couple concerns - I'll address them during merge this evening.

Thank you for contribution here!

venilnoronha:
Thanks for reviewing my code! I've addressed the remaining concerns in 56950cd.

artembilan (approved):
LGTM
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b1865dab3adf8796330b743ed6ddcd7bc7295b92 after some minor polishing.

Great contribution!
Looking forward to review and merge all other your PRs.

venilnoronha:
Thanks for merging! Will fix other PRs soon.

PR #2244, opened 2017-09-20
INT-4344: AbstRemoteFileOutGw: Add assert for NPE

artembilan:
JIRA: https://jira.spring.io/browse/INT-4344

The `FtpOutboundGateway` provides ctors without expression for remote path.
In this case it is treated as a `working directory` but only for the
`LS`, `NLST`, `PUT` and `MPUT` commands.

* Add assertion in the `AbstractRemoteFileOutboundGateway` to discard
configuration for all other commands when `expression` is `null`

**Cherry-pick to 4.3.x**


garyrussell:
Merged as 077b5ffffcd48e33fdbefb6b1054ec389f92911d; cherry-picked as 7d24b8ec6775bf176b016ee855ca37bb09d2285c

PR #2254, opened 2017-10-01
INT-3876 Identify misconfigured Aggregator

meherzad:
JIRA - https://jira.spring.io/browse/INT-3876

Introduce UniqueExpiryCallback to ensure message stores are not reused across aggregators.


artembilan:
@meherzad ,

Thank you for contribution of course!

Unfortunately we haven't agreed the solution yet. So, your PR will hang on here for a while.

For more context here is my internal conclusion:

> IMO that is definitely bug that we expire not our groups. The suggested filtering logic will eliminate it and cover most use-cases with the shared `MessageStore`.
Saying end-user (discarding config or either `WARN`) that (s)he isn't right because of shared `MessageStore` isn't correct in this case. That is our Framework component issue do not perform logic properly.

> The pub-sub concern is somehow not related and in one cases that is fully OK to group from different aggregator instances to the same group on the same MS (as we realized with you only one aggregator will be able to release that group anyway because of `groupNow.isComplete()`), but in other cases even if we use separate MS instances the target store might be the same and we end up with unexpected problem storing to the same group again. But that already won't be caught by `UniqueExpiryCallback` trick.

> Therefore my suggestion to fix the bug via filtering and replace important note in the Reference Manual to the use-cases with the same group between different aggregators: https://docs.spring.io/spring-integration/reference/html/messaging-routing-chapter.html#reaper. Seek for the words:

>   When using a `MessageGroupStoreReaper`, it is generally recommended to use a separate `MessageStore` for each correlating endpoint. 

Now imaging that you have several application looking to the same store and this `UniqueExpiryCallback` won't help again: one aggregator will pick up all the groups from that `MS` and we come back to our original problem again.

What I want with the `Set groupIds` helps everywhere: we won't expire not our groups at all!

Sending messages to different aggregators with the same `correlationKey`, for different logics, is indeed misconfiguration, but that's already somehow out-of-scope of Framework. We don't carry responsibility for wrong usage of our components. There definitely might be other unexpected usages we are not aware of. 

So, having the bug fix and an important note in the Doc for known concern is fully appropriate from the Open Source perspective.

@garyrussell  is on vacation this week - we might need to wait for his vision until then, but for now you should know that I don't accept this `UniqueExpiryCallback` solution.

meherzad:
@artembilan I understand the points raised by you. Also after making changes, I can see some of the test cases are failing, I realized the points which you have raised. I'll continue working and think of some more cleaner option.

garyrussell:
While I think your idea of having the aggregator keep track of its own groups is simpler, there is a corner case where two aggregators share the same store and might use the same group concurrently. Improbable, perhaps, but possible.

That said, if they have that environment today, it would be breaking too, so it's no worse.

The purpose of this JIRA was to detect a bad configuration and fail fast rather than getting some odd unexplained side effects that might be hard to debug.

Your solution solves the root cause effect in most cases, but not all.

I can live with that as long as we __strongly__ urge using a different MS instance for each aggregator.

artembilan:
> we strongly urge using a different MS instance for each aggregator.

That's wrong to say. We may have several instances in the cluster with the shared message store. And we definitely would like to distribute aggregation logic between those several instances of the aggregator. So, that is good to have there the same `MS`.

Those corner cases you mention are really just about wrong design from the target application where semantically we have different groups but physically use the same one. That's already not Framework responsibility to track these case. More over we don't know all of them and can't predict what end-user may "cook" with our components.

We have here a bug - expire not our groups. And that's what has to be fixed in the Framework. If you want I'll rename that JIRA to properly reflect the problem or create a new one to point exactly to this problem.

OTOH the suggested solution here and your wish to disallow aggregator clustering narrows the functionality of the component. 

Yeah... Unfortunately I don't have more arguments to convince you do not introduce that `UniqueExpiryCallback` 😄

garyrussell:
>That's wrong to say. 

I think you are missing my point; I am NOT talking about multiple aggregator instances in a cluster; I AM talking about 2 different aggregators in the same application (or different applications) that are incorrectly configured to use the same `MessageStore` instance.

    ...split1->...->agg1->...->split2->...->agg2...

If the group used by agg1 and agg2 happen to be the same, they can still remove each other's group(s).

or

     App1 ...->agg->

     App2 ...->agg->...

Since we can not guarantee these apps use a UUID for the group, it is possible they can remove each other's groups.

artembilan:
```
 App1 ...->agg->

 App2 ...->agg->...
```

I wonder how are you going to detect this situation...
Even this `UniqueExpiryCallback` doesn't help. That is really already a matter of documentation as I suggested in the JIRA. 

> it is possible they can remove each other's groups.

Without tracking your own groups (even if group ids are unique) you still will be able to remove those what are not yours, when the `MS` is shared between different apps.

garyrussell:
Good point; my first case still applies but even my solution is not water-tight; we've spent too much time on this already; please go ahead with your solution.

artembilan:
@meherzad ,

Are you still willing to help us with the contribution?
Any your thoughts on the matter?

As you see we have just worked out some compromise with Gary: we add groups tracking to the `AbstractCorrelatingMessageHandler` do not expire not our groups and modify Docs to mention how it is important to properly select the `correlationKey` strategy when target application is going to use the same `MessageStore`. 

Do you have any objections regarding this solution?
Would you mind to proceed with it?

Any feedback is welcome!

Thank you for your patience!

And of course, thanks to Gary for his valuable time and wisdom.

meherzad:
@artembilan Ill proceed with the agreed solution. Thanks for the clarification.

meherzad:
@artembilan - I have made the fix, my only concern here is keeping track of the groups. I think so the operations should be atomic where we store the group ids and store messages in store. The current fix which I have made I have made sure the group id tracking happens after message store operations.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -673,7 +683,10 @@ protected void forceComplete(MessageGroup group) {
 
 	void remove(MessageGroup group) {
 		Object correlationKey = group.getGroupId();
-		this.messageStore.removeMessageGroup(correlationKey);
+		if (this.groupIds.contains(correlationKey)) {
```
artembilan: That's not logic for this method.
If we are here, we have to be already sure that it is our own group.
The real problem is definitely in the `ForceReleaseMessageGroupProcessor` implementation.
We have to check for its presence there.
All other logic should not be affected.

Please, make this `remove()` method `protected`.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -477,11 +486,12 @@ private void removeEmptyGroupAfterTimeout(MessageGroup messageGroup, long timeou
 							boolean removeGroup = groupNow.size() == 0 &&
 									groupNow.getLastModified()
 											<= (System.currentTimeMillis() - this.minimumTimeoutForEmptyGroups);
-							if (removeGroup) {
+							if (removeGroup && this.groupIds.contains(groupId)) {
```
artembilan: See my comment for the `remove()`  method changes. 
This should not be changed at all.
Although I would see `remove(groupNow)` call instead of direct `removeMessageGroup()`

artembilan (changes requested):
That's all.

Well done though!

artembilan:
@meherzad ,

let us know if you are going to fix requested concerns or we should go ahead ourselves.

The problem that we need to fix `5.0 GA` the next Tuesday and would like to see this fix there.

Thank you for your effort!

meherzad:
Apologies for delay @artembilan.

Have made the required fix.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -846,8 +846,7 @@ If the flag is set to true, then when the expiry callback is invoked, any unmark
 
 [IMPORTANT]
 =====
-When using a `MessageGroupStoreReaper`, it is generally recommended to use a separate `MessageStore` for each correlating endpoint.
-Otherwise, unexpected results may occur because one endpoint may remove another endpoint's groups.
+When using a `MessageGroupStoreReaper`, with a common `MessageStore` for multiple correlating endpoints, ensure proper `CorrelationStrategy` is used when same `MessageStore` is used for multiple aggregators to ensure unique correlationkey is being assigned to the messages.
```
artembilan: This sentence is a bit awkward.
Plus would be better to place here an answer for the question "Why?".
That  might not be so clear for end-users what is the reason to worry about this restriction.

artembilan (changes requested):
That's all.
Only single documentation issue.
Everything rest - LGTM!

meherzad:
@artembilan I have reworded the doc the change in the latest version and what can happen if proper Correlation Strategy is not used.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -846,7 +846,9 @@ If the flag is set to true, then when the expiry callback is invoked, any unmark
 
 [IMPORTANT]
 =====
-When using a `MessageGroupStoreReaper`, with a common `MessageStore` for multiple correlating endpoints, ensure proper `CorrelationStrategy` is used when same `MessageStore` is used for multiple aggregators to ensure unique correlationkey is being assigned to the messages.
+Since _version 5.0_ Correlating endpoint's keep track of correlation key of message's processed by the endpoint.
```
artembilan: How about this instead?

> When a shared `MessageStore` is used for different correlation endpoints, it is necessary to configure a proper `CorrelationStrategy` to ensure uniqueness for group ids.
> Otherwise unexpected behavior may happen when one correlation endpoint may release or expire messages from others - messages with the same correlation key are stored in the same message group.

Let me know and I'll fix on merge

meherzad:
Looks good.

-----Original Message-----

src/reference/asciidoc/aggregator.adoc
meherzad: Looks good. Thanks.

artembilan: The main my point here that the issue is not only related to the `MessageGroupStoreReaper` and expiration. The aggregator definitely polls the whole group from the store and may decide to release it.

So, you might be surprised that another your aggregator has somehow lost its messages, meanwhile the current one may produce some strange output container.

Gary once said that we are providing `region` option on the message store for this kind of "the same correlation key" scenarios. Right, we have it, but with this IMPORTANT message we should protect ourselves from the configuration when end-user is missing the `region` option or some other flag for distinguishing.

I still don't see how can we catch the problem in the distributed environment, therefore I'm pretty satisfied just having this message in the Reference Manual. 😄 

Merging...

Thank you, @meherzad , for your effort and patience during our review.
Looking forward for more from your side!

artembilan (approved):

garyrussell:
👍 Thanks @meherzad for your assistance with this.

artembilan:
Please, reconsider your Git properties to include your real full name.
Now it is like `meherzad <meherzad4u@gmail.com>`, but I believe it must be like `Meherzad Lahewala <meherzad4u@gmail.com>`.

Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/96aa7280a0700a3c296aeb26443729fc807a7179

Issue #2261, opened 2017-10-10
ScatterGatherHandler.stop() calling GatherEndpoint.start instead of stop


abdulbeard:
https://github.com/spring-projects/spring-integration/blob/00807d47e2880e780b1430c3aba3a588490572da/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java#L164

artembilan:
Great catch! 

Would you mind fixing the issue as contribution: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc ?

Thank you!

abdulbeard:
https://github.com/spring-projects/spring-integration/pull/2262

PR #2262, opened 2017-10-11
Correcting ScatterGatherHandler's internal handling of GatherEndpoint…

abdulbeard:
Correcting ScatterGatherHandler's internal handling of GatherEndpoint as per ISSUE 2261: https://github.com/spring-projects/spring-integration/issues/2261

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/scattergather/config/ScatterGatherParserTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/scattergather/config/ScatterGatherParserTests.java
@@ -105,10 +105,10 @@ public class ScatterGatherParserTests {
 	public void testStop() {
 		MessageHandler scatterGather = this.beanFactory.getBean("scatterGather2.handler", MessageHandler.class);
 		assertSame(this.beanFactory.getBean("gatherChannel"),
-				   TestUtils.getPropertyValue(scatterGather, "gatherChannel"));
+						TestUtils.getPropertyValue(scatterGather, "gatherChannel"));
 		assertNotNull(TestUtils.getPropertyValue(scatterGather, "gatherEndpoint"));
 		Lifecycle gatherEndpoint = TestUtils.getPropertyValue(scatterGather, "gatherEndpoint", Lifecycle.class);
-		assertTrue(((ScatterGatherHandler)scatterGather).isRunning());
+		assertTrue(((ScatterGatherHandler) scatterGather).isRunning());
 		assertTrue(gatherEndpoint.isRunning());
 		((ScatterGatherHandler) scatterGather).stop();
 		assertFalse((gatherEndpoint).isRunning());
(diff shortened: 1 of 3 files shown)


artembilan:
Cool! Thanks!

Would you mind adding some test improvement into the `ScatterGatherParserTests` calling the mentioned `stop()` and assert afterwards that `gatherEndpoint` property is stopped.

abdulbeard:
gatherEndpoint is a private, and is created internally. In the case of `TestDistribution` and `TestAuction` this `gatherEndpoint` is null, which is why `.stop()` has no effect on it.
So a test can't be written for the state of `gatherEndpoint` without editing the `ScatterGatherHandler` class.
Or so it seems to me at least.

artembilan:
That's not true:

https://github.com/spring-projects/spring-integration/blob/master/spring-integration-core/src/test/java/org/springframework/integration/scattergather/config/ScatterGatherParserTests.java#L76

So, that assert confirms that `gatherEndpoint` property is not null.
You can cast `scatterGather` variable to the `Lifecycle` and call its `stop()`:

```
Lifecycle gatherEndpoint = TestUtils.getPropertyValue(scatterGather, "gatherEndpoint", Lifecycle.class);
...
((Lifecycle) scatterGather).stop();
assertFalse(gatherEndpoint.isRunning());
```

Does it work for you?

abdulbeard:
That works for me. PR updated 👍

spring-integration-core/src/test/java/org/springframework/integration/scattergather/config/ScatterGatherParserTests.java
```diff
@@ -17,11 +17,7 @@
 package org.springframework.integration.scattergather.config;
 
 import static org.hamcrest.Matchers.instanceOf;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
artembilan: No asterisk imports, please.

I'll fix on merge tomorrow.

Thank you for the update!

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/scattergather/config/ScatterGatherParserTests.java
abdulbeard: IntelliJ got greedy and did this behind the scenes. Updated now 👍

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f202b257752ce9687735eed5d053db9ff885ecb5 and cherry-picked to `4.3.x`.

@abdulbeard ,

Thank you for contribution.
Looking forward for more!

PR #2265, opened 2017-10-12
Emits event when LockRegistryLeaderInitiator fails to get lock.

cppwfs:
resolves INT-4357

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/leader/event/LeaderEventPublisher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/leader/event/LeaderEventPublisher.java
@@ -23,6 +23,7 @@ import org.springframework.integration.leader.Context;
  *
  * @author Janne Valkealahti
  * @author Gary Russell
+ * @author Glenn Renfro
  *
  */
 public interface LeaderEventPublisher {
@@ -44,5 +45,15 @@ public interface LeaderEventPublisher {
 	 * @param role the role of the leader
 	 */
 	void publishOnRevoked(Object source, Context context, String role);
+
+	/**
+	 * Publish a failure to acquire lock event.
+	 *
+	 * @param source the component generated this event
+	 * @param context the context associated with event
+	 * @param role the role of the leader
+	 */
+	void publishOnFailedToAcquireLock(Object source, Context context, String role);
+
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/leader/event/LeaderEventPublisher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/leader/event/LeaderEventPublisher.java
@@ -30,7 +30,6 @@ public interface LeaderEventPublisher {
 
 	/**
 	 * Publish a granted event.
-	 *
 	 * @param source the component generated this event
 	 * @param context the context associated with event
 	 * @param role the role of the leader
@@ -39,7 +38,6 @@ public interface LeaderEventPublisher {
 
 	/**
 	 * Publish a revoked event.
-	 *
 	 * @param source the component generated this event
 	 * @param context the context associated with event
 	 * @param role the role of the leader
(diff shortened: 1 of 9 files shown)


spring-integration-core/src/main/java/org/springframework/integration/leader/event/LeaderEventPublisher.java
```diff
@@ -45,4 +46,14 @@
 	 */
 	void publishOnRevoked(Object source, Context context, String role);
 
+	/**
+	 * Publish a failure to acquire lock event.
+	 *
+	 * @param source the component generated this event
+	 * @param context the context associated with event
+	 * @param role the role of the leader
+	 */
```
artembilan: Need `@since 5.0`
And no nay blank lines in the method JavaDocs.
All the affected classes should be updated for the current year in their Copyright.

(I can fix all on merge 😄 )

spring-integration-core/src/main/java/org/springframework/integration/leader/event/LeaderEventPublisher.java
```diff
@@ -45,4 +46,14 @@
...
+	void publishOnFailedToAcquireLock(Object source, Context context, String role);
```
artembilan: Well, the `LeaderEventPublisher` is used not only from the `LockRegistryLeaderInitiator` and there is really might be some state in other implementation which may omit similar even.
I really would like to make it as generic as possible then, even if it is right now only for locks.
Therefore the name like `publishOnFailedToAcquire()` looks sufficient and enough to cover locks case as well.

spring-integration-core/src/main/java/org/springframework/integration/leader/event/OnFailedToAcquireLockEvent.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.integration.leader.Context;
+
+/**
+ * Generic event representing that a lock could not be acquired during leader election.
+ *
+ * @author Glenn Renfro
+ *
+ * @since 5.0.0
+ */
+public class OnFailedToAcquireLockEvent extends AbstractLeaderEvent {
```
artembilan: As I talked above let's try to avoid the tying with *lock* entity!
Maybe `mutex` would be better?..
Seems for me it fits to the "lock" subject and is used in the Curator for the Zookeeper `LeaderInitiator`

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -111,6 +112,13 @@
 	 */
 	private long busyWaitMillis = DEFAULT_BUSY_WAIT_TIME;
 
+	/**
+	 * A boolean that if true, failures that occur during leader election will
+	 * be published to the specified applicationEventPublisher.  If false,
+	 * no failures will be published.  Default is false.
```
artembilan: Sounds like should be moved to the setters JavaDoc.
Plus `@since 5.0`

Also need to give people a hint "why?".
Really: why can't we emit such an event unconditionally all the time we fail to obtain the lock?

Thanks

artembilan (changes requested):
Nothing critical: just my regular nit-picking.

Also would be great to have some docs on the matter in the `endpoint.adoc`. There is the `leadership-event-handling` in the end. And that doc must explain this new event type, when it happens and why do we need to react to it. Your use-case should be enough.

Thank you, Glenn!

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
cppwfs: Will move the comment to the setPublishFailedEvents.
The flag was Gary's idea, so we won't flood the users with these events when its in the loop.

spring-integration-core/src/main/java/org/springframework/integration/leader/event/LeaderEventPublisher.java
cppwfs: :thumbsup:

cppwfs:
Just updated the PR with code review comments and updated the adoc as well.  Thanks for the review!!!!

artembilan (approved):
LGTM
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/1979f91cf544203d5c280f5b20d130cb48f3196a

PR #2266, opened 2017-10-15
INT-4081 Inject PreparedStatementSetter in Store

meherzad:
JIRA - INT-4081

Allow to inject PreparedStatementSetter in JdbcMessageChannelStore. Introduce setter to inject PreparedStatementSetter similar to row mapper for a customized statement.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
@@ -260,7 +260,7 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 	}
 
 	/**
-	 * Allows for passing in a custom {@link MessageGroupPreparedStatementSetter}. 
+	 * Allows for passing in a custom {@link MessageGroupPreparedStatementSetter}.
 	 * The {@link MessageGroupPreparedStatementSetter} is used to insert message in the database.
 	 *
 	 * @param messageGroupPreparedStatementSetter Must not be null
@@ -400,11 +400,11 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 
 	/**
 	 * Check mandatory properties ({@link DataSource} and
-	 * {@link #setChannelMessageStoreQueryProvider(ChannelMessageStoreQueryProvider)}). If no {@link MessageRowMapper} 
+	 * {@link #setChannelMessageStoreQueryProvider(ChannelMessageStoreQueryProvider)}). If no {@link MessageRowMapper}
 	 * and {@link MessageGroupPreparedStatementSetter} was explicitly set using
-	 * {@link #setMessageRowMapper(MessageRowMapper)} and 
+	 * {@link #setMessageRowMapper(MessageRowMapper)} and
 	 * {@link #setMessageGroupPreparedStatementSetter(MessageGroupPreparedStatementSetter)}  respectively, the default
-	 * {@link MessageRowMapper} and {@link MessageGroupPreparedStatementSetter} will be instantiate using the 
+	 * {@link MessageRowMapper} and {@link MessageGroupPreparedStatementSetter} will be instantiate using the
 	 * specified {@link #deserializer} and {@link #lobHandler}.
 	 *
 	 * Also, if the jdbcTemplate's fetchSize property ({@link JdbcTemplate#getFetchSize()})
@@ -447,7 +447,8 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 
 				if (JdbcChannelMessageStore.this.priorityEnabled && priority != null) {
 					preparedStatement.setInt(5, priority);
-				} else {
+				}
+				else {
 					preparedStatement.setNull(5, Types.NUMERIC);
 				}
 
@@ -471,7 +472,8 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 		try {
 			this.jdbcTemplate.update(getQuery(this.channelMessageStoreQueryProvider.getCreateMessageQuery()),
 					ps -> this.messageGroupPreparedStatementSetter.setValues(ps, message, groupId));
-		} catch (DuplicateKeyException e) {
+		}
+		catch (DuplicateKeyException e) {
 			if (logger.isDebugEnabled()) {
 				String messageId = getKey(message.getHeaders().getId());
 				logger.debug("The Message with id [" + messageId + "] already exists.\nIgnoring INSERT...");
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
@@ -427,7 +427,8 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 		}
 
 		if (this.messageGroupPreparedStatementSetter == null) {
-			this.messageGroupPreparedStatementSetter = new MessageGroupPreparedStatementSetter(serializer, lobHandler);
+			this.messageGroupPreparedStatementSetter = new MessageGroupPreparedStatementSetter(this.serializer,
+					this.lobHandler);
 		}
 		this.jdbcTemplate.afterPropertiesSet();
 	}
@@ -447,7 +448,8 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 			this.jdbcTemplate.update(getQuery(this.channelMessageStoreQueryProvider.getCreateMessageQuery()),
 					ps -> this.messageGroupPreparedStatementSetter.setValues(ps, message, groupId, this.region,
 							this.priorityEnabled));
-		} catch (DuplicateKeyException e) {
+		}
+		catch (DuplicateKeyException e) {
 			if (logger.isDebugEnabled()) {
 				String messageId = getKey(message.getHeaders().getId());
 				logger.debug("The Message with id [" + messageId + "] already exists.\nIgnoring INSERT...");
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
@@ -40,7 +40,6 @@ import org.springframework.core.serializer.Deserializer;
 import org.springframework.core.serializer.Serializer;
 import org.springframework.core.serializer.support.SerializingConverter;
 import org.springframework.dao.DuplicateKeyException;
-import org.springframework.integration.IntegrationMessageHeaderAccessor;
 import org.springframework.integration.jdbc.store.channel.ChannelMessageStoreQueryProvider;
 import org.springframework.integration.jdbc.store.channel.MessageGroupPreparedStatementSetter;
 import org.springframework.integration.jdbc.store.channel.MessageRowMapper;
@@ -428,32 +427,7 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 		}
 
 		if (this.messageGroupPreparedStatementSetter == null) {
-			this.messageGroupPreparedStatementSetter = (preparedStatement, message, groupId) -> {
-				String groupKey = getKey(groupId);
-				long createdDate = System.currentTimeMillis();
-				String messageId = getKey(message.getHeaders().getId());
-				byte[] messageBytes = this.serializer.convert(message);
-
-				if (logger.isDebugEnabled()) {
-					logger.debug("Inserting message with id key=" + messageId);
-				}
-
-				preparedStatement.setString(1, messageId);
-				preparedStatement.setString(2, groupKey);
-				preparedStatement.setString(3, this.region);
-				preparedStatement.setLong(4, createdDate);
-
-				Integer priority = message.getHeaders().get(IntegrationMessageHeaderAccessor.PRIORITY, Integer.class);
-
-				if (JdbcChannelMessageStore.this.priorityEnabled && priority != null) {
-					preparedStatement.setInt(5, priority);
-				}
-				else {
-					preparedStatement.setNull(5, Types.NUMERIC);
-				}
-
-				this.lobHandler.getLobCreator().setBlobAsBytes(preparedStatement, 6, messageBytes);
-			};
+			this.messageGroupPreparedStatementSetter = new MessageGroupPreparedStatementSetter(serializer, lobHandler);
 		}
 		this.jdbcTemplate.afterPropertiesSet();
 	}
@@ -471,9 +445,9 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 	public MessageGroup addMessageToGroup(Object groupId, final Message<?> message) {
 		try {
 			this.jdbcTemplate.update(getQuery(this.channelMessageStoreQueryProvider.getCreateMessageQuery()),
-					ps -> this.messageGroupPreparedStatementSetter.setValues(ps, message, groupId));
-		}
-		catch (DuplicateKeyException e) {
+					ps -> this.messageGroupPreparedStatementSetter.setValues(ps, message, groupId, this.region,
+							this.priorityEnabled));
+		} catch (DuplicateKeyException e) {
 			if (logger.isDebugEnabled()) {
 				String messageId = getKey(message.getHeaders().getId());
 				logger.debug("The Message with id [" + messageId + "] already exists.\nIgnoring INSERT...");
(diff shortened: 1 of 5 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,43 @@
...
+ * {@link JdbcChannelMessageStore}.
+ *
+ * @author Meherzad Lahewala
+ * @since 5.0
+ * @see PreparedStatementSetter
+ */
+@FunctionalInterface
+public interface MessageGroupPreparedStatementSetter {
+
+	void setValues(PreparedStatement ps, Message<?> requestMessage, Object groupId) throws SQLException;
```
artembilan: I would say that this is very high from the requirements perspective. A target developer indeed must know all the `JdbcChannelMessageStore` internals to implement this properly.

I would suggest an approach similar to the `MessageRowMapper` and definitely populate the required statement fields  in this external class. I would make `serializer` and `lobHandler` as optional properties do not perform those hard operation in the custom implementation and leave the rest of logic for the end-user code.
The `priorityEnabled` should be as one more method argument.
The empty ctor  (with nulls for the `serializer` and `lobHandler`) should be protected to be available only for the implementors.

artembilan (changes requested):
Great job of course, @meherzad !

In addition we need `What's New` note and some description of the feature in the `jdbc.adoc`

meherzad:
@artembilan have made the changes as requested. I have included region also as one of the argument in the method. Update the whatnew and jdbc doc.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -18,26 +18,92 @@
...
+		this.lobHandler = lobHandler;
+	}
+
+	/**
+	 * Default Constructor using default serializer as
+	 * {@link SerializingConverter} and lobHandler as {@link DefaultLobHandler}.
+	 */
+	protected MessageGroupPreparedStatementSetter() {
+		this.serializer = new SerializingConverter();
+		this.lobHandler = new DefaultLobHandler();
```
artembilan: Thanks, @meherzad , for update.

Well, that's not what I want. I really would like to see them both as `null` here and add null checks to the `setValues()` method around them.
This way an implementor may decide to get some other logic for the message to store, e.g. convert to simple string representation. This way (s)he is able to override `setValues()` and don't count with serializer/lobHandler.

I think the way with nul checks we somehow minimize the time for the actions under `preparedStatement`.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -289,3 +289,8 @@ See <<ip>> for more information.
 The `GemfireMetadataStore` now implements `ListenableMetadataStore`, allowing users to listen to cache events by providing `MetadataStoreListener` instances to the store.
 
 See <<gemfire>> for more information.
+
+==== Jdbc Message Channel Store Changes
+
+The `JdbcMessageChannelStore` now provides setter to set `PreparedStatementSetter`, allowing users to easily customize
+`PreparedStatementSetter` in the store.
```
artembilan: In the end need a like to the target chapter:
```
See <<jdbc-message-store-channels>> for more information.
```

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -18,26 +18,92 @@
...
  * @author Meherzad Lahewala
  * @since 5.0
  * @see PreparedStatementSetter
  */
-@FunctionalInterface
-public interface MessageGroupPreparedStatementSetter {
+public class MessageGroupPreparedStatementSetter {
 
-	void setValues(PreparedStatement ps, Message<?> requestMessage, Object groupId) throws SQLException;
+	private static final Log logger = LogFactory.getLog(MessageGroupPreparedStatementSetter.class);
```
artembilan: Let's remove this logging hooks altogether!

artembilan (changes requested):
Nothing critical, but still need some improvements.

Thanks

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
meherzad: My only thinking to have serializer/lob handler as final to ensure the variables are not changed. But understand your point what of these variables are not required.

src/reference/asciidoc/whats-new.adoc
meherzad: Ill check and update.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
artembilan: Right, we just make like this:
```
protected MessageGroupPreparedStatementSetter() {
		this.serializer = null;
		this.lobHandler = null;
}
```
If you don't need them as `null`, you have another ctor for choice.

meherzad:
@artembilan Have fixed all the open items. Build is green.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -255,6 +259,19 @@ public void setMessageRowMapper(MessageRowMapper messageRowMapper) {
 	}
 
 	/**
+	 * Allows for passing in a custom {@link MessageGroupPreparedStatementSetter}.
+	 * The {@link MessageGroupPreparedStatementSetter} is used to insert message in the database.
```
artembilan: message**s** in**to** the database

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -255,6 +259,19 @@ public void setMessageRowMapper(MessageRowMapper messageRowMapper) {
 	}
 
 	/**
+	 * Allows for passing in a custom {@link MessageGroupPreparedStatementSetter}.
```
artembilan: Set/Inject a custom.

But JavaDocs on methods must definitely be in imperative.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -255,6 +259,19 @@ public void setMessageRowMapper(MessageRowMapper messageRowMapper) {
...
+	 * The {@link MessageGroupPreparedStatementSetter} is used to insert message in the database.
+	 *
+	 * @param messageGroupPreparedStatementSetter Must not be null
```
artembilan: The param must have some description - `not null` is not enough.
Also `@since 5.0`

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,97 @@
...
+ * Callback to be used with {@link JdbcChannelMessageStore}.
+ * <p>
+ * Behavior is same as standard {@link PreparedStatementSetter}, it takes in
+ * extra {@code Message<?> requestMessage}, {@code Object groupId},
+ * {@code String region} and {@code boolean priorityEnabled} as parameters used
+ * for {@code addMessageToGroup} method in {@link JdbcChannelMessageStore}.
+ *
+ * @author Meherzad Lahewala
+ * @since 5.0
+ * @see PreparedStatementSetter
```
artembilan: If you use a class only in the JavaDocs, the fully qualified name must be here.
Otherwise we end up with the `Unused import` during JavaDocs generation phase.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public class MessageGroupPreparedStatementSetter {
+
+	private static final Log logger = LogFactory.getLog(MessageGroupPreparedStatementSetter.class);
```
artembilan: Let's remove the extra noise with the logger!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,97 @@
...
+
+	protected MessageGroupPreparedStatementSetter() {
+		this.serializer = null;
+		this.lobHandler = null;
+	}
+
+	public void setValues(PreparedStatement preparedStatement, Message<?> requestMessage, Object groupId, String region,
+			boolean priorityEnabled) throws SQLException {
+		Assert.notNull(this.serializer, "The serializer should not be null.");
+		Assert.notNull(this.lobHandler, "The lobhandler should not be null.");
```
artembilan: I guess you don't have appropriate test on the matter.
We agreed before that have them as `null` is OK by design for the inheritors who don't care about this extra steps.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,97 @@
...
+		String groupKey = getKey(groupId);
+		long createdDate = System.currentTimeMillis();
+		String messageId = getKey(requestMessage.getHeaders().getId());
+		byte[] messageBytes = this.serializer.convert(requestMessage);
```
artembilan: This must be wrapped with the `if (this.serializer != null)`

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,97 @@
...
+
+		if (logger.isDebugEnabled()) {
+			logger.debug("Inserting message with id key=" + messageId);
```
artembilan: No logs in this class, please

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,97 @@
...
+
+		Integer priority = requestMessage.getHeaders().get(IntegrationMessageHeaderAccessor.PRIORITY, Integer.class);
+
+		if (priorityEnabled && priority != null) {
+			preparedStatement.setInt(5, priority);
+		}
+		else {
+			preparedStatement.setNull(5, Types.NUMERIC);
+		}
+		this.lobHandler.getLobCreator().setBlobAsBytes(preparedStatement, 6, messageBytes);
```
artembilan: This with the `if (this.lobHandler != null)` as well

src/reference/asciidoc/whats-new.adoc
```diff
@@ -289,3 +289,10 @@ See <<ip>> for more information.
 The `GemfireMetadataStore` now implements `ListenableMetadataStore`, allowing users to listen to cache events by providing `MetadataStoreListener` instances to the store.
 
 See <<gemfire>> for more information.
+
+==== Jdbc Message Channel Store Changes
+
+The `JdbcMessageChannelStore` now provides setter to set `PreparedStatementSetter`, allowing users to easily customize
```
artembilan: to set `MessageGroupPreparedStatementSetter`

src/reference/asciidoc/jdbc.adoc
```diff
@@ -365,6 +365,9 @@ If your database is not listed, you can easily extend the `AbstractChannelMessag
 
 Since _version 4.0_, the `MESSAGE_SEQUENCE` column has been added to the table to ensure first-in-first-out (FIFO) queueing even when messages are stored in the same millisecond.
 
+Since_version 5.0, by overloading `MessageGroupPreparedStatementSetter` class you can provide custom implementation to 
+ `PreparedStatementSetter` to support different columns or table structure.
```
artembilan: Well, it is somehow wrong to say `PreparedStatementSetter` because the `MessageGroupPreparedStatementSetter` doesn't implement it, really. 

> Starting with _version 5.0_, the `JdbcChannelMessageStore` can be supplied with the custom `MessageGroupPreparedStatementSetter`  to support different columns or table structure or data types.
For example, instead of default message serialization to the `byte[]`, we can store its structure as JSON string.

artembilan (changes requested):

src/reference/asciidoc/jdbc.adoc
meherzad: Yes. But the functionality is almost similar.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
meherzad: What should go in the else case then? We need to store the message. Should we just the `messageBytes` if available or create text representation from the request message?

artembilan: We left that up to the target implementation. And that’s why I requested it is as a `protected` ctor. More over you say yourself: in case of different table structure. 

One more time: my idea to let end-user to call `super` for common columns, but use custom logic e.g. for different message serialization way.

src/reference/asciidoc/jdbc.adoc
artembilan: Right, you say that in its JavaDocs, but that’s fully wrong to mislead target developer to say him/her to implement wrong contruct. That really might be also good to add simple sample here in Doc. For that original idea for just simple storing message in varchar column, leaving the rest in the same structure.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
```diff
@@ -0,0 +1,88 @@
...
+ * Behavior is same as standard {@link org.springframework.jdbc.core.PreparedStatementSetter}, it takes in
+ * extra {@code Message<?> requestMessage}, {@code Object groupId},
+ * {@code String region} and {@code boolean priorityEnabled} as parameters used
+ * for {@code addMessageToGroup} method in {@link org.springframework.integration.jdbc.store.JdbcChannelMessageStore}.
+ *
+ * @author Meherzad Lahewala
+ * @since 5.0
+ * @see org.springframework.jdbc.core.PreparedStatementSetter
+ */
+public class MessageGroupPreparedStatementSetter {
```
artembilan: Well, after closer look, it really isn't a `group`. It just does nothing with groups. This is exactly about insertion a single message.
Will rename to the `MessagePreparedStatementSetter` during merge...

artembilan (approved):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageGroupPreparedStatementSetter.java
artembilan: Hm. We have already one - `org.springframework.integration.jdbc.MessagePreparedStatementSetter`.
so, no choice: this one will be `ChannelMessageStorePreparedStatementSetter` - to really distinguish this one for target purpose and don't let end-user to be confused.

artembilan:
@meherzad ,

Please, consider to fix your Git config for you real name - Meherzad Lahewala.
Right now it is like - `meherzad <meherzad4u@gmail.com>`, but would be better to reflect your proper personality.

Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/34d0859515d68ab213b7e97ac4d526e4e374efe5 with subsequent polishing and improvements like https://github.com/spring-projects/spring-integration/commit/d7d56a04445a04e0b8e9757e8bdb2dc7ee4bce4a.

Thank you for your contribution.
Looking forward for more!
And hope my polishing isn't so offending since I've almost overwritten your code 😄

meherzad:
@artembilan Thanks for pointing out the embarrassing email in the git config. 

No issues with the polishing part, have gone through it, good learning for me.

PR #2270, opened 2017-10-26
Some Java DSL improvements

artembilan:
* Register `BeanDefinition`s for DSL components.
That way we get a gain when the `BeanDefinition` is involved,
e.g. Spring Cloud Function uses this approach to determine the generics
of the target `Function` class
* This programmatic `BeanDefinition` registration approach
(available since SF-5.0) allows us to avoid some manual lifecycle processes:
bean initialization, autowiring, event listener registration etc.
* Rework `IntegrationFlowContext` to register and remove `BeanDefinition`s as well
* Remove unused `registerComponents`  flag from the `StandardIntegrationFlow`.
With the proper logic in the `IntegrationFlowContext` it does not make sense any more
* Remove redundant `AnnotationGatewayProxyFactoryBean` in the `IntegrationFlows`;
populate `defaultRequestChannel` directly to the `GatewayProxyFactoryBean` instance
* Add an overloaded `IntegrationFlows.from(Class, String)` to allow to specify an
explicit bean name for the target gateway proxy
* Extract and populate bean name from the `@MessagingGateway.name()` in the
`AnnotationGatewayProxyFactoryBean`
* Stop lifecycles in the `AmqpTests` after using for proper test suit shutdown

### DIFF
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/dsl/AmqpTests.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/dsl/AmqpTests.java
@@ -40,6 +40,7 @@ import org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer;
 import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.Lifecycle;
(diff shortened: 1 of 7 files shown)


artembilan:
@olegz ,

This is what we talked with you about `BeanDefinition` in case of SI Java DSL.

Although there is still problem in Spring Cloud Function: https://github.com/spring-cloud/spring-cloud-function/issues/118. I don't consider here in SI any struts because there is no fault in SI on the matter.

Thank you for your attention!

PR #2272, opened 2017-10-30
INT-4360: Add `ClientStompEncoder` support

artembilan:
The `StompSubProtocolHandler` explicitly sets `stompCommand` header
to the `MESSAGE` value ignoring any client inputs.
In this case the message is treated as from the server and ignored on
the STOMP Broker side from the client session.

* Introduce `ClientStompEncoder` for the client side to be injected
into the `StompSubProtocolHandler` for the proper client side messages
encoding/decoding.
Override `stompCommand` header to the `SEND` value if it is `MESSAGE`
before encoding to the `byte[]` to send to the session

JIRA: https://jira.spring.io/browse/INT-4360

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-websocket/src/test/java/org/springframework/integration/websocket/client/StompIntegrationTests.java
+++ b/spring-integration-websocket/src/test/java/org/springframework/integration/websocket/client/StompIntegrationTests.java
@@ -55,7 +55,6 @@ import org.springframework.integration.channel.QueueChannel;
 import org.springframework.integration.config.EnableIntegration;
 import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.event.inbound.ApplicationEventListeningMessageProducer;
-import org.springframework.integration.test.support.LogAdjustingTestSupport;
 import org.springframework.integration.transformer.ExpressionEvaluatingTransformer;
 import org.springframework.integration.websocket.ClientWebSocketContainer;
 import org.springframework.integration.websocket.IntegrationWebSocketContainer;
@@ -85,7 +84,7 @@ import org.springframework.messaging.support.MessageBuilder;
 import org.springframework.stereotype.Controller;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.socket.client.WebSocketClient;
 import org.springframework.web.socket.client.standard.StandardWebSocketClient;
@@ -108,10 +107,10 @@ import org.springframework.web.socket.sockjs.client.WebSocketTransport;
  *
  * @since 4.1
  */
-@ContextConfiguration
-@RunWith(SpringJUnit4ClassRunner.class)
-@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
-public class StompIntegrationTests extends LogAdjustingTestSupport {
+@ContextConfiguration(classes = StompIntegrationTests.ClientConfig.class)
+@RunWith(SpringRunner.class)
+@DirtiesContext
+public class StompIntegrationTests {
 
 	@Value("#{server.serverContext}")
 	private ApplicationContext serverContext;
@@ -130,11 +129,6 @@ public class StompIntegrationTests extends LogAdjustingTestSupport {
 	@Autowired
 	@Qualifier("webSocketEvents")
 	private QueueChannel webSocketEvents;
-
-	public StompIntegrationTests() {
-		super("org.springframework", "org.springframework.integration", "org.apache.catalina");
-	}
-
 
 	@Test
 	public void sendMessageToController() throws Exception {
@@ -334,7 +328,7 @@ public class StompIntegrationTests extends LogAdjustingTestSupport {
 
 	@Configuration
 	@EnableIntegration
-	public static class ContextConfiguration {
+	public static class ClientConfig {
 
 		@Bean
 		public TomcatWebSocketTestServer server() {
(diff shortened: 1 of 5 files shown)


spring-integration-websocket/src/test/java/org/springframework/integration/websocket/client/StompIntegrationTests.java
```diff
@@ -350,7 +353,9 @@ public IntegrationWebSocketContainer clientWebSocketContainer() {
 
 		@Bean
 		public SubProtocolHandler stompSubProtocolHandler() {
-			return new StompSubProtocolHandler();
+			StompSubProtocolHandler stompSubProtocolHandler = new StompSubProtocolHandler();
+			stompSubProtocolHandler.setEncoder(new ClientStompEncoder());
+			return stompSubProtocolHandler;
 		}
```
garyrussell: Shouldn't we use a different protocol handler on the client/server side? This should only be used on the client side, right?

Also, the test is checking the switch occurred on the server side, which seems wrong to me.

Surely, on the inbound side, we are expecting a `MESSAGE` frame.

By the way, these `sendMessageToController*` tests always fail for me in my IDE; it seems the event listener is not registered/started...

>15:18:58.747 TRACE [WebSocketClient-AsyncIO-1][org.springframework.context.support.GenericApplicationContext] Publishing event in org.springframework.context.support.GenericApplicationContext@2d2e5f00: SessionConnectedEvent[GenericMessage [payload=byte[0], headers={simpMessageType=OTHER, stompCommand=CONNECTED, nativeHeaders={heart-beat=[0,0]}, simpSessionAttributes={}, simpHeartbeat=[J@1bad8bb9, simpSessionId=53826b559db84ae3afe7e3a9309ba9ae}]]
>15:19:18.712 TRACE [main][org.springframework.integration.channel.QueueChannel] postReceive on channel 'webSocketEvents', message is null


Works in gradle...

>15:18:28.376 TRACE [WebSocketClient-AsyncIO-1][org.springframework.context.support.GenericApplicationContext] Publishing event in org.springframework.context.support.GenericApplicationContext@3aa3b61a: SessionConnectedEvent[GenericMessage [payload=byte[0], headers={simpMessageType=OTHER, stompCommand=CONNECTED, nativeHeaders={heart-beat=[0,0]}, simpSessionAttributes={}, simpHeartbeat=[J@21741628, simpSessionId=06af94fa8ce54e14afdc9bd185433e51}]]
15:18:28.377 DEBUG [WebSocketClient-AsyncIO-1][org.springframework.beans.factory.support.DefaultListableBeanFactory] Returning cached instance of singleton bean 'webSocketEventListener'
15:18:28.377 DEBUG [WebSocketClient-AsyncIO-1][org.springframework.integration.channel.QueueChannel] preSend on channel 'webSocketEvents', message: GenericMessage [payload=SessionConnectedEvent[GenericMessage [payload=byte[0], headers={simpMessageType=OTHER, stompCommand=CONNECTED, nativeHeaders={heart-beat=[0,0]}, simpSessionAttributes={}, simpHeartbeat=[J@21741628, simpSessionId=06af94fa8ce54e14afdc9bd185433e51}]], headers={id=f1ba3bbd-21d0-21fe-8933-d411ae4f26b6, timestamp=1509391108377}]
15:18:28.377 DEBUG [Test worker][org.springframework.integration.channel.QueueChannel] postReceive on channel 'webSocketEvents', message: GenericMessage [payload=SessionConnectedEvent[GenericMessage [payload=byte[0], headers={simpMessageType=OTHER, stompCommand=CONNECTED, nativeHeaders={heart-beat=[0,0]}, simpSessionAttributes={}, simpHeartbeat=[J@21741628, simpSessionId=06af94fa8ce54e14afdc9bd185433e51}]], headers={id=f1ba3bbd-21d0-21fe-8933-d411ae4f26b6, timestamp=1509391108377}]

...investigating...

garyrussell (changes requested):

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/client/StompIntegrationTests.java
artembilan: > Shouldn't we use a different protocol handler on the client/server side? 

We do. This `StompIntegrationTests` has `ContextConfiguration` - for client side. See there `clientWebSocketContainer`  bean for both channel adapters. And there is a separate `ServerConfig` picked up by the `TomcatWebSocketTestServer` bean definition.

>  the test is checking the switch occurred on the server side, which seems wrong to me.

The test indeed checks the state on the server side. See controller usage:
```
SimpleController controller = this.serverContext.getBean(SimpleController.class);
```
The goal is to be sure that client has sent the proper frame and server accepts it without any modifications.

> Surely, on the inbound side, we are expecting a MESSAGE frame.

Not sure what you mean since we test on server messages sent by the web socket client.

> By the way, these sendMessageToController* tests always fail for me in my IDE;

No, for me it's fine on the Windows...

garyrussell:
Interesting (weird) - when running in eclipse, if I `@Autowire` the `webSocketEventListener`, I get the bean from the server context - so it appears the events are going there (and ignored). If I change the client and server event listener beans to have different bean names, the tests pass ok.

Some kind of weird interaction with the embedded application context in the test tomcat server.

garyrussell:
I see what's happening - the test context has both beans; when they have the same name, one overwrites the other. spring-test loads both configs.

This is the proper fix...

    @ContextConfiguration(classes = StompIntegrationTests.Config.class)

(renamed config class and explicit reference).

artembilan:
Pushed the fix for tests.

Thank you Gary for pointing that out! 😄

garyrussell:
Merged as 8766262399ce721d1b9bf5d2fc866c6cd24bb2f3; cherry-picked as 26819f5a950bad6bdd957363195b7d1ec342b72d

PR #1483, opened 2015-06-24
INT-3617: ZK Leader Event Processing

garyrussell:
JIRA: https://jira.spring.io/browse/INT-3617

Start/stop `SmartLifecyle` beans with leader election/revocation.

**needs my PRs to spring-cloud-cluster:** https://github.com/spring-cloud/spring-cloud-cluster/pulls

(local mvn package install)

PR #2277, opened 2017-11-07
INT-4361: Add a role() hook to Java DSL

artembilan:
JIRA: https://jira.spring.io/browse/INT-4361

* To get a gain of the method chain code flow and avoid extra annotation
parsing, add `.role(String)` hook to the `EndpointSpec`
* Delegate the provided `role` property to the `AbstractEndpoint`
* Register `AbstractEndpoint` as itself `SmartLifecycle` in the
`SmartLifecycleRoleController`
* Add `destroy()` to the `AbstractEndpoint` and remove it from the
`SmartLifecycleRoleController`
* Provide some Java 8 code style refactoring

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
@@ -20,6 +20,7 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.SmartLifecycle;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.context.IntegrationObjectSupport;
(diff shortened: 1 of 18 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
```diff
@@ -84,6 +108,21 @@ protected void onInit() throws Exception {
 				}
 			}
 		}
+
+		if (StringUtils.hasText(this.role)) {
+			this.roleController = getBeanFactory()
+					.getBean(IntegrationContextUtils.INTEGRATION_LIFECYCLE_ROLE_CONTROLLER,
+					SmartLifecycleRoleController.class);
+
+			this.roleController.addLifecycleToRole(this.role, this);
```
garyrussell: Needs null check - also, if endpoints now register themselves, shouldn't we use the parsers to use the same technique? https://github.com/spring-projects/spring-integration/pull/1483/files#diff-1fda9b32dc8bd19e9984d90ebf376ef6

artembilan: Yeah... We have to catch:
```
catch (NoSuchBeanDefinitionException e) {
		this.logger.error("No LifecycleRoleController in the context");
}
```

Well, we can't use `addLifecycleToRole()` from the XML parser since we don't have bean instances there.
The manipulation with the `BeanDefinition` attributes is correct there.
IMO...

garyrussell: ? Can't we just set the `role` property on the endpoint BD from the xml?

artembilan:
@garyrussell ,

Please, take a look.

That indeed was good suggestion to go.
This way I have observed several `AbstractEndpoint` implemetation which didn't delegate to `super.onInit()`.

I guess we need to back port that fix to `4.3.x` 😄

garyrussell:
Merged as f7e75223c79996e1350c87f35724491adfe6b39b

Issue #2268, opened 2017-10-24
RedisInboundChannelAdapter "drops" channel ID


matthjes:
Hi,

The current implementation of the `RedisInboundChannelAdapter` does not add the channel ID to the header when a message is received. This makes it impossible to filter / route messages based on the topic / channel ID. Though I can subscribe to multiple topics, I do not get any information about which message belongs to which topic.

I have a small fix for this (see below). You only have to change the `handleMessage()` method of the MessageListenerDelegate to accept the channel ID as a parameter and then also change the `convertMessage` appropriately. Note that the default `MessageConverter` is a `SimpleMessageConverter`, which also ignores the header, so you have to use a custom `MessageConverter`.

```java
public class RedisInboundChannelAdapter extends MessageProducerSupport {
    ...
    private Message<?> convertMessage(Object object, String channel) {
        Map<String, Object> headers = new HashMap<>();
        headers.put("topic", channel);
        return this.messageConverter.toMessage(object, new MessageHeaders(headers));
    }

    private class MessageListenerDelegate {

	@SuppressWarnings("unused")
	public void handleMessage(Object object, String channel) {
		sendMessage(convertMessage(object, channel));
    }
}
```

I'll submit a PR when I find the time to fix this.

artembilan:
The project is called **Spring**.

Please, be careful if you would like to get a help from us.

matthjes:
Hm, ok, I‘m deeply deeply sorry for the typo...

artembilan:
I think your request is valid.
Please, go ahead with the PR on the matter.

Consider to add `SOURCE` constant to the `RedisHeaders`.
It must be source because of: https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java#L967

Also we should wrap into the `if (source != null)` and don't create those extra headers object for nothing.

Thank you for considering contribution and proposing an improvement!

PR #2279, opened 2017-11-07
GH-2268: Add RedisHeaders.MESSAGE_SOURCE header

artembilan:
Resolves: /spring-projects/spring-integration#2268

* To indicate the source the Redis message in the `RedisInboundChannelAdapter`
populate the `RedisHeaders.MESSAGE_SOURCE` header to the messages to produce
* Fix the `SimpleMessageConverter` to populate the provided `MessageHeaders`
to the message to produce

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.gateway;
 
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicLong;
 
@@ -792,11 +793,14 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 		}
 
 		@Override
-		public Message<?> toMessage(Object object) throws Exception {
+		public Message<?> toMessage(Object object, @Nullable Map<String, Object> headers) throws Exception {
 			if (object instanceof Message<?>) {
 				return (Message<?>) object;
 			}
-			return (object != null) ? this.messageBuilderFactory.withPayload(object).build() : null;
+
+			return object != null
+					? this.messageBuilderFactory.withPayload(object).copyHeadersIfAbsent(headers).build()
+					: null;
 		}
 
 	}
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/converter/SimpleMessageConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/converter/SimpleMessageConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,14 +28,16 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.support.MessageBuilder;
 
 /**
  * @author Mark Fisher
  * @author Gary Russell
  * @author Artem Bilan
+ *
  * @since 2.0
  */
-@SuppressWarnings({"unchecked", "rawtypes"})
+@SuppressWarnings({ "unchecked", "rawtypes" })
 public class SimpleMessageConverter implements MessageConverter, BeanFactoryAware {
 
 	private volatile InboundMessageMapper inboundMessageMapper;
@@ -103,7 +105,14 @@ public class SimpleMessageConverter implements MessageConverter, BeanFactoryAwar
 	@Override
 	public Message<?> toMessage(Object object, MessageHeaders headers) {
 		try {
-			return this.inboundMessageMapper.toMessage(object);
+			Message message = this.inboundMessageMapper.toMessage(object);
+			if (headers != null) {
+				message =
+						MessageBuilder.fromMessage(message)
+								.copyHeadersIfAbsent(headers)
+								.build();
+			}
+			return message;
 		}
 		catch (Exception e) {
 			throw new MessageConversionException("failed to convert object to Message", e);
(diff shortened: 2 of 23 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/converter/SimpleMessageConverter.java
```diff
@@ -103,7 +105,14 @@ protected MessageBuilderFactory getMessageBuilderFactory() {
...
 	public Message<?> toMessage(Object object, MessageHeaders headers) {
 		try {
-			return this.inboundMessageMapper.toMessage(object);
+			Message message = this.inboundMessageMapper.toMessage(object);
+			if (headers != null) {
+				message =
+						MessageBuilder.fromMessage(message)
+								.copyHeadersIfAbsent(headers)
+								.build();
+			}
```
garyrussell: Hmmm - this will impact every messaging gateway. I wonder if we could have the mapper return the builder (perhaps a `default` method)?

Or provide an overloaded method that takes extra headers.

Or create a new converter that's only used by Redis.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisInboundChannelAdapter.java
```diff
@@ -78,6 +82,12 @@ public void setMessageConverter(MessageConverter messageConverter) {
 		this.messageConverter = messageConverter;
 	}
 
+	/**
+	 * Specify an {@link Executor} used for running the message listeners when messages are received.
+	 * @param taskExecutor the Executor to use for listener container.
+	 * @since 4.3.13
```
garyrussell: Does this __really__ need to be back-ported? There's a work around (header enricher).

src/reference/asciidoc/redis.adoc
```diff
@@ -103,7 +103,7 @@ Instead those Messages are passed through Redis allowing you to rely on its supp
 [[redis-inbound-channel-adapter]]
 ==== Redis Inbound Channel Adapter
 
-The Redis-based Inbound Channel Adapter adapts incoming Redis messages into Spring Integration Messages in the same way as other inbound adapters.
+The Redis-based Inbound Channel Adapter (`RedisInboundChannelAdapter`) adapts incoming Redis messages into Spring Integration Messages in the same way as other inbound adapters.
```
garyrussell: `spring-messaging` messages?

garyrussell (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisInboundChannelAdapter.java
artembilan: Well, that's just a JavaDocs from the previous commit.

We indeed provide this `MESSAGE_SOURCE` feature only for `5.0`.

At the same time I'm not sure how you are going to workaround this header when we already step out of this `RedisInboundChannelAdapter`

spring-integration-core/src/main/java/org/springframework/integration/support/converter/SimpleMessageConverter.java
artembilan: As you see I have switched `RedisInboundChannelAdapter` to use `org.springframework.messaging.converter.SimpleMessageConverter` instead.

But I agree, I'm looking now to something "not very impacting"...

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -792,11 +793,14 @@ void setMessageBuilderFactory(MessageBuilderFactory messageBuilderFactory) {
...
 		@Override
-		public Message<?> toMessage(Object object) throws Exception {
+		public Message<?> toMessage(Object object, @Nullable Map<String, Object> headers) throws Exception {
 			if (object instanceof Message<?>) {
 				return (Message<?>) object;
 			}
-			return (object != null) ? this.messageBuilderFactory.withPayload(object).build() : null;
+
+			return object != null
+					? this.messageBuilderFactory.withPayload(object).copyHeadersIfAbsent(headers).build()
```
garyrussell: We should add `@Nullable` to `copyHeadersIfAbsent` (it's present on the Spring Messaging MB).

artembilan: Is it related to this PR ?

There are too much those Kotlin "love" in SF...

garyrussell: Yes, it's related because you now unconditionally `copyHeadersIfAbsent(headers)` - I had to go and check the code to see if it's protected. `@Nullable` documents it.

artembilan: ??? That was always unconditional because there is `if (headersToCopy != null) {`.

I don't understand your concern regarding this `@Nullable`. 
This is something new in SF `5.0` and we have lived without it for a while already.

What is the point to jumt into this subject now?
Thanks

garyrussell: ? This cHIA () is new code in this commit and the arg is @Nullable.

artembilan: How is that new? 
I just use here existing `MessageBuilder` API.
What am I missing?

Right, I added a new `headers` arg to the `toMessage()` method and just followed its interface contract with adding `@Nullable`. And this is definitely new code. But why do I have to go from here everywhere to fix all the possible `@Nullable`?

Cant't we address it on demand at least?

garyrussell: You don't have to "go everywhere". Since you can send null here I had to check if it was safe. So this is indeed "on demand". This discussion is more effort than it would have taken ro fix it.

artembilan:
Pushed.

Thank you for the patience, Gary, and for the clean explanation what you want to see.
Although I still don't see a gain of all those `@Nullable`s 😄

garyrussell:
Merged as 7ca20e53f78688f46617775b0d45ba8266e817e1

Issue #2207, opened 2017-08-03
Trying to use RedisMetadataStore on a Redis cluster


tkvangorder:
Using
Spring Integration 4.3.8
Spring Boot 1.4.5
Spring Core 4.3.7

We have a cluster of servers that are are monitoring a shared network mount for processing EDI files. We recently added code to use the RedisMetadataStore as follows: 

```java

	@Bean
	public ConcurrentMetadataStore metadataStore() {
        return new RedisMetadataStore(redisConnectionFactory);
     }

	@Bean 
	public FileSystemPersistentAcceptOnceFileListFilter persistentAcceptOnceFileFilter() {
		return new FileSystemPersistentAcceptOnceFileListFilter(metadataStore(), "edi-file-locks");
	}

	@Bean
	public IntegrationFlow flowInboundNetTransferFile(
			@Value("${edi.incoming.directory.netTransfers}") String inboundDirectory,
			@Value("${edi.incoming.age-before-ready-seconds:30}") int ageBeforeReadySeconds,
			@Value("${taskExecutor.inboundFile.corePoolSize:4}") int corePoolSize,
			@Qualifier("taskExecutorInboundFile") TaskExecutor taskExecutor) {
		
		//Setup a filter to only pick up a files older than a certain age, relative to the current time. This prevents cases
		//where something is writing to the file as the EDI processor is moving that file. 
		LastModifiedFileListFilter lastModifiedFilter = new LastModifiedFileListFilter();
		lastModifiedFilter.setAge(ageBeforeReadySeconds);
        return IntegrationFlows
        		.from(
        			Files
        				.inboundAdapter(new File(inboundDirectory))
        				.locker(ediDocumentLocker())
        				.filter(new ChainFileListFilter<File>())
        				.filter(new IgnoreHiddenFileListFilter())
        				.filter(lastModifiedFilter)
        				.filter(persistentAcceptOnceFileFilter()),
        			e -> e.poller(Pollers.fixedDelay(20000).maxMessagesPerPoll(corePoolSize).taskExecutor(taskExecutor)))
                .channel(channelInboundFile())
                .get();		
    }
```

This was working fine in our lower environments, however, we use a Redis cluster in our production environment and when we deployed to that environment we got the following exception: 

```java
org.springframework.dao.InvalidDataAccessApiUsageException: WATCH is currently not supported in cluster mode.
[...]

PR #2281, opened 2017-11-07
GH-2207: Clarify RedisMetadataStore in cluster

artembilan:
Resolves: spring-projects/spring-integration#2207

* Since the `WATCH` command isn't supported on clustered connections,
the `RedisMetadataStore.replace()` operation can't be used there.
Document this limitation


garyrussell:
The exception says `WATCH is currently not supported in cluster mode.` (note `currently`). We should reflect the same or add "at the time of writing, `WATCH` is not supported..."

Can change on merge if you agree.

artembilan:
Sure! Makes sense 😄 

Please, fix on merge.

Thank you for pointing that out!

PR #2283, opened 2017-11-13
INT-4362: MMIH Catch Conversion Exception

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4362

INT-4312 added a JSON conversion when there is a single method with a non-String parameter.

Add more tests for parameters of type `Message`.

Catch conversion exceptions and use the message as-is if the conversion fails.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -259,27 +259,38 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 
 	@SuppressWarnings("unchecked")
 	public T process(Message<?> message) throws Exception {
-		Message<?> messageToProcess = message;
-		/*
-		 * If there's a single method, the content is JSON, the payload is a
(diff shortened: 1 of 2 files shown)


artembilan:
I know that isn't related to your changes, but still would be great to have your double check for the Travis failure report:
```
org.springframework.integration.ip.tcp.TcpConfigOutboundGatewayTests > testOutboundInsideChain FAILED

    org.springframework.messaging.MessagingException at TcpConfigOutboundGatewayTests.java:180

        Caused by: org.springframework.messaging.MessagingException at TcpConfigOutboundGatewayTests.java:180

            Caused by: java.net.ConnectException at TcpConfigOutboundGatewayTests.java:180
```
Thanks

garyrussell:
That requires a large change and should not be attached to this PR. [INT-4364 | https://jira.spring.io/browse/INT-4364].

artembilan (approved):
LGTM

Will be merged when our CI comes back to life.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/70c166fbfdeafafa08a9445f191480ec4a75997d after rebase and resolving conflicts in the `MethodInvokingMessageProcessorTests`

PR #2284, opened 2017-11-13
INT-4364: Deprecate SocketUtils and remove all use

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4364

Always let the OS choose the ports for tests.

Just one test remains (RMI) because there appears to be no way to obtain
the port after creating a registry with a 0 port.

This test has been switched to use the Spring utility.


spring-integration-syslog/src/main/java/org/springframework/integration/syslog/inbound/TcpSyslogReceivingChannelAdapter.java
```diff
@@ -61,7 +71,7 @@ public String getComponentType() {
 	protected void onInit() {
 		super.onInit();
 		if (this.connectionFactory == null) {
-			this.connectionFactory = new TcpNioServerConnectionFactory(this.getPort());
+			this.connectionFactory = new TcpNioServerConnectionFactory(super.getPort());
```
artembilan: I think we don't need this change. 
We have similar structure in the `UdpSyslogReceivingChannelAdapter` below but you don't override the `onInit()` logic there.

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/SocketUtils.java
```diff
@@ -28,15 +28,20 @@
...
+ * @deprecated - it's generally better to set the server port to 0; let the operating
+ * system choose a port; wait until the server starts (see TestingUtilities in the ip
+ * module for an example), then set the port on the client factory.
+ *
  * @author Gunnar Hillert
  * @author Gary Russell
  * @since 2.2
  */
+@Deprecated
 public final class SocketUtils {
```
artembilan: Why just don't remove it?
Otherwise it will wait until `6.0`...

artembilan (changes requested):
Just a couple concerns

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/SocketUtils.java
garyrussell: I don't think it hurts to leave it there until 6.0; we're not using it. I think it's too close to GA to remove it now (and deprecate in 4.3.x).

artembilan (approved):
LGTM
Will merge when CI comes back

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4207f36a1749d16e512d19c1474805d345bfa93e after rebase.

PR #2285, opened 2017-11-14
INT-4363: Configurable MessageHandlerMethodFactory

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4363

The JIRA requests making the `MessageHandlerMethodFactory` a bean.
This is not possible without a major rework of the `MessagingMethodInvokerHelper`.

The problem is that the `InvokerHandlerMethod` s are created before the factory
is initialized. In fact, it only works at all since the IHMs have a hard reference
to the factory's argument resolvers so, when they are initialized in `initialize()`
each handler sees the resolvers.

This really needs to be improved but, since we are so close to GA, the compromise
was to add a `HandlerMethodArgumentResolversHolder` which holds the standard resolvers
which are then wired into the factory (along with the message converter).

Delaying the creation of the `InvokerHandlerMethod` is not trivial; today they are
built in the CTOR; moving it to `initialize()` would prevent the fast failure for
a badly configured bean; the proper solution might be to make the MMIHs beans themselves.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/HandlerMethodArgumentResolversHolder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/HandlerMethodArgumentResolversHolder.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.config;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
+
+/**
+ * A holder for the configured argument resolvers.
+ *
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public class HandlerMethodArgumentResolversHolder {
+
+	private final List<HandlerMethodArgumentResolver> resolvers;
+
+	public HandlerMethodArgumentResolversHolder(List<HandlerMethodArgumentResolver> resolvers) {
+		this.resolvers = new ArrayList<>(resolvers);
+	}
+
+	public List<HandlerMethodArgumentResolver> getResolvers() {
+		return new ArrayList<>(this.resolvers);
+	}
+
+	public void addResolver(HandlerMethodArgumentResolver resolver) {
+		this.resolvers.add(resolver);
+	}
+
+	public boolean removeResolver(HandlerMethodArgumentResolver resolver) {
+		return this.resolvers.remove(resolver);
+	}
+
+}
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/HandlerMethodArgumentResolversHolder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/HandlerMethodArgumentResolversHolder.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.config;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
@@ -37,7 +38,7 @@ public class HandlerMethodArgumentResolversHolder {
 	}
 
 	public List<HandlerMethodArgumentResolver> getResolvers() {
-		return new ArrayList<>(this.resolvers);
+		return Collections.unmodifiableList(this.resolvers);
 	}
 
 	public void addResolver(HandlerMethodArgumentResolver resolver) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -30,7 +30,6 @@ import org.springframework.beans.factory.BeanClassLoaderAware;
 import org.springframework.beans.factory.ListableBeanFactory;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.BeanDefinitionHolder;
-import org.springframework.beans.factory.config.BeanReference;
 import org.springframework.beans.factory.config.PropertiesFactoryBean;
 import org.springframework.beans.factory.config.RuntimeBeanReference;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
@@ -57,7 +56,7 @@ import org.springframework.integration.support.converter.ConfigurableCompositeMe
 import org.springframework.integration.support.converter.DefaultDatatypeChannelMessageConverter;
 import org.springframework.integration.support.utils.IntegrationUtils;
 import org.springframework.messaging.converter.CompositeMessageConverter;
-import org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory;
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
 import org.springframework.util.ClassUtils;
 
 /**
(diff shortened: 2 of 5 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/HandlerMethodArgumentResolversHolder.java
```diff
@@ -0,0 +1,51 @@
...
+public class HandlerMethodArgumentResolversHolder {
+
+	private final List<HandlerMethodArgumentResolver> resolvers;
+
+	public HandlerMethodArgumentResolversHolder(List<HandlerMethodArgumentResolver> resolvers) {
+		this.resolvers = new ArrayList<>(resolvers);
+	}
+
+	public List<HandlerMethodArgumentResolver> getResolvers() {
+		return new ArrayList<>(this.resolvers);
```
artembilan: `Collections.unmodifableList()`?

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -414,14 +423,91 @@ private void registerDefaultDatatypeChannelMessageConverter(BeanDefinitionRegist
 	 * @param registry the registry.
 	 */
 	private void registerArgumentResolverMessageConverter(BeanDefinitionRegistry registry) {
-		if (!registry.containsBeanDefinition(IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME)) {
-			BeanDefinitionBuilder postProcessorBuilder = BeanDefinitionBuilder
+		if (!registry.containsBeanDefinition(
+				IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME)) {
+			BeanDefinitionBuilder resolverBuilder = BeanDefinitionBuilder
```
artembilan: I agree that previous `postProcessorBuilder` name wasn't good, but still isn't `Resolver`. It is `MessageConverter`

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -414,14 +423,91 @@ private void registerDefaultDatatypeChannelMessageConverter(BeanDefinitionRegist
...
 					.genericBeanDefinition(ConfigurableCompositeMessageConverter.class);
 			registry.registerBeanDefinition(IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME,
-					postProcessorBuilder.getBeanDefinition());
+					resolverBuilder.getBeanDefinition());
+		}
+	}
+
+	/**
+	 * Register the default {@link MessageHandlerMethodFactory} for handler method
```
artembilan: Looks like this method doesn't register `MessageHandlerMethodFactory`

spring-integration-core/src/main/java/org/springframework/integration/config/HandlerMethodArgumentResolversHolder.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
+
+/**
+ * A holder for the configured argument resolvers.
+ *
+ * @author Gary Russell
+ * @since 5.0
+ *
+ */
+public class HandlerMethodArgumentResolversHolder {
```
artembilan: Can't we use `HandlerMethodArgumentResolverComposite` from SF instead?
Or there might be really a clash that it is a `HandlerMethodArgumentResolver` per se?

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -414,14 +423,91 @@ private void registerDefaultDatatypeChannelMessageConverter(BeanDefinitionRegist
...
+				IntegrationContextUtils.LIST_ARGUMENT_RESOLVERS_BEAN_NAME)) {
+			registry.registerBeanDefinition(IntegrationContextUtils.LIST_ARGUMENT_RESOLVERS_BEAN_NAME,
+					internalArgumentResolversBuilder(registry, true).getBeanDefinition());
+		}
+	}
+
+	private BeanDefinitionBuilder internalArgumentResolversBuilder(BeanDefinitionRegistry registry,
+			boolean listCapable) {
+		ManagedList<BeanReference> resolvers = new ManagedList<>();
+		if (!registry.containsBeanDefinition(IntegrationContextUtils.PAYLOAD_RESOLVER_BEAN_NAME)) {
```
artembilan: Why do we need to have the all as beans?

I might agree that we have to register two sets of resolvers - single entry and listCapable, but I don't see reason to register each `ArgumentResolver` as a bean.

Seems for me just a simple `BeanDefinition` as reference to the `ManagedList` does the trick and those "ephemeral" beans are populated with the `BeanFactory` if that...

Thanks

artembilan (changes requested):
Nothing critical - just several concerns to explain and maybe discuss.

Now I remember why I haven't exposed that `MessageHandlerMethodFactory` as a bean. Just because we have that `canProcessMessageList`. So, just a single generic bean is not enough. Plus yes, Gary, is right - the `MessagingMethodInvokerHelper` in most cases is used from the ctors, a bit earlier than we are aware about `BeanFactory`. So, we are able to fail fast.

Anyway, the compromise provided here is good enough facilitate the SCSt requirements and we always can override these holder beans for the SI with an appropriate set of resolvers.

/CC @olegz

spring-integration-core/src/main/java/org/springframework/integration/config/HandlerMethodArgumentResolversHolder.java
garyrussell: Ironically, that class is not spring-friendly (XML); it can't be configured as a `<bean/>`.

artembilan: Ha! I see 😄 
OK. Let's stay with our version!
Plus we will be sure that there won't be any conflicts since `HandlerMethodArgumentResolverComposite` is a `HandlerMethodArgumentResolver` as well.

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -459,52 +458,11 @@ private void registerListCapableArgumentResolvers(BeanDefinitionRegistry registr
...
-		if (!registry.containsBeanDefinition(IntegrationContextUtils.MAP_RESOLVER_BEAN_NAME)) {
-			BeanDefinitionBuilder resolverBuilder = BeanDefinitionBuilder.genericBeanDefinition(
-					MapArgumentResolver.class);
-			registry.registerBeanDefinition(IntegrationContextUtils.MAP_RESOLVER_BEAN_NAME,
-					resolverBuilder.getBeanDefinition());
-		}
-		resolvers.add(new RuntimeBeanReference(IntegrationContextUtils.MAP_RESOLVER_BEAN_NAME));
-
+		ManagedList<HandlerMethodArgumentResolver> resolvers = new ManagedList<>();
+		resolvers.add(new PayloadExpressionArgumentResolver());
```
artembilan: Well, unfortunately this doesn't trigger the `BeanFactory` callbacks. Must be like:
```
resolvers.add(new RootBeanDefinition(PayloadExpressionArgumentResolver.class));
```
Will fix on merge if only concern.

garyrussell: Thanks.

garyrussell: Interesting that all out tests pass, though - I guess we don't have tests that actually need those callbacks 😦

artembilan: I think so. I'll try to come up with something simple during merge. For example something like `#jsonPath()` from the `@Payload(expression)`

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -513,43 +515,74 @@ private void initializeHandler(HandlerMethod candidate) {
...
+
+		List<HandlerMethodArgumentResolver> customArgumentResolvers = new LinkedList<>();
+		customArgumentResolvers.add(payloadExpressionArgumentResolver);
+		customArgumentResolvers.add(payloadsArgumentResolver);
+		customArgumentResolvers.add(collectionArgumentResolver);
+		customArgumentResolvers.add(mapArgumentResolver);
+
+		this.messageHandlerMethodFactory.setCustomArgumentResolvers(customArgumentResolvers);
+
+		if (getBeanFactory() != null &&
```
artembilan: Redundant code. If we are in this `configureLocalMessageHandlerFactory()`, then no `beanFactory` or `ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME`. Therefore this code is dead.
Will fix on merge

garyrussell: Not really - this is called from 2 places - bf exists but doesn't have the arg resolver beans.

artembilan: Yeah... Makes sense. Thank you!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/986d5fc0cdf663d9007a27643b948e7132ef0d7d.

With the subsequent polishing: https://github.com/spring-projects/spring-integration/commit/9414ceef869b2bcf8ba9aabdb1c0a0204e2c220f

PR #2287, opened 2017-11-17
INT-4365: Improve notPropagatedHeaders function

artembilan:
JIRA: https://jira.spring.io/browse/INT-4365

It is much useful to configure the `notPropagatedHeaders` as a set of
patterns to match.
In this case we can filter a group of headers with the common prefix or
suffix

* Allow to configure `AbstractMessageProducingHandler.setNotPropagatedHeaders`
as simple patterns; the `*` means filter all - not copy request headers
at all - similar to `transformer` behavior
* Add `ConsumerEndpointSpec.notPropagatedHeaders()` for Java DSL
* Add `not-propagated-headers` to the `<service-activator>`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
@@ -43,6 +43,7 @@ import org.springframework.messaging.core.DestinationResolutionException;
 import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
+import org.springframework.util.PatternMatchUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.util.concurrent.ListenableFuture;
 import org.springframework.util.concurrent.ListenableFutureCallback;
@@ -64,17 +65,19 @@ import reactor.core.publisher.Mono;
 public abstract class AbstractMessageProducingHandler extends AbstractMessageHandler
 		implements MessageProducer, HeaderPropagationAware {
 
-	private final Set<String> notPropagatedHeaders = new HashSet<String>();
+	private final Set<String> notPropagatedHeaders = new HashSet<>();
 
 	protected final MessagingTemplate messagingTemplate = new MessagingTemplate();
 
-	private volatile MessageChannel outputChannel;
-
-	private volatile String outputChannelName;
-
-	private volatile boolean async;
+	private boolean async;
 
 	private boolean selectiveHeaderPropagation;
+
+	private boolean shouldCopyRequestHeaders = true;
+
+	private String outputChannelName;
+
+	private volatile MessageChannel outputChannel;
 
 	/**
 	 * Set the timeout for sending reply Messages.
@@ -115,10 +118,13 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	}
 
 	/**
-	 * Set headers that will NOT be copied from the inbound message if
+	 * Set header patterns ("xxx*", "*xxx", "*xxx*" or "xxx*yyy")
+	 * that will NOT be copied from the inbound message if
 	 * {@link #shouldCopyRequestHeaders() shouldCopyRequestHeaaders} is true.
+	 * '*' means do not copy headers at all.
 	 * @param headers the headers to not propagate from the inbound message.
 	 * @since 4.3.10
+	 * @see org.springframework.util.PatternMatchUtils
 	 */
 	@Override
 	public void setNotPropagatedHeaders(String... headers) {
@@ -134,14 +140,17 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 			this.notPropagatedHeaders.addAll(Arrays.asList(headers));
 		}
 		this.selectiveHeaderPropagation = this.notPropagatedHeaders.size() > 0;
+
+		this.shouldCopyRequestHeaders = !this.notPropagatedHeaders.contains("*");
 	}
 
 	/**
-	 * Get the header names this handler doesn't propagate.
+	 * Get the header patterns this handler doesn't propagate.
 	 * @return an immutable {@link java.util.Collection} of headers that will not be
 	 * copied from the inbound message if {@link #shouldCopyRequestHeaders()} is true.
 	 * @since 4.3.10
 	 * @see #setNotPropagatedHeaders(String...)
+	 * @see org.springframework.util.PatternMatchUtils
 	 */
 	@Override
 	public Collection<String> getNotPropagatedHeaders() {
@@ -149,7 +158,8 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	}
 
 	/**
-	 * Add headers that will NOT be copied from the inbound message if
+	 * Add header patterns ("xxx*", "*xxx", "*xxx*" or "xxx*yyy")
+	 * that will NOT be copied from the inbound message if
 	 * {@link #shouldCopyRequestHeaders()} is true, instead of overwriting the existing
 	 * set.
 	 * @param headers the headers to not propagate from the inbound message.
@@ -344,7 +354,7 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	protected Message<?> createOutputMessage(Object output, MessageHeaders requestHeaders) {
 		AbstractIntegrationMessageBuilder<?> builder = null;
 		if (output instanceof Message<?>) {
-			if (!this.shouldCopyRequestHeaders()) {
+			if (!shouldCopyRequestHeaders()) {
 				return (Message<?>) output;
 			}
 			builder = this.getMessageBuilderFactory().fromMessage((Message<?>) output);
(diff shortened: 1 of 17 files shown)


src/reference/asciidoc/message.adoc
```diff
@@ -394,9 +394,16 @@ Also, a header is only propagated if it does not already exist in the outbound m
 
 Starting with _version 4.3.10_, you can configure message handlers (that modify messages and produce output) to suppress the propagation of specific headers.
 Call the `setNotPropagatedHeaders()` or `addNotPropagatedHeaders()` methods on the `MessageProducingMessageHandler` abstract class, to configure the header(s) you don't want to be copied.
+
 You can also globally suppress propagation of specific message headers by setting the `readOnlyHeaders` property in `META-INF/spring.integration.properties` to a comma-delimited list of headers.
 
-IMPORTANT: Header propagation suppression does not apply to those endpoints that don't modify the message, e.g. <<bridge, bridges>> and <<router, routers>>
+Starting with _version 5.0_, the `setNotPropagatedHeaders()` implementation on the `AbstractMessageProducingHandler` applies the simple patterns (`xxx*`, `*xxx`, `*xxx*` or `xxx*yyy`) to allow to filter header with common suffix or prefix.
+See `PatternMatchUtils` JavaDocs for more information.
+When `*` (asterisk) is used, no any headers are propagated.
```
andybaer: When `*` (asterisk) is used, no headers are propagated.

m-terra (commented):
Does this PR included requires-reply xml support for Transformer?
These changes will give us the flexibility and control we need in our complex flows. 
Thanks!

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -134,22 +140,26 @@ private void updateNotPropagatedHeaders(String[] headers, boolean merge) {
 			this.notPropagatedHeaders.addAll(Arrays.asList(headers));
 		}
 		this.selectiveHeaderPropagation = this.notPropagatedHeaders.size() > 0;
+
+		this.shouldCopyRequestHeaders = !this.notPropagatedHeaders.contains("*");
```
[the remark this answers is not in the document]
artembilan: That's not. Because `notPropagatedHeaders` is a `Set` not simple `String`. So, we don't copy headers at all if one of the `notPropagatedHeaders` elements is exactly asterisk, e.g. `foo, b*, *, b*z`.
So, all other patterns will be ignored just because we have one of them as `*`. It has a highest priority.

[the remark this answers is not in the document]
andybaer: Of course, sry!

artembilan:
Thank you, @andybaer , for the valuable fieedback!

Well, I did requested changes - please take a look again.

Right, I was in doubt last week to expose `requiresReply` for the `<transformer>` or not. 
That is not the component to return a null as simple result of the transformation function.
Although that's for now... I still hope that we will introduce once `null payload` support in the future: https://jira.spring.io/browse/INT-997

But after some weekend debates in my head I decided to relax a restriction and expose it as is.
In the end that doesn't hurt the Framework and everything rest is up to the end-user application. We provide an option and you decide to use it or not and how.

andybaer:
Thanks Artem!
For our application, it would be either 
- using transformer everywhere and mark possible 'flow aborters' with requires-reply=false
or
- using service activators everywhere and mark possible incorrect header merges with not-progated-headers=*
For readability I would prefer the first one but most important is having an option to deal with the scenarios.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -133,23 +139,34 @@ private void updateNotPropagatedHeaders(String[] headers, boolean merge) {
 			}
 			this.notPropagatedHeaders.addAll(Arrays.asList(headers));
 		}
+		boolean hasAsterisk = this.notPropagatedHeaders.contains("*");
+
+		if (hasAsterisk) {
+			this.notPropagatedHeaders.clear();
+			this.notPropagatedHeaders.add("*");
+		}
```
garyrussell: I would just set a boolean in this case so we can optimize later.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -355,12 +372,15 @@ else if (output instanceof AbstractIntegrationMessageBuilder) {
 		else {
 			builder = this.getMessageBuilderFactory().withPayload(output);
 		}
-		if (this.shouldCopyRequestHeaders()) {
+		if (shouldCopyRequestHeaders()) {
```
garyrussell: If (shouldCopyRequestHeaders() && !this.copyNoHeaders) {...

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
garyrussell: I would just set a boolean in this case so we can optimize later e.g `this.copyNoHeaders=true`.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -355,12 +372,15 @@ else if (output instanceof AbstractIntegrationMessageBuilder) {
...
 			if (this.selectiveHeaderPropagation) {
-				Map<String, Object> headersToCopy = new HashMap<String, Object>(requestHeaders);
-				for (String header : this.notPropagatedHeaders) {
-					headersToCopy.remove(header);
-				}
+				String[] patterns = this.notPropagatedHeaders.toArray(new String[this.notPropagatedHeaders.size()]);
```
garyrussell: We can optimize here rather than building an array for every message.

src/reference/asciidoc/message.adoc
```diff
@@ -394,9 +394,16 @@ Also, a header is only propagated if it does not already exist in the outbound m
 
 Starting with _version 4.3.10_, you can configure message handlers (that modify messages and produce output) to suppress the propagation of specific headers.
 Call the `setNotPropagatedHeaders()` or `addNotPropagatedHeaders()` methods on the `MessageProducingMessageHandler` abstract class, to configure the header(s) you don't want to be copied.
+
 You can also globally suppress propagation of specific message headers by setting the `readOnlyHeaders` property in `META-INF/spring.integration.properties` to a comma-delimited list of headers.
 
-IMPORTANT: Header propagation suppression does not apply to those endpoints that don't modify the message, e.g. <<bridge, bridges>> and <<router, routers>>
+Starting with _version 5.0_, the `setNotPropagatedHeaders()` implementation on the `AbstractMessageProducingHandler` applies the simple patterns (`xxx*`, `*xxx`, `*xxx*` or `xxx*yyy`) to allow to filter header with common suffix or prefix.
```
garyrussell: `s/to allow to filter with common/to allow filtering headers with a common/`

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
artembilan: I do that. See below:
```
this.shouldCopyRequestHeaders = !hasAsterisk;
```
And `shouldCopyRequestHeaders()` body.
What am I missing?

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
artembilan: I don't see reason in that `copyNoHeaders` `shouldCopyRequestHeaders` does the trick.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
garyrussell: Ah - didn't see that but, since subclasses can (and do) override that method, we need another boolean.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
artembilan: Remember:
> if * {@link #shouldCopyRequestHeaders() shouldCopyRequestHeaaders} is true. 

?

So, independently of what you configure for the `headers`, but if `shouldCopyRequestHeaders()` returns `false`, they all are ignored.

garyrussell: But what if a subclass overrides `shouldCopyRequestHeaders()` to return `true` and `*` is the only pattern?

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
artembilan: Right, and they always override it to `false` eliminating all the purpose of this `notPropagatedHeaders`

garyrussell: While ours might override to `false`, we can't guarantee user handlers do the same.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
artembilan: Ha! That was your suggestion to go pattern way 😄 
I was just suggested simple boolean flag - for symmetry of existing `shouldCopyRequestHeaders()`.

I don't know (yet) how to be...

Need to think, but I somehow don't like too much boolean flags...

garyrussell: Perhaps we could call `shouldCopyRequestHeaders()` just once instead of on every message. It doesn't really make much sense to call it on every message anyway (unless a subclass used a `ThreadLocal`).

artembilan:
OK. Pushed some fix.
I think ready for one more review round.
Thanks

garyrussell:
Merged as 90c46f5a79894a130badad77d2ee7ede9b33f697

PR #2290, opened 2017-11-21
Don't wrap SpEL result message to another message

artembilan:
The `ExpressionEvaluatingTransactionSynchronizationProcessor` wraps
an expression evaluation result to the `Message` unconditionally

* Don't wrap one message to another if SpEL result is a `Message` per se
* Don't wrap the received message to another if the `expression` isn't
configured.

**Cherry-pick to 4.3.x**


garyrussell:
I don't think we can cherry-pick - it's a behavior change.

We should probably update the migration guide too.

artembilan:
Done: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-4.3-to-5.0-Migration-Guide#expressionevaluatingtransactionsynchronizationprocessor-and-message-to-send.

No problems with no cherry-picking!

PR #2291, opened 2017-11-21
INT-2576: Optimize JdbcMS.removeMessageGroup

artembilan:
JIRA: https://jira.spring.io/browse/INT-2576

* Do not fetch message ids and then iterate over them to remove
one by one - just use an appropriate `DELETE` query for all messages

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
@@ -19,7 +19,6 @@ package org.springframework.integration.jdbc.store;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
@@ -48,7 +47,6 @@ import org.springframework.integration.support.converter.WhiteListDeserializingC
 import org.springframework.integration.util.UUIDConverter;
 import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.jdbc.core.JdbcTemplate;
-import org.springframework.jdbc.core.RowCallbackHandler;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.SingleColumnRowMapper;
 import org.springframework.jdbc.support.lob.DefaultLobHandler;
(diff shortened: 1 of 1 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
```diff
@@ -105,12 +103,9 @@
...
 		COUNT_ALL_MESSAGES_IN_GROUP("SELECT COUNT(MESSAGE_ID) from %PREFIX%GROUP_TO_MESSAGE where GROUP_KEY=? and REGION=?"),
 
-		LIST_MESSAGEIDS_BY_GROUP_KEY("select MESSAGE_ID, CREATED_DATE " +
-				"from %PREFIX%MESSAGE where MESSAGE_ID in (select MESSAGE_ID from %PREFIX%GROUP_TO_MESSAGE where GROUP_KEY=? and REGION=?) " +
-				"ORDER BY CREATED_DATE"),
-
 		LIST_MESSAGES_BY_GROUP_KEY("SELECT MESSAGE_ID, MESSAGE_BYTES, CREATED_DATE " +
-				"from %PREFIX%MESSAGE where MESSAGE_ID in (SELECT MESSAGE_ID from %PREFIX%GROUP_TO_MESSAGE where GROUP_KEY = ?) and REGION=? " +
+				"from %PREFIX%MESSAGE where MESSAGE_ID in " +
+				"(SELECT MESSAGE_ID from %PREFIX%GROUP_TO_MESSAGE where GROUP_KEY = ? and REGION = ?) and REGION = ? " +
```
garyrussell: Why do we need REGION twice? Sure the `in` clause will only have ids in that region so the and part will be a no-op.

garyrussell (changes requested):
That's all; otherwise LGTM.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
artembilan: That's not true. Primary key is:
```
constraint MESSAGE_PK primary key (MESSAGE_ID, REGION)
```
So, definitely the same message may be in different regions and, therefore, selecting just ids for messages is not enough.

Let me know what ma I missing.

garyrussell (approved):

PR #2292, opened 2017-11-22
Revert to Spring Social Twitter 1.1.2.RELEASE

habuma:
Reverts Spring Integration to build against Spring Social Twitter 1.1.2.RELEASE, as Spring Social Twitter 2.0.0 may not be ready in time for Spring Integration 5.0.0


artembilan (approved):
Thank, @habuma !

So, it is green and since there is no other SST version, that is really the way to go for our GA - with your latest GA! Right, we are not compatible with SST-2.0, but since it isn't GA we haven't promised anything otherwise. 😄 

@garyrussell , any objections do not go this way?
As I mentioned in the email discussion we can consider to move it to `extensions` in `5.1`

PR #2297, opened 2017-11-28
INT-4365 Fixed NPE generated by previous commit

olegz:


olegz:
FWIW, it seems that in could be cleaner if internally `notPropagatedHeaders` were `List<String>` instead of the array. . . something to consider, but for now I just initialized to an empty array.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -73,7 +73,7 @@
 
 	private MessageChannel outputChannel;
 
-	private String[] notPropagatedHeaders;
+	private String[] notPropagatedHeaders = new String[] {};
```
artembilan: We can't make it as a `List` (it was a `Set`), because the target runtime logic is really based on the array.
See `createOutputMessage()`:
```
headersToCopy.entrySet()
						.removeIf(entry -> PatternMatchUtils.simpleMatch(this.notPropagatedHeaders, entry.getKey()));
```

On another hand I wouldn't create an empty object here. In most cases we definitely don't have any headers restrictions. So, the `null` is more optimal here.

What we can do is just this in the `updateNotPropagatedHeaders()`:
```
if (merge && this.notPropagatedHeaders != null) {
		headerPatterns.addAll(Arrays.asList(this.notPropagatedHeaders));
}
```

spring-integration-core/src/test/java/org/springframework/integration/handler/AbstractReplyProducingMessageHandlerTests.java
```diff
@@ -109,6 +108,33 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
...
+		handler.handleMessage(MessageBuilder.withPayload("hello")
+				.setHeader("boom", "FOO")
+				.setHeader("bar", "BAR")
+				.setHeader("baz", "BAZ")
+				.build());
+		Message<?> out = captor.getValue();
+		assertThat(out, notNullValue());
+		assertThat(out.getHeaders().get("boom"), nullValue());
+		assertThat(out.getHeaders().get("bar"), equalTo("RAB"));
+		assertThat(out.getHeaders().get("baz"), equalTo("BAZ"));
```
artembilan: There is a `HeaderMatcher` in the `org.springframework.integration.test.matcher`.
Just FYI, for the future.
We can live with your current test.

artembilan (changes requested):
A couple comments.
I can address on merge if you are OK.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
olegz: I am ok either way (as I don;'t have the full context), so your choice

artembilan (approved):
Good.

So, merging with the mentioned polishing.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b58087881ac7e55794fa39611f2fb92450034f10

Issue #2300, opened 2017-12-07
WebFlux.outboundGateway has no way of producing Flux<Anything>


asarkar:
In `WebFluxRequestExecutingMessageHandler`, I see the following:
```
if (expectedResponseType instanceof ParameterizedTypeReference<?>) {
  bodyMono = response.body(BodyExtractors.toMono((ParameterizedTypeReference<?>) expectedResponseType));
}
else if (expectedResponseType != null) {
	bodyMono = response.body(BodyExtractors.toMono((Class<?>) expectedResponseType));
}
```

What I'm expecting as response is `Flux<DataBuffer>` that I can do some processing on before sending to an output channel. However, the above code always forces the body to be converted to `Mono`.

The documentation for webflux support merely regurgitates what's in the code. 

> The WebClient exchange() operation returns a Mono<ClientResponse> which is mapped to the AbstractIntegrationMessageBuilder reactive support (using Mono.map()) as the output from the WebFluxRequestExecutingMessageHandler. Together with the ReactiveChannel as an outputChannel, the Mono<ClientResponse> evaluation is deferred until a downstream subscription is made. Otherwise, it is treated as an async mode and the Mono response is adapted to an SettableListenableFuture for an asynchronous reply from the WebFluxRequestExecutingMessageHandler.

It'd be nice to also see some examples for how to get a `Flux<UserDataType>`,  `Flux<DataBuffer>`, `Mono<ClientResponse>` etc. As with `RestTemplate`, where I've the option to get the `ResponseEntity` if I want to do the processing myself, I should also be able to get `Mono<ClientResponse>` should I so choose.

artembilan:
Same here. The PR on the matter #2304

artembilan:
Correct. That is not Spring Integration and Messaging architecture responsibility to expose such a low level resource. If you really need the access to it and do low-level logic around, consider to use `WebClient` directly in some POJO for the `@ServiceActivator`.

Anyway in this case we provide the body converted to the `payload` via provided `HttpMessageReader`s and HTTP headers are copied into the `MessageHeaders` according `HeaderMapper`. So, technically you have everything in the `Message` what you could get from the `ClientResponse` directly.

asarkar:
>  you have everything in the Message what you could get from the ClientResponse directly.

[...]

PR #2304, opened 2017-12-14
GH-2300: Add Flux support in WebFluxRequestExecMH

artembilan:
Fixes: spring-projects/spring-integration#2300

To allow to consume a streaming HTTP response downstream expose
`replyToFlux` option on the `WebFluxRequestExecutingMessageHandler`.
This way the body of the HTTP response can be converted now to the
`Flux` for subsequent output message.
The option is `false` by default; can be changed to `true` in the `5.1`

### DIFF
--- a/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
+++ b/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
@@ -29,6 +29,7 @@ import org.springframework.expression.common.LiteralExpression;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.ReactiveHttpInputMessage;
 import org.springframework.http.ResponseEntity;
 import org.springframework.integration.expression.ValueExpression;
 import org.springframework.integration.http.outbound.AbstractHttpRequestExecutingMessageHandler;
@@ -36,12 +37,14 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.util.Assert;
 import org.springframework.util.MimeType;
+import org.springframework.web.reactive.function.BodyExtractor;
 import org.springframework.web.reactive.function.BodyExtractors;
 import org.springframework.web.reactive.function.BodyInserters;
 import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.WebClient;
 import org.springframework.web.reactive.function.client.WebClientResponseException;
 
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 /**
@@ -58,6 +61,8 @@ import reactor.core.publisher.Mono;
 public class WebFluxRequestExecutingMessageHandler extends AbstractHttpRequestExecutingMessageHandler {
 
 	private final WebClient webClient;
+
+	private boolean replyToFlux;
 
 	/**
 	 * Create a handler that will send requests to the provided URI.
@@ -108,6 +113,20 @@ public class WebFluxRequestExecutingMessageHandler extends AbstractHttpRequestEx
 		super(uriExpression);
 		this.webClient = (webClient == null ? WebClient.create() : webClient);
 		this.setAsync(true);
+	}
+
+	/**
+	 * The boolean flag t identify if the reply payload should be as a {@link Flux} from the response body
+	 * or as resolved value from the {@link Mono} of the response body.
+	 * Defaults to {@code false} - simple value is pushed downstream.
+	 * Makes sense when {@code expectedResponseType} is configured.
+	 * @param replyToFlux represent reply payload as a {@link Flux} or as a value from the {@link Mono}.
+	 * @since 5.0.1
+	 * @see #setExpectedResponseType(Class)
+	 * @see #setExpectedResponseTypeExpression(Expression)
+	 */
+	public void setReplyToFlux(boolean replyToFlux) {
+		this.replyToFlux = replyToFlux;
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
```diff
@@ -110,6 +115,20 @@ public WebFluxRequestExecutingMessageHandler(Expression uriExpression, WebClient
 		this.setAsync(true);
 	}
 
+	/**
+	 * The boolean flag t identify if the reply payload should be as a {@link Flux} from the response body
+	 * or as resolved value from the {@link Mono} of the response body.
```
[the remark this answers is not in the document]
garyrussell: I was going to fix that on merge, but the discussion on the issue would indicate more changes are coming...

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
```diff
@@ -169,13 +188,35 @@ protected Object exchange(Supplier<URI> uriSupplier, HttpMethod httpMethod, Http
...
 												.headers(response.headers().asHttpHeaders());
 
-								Mono<?> bodyMono = Mono.empty();
-
-								if (expectedResponseType instanceof ParameterizedTypeReference<?>) {
-									bodyMono = response.body(BodyExtractors.toMono((ParameterizedTypeReference<?>) expectedResponseType));
+								Mono<?> bodyMono;
+
+								if (expectedResponseType != null) {
+									if (this.replyToFlux) {
```
asarkar: Perhaps the property `replyToFlux` should be named `replyAsFlux` or better `responseAsFlux` that's more in line with `responseType`.

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
```diff
@@ -169,13 +188,35 @@ protected Object exchange(Supplier<URI> uriSupplier, HttpMethod httpMethod, Http
...
+										else {
+											extractor = BodyExtractors.toFlux((Class<?>) expectedResponseType);
+										}
+										Flux<?> flux = response.body(extractor);
+										bodyMono = Mono.just(flux);
+									}
+									else {
+										BodyExtractor<? extends Mono<?>, ReactiveHttpInputMessage> extractor;
+										if (expectedResponseType instanceof ParameterizedTypeReference<?>) {
+											extractor = BodyExtractors.toMono(
```
asarkar: Here, a additional check can be made to see if the wrapped type is `ClientResponse`, and if so, skip the `BodyExtractor`. I don't exactly remember how to get the type out of `ParameterizedTypeReference`, but I think Spring has some utilities for that.
Having the client provide a no-op `BodyExtractor` is not very convenient because they also have to be mindful of the order of the various `BodyExtractor`; and technically, we would not be extracting the body at all, so, a no-op `BodyExtractor` would be misleading.

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
```diff
@@ -169,13 +188,35 @@ protected Object exchange(Supplier<URI> uriSupplier, HttpMethod httpMethod, Http
 										ResponseEntity.status(response.statusCode())
 												.headers(response.headers().asHttpHeaders());
 
-								Mono<?> bodyMono = Mono.empty();
-
-								if (expectedResponseType instanceof ParameterizedTypeReference<?>) {
-									bodyMono = response.body(BodyExtractors.toMono((ParameterizedTypeReference<?>) expectedResponseType));
+								Mono<?> bodyMono;
+
+								if (expectedResponseType != null) {
```
asarkar: Is a check made somewhere to see if the response is 2xx? Should we be trying to parse the response if it's not?

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
asarkar: ON 2nd thought, this property seems redundant. If you set `responseType=ParameterizedTypeReference<Flux<T>>`, what'd you expect to get back?

artembilan: No, that's not. I accept neither options. I try to be aligned here with `WebClient` API.

So, we have there these in the `ClientResponse`:
```
<T> T body(BodyExtractor<T, ? super ClientHttpResponse> extractor);

<T> Mono<T> bodyToMono(Class<? extends T> elementClass);

<T> Mono<T> bodyToMono(ParameterizedTypeReference<T> typeReference);

<T> Flux<T> bodyToFlux(Class<? extends T> elementClass);

<T> Flux<T> bodyToFlux(ParameterizedTypeReference<T> typeReference);
```

So, `expectedResponseType` is indeed just a type for the response element, meanwhile `toMono/toFlux` is an API level, so that would be a bit confusing for end-user to say that to select `flux/mono` API, please, specify the particular type. And I named it `toFlux` definitely after that `bodyToFlux` API. Again - for consistency.

This is indeed not about `response`. That is what we will have as a `payload` of reply message - the result of `bodyToMono()` or `bodyToFlux()` methods invocation underneath.

I am also against checking the type for the `ClientResponse`. As I mentioned we have that ` body(BodyExtractor<T, ? super ClientHttpResponse> extractor);` API on the `WebClient` level, so that should be very clear and consistent if we provide a hook to inject any customization  to the `ClientResponse` via `BodyExtractor`. Right, we might call it like `IdentityBodyExtractor` instead, without any low-level operation but just return for the `ClientHttpResponse`. And that is how we will skip any conversion underneath.

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
artembilan: See my comment above.

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
artembilan: Yes, we have that one there:
```
if (httpStatus.isError()) {
	return response.body(BodyExtractors.toDataBuffers())
			.reduce(DataBuffer::write)
                         ...
			.map(bodyBytes -> {
					throw new WebClientResponseException(...);
					}
			);
}
else {
	return Mono.just(response);
}
```
See `WebFluxRequestExecutingMessageHandler.exchange()` source code.

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
artembilan: Right, but since Spring Integration WebFlux is fully based on the Srping WebFlux that really would be better to stay consistence with their API. So, we are going to use existing `BodyExtractor` strategy as an option to inject into this gateway. But at the same time since there is no such a "void" implementation in the Spring WebFlux for us, we are going to place it exactly in the `org.springframework.integration.webflux.support`.

Right, we will also talk to Spring team about having that impl there, in their package, but I'm afraid they just answer to us with an argument that they have already `requestSpec.exchange()` which returns exactly `Mono<ClientResponse>` for your consideration. 

Since we don't have such a luxury here in SI to cover all the `WebClient` API, I think like middle ground with the `BodyExtractor` injection and some `IdentityBodyExtractor` as a point to get access to the ClientResponse` as is, should be enough to cover as much as possible from your original request.

Sounds good ?

asarkar: Sounds good to me. Thanks for making the PR.

garyrussell:
Merged as bd872846fb470ebc32d0d664dac82d1daa7599bf with minor doc polishing.

artembilan:
> Merged as ...

I'm afraid that we also talked about `BodyExtractor` option as well, but I guess that can be addressed in the separate PR. NP.

asarkar:
@artembilan I was trying to use this new feature, but `WebFluxMessageHandlerSpec` doesn't expose it. The only way is to create a `WebFluxRequestExecutingMessageHandler` myself, which defeats the purpose of the spec.

artembilan:
@asarkar,

Thank you for the pointer!
Indeed that has been missed here, but fixed now in the #2307

PR #2307, opened 2017-12-20
WebFlux improvements

artembilan:
* add `BodyExtractor` support for Outbound part
* add `ClientHttpResponseBodyExtractor` as identity function
* add XML configuration for the Inbound part
* document `BodyExtractor` and Inbound XML support

### DIFF
--- a/src/reference/asciidoc/webflux.adoc
+++ b/src/reference/asciidoc/webflux.adoc
@@ -100,7 +100,7 @@ Together with the `ReactiveChannel` as an `outputChannel`, the `Mono<ClientRespo
 Otherwise, it is treated as an `async` mode and the `Mono` response is adapted to an `SettableListenableFuture` for an asynchronous reply from the `WebFluxRequestExecutingMessageHandler`.
 The target payload of the output message depends on the `WebFluxRequestExecutingMessageHandler` configuration.
 The `setExpectedResponseType(Class<?>)` or `setExpectedResponseTypeExpression(Expression)` identifies the target type of the response body element conversion.
-If the `replyToFlux` is set to `true`, the response body is converted to a `Flux` with the provided `expectedResponseType` for each element and this `Flux` is sent as the payload downstream.
+If the `replyPayloadToFlux` is set to `true`, the response body is converted to a `Flux` with the provided `expectedResponseType` for each element and this `Flux` is sent as the payload downstream.
 A <<splitter,splitter>> afterwards can be used to iterate over this `Flux` in a reactive manner.
 
 In addition a `BodyExtractor<?, ClientHttpResponse>` can be injected into the `WebFluxRequestExecutingMessageHandler` instead of `expectedResponseType` and `replyToFlux` properties.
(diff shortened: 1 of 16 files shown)


asarkar:
@artembilan Since this PR is still open, I wanna make a follow up comment for your consideration. We had talked about whether `replyToFlux` is the best naming choice, and after some thought, I think it is misleading. Yes, in core web, `bodyToXXX` makes sense, but in a messaging world, `replyTo` has a very specific meaning (replying to another component, usually sender), and I feel `replyToFlux` is misleading in that sense. I still suggest `responseAsFlux`.
What do you think?

artembilan:
OK. Since it is already confusing you and you are definitely a Community member, we will respect your feeling and we'll come to the rescue.

Usually such a confusing names we try to explain and clarify via JavaDocs:

```
/**
 * The boolean flag to identify if the reply payload should be as a {@link Flux} from the response body
 * or as resolved value from the {@link Mono} of the response body.
 * Defaults to {@code false} - simple value is pushed downstream.
 * Makes sense when {@code expectedResponseType} is configured.
 * @param replyToFlux represent reply payload as a {@link Flux} or as a value from the {@link Mono}.
 * @since 5.0.1
 * @see #setExpectedResponseType(Class)
 * @see #setExpectedResponseTypeExpression(Expression)
 */
public void setReplyToFlux(boolean replyToFlux) {
	this.replyToFlux = replyToFlux;
}
```
but yeah, if the naming is goal and only a source of the info, let's reconsider!

I still insist that `response` in terms of HTTP-based component is still to broad and more confusing since it is strictly associated with the whole HTTP Response object. Therefore I would stay away from that.

My suggestion is like `replyAsFlux` or maybe a bit broader but cleaner - `replyPayloadAsFlux`.

WDYT?

asarkar:
Thanks for understanding. `replyAsFlux` or `replyBodyAsFlux` or `replyPayloadAsFlux` are all reasonable - pick whatever sounds good to you. I don’t put much weight on documentation tryin to explain names, code should speak for itself as much as possible.

artembilan:
I ended up with the `replyPayloadToFlux`. See the latest commit.

asarkar:
@artembilan But you said `replyPayloadAsFlux`, not to. You changed your mind. That’s ok though.

artembilan:
Yeah... I do sometimes during development 😄 

The `as` is associated in my mind with the `cast` operation: http://groovy-lang.org/operators.html#_coercion_operator. Meanwhile `to` is more like conversion or wrapping.

Agree: the naming, especially short, self-explanatory, is hard. But together we definitely may work out the final decision. Not too late to rename though!

asarkar:
Thanks for accommodating my request. Looking forward to this PR merged

src/reference/asciidoc/webflux.adoc
```diff
@@ -100,9 +100,13 @@ Together with the `ReactiveChannel` as an `outputChannel`, the `Mono<ClientRespo
 Otherwise, it is treated as an `async` mode and the `Mono` response is adapted to an `SettableListenableFuture` for an asynchronous reply from the `WebFluxRequestExecutingMessageHandler`.
 The target payload of the output message depends on the `WebFluxRequestExecutingMessageHandler` configuration.
 The `setExpectedResponseType(Class<?>)` or `setExpectedResponseTypeExpression(Expression)` identifies the target type of the response body element conversion.
-If `replyToFlux` is set to `true`, the response body is converted to a `Flux` with the provided `expectedResponseType` for each element and this `Flux` is sent as the payload downstream.
+If the `replyPayloadToFlux` is set to `true`, the response body is converted to a `Flux` with the provided `expectedResponseType` for each element and this `Flux` is sent as the payload downstream.
 A <<splitter,splitter>> afterwards can be used to iterate over this `Flux` in a reactive manner.
 
+In addition a `BodyExtractor<?, ClientHttpResponse>` can be injected into the `WebFluxRequestExecutingMessageHandler` instead of `expectedResponseType` and `replyToFlux` properties.
```
garyrussell: `s/replyToFlux/replyPayloadToFlux`

garyrussell (approved):
That's all; will fix on merge.

garyrussell:
Merged as b5ec98f025722554f820829a1a0914fdf546a678

PR #2310, opened 2017-12-26
INT-4371: Add DeferredAcknowledgmentMessageSource

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4371

Add an abstract message source that populates the message with a header containing
an abstraction that allows the application to acknowledge/reject/requeue the message.
(In amqp, for example, this would map to channel.basicAck(tag),
channel.basicReject(tag, false), and channel.basicReject(tag, true).

With some brokers (such as kafka), acknowledge and reject might be equivalent
since there is no automatic DLQ processing in kafka both would simply commit the offset.

Also include a strategy interface for a factory to provide the acknowledgment header.

__TODO before merge: docs, <strike>kafka impl</strike>__

(I have now done the Kafka implementation and didn't need any changes to the SI core code).

I think we can wait until/if someone asks for XML config.


spring-integration-core/src/main/java/org/springframework/integration/support/AcknowledgmentCallbackFactory.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+public interface AcknowledgmentCallbackFactory<T> {
+
+	/**
+	 * Create the callback.
+	 * @param supplier a supplier of T
+	 * @return the callback
+	 */
+	AcknowledgmentCallback createCallback(Supplier<T> supplier);
```
artembilan: I wonder why would one need to customize such a low level resource...
Just curious what is the point of this abstraction.
And why have you made it based on the `Supplier`... Why simple `T` is not enough?..

Thanks

spring-integration-core/src/main/java/org/springframework/integration/endpoint/SourcePollingChannelAdapter.java
```diff
@@ -214,8 +218,22 @@ protected void handleMessage(Message<?> message) {
 		}
 		try {
 			this.messagingTemplate.send(getOutputChannel(), message);
+			if (this.source instanceof DeferredAcknowlegmentMessageSource) {
```
artembilan: Why do we need this new abstraction when there really would be enough to get an `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` header from the message and perform the logic if it isn't `null`?

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractDeferredAcknowlegmentMessageSource.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Gary Russell
+ * @since 5.0.1
+ *
+ */
+public abstract class AbstractDeferredAcknowlegmentMessageSource<T> extends AbstractMessageSource<T>
+		implements DeferredAcknowlegmentMessageSource<T> {
+
+	@Override
+	protected Object doReceive() {
+		return receiveMessageBuilder();
```
artembilan: Really no any logic here, so simple `AbstractMessageSource` for AMQP and all the "hard" `Acknowlegment` logic in the `doReceive()` implementation.

What am I missing?

spring-integration-core/src/main/java/org/springframework/integration/channel/MessageSourcePollableChannel.java
```diff
@@ -0,0 +1,62 @@
...
+ * <p>
+ *
+ * ####### THIS IS JUST A TRIAL BALOON #######
+ *
+ * It's an easy way to apply interceptors to messages created by message sources which
+ * can be polled on an ad-hoc basis.
+ * But another option would be to inject an interceptor list into AbstractMessageSource
+ * itself.
+ *
+ * Thoughts?
```
artembilan: Looks like you try to repeat what we really have with the `SourcePollingChannelAdapter` and its `adviceChain()`.
That really doesn't look like functionality for `channel` abstraction: http://www.enterpriseintegrationpatterns.com/patterns/messaging/MessageChannel.html.

I don't have too much arguments against, but I wouldn't commit it into the `master` since it is not a design we follow in the Framework. Isn't it? 

It may live in the tests though, but we even once have deprecated and removed `MessageTransformingChannelInterceptor` although that is really possible...

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpMessageSource.java
```diff
@@ -0,0 +1,234 @@
...
+				}
+			}
+			catch (IOException e) {
+				throw RabbitExceptionTranslator.convertRabbitAccessException(e);
+			}
+			finally {
+				RabbitUtils.closeChannel(this.ackInfo.getChannel());
+				RabbitUtils.closeConnection(this.ackInfo.getConnection());
+			}
+			this.acknowledged = true;
```
artembilan: Is it OK do not mark it as `acknowledged` when we get an exception during this process?

I mean this is outside of the `finally` block, so won't be reachable in case of the exception during `basicAck/basicReject`.

So, any thoughts on the matter, please?

artembilan (changes requested):
I left some comments and explanation requests.
Thanks

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpMessageSource.java
garyrussell: Well, strictly speaking if the ack/nack fails then it should remain false. But I agree it's likely to fail again if tried a second time so might just add noise to the log. OK.

spring-integration-core/src/main/java/org/springframework/integration/channel/MessageSourcePollableChannel.java
garyrussell: The problem is I need a way in SCSt to add interceptors to modify the message produced by the source before  the app gets the message and I am trying to avoid adding a bridge for that purpose (extra hop).

So I have two choices - add some kind of interceptor chain to the message source so that `receive()` returns a modified message, or use this technique and use channel interceptors to `postReceive()` the message.

In the first case, the app would have to use `source.receive()` in the second case he just gets a `PollableChannel` - which is closer to the current message-driven `@StreamListener`.

I suppose we could put this "dummy" channel in SCSt instead.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractDeferredAcknowlegmentMessageSource.java
garyrussell: Yeah - I originally had more logic in this abstract class but determined that it really had to go in the actual implementations.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/SourcePollingChannelAdapter.java
garyrussell: Makes sense.

spring-integration-core/src/main/java/org/springframework/integration/support/AcknowledgmentCallbackFactory.java
garyrussell: I thought about keeping it simple but decided that using a factory would enable users to customize the ack behavior. For example, with Kafka (or Rabbit), the reject could be modified to send the message to another topic.

I think these extension points are useful.

garyrussell: But I agree, the Supplier is not needed.

artembilan: OK with factory, but not OK with `Supplier` 😄

artembilan (approved):
LGTM though, but I would like to see `2018` for Copyright and docs.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/endpoint/SourcePollingChannelAdapter.java
```diff
@@ -212,10 +215,17 @@ protected void handleMessage(Message<?> message) {
 		if (this.shouldTrack) {
 			message = MessageHistory.write(message, this, this.getMessageBuilderFactory());
 		}
+		AcknowledgmentCallback callback = StaticMessageHeaderAccessor.getAcknowledgmentCallback(message);
 		try {
 			this.messagingTemplate.send(getOutputChannel(), message);
+			if (callback != null && !callback.isAcknowledged()) {
+				callback.acknowledge(Status.ACCEPT);
```
artembilan: It's a bit speculative to always do this unconditionally.
The `this.messagingTemplate.send()` might be to the thread hands off `MessageChannel` and end-user still would like to ack in his/her code downstream somewhere. But if we do this way we deprive the general purpose of the `AcknowledgmentCallback`.
Any thoughts?

spring-integration-core/src/main/java/org/springframework/integration/support/AcknowledgmentCallbackFactory.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ */
+public interface AcknowledgmentCallbackFactory<T> {
+
+	/**
+	 * Create the callback.
+	 * @param info information for the callback to process the acknowledgment.
+	 * @return the callback
+	 */
+	AcknowledgmentCallback createCallback(T info);
```
artembilan: So, `@FunctionalInterface` ?

src/reference/asciidoc/polling-consumer.adoc
```diff
@@ -59,6 +59,76 @@ _Version 4.2_ added more flexibility in this area - see <<conditional-pollers>>.
 This chapter is meant to only give a high-level overview regarding Polling Consumers and how they fit into the concept of message channels - <<channel>> and channel adapters - <<channel-adapter>>.
 For more in-depth information regarding Messaging Endpoints in general and Polling Consumers in particular, please see <<endpoint>>.
 
+[[deferred-acks-message-source]]
+==== Deferred Acknowledgment Pollable Message Source
+
+Starting with _version 5.0.1_, certain modules provide `MessageSource` implementations that support deferring acknowledgment until the downstream flow completes (or hands off the message to another thread).
+This is currently limited to the `AmqpMessageSource` and the `KafkaMessageSource` provided by the https://github.com/spring-projects/spring-integration-kafka[spring-kafka-integration extension project, version 3.0.1 or higher].
+
+With these message sources, the `ACKNOWLEDGMENT_CALLBACK` header (see <<message-header-accessor>>) is added to the message.
```
artembilan: `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` because the real header key is `acknowledgmentCallback`.

src/reference/asciidoc/polling-consumer.adoc
```diff
@@ -59,6 +59,76 @@ _Version 4.2_ added more flexibility in this area - see <<conditional-pollers>>.
...
+The value of the header is an instance of `AcknowledgmentCallback`:
+
+[source, java]
+----
+@FunctionalInterface
+public interface AcknowledgmentCallback {
+
+	void acknlowledge(Status status);
```
artembilan: Please, whitespaces instead of tabs in the Docs.

src/reference/asciidoc/polling-consumer.adoc
```diff
@@ -59,6 +59,76 @@ _Version 4.2_ added more flexibility in this area - see <<conditional-pollers>>.
...
+
+...
+
+StaticMessageHeaderAccessor.getAcknowledgmentCallback(received)
+        .acknowledge(Status.ACCEPT);
+----
+
+If the `MessageSource` is wired into a `SourcePollingChannelAdapter`, when the poller thread returns to the adapter after the downstream flow completes, the adapter will check if the acknowledgment has already been acknowledged and, if not, `ACCEPT` it (or `REJECT` it if the flow throws an exception).
+
+To perform ad-hoc polling of a `MessageSource` a `MessageSourcePollingTemplate` is provided; this, too will take care of `ACCEPT` ing or `REJECT` ing the AcknowledgmentCallback when the `MessageHandler` callback returns (or throws an exception).
```
artembilan: Wrap `AcknowledgmentCallback` to the code.

artembilan (changes requested):
Something like that...

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpMessageSource.java
```diff
@@ -0,0 +1,234 @@
...
+	 */
+	public static class AmqpAckInfo {
+
+		private final Connection connection;
+
+		private final Channel channel;
+
+		private final GetResponse getResponse;
+
+		public AmqpAckInfo(Connection connection, Channel channel, GetResponse getResponse) {
```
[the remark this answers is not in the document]
garyrussell: It needs to be public so that users can subclass `AmqpAckCallback`.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/inbound/AmqpMessageSourceIntegrationTests.java
```diff
@@ -0,0 +1,216 @@
...
+					}
+					return bean;
+				}
+
+			};
+		}
+
+		@Bean
+		public ConnectionFactory connectionFactory() {
+			return new CachingConnectionFactory("localhost");
```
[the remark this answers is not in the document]
garyrussell: >I also think if that would be useful to expose CachingConnectionFactory...

We can't do that - spring-rabbit-junit can't depend on spring-rabbit.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/inbound/AmqpMessageSourceTests.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * @author Gary Russell
+ * @since 5.0.1
+ *
+ */
+public class AmqpMessageSourceTests {
+
+	@Test
+	public void testAck() throws Exception {
```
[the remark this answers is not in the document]
garyrussell: Seems like a JUnit/Mockito startup cost; I added another test that runs this one 10 times, result:

screen shot 2018-01-04 at 3 28 30 pm

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageSourcePollingTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+				handler.handleMessage(message);
+				if (acknowledgmentCallback != null && !acknowledgmentCallback.isAcknowledged()) {
+					acknowledgmentCallback.acknowledge(Status.ACCEPT);
+				}
+			}
+			catch (Exception e) {
+				if (acknowledgmentCallback != null && !acknowledgmentCallback.isAcknowledged()) {
+					acknowledgmentCallback.acknowledge(Status.REJECT);
+				}
+				throw new MessageHandlingException(message, e);
```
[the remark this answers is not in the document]
garyrussell: I think this is needed - there is no guarantee that the handler will populate the `failedMessage`.

It's also conditional in `MethodInvokingMessageProcessor`.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/SourcePollingChannelAdapter.java
garyrussell: That's a good point; however, at least with Kafka, that won't work because the consumer is not thread-safe; need to think.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpInboundPolledChannelAdapterSpec.java
```diff
@@ -37,6 +38,11 @@
 		this.target = new AmqpMessageSource(connectionFactory, queue);
 	}
 
+	public AmqpInboundPolledChannelAdapterSpec(ConnectionFactory connectionFactory,
```
artembilan: We usually don't allow to create `Spec` directly - only the factory above.
Although I wouldn't mind to start to make these ctors `protected` for the cases when we missed some property to expose but end-user still would like to use our `Spec` and extend it for omitted property.

spring-integration-core/src/main/java/org/springframework/integration/support/AckUtils.java
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support;
+
+import org.springframework.integration.support.AcknowledgmentCallback.Status;
+
+/**
+ * @author Gary Russell
```
artembilan: I think we need some general JavaDoc here as well. I'm sure users will find it as useful utility.

artembilan (changes requested):
Just a couple simple concerns.
And let's merge it finally!

We can improve/redo in the future 😄

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpInboundPolledChannelAdapterSpec.java
garyrussell: 😄 - I let eclipse create the ctor for me

artembilan: I know. That's pain. And for me IDEA doesn't add `this.` when I generate getters 😢

artembilan (approved):
LGTM.
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/bab850595a9e3c1647886f192df3ac50a787efa9

PR #2311, opened 2017-12-27
Add StaticMessageHeaderAccessor

garyrussell:
- avoids object creation when getting type-safe well-known headers.


garyrussell:
Moved to #2315

PR #2315, opened 2017-12-28
INT-4369: Add DeliveryAttempt header

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4369

Also add StaticMessageHeaderAccessor

- avoids object creation when getting type-safe well-known headers.


artembilan (changes requested):
LGTM.
But I think need to document the feature in the https://docs.spring.io/spring-integration/docs/5.0.0.RELEASE/reference/html/messaging-construction-chapter.html#message-header-accessor.
Plus I guess we need to mentioned there `CLOSEABLE_RESOURCE` header as well.

Thanks

PR #2321, opened 2018-01-08
INT-4376: Upgage to Log4J 2

artembilan:
JIRA: https://jira.spring.io/browse/INT-4376

* Deprecate Log4J 1.x components in favor of newly added
a `Log4j2LevelAdjuster` JUnit `@Rule`
* Update all the logging configuration to Log4J 2


artembilan:
@wilkinsona ,

tell us, please, what a trick we may add to our Gradle config to ignore `compile ('log4j:log4j:1.2.17', optional)` dependency during `springIoCheck` task? Or maybe even whole `spring-integration-test-support` module?
Or the easiest way to reinstate Log4J 1.x dependency in IO Platform? Than we will stick with it there until our Spring Integration `5.1`.

Thank you for any wisdom and advises!

wilkinsona:
I've reinstated Log4j 1 dependency management in Cairo. You should see the benefits of that once I've fixed up a few other things and can then get a new snapshot published

PR #2322, opened 2018-01-11
INT-4371: Polishing - raw message header; docs

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4371


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpMessageSource.java
```diff
@@ -132,6 +135,20 @@ public void setMessageConverter(MessageConverter messageConverter) {
...
+		return this.rawMessageHeader;
+	}
+
+	/**
+	 * Set to true to include the raw spring-amqp message as a header
+	 * with key {@link AmqpMessageHeaderErrorMessageStrategy#AMQP_RAW_MESSAGE},
+	 * enabling callers to have access to the message to process errors.
+	 * @param rawMessageHeader true to include the header.
+	 */
+	public void setRawMessageHeader(boolean rawMessageHeader) {
```
artembilan: Looks like this option must go to Java DSL as well

artembilan (changes requested):
I also don't see any GH hooks activity on the PR, but I guess that is some kind of recent GH downtime.

PR #2325, opened 2018-01-12
Fix SpelPropertyAccessorRegistrar

artembilan:
The `SpelPropertyAccessorRegistrar` ctor and
`add(PropertyAccessor... propertyAccessors)` method use wrong variable
to build a key for the provided `PropertyAccessor`.
Therefore during iteration the next `PropertyAccessor` overrides the
previous and we end up just only with one instance in the registry

* Extract `private static obtainAccessorKey()` method to build
decapitalized key for the `PropertyAccessor`
* Change the map store to the `LinkedHashMap` to keep order of the
provided `PropertyAccessor` s
* Modify `EnableIntegrationTests` to ensure that we support several
`PropertyAccessor` s and in the proper order

**Cherry-pick to 4.3.x**


artembilan:
Has been observed during fix for the SCSt: https://github.com/spring-cloud/spring-cloud-stream/pull/1178

PR #2327, opened 2018-01-15
[INT-3945] Adding async property for ServiceActivator annotation.

yilinwei:
Adding async property for `ServiceActivator`.

This is a continuation of this particular [issue](https://jira.spring.io/browse/INT-3945).

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/config/annotation/AnnotatedEndpointActivationTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/config/annotation/AnnotatedEndpointActivationTests.java
@@ -145,7 +145,7 @@ public class AnnotatedEndpointActivationTests {
 	@SuppressWarnings("unused")
 	private static class AnnotatedEndpoint3 {
 
-	    @ServiceActivator(inputChannel = "input", outputChannel = "output", async = "true")
+		@ServiceActivator(inputChannel = "input", outputChannel = "output", async = "true")
 		public ListenableFuture<String> process(String message) {
 			count++;
 			SettableListenableFuture<String> future = new SettableListenableFuture<>();
--- a/spring-integration-core/src/test/java/org/springframework/integration/config/annotation/AnnotatedEndpointActivationTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/config/annotation/AnnotatedEndpointActivationTests.java
@@ -37,6 +37,8 @@ import org.springframework.messaging.support.GenericMessage;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.springframework.util.concurrent.ListenableFuture;
+import org.springframework.util.concurrent.SettableListenableFuture;
 
 /**
  * @author Dave Syer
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@yilinwei Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2327)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2327) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@yilinwei Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2327)!

artembilan:
Thank you for the contribution, @yilinwei!

Would you mind adding some tests on the matter and simple doc description?

yilinwei:
@artembilan Sure; do you mean site docs?

artembilan:
Well, I mean Reference Manual, especially this chapter: https://github.com/spring-projects/spring-integration/blob/master/src/reference/asciidoc/configuration.adoc, but looks like we don’t have there proper description for all other attributes, so now I think the good JavaDoc should be enough.

Anyway tests must be presented.

yilinwei:
@artembilan OK. I've added a test. Do I need to make the javadoc more detailed or is it OK?

artembilan:
I’ll review tomorrow in the morning. Thanks

spring-integration-core/src/test/java/org/springframework/integration/config/annotation/AnnotatedEndpointActivationTests.java
```diff
@@ -80,6 +82,7 @@ public void sendAndReceive() {
 
 		assertTrue(this.applicationContext.containsBean("annotatedEndpoint.process.serviceActivator"));
 		assertTrue(this.applicationContext.containsBean("annotatedEndpoint2.process.serviceActivator"));
+		assertTrue(this.applicationContext.containsBean("annotatedEndpoint3.process.serviceActivator"));
```
artembilan: The test isn't full. It doesn't prove that `async` is applied. 
I'm sure the test will pass even without that `async = "true"`.

Please, develop the test-case to reflect the expectations.

artembilan (changes requested):
Please, update the affected classes to the current year in their Copyright and add your name to the author list.

spring-integration-core/src/test/java/org/springframework/integration/config/annotation/AnnotatedEndpointActivationTests.java
yilinwei: That's not entirely true; it will fail sometimes since the `ListenableFuture` doesn't get unwrapped. I can separate it out into a separate test case though if you would prefer?

spring-integration-core/src/test/java/org/springframework/integration/config/annotation/AnnotatedEndpointActivationTests.java
```diff
@@ -82,6 +93,15 @@ public void sendAndReceive() {
 		assertTrue(this.applicationContext.containsBean("annotatedEndpoint2.process.serviceActivator"));
 	}
 
+	@Test
+	public void sendAndReceiveAsync() {
+		this.inputAsync.send(new GenericMessage<String>("foo"));
+		Message<?> message = this.outputAsync.receive(100);
+		assertNotNull(message);
+		assertEquals("foo", message.getPayload());
+		assertTrue(this.applicationContext.containsBean("annotatedEndpoint3.process.serviceActivator"));
```
artembilan: If code in the beginning of the test passes, we don't need this assertion any more.

I will play with test during merge.

Thank you for the update! LGTM

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/e6ec86c505ae741326708666ed3712ffd59c0ed1.

Thank you for the contribution, @yilinwei .

Looking forward for more!

PR #2328, opened 2018-01-16
Minor Fixes

garyrussell:
- suppress `rawtypes` compiler warning
- use passive mode for FTP tests (acive fails on one of my machines for some reason)


artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/1bb4f8615103c3702236bab98263c6f31233ca21.

I decided to fix generic args instead of `rawtypes`, but good catch! Looks like one more IDEA flaw 😄

PR #2329, opened 2018-01-16
INT-4366: Fix MulticastSendingMessageHandler

artembilan:
JIRA: https://jira.spring.io/browse/INT-4366

Fix race condition in the `MulticastSendingMessageHandler` around
`multicastSocket` and super `socket` properties.

* Synchronize around `this` and check for the `multicastSocket == null`.
This let the `MulticastSendingMessageHandler` to fully configure and
prepare the socket for use.
* Remove `socket.setInterface(whichNic)` since it is populated by the
`InetSocketAddress` ctor before

**Cherry-pick to 4.3.x**


garyrussell:
Unfortunately, there are many conflicts when cherry-picking - perhaps I should just backport the JIRA fix?

artembilan:
> perhaps I should just backport the JIRA fix?

Fine with me.
Thanks

PR #2330, opened 2018-01-17
Optimize AbstractMessageSources

artembilan:
To avoid `Message` re-creation during `AbstractMessageSource.receive()`
logic, refactor `AbstractMessageSource` implementations
to return `AbstractIntegrationMessageBuilder`

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -51,6 +51,7 @@ import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.file.filters.AcceptOnceFileListFilter;
 import org.springframework.integration.file.filters.FileListFilter;
 import org.springframework.integration.file.filters.ResettableFileListFilter;
+import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
 import org.springframework.util.Assert;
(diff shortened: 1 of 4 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
```diff
@@ -350,8 +350,21 @@ protected void onInit() {
 
 	@Override
 	public Message<File> receive() throws MessagingException {
+		AbstractIntegrationMessageBuilder<File> messageBuilder = doReceive();
+
 		Message<File> message = null;
 
+		if (messageBuilder != null) {
+			message = messageBuilder.build();
```
garyrussell: Why are we building here? Just for the log message?

I guess we could move that log up to the abstract class and have a boolean flag.

garyrussell:
I was surprised to see so few; I guess we should consider refactoring other MS (such as `MailReceivingMessageSource`) to extend AMS - but I guess that's a different task.

spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
artembilan: Because this is `FileReadingMessageSource` - clean `MessageSource` implementation which `receive()` method contract requires the `Message` to be returned.

Not sure what I am missing here...

artembilan:
We can refactor others to the `AMS` but technically for those (like `MailReceivingMessageSource`) it won't matter to match, unless we get additional headers support and metrics.

spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
garyrussell: Hmmm - since I was just looking at the changes, I assumed this was an AMS - since the title of this PR is "Optimize AbstractMessageSources". But now I see it is not.

So, why is this refactoring in the PR at all, without a clarifying mention in the commit comment?

garyrussell:
>unless we get additional headers support

Now that we added to the DSL spec; any implementation of AMS will get headers for free (via the DSL, but not XML, of course).

spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
artembilan: I repushed with an updated commit message.
Was in a hurry in the evening to PR'ing this.

artembilan:
> any implementation of AMS will get headers for free 

Agree, but let's consider that in the next `5.1` already! There might be some breaking changes. For example that `logger.info("Created message: [" + message + "]");` will go to the `AMS` and I guess already as `debug`. There might be some other hidden gems.

PR #2332, opened 2018-01-18
INT-4379: JMS OG Shutdown reply container on stop

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4379

- shutdown the container when the gateway is stopped

Also, improve test suite

- at the end of the tests, hundreds of threads are running, some caused by the above
  but others because `TaskExecutor`s are not shut down
- reduce the number of iterations in the JMS pipeline tests to speed things up
- change more tests to extend `ActiveMQMultiContextTests`, to keep a single broker up

__cherry-pick to 4.3.x__ (perhaps just the gateway fix)

### DIFF
--- a/spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundGatewayTests.java
+++ b/spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundGatewayTests.java
@@ -28,6 +28,7 @@ import static org.mockito.Mockito.when;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -67,7 +68,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @since 2.2.4
  */
-public class JmsOutboundGatewayTests {
+public class JmsOutboundGatewayTests extends ActiveMQMultiContextTests {
 
 	private final Log logger = LogFactory.getLog(this.getClass());
 
@@ -100,8 +101,9 @@ public class JmsOutboundGatewayTests {
 		gateway.setUseReplyContainer(true);
 		ReplyContainerProperties replyContainerProperties = new ReplyContainerProperties();
 		final List<Throwable> errors = new ArrayList<Throwable>();
+		ExecutorService exec = Executors.newFixedThreadPool(10);
 		ErrorHandlingTaskExecutor errorHandlingTaskExecutor =
-				new ErrorHandlingTaskExecutor(Executors.newFixedThreadPool(10), t -> {
+				new ErrorHandlingTaskExecutor(exec, t -> {
 					errors.add(t);
 					throw new RuntimeException(t);
 				});
@@ -157,6 +159,7 @@ public class JmsOutboundGatewayTests {
 		}
 		finally {
 			gateway.stop();
+			exec.shutdownNow();
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundGatewayTests.java
```diff
@@ -67,7 +68,7 @@
  *
  * @since 2.2.4
  */
-public class JmsOutboundGatewayTests {
+public class JmsOutboundGatewayTests extends ActiveMQMultiContextTests {
```
artembilan: How does this `extends ActiveMQMultiContextTests` help here?
I don't see any `connectionFactory` reusing below.
Thanks

spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundGatewayTests.java
```diff
@@ -157,6 +159,7 @@ public void testReplyContainerRecovery() throws Exception {
 		}
 		finally {
 			gateway.stop();
+			exec.shutdownNow();
```
artembilan: Oh! That's cool catch!

I guess we need to revise all the tests in project for similar vulnerability 😄 

I wonder how did you come here? What helped you to figure out the culprit?

Thanks

spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundInsideChainTests.java
```diff
@@ -36,11 +36,12 @@
  * //INT-2275
  *
  * @author Artem Bilan
+ * @author Gary Russell
  */
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration
 @DirtiesContext
-public class JmsOutboundInsideChainTests {
+public class JmsOutboundInsideChainTests extends ActiveMQMultiContextTests {
```
artembilan: Yeah... The same here.

Or does that `new ActiveMQConnectionFactory("vm://localhost?broker.persistent=false")` really mean this:

> to keep a single broker up

and that already doesn't matter how much `new ActiveMQConnectionFactory("vm://localhost?broker.persistent=false")` we are going to call yet?

artembilan (approved):
Just a couple comments awaiting explanations.

I think we can fully backport it into `4.3.x` since we support there Java 8 level for tests.

spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundGatewayTests.java
garyrussell: There are a couple of tests in this class that create and destroy CCFs. When there are no connections, the broker stops. Subclassing this keeps the broker running across all the tests in the class.

spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundGatewayTests.java
garyrussell: I was just lucky; I happened to take a `jstack <pid>` and noticed hundreds of threads including many listener container threads; since we have `@DirtiesContext` on all tests, something didn't seem right. Hence I found the bug in the gateway. I also noticed a bunch of task executor threads so I started looking at some of the tests.

spring-integration-jms/src/test/java/org/springframework/integration/jms/JmsOutboundInsideChainTests.java
garyrussell: Exactly - `vm://...` will connect to an existing broker, or start a new one.

artembilan:
Good. Thank you for all the explanation!

Testing now backport to `4.3.x`.
And at the same time I see a bunch of `Executors` without `shutdown`.

artembilan:
Cherry-picked to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/34c938fe4fc0fe3505faf93bbf10449cbad72859 after resolving conflicts.

PR #2334, opened 2018-01-18
Shutdown test executors in -core

garyrussell:
- also add log adjuster for ftp test that sometimes fails

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/aggregator/BarrierMessageHandlerTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/aggregator/BarrierMessageHandlerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2018 the original author or authors.
  *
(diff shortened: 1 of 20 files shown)


spring-integration-core/src/test/java/org/springframework/integration/aggregator/BarrierMessageHandlerTests.java
```diff
@@ -141,8 +142,8 @@ public void testReplyBeforeRequest() throws Exception {
 		handler.setOutputChannel(outputChannel);
 		handler.setBeanFactory(mock(BeanFactory.class));
 		handler.afterPropertiesSet();
-		Executors.newSingleThreadExecutor()
-				.execute(() -> handler.trigger(MessageBuilder.withPayload("bar").setCorrelationId("foo").build()));
+		ExecutorService exec = Executors.newSingleThreadExecutor();
```
artembilan: In the #2329 I went the way to expose a `SimpleAsyncTaskExecutor` as a class property do not instantiate per test and don't care about `shutdown`, but I think it really doesn't matter from test perspective.

So, merging...

artembilan (approved):

PR #2336, opened 2018-01-19
INT-4382: Fix deadlock in FileWritingMH

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4382

There is a deadlock in the FMWH due to lock ordering.

Writes try to lock the handler monitor while holding the file lock.
The `Flusher` tries to lock the file lock while holding th handler monitor.

- Change the `Flusher`, `stop` and `flushIfNeeded` methods to not call close
  on the file state while holding the handler monitor.

__cherry pick to 4.3.x__ (fix <> for Java 6)

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
@@ -481,7 +481,25 @@ public class FileWritingMessageHandler extends AbstractReplyProducingMessageHand
 				this.flushTask = null;
 			}
 		}
-		new Flusher().run();
+		Flusher flusher = new Flusher();
+		flusher.run();
+		boolean needInterrupt = this.fileStates.size() > 0;
+		int n = 0;
+		while (n++ < 10 && this.fileStates.size() > 0) {
+			try {
+				Thread.sleep(1);
+			}
+			catch (InterruptedException e) {
+				// cancel the interrupt
+			}
+			flusher.run();
+		}
+		if (this.fileStates.size() > 0) {
+			this.logger.error("Failed to flush after multiple attempts, while stopping: " + this.fileStates.keySet());
+		}
+		if (needInterrupt) {
+			Thread.currentThread().interrupt();
+		}
 	}
 
 	@Override
@@ -864,10 +882,10 @@ public class FileWritingMessageHandler extends AbstractReplyProducingMessageHand
 
 	private synchronized FileState getFileState(final File fileToWriteTo, boolean isString)
 			throws FileNotFoundException {
-		String absolutePath = fileToWriteTo.getAbsolutePath();
 		FileState state;
 		boolean appendNoFlush = FileExistsMode.APPEND_NO_FLUSH.equals(this.fileExistsMode);
 		if (appendNoFlush) {
+			String absolutePath = fileToWriteTo.getAbsolutePath();
 			state = this.fileStates.get(absolutePath);
 			if (state != null && ((isString && state.stream != null) || (!isString && state.writer != null))) {
 				state.close();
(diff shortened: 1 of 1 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -977,6 +987,32 @@ private synchronized void clearState(final File fileToWriteTo, final FileState s
...
+				}
+			}
+			else { // interrupted (stop), re-add
+				interrupted = true;
+				toRestore.put(entry.getKey(), entry.getValue());
+			}
+		}
+		if (interrupted) {
+			if (FileWritingMessageHandler.this.logger.isDebugEnabled()) {
+				FileWritingMessageHandler.this.logger.debug("Flushed: " + toRestore.keySet());
```
artembilan: Maybe do you mean `Restored:` ?

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -977,6 +987,32 @@ private synchronized void clearState(final File fileToWriteTo, final FileState s
...
+		boolean interrupted = false;
+		for (Entry<String, FileState> entry : toRemove.entrySet()) {
+			if (!interrupted && entry.getValue().close()) {
+				if (FileWritingMessageHandler.this.logger.isDebugEnabled()) {
+					FileWritingMessageHandler.this.logger.debug("Flushed: " + entry.getKey());
+				}
+			}
+			else { // interrupted (stop), re-add
+				interrupted = true;
+				toRestore.put(entry.getKey(), entry.getValue());
```
artembilan: OK. I guess we got here from the `InterruptedException` during `this.flushTask.cancel(true);` in the `stop()`.
Shouldn't we wrap this `put()` to the `synchronized (this) {` as well?
Or `new Flusher().run()` won't be called until the `this.flushTask.cancel(true);` finishes properly and all the puts have happened?

Thanks

artembilan (changes requested):
I see some optimization in the `getFileState()` - we don't need to call `fileToWriteTo.getAbsolutePath()` if `!appendNoFlush`. Plus I think we can refactor `flushIfNeeded()` methods do not duplicate code, but that already Java 8 `Predicate` matter and only for `master` therefore.

Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
garyrussell: ?? this is just putting to a local variable; we grab the lock below to restore the values to the field.

artembilan: Doh. I see now. Confused with `put` here and `put` there. Never mind.
Although anyway: can't this `this.fileStates.putIfAbsent` be called already after that `Flusher.run()`?

I mean aren't there a situation when we may still have some entries even after `stop()`.

Thanks

garyrussell: It's true that if the thread calling stop() is interrupted, there could be some left; I'll give it some more thought.

artembilan: Maybe `Semaphore` to permit `Flusher` to clean up ?
Just a thought...

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -481,7 +481,25 @@ public void stop() {
...
+			catch (InterruptedException e) {
+				// cancel the interrupt
+			}
+			flusher.run();
+		}
+		if (this.fileStates.size() > 0) {
+			this.logger.error("Failed to flush after multiple attempts, while stopping: " + this.fileStates.keySet());
+		}
+		if (needInterrupt) {
+			Thread.currentThread().interrupt();
```
artembilan: Why do you interrupt the thread who calls this `stop()`?

garyrussell: Because if we don't complete the flush in the first attempt, the user's thread was interrupted; we can't "eat" that interrupt so after we are complete (or give up) we re-interrupt the thread. That said, I see a problem; pushing another commit...

garyrussell: Forget that; all is ok; we reset the interrupt flag each time in the loop.

artembilan:
I'm fine to merge, but we have failed Travis.
More over we have failed CI: https://build.spring.io/browse/INT-MASTER.

So, I kinda would like to have my PR #2337 merged first to try to bring CI to life.
Then come back to other not TCP/IP related tasks.

Thanks

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/55d92edd4e9d27bc7157534a76273152f4b17481 and cherry-picked as https://github.com/spring-projects/spring-integration/commit/21c4932abe0251541e2543444c247ccb6933dadc after resolving conflicts.

PR #2337, opened 2018-01-19
Fix NPE in the UnicastSendingMessageHandler

artembilan:
https://build.spring.io/browse/INT-MASTER-898
https://build.spring.io/browse/INT-SI43X-224

There are situations when `socket` is `null`'ed, but
`UnicastSendingMessageHandler.run()` method calls `getSocket()` which is
overridden in the `MulticastSendingMessageHandler`.
And that one has the logic around its `multicastSocket` what causes
missing the call for the `createSocket()` to refresh the `socket`
property.

* Override `closeSocketIfNeeded()` in the `MulticastSendingMessageHandler`
and `null` `multicastSocket` property, so the next `getSocket()` will
renew properties
* Fix "No `BeanFactory`" warning in the `UdpChannelAdapterTests`

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/MulticastSendingMessageHandler.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/MulticastSendingMessageHandler.java
@@ -201,4 +201,10 @@ public class MulticastSendingMessageHandler extends UnicastSendingMessageHandler
 		}
 	}
 
+	@Override
+	protected void closeSocketIfNeeded() {
+		this.multicastSocket = null;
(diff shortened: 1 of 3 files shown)


artembilan:
Not sure if this is the reason of those CI failures, but at least need to try.
Otherwise that might be some Linux specifics, because it doesn't fail on Windows and I guess for you on Mac.

Any other ideas?

artembilan:
Hm. Travis hangs on JMS tests in this my PR as well.
Works fine locally though... 
Some Linux incompatibility again?

artembilan:
Superseded by #2342

PR #2338, opened 2018-01-19
INT-4384: Micrometer Support

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4384

Add support for micrometer metrics collection.

Initial commit.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/management/CounterFacade.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/management/CounterFacade.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.management;
+
+import java.time.Duration;
+
+/**
+ * A facade abstraction over some timer that tracks {@link Duration}.
+ *
+ * @author Gary Russell
+ * @since 5.0.1
+ *
+ */
+public interface CounterFacade {
+
+	/**
+	 * Increment the counter.
+	 */
+	void increment();
+
+}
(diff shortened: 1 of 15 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/management/TimerFacade.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.management;
+
+import java.time.Duration;
+
+/**
+ * A facade abstraction over some counter.
```
jkschneider: "timer" instead of "counter"

spring-integration-core/src/main/java/org/springframework/integration/support/management/CounterFacade.java
```diff
@@ -0,0 +1,35 @@
...
+import java.time.Duration;
+
+/**
+ * A facade abstraction over some timer that tracks {@link Duration}.
+ *
+ * @author Gary Russell
+ * @since 5.0.1
+ *
+ */
+public interface CounterFacade {
```
jkschneider: Micrometer is intended to be the facade over a bunch of monitoring systems. Why have a facade of the facade?

spring-integration-core/src/main/java/org/springframework/integration/support/management/DefaultMessageChannelMetrics.java
```diff
@@ -114,7 +152,11 @@ public void afterSend(MetricsContext context, boolean result) {
 			if (isFullStatsEnabled()) {
 				long now = System.nanoTime();
```
jkschneider: A consequence of adding `TimerFacade` -- it would be preferable to use Micrometer's `Timer#record(Runnable)` or variants so that the provided `Clock` is honored. `System.nanoTime` bypasses the provided clock.

garyrussell: `record(Runnable)` would require some major changes which we can't really do in a point release, and in any case would be tough to do by Wednesday, which is our release date.

We can rework in 5.1; if this is a show-stopper we will have to rethink.

spring-integration-core/src/main/java/org/springframework/integration/support/management/CounterFacade.java
garyrussell: The problem is timing (pun); micrometer is still RC so we can't have a hard dependency on it in our 5.0.1 release (due next Wednesday).

spring-integration-core/src/main/java/org/springframework/integration/support/management/DefaultMessageChannelMetrics.java
jkschneider: Not a show-stopper, just something to keep in mind going forward.

garyrussell:
Closing for now - deferred to 5.0.2

garyrussell:
Replaced by #2339

PR #2339, opened 2018-01-22
INT-4384: Micrometer Support

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4384

Add support for micrometer metrics collection.

Initial commit.

__DO NOT MERGE until after 5.0.1__

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -319,7 +319,7 @@ project('spring-integration-core') {
 		}
 		compile("io.fastjson:boon:$boonVersion", optional)
 		compile("com.esotericsoftware:kryo-shaded:$kryoShadedVersion", optional)
-		compile("io.micrometer:micrometer-core:$micrometerVersion", optional)
+		compile "io.micrometer:micrometer-core:$micrometerVersion"
 
 		testCompile ("org.aspectj:aspectjweaver:$aspectjVersion")
 		testCompile "io.projectreactor:reactor-test:$reactorVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ subprojects { subproject ->
 		jythonVersion = '2.5.3'
 		kryoShadedVersion = '3.0.3'
 		log4jVersion = '2.10.0'
-		micrometerVersion = '1.0.0-rc.7'
+		micrometerVersion = '1.0.0-rc.8'
 		mockitoVersion = '2.11.0'
 		mysqlVersion = '6.0.6'
 		pahoMqttClientVersion = '1.2.0'
--- a/build.gradle
+++ b/build.gradle
@@ -274,6 +274,7 @@ project('spring-integration-test-support') {
 		compile ("org.mockito:mockito-core:$mockitoVersion") {
 			exclude group: 'org.hamcrest', module: 'hamcrest-core'
 		}
+		compile "org.assertj:assertj-core:$assertjVersion"
 		compile "org.springframework:spring-context:$springVersion"
 		compile "org.springframework:spring-messaging:$springVersion"
 		compile "org.springframework:spring-test:$springVersion"
@@ -318,9 +319,8 @@ project('spring-integration-core') {
 		}
 		compile("io.fastjson:boon:$boonVersion", optional)
 		compile("com.esotericsoftware:kryo-shaded:$kryoShadedVersion", optional)
-		compile "io.micrometer:micrometer-core:$micrometerVersion"
+		compile("io.micrometer:micrometer-core:$micrometerVersion", optional)
 
-		testCompile "org.assertj:assertj-core:$assertjVersion"
 		testCompile ("org.aspectj:aspectjweaver:$aspectjVersion")
 		testCompile "io.projectreactor:reactor-test:$reactorVersion"
 		testCompile ("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jackson2Version")
--- a/build.gradle
+++ b/build.gradle
@@ -88,8 +88,9 @@ subprojects { subproject ->
 
 	ext {
 		activeMqVersion = '5.15.2'
-		aspectjVersion = '1.8.13'
 		apacheSshdVersion = '1.6.0'
+		aspectjVersion = '1.8.13'
+		assertjVersion = '2.6.0'
 		boonVersion = '0.34'
 		commonsDbcp2Version = '2.1.1'
 		commonsIoVersion = '2.4'
@@ -119,6 +120,7 @@ subprojects { subproject ->
 		jythonVersion = '2.5.3'
 		kryoShadedVersion = '3.0.3'
 		log4jVersion = '2.10.0'
+		micrometerVersion = '1.0.0-rc.7'
 		mockitoVersion = '2.11.0'
 		mysqlVersion = '6.0.6'
 		pahoMqttClientVersion = '1.2.0'
@@ -316,7 +318,9 @@ project('spring-integration-core') {
 		}
 		compile("io.fastjson:boon:$boonVersion", optional)
 		compile("com.esotericsoftware:kryo-shaded:$kryoShadedVersion", optional)
+		compile "io.micrometer:micrometer-core:$micrometerVersion"
 
+		testCompile "org.assertj:assertj-core:$assertjVersion"
 		testCompile ("org.aspectj:aspectjweaver:$aspectjVersion")
 		testCompile "io.projectreactor:reactor-test:$reactorVersion"
 		testCompile ("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jackson2Version")
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -432,13 +432,19 @@ public abstract class AbstractMessageChannel extends IntegrationObjectSupport
 					return false;
 				}
 			}
-			if (countsEnabled) {
-				metrics = channelMetrics.beforeSend();
+			if (channelMetrics.getTimer() != null && countsEnabled) {
+				final Message<?> messageToSend = message;
+				sent = channelMetrics.getTimer().recordCallable(() -> doSend(messageToSend, timeout));
 			}
-			sent = this.doSend(message, timeout);
-			if (countsEnabled) {
-				channelMetrics.afterSend(metrics, sent);
-				metricsProcessed = true;
+			else {
+				if (countsEnabled) {
+					metrics = channelMetrics.beforeSend();
+				}
+				sent = doSend(message, timeout);
+				if (countsEnabled) {
+					channelMetrics.afterSend(metrics, sent);
+					metricsProcessed = true;
+				}
 			}
 
 			if (debugEnabled) {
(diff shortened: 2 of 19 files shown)


garyrussell:
@snicoll @philwebb @jkschneider

Since Boot auto configures a `MeterRegistry`, Spring Integration autoconfiguration could configure an SI `MicrometerMetricsFactory` (that gets a reference to the `MeterRegistry` - see [here](https://github.com/spring-projects/spring-integration/pull/2339/files#diff-94e519d70db721b42405d50a4b146596R103) ).

This would switch SI metrics to use Micrometer instead of its legacy metrics.

On the other hand, existing SI users might be thrown off by that, but they could easily switch back the the default by adding their own bean definition for an SI `MetricsFactory`

What would you recommend?

philwebb:
@garyrussell I think we should go with micrometer by default and document how to override.

garyrussell:
ok; I'll submit a PR once this is merged (after 5.0.1 this week).

jkschneider:
Is the idea that in a future major release that we will eliminate `MetricsFactory` altogether? After all, if SI users need some custom metrics exposition format or publishing scheme (pulling scheme), they can accomplish that with a custom `MeterRegistry` without the need for some further intermediate interface.

If so, could we mark `MetricsFactory` deprecated?

garyrussell:
>If so, could we mark MetricsFactory deprecated?

I don't follow; the `IntegrationManagementConfigurer` iterates over all the integration components in the context (`MessageHandler`s, `MessageChannel`s etc) then delegates to the `MetricsFactory` bean (which in turn delegates to the `MeterRegistry`) to create the timers and counters.

We will certainly deprecate the `DefaultMetricsFactory` (and simplify things further) in 5.1, but we'll still need "something" to create the `Meter`s, right?

garyrussell:
Ah - perhaps you mean the components could get their own `Meter`s from the registry rather than having the configurer do that.

Yes, we could do that.

jkschneider:
> but we'll still need "something" to create the Meters, right?

The function of a `MeterRegistry` is to create and hold `Meter`s. I'm wondering if we need `MetricsFactory` at all.

artembilan:
Would you mind rebasing this to current `master` and maybe even update dependencies for the Micormeter to the latest?
Then it will be good to go for review/merging.

Thanks

artembilan:
Gary, gentle ping.

Thanks

garyrussell:
Doh - I did it on my old INT-4384 branch 😦

build.gradle
```diff
@@ -88,8 +88,9 @@ subprojects { subproject ->
 
 	ext {
 		activeMqVersion = '5.15.2'
-		aspectjVersion = '1.8.13'
 		apacheSshdVersion = '1.6.0'
+		aspectjVersion = '1.8.13'
+		assertjVersion = '2.6.0'
```
artembilan: Why do we need to introduce one more testing dependency?

artembilan (changes requested):
One concern at a glance.

Will dig it more in the morning

build.gradle
garyrussell: Over time, I would like to migrate everything to AssertJ to be consistent with Boot and cloud - I just don't have Phil Webb's diligence to do it all in one go. 

Similar to BDDMockito migration.

It's becoming a bit painful to use different libraries in the various projects I contribute to.

If you don't agree, I will reconsider.

artembilan: Yes, I agree about consistency and I definitely like AssertJ with all its smart API.

Only the problem that we even have failed for several years already to update our imports order to align at least with Spring Framework. And also we have failed the same way to turn one all the Checkstyle rules.

So, that's how I'm afraid we will fail with AssertJ as well.

As you said: too much various projects we contribute to...

But at least we can start to use AssertJ in new tests! 😄 
OK, let's see how is it going to work!

jkschneider: I think at one point I wrote a [Rewrite](https://github.com/Netflix-Skunkworks/rewrite) rule to automatically convert JUnit style asserts to AssertJ, if that's interesting to you.

build.gradle
```diff
@@ -316,7 +318,9 @@ project('spring-integration-core') {
 		}
 		compile("io.fastjson:boon:$boonVersion", optional)
 		compile("com.esotericsoftware:kryo-shaded:$kryoShadedVersion", optional)
+		compile "io.micrometer:micrometer-core:$micrometerVersion"
 
+		testCompile "org.assertj:assertj-core:$assertjVersion"
```
artembilan: Since we agreed about supporting AssertJ this has to go to the `spring-integration-test-support` to let all the modules to pull transitive AssertJ for free.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -432,13 +432,19 @@ public boolean send(Message<?> message, long timeout) {
 					return false;
 				}
 			}
-			if (countsEnabled) {
-				metrics = channelMetrics.beforeSend();
+			if (channelMetrics.getTimer() != null && countsEnabled) {
```
artembilan: For better performance should be swapped: `if (countsEnabled && channelMetrics.getTimer() != null)`.
OTOH doesn't this look cleaner:
```
	if (countsEnabled) {
				if (channelMetrics.getTimer() != null) {
					final Message<?> messageToSend = message;
					sent = channelMetrics.getTimer().recordCallable(() -> doSend(messageToSend, timeout));
				}
				else {
					metrics = channelMetrics.beforeSend();
					sent = doSend(message, timeout);
					channelMetrics.afterSend(metrics, sent);
					metricsProcessed = true;
				}
			}
			else {
				sent = doSend(message, timeout);
			}
```
?

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageHandler.java
```diff
@@ -135,17 +135,39 @@ public void handleMessage(Message<?> message) {
 			if (this.shouldTrack) {
 				message = MessageHistory.write(message, this, this.getMessageBuilderFactory());
 			}
-			if (countsEnabled) {
-				start = handlerMetrics.beforeHandle();
+			if (handlerMetrics.getTimer() != null && countsEnabled) {
```
artembilan: DITTO here.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageHandler.java
```diff
@@ -135,17 +135,39 @@ public void handleMessage(Message<?> message) {
...
+				final Message<?> messageToSend = message;
+				handlerMetrics.getTimer().record(() -> {
+					try {
+						handleMessageInternal(messageToSend);
+					}
+					catch (Exception e) {
```
artembilan: Can't we use `recordCallable()` with `return null;` instead ?

spring-integration-core/src/main/java/org/springframework/integration/support/management/AbstractMessageChannelMetrics.java
```diff
@@ -32,10 +35,20 @@
...
+
+	private final Counter errorCounter;
+
 	private volatile boolean fullStatsEnabled;
 
 	public AbstractMessageChannelMetrics(String name) {
+		this(name, null, null);
+	}
+
+	public AbstractMessageChannelMetrics(String name, Timer timer, Counter errorCounter) {
```
artembilan: JavaDocs?

spring-integration-core/src/main/java/org/springframework/integration/support/management/AbstractMessageChannelMetrics.java
```diff
@@ -72,6 +85,14 @@ protected boolean isFullStatsEnabled() {
 	 */
 	public abstract void reset();
 
+	public Timer getTimer() {
+		return this.timer;
+	}
+
+	public Counter getErrorCounter() {
```
artembilan: JavaDocs on new methods ?

spring-integration-core/src/main/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsFactory.java
```diff
@@ -0,0 +1,133 @@
...
+
+	public Function<String, String> counterNameProvider = n -> n + ".counter";
+
+	public Function<String, String> errorCounterNameProvider = n -> n + ".errorCounter";
+
+	public Function<String, String[]> timerTagProvider = NO_TAGS;
+
+	public Function<String, String[]> counterTagProvider = NO_TAGS;
+
+	public Function<String, String[]> errorCounterTagProvider = NO_TAGS;
```
artembilan: Properties must go before any methods in the class.
More over I think they have to be `private`.

spring-integration-core/src/main/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsFactory.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ */
+public class MicrometerMetricsFactory implements MetricsFactory, MessageSourceMetricsConfigurer {
+
+	private static final Function<String, String[]> NO_TAGS = n -> new String[0];
+
+	private final MeterRegistry meterRegistry;
+
+	public MicrometerMetricsFactory(MeterRegistry meterRegistry) {
+		this.meterRegistry = meterRegistry;
```
artembilan: `Assert.notNull()` ?

spring-integration-core/src/main/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsFactory.java
```diff
@@ -0,0 +1,133 @@
...
+import org.springframework.integration.support.management.DefaultMessageHandlerMetrics;
+import org.springframework.integration.support.management.MessageSourceMetrics;
+import org.springframework.integration.support.management.MessageSourceMetricsConfigurer;
+import org.springframework.integration.support.management.MetricsFactory;
+import org.springframework.util.Assert;
+
+import io.micrometer.core.instrument.MeterRegistry;
+
+/**
+ * Micrometer implementation.
```
artembilan: I think need more JavaDocs, e.g. explain the reason of the `MessageSourceMetricsConfigurer`.

artembilan (changes requested):
That's all.

Nothing critical so far.

Good stuff and if we agree that this new non-`optional` dependency is fine in the point release, we need to have `What's New` entry and good to go to final polishing and merging!

build.gradle
garyrussell: @jkschneider I'd certainly be interested in taking a look, but I'd probably start with a small project than SI, though 😄

garyrussell:
@philwebb @jkschneider After this is merged, [this is my proposed boot change](https://github.com/garyrussell/spring-boot/commit/abeb459f3b81963e007936d9634b327d76a68896) to enable Micrometer by default (in case you want to take an early look).

build.gradle
```diff
@@ -318,9 +319,8 @@ project('spring-integration-core') {
 		}
 		compile("io.fastjson:boon:$boonVersion", optional)
 		compile("com.esotericsoftware:kryo-shaded:$kryoShadedVersion", optional)
-		compile "io.micrometer:micrometer-core:$micrometerVersion"
+		compile("io.micrometer:micrometer-core:$micrometerVersion", optional)
```
artembilan: No, I think it's not going to work.
Look here: `AbstractMessageSource implements MessageSourceMetrics` and the last one has hard dependency on the `io.micrometer.core.instrument.Counter`. Therefore no any application is going to work without `micrometer-core` on CP.

Therefore this is a breaking change.

I'm OK to live with a hard dependency here. In fact it comes transitively.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -300,3 +300,9 @@ The `JdbcMessageChannelStore` now provides setter for the `ChannelMessageStorePr
 The `ExpressionEvaluatingSqlParameterSourceFactory` now provides setter for the sqlParameterTypes allowing users to customize sql types of the parameters.
 
 See <<jdbc>> for more information.
+
+
+==== Metrics Changes
+
+http://micrometer.io/[Micrometer] application monitoring is now supported (since _version 5.0.2).
```
artembilan: Missed `_` after 5.0.2

artembilan (changes requested):

artembilan (approved):

PR #2341, opened 2018-01-22
Fixes issues/2340 WebFluxMessageHandlerSpec should support chaining

asarkar:

### DIFF
--- a/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/dsl/WebFluxMessageHandlerSpec.java
+++ b/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/dsl/WebFluxMessageHandlerSpec.java
@@ -35,6 +35,7 @@ import reactor.core.publisher.Mono;
  *
  * @author Shiliang Li
  * @author Artem Bilan
+ * @author Abhijit Sarkar
  *
  * @since 5.0
  *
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Cool!

Any chance to have some simple test-case modification for confirmation? See `WebFluxDslTests`.

Plus your name to the `@author` list for each affected class and also would be great to have a proper commit message: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc#provide-a-link-to-the-jira-issue-in-the-associated-pull-request

You can reject all the request from and I'll fix everything during merge.

Thank you!

asarkar:
Added my name to author list, and updated commit message. Tests for those 2 methods don't exist, so it seems like outside the scope of this PR to add new ones.

artembilan:
Well, definitely yes, it is the scope of this PR. Since we hadn't had tests before we have bumped to this issue.
So, if we are fixing it here, let's do that proper way do not come back to the problem anymore.

Moreover the rule is like: any PR should be supplied with tests or should provide a confirmation that existing tests cover the fix.

On the other hand we don't need extra test. I believe that we can modify existing tests at least to call these method for confirmation that the method chain issue is covered by test suit.

WDYT?

Thanks for understanding.

asarkar:
I don't think it's fair to force me to write tests that someone else didn't, but I'm waiting to get my code working with this fix, so I'll try to add tests and report back.

asarkar:
I wrote a test (see attached) but it keeps failing, and being new to the codebase, I'm not exactly sure why. This is taking longer than I expected, so if this can't be fixed quickly, I'll bail. If that means the PR can't be merged, so be it.
[WebFluxDslTests.txt](https://github.com/spring-projects/spring-integration/files/1654005/WebFluxDslTests.txt)

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/56ffed080120ebb32c2e9e8db39388cfe85ad989.

@asarkar ,

Thank you for your effort!
I've utilized something from your test-case and minimized it to exactly the feature we would like to test.

Looking forward for more from you!

PR #2342, opened 2018-01-22
INT-4366: Fix MulticastSendingMH race condition

artembilan:
JIRA: https://jira.spring.io/browse/INT-4366

The `MulticastSendingMessageHandler.getSocket()` doesn't guard around
`this.multicastSocket` property causing `NPE` and other inconsistency
in the multi-threaded environment

* Make the whole `MulticastSendingMessageHandler.getSocket()` as
`synchronized` like it is with the super method
* Reuse `closeSocketIfNeeded()` in the
`UnicastSendingMessageHandler.handleMessageInternal()`
* Fix type in the `UnicastSendingMessageHandler` logging message
* Fix `UdpChannelAdapterTests` for missed `BeanFactory` for the SpEL
and also `MulticastSendingMessageHandler.stop()` in one missed places

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/MulticastSendingMessageHandler.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/MulticastSendingMessageHandler.java
@@ -125,13 +125,11 @@ public class MulticastSendingMessageHandler extends UnicastSendingMessageHandler
 	}
 
 	@Override
-	protected DatagramSocket getSocket() throws IOException {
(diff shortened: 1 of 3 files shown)


artembilan:
The CI Test Plan confirms that the fix works on Linux: https://build.spring.io/browse/INT-BSP32-43

artembilan:
Cherry-picked to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/6a86f5231d300a61ce9e82f32a817261d714e3b6.

PR #2345, opened 2018-01-30
INT-4388: Add UdpServerListeningEvent

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4388

Publish an event when the UDP server socket is established.

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UdpServerListeningEvent.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UdpServerListeningEvent.java
@@ -17,7 +17,6 @@
 package org.springframework.integration.ip.udp;
 
 import org.springframework.integration.ip.event.IpIntegrationEvent;
-import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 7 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UdpServerListeningEvent.java
```diff
@@ -0,0 +1,46 @@
...
+ * @since 5.0.2
+ */
+@SuppressWarnings("serial")
+public class UdpServerListeningEvent extends IpIntegrationEvent {
+
+	private final int port;
+
+	public UdpServerListeningEvent(UnicastReceivingChannelAdapter adapter, int port) {
+		super(adapter);
+		Assert.notNull(adapter, "'adapter' cannot be null");
```
artembilan: We don't need this. See super-super-super class:
```
 public EventObject(Object source) {
        if (source == null)
            throw new IllegalArgumentException("null source");

        this.source = source;
    }
```

artembilan (changes requested):

PR #2347, opened 2018-02-01
Fixed issue where the Java DSL will always create NIO connection factories regardless of which type you specified.

### LINKED ISSUE #2346: Spring Integration IP Java DSL creates wrong connection factories
In version 5.0 the Java DSL will always create NIO connection factories regardless of which type you specified.

TYsewyn:
Fixes #2346

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/Tcp.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/Tcp.java
@@ -23,6 +23,7 @@ import org.springframework.integration.ip.tcp.connection.AbstractConnectionFacto
  * Factory methods for TCP.
  *
  * @author Gary Russell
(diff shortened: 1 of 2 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/Tcp.java
```diff
@@ -37,7 +38,7 @@
 	 * Boolean indicating the connection factory should not use NIO
 	 * (default).
 	 */
-	public static final boolean NET = true;
+	public static final boolean NET = false;
```
artembilan: Uh! That's nasty.

I suggest to deprecate both these constants altogether because they are not used in the Framework at all.

And use just simply `true/false` in those factories.

artembilan (changes requested):

PR #2348, opened 2018-02-01
Fix UDP Tangles

garyrussell:
- Class tangle in new event
- Package tangle udp <-> util

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UdpServerListeningEvent.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UdpServerListeningEvent.java
@@ -32,7 +32,7 @@ public class UdpServerListeningEvent extends IpIntegrationEvent {
 
 	private final int port;
 
-	public UdpServerListeningEvent(UnicastReceivingChannelAdapter adapter, int port) {
+	public UdpServerListeningEvent(Object adapter, int port) {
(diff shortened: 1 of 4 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UdpServerListeningEvent.java
```diff
@@ -32,7 +32,7 @@
 
 	private final int port;
 
-	public UdpServerListeningEvent(UnicastReceivingChannelAdapter adapter, int port) {
+	public UdpServerListeningEvent(Object adapter, int port) {
```
artembilan: Gary, I don't understand this problem.
I don't see how `org.springframework.integration.ip.util` and `org.springframework.integration.ip.udp` packages use each other. Unless you have included to your report `test` directory as well...

garyrussell: The event and the adapter reference each other (class tangle)

`DatagramPacketMessageMapper` uses `RegexUtils` and `TestingUtilities` referenced the `UnicastReceivingChannelAdapter` (package tangle)

artembilan (approved):
Cool! Thanks

I approve the PR, but feel free to add more fixes!

Will merge somewhere in the afternoon.

PR #2349, opened 2018-02-02
INT-4394: Fix compatibility with Jackson JSON

artembilan:
JIRA: https://jira.spring.io/browse/INT-4394

Looks like Jackson has became much smarter and now it stores
the target type for the object, not returned interface.

We can't use Jackson annotations on the Framework classes and
we can't restrict the `messagingAwareMapper()` `ObjectMapper` just
to use fields for all the object passed through it.

As a compromise solution we shouldn't use
`Collections.unmodifiableList()` in the getter.
The copy of the internal collection is pretty enough to protect
our object from mutation.

**Cherry-pick to 4.3.x (excluding `build.gradle` change)**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,7 +105,7 @@ subprojects { subproject ->
 		hibernateVersion = '5.2.10.Final'
 		hsqldbVersion = '2.4.0'
 		h2Version = '1.4.196'
-		jackson2Version = '2.9.1'
+		jackson2Version = '2.9.4'
 		javaxActivationVersion = '1.1.1'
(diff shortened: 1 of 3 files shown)


artembilan:
@garyrussell ,

any chances to review/merge this?

We are failing against IO Cairo for a long time already: https://build.spring.io/browse/INT-FATS5IC

Thanks

artembilan:
... and cherry-picked to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/4f0f3d11fd2c8eaf7c4f8c42452fbf0b806111e1

PR #2355, opened 2018-02-09
INT-4396: Add retrying lock in case of exception

artembilan:
JIRA: https://jira.spring.io/browse/INT-4396

When target distributed `Lock` implementation throws an exception, e.g.
in case of no connection to the service, the `LockRegistryLeaderInitiator`
exists the loop and can come back to the elections only after restart

* Catch all the exception on `this.lock.tryLock()` and resubmit
`LeaderSelector` for a new locking cycle if
`LockRegistryLeaderInitiator.isRunning()`

* Remove diagnostics from the `JdbcLockRegistryLeaderInitiatorTests`
since this fix confirms that we just didn't have a reconnect logic before
when this test failed sporadically

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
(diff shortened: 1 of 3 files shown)


garyrussell:
Conflicts on cherry-pick - probably easier for you to figure out the differences; thanks.

artembilan:
OK. Will do.
Sorry for the conflicts.

artembilan:
Cherry-picked as https://github.com/spring-projects/spring-integration/commit/935da81ab3cc8c7a86345e683299ab02e9bf0da3

Issue #2354, opened 2018-02-09
Allow ExecutorService to be injected


boaty82:
Please, please, please can this be injected through an overloaded constructor, and make the existing constructors chain on with a default value as per current defaulting logic.

I have need to use this class with another executorservice/threadfactory and it's looks like I'm going to have to now duplicate a lot of code

https://github.com/spring-projects/spring-integration/blob/e53e8941a18cb3bec30187abe32d66a14afa48ac/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java#L78-L79

artembilan:
Yes, we can do that.

[CONTRIBUTING](https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc) is welcome!

boaty82:
Okie dokie I shall certainly do so!

boaty82:
I did start refactoring the class too, to make it smaller and have a single responsibility, but felt I should possibly do this under another pull request.

May have some other feature requests (with PoCs) and some other helpful/reusable implementations I can submit

artembilan:
I don't find that class too bad by design so we need to refactor it, but you are feel free to share your vision anyway. However right: let's do that in the separate PR!

boaty82:
I just felt the class in particular didn’t follow the open/closed principle. It’s complexity is quite large too. But yes agree a different PR

PR #2357, opened 2018-02-10
INT-2354 allow ExecutorService to be injected via constructor

boaty82:
INT-2354: Add support for ExecutorService injection in LockRegistryLeaderInitiator

https://github.com/spring-projects/spring-integration/issues/2354

* add 2 constructor overloads

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -27,6 +27,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.context.SmartLifecycle;
@@ -57,10 +58,11 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Vedran Pavic
  * @author Glenn Renfro
+ * @author Kiel Boatman
  *
  * @since 4.3.1
  */
-public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBean, ApplicationEventPublisherAware {
+public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBean, ApplicationEventPublisherAware, InitializingBean {
 
 	public static final long DEFAULT_HEART_BEAT_TIME = 500L;
 
@@ -73,10 +75,15 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 	private final Object lifecycleMonitor = new Object();
 
 	/**
+	 * flag to denote whether the executor service should be shutdown when {@link #destroy()} is called
+	 * when a {@link ExecutorService} is provided via the setter then this should not be invoked
+	 */
+	private boolean shutdownExecutorService;
+
+	/**
 	 * Executor service for running leadership daemon.
 	 */
-	private final ExecutorService executorService =
-			Executors.newSingleThreadExecutor(new CustomizableThreadFactory("lock-leadership-"));
+	private ExecutorService executorService;
 
 	/**
 	 * A lock registry. The locks it manages should be global (whatever that means for the
@@ -169,6 +176,14 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 	}
 
 	@Override
+	public void afterPropertiesSet() throws Exception {
+		if (this.executorService == null) {
+			this.executorService = Executors.newSingleThreadExecutor(new CustomizableThreadFactory("lock-leadership-"));
+			this.shutdownExecutorService = true;
+		}
+	}
+
+	@Override
 	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
 		this.applicationEventPublisher = applicationEventPublisher;
 	}
@@ -183,10 +198,20 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 
 	/**
 	 * Sets the {@link LeaderEventPublisher}.
+	 *
 	 * @param leaderEventPublisher the event publisher
 	 */
 	public void setLeaderEventPublisher(LeaderEventPublisher leaderEventPublisher) {
 		this.leaderEventPublisher = leaderEventPublisher;
+	}
+
+	/**
+	 * Sets the {@link ExecutorService}
+	 *
+	 * @param executorService the executor service
+	 */
+	public void setExecutorService(ExecutorService executorService) {
+		this.executorService = executorService;
 	}
 
 	/**
@@ -274,7 +299,9 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 	@Override
 	public void destroy() throws Exception {
 		stop();
-		this.executorService.shutdown();
+		if (this.shutdownExecutorService) {
+			this.executorService.shutdown();
+		}
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -88,7 +88,7 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 	private final Candidate candidate;
 
 	/**
-	 * flag to denote whether the {@link ExecutorService} was provided via the setter and
+	 * Flag to denote whether the {@link ExecutorService} was provided via the setter and
 	 * thus should not be shutdown when {@link #destroy()} is called
 	 */
 	private boolean executorServiceExplicitlySet;
@@ -175,7 +175,8 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 	}
 
 	/**
-	 * Sets the {@link ExecutorService}
+	 * Sets the {@link ExecutorService}, where is not provided then a default of
+	 * single thread Executor will be used
 	 * @param executorService the executor service
 	 * @since 5.0.2
 	 */
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -57,7 +57,6 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Vedran Pavic
  * @author Glenn Renfro
- * @author Kiel Boatman
  *
  * @since 4.3.1
  */
@@ -76,7 +75,8 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 	/**
 	 * Executor service for running leadership daemon.
 	 */
-	private final ExecutorService executorService;
+	private final ExecutorService executorService =
+			Executors.newSingleThreadExecutor(new CustomizableThreadFactory("lock-leadership-"));
 
 	/**
 	 * A lock registry. The locks it manages should be global (whatever that means for the
@@ -162,33 +162,10 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 	 * @param candidate leadership election candidate
 	 */
 	public LockRegistryLeaderInitiator(LockRegistry locks, Candidate candidate) {
-		this(locks, candidate, Executors.newSingleThreadExecutor(new CustomizableThreadFactory("lock-leadership-")));
-	}
-
-	/**
-	 * Create a new leader initiator. The candidate implementation is provided by the user
-	 * to listen for leadership events and carry out business actions.
-	 * @param locks lock registry
-	 * @param executorService for running leadership election daemon
-	 */
-	public LockRegistryLeaderInitiator(LockRegistry locks, ExecutorService executorService) {
-		this(locks, new DefaultCandidate(), executorService);
-	}
-
-	/**
-	 * Create a new leader initiator. The candidate implementation is provided by the user
-	 * to listen for leadership events and carry out business actions.
-	 * @param locks lock registry
-	 * @param candidate leadership election candidate
-	 * @param executorService for running leadership election daemon
-	 */
-	public LockRegistryLeaderInitiator(LockRegistry locks, Candidate candidate, ExecutorService executorService) {
 		Assert.notNull(locks, "'locks' must not be null");
 		Assert.notNull(candidate, "'candidate' must not be null");
-		Assert.notNull(executorService, "'executorService' must not be null");
 		this.locks = locks;
 		this.candidate = candidate;
-		this.executorService = executorService;
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -57,6 +57,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Vedran Pavic
  * @author Glenn Renfro
+ * @author Kiel Boatman
  *
  * @since 4.3.1
  */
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -162,10 +162,33 @@ public LockRegistryLeaderInitiator(LockRegistry locks) {
...
+	}
+
+	/**
+	 * Create a new leader initiator. The candidate implementation is provided by the user
+	 * to listen for leadership events and carry out business actions.
+	 * @param locks lock registry
+	 * @param candidate leadership election candidate
+	 * @param executorService for running leadership election daemon
+	 */
+	public LockRegistryLeaderInitiator(LockRegistry locks, Candidate candidate, ExecutorService executorService) {
```
artembilan: I would say this option is not so important.
Let's move it just into a setter!

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
```diff
@@ -76,17 +78,40 @@ public void init() {
...
+	public void startAndStopWithRegistryCandidateAndExecutorService() throws Exception {
+		startStopAndVerifyInitiatorLeadership(
+				new LockRegistryLeaderInitiator(this.registry,
+						new DefaultCandidate(),
+						Executors.newSingleThreadExecutor(new CustomizableThreadFactory("lock-leadership-")))
+		);
+	}
+
+	@Test
+	public void startAndStopWithRegistryAndExecutorService() throws Exception {
```
artembilan: I think your new tests don't bring any value.

Would be better to add setter, set an external executor, call `destroy()` on the `LockRegistryLeaderInitiator` and in the end ensure that exactly this external executor is used. Plus its `shutdown()` hasn't been called.

TO repeat a test-case with the same logic doesn't expose any vulnerability.

artembilan (changes requested):
Thank you for contribution, but, please, take a look into my comments.

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
boaty82: OK, was just trying to keep the changes to a minimum.

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
boaty82: Fair points, will make the changes and was just trying to cover my changes.  Certainly the destroy() and shutdown() verifies make sense.  will make changes and update PR

artembilan: Good. Thank you very much for the update! I will come to you for review tomorrow. 

For this particular case to check an internal executor you can consider to use `TestUtils.getPropertyValue()` where we use reflection on the matter

boaty82:
Ok awesome that should work then. I’ll finish off the changes and push tonight/tomorrow

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
boaty82: Ok awesome. I’ll add the test case and then push, most likely tomorrow

boaty82:
OK, well the first failure was my own fault I did add the change to my staged changes before commit and pushing.

However, the second failure doesn't appear to be anything I've done.  The test failing in CI is not failing for me locally.  Have ran the test suite 3 times and the only issue with a test is

`java.lang.AssertionError: 
Expected: an instance of java.net.UnknownHostException
     but: <java.net.ConnectException: Operation timed out> is a java.net.ConnectException
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.springframework.integration.ip.tcp.connection.ConnectionEventTests.testFailConnect(ConnectionEventTests.java:339)
`

but this failed even before I made any changes after forking

artembilan:
I think the problem with the `PollingConsumerEndpointTests` that it uses shared mock object in between tests and relies on the `Mockito.reset()`. In other words it doesn’t follow Mockito recommendations. Different story, though, but you feel free to come back to us with the contribution on the matter as well!

boaty82:
OK sure, I can read up on that on the train tomorrow morning (9:15PM now, so going to log off).  But what to do with the current situation?  Can you trigger the build again?

artembilan:
The `ConnectionEventTests.testFailConnect()` should check for both exceptions, because we don’t have guarantee which one is thrown on which OS and on what Java version. Seems for me we had similar problem in other tests before. Or we can just use there a mock for connection factory. Different story as well. Not a part of this OR.

artembilan:
Hm. I see there is a new problem:
```
org.springframework.integration.redis.leader.RedisLockRegistryLeaderInitiatorTests > testDistributedLeaderElection FAILED
    java.lang.NullPointerException at RedisLockRegistryLeaderInitiatorTests.java:62
```
That’s might be related to your changes...
If not, good to have any way. There is always a room for improvements!

boaty82:
Yes that indeed seems to be related to my change, apologies. I’ll clean, rerun and manually check the entire suite and go from there

boaty82:
My local test run didn't pick up the Redis class due to a ''fail fast' approach for the build and the before mentioned issue with `ConnectionEventTests.testFailConnect` - however, I should've just simply traced usages of the constructor within the project and gone from there.

Now, looking at `PollingConsumerEndpointTests.heavierLoadTest()`

Firstly, I'm not quite sure what this adds over the existing tests - other than just running the test `multipleMessages` many, many times

Anyway, I believe the solution is quite simple.  In the `init` method, before constructing the endpoint, create a new Mock instance of PollableChannel and then remove the line at end of init
`Mockito.reset(channelMock)`

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -168,6 +175,14 @@ public LockRegistryLeaderInitiator(LockRegistry locks, Candidate candidate) {
 		this.candidate = candidate;
 	}
 
+	@Override
+	public void afterPropertiesSet() throws Exception {
```
artembilan: We don't need this method here. We can just simply have that 
```
private ExecutorService executorService =
			Executors.newSingleThreadExecutor(new CustomizableThreadFactory("lock-leadership-"));
```
 as before.

Indeed we need some `boolean` property to control external executor. But in other places we call it like `executorServiceExplicitlySet` - much cleaner and consistent.
You set it to `true` in the setter

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -183,12 +198,22 @@ public void setBusyWaitMillis(long busyWaitMillis) {
...
 	 * @param leaderEventPublisher the event publisher
 	 */
 	public void setLeaderEventPublisher(LeaderEventPublisher leaderEventPublisher) {
 		this.leaderEventPublisher = leaderEventPublisher;
 	}
 
+	/**
+	 * Sets the {@link ExecutorService}
+	 *
+	 * @param executorService the executor service
```
artembilan: The method JavaDocs should be like: 
- Use an imperative style (i.e. Return and not Returns) for the first sentence.
- No blank lines
- Must be `@since` when it is new public API.

https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/leader/JdbcLockRegistryLeaderInitiatorTests.java
```diff
@@ -74,6 +74,7 @@ public void testDistributedLeaderElection() throws Exception {
 			LockRegistryLeaderInitiator initiator = new LockRegistryLeaderInitiator(
 					new JdbcLockRegistry(lockRepository),
 					new DefaultCandidate("foo", "bar"));
+			initiator.afterPropertiesSet();
```
artembilan: When you reinstate the ctor behavior in the `LockRegistryLeaderInitiator`, all these changes in these tests won't be necessary.

artembilan (changes requested):
Here is a review for your changes.

artembilan:
If you would like to fix those non-related tests, please, consider to do that in the separate PRs.
If you don't know how to proceed with them, please, raise issues on the matter and will see how to be.

The main concern here that they are not related to the subject of the PR and therefore such a discussion here is confusing a bit.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
boaty82: OK, I'll go back to that.  I just personally don't like setter doing anything other than setting the target variable.  Also I thought `afterPropertiesSet()` would given more chance for extension and would also not create an `ExecutorService` & `ThreadFactory` when not needed.

boaty82:
Sure, I will create another PR to fix the other test issue

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -92,6 +87,17 @@
 	 */
 	private final Candidate candidate;
 
+	/**
+	 * flag to denote whether the {@link ExecutorService} was provided via the setter and
```
artembilan: I would say JavaDocs on the properties don't make sense, more over if they don't start from the capital letter 😄

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -168,6 +174,16 @@ public LockRegistryLeaderInitiator(LockRegistry locks, Candidate candidate) {
 		this.candidate = candidate;
 	}
 
+	/**
+	 * Sets the {@link ExecutorService}
```
artembilan: I told already that method JavaDocs should go in the imperative style: https://github.com/spring-projects/spring-integration/pull/2357#discussion_r167612477.

Also would be great to mention what is default.

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
```diff
@@ -284,17 +290,41 @@ public void testExceptionFromLock() throws Exception {
...
 
-		initiator.stop();
+		another.stop();
+	}
+
+	@Test
+	public void shouldShutdownInternalExecutorService() throws Exception {
+
+		this.initiator.start();
+		this.initiator.destroy();
```
artembilan: I think there is no guarantee in the method execution order in one test class.
So, this `destroy()` on the shared variable is going to be dangerous when this test method will be called before any other.

Does it make sense to you?

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
```diff
@@ -284,17 +290,41 @@ public void testExceptionFromLock() throws Exception {
...
+
+		ExecutorService executorService = (ExecutorService) TestUtils.getPropertyValue(this.initiator, "executorService");
```
artembilan: There is an overloaded `getPropertyValue()` with an explicit class and generic argument to return.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
boaty82: I was following other setter docs in the class

```
/**
 * Sets the {@link LeaderEventPublisher}.
 * @param leaderEventPublisher the event publisher
 */
```

I will add to description
`where is not provided then a default of single thread Executor will be used`

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
boaty82: JUnit instantiates an instance of the test class, per test method, and as none of the variables in the test are static there are no variables shared between tests

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
boaty82: Given that the variable was renamed to `executorServiceExplicitlySet`, based on your feedback, it doesn't really convey any meaning as to its use, hence why I previously called it `shutdownExecutorService` 

Without a more meaningful name I felt that putting a JavaDoc on it to explain its use was wise, even though the code should explain what is doing rather than any commentary

I have capitalised the first letter of the docs, if you don't want any docs on this property then I can remove - if you feel the comment is not appropriate then please provide an example of what you feel would be more appropriate

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -168,6 +174,17 @@ public LockRegistryLeaderInitiator(LockRegistry locks, Candidate candidate) {
 		this.candidate = candidate;
 	}
 
+	/**
+	 * Sets the {@link ExecutorService}, where is not provided then a default of
```
artembilan: You don't listen to me: must be imperative. Therefore `Set`, not `Seets`

> I was following other setter docs in the class

Then the other one is wrong and must be fixed as well.

Thanks

spring-integration-core/src/test/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiatorTests.java
artembilan: OK. I see what you mean. Good to know 😄

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -92,6 +87,17 @@
 	 */
 	private final Candidate candidate;
 
+	/**
+	 * Flag to denote whether the {@link ExecutorService} was provided via the setter and
+	 * thus should not be shutdown when {@link #destroy()} is called
+	 */
+	private boolean executorServiceExplicitlySet;
```
artembilan: > Given that the variable was renamed to `executorServiceExplicitlySet`, based on your feedback, it doesn't really convey any meaning as to its use, hence why I previously called it `shutdownExecutorService`.

The `shutdownExecutorService` is more misleading for me. That may mean that there is a special configurator for this property and it really isn't related to the `executorService`.

The `executorServiceExplicitlySet` explicitly tells us that it related to the action around the `executorService` property and cannot be configured separately.

That is really self-explanatory for me. Not sure what is your confuse. But I'm fine to leave the JavaDoc here.
Only the concern is to add a period in the end of the sentence.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
boaty82: OK I'll update both my doc and the existing one.

I was confused by you saying mine didn't follow convention when then existing didn't either but was already in the source.

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
boaty82: >The executorServiceExplicitlySet explicitly tells us that it related to the action around the executorService property and cannot be configured separately.

OK I take your point, but again I am used to self documenting code hence I went for a descriptive name.  

Not a problem, every day is a school day! 😄

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
artembilan: Yeah... Sorry for confuse.
What I can tell about that code that it was there before obtaining conventions for the JavaDocs.
Or we have just missed it during review 😄 

Good to have you with a perceptive eye to spot and point for that issue 😉 !

artembilan (approved):
LGTM. Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/78b80ef79fed817143833a5f0d08b2753966a300.

Looking forward for more!

PR #2361, opened 2018-02-13
GH-2358: Register IntGraphController for WebFlux

### LINKED ISSUE #2358: check for both Spring MVC and for Spring WebFlux when deciding to register the Integration Graph controller
`IntegrationGraphControllerRegistrarImportSelector` checks for the `DispatcherServlet`, which fails in a Spring WebFlux application. I copied and pasted the relevant code to a separate class hiearchy, commented out the test for Spring MVC and then ran it in a Spring WebFlux app, and it seems to work.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2358

The WebFlux on the server side can deal with MVC `@Controllers` as well
without any Servlet environment.

* Enable `IntegrationGraphController` when we not only in the
Servlet environment, but also in the WebFlux
* Upgrade to SF-5.0.4 and fix WebFlux tests to meet the latest requirements
* Move `WEB_FLUX_PRESENT` to the `HttpContextUtils`;
deprecate it in the `WebFluxContextUtils`
* Mention WebFlux ability for the `IntegrationGraphController` in the
`graph.adoc`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -138,7 +138,7 @@ subprojects { subproject ->
 		springSecurityVersion = '5.0.1.RELEASE'
 		springSocialTwitterVersion = '1.1.2.RELEASE'
 		springRetryVersion = '1.2.2.RELEASE'
-		springVersion = project.hasProperty('springVersion') ? project.springVersion : '5.0.3.RELEASE'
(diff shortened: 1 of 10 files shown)


garyrussell:
Merged as 11240be55583ae7f914422b4ad53896def37494c with minor javadoc polishing.

PR #2363, opened 2018-02-15
INT-4402: Apply global interceptors at runtime

artembilan:
JIRA: https://jira.spring.io/browse/INT-4402

Make `GlobalChannelInterceptorProcessor` as a `BeanPostProcessor`,
so it can apply global `ChannelInterceptor`s to any initialized channel
beans, even those created at runtime, like in case of dynamic flows with
Java DSL

**Cherry-pick to 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.ListableBeanFactory;
 import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.core.OrderComparator;
 import org.springframework.integration.channel.ChannelInterceptorAware;
 import org.springframework.integration.channel.interceptor.GlobalChannelInterceptorWrapper;
@@ -54,7 +55,8 @@ import org.springframework.util.StringUtils;
  *
  * @since 2.0
  */
-final class GlobalChannelInterceptorProcessor implements BeanFactoryAware, SmartInitializingSingleton {
+final class GlobalChannelInterceptorProcessor
+		implements BeanFactoryAware, SmartInitializingSingleton, BeanPostProcessor {
 
 	private static final Log logger = LogFactory.getLog(GlobalChannelInterceptorProcessor.class);
 
@@ -66,6 +68,8 @@ final class GlobalChannelInterceptorProcessor implements BeanFactoryAware, Smart
 	private final Set<GlobalChannelInterceptorWrapper> negativeOrderInterceptors = new LinkedHashSet<>();
 
 	private ListableBeanFactory beanFactory;
+
+	private volatile boolean singletonsInstantiated;
 
 	@Override
 	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
@@ -96,6 +100,21 @@ final class GlobalChannelInterceptorProcessor implements BeanFactoryAware, Smart
 				addMatchingInterceptors(entry.getValue(), entry.getKey());
 			}
 		}
+
+		this.singletonsInstantiated = true;
+	}
+
+	@Override
+	public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
+		return bean;
+	}
+
+	@Override
+	public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+		if (this.singletonsInstantiated && bean instanceof ChannelInterceptorAware) {
+			addMatchingInterceptors((ChannelInterceptorAware) bean, beanName);
+		}
+		return bean;
 	}
 
 	/**
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
@@ -23,6 +23,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Properties;
 import java.util.Set;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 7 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
```diff
@@ -96,6 +100,21 @@ public void afterSingletonsInstantiated() {
...
+
+	@Override
+	public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
+		return bean;
+	}
+
+	@Override
+	public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+		if (this.singletonsInstantiated && bean instanceof ChannelInterceptorAware) {
+			addMatchingInterceptors((ChannelInterceptorAware) bean, beanName);
```
garyrussell: I think we need to make this optional somehow, default false - since it's a behavior change - some people might get channels intercepted that they didn't previously - they might be already manually adding their interceptor to channels they register manually.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
```diff
@@ -96,6 +100,21 @@ public void afterSingletonsInstantiated() {
...
 			}
 		}
+
+		this.singletonsInstantiated = true;
+	}
+
+	@Override
+	public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
+		return bean;
+	}
```
olegz: You don't need it here, only for backport since it uses older spring

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
olegz: Also, we need to track what channels have been post processed otherwise you'll keep on adding interceptors with each invocation of this operation

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
artembilan: That's not related if we are going to back-port the fix.

It is much easier to merge and just cherry-pick without any modifications.

If there is really an interest to chase only Java 8 style, there is always a room for extra subsequent polishing, but only for the `master`.

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
artembilan: Not, sure, Oleg, what you mean since this is how `BPP` works.

With the `this.singletonsInstantiated` we are going to apply these global interceptors only to the channels created dynamically, at runtime.

garyrussell: Why? This will only be called once per bean.

artembilan: OK, Gary. Then plan is like this:

1. Global Integration property - `false` by default.
2. Remove this property in `5.1` and just rely on the standard `BPP` behavior here.
3. Make this class `public` for Oleg's solution in the https://github.com/spring-cloud/spring-cloud-stream/pull/1227.
4. Leave Oleg's fix in the SCSt since turning on the mentioned property there is going to be a breaking change as well and we are still requested to back-port the fix.

Or another way:

Push this issue back to `5.1` as a normal `BPP` solution without any flags. As we all know there is a workaround - adding these global interceptors manually. In this case we may leave for the fix in the SCSt for a while.

(I might missed something...)

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
olegz: IMHO the interest of keeping code clean (using the tools that are currently available to you) should supersede the desire for clean cherry-pick, so it really isn't about java 8

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
olegz: @artembilan in your test call this twice 
```
beanFactory.initializeBean(testChannel, "testChannel");
```
and you'll see what I mean.

garyrussell: >Make this class public

Why? SCSt can simply set the property to true and it will all work.

>call this twice

That's not really a valid thing to do; I don't think we have to protect against foolishness.

garyrussell: @artembilan Oh; I see what you mean - if SCSt sets the flag to true their own channels might be double intercepted - OK.

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
artembilan: Gary, I see you added +1 to Oleg's comment. So, is it a new codding style I'm not aware?
We really got used to keep mergeable commits before...
Or am I missing here anything again?
Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
artembilan: Correct. That's why I say that the flag to `true` in SCSt is a breaking change as well.
So, if we are not going this "let it break anyway!" way, the only solution as I've chosen in the beginning - move the fix to `5.1` and leave with the workaround in SCSt which is clearly applied only to the internally created channels by the `BinderAwareRouterBeanPostProcessor`.

garyrussell: No; I am concurring with your 1,2,3,4 plan above.

artembilan: Good. I'll also add a note to the `Global Channel Interceptor Configuration` section to describe how it works now and what option should be turned on otherwise.

garyrussell (approved):

spring-integration-core/src/main/java/org/springframework/integration/config/GlobalChannelInterceptorProcessor.java
garyrussell: @artembilan I was just indicating that I will remove this on master after cherry-picking to 4.3.x; that's all.

It doesn't really matter to me which way we do it, but it's probably easiest to use Java 8 on master and fix during backport (like we do with lambdas).

garyrussell:
Merged as 800184270a575479c8f685ad34fd6f02d19933e2 cherry-picked as c7c6264642129993e649d220741de0f5a4f5e7ef

Issue #2364, opened 2018-02-18
Fix intermittent failure of test


boaty82:
Whilst working on another PR I had issue with the test

org.springframework.integration.endpoint.PollingConsumerEndpointTests#heavierLoadTest

Due to use of Mockito.reset in the init method.

The docs advise to not use this method.  I will create a PR to correct this issue

PR #2365, opened 2018-02-18
INT-2364 fix intermittent failing test

boaty82:
Test updates to fix issue https://github.com/spring-projects/spring-integration/issues/2364

JUnit instantiates a new instance of the test class for each test method, so can simply create a new mock for each test rather than calling Mockito.reset


garyrussell:
Copyright needs to be bumped to 2018; will fix on merge.

garyrussell:
Merged as 4bb4a0187a8b92c32a38e3a75c5d7272c60d7807

PR #2373, opened 2018-02-21
Add Pausable

garyrussell:
See https://github.com/spring-projects/spring-integration-kafka/issues/198

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/Pausable.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/Pausable.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/Pausable.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author Gary Russell
+ * @since 5.0.3
+ *
+ */
+public interface Pausable extends Lifecycle {
+
+	/**
+	 * Pause the endpoint.
+	 */
+	void pause();
```
artembilan: The contract of the Kafka `Consumer` is like:
```
public void pause(Collection<TopicPartition> partitions);

public void resume(Collection<TopicPartition> partitions);
```

Maybe we can make this as `Supplier<T>` as well?
Or at least as an overloaded variants:
```
default void pause(Supplier<T> pauseOn) {
    pause();
}
```
?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/endpoint/Pausable.java
garyrussell: I don't think we need to support partial pause; this is what I have...

```java
	@Override
	public void onApplicationEvent(ListenerContainerIdleEvent event) {
		Consumer<?, ?> consumer = event.getConsumer();
		pauseIfNecessary(consumer);
		if (this.resumePending.getAndSet(false) && this.paused.getAndSet(false)) {
			consumer.resume(consumer.paused());
		}
	}

	private void pauseIfNecessary(Consumer<?, ?> consumer) {
		if (this.pausePending.getAndSet(false) && !this.paused.getAndSet(true)) {
			consumer.pause(consumer.assignment());
		}
	}
```

artembilan: OK. Merging...

PR #2374, opened 2018-02-21
INT-4411: DSL: Fix sub-flows for dynamic routers

artembilan:
JIRA: https://jira.spring.io/browse/INT-4411

The `RouterSpec.RouterMappingProvider` relies on the
`ContextRefreshedEvent` which happens only during application start up.
When we register `IntegrationFlow` at runtime, this event doesn't
happen and therefore sub-flow mappings don't populated.

* Fix `RouterSpec.RouterMappingProvider` to parse sub-flow mappings in
the `onInit()`
* Reorder components registration for the router in the
`IntegrationFlowDefinition` to let lately sub-flows to start earlier,
then lifecycles in the main flow

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
```diff
@@ -197,31 +191,30 @@ void addMapping(Object key, NamedComponent channel) {
 		}
 
 		@Override
-		public void onApplicationEvent(ContextRefreshedEvent event) {
-			if (event.getApplicationContext() == getApplicationContext() && !this.initialized.getAndSet(true)) {
-				ConversionService conversionService = getConversionService();
-				if (conversionService == null) {
-					conversionService = DefaultConversionService.getSharedInstance();
+		protected void onInit() throws Exception {
```
garyrussell: Just curious - you must have had a reason to defer this until the context refreshed.

If so, what has changed?

artembilan: As I said in the second part of the change - proper order for the comments registration.
And I guess previously that might not be a `LinkedHashMap` as well.

Now the `IntegrationFlow` bean for the sub-flow is processed before this `RouterMappingProvider`, therefore all the implicit channel beans got their bean names already before this `onInit()` is called.

PR #2377, opened 2018-02-23
MMIH: Convert from JSON only if SpELInvoker

artembilan:
If there is `json`-aware `contentType` header in the message to process
and payload type is `String` or `byte[]` and it isn't equal to the
expected exclusive method argument type (even generic from the `Message<>`),
the conversion should happen only if SpelInvoker is no method.

The regular `InvocableHandlerMethod` takes care about payload conversion
via configured `MessageConverter` in the `PayloadArgumentResolver`

* Fix `MessagingMethodInvokerHelper` to call JSON conversion only when
`this.handlerMethod.spelOnly`

NOTE: We can't do generics conversion right now because it is going
to be a breaking change around
`org.springframework.integration.support.json.JsonObjectMapper` property
in the `MessagingMethodInvokerHelper`.
We can do that only in `5.1` and rely there only on the functionality
from the Jackson Object mapper

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
@@ -308,48 +308,8 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 	}
 
 	public T process(Message<?> message) throws Exception {
-		Message<?> messageToProcess = possiblyConvert(message);
-		ParametersWrapper parameters = new ParametersWrapper(messageToProcess);
+		ParametersWrapper parameters = new ParametersWrapper(message);
 		return processInternal(parameters);
-	}
-
-	/*
-	 * If there's a single method, it is SpEL only, the content is JSON,
-	 * the payload is a String or byte[], the parameter doesn't match the payload,
-	 * and there is a Json Object Mapper on the CP, convert.
-	 */
-	private Message<?> possiblyConvert(Message<?> message) {
-		if (this.handlerMethod != null &&
-				this.handlerMethod.exclusiveMethodParameter != null &&
-				this.jsonObjectMapper != null &&
-				this.handlerMethod.spelOnly) {
-
-			Class<?> type = this.handlerMethod.targetParameterType;
-			if ((message.getPayload() instanceof String && !type.equals(String.class)
-					|| message.getPayload() instanceof byte[] && !type.equals(byte[].class))
-					&& contentTypeIsJson(message)) {
-
-				try {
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -314,22 +314,26 @@ public T process(Message<?> message) throws Exception {
...
 	 */
-	private Message<?> possiblyConvert(Message<?> message) throws Exception {
-		if (this.handlerMethod != null && this.handlerMethod.getTargetParameterType() != null &&
-				this.jsonObjectMapper != null) {
-			Class<?> type = this.handlerMethod.getTargetParameterType();
+	private Message<?> possiblyConvert(Message<?> message) {
+		if (this.handlerMethod != null &&
+				this.handlerMethod.exclusiveMethodParameter != null &&
+				this.jsonObjectMapper != null &&
+				this.handlerMethod.spelOnly) {
```
garyrussell: It's not just `spelOnly` it could be `this.useSpelInvoker` too - see `processInternal()`.

Furthermore, we might fall back to SpEL in `invokeHandlerMethod` before `spelOnly` is forced to true.

I think this might need to be moved to `invokeExpression` so the conversion is done, if necessary, whenever SpEL is involved, which would mean making `ParametersWrapper.message` non-`final`.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
artembilan: Yes, sounds better.

Will take a look later.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingMethodInvokerHelper.java
```diff
@@ -1013,15 +1025,17 @@ private static boolean isMethodDefinedOnObjectClass(Method method) {
...
 
 		private volatile boolean spelOnly;
 
 		private volatile UseSpelInvoker useSpelInvoker;
 
 		private volatile boolean initialized;
-
 		// The number of times InvocableHandlerMethod was attempted and failed - enables us to eventually
+
 		// give up trying to call it when it just doesn't seem to be possible.
```
garyrussell: Why this blank line?

garyrussell (approved):
will fix on merge

garyrussell:
Merged as 2c3ee7d68cb288095939cea61089fd4a5c6e37b5

PR #2378, opened 2018-02-27
Fix JDBC tests and some upgrades

artembilan:
https://build.spring.io/browse/INT-FATS5IC-430

Looks like there is a race condition when our polling rate around
data base is too fast and we have access to the DB files even during
application context close.

* Stop polling channel adapter in the tests explicitly after test methods
* Increase some tests performance decreasing timeouts to wait
* Upgrade to Reactor-3.1.5, AssertJ-3.9.1, Derby-10.14.1.0 and some
Gradle plugins


build.gradle
```diff
@@ -126,7 +126,7 @@ subprojects { subproject ->
 		pahoMqttClientVersion = '1.2.0'
 		postgresVersion = '42.0.0'
 		reactorNettyVersion = '0.7.4.RELEASE'
-		reactorVersion = '3.1.4.RELEASE'
```
garyrussell: Shall we wait for netty before merging?

artembilan: Right, I don't have any objects.
This is here just in case you would like to review.

Thanks

artembilan:
We got all the upgrades here in this PR.
So, nothing holds us from release, unless Micrometer support Docs.

Thanks

garyrussell:
Merged as cb0d43db6bd3d2fe6d96b7e36099176d03e151b0

PR #2380, opened 2018-02-28
INT-4420 spring-integration-test-support hamcrest-all dependency

drewtul:
Use hamcrest-core and hamcrest-library in place of hamcrest-all to match behaviour of other spring projects and avoid multiple JARs containing the same classes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -267,13 +267,10 @@ subprojects { subproject ->
 project('spring-integration-test-support') {
 	description = 'Spring Integration Test Support - **No SI Dependencies Allowed**'
 	dependencies {
-		compile ("junit:junit:$junitVersion") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
-		compile "org.hamcrest:hamcrest-all:$hamcrestVersion"
-		compile ("org.mockito:mockito-core:$mockitoVersion") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
+		compile "org.hamcrest:hamcrest-core:$hamcrestVersion"
+		compile "org.hamcrest:hamcrest-library:$hamcrestVersion"
+		compile "junit:junit:$junitVersion"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -267,13 +267,10 @@ subprojects { subproject ->
...
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
-		compile "org.hamcrest:hamcrest-all:$hamcrestVersion"
-		compile ("org.mockito:mockito-core:$mockitoVersion") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
+		compile "org.hamcrest:hamcrest-core:$hamcrestVersion"
+		compile "org.hamcrest:hamcrest-library:$hamcrestVersion"
+		compile "junit:junit:$junitVersion"
+		compile "org.mockito:mockito-core:$mockitoVersion"
```
artembilan: @drewtul ,

Would you mind to share some of those "spring modules" who use this style of dependencies declaration?

That also would be great to have such an info in the commit message. For the history.

Thanks

drewtul: @artembilan I think I was primarily bumping into it from here:
https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-starters/spring-boot-starter-test/pom.xml#L47

And via third party deps of restassured that spring boot included.

I'm really just looking for the most compatible way locally I can import multiple spring project BOM's without ending up with multiple JARs containing the same classes.

I'm not sure if this is the perfect balance either, as there's no one clear right answer for this dependency.

I can work around this locally by effectively dropping the BOM POM's deps into our project and adding exclusions for this.

drewtul: If you're happy with that as a good enough reason I can update the commit message to reference these items.

artembilan: Yeah... Fair enough.

Merging with appropriate commit message polishing...

Thank you very much!

PR #2381, opened 2018-02-28
INT-4418: Micrometer docs and polishing

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4418

Add docs for the rework and optimize `Meter` creation.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
@@ -92,6 +92,10 @@ public abstract class AbstractMessageChannel extends IntegrationObjectSupport
 
 	private MeterRegistry meterRegistry;
 
+	private Timer successTimer;
+
+	private Timer failureTimer;
+
 	public AbstractMessageChannel() {
 		this.interceptors = new ChannelInterceptorList(logger);
 	}
@@ -458,13 +462,7 @@ public abstract class AbstractMessageChannel extends IntegrationObjectSupport
 				}
 				sent = doSend(message, timeout);
 				if (sample != null) {
-					sample.stop(Timer.builder(SEND_TIMER_NAME)
-							.tag("type", "channel")
-							.tag("name", getComponentName() == null ? "unknown" : getComponentName())
-							.tag("result", sent ? "success" : "failure")
-							.tag("exception", "none")
-							.description("Subflow process time")
-							.register(this.meterRegistry));
+					sample.stop(sendTimer(sent));
 				}
 				channelMetrics.afterSend(metrics, sent);
 				metricsProcessed = true;
@@ -490,7 +488,7 @@ public abstract class AbstractMessageChannel extends IntegrationObjectSupport
 							.tag("name", getComponentName() == null ? "unknown" : getComponentName())
 							.tag("result", "failure")
 							.tag("exception", e.getClass().getSimpleName())
-							.description("Subflow process time")
+							.description("Send processing time")
 							.register(this.meterRegistry));
 				}
 				channelMetrics.afterSend(metrics, false);
@@ -503,6 +501,33 @@ public abstract class AbstractMessageChannel extends IntegrationObjectSupport
 			}
 			throw new MessageDeliveryException(message,
 					"failed to send Message to channel '" + this.getComponentName() + "'", e);
+		}
+	}
+
+	private Timer sendTimer(boolean sent) {
+		if (sent) {
+			if (this.successTimer == null) {
+				this.successTimer = Timer.builder(SEND_TIMER_NAME)
+					.tag("type", "channel")
+					.tag("name", getComponentName() == null ? "unknown" : getComponentName())
+					.tag("result", "success")
+					.tag("exception", "none")
+					.description("Send processing time")
+					.register(this.meterRegistry);
+			}
+			return this.successTimer;
+		}
+		else {
+			if (this.failureTimer == null) {
+				this.failureTimer = Timer.builder(SEND_TIMER_NAME)
+					.tag("type", "channel")
+					.tag("name", getComponentName() == null ? "unknown" : getComponentName())
+					.tag("result", "failure")
+					.tag("exception", "none")
+					.description("Send processing time")
+					.register(this.meterRegistry);
+			}
+			return this.failureTimer;
 		}
 	}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractMessageSource.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractMessageSource.java
@@ -60,13 +60,13 @@ public abstract class AbstractMessageSource<T> extends AbstractExpressionEvaluat
 
 	private String managedName;
 
-	private Counter counter;
-
 	private volatile boolean countsEnabled;
 
 	private volatile boolean loggingEnabled = true;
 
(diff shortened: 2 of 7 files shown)


garyrussell:
cc/ @jkschneider

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -506,6 +504,33 @@ public boolean send(Message<?> message, long timeout) {
 		}
 	}
 
+	private Timer sendTimer(boolean sent) {
```
artembilan: Gary, I've taken a courage to refactor it like this:
```
private Timer sendTimer(boolean sent) {
		if (sent) {
			if (this.successTimer == null) {
				this.successTimer = buildSendTimer(true, "none");
			}
			return this.successTimer;
		}
		else {
			if (this.failureTimer == null) {
				this.failureTimer = buildSendTimer(false, "none");
			}
			return this.failureTimer;
		}
	}

	private Timer buildSendTimer(boolean success, String exception) {
		return Timer.builder(SEND_TIMER_NAME)
				.tag("type", "channel")
				.tag("name", getComponentName() == null ? "unknown" : getComponentName())
				.tag("result", success ? "success" : "failure")
				.tag("exception", exception)
				.description("Send processing time")
				.register(this.meterRegistry);
	}
```
So, now we even can use that `buildSendTimer()` for the exception part:
```
sample.stop(buildSendTimer(false, e.getClass().getSimpleName()));
```

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageHandler.java
```diff
@@ -174,7 +170,7 @@ public void handleMessage(Message<?> message) {
 						.tag("name", getComponentName() == null ? "unknown" : getComponentName())
 						.tag("result", "failure")
 						.tag("exception", e.getClass().getSimpleName())
-						.description("Subflow process time")
+						.description("Send processing time")
```
artembilan: It's `AbstractMessageHandler`, so, I guess `Handle processing time` would be better. Or something similar, but definitely not `send`. The same is applied for the `sendTimer()` method name below.

Also if you apply my polishing in the `AbstractMessageChannel`, I think similar refactoring should be done in this class as well.

src/reference/asciidoc/metrics.adoc
```diff
@@ -100,31 +98,62 @@ Starting with _version 5.0.2_, the framework will automatically detect if there
...
+- `tag` : `type:handler`
+- `tag` : `name:<componentName>`
+- `tag` : `result:(success|failure)`
+- `tag` : `exception:(none|exception simple class name)`
+- `description` : `Send processing time`
+
+The `Counter` meters for message sources have the following names/tags:
+
+- `name` : `spring.integration.receive`
+- `tag` : `type:channel`
```
artembilan: The `source` must be here instead.

artembilan (changes requested):
That's all

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageHandler.java
jkschneider: @artembilan Note that the same timer name is used by handlers and channels (which is intentional). Monitoring systems that receive description text will only display one of them and we don't want that to be non-deterministic. We should maintain a unified description text between handlers and channels whatever we do.

artembilan: Not sure why do you think it the same matter.

When we use an `ExecutorChannel` or `FluxChannel` (or even `QueueChannel`), our `send` is performed on one, calling thread. But `handle` will happen definitely in a different thread. That is only a case of `DirectChannel` that `send` and `handle` overlap. 

So, now you are confusing me even more. IMO we indeed need a separate `send`, `handle` and `receive` metrics.
What's wrong with that?

Thanks

garyrussell: In discussion with Jon, given the nature of how the back-ends work, we need a single metric name; we decided on `send` after struggling for some time with `flow`, `subflow`, `process` etc.

artembilan: Just great! No any clues on the matter throughout the PR...

How about to mention this in the Docs as well?
At least make me happy do not ask similar questions in the future.

Thanks and sorry for such a noise during review.

garyrussell: It's not noise; they are valid questions.

garyrussell: We can enhance the docs later; we will work with @jkschneider and clean up cafe sample app to create some sexy screen shots which will explain better why the naming strategy was chosen.

artembilan: OK. So, what we do with this PR then?
Wait for screenshots?
Shell I polish the mentioned code and merge?
Anything else ?

Thanks

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
garyrussell: Cool; thanks.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageHandler.java
garyrussell: Screenshots can follow after 5.0.3; let's go ahead (with your polishing) and get the 5.0.3.RELEASE train rolling 😄

artembilan (approved):
Merging with mentioned polishing...

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractMessageSource.java
```diff
@@ -131,8 +131,8 @@ public void setLoggingEnabled(boolean loggingEnabled) {
 	}
 
 	@Override
+	@Deprecated
```
artembilan: @garyrussell ,

if you are still here, I would like to ask about this deprecation.

Should we do the same for the super method in the `MessageSourceMetrics` at all? And therefore remove it altogether from here since it is a `default` one anyway there ?

garyrussell: Ah; yes, we should; thanks.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -92,6 +92,10 @@
 
 	private MeterRegistry meterRegistry;
 
+	private Timer successTimer;
```
artembilan: I apply the same optimization to the `NullChannel` as well.

garyrussell: thanks

artembilan:
Gary, here is a polishing for your confirmation: https://github.com/artembilan/spring-integration/commit/92616f3d1c924ac6ec3e9ef7ab0710ea074ac35b

PR #2382, opened 2018-03-01
INT-4417: Add @EndpointId

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4417

Provide more flexibility with bean naming when using EIP Annotations with
Java configuration.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/annotation/EndpointId.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/annotation/EndpointId.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * When used alongside an EIP annotation (and no {@code @Bean}) specifies the bean name of
+ * the consumer bean with the handler bean being {@code id.handler} (for a consuming
+ * endpoint) or {@code id.source} for a message source (e.g. inbound channel adapter).
+ * <p>
+ * When there is also a {@code @Bean} annotation, this is the name of the consumer or
+ * source polling bean (the handler or source gets the normal {@code @Bean} name). When
+ * using on a {@code MessageHandler @Bean}, it is recommended to name the bean
+ * {@code foo.handler} when using {@code @EndpointId("foo"}. This will align with
+ * conventions in the framework. Similarly, for a message source, use
+ * {@code @Bean("bar.source"} and {@code @EndpointId("bar")}.
+ *
+ * @author Gary Russell
+ *
+ * @since 5.0.4
+ */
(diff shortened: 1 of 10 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java
```diff
@@ -255,16 +252,24 @@ protected boolean recursiveFindAnnotation(Class<? extends Annotation> annotation
...
+			name = baseName;
+			int count = 1;
+			while (this.beanFactory.containsBean(name)) {
+				name = baseName + "#" + (++count);
+			}
 		}
 		return name;
 	}
 
+	public static String endpointIdValue(Method method) {
```
artembilan: Can we move this utility into the `MessagingAnnotationUtils` ?

src/reference/asciidoc/configuration.adoc
```diff
@@ -612,12 +613,13 @@ This works like there is no Messaging Annotation on the `@Bean` method.
 * The `AbstractEndpoint` bean name is generated with the pattern:
 `[configurationComponentName].[methodName].[decapitalizedAnnotationClassShortName]`.
 For example the endpoint (`SourcePollingChannelAdapter`) for the `consoleSource()` definition above gets a bean name like:
-`myFlowConfiguration.consoleSource.inboundChannelAdapter`.
+`myFlowConfiguration.consoleSource.inboundChannelAdapter`. Also see <<endpoint-bean-names>>.
```
artembilan: New sentence from a new line.

artembilan (changes requested):
Nothing critical though

artembilan:
Ping

spring-integration-core/src/main/java/org/springframework/integration/annotation/EndpointId.java
```diff
@@ -0,0 +1,53 @@
...
+ * conventions in the framework. Similarly, for a message source, use
+ * {@code @Bean("bar.source"} and {@code @EndpointId("bar")}.
+ *
+ * @author Gary Russell
+ *
+ * @since 5.0.4
+ */
+@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
```
artembilan: Not sure that `@Inherited` is OK here. 
If you override method and then register it in the application context with some Messaging Annotation, you definitely have to have its own distinct endpoint id. Therefore `@EndpointId` is required on the overridden method.

WDYT?

artembilan (changes requested):
While it is a good conventional addition I still think there is a use-case when we can have several Messaging Annotations on the same method. Right, it sounds awkward and might not have a logical reason, but when we add this `@EndpointId` to such a definition we end up with the when it will be overridden with the unique bean name based on the index.

Saying that we might need to document the order how those annotations are parsed:
```
this.postProcessors.put(Filter.class, );
this.postProcessors.put(Router.class, );
this.postProcessors.put(Transformer.class, );
this.postProcessors.put(ServiceActivator.class, );
this.postProcessors.put(Splitter.class, );
this.postProcessors.put(Aggregator.class, );
```

Or just disallow to configure `@EndpointId` on several Messaging Annotations.

artembilan:
See `MessagingAnnotationsWithBeanAnnotationTests.ContextConfiguration.skippedMessageHandler()` bean definition for more info.

spring-integration-core/src/main/java/org/springframework/integration/annotation/EndpointId.java
garyrussell: Looks like this is wrong on several of the EIP annotations...
```
 * <p>Note that this meta-annotation type has no effect if the annotated
 * type is used to annotate anything other than a class.  Note also
 * that this meta-annotation only causes annotations to be inherited
 * from superclasses; annotations on implemented interfaces have no
 * effect.
```

artembilan: Hm. So, sounds like fixing that won't cause any breaking changes.
Isn't it ?

artembilan (approved):
OK. That's cool.

Will we remove `@Inherited` from other EIP method-based annotations in `5.1` ?

garyrussell:
Yes; I think we should wait for 5.1.

artembilan:
>  I think we should wait for 5.1.

Good. I'm raising JIRA then...

artembilan:
For better traceability the JIRA about `@Inherited` on method-based annotations: https://jira.spring.io/browse/INT-4424

PR #2383, opened 2018-03-02
Channel Incorrectly started a second timer

garyrussell:


spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -457,9 +457,6 @@ public boolean send(Message<?> message, long timeout) {
 			}
 			if (countsEnabled) {
 				metrics = channelMetrics.beforeSend();
-				if (this.meterRegistry != null) {
```
artembilan: I think that one before `try {` block is wrong. For consistency with the existing behavior.
Otherwise the `sample.stop(sendTimer(sent))`  has to go already *after* all the `interceptors.postSend(message, this, sent);`. For consistency to include the full stack of converting, preSend, doSend and finally postSend.

Am I missing anything ?

Thanks

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
garyrussell: I discussed this with Jon and I agreed that the interception should be part of the timing, but on reflection; I think you are correct - can you fix on merge?

artembilan: OK. Merging with an agreed fix.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/1c4db65b1cbd346972c3d87c596ac8bc1600f287

PR #2384, opened 2018-03-03
INT-4421: Fix failedMessage in some exceptions

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4421

If a component invoked by an SPCA threw a `MessagingException` with no
`failedMessage`, the resulting ErrorMessage payload had no `failedMessage`.

The `UnicastingDispatcher` had a check for this so it wasn't an issue
as long as at least one `DirectChannel` was between the poller and the
component.

Promote the wrapping code to `IntgrationUtils` and invoke it from
places that blindly rethrew `MessagingException`s.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/AbstractDispatcher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/AbstractDispatcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dispatcher/AbstractDispatcher.java
```diff
@@ -123,6 +123,23 @@ protected boolean tryOptimizedDispatch(Message<?> message) {
...
+	 * If the exception is not a {@link MessagingException} or does not have a
+	 * {@link MessagingException#getFailedMessage() failedMessage}, wrap it in a new
+	 * {@link MessagingException} with the message. There is some inconsistency here in
+	 * that {@link MessagingException}s are wrapped in a {@link MessagingException} whereas
+	 * {@link Exception}s are wrapped in {@link MessageDeliveryException}. It is retained
+	 * for backwards compatibility and will be resolved in 5.1.
+	 * It also does not wrap other {@link RuntimeException}s.
+	 * TODO: Remove this in favor of
+	 * {@code #wrapInDeliveryExceptionIfNecessary(Message, Supplier, Exception)} in 5.1.
+	 * @deprecated in favor of
```
artembilan: I think `@deprecated` must be in the end.
See Checkstyle rule:
```
<!--  		<module name="AtclauseOrder"> -->
<!-- 			<property name="target" value="METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/> -->
<!--     		<property name="tagOrder" value="@param, @return, @throws, @since, @deprecated, @see"/> -->
<!-- 		</module> -->
```

Right, it is commented out, but may be we should consider to switch it on already...

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
```diff
@@ -145,7 +145,8 @@ private boolean doDispatch(Message<?> message) {
 				success = true; // we have a winner.
 			}
 			catch (Exception e) {
-				RuntimeException runtimeException = this.wrapExceptionIfNecessary(message, e);
+				@SuppressWarnings("deprecation")
+				RuntimeException runtimeException = wrapExceptionIfNecessary(message, e);
```
artembilan: Why don't use a new utility here?
Is there some breaking changes in the new utility method?
Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
```diff
@@ -17,6 +17,7 @@
 package org.springframework.integration.support.utils;
 
 import java.io.UnsupportedEncodingException;
+import java.util.function.Supplier;
```
artembilan: I think Copyright of this `IntegrationUtils` class has to be updated.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
garyrussell: Yes; the utility always wraps in `MessageDeliveryException` see the comment in `AbstractDispatcher`; need to wait until 5.1 to switch to the utility.

artembilan (approved):

artembilan:
Doh. Merge two commits without rebasing :cry:

PR #2386, opened 2018-03-05
Support JUnit5 in tests

garyrussell:
Framework tests can now use JUnit5.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.3.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.6-bin.zip
```
artembilan: Did you call `warapper` after upgrading?
Usually they make some changes in the command line scripts as well

artembilan (changes requested):

gradle/wrapper/gradle-wrapper.properties
garyrussell: Good catch - I just pushed the same to RabbitMQ

PR #2389, opened 2018-03-12
Update Master to 5.1.0.BUILD-SNAPSHOT

garyrussell:
`JedisConnection.execute()` no longer accepts `null` arguments.

https://github.com/spring-projects/spring-data-redis/commit/8adea79e67b1ab55cfbeb12a8c215e8ed4bf21a2


artembilan:
Cool!

May we go to Spring Security `5.1` and Reactor `3.2` here as well? 

Also since you have just mentioned Redis, how about to use `Lettuce` client instead?
Right, might be a different story and as some discussion with Spring Data team...

garyrussell:
We switched to Lettuce once before [INT-3113](https://jira.spring.io/browse/INT-3394) but switched back [INT-3394](https://jira.spring.io/browse/INT-3394). I don't remember why, and the JIRA/Commit don't help 😦

garyrussell:
Maybe add a JIRA to the backlog if it's now the preferred client?

PR #2390, opened 2018-03-12
INT-4392: Fix tangles

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4392
JIRA: https://jira.spring.io/browse/INT-4391
JIRA: https://jira.spring.io/browse/INT-4393

Resolve class and leaf package tangles in:

- core (except DSL)
- file
- http

Phase I

Phase II

Phase III

Phase IV


spring-integration-core/src/main/java/org/springframework/integration/support/management/MetricsCaptorLoader.java
```diff
@@ -0,0 +1,88 @@
...
+ * Iterates over a map of classes to check for presence (key) and classes to try to load
+ * (value) (if the key is on the classpath). The class to load must be a
+ * {@link MetricsCaptor} and have a static method {@code loadCaptor(ApplicationContext)}.
+ * User code can add to or replace the default {@link MicrometerMetricsCaptor}.
+ *
+ * @author Gary Russell
+ * @since 5.1
+ *
+ */
+public final class MetricsCaptorLoader {
```
artembilan: Why did you decide to introduce this class?

I agree that it would be better do not mix concerns, but now we have more code around `ClassUtils.isPresent()` and it's so complicated.

Thanks

artembilan (changes requested):
Yeah... Only one  concern.
If tangles are fixed (and I believe so), I'll merge after your answer.

spring-integration-core/src/main/java/org/springframework/integration/support/management/MetricsCaptorLoader.java
garyrussell: I just thought that, since I was externalizing it, it could be more configurable, but yes; agree, will simplify.

spring-integration-core/src/main/java/org/springframework/integration/support/management/IntegrationManagementConfigurer.java
```diff
@@ -210,7 +212,20 @@ public void afterSingletonsInstantiated() {
...
-		this.metricsCaptor = MetricsCaptorLoader.loadCaptor(this.applicationContext);
+		ClassLoader classLoader = IntegrationManagementConfigurer.class.getClassLoader();
+		if (ClassUtils.isPresent("io.micrometer.core.instrument.MeterRegistry",
+				classLoader)) {
+			try {
+				Class<?> captor = ClassUtils.forName(
+						"org.springframework.integration.support.management.micrometer.MicrometerMetricsCaptor",
+						classLoader);
+				Method method = captor.getDeclaredMethod("loadCaptor", ApplicationContext.class);
+				this.metricsCaptor = (MetricsCaptor) method.invoke(null, applicationContext);
```
artembilan: Why do we need all these reflection kung-fu?
Seems for me since `MicrometerMetricsCaptor.loadCaptor()` is `static` we are good around proper class loading.

Thank

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/management/IntegrationManagementConfigurer.java
garyrussell: I tried that and, even with a fully qualified class name (no import), I got a CNFE in a test app that didn't have Micrometer on the CP.

garyrussell: Must have been pilot error - works now.

garyrussell: Actually, no; my memory is bad; reflection is needed to "avoid" the package tangle between `management` and `micrometer`.

The alternative would be to move all the classes in `micrometer` up to `management`. 

WDYT?

artembilan: Good point!

Let's add a comment here on these lines to explain the reason for the reflection and that's all!

Please, also fix `this.` prefix for this `applicationContext` property.

artembilan (changes requested):

artembilan (approved):

PR #2394, opened 2018-03-14
INT-4390: Phase I: flow context/registration cycle

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4390

- extract interfaces
- rename implementations to `Standard...`


garyrussell:
`BEFORE ................................. AFTER`

screen shot 2018-03-14 at 4 28 05 pm screen shot 2018-03-14 at 4 26 34 pm

screen shot 2018-03-14 at 4 28 18 pm screen shot 2018-03-14 at 4 26 57 pm

Still one (tough) tangle to resolve:

screen shot 2018-03-14 at 4 29 31 pm
screen shot 2018-03-14 at 4 29 52 pm

artembilan:
Cool!

Will review tomorrow!

Let's resolve that last one in the separate PR!
This one is already big enough 😄

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.springframework.integration.config.dsl;
```
artembilan: I don't think it's too strict to move classes in this package to some other place.

May just fix `@see` JavaDocs in those classes who uses this one?

artembilan (changes requested):
Only one concern.
I wonder why that Structure 101 considers JavaDocs as tangles as well :smile:

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
garyrussell: Sorry; I don't know what you mean; S101 doesn't look at javadocs, just bytecode.

The **package** tangle between `config` and `dsl` is because this class references lots of classes in `dsl` and lots of `...Spec` classes in `dsl` reference classes in `config` (e.g. `CEFB`, `SPCAFB`). So the simplest fix was to move this to the dsl. Happy to do something else if you have a suggestion.

artembilan: Ah! I see now. That's not a simple class-to-class relationship. That's how those in `dsl` are consuming classes in this `config` package. And later opposite the `IntegrationFlowBeanPostProcessor` uses classes from the `dsl` package. 

Good to know. There is always something to study every day :smile: 
Thank you!

So, never mind. I'm fine with your current solution on the matter.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/82f252dd00d2f352340107aab7011e8d1d72b2de with minor polishing: mostly Copyright

PR #2395, opened 2018-03-15
INT-4430: Make Iterator-based FileSplitter close it's underlying Read…

xak2000:
…er when exception happens in downstream flow

JIRA: https://jira.spring.io/browse/INT-4430

When Iterator-based FileSplitter splits the file, and an exception throws in downstream flow (in the same thread), the exception propagates to the caller leaving underlying file reader opened.

This commit changes AbstractMessageSplitter the way, that, when any exception happens, if Iterator implements java.io.Closeable, it's close() method will be called before propagating exception.

Also FileSplitter's underlying iterator implements Closeable now.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2018 the original author or authors.
+ * Copyright 2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,16 +25,7 @@ import java.util.Iterator;
  *
  * @author Ruslan Stelmachenko
  *
- * @since 4.3.15, 5.0.4
+ * @since 4.3.15
  */
 public interface CloseableIterator<E> extends Iterator<E>, Closeable {
-
-	/**
-	 * Closes this iterator and releases any system resources associated
-	 * with it. If the iterator is already closed then invoking this
-	 * method has no effect.
-	 */
-	@Override
-	void close();
-
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.util;
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+/**
+ * A {@link CloseableIterator} is intended to be used when it may hold resources (such as file or socket handles).
+ * This allows implementations to clean up any resources they need to keep open to iterate over elements.
+ *
+ * @author Ruslan Stelmachenko
+ *
+ * @since 4.3.15, 5.0.4
+ */
+public interface CloseableIterator<E> extends Iterator<E>, Closeable {
+
+	/**
+	 * Closes this iterator and releases any system resources associated
+	 * with it. If the iterator is already closed then invoking this
+	 * method has no effect.
+	 */
+	@Override
+	void close();
+
+}
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
@@ -29,7 +29,7 @@ import java.util.function.Function;
  * @author Ruslan Stelmachenko
  * @since 4.1
  */
-public class FunctionIterator<T, V> implements Iterator<V>, Closeable {
+public class FunctionIterator<T, V> implements CloseableIterator<V> {
 
 	private final Iterator<T> iterator;
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2017 the original author or authors.
+ * Copyright 2014-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 
 package org.springframework.integration.util;
 
+import java.io.Closeable;
+import java.io.IOException;
 import java.util.Iterator;
 import java.util.function.Function;
 
@@ -24,9 +26,10 @@ import java.util.function.Function;
  * {@link #iterator} to a new object applying the {@link #function} on {@link #next()}.
  *
  * @author Artem Bilan
+ * @author Ruslan Stelmachenko
  * @since 4.1
  */
-public class FunctionIterator<T, V> implements Iterator<V> {
+public class FunctionIterator<T, V> implements Iterator<V>, Closeable {
 
 	private final Iterator<T> iterator;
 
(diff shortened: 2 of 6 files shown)


xak2000:
The `org.springframework.integration.file.remote.StreamingInboundTests#testLineByLine` test fails on line:

```
// close by list, splitter
verify(new IntegrationMessageHeaderAccessor(receivedStream).getCloseableResource(), times(2)).close();
```

It checks how many times `close()` was called on `closeableResource`. And it was called 3 times now.

I understand that this is because of additional calling of `close()` from Closeable Iterator. But is this bad or should I have just increase expected number of calls in tests?

spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.Closeable;
+import java.util.Iterator;
+
+/**
+ * A {@link CloseableIterator} is intended to be used when it may hold resources (such as file or socket handles).
+ * This allows implementations to clean up any resources they need to keep open to iterate over elements.
+ *
+ * @author Ruslan Stelmachenko
+ *
+ * @since 4.3.15, 5.0.4
```
artembilan: Just only `4,3.15` 

Also we prefer only the current year in the Copyright for new classes

spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface CloseableIterator<E> extends Iterator<E>, Closeable {
+
+	/**
+	 * Closes this iterator and releases any system resources associated
```
artembilan: Must be imperative: `Close` not `Closes` : https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
```diff
@@ -0,0 +1,40 @@
...
+	 * with it. If the iterator is already closed then invoking this
+	 * method has no effect.
```
artembilan: I would say that this second sentence with the `if` is too bold: we can't guarantee the target implementation.
May be just remove it altogether?

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
```diff
@@ -24,9 +26,10 @@
  * {@link #iterator} to a new object applying the {@link #function} on {@link #next()}.
  *
  * @author Artem Bilan
+ * @author Ruslan Stelmachenko
  * @since 4.1
  */
-public class FunctionIterator<T, V> implements Iterator<V> {
+public class FunctionIterator<T, V> implements Iterator<V>, Closeable {
```
artembilan: Why not `CloseableIterator` ?

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
```diff
@@ -263,7 +265,7 @@ public void close() throws IOException {
 			public boolean hasNext() {
 				this.hasNextCalled = true;
 				try {
-					if (this.line == null && !this.done) {
+					if (!this.done && this.line == null) {
```
artembilan: Why did you swap conditions?

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
```diff
@@ -280,8 +282,8 @@ public boolean hasNext() {
 				}
 				catch (IOException e) {
 					try {
-						bufferedReader.close();
 						this.done = true;
+						bufferedReader.close();
```
artembilan: Why did you swap these lines?

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
```diff
@@ -344,6 +346,17 @@ public Object next() {
 						.setHeader(FileHeaders.MARKER, fileMarker.mark.name());
 			}
 
+			@Override
+			public void close() {
+				try {
+					this.done = true;
+					bufferedReader.close();
```
artembilan: Well, I think I agree that it would be better to mark as `done` before closing `reader`.
Otherwise we may end up with the race condition where we have closed `reader`, but still not `done`, therefore we jump to the next `bufferedReader.readLine()` and catch an `IOException`.

Am I right?

artembilan (changes requested):
That's all.
Just a couple concerns requiring explanation.

I really think that we need to adjust `StreamingInboundTests#testLineByLine` to the current state. because we indeed call `close()` on the reader one more time in the end of iterator.

Otherwise LGTM :+1:

spring-integration-core/src/main/java/org/springframework/integration/util/CloseableIterator.java
xak2000: Actually, this sentence is from [Closeable](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html#close--) interface. I think it is more like a contract which all implementors should follow. But I agree that, in this formulation it is too vague. Maybe just remove it is better.

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
xak2000: It is interesting. Actually first I do this, but then refactored, because `close()` method in `CloseableIterator` doesn't declare any exception, and `Closeable` does. If we implement `CloseableIterator` here, we should catch and ignore inner iterator's `IOException` or wrap it with runtime exception before throw inside `FunctionIterator.close()`.

Maybe I should just add `IOException` to `CloseableIterator.close`? What do you think?

The reason I removed this exception from `CloseableIterator.close` declaration is because the analogous interface from spring-data project also doesn't have it. I thought the intention was to make clear that iterator.close() should not throw `IOException` and simplify it's use, but I'm not sure.

What do you think?

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
xak2000: Because one line below is:

```
boolean ready = !this.done && this.line != null;
```

And the code reading is easier when both expressions have conditions in the same order.

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
artembilan: Good catch, indeed!

I think let's stick with the `Closeable` contract and have that `IOException` in our `close()` method as well! Although from here I think we won't need that method declaration in the `CloseableIterator` at all.

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
xak2000: It is more because `bufferedReader.close();` can throw exception itself, and if it happens, then `done` will not be set to `true`. Another option is to do this in finally block.

artembilan: OK. I see. But do you agree with my logic?

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
xak2000: The reason the same as in `close()` method below:
if `bufferedReader.close();` thows any exception, then `this.done = true;` will not run and we end up in undeterministic state.

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
xak2000: Yes, of course, if iterator will be used from multiple threads, this race condition could happen. I'm not sure that it is thread-safe and intended to be used from multiple threads though.

artembilan: It is. That's how we have to be careful and atomic.
I think what we have so far with your fix is good.
Just need to have those agreed polishing and we will merge soon.

xak2000: @artembilan Sorry for the noob question, but should I fix this in the same commit and then force-push, or create another commit and they be squashed later?

artembilan: Right, you just fix commit with an appropriate comment and push.
We are squashing during merge.

BTW here is an article about good commit message: https://chris.beams.io/posts/git-commit/

xak2000:
Adjusted tests and addressed mentioned concerns.

artembilan:
Ruslan, I got notification about your new commits. Only the inconvenience that it is 22:00 already here. I’ll come to work in the morning 😀

artembilan:
Cherry-picked to `5.0.x` and back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/c45f9a6544d718c94b79f3ea0f1914deea6efd59.

Thank you very much, @xak2000 , for contribution.
Looking forward for more!

PR #2396, opened 2018-03-15
Upgrade to Micrometer 1.0.2

jkschneider:
Release notes [here](https://github.com/micrometer-metrics/micrometer/releases/tag/v1.0.2)


artembilan:

Thanks, Jon!

I the future there would be just enough to notify us about your new version.
We will come back to you if something is wrong after upgrade :smile:

PR #2398, opened 2018-03-20
    INT-4432: fix output channel not correctly set bug.

fifman:
JIRA: https://jira.spring.io/browse/INT-4432

    The `MockIntegrationContext.substituteMessageHandlerFor()` method
    cannot get the correct output channel of target message handler
    since the `outputChannel` property is not initialized.

    * Use `getOutputChannel()` method instead of directly accessing
    the `outputChannel` property. The `getOutputChannel()` method
    guarantees the correct value is retrieved.


artembilan:
Closed in favor of #2402

PR #2399, opened 2018-03-20
INT-4433: Optimize @Publisher metadata

artembilan:
JIRA: https://jira.spring.io/browse/INT-4433

The current `MessagePublishingInterceptor` behavior is to parse expressions
on each method invocation what is not so efficient at runtime

* Introduce `default` `Expression`-based method to the `PublisherMetadataSource`
contract and call existing String-based methods for backward compatibility.
* Deprecate String-based `PublisherMetadataSource` methods in favor of newly
introduced `Expression`-based
* Implement new `getExpressionForPayload()` and `getExpressionsForHeaders()`
in all the `PublisherMetadataSource` implementations
* Cache parsed `Expression` s during initialization in the `PublisherMetadataSource`
implementations or do that on demand in the `MethodAnnotationPublisherMetadataSource`
by provided method basis
* Introduce `MethodAnnotationPublisherMetadataSource#metadataCacheLimit` and populate
its value from the `@EnablePublisher` or `<int:annotation-configuration>`
* Implement `entrySet()` and `values()` in the `ExpressionEvalMap`

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aop/MessagePublishingInterceptor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aop/MessagePublishingInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


garyrussell:
LGTM; although I'm not convinced that we need the complexity of the LRU caches. The number of channels/expressions is bounded by whatever `@Publisher` annotations exist and is fixed. 

What am I missing?

artembilan:
Doh. I think I see what you mean. Since all those `@Publisher` methods exist in our application it fully doesn't hurt to cache their expressions permanently. 

Is that what you mean ?

garyrussell:
Yes; exactly.

garyrussell:
It's a pretty big change too, so I am not sure we should cherry-pick.

artembilan:
Let's review it one more time when I remove that LRU logic! :smile:

artembilan:
OK! Now it's just 8 classes to review. Maybe this is sufficient for back-porting? :wink:

garyrussell:
LGTM; still not sure about cherry-picking, though - I don't recall any complaints about performance (but I can be convinced 😄 )

artembilan:
Hm... We did all those breaking changes for the Micrometer support, but you don't want to cherry-pick performance improvement which is fully backward compatible.

I'm fine to cherry-pick it myself though or if you insist let's don't do that at all!

I'm preparing Java DSL fix which is much bigger change than this, but that one we still consider as a regression and it must be back-ported to `5.0.x`, as well as to `1.2.x`.

So, things happen and we can work out a decision for any our solution.

garyrussell:
OK (I said I could be convinced 😄 ) - merged as b1a3ca764c73e5da4bf10eb4b6be1116375ac227 and cherry-picked.

PR #2401, opened 2018-03-20
LockRegistryLeaderInitiator: Add DEBUG for errors

artembilan:
For better traceability for errors during lock acquiring add DEBUG
logging message in the `catch` block before returning back to the main
loop for the next acquiring attempt

/CC @vpavic

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -400,6 +400,9 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 							}
 							return null;
 						}
+						else if (logger.isDebugEnabled()) {
+							logger.debug("Error to acquire the lock. " + (isRunning() ? "Retrying..." : ""), e);
+						}
 					}
 				}
 			}


spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -400,6 +400,9 @@ else if (acquired) {
 							}
 							return null;
 						}
+						else if (logger.isDebugEnabled()) {
+							logger.debug("Error to acquire the lock. " + (isRunning() ? "Retrying..." : ""), e);
```
vpavic: Perhaps the error message could be worded like _Error acquiring the lock_? Maybe it's just me but _Error to acquire the lock_ sounds a bit confusing at first.

vpavic (commented):
Thanks for proposing this change @artembilan and pinging me. This will be very helpful for debugging purposes, I've just left one minor comment concerning the logged message.

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
artembilan: :smile: 
Fine with me. English is far away from my fluent language, so I'll deffer the final wording polishing to @garyrussell 

Glad to hear that it is going to be helpful, @vpavic !

Thank you!

garyrussell:
Merged as 12301f7fae4e3715d486d142329dc4b56893be98 with suggested change (and cherry-picked).

PR #2402, opened 2018-03-21
INT-4432: fix output channel not set bug.

fifman:
JIRA: https://jira.spring.io/browse/INT-4432

The `MockIntegrationContext.substituteMessageHandlerFor()` method
cannot get the correct output channel of target message handler
since the `outputChannel` property is not initialized.

* Use `getOutputChannel()` method instead of directly accessing
the `outputChannel` property. The `getOutputChannel()` method
guarantees the correct value is retrieved.


artembilan (approved):
LGTM

Merging with minor code style polishing and author name.

:+1:

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/ed00f62549f3ee8ab5585abfc20a2bf4cb5feac4 and cherry-picked to `5.0.x`.

Looking forward for more!

PR #2403, opened 2018-03-21
INT-4434: Allow to use sub-flows from beans

artembilan:
JIRA: https://jira.spring.io/browse/INT-4434

There were a restriction introduced since Java DSL `1.2` do not use
`IntegrationFlow` beans for sub-flow definitions, e.g. in routers.
It is considered as regression by community because it worked before
in version `1.1`

* Introduce `IntegrationFlow.getInputChannel()` to be able to bridge
from the main flow to the flow which is treated as sub-flow.
In most cases we talk about an independent bean for the `IntegrationFlow`
which can be used as a stand along one and as a sub-flow in other flow

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/EndpointSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/EndpointSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,8 +23,10 @@ import java.util.function.Function;
 import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.context.SmartLifecycle;
 import org.springframework.core.ResolvableType;
+import org.springframework.integration.channel.DirectChannel;
 import org.springframework.integration.endpoint.AbstractPollingEndpoint;
 import org.springframework.integration.scheduling.PollerMetadata;
+import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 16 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/EndpointSpec.java
```diff
@@ -141,4 +143,24 @@ protected void assertHandler() {
 		Assert.state(this.handler != null, "'this.handler' must not be null.");
 	}
 
+	protected MessageChannel obtainInputChannelFromFlow(IntegrationFlow subFlow) {
+		return obtainInputChannelFromFlow(subFlow, true);
+	}
+
+	protected MessageChannel obtainInputChannelFromFlow(IntegrationFlow subFlow, boolean evaluateInternalBuilder) {
```
garyrussell: Why false on routers and true everywhere else (including recipient list router)? Perhaps javadocs on these methods would help.

artembilan: Yeah... That's hard to explain so far. See `route(S routerSpec, Consumer<S> routerConfigurer)`.
I think the main reason to do that to be able to add beans into the `BeanFactory` in the proper order. So, since sub-flows (somehow) in router depend on their root router component, I register it first and only after that I build sub-flows from their builder.

Looks like in other place we don't have such a strong dependency, so it's fine to register sub-flows before their main component.

I thought briefly to change such a logic, but decided to do that in the separate JIRA (if that is possible at all) and already without backporting.

Right,  I'll add JavaDocs to them to explain things a bit.

Thanks

P.S. I won't mind if you (or someone else) will take a look to this subject independently. Another pair of eyes sometimes gives a good unexpected solution.

spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
```diff
@@ -69,12 +70,35 @@
...
+	public MessageChannel getInputChannel() {
+		if (this.inputChannel == null) {
+			this.inputChannel =
+					this.integrationComponents.keySet()
+							.stream()
+							.filter(MessageChannel.class::isInstance)
+							.map(MessageChannel.class::cast)
+							.findFirst()
+							.orElseThrow(() -> new IllegalStateException("The 'IntegrationFlow' [" + this + "] " +
+									"doesn't start with 'MessageChannel' for direct message sending."));
```
garyrussell: Is it really valid to search for the first channel in a flow? Shouldn't we simply reject subflows that don't start with a channel?

garyrussell (commented):
To answer your question, I wonder if it would be simpler to just check if the flow passed into a subflow method already exists as a bean in the application context and, if so, just wire us up to the input channel.

spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
artembilan: I thought about that one more time, just because the previous version really rejected flows from non-channel in the beginning. But right now I think it doesn't matter: we always are good to send a message to the output channel of any message producer in the beginning of the flow. In other words essentially we are selecting here the *first* channel anyway.
See this:
```
IntegrationFlows.from(MessageSource).handle().get();
...
IntegrationFlows.from(MessageProducer).handle().get();
```
So, what we select here is an `inputChannel` of that `handle()`. What is the point to reject it?

That's might not be obvious at a glance, but it becomes reasonable to reuse existing flows even if they are not starting from the channel. We may document it though or just leave as silent feature until someone will ask :popcorn:

garyrussell: OK; makes sense; did you see the comment in my review overview?

artembilan:
>  to just check if the flow passed into a subflow method already exists as a bean in the application context

We can't do that on the DSL level, we don't have any `BeanFactory` context yet there from builder API.

I can make `StandardIntegrationFlow` as `BeanFactoryAware` but that doesn't matter in the end: we still need that `inputChannel` from the flow. Therefore extra `is_BeanFactory_wired_API` doesn't bring too much value.

garyrussell:
Merged as 74504326c98208b46f6dfbdbf693601f1bd4b1bd and cherry-picked as f9da4382c24936348301c7c415b5ff5837d90f09

PR #2404, opened 2018-03-22
INT-4440: Support serialized UUID

nathankurtyka:
Support for serialized UUIDs (INT-4440)

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2017 the original author or authors.
+ * Copyright 2015-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.support;
 
+import java.nio.ByteBuffer;
 import java.util.Map;
 import java.util.UUID;
 
@@ -30,6 +31,7 @@ import org.springframework.messaging.MessageHeaders;
  * @author Stuart Williams
  * @author David Turanski
  * @author Artem Bilan
+ * @author Nathan Kurtyka
  *
  * @since 4.2
  */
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):

nathankurtyka:
Excellent feedback!

All incorporated and pushed.

spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java
```diff
@@ -76,4 +70,20 @@ public Object remove(Object key) {
...
 
+	private static Long extractTimestamp(Map<String, Object> headers) {
+		if (headers != null && headers.containsKey(MessageHeaders.TIMESTAMP)) {
+			Object timestamp = headers.get(MessageHeaders.TIMESTAMP);
+			return (timestamp instanceof String) ? Long.parseLong((String) timestamp) : (Long) timestamp;
+		}
+		return null;
+	}
+
+	private static UUID extractUuid(Map<String, Object> headers) {
```
artembilan: I would call it `extractId()`.

Also, please, swap the methods: this one goes first in the class definition. Just because it is used first

spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java
```diff
@@ -76,4 +70,20 @@ public Object remove(Object key) {
...
+		if (headers != null && headers.containsKey(MessageHeaders.ID)) {
+			Object uuid = headers.get(MessageHeaders.ID);
+			return (uuid instanceof String) ? UUID.fromString((String) uuid) : (UUID) uuid;
```
artembilan: How about `UUID.nameUUIDFromBytes(byte[] name)` as well ?
There is no guarantee that we always serialize only to JSON.

Or am I paranoid? :smile:

artembilan (changes requested):
Only a couple simple concerns.

Also, please, add your name to the `@author` list and look if we need to update Copyright to the current year.

Thank you!

spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java
```diff
@@ -38,13 +40,7 @@
...
 
 	public MutableMessageHeaders(@Nullable Map<String, Object> headers) {
-		this(headers,
-				(headers != null ?
-						(UUID) headers.get(MessageHeaders.ID)
-						: null),
-				(headers != null ?
-						(Long) headers.get(MessageHeaders.TIMESTAMP)
-						: null));
+		super(headers, MutableMessageHeaders.extractId(headers), MutableMessageHeaders.extractTimestamp(headers));
```
artembilan: We don't need class name if `static` method is here as well.

Also your name for the modified test class.

I'll fix on merge.

Thank you!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/ae2aa8b6d10c20b09413776034f718de31a9915c and cherry-picked to `5.0.x`.

@nathankurtyka ,

Looking forward for more!

A couple comments while you are still here:

1. Please, don't squash commits when push changes to the PR. It would be much easier for us to trace the progress. We are squashing during merge anyway.
2. Please, make changes in the top-level branches, not `master`. Right now you have to `reset --hard` your origin because of changes in the upstream.

nathankurtyka:
Thank you for your patience!

Yes, my next fix will be much smoother :)

PR #2406, opened 2018-03-22
Fix AMQP MessageSource Tests

garyrussell:
https://build.spring.io/browse/INT-MASTER-983/

Race between the poller and template to get the requeued message.

Also add trace logging to the acknowledgment.

**cherry-pick to 5.0.x**


artembilan:

PR #2408, opened 2018-03-23
INT-4443: Use SimpleEC for uriVariablesExpression

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4443

__cherry-pick to 5.0.x, 4.3.x__ (what's new will need polishing)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -141,7 +141,7 @@ subprojects { subproject ->
 		springSecurityVersion = '5.1.0.BUILD-SNAPSHOT'
 		springSocialTwitterVersion = '1.1.2.RELEASE'
 		springRetryVersion = '1.2.2.RELEASE'
-		springVersion = project.hasProperty('springVersion') ? project.springVersion : '5.0.4.RELEASE'
+		springVersion = project.hasProperty('springVersion') ? project.springVersion : '5.0.5.BUILD-SNAPSHOT'
 		springWsVersion = '3.0.1.RELEASE'
 		tomcatVersion = "8.5.23"
 		xmlUnitVersion = '1.6'
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package org.springframework.integration.context;
 import java.util.Properties;
 
 import org.springframework.beans.factory.BeanFactory;
+import org.springframework.expression.spel.support.SimpleEvaluationContext;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
 import org.springframework.integration.config.IntegrationConfigUtils;
 import org.springframework.integration.metadata.MetadataStore;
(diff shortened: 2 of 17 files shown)


build.gradle
```diff
@@ -141,7 +141,7 @@ subprojects { subproject ->
 		springSecurityVersion = '5.1.0.BUILD-SNAPSHOT'
 		springSocialTwitterVersion = '1.1.2.RELEASE'
 		springRetryVersion = '1.2.2.RELEASE'
-		springVersion = project.hasProperty('springVersion') ? project.springVersion : '5.0.4.RELEASE'
+		springVersion = project.hasProperty('springVersion') ? project.springVersion : '5.0.5.BUILD-SNAPSHOT'
```
artembilan: What version to use for `4.3.x` back-port ?

Thanks

artembilan (approved):
OK. Only one concern for back-porting to `4.3.x`.
However I agree that `changes-*.adoc` have to be reverted there. Not sure about `What's New` on the matter though...

artembilan (changes requested):
Huh! Travis failed for many tests:
```
org.springframework.integration.http.outbound.HttpRequestExecutingMessageHandlerTests > testInt2455UriNotEncoded FAILED
    java.lang.IllegalArgumentException at HttpRequestExecutingMessageHandlerTests.java:761
```

build.gradle
garyrussell: https://github.com/spring-projects/spring-framework/commit/4bc3e0c2b4c703b3da90774e948f8adb18f334f9

4.3.15.BS

garyrussell:
Strange; built ok for me.

garyrussell:
Ah; new SPR commit: https://github.com/spring-projects/spring-framework/commit/38dcd179d5c2ad7630310cba292e80471b000d5f

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -131,12 +134,23 @@ public static TaskScheduler getRequiredTaskScheduler(BeanFactory beanFactory) {
 
 	/**
 	 * @param beanFactory BeanFactory for lookup, must not be null.
-	 * @return the instance of {@link StandardEvaluationContext} bean whose name is "integrationEvaluationContext" .
+	 * @return the instance of {@link StandardEvaluationContext} bean whose name is
+	 * {@value #INTEGRATION_SIMPLE_EVALUATION_CONTEXT_BEAN_NAME}.
```
artembilan: Must be `INTEGRATION_EVALUATION_CONTEXT_BEAN_NAME`.

Will fix on merge if only concern.

spring-integration-core/src/test/java/org/springframework/integration/configuration/EnableIntegrationTests.java
```diff
@@ -706,7 +707,8 @@ public void testSourcePollingChannelAdapterOutputChannelLateBinding() {
 
 	@Test
 	public void testIntegrationEvaluationContextCustomization() {
-		EvaluationContext evaluationContext = this.context.getBean(EvaluationContext.class);
+		EvaluationContext evaluationContext = this.context
+				.getBean(IntegrationContextUtils.INTEGRATION_EVALUATION_CONTEXT_BEAN_NAME, EvaluationContext.class);
```
artembilan: Looks like the `this.context.getBean(StandardEvaluationContext.class);`
And this way we really ensure that `simple` variant is really not a `StandardEvaluationContext` impl.

Or do you have something else in mind on the matter ?

artembilan (approved):
Only two simple concerns.

I have them fixed (and some Copyrights) in hands: just need your approval and I'll merge this already.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
garyrussell: Oops; good catch.

spring-integration-core/src/test/java/org/springframework/integration/configuration/EnableIntegrationTests.java
garyrussell: No; that's good; I was just fixing the test because there are now 2 ECs.

garyrussell:
Thanks; please go ahead and merge with your polishing.

artembilan:
Back-port to `4.3.x` is blocked by the missed API in the SF-4.3.x.

Will come back to that when upstream is ready.

artembilan:
Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/705d33a30c62d86774d7be9c2c606bdce3da08db

PR #2409, opened 2018-03-26
INT-4438 provide integration test for allEndpointsRunning

m-terra:
Details described here: https://jira.spring.io/browse/INT-4438

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/support/SmartLifecycleRoleControllerTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/support/SmartLifecycleRoleControllerTests.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.support;
 
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
@@ -24,6 +25,7 @@ import org.junit.Test;
 import org.mockito.InOrder;
 
 import org.springframework.context.SmartLifecycle;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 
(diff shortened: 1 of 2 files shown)


artembilan:
@andybaer ,

Thank you for sharing the code!

I'm not sure that we are going to merge it because I see we have a pretty good `EndpointRoleParserTests`.

I'm adjusting that one with the `allEndpointsRunning()` assertion, but meanwhile, please, confirm that this test case fails with the mentioned `java.lang.IllegalStateException: Duplicate key`.

That's still unclear why are you sticking with that `errorLogger`...

andybaer:
Hi Artem
I'm sticking to that {{errorLogger}} because I fetched all {{NamedComponent}} beans from the context and found that the {{errorLogger}}  and it's handler have the same component name which again is used as key in that {{Map}}. I don't understand what is not clear about my statement. Anyway, if you try to make this test work, you will find out.
Regards, Andy

artembilan:
Yep! Confirmed. It fails.

And I know "Why?": https://github.com/spring-projects/spring-integration/pull/2407/files#diff-0a70ed47bf7fc1556df5d76b2bd0ce5cL242

The `AbstractPollingEndpoint` registers itself in the `SmartLifecycleRoleController` twice: in the `afterPropertiesSet()` and then after `stop()/start()`.

Now I have fixed that in my PR. I'll accept  a `assertTrue(this.controller.allEndpointsRunning("cluster"));` addition to the mentioned `EndpointRoleParserTests` to cover this use-case.

I think you still have been misled by that the same component name: the `LoginHandler` can't make it into that map just because it isn't `SmartLifecycle` as it is required by the `SmartLifecycleRoleController`.

More over that `errorLogger` endpoint can't make it into this `SmartLifecycleRoleController` just because it isn't assigned to any role.

You need to filter `SmartLifecycle` from the application context, first of all.
Then filter them for the `role` they are assigned. 
And only after that check their `NamedComponent`  state. 

This way you won't bump to the `errorLogger` at all.

So, why do you still think that `errorLogger` is a culprit?

Does it all make sense to you now ?

artembilan:
Closed in favor of #2407 .

I've just pushed your concern to the `EndpointRoleParserTests`. This one and its config really reflects what you suggest. The real problem that you have an `inbound-channel-adapter` (doesn't matter which one). And it's `AbstractPollingEndpoint` re-registers itself in the `SmartLifecycleRoleController`after `stop()/start()`. 

This is fixed now in that PR.

Again: nothing bad with the `errorLogger` and we are not going to fix its component name. Fully unrelated and isn't involved in the role control process.

Cheers

andybaer:
Ok sorry about that. As long as the issue is fixed i'm happy. It actually prevents us to migrate to spring boot 2.

Thanks, Andy

PR #2411, opened 2018-03-27
[DO NOT MERGE YET] Add support Reactor's `errorStrategyContinue`

artembilan:


artembilan:
Closed as invalid. The Reactor's error strategy is for the intermediate operators, but not for the target `Subscriber` what we have exactly with our `AbstractMessageHandler`.

PR #2412, opened 2018-03-28
Add BoundRabbitChannelAdvice

garyrussell:


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdvice.java
```diff
@@ -0,0 +1,95 @@
...
+			return this.operations.invoke(operations -> {
+				try {
+					Object result = invocation.proceed();
+					if (this.waitForConfirmsTimeout != null) {
+						this.operations.waitForConfirmsOrDie(this.waitForConfirmsTimeout.toMillis());
+					}
+					return result;
+				}
+				catch (Throwable t) { // NOSONAR - rethrown below
+					throw new ThrowableWrapper(t);
```
artembilan: `ReflectionUtils.rethrowRuntimeException()` ?

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdviceTests.java
```diff
@@ -0,0 +1,133 @@
...
+		verify(this.config.channel).confirmSelect();
+		verify(this.config.channel).basicPublish(eq(""), eq("rk"), anyBoolean(), any(), eq("A".getBytes()));
+		verify(this.config.channel).basicPublish(eq(""), eq("rk"), anyBoolean(), any(), eq("B".getBytes()));
+		verify(this.config.channel).basicPublish(eq(""), eq("rk"), anyBoolean(), any(), eq("C".getBytes()));
+		verify(this.config.channel).waitForConfirmsOrDie(10_000L);
+	}
+
+	@Configuration
+	@EnableIntegration
+	@IntegrationComponentScan
```
artembilan: M-m-m. Seems for me the style `IntegrationFlows.from(Gate.class)` doesn't need this annotation.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdviceTests.java
```diff
@@ -0,0 +1,133 @@
...
+	private Config config;
+
+	@Test
+	public void testAdvice() throws Exception {
+		this.gate.send("a,b,c");
+		verify(this.config.connection, times(1)).createChannel();
+		verify(this.config.channel).confirmSelect();
+		verify(this.config.channel).basicPublish(eq(""), eq("rk"), anyBoolean(), any(), eq("A".getBytes()));
+		verify(this.config.channel).basicPublish(eq(""), eq("rk"), anyBoolean(), any(), eq("B".getBytes()));
+		verify(this.config.channel).basicPublish(eq(""), eq("rk"), anyBoolean(), any(), eq("C".getBytes()));
```
artembilan: Well, I guess this test is going to work the same way even without `AmqpDedicatedChannelAdvice`.
I mean the order how you verify messages appearance. Just because all the calls use the same channel and the order is guaranteed by the regular call stack.
But let it be since you verify all other interactions with the channel caused by the mentioned advice.

src/reference/asciidoc/amqp.adoc
```diff
@@ -1504,6 +1504,58 @@ Negated patterns get priority, so a list such as
...
+            .handle(Amqp.outboundAdapter(template).routingKey("rk"))
+            .get();
+}
+----
+
+Let's say we send a message `a,b,c` to the gateway, while it is likely that messages `A`, `B`, `C` will be sent in order, there is no guarantee.
+This is because the template "borrows" a channel from the cache for each send and there is no guarantee that the same channel will be used for each.
+One solution is to start a transaction before the splitter, but transactions are very expensive in RabbitMQ and can reduce performance several hundred fold.
+
+To solve this problem in a more efficient manner, Spring Integration provides the `AmqpDedicatedChannelAdvice`.
```
artembilan: We need to say `since 5.1` and also explain that this is exactly a `HandleMessageAdvice` and point to the `<<handle-message-advice>>` chapter for more background.

artembilan (changes requested):
Cool stuff!

I'd be glad to be involved in the discussions where these ideas get birth. :smile:

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdvice.java
garyrussell: ? 

`proceed()` throws `Throwable`.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdviceTests.java
garyrussell: yup

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdvice.java
```diff
@@ -74,6 +78,14 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
 				catch (Throwable t) { // NOSONAR - rethrown below
 					throw new ThrowableWrapper(t);
 				}
+			}, (tag, multiple) -> {
+				if (logger.isDebugEnabled()) {
+					logger.debug("Publisher confirm ack: " + tag + ", multiple: " + multiple);
+				}
+			}, (tag, multiple) -> {
+				if (logger.isDebugEnabled()) {
+					logger.debug("Publisher confirm ack: " + tag + ", multiple: " + multiple);
```
artembilan: `nack` ?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdvice.java
artembilan: ?
```
public static void rethrowRuntimeException(Throwable ex) {
```
Then in the main `catch` you can use `UndeclaredThrowableException.getUndeclaredThrowable()`

garyrussell:
So I did this as a PoC but I think it's worthwhile; if you agree, I'll create a JIRA for the release notes.

garyrussell:
Can you think of a better name for the advice?

artembilan:
Yeah! Fine with me!
Although I can live without JIRA since we have started to build release notes from the commits history, e.g.: https://github.com/spring-projects/spring-integration/releases/tag/v5.0.3.RELEASE

artembilan:
> Can you think of a better name for the advice?

`RabbitOperationsInvokeAdvice` ? Because there is no guarantee from the advice perspective that end-user is really going to use the same `RabbitTemplate` or won't hand off to different that, but the goal of an advice is still there - call `RabbitOperations.invoke()`

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdviceIntegrationTests.java
```diff
@@ -0,0 +1,109 @@
...
+	@Autowired
+	private Config.Gate gate;
+
+	@Autowired
+	private Config config;
+
+	@Test
+	public void testAdvice() throws Exception {
+		this.gate.send("a,b,c");
+		assertTrue(this.config.latch.await(10, TimeUnit.SECONDS));
```
artembilan: This test doesn't bring too much value and doesn't confirm how the new advice is cool.
I'm not sure what and how we should test here, but the result I'm sure I can get even without this advice.

Maybe really to check an order of messages with a single-threaded consumer?

artembilan (changes requested):

garyrussell:
I went with `RabbitOperationsInvokingAdvice` - the commit headline will need polishing when squashing - shall I squash with the next commit?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/AmqpDedicatedChannelAdvice.java
garyrussell: Oh - I was looking at the wrong `rethrow...` method.

Fixing.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/RabbitOperationsInvokingAdvice.java
```diff
@@ -0,0 +1,107 @@
...
+ * on the same channel, as long as there are no thread handoffs.
+ * The same RabbitOperations must be used in this and all downstream components.
+ * Typically used with a splitter or some other mechanism that would cause
+ * multiple messages to be sent.
+ *
+ * @author Gary Russell
+ * @since 5.1
+ *
+ */
+public class RabbitOperationsInvokingAdvice implements HandleMessageAdvice {
```
artembilan: M-m-m. Maybe just `RabbitOperationsAdvice`? Because that our `OperationsCallback` really assumes to let us to call all the possible `RabbitOperations` during `invoke()`

garyrussell: Hmmm `RabbitOperationsAdvice` implies it advises the operations.

artembilan: Just `RabbitChannelAdvice`? With the meaning that we wrap the method with a dedicated thread-bound Rabbit channel. 
Well, anything works for me since I understand what is this advice about. We just need to come up with some fluent name to let end users to determine the purpose clearly or at least close.

Sorry for being so nit-pick, but for me, as non-English speaker, all those `ing`-ings are hard to translate since there is no such a structure in our language :smile: 

I think we will stop with the next your name. It took already a plenty of time to discuss almost for nothing.

garyrussell: No; I think it's worth spending time to get it right.

`RabbitBoundChannelAdvice`
`RabbitBindChannelAdvice`
`RabbitReserveChannelAdvice`
`BindRabbitChannelAdvice`
`BoundRabbitChannelAdvice`

I think I like the last one - describes what it does - creates a channel that is bound to the thread.

IMHO we have too many "bindings" - SCSt, RabbitMQ bindings, so I think we should avoid Bind, Binding in the name.

artembilan (approved):

garyrussell:
Polish is good; thanks.

PR #2413, opened 2018-03-29
INT-4351: Add DiscardAwareFileListFilter

artembilan:
JIRA: https://jira.spring.io/browse/INT-4351

The `WatchService` reacts to the events in the file system and keep
track ove the events until we poll them, e.g. via `listEligibleFiles()`
in the `WatchServiceDirectoryScanner`.
On the other hand the `SourcePollingChannelAdapter` calls the mentioned
`listEligibleFiles()` according its polling period.
At this moment the `FileListFilter` is applied to the polled files.
It may happen that `LastModifiedFileListFilter` can't accept too young
files yet and they are lost for the future consideration.

* To allow, for example, to retain young files by the
`LastModifiedFileListFilter` judgment for the future cycles add
`DiscardAwareFileListFilter` with the `DiscardCallback` support.
The `WatchServiceDirectoryScanner` now registers such a callback into
the filter and stores discarded files into the `filesToPoll` queue
for the future poll cycle.

**Cherry-pick to 5.0**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,17 +41,20 @@ import org.springframework.util.Assert;
  *
  *
  */
-public class CompositeFileListFilter<F> implements ReversibleFileListFilter<F>, ResettableFileListFilter<F>, Closeable {
+public class CompositeFileListFilter<F>
+		implements ReversibleFileListFilter<F>, ResettableFileListFilter<F>, DiscardAwareFileListFilter<F>, Closeable {
 
 	protected final Set<FileListFilter<F>> fileFilters; // NOSONAR
 
+	private DiscardCallback<F> discardCallback;
+
 
 	public CompositeFileListFilter() {
-		this.fileFilters = new LinkedHashSet<FileListFilter<F>>();
+		this.fileFilters = new LinkedHashSet<>();
 	}
 
 	public CompositeFileListFilter(Collection<? extends FileListFilter<F>> fileFilters) {
-		this.fileFilters = new LinkedHashSet<FileListFilter<F>>(fileFilters);
+		this.fileFilters = new LinkedHashSet<>(fileFilters);
 	}
 
 
@@ -65,7 +68,7 @@ public class CompositeFileListFilter<F> implements ReversibleFileListFilter<F>, 
 	}
 
 	public CompositeFileListFilter<F> addFilter(FileListFilter<F> filter) {
-		return this.addFilters(Collections.singletonList(filter));
+		return addFilters(Collections.singletonList(filter));
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/filters/DiscardAwareFileListFilter.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Artem Bilan
+ *
+ * @since 5.0.4
+ */
+public interface DiscardAwareFileListFilter<F> extends FileListFilter<F> {
+
+	void addDiscardCallback(DiscardCallback<F> discardCallback);
+
+	@FunctionalInterface
+	interface DiscardCallback<F> {
```
garyrussell: Do we really need a new interface? Why not just `Consumer<File>` ?

spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
```diff
@@ -101,6 +107,17 @@ public void close() throws IOException {
...
 
+	@Override
+	public void addDiscardCallback(DiscardCallback<F> discardCallback) {
+		this.discardCallback = discardCallback;
+		if (this.discardCallback != null) {
+			this.fileFilters
+					.stream()
+					.filter(DiscardAwareFileListFilter.class::isInstance)
+					.map(f -> (DiscardAwareFileListFilter<F>) f)
+					.forEach(f -> f.addDiscardCallback(discardCallback));
```
garyrussell: We must only call the callback once per file - remember the composite filter passes all files to all filters.

Or, change the queue to linked set.

garyrussell (changes requested):
That's all for now. Otherwise technique looks good.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
artembilan: Well, as far as `CompositeFileListFilter` calls `filterFiles()` for each its delegate, it really a part of the API that `discardCallback` will be called from them. But pay attention that this `CompositeFileListFilter` doesn't call `discardCallback` - it just populates to the delegates.

But I agree: let's reconsider the `WatchServiceScanner` to use a some collection with uniqueness to avoid a double call for the same file from different filters!

garyrussell:
Merged as 0ef659b8a7c2833640e5f95246225893f03ffc95

yihfenglow:
Would this fix the issue where if you use .preventDuplicate(true) and a LastModifiedFileListFilter it behaves unexpectedly due to https://stackoverflow.com/questions/39604652/acceptoncefilefilter-keeps-other-filters-from-working-in-a-compositefilelistfilt? 

I believes this quirk throws a lot of people off. If this doesn't fix it, it feels like something that can potentially be extended to do so?

Thanks!

artembilan:
I think that question is not related to the problem we have fixed here. I believe what you need is called `ChainFileListFilter` instead of `CompositeFileListFilter`: https://docs.spring.io/spring-integration/docs/5.0.4.RELEASE/reference/html/files.html#file-reading

yihfenglow:
What I meant was that if you use .preventDuplicates(true) when setting up the file watcher, it defaults into a CompositeFileListFilter, and when used with a LastModifiedFileListFilter it behaves unexpectedly. 

You can work around it by using a ChainFileListFilter instead, but that often throws people off. 

Hope that makes sense.

artembilan:
I’m not sure what you mean with the “throws people off”, but GH is not the place to discuss opinions. 

If you have strong position how that should be, feel free to raise a JIRA to discuss. This one was about exactly `LastModified`, but not `AcceptOnce`.

PR #2415, opened 2018-04-09
INT-4445: Fix JDBC tests for Derby

artembilan:
JIRA: https://jira.spring.io/browse/INT-4445

Looks like `/dataSource` temporary directory is busy in between different
tests.
The thought is like a high-frequently polling endpoint keeps DB resource
from removing.

* Explicitly `stop()` endpoint in the `JdbcMessageStoreChannelTests`
* Optimize
`JdbcMessageStoreChannelTests.testSendAndActivateTransactionalSend()`
to rely on the short `1` millisecond TX timeout.
This safes for us 10 seconds of the tests executions.
* Configure all the embedded DB beans for the `ignore-failures="ALL"`
as a fallback option if polling endpoint is not a cause of the concurrent
resource access.

**Cherry-pick to 5.0.x**


garyrussell:
and cherry-picked as 7d8fd16e0c142518d835ee2bf29f9a092ba8749c

PR #2417, opened 2018-04-10
Upgrade to SF 5.1; Remove Deprecated Code

garyrussell:


spring-integration-test-support/src/main/java/org/springframework/integration/test/rule/Log4jLevelAdjuster.java
```diff
@@ -1,112 +0,0 @@
...
- * @author Dave Syer
- * @author Gary Russell
- *
- * @deprecated since 5.0.1 in favor of {@link Log4j2LevelAdjuster}.
- * Will be removed in 5.1.
- *
- */
-@Deprecated
-@SuppressWarnings("deprecation")
-public class Log4jLevelAdjuster implements MethodRule {
```
artembilan: Sounds like we also need to remove such a dependency as well. Finally! :smile:

artembilan (changes requested):

artembilan:
Cool!
With that we found a usage of deprecated API :smile: 
```
:spring-integration-ip:compileTestJava/home/travis/build/spring-projects/spring-integration/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/InterceptedSharedConnectionTests.java:25: error: package org.apache.log4j does not exist
import org.apache.log4j.Level;
                       ^
/home/travis/build/spring-projects/spring-integration/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/InterceptedSharedConnectionTests.java:26: error: package org.apache.log4j does not exist
import org.apache.log4j.LogManager;
                       ^
/home/travis/build/spring-projects/spring-integration/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/InterceptedSharedConnectionTests.java:70: error: cannot find symbol
	private static Level existingLogLevel;
```

garyrussell:
Weird:

BUILD SUCCESSFUL in 7m 17s
356 actionable tasks: 353 executed, 3 up-to-date

build.gradle
```diff
@@ -485,6 +485,7 @@ project('spring-integration-jmx') {
 		compile project(":spring-integration-core")
 		testCompile "org.aspectj:aspectjweaver:$aspectjVersion"
 		testCompile "com.hazelcast:hazelcast:$hazelcastVersion"
+		testRuntime 'log4j:log4j:1.2.17' // Needed by hazelcast
```
artembilan: FYI: http://docs.hazelcast.org/docs/latest-development/manual/html/Setting_Up_Clusters/Logging_Configuration.html.

So, I think we still can avoid this dependency.

In the Hazelcast extension I use this:
```
jvmArgs "-Dhazelcast.logging.type=slf4j"
```
In the test task definition.

Maybe that will work here as well for log4j2?

artembilan (changes requested):

garyrussell:
Didn't work - that must be for the server. Found it though...

```java
		@Bean
		public HazelcastInstance hazelcastInstance() {
			return Hazelcast.newHazelcastInstance(new Config().setProperty("hazelcast.logging.type", "log4j"));
		}
```

artembilan (approved):
Travis failure not related to your changes:
```
org.springframework.integration.jdbc.store.JdbcMessageStoreTests > testAddAndUpdateWithChange FAILED
    org.springframework.transaction.CannotCreateTransactionException
        Caused by: java.sql.SQLException
            Caused by: org.apache.derby.iapi.error.StandardException
```
I guess that is my recent `ignore-failures="ALL"` fix which hides from us the real `Cannot delete DataSource directory` error. 

I wish I could understand why we have started to get such a problem recently.
Doesn't look like we have introduced some polluting JDBC test.

Well, LGTM.
Pulling locally for final testing and merging...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b940e3872b15baf7922539a252b9368b2d2b7f03

PR #2420, opened 2018-04-17
INT-4448, INT-4449: Fix Gateway for no-arg method

artembilan:
JIRA: https://jira.spring.io/browse/INT-4448
JIRA: https://jira.spring.io/browse/INT-4449

When we are not interested in the `payload` to send, we use a gateway
method without any args, but in this case for send operation (or
send-and-receive) we should specify a default `payloadExpression`

The `MessagingGatewayRegistrar` fails with `NPE` if we don't have a
any global headers and have `defaultPayloadExpression`

Also in this case the `GatewayProxyFactoryBean` fails to send
and fallbacks to receive with the meaning "no args, not payloadExpression"

* Fix `MessagingGatewayRegistrar` to check `hasDefaultHeaders` before
processing them
* Fix `GatewayProxyFactoryBean` to consult `this.globalMethodMetadata`
if there is no `payloadExpression` for the method specific metadata

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ addons:
     - mongodb-3.0-precise
     packages:
     - mongodb-org-server
-    - oracle-java8-installer
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
 cache:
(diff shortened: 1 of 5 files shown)


artembilan:
@garyrussell ,

Would you mind to take a look/merge.

This is kinda blocker since no one new PR can't pass Travis for proper coverage.

Thanks

garyrussell:
cherry-picked to 5.0.x, 4.3.x

PR #2421, opened 2018-04-17
INT-4447: LockRegLeaderInit: Catch unlock errors

artembilan:
JIRA: https://jira.spring.io/browse/INT-4447

When we get an exception during `this.lock.unlock()`, we don't revoke
leadership.
In case of external resource (e.g. JDBC) this may cause a race condition
when the second candidate is selected as leader when connection comes back

* Catch `this.lock.unlock()` exceptions and log them under DEBUG.
This way we proceed to the `handleRevoked()` logic

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -381,8 +381,15 @@ public class LockRegistryLeaderInitiator implements SmartLifecycle, DisposableBe
 					}
 					catch (Exception e) {
 						if (this.locked) {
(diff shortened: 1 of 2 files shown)


garyrussell:
Cherry-picked as 38ec9a16a66a827d5d01f70fd8686812016dd10e to 5.0.x.

4.3.x needs some backport work; I don't recall why, but the code is different there due to this: https://github.com/spring-projects/spring-integration/commit/0fe4c2e8aecdbbd3591a3e6ea804cdc135166cae

artembilan:
OK. I will issue back-port PR later.

artembilan:
Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/9c0a628f78ec5789a2dfbdc3f35ce6bdba5dc59f

PR #2423, opened 2018-04-19
Upgrade to Gradle 4.7 and others, polishing build

artembilan:
* Fix `MqttAdapterTests` according requirement of the latest Mockito
* Move `int-derby.properties` to the `/resources` from `/java`


build.gradle
```diff
@@ -298,12 +285,7 @@ project('spring-integration-amqp') {
 	description = 'Spring Integration AMQP Support'
 	dependencies {
 		compile project(":spring-integration-core")
-		compile("org.springframework.amqp:spring-rabbit:$springAmqpVersion") {
-			exclude group: 'org.springframework', module: 'spring-aop'
-			exclude group: 'org.springframework', module: 'spring-context'
-			exclude group: 'org.springframework', module: 'spring-core'
-			exclude group: 'org.springframework', module: 'spring-tx'
```
garyrussell: Why are you removing all these exclusions? (I don't remember why we added them).

garyrussell (commented):
That's all; just one question.

build.gradle
artembilan: We added them to avoid classpath conflicts, mostly in case of IO Platform compatibility tests.
1. Now we don't have them
2. Now we keep all our Spring dependencies aligned with the latest SF.
3. Pay attention how I didn't remove exclusion from the `spring-retry` - it is still based on the SF `4.3.x`.

artembilan:
```
org.springframework.integration.jdbc.store.JdbcMessageStoreTests > testAddAndUpdateWithChange FAILED
    org.springframework.transaction.CannotCreateTransactionException
        Caused by: java.sql.SQLException
            Caused by: org.apache.derby.iapi.error.StandardException
```
M-m-m. I already hate that instability...

In the next PR I'm going to allow all the failures for Derby: hope we will figure out with subsequent failures what's going on. Can't recall what kind of change we did there that we have a problem with that DB resource :cry:

PR #2428, opened 2018-04-24
INT-4455: Fix NPE in Exception router

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4455

When using java config, `setChannelMapping` NPEs due to no AC.

**cherry-pick to 4.3.x, and to master, removing AC check**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/router/ErrorMessageExceptionTypeRouter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/router/ErrorMessageExceptionTypeRouter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,6 +41,7 @@ import org.springframework.util.ClassUtils;
  * @author Mark Fisher
  * @author Oleg Zhurakousky
  * @author Artem Bilan
+ * @author Gary Russell
  */
 public class ErrorMessageExceptionTypeRouter extends AbstractMappingMessageRouter {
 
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/router/ErrorMessageExceptionTypeRouter.java
```diff
@@ -67,7 +68,9 @@ private void populateClassNameMapping(Set<String> classNames) {
 
 	private Class<?> resolveClassFromName(String className) {
 		try {
-			return ClassUtils.forName(className, getApplicationContext().getClassLoader());
+			ClassLoader classLoader = getApplicationContext() == null ? getClass().getClassLoader()
```
artembilan: I would prefer do not fallback to the `getClass().getClassLoader()`.
The `Assert.state()` in this method should be great compromise.

spring-integration-core/src/main/java/org/springframework/integration/router/ErrorMessageExceptionTypeRouter.java
```diff
@@ -78,9 +81,12 @@ private void populateClassNameMapping(Set<String> classNames) {
 	@ManagedOperation
 	public void setChannelMapping(String key, String channelName) {
 		super.setChannelMapping(key, channelName);
-		Map<String, Class<?>> newClassNameMappings = new ConcurrentHashMap<>(this.classNameMappings);
-		newClassNameMappings.put(key, resolveClassFromName(key));
-		this.classNameMappings = newClassNameMappings;
+		// TODO: remove application context check in 5.1, here for backwards compatibility
+		if (this.initialized || getApplicationContext() != null) {
```
artembilan: Why do we really need to check here for `AC` ?
The `initialized` flag should be enough: we don't go to load classes until `onInit()` if otherwise.
Why should one worry about class loading if its not a bean yet ?

artembilan (changes requested):

Dmitrin (approved):

artembilan (approved):
Oh! You fixed it in the latest commit.
Good. I'm happy then. Merging... :smiley:

artembilan:
... and cherry-picked to `4.3.x` with resolving diamond operators for Java 6 compatibility.
And to `5.1` with resolving TODO.

PR #2429, opened 2018-04-26
Add Reactive mode for AbstractPollingEndpoint

artembilan:
* When `SourcePollingChannelAdapter.outputChannel` is a
`ReactiveStreamsSubscribableChannel`, use `Flux.generate()` for polling
* Refactor `AbstractPollingEndpoint` to remove redundant `Poller` class
in favor of lambda
* Extract `pollForMessage()` method to handle TX states instead of
`Poller` class previously

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -16,7 +16,9 @@
 
 package org.springframework.integration.endpoint;
 
+import java.time.Duration;
 import java.util.Collection;
+import java.util.Date;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
@@ -24,6 +26,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.stream.Collectors;
 
 import org.aopalliance.aop.Advice;
+import org.reactivestreams.Subscription;
 
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.beans.factory.BeanClassLoaderAware;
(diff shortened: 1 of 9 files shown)


artembilan:
@garyrussell ,

any chances that this can make it into our current `5.1`?
Then I will shade your one on the matter: #2433 . With slightly different approach since our `FluxMessageChannel` have some attraction already, especially after adoption into the Spring Cloud Stream.

I mean that I had recently a question about Reactive JMS on Gitter: https://gitter.im/spring-projects/spring-integration?at=5b9644fef3c26b08f65608d4

This Reactive polling adaption will do for us a trick when we won't ask broker for more messages until downstream demand.

I believe we can do something similar with the event-driven channel adapters when we will block consuming thread until demand from the downstream. Different story though...

Thanks

garyrussell:
Sounds like a plan if it can make the RC1 on 9/20.

artembilan:
So, Gary, can we merge it now or you would like to see anything else here?

I would address Docs in the separate PR somewhere in between RC and GA.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -66,23 +74,27 @@
 
 	private boolean syncExecutor = true;
 
+	private ClassLoader beanClassLoader = ClassUtils.getDefaultClassLoader();
+
+	private Trigger trigger = new PeriodicTrigger(10);
```
garyrussell: Default 10ms? Really?

If we **must** have a default, it should be much larger than this.

Why wouldn't we just say it's invalid to configure a reactive poller without a trigger?

garyrussell (changes requested):
Just one issue.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
artembilan: Huh? This is an existing code. I just reformatted it a bit to remove some `volatile`.
We can reconsider the default in other issue, although I think Mark did this way to make as close to the event-driven behavior as possible

garyrussell: Doh; merging.

garyrussell (approved):

PR #2430, opened 2018-04-27
INT-4456: Explicit methods for Kotlin lambdas

artembilan:
JIRA: https://jira.spring.io/browse/INT-4456

Since Kotlin compiles lambdas different way than Java, they are not
synthetic classes at runtime anymore.
Therefore we fallback to the method invocation logic, but since Java 8
interfaces has `default` methods as well the `MessagingMethodInvokerHelper`
can't select the target method for invocation

* Use explicit `Function.apply()` for non-synthetic implementations
* Add `RouterDslTests` Kotlin-based test-case

**Cherry-pick to 5.0.x**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,12 @@
 buildscript {
+	ext.kotlinVersion = '1.2.31'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
 	dependencies {
 		classpath 'io.spring.gradle:docbook-reference-plugin:0.3.1'
 		classpath 'org.asciidoctor:asciidoctor-gradle-plugin:1.5.3'
+		classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
 	}
 }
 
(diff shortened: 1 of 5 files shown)


artembilan:
If this is OK, I can go ahead and add similar explicit methods for all other lambda-based DSL methods.
Seems for me with that we will have a better performance since `Method` would be already pre-selected.

build.gradle
```diff
@@ -72,11 +75,18 @@ subprojects { subproject ->
 		targetCompatibility = 1.8
 	}
 
+	compileKotlin {
```
garyrussell: compiltTestKotlin?

garyrussell (commented):
Cool - that's all - just one comment; can change during merge if you agree.

garyrussell:
>If this is OK, I can go ahead and add similar explicit methods for all other lambda-based DSL methods.

Makes sense.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/0c1423e78a489c34480322f7c4f2fcecee08050e and cherry-picked to `5.0.x`.

PR #2432, opened 2018-04-30
INT-4458: Do not expose recursive generics API

artembilan:
JIRA: https://jira.spring.io/browse/INT-4458

According Kotlin generics system restrictions we can't expose API
based on the recursive generics, like we have with Java DSL for JMS

* Replace explicit recursive generics on factory method with the wildcard (`?`).
This way both Kotlin and Java are able to instantiate target object and perform
the proper chain API auto-completion
* Fix `JmsInboundChannelAdapterSpec` to deal with provided `S` type for the
`configureListenerContainer()` - the way we can configure a `JmsDefaultListenerContainerSpec`
* Add `kotlin-spring` Gradle plugin to avoid extra `open` modificator on `@Bean` methods
* Add `JmsDslKotlinTests.kt`

**Cherry-pick to 5.0.x excluding Kotlin support**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlinVersion = '1.2.40'
+	ext.kotlinVersion = '1.2.41'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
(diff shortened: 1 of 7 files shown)


artembilan:
If that's OK, I can go ahead and fix similar problem in AMQP DSL and other.

artembilan:
Well, I can do in the separate PR(s), but then it's more work during cherry-picking.
So, I'd prefer to do here as one unit of work, because this PR doesn't address all the problems.
Therefore [DO NOT MERGE YET] :smile:

artembilan:
Looks like nothing more to fix.
Ready for final review.

Thanks

artembilan:
Oh! Looks like we have back-ported Kotlin tests to `5.0.x`! 
So, this is simple to merge then.
Let me push some polishing to the test-case! :smile:

PR #2434, opened 2018-05-04
INT-4461: Support byte[] in #jsonPath

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4461

Convert `byte[]` to `String` using `URF-8` by default.

Add optional evaluate methods to JsonPathUtils with a Charset to use when
converting `byte[]` to `String`.

This is not currently exposed using SpEL. It can be done, but probably not worth
the effort until somebody asks for it.


garyrussell:
I prefer your InputStream suggestion - will rework.

artembilan:
... and cherry-picked to `5.0.x` and `4.3.x` after some minor conflicts fixes.

PR #2435, opened 2018-05-04
Use Lettuce Redis client; Redis module fixes

artembilan:
The Lettuce client is based on the Netty and more stable, than Jedis
therefore we get a performance improvement for tests
(it saves us at about 30 seconds).
Also this client doesn't fail for me on Windows sporadically (very often)
 with the `ConnectionClosedException`

* After switching to the Netty-based client, we expose the interrupted
Thread issue in the `LockRegistryLeaderInitiator`.
If we interrupted (expected behavior), we try to unlock calling
`RedisLockRegistry`, but Netty client reject our request because the
thread is interrupted, therefore we never delete the lock when we yield
our leadership.
Fix the issue with shifting a `RedisTemplate.delete()` operation to the
`ExecutorService` when the current thread is interrupted
* Allow to configure such an `ExecutorService` and timeout to wait for
the `submit()` result
* Refactor `RedisAvailableRule` and all the Redis tests do not expose
the target `RedisConnectionFactory` implementation.
* Make all the test-cases based on the `connectionFactory` created by
the `RedisAvailableTests.setupConnectionFactory()`
* Tweak some unnecessary timeouts and sleeps for better tests task
throughput

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,7 +108,6 @@ subprojects { subproject ->
 		jackson2Version = '2.9.5'
 		javaxActivationVersion = '1.1.1'
 		javaxMailVersion = '1.6.1'
-		jedisVersion = '2.9.0'
 		jmsApiVersion = '2.0.1'
 		jpa21ApiVersion = '1.0.0.Final'
 		jpaApiVersion = '2.1.1'
@@ -120,6 +119,7 @@ subprojects { subproject ->
 		junitPlatformVersion = '1.1.1'
 		jythonVersion = '2.5.3'
 		kryoShadedVersion = '3.0.3'
+		lettuceVersion = '5.0.3.RELEASE'
 		log4jVersion = '2.11.0'
 		micrometerVersion = '1.0.3'
 		mockitoVersion = '2.18.0'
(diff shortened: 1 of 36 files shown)


artembilan:
We may consider this for back-porting, especially the fix for `RedisLockRegistry`.

garyrussell:
>org.springframework.integration.redis.leader.RedisLockRegistryLeaderInitiatorTests > testDistributedLeaderElection FAILED
    java.lang.AssertionError at RedisLockRegistryLeaderInitiatorTests.java:144
142 tests completed, 1 failed, 3 skipped

artembilan:
The Travis is green after the second attempts, but there is indeed something else to do there.
Let's see how it works for you and on CI server!
The `Log4j2LevelAdjuster` should help us.

artembilan:
@garyrussell ,

Do you want a separate PR for back-porting or are you OK to cherry-pick this fully ?

garyrussell:
Before backporting, I think we need to be sure the src/main changes still work with the Jedis library.

If you've already done that, we can cherry-pick; if not, we should backport just the src/main changes first.

PR #2439, opened 2018-05-09
INT-4465: Fix delay in close propagation with NIO

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4465

There is a one second delay before a socket close is propagated if there is an active
assembler. This is generally only a problem with deserializers that use EOF to signal
message end (such as the `ByteArrayElasticRawDeserializer`).

Attempt to insert an EOF marker into the buffer queue so that the `getNextBuffer()` will
exit immediately on `close()` if it is blocked awaiting a buffer.


artembilan:
Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/f912aadb4be72a277ffe0f35fd00cba13fc09ed5 after some conflicts resolution - the `ApplicationEventPublisher` is not a functional interface in the SF `4.3`...

PR #2440, opened 2018-05-10
Wrap non-`StandardIntegrationFlow` into Proxy

artembilan:
In previous version all the `IntegrationFlow` beans have been replaced
by the `StandardIntegrationFlow` i the `IntegrationFlowBeanPostProcessor`
if they are lambda.
That works for Java, but doesn't with Kotlin, since lambdas i last one
are not synthetic classes.
Therefore some Java DSL definitions (especially `.subFlowMapping()`)
don't work consistently in two languages.

* Introduce `IntegrationFlowLifecycleAdvice` to wrap all the
non-`StandardIntegrationFlow`s (excluding `IntegrationFlowAdapter`)
into the `Proxy` to expose `SmartLifecycle` and `getInputChannel()`
operations and delegate them to the internal `StandardIntegrationFlow`
created by the `IntegrationFlowBeanPostProcessor`.
* This way any custom `IntegrationFlow` implementations can be used
for manual flow registration via `IntegrationFlowContext`
* Polish `RouterDslTests.kt` for the `@Bean`s for sub-flows.
* Document in the `dsl.adoc` a request-reply approach for the case
when `.subFlowMapping()` refers to an `IntegrationFlow` `@Bean`.
* Polishing for the `FlowServiceTests` since all the
non-`StandardIntegrationFlow`s and not-`IntegrationFlowAdapter`s are
wrapped now to the Proxy.
* Add missing `from()` delegations into the `IntegrationFlowAdapter`
* Polishing for the `ManualFlowTests` since all the `IntegrationFlow`
now are `Lifecycle` after wrapping to the Proxy.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
@@ -253,6 +253,15 @@ public class IntegrationFlowBeanPostProcessor
 		return flow;
 	}
 
+	/**
+	 * Only invoked for {@link IntegrationFlow} instances that are not
+	 * {@link StandardIntegrationFlow}s; typically lambdas. Creates a new
+	 * {@link StandardIntegrationFlow} with an input channel named {@code beanName.input}
+	 * and the flow defined by the flow parameter. If the flow is not an
+	 * {@link IntegrationFlowAdapter} the original, user-provided {@link IntegrationFlow}
+	 * is wrapped in a proxy and advised with a {@link IntegrationFlowLifecycleAdvice};
+	 * see its javadocs for more information.
+	 */
 	private Object processIntegrationFlowImpl(IntegrationFlow flow, String beanName) {
 		IntegrationFlowBuilder flowBuilder = IntegrationFlows.from(beanName + ".input");
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
@@ -20,6 +20,8 @@ import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
+import org.springframework.aop.framework.ProxyFactory;
+import org.springframework.aop.support.NameMatchMethodPointcutAdvisor;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanCreationNotAllowedException;
 import org.springframework.beans.factory.BeanFactory;
@@ -33,6 +35,7 @@ import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.support.AbstractBeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.context.SmartLifecycle;
 import org.springframework.core.io.DescriptiveResource;
 import org.springframework.integration.channel.AbstractMessageChannel;
 import org.springframework.integration.channel.DirectChannel;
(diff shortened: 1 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowLifecycleAdvice.java
```diff
@@ -0,0 +1,94 @@
...
+package org.springframework.integration.dsl;
+
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+
+import org.springframework.context.SmartLifecycle;
+import org.springframework.util.ObjectUtils;
+
+/**
+ * @author Artem Bilan
```
garyrussell: Can we have some javadocs, please?

For example, it's not at all clear why we first try to get the input channel from the flow and, if null, from the delegate.

It's also not clear why we are invoking lifecycle methods to both the flow and the delegate, with the `result` always being from the delegate.

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowLifecycleAdvice.java
```diff
@@ -0,0 +1,94 @@
...
+
+		if ("getInputChannel".equals(method)) {
+			result = invocation.proceed();
+			if (result == null) {
+				result = this.delegate.getInputChannel();
+			}
+		}
+		else {
+			if (target instanceof SmartLifecycle) {
+				result = invocation.proceed();
```
garyrussell: Why do we save the result here? It's always overwritten by the delegate result.

src/reference/asciidoc/dsl.adoc
```diff
@@ -643,6 +643,42 @@ public IntegrationFlow routeFlow() {
...
+    return f -> f.handle(m -> System.out.println("odd"));
+}
+
+@Bean
+public IntegrationFlow evenFlow() {
+    return f -> f.handle((p, h) -> "even");
+}
+----
+
+In this case, when you need to receive a reply from such a sub-flow and continue the main flow, this `IntegrationFlow` bean reference (or its input channel) has to be wrapped with the `.gateway()` as shown above.
```
garyrussell: with __a__ gateway.

src/reference/asciidoc/dsl.adoc
```diff
@@ -643,6 +643,42 @@ public IntegrationFlow routeFlow() {
...
+The `oddFlow()` referece in the sample above isn't not wrapped to the `.gateway()` - therefore we don't expect a reply from this routing branch.
+Otherwise you end up with an exception like:
+....
+Caused by: org.springframework.beans.factory.BeanCreationException: The 'currentComponent' (org.springframework.integration.router.MethodInvokingRouter@7965a51c) is a one-way 'MessageHandler' and it isn't appropriate to configure 'outputChannel'. This is the end of the integration flow.
+....
+When sub-flow is configured as a lambda, the Framework takes care about request-reply interaction with the sub-flow.
```
garyrussell: When __a__ sub-flow ... the Framework handles the request-reply interaction with the sub-flow and a gateway is not needed.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
```diff
@@ -252,9 +255,33 @@ else if (component instanceof AnnotationGatewayProxyFactoryBean) {
...
+					"start",
+					"stop",
+					"isRunning",
+					"isAutoStartup",
+					"getPhase");
+
+			ProxyFactory proxyFactory = new ProxyFactory(flow);
+			proxyFactory.addAdvisor(integrationFlowAdvice);
+			if (!(flow instanceof SmartLifecycle)) {
+				proxyFactory.addInterface(SmartLifecycle.class);
```
garyrussell: Suggested method comments here https://github.com/garyrussell/spring-integration/commit/ee8015a87998106f2c4ae5e5d5a7fae8102bd2b2

I wonder if we should use `ClassUtils.getAllInterfaces`, and add them here, in case a user implements `IntegrationFlow` and additional interfaces.

artembilan: We don't need that information. The `Advice` is applied only to the restricted set of methods. Everything else is left untouched.

We add a `SmartLifecycle` here explicitly because we need it for the Manual flow registration and really for propagation lifecycle control to the `target`, which is not available from the application context anymore.

artembilan: Interesting "why?". This is JDK Proxy, so all original interfaces must be on the target wrapped object...

artembilan:
```
public interface Foo {

}

@Component
public static class MyFlow implements IntegrationFlow, Foo {
```

and in the test case:
```
assertThat(this.myFlow, instanceOf(Foo.class));
```

What am I missing?

artembilan: I also have just added `SmartLifecycle` into that `MyFlow` class definition. Works well.
:man_shrugging:

garyrussell: Doh; yes; of course; I was thinking CGLib.

garyrussell (approved):

PR #2448, opened 2018-05-17
Fix IntegrationFlowContext concurrency issue

artembilan:
When we register `IntegrationFlow` s concurrently at runtime, we may
end up with the problem when we register the same object with the same
bean name, but in different places.
Or when we turn off bean overriding, we end up with the exception that
bean with the name already registered

* Wrap `IntegrationFlow` bean registration in the
`StandardIntegrationFlowContext` into the `Lock` when its bean name
is generating
* Make `StandardIntegrationFlowContext.registry` as `ConcurrentHashMap`
to avoid `ConcurrentModificationException` during `put()` and `remove()`
* Fix concurrency for beans registration with the generation names in
the `IntegrationFlowBeanPostProcessor` using an `IntegrationFlow` id
as a prefix for uniqueness.

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowBeanPostProcessor.java
@@ -123,7 +123,7 @@ public class IntegrationFlowBeanPostProcessor
 				String id = endpointSpec.getId();
 
 				if (id == null) {
-					id = generateBeanName(endpoint, entry.getValue());
+					id = generateBeanName(endpoint, flowNamePrefix, entry.getValue());
 				}
 
 				Collection<?> messageHandlers =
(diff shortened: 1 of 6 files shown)


artembilan:
See https://stackoverflow.com/questions/50393906/spring-integration-java-dsl-dynamically-create-integrationflows.

And also https://jira.spring.io/browse/SPR-16837

artembilan:
I think there is no reason in locking/synchronization in the `IntegrationFlowBeanPostProcessor` - we simply can generate bean names based on the flow id as a prefix.

Will come back with such a fix later today...

artembilan:
Now it's ready for review and merging.

Thanks

garyrussell:
Do you think the travis failure is legitimate, or should I try to build locally?

garyrussell:
I am a bit uncomfortable cherry-picking the bean name change to 5.0.x since it's a breaking change. But perhaps we have no choice.

We should add something to the what's new, DSL chapter, and migration guide(s) for the bean name change.

artembilan:
Right, will do.

Thanks for the reminders!

artembilan:
I will add Migration Guide note after this is merged.

garyrussell:
Doc polishing for your review: https://github.com/garyrussell/spring-integration/commit/e572c682efe2d709c455d13806d3a951bc85f852

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b21dc0dff353d735f7acf915ab189550297c2ceb.

Cherry-picked to `5.0.x` as: https://github.com/spring-projects/spring-integration/commit/d0603a80ff5d89d14681ed6acec17c94c08e9564

PR #2449, opened 2018-05-18
Fix IntegrationFlow <-> IntFlowDefinition tangles

artembilan:


artembilan:
@garyrussell ,

Let me know, please, if this is a valid classes tangle fix.

Related to https://jira.spring.io/browse/INT-4390.

I moved `IntegrationFlowDefinition` in to the `IntegrationFlow.Definition` inner class, because they use each other. 
The same I did for the `IntegrationFlowBuilder` - moved to the inner `IntegrationFlow.Builder`.

Would you also mind to share the latest Structure 21 report, please?

Thank you!

garyrussell:
Cool - I looked at something similar - moving IF into IFD - but I thought that would be too much of a breaking change.

So this fixes 2 of the tangles; we're left with tangles between IF and several specs (due to subflows I believe); also `EndpointSpec.obtainInputChannelFromFlow()` and a couple of others. I can give you complete details if you need.

Before this PR:

screen shot 2018-05-18 at 4 20 53 pm

After:

screen shot 2018-05-18 at 4 26 00 pm

artembilan:
Cool! Thank you!

At least now I start feeling what is the tangle and how to fix it :smile: 

Although it's still unclear to me what is the harm with this, since compiler lets us to proceed with the  code structure we have so far...

I'll proceed then with polishing on the matter.

I think this should not be a big breaking change for end-users since mostly they use it via lambdas or `IntegrationFlows` factory.

However would be better to have #2448 merged before this painful change :smile:

garyrussell:
It LGTM; just waiting for travis.

artembilan:
@garyrussell ,

You know this fix might not be correct from the target Specs perspective, even if we fix an internal API do not use Builder directly for sub-flows, the public API would be still the same, e.g.:
```
public S defaultSubFlowMapping(IntegrationFlow subFlow) {
```

I mean with the current state we may remove tangling with the `IntegrationFlowDefinition`, but we can't with the `IntegrationFlow`, because both `IntegrationFlow` and `AbstractRouterSpec` are public API. 
And since we have moved here a `Definition` inside the `IntegrationFlow`, the tangle with `AbstractRouterSpec` has moved there as well.

So, this fix is even much worse, than it was before.

Need to think more how to be...

Anyway feel free to take a look and provide your feedback.

artembilan:
However with the contract like `void configure(IntegrationFlowDefinition<?> flow);` in the `IntegrationFlow` we still have the tangle from those components which provide a sub-flow support. Therefore it doesn't matter if we accept this PR or not: the problem is still there.

No ideas so far...

artembilan:
I think the situation is like we can't fix `IntegrationFlow <-> Spec` tangles, since the sub-flow is really a feature and it looks and feels naturally that they call each each recursively. The API is aligned and very clear from IDE auto-completion assistant, especially when we come to the request for external bean for the sub-flow and when we deal with strict typing model in the Kotlin.

I see only one ugly solution like place all the Specs into the `IntegrationFlow` class. :cry:

garyrussell:
>place all the Specs

Let's not go there. I agree this is really hard (if even possible) to fix; we may just have to live with it.

artembilan:
Well, then this one is ready for merge. 
Or let's just reject it and continue to live with top-level classes for `Definition` and `Builder`!

WDYT?

artembilan:
@garyrussell ,

I'm intended to close this PR without any merging since we agreed that there is no way to fix DSL builders tangles and it feels for me better to keep the classes moved in this PR on top level - they are used from the end-user perspective anyway.

So, WDYT?

garyrussell:
Unfortunate, but yes, agreed.

PR #2450, opened 2018-05-21
INT-4473: Support prefix bean names with flow id

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4473

Previously, dynamic registration of integration flows with components configured
with the same `id` would fail with duplicate bean names.

Add `useFlowIdAsPrefix()` to the registration builder to enable the option.
Then, in the BPP, check the flag before naming the beans.

**cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/StandardIntegrationFlowContext.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/StandardIntegrationFlowContext.java
@@ -37,6 +37,7 @@ import org.springframework.integration.dsl.IntegrationFlow;
 import org.springframework.integration.support.context.NamedComponent;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Standard implementation of {@link IntegrationFlowContext}.
@@ -50,6 +51,8 @@ import org.springframework.util.Assert;
 public final class StandardIntegrationFlowContext implements IntegrationFlowContext, BeanFactoryAware {
 
 	private final Map<String, IntegrationFlowRegistration> registry = new ConcurrentHashMap<>();
+
+	private final Map<String, Boolean> useFlowIdAsPrefix = new ConcurrentHashMap<>();
 
 	private final Lock registerFlowsLock = new ReentrantLock();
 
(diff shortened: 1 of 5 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/context/StandardIntegrationFlowContext.java
```diff
@@ -78,6 +81,12 @@ public StandardIntegrationFlowRegistrationBuilder registration(IntegrationFlow i
 		return new StandardIntegrationFlowRegistrationBuilder(integrationFlow);
 	}
 
+	@Override
+	public boolean isUseIdAsPrefix(String flowId) {
+		Boolean use = this.useFlowIdAsPrefix.get(flowId);
+		return use == null ? false : use;
```
artembilan: `return Boolean.TRUE.equals(use)` ?

src/reference/asciidoc/dsl.adoc
```diff
@@ -933,10 +933,44 @@ Usually those additional beans are connection factories (AMQP, JMS, (S)FTP, TCP/
 Such a dynamically registered `IntegrationFlow` and all its dependant beans can be removed afterwards using `IntegrationFlowRegistration.destroy()` callback.
 See `IntegrationFlowContext` JavaDocs for more information.
 
-NOTE: Starting with _version 5.0.5_, all generated bean names in an `IntegrationFlow` definition are prepended with flow id as a prefix.
+NOTE: Starting with _version 5.0.6_, all generated bean names in an `IntegrationFlow` definition are prepended with flow id as a prefix.
 It is recommended to always specify an explicit flow id, otherwise a synchronization barrier is initiated in the `IntegrationFlowContext` to generate the bean name for the `IntegrationFlow` and register its beans.
 We synchronize on these two operations to avoid a race condition when the same generated bean name may be used for different `IntegrationFlow` instances.
 
+Also, starting with _version 5.0.6_, the registration builder API has a new method `useFlowIdAsPrefix()`.
+This is useful if you wish to declare multiple instances of the same flow and avoids bean name collisions if components in the flows have the same id.
```
artembilan: s/avoids/avoid ?

artembilan (changes requested):
Just a couple nit-picks.

I can address on merge.

Also I believe it won't be so easy to cherry-pick since classes for the `context` have been moved a bit.

Not a big deal though. Just let me know!

spring-integration-core/src/main/java/org/springframework/integration/dsl/context/StandardIntegrationFlowContext.java
garyrussell: 👍

src/reference/asciidoc/dsl.adoc
garyrussell: 👍

artembilan:
Gary, I did more changes, than expected.
See polishing here: https://github.com/artembilan/spring-integration/commit/10d50d6042ffc85823160323d54fa355c645f228

PR #2453, opened 2018-05-24
Changes to enable build with JDK10

garyrussell:
- ambiguous `.from()` in `IntegrationFlows`
- new Integer|Long(n) deprecated
- `Class.newInstance()` deprecated
- --add-modules where needed
- illegal reflection


spring-integration-core/src/test/java/org/springframework/integration/dsl/manualflow/ManualFlowTests.java
```diff
@@ -494,7 +495,7 @@ public Date foo() {
 
 		@Override
 		protected IntegrationFlowDefinition<?> buildFlow() {
-			return from(() -> new GenericMessage<>("flowAdapterMessage"),
+			return from((MessageSource<?>) () -> new GenericMessage<>("flowAdapterMessage"),
```
artembilan: I wonder if that would better to rename to `fromSupplier()`. And yeah... remove that `from(Supplier)` altogether.

Breaking change, agreed, but no choice.

I will find this explicit casting style more annoying than removed method.

WDYT?

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionTests.java
```diff
@@ -254,6 +254,7 @@ public void testMemoryLeak() throws Exception {
 	}
 
 	@Test
+	@EnabledOnJre(JRE.JAVA_8)
```
artembilan: Would you mind elaborate more for this use-case?

Is something wrong with reflection?

spring-integration-jpa/src/test/java/org/springframework/integration/jpa/core/EclipseLinkJpaOperationsTests.java
```diff
@@ -41,8 +43,8 @@
  * @since 2.2
  *
  */
-@RunWith(SpringJUnit4ClassRunner.class)
-@ContextConfiguration
+@SpringJUnitConfig
+@EnabledOnJre(JRE.JAVA_8)
```
artembilan: Maybe it's time to remove Eclipse Link tests altogether ?

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/support/MongoDbMessageBytesConverter.java
```diff
@@ -32,14 +32,17 @@
...
  * A {@link GenericConverter} implementation to convert {@link Message} to
- * serialized {@link byte[]} to store {@link Message} to the MongoDB.
- * And vice versa - to convert {@link byte[]} from the MongoDB to the {@link Message}.
+ * serialized {@code byte[]} to store {@link Message} to the MongoDB.
+ * And vice versa - to convert {@code byte[]} from the MongoDB to the {@link Message}.
 
  * @author Artem Bilan
  * @author Gary Russell
  * @since 4.2.10
  * @deprecated since 5.0 in favor of {@link MessageToBinaryConverter} and {@link BinaryToMessageConverter}
```
artembilan: Maybe we can remove this class now?

I think that is a good catch for wrong JavaDocs - I don't see reason to add comment for that :smile: 

But thanks anyway!

spring-integration-scripting/src/test/java/org/springframework/integration/scripting/config/jsr223/Jsr223ServiceActivatorTests.java
```diff
@@ -123,6 +124,7 @@ public void withScriptVariableGenerator() throws Exception {
 	}
 
 	@Test
+	@EnabledOnJre(JRE.JAVA_8)
```
artembilan: What's wrong here, please ?

spring-integration-xml/src/test/java/org/springframework/integration/xml/config/XPathSplitterParserTests.java
```diff
@@ -63,11 +62,18 @@
...
 	public void testXpathSplitterConfig() {
 		assertTrue(TestUtils.getPropertyValue(this.xpathSplitter, "createDocuments", Boolean.class));
 		assertFalse(TestUtils.getPropertyValue(this.xpathSplitter, "applySequence", Boolean.class));
 		assertFalse(TestUtils.getPropertyValue(this.xpathSplitter, "iterator", Boolean.class));
 		assertSame(this.outputProperties, TestUtils.getPropertyValue(this.xpathSplitter, "outputProperties"));
+		/*
+		 * JDK 10
+		 * WARNING: An illegal reflective access operation has occurred
+		 * WARNING: Illegal reflective access by org.springframework.util.ReflectionUtils
+		 * to field com.sun.org.apache.xpath.internal.jaxp.XPathExpressionImpl.xpath
```
artembilan: Interesting...
Maybe we can just end up with the `"xpathExpression.expression` path?

Anyway what is the premise of the error?
Is there any ideas ?

Thanks

spring-integration-xml/src/test/java/org/springframework/integration/xml/transformer/jaxbmarshaling/JaxbMarshallingIntegrationTests.java
```diff
@@ -24,23 +24,26 @@
...
 import org.springframework.xml.transform.StringSource;
 
 /**
  * @author Jonas Partner
+ * @author Gary Russell
  */
-@ContextConfiguration
-public class JaxbMarshallingIntegrationTests extends AbstractJUnit4SpringContextTests {
+@SpringJUnitConfig
+@EnabledOnJre(JRE.JAVA_8)
```
artembilan: What here?

artembilan (changes requested):
The change is definitely good, but I left several doubts and questions.

Thanks

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionTests.java
garyrussell: Invalid reflection at line 276 - no such field in JDK10.

spring-integration-scripting/src/test/java/org/springframework/integration/scripting/config/jsr223/Jsr223ServiceActivatorTests.java
garyrussell:
```
java.lang.AssertionError
	at org.jruby.RubyModule.fetchConstant(RubyModule.java:4197)
	at org.jruby.RubyModule.getConstantAt(RubyModule.java:3681)
	at org.jruby.RubyModule.getConstantAt(RubyModule.java:3677)
	at org.jruby.javasupport.Java.getJavaPackageModule(Java.java:800)
	at org.jruby.javasupport.Java.getJavaPackageModule(Java.java:780)
	at org.jruby.javasupport.binding.ClassInitializer.initialize(ClassInitializer.java:47)
	at org.jruby.javasupport.binding.Initializer.setupProxyClass(Initializer.java:99)
```

spring-integration-xml/src/test/java/org/springframework/integration/xml/config/XPathSplitterParserTests.java
garyrussell: In a future compiler, such illegal accesses will be denied.

spring-integration-xml/src/test/java/org/springframework/integration/xml/transformer/jaxbmarshaling/JaxbMarshallingIntegrationTests.java
artembilan: Does this help?

https://discuss.gradle.org/t/gradle-does-not-honor-add-modules-jvm-argument-in-jdk9/20403

spring-integration-xml/src/test/java/org/springframework/integration/xml/config/XPathSplitterParserTests.java
artembilan: In other words all our "Direct fields access" magic is going to be banned?

spring-integration-scripting/src/test/java/org/springframework/integration/scripting/config/jsr223/Jsr223ServiceActivatorTests.java
artembilan: That's fine.
We may even remove all the Ruby tests.
I tried once to upgrade to the latest version and we have some conflicts with the Jython - they override some classes, but not fully :cry: 

I would say that JavaScript compatibility should be enough for us to trace: everything rest is just a target engine implementation and we are not responsible for their incompatibility with Java 10.

OK. Let's leave like you did. Maybe some comment on lines for the future consideration when Java 11 is going to be a base line ?

artembilan:
There is something like this in Gradle logs:
```
> Task :spring-integration-xmpp:javadoc
javadoc: warning - You have not specified the version of HTML to use.
The default is currently HTML 4.01, but this will change to HTML5
in a future release. To suppress this warning, please specify the
version of HTML used in your documentation comments and to be
generated by this doclet, using the -html4 or -html5 options.
```

I see similar warn for all our modules.

Also all the SFTP tests fail for me on Windows:
```
java.lang.NoClassDefFoundError: Could not initialize class javax.crypto.JceSecurity

	at java.base/javax.crypto.JceSecurityManager.<clinit>(JceSecurityManager.java:65)
	at java.base/javax.crypto.Cipher.getConfiguredPermission(Cipher.java:2613)
	at java.base/javax.crypto.Cipher.getMaxAllowedKeyLength(Cipher.java:2635)
	at org.apache.sshd.common.cipher.Cipher.checkSupported(Cipher.java:83)
	at org.apache.sshd.common.cipher.BuiltinCiphers.<init>(BuiltinCiphers.java:102)
	at org.apache.sshd.common.cipher.BuiltinCiphers.<clinit>(BuiltinCiphers.java:54)
	at org.apache.sshd.common.BaseBuilder.<clinit>(BaseBuilder.java:71)
	at org.apache.sshd.server.SshServer.setUpDefaultServer(SshServer.java:412)
	at org.springframework.integration.sftp.SftpTestSupport.createServer(SftpTestSupport.java:61)

```

Maybe it requires some `-add-modules` to include as well?
No ideas so far. Also interesting how it works for you...

Thanks

garyrussell:
Yes, I noticed those javadoc warnings. Still TODO.

I did see a couple of SFTP failures on Mac OS. I don't recall if it was that error (but it looks familiar), but the last few times I had no problems, so I don't think it's as simple as needing `add-modules`; more of an intermittent problem; at least on Mac OS.

artembilan (changes requested):

build.gradle
```diff
@@ -700,6 +711,14 @@ project('spring-integration-xml') {
 
 		testCompile "xmlunit:xmlunit:$xmlUnitVersion"
 	}
+
+	if (JavaVersion.current() != JavaVersion.VERSION_1_8) {
+		[compileTestJava]*.options*.compilerArgs = ['--add-modules', 'java.xml.bind']
```
artembilan: You are missing in all these manipulation our global `xLintArg` option.
I wonder if we should move all the `--add-modules` to the top level config, along side with our `[compileJava, compileTestJava]*.options*.compilerArgs = [xLintArg]` there...

artembilan (changes requested):
Unfortunately I can't find what is wrong with the SSL in case of SFTP module. Sounds like we need to escalate it up to the Apache `sshd` or even Java 10 :cry:

build.gradle
garyrussell: Perhaps `+=` works here?

artembilan: No, I tried :cry:

build.gradle
```diff
@@ -694,6 +701,10 @@ project('spring-integration-ws') {
 
 		testRuntime "com.sun.mail:mailapi:$javaxMailVersion"
 		testRuntime "com.sun.mail:javax.mail:$javaxMailVersion"
+
+	}
+	if (JavaVersion.current() != JavaVersion.VERSION_1_8) {
+		[compileTestJava]*.options*.compilerArgs = ["$xLintArg", '--add-modules', 'java.xml.ws']
```
artembilan: See some further instructions about new Java version dependencies configuration: https://spring.io/blog/2018/09/19/spring-web-services-3-0-4-2-4-3-released

artembilan:
Superseded with https://github.com/spring-projects/spring-integration/pull/3310 to support Java 14 already.

PR #2457, opened 2018-05-31
INT-4477: Add getRole() to LeaderInitiator Context

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4477

When running multiple roles, it is useful to be able to determine the
role for a particular context, e.g. in an event.

__cherry-pick to 5.0.x__


artembilan (approved):
LGTM, but shouldn't we add some not to the Docs as well?
Thanks

artembilan:

PR #2458, opened 2018-05-31
Fix LambdaMessageProcessor for Map payload

artembilan:
JIRA: https://jira.spring.io/browse/INT-4478

When we use a `GenericHandler` and an incoming payload is a `Map`, we
copy it into both arguments into the `Object` for the payload
and `Map` for the headers.
This way we lose headers in the target lambda

* Check a size of arguments on the target lambda and don't set a
payload into the `Map` argument if we have more than 1 arguments

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
@@ -100,7 +100,7 @@ public class LambdaMessageProcessor implements MessageProcessor<Object>, BeanFac
 				args[i] = message;
 			}
 			else if (Map.class.isAssignableFrom(parameterType)) {
-				if (message.getPayload() instanceof Map) {
+				if (message.getPayload() instanceof Map && this.parameterTypes.length == 1) {
 					args[i] = message.getPayload();
 				}
 				else {
--- a/spring-integration-webflux/src/test/java/org/springframework/integration/webflux/dsl/WebFluxDslTests.java
+++ b/spring-integration-webflux/src/test/java/org/springframework/integration/webflux/dsl/WebFluxDslTests.java
@@ -299,8 +299,10 @@ public class WebFluxDslTests {
 		public IntegrationFlow sseFlow() {
 			return IntegrationFlows
 					.from(WebFlux.inboundGateway("/sse")
-							.requestMapping(m -> m.produces(MediaType.TEXT_EVENT_STREAM_VALUE)))
-					.handle((p, h) -> Flux.just("foo", "bar", "baz"))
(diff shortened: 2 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
```diff
@@ -100,7 +100,7 @@ public Object processMessage(Message<?> message) {
 				args[i] = message;
 			}
 			else if (Map.class.isAssignableFrom(parameterType)) {
-				if (message.getPayload() instanceof Map) {
+				if (message.getPayload() instanceof Map && this.parameterTypes.length == 1) {
```
garyrussell: A bit contrived, perhaps, but I wonder if we need to protect against user code such as this...

```
interface MyFunc {

	Object handle(Map<?, ?> payload, Map<String, Object> headers);

}
```

and

```java
MyFunc func = (p, h) -> p;
LambdaMessageProcessor mp = new LambdaMessageProcessor(func, null);
Message<Map<?, ?>> m = new GenericMessage<>(Collections.singletonMap("foo", "bar"),
		Collections.singletonMap("baz", "qux"));
assertThat(((Map<?, ?>) mp.processMessage(m)).get("foo")).isEqualTo("bar");
```
I suppose we could assert that the first parameter type is `Object` (or emit a WARN log).

artembilan: Well, the sample you show is really classical ambiguity: fully unclear which argument to use for `payload` and which for `headers`. Also the problem that we can't use parameter annotations to distinguish what is for what.
How about this: if more than one argument, use `Map` payload only for those who are not a `Map`? Otherwise - only headers.

garyrussell: That's a good compromise; but I am ok with the status quo if you like - I was just trying to think of examples where the new logic might break some existing code.

artembilan: Well, as far as I remember, originally this `LambdaMessageProcessor` wasn't a `public` class, so what we have here now is fully enough for those use-case where we suggest only our lambdas.

The situation is like the current behavior is really wrong and fully lose headers in the lambda and there is no so obvious way to overcome the problem.

Let's merge it as it states and see if there is some complains on the matter! Just for curiosity what is the reason to use this `LambdaMessageProcessor` outside of the Java DSL :smile:

garyrussell: I concur - just that it's public - but I doubt anyone would use it natively.

Merging.

PR #2459, opened 2018-05-31
INT-4471: PubSubChannel: Add errorHandler warn

artembilan:
JIRA: https://jira.spring.io/browse/INT-4471

* When an `Executor` is not provided, log warn that the provided
`ErrorHandler` is ignored.


spring-integration-core/src/main/java/org/springframework/integration/channel/PublishSubscribeChannel.java
```diff
@@ -149,6 +146,11 @@ public final void onInit() throws Exception {
 			getDispatcher().setApplySequence(this.applySequence);
 			getDispatcher().setMinSubscribers(this.minSubscribers);
 		}
+		else if (this.errorHandler != null) {
+			logger.warn("When 'executor' is not provided the 'errorHandler' is ignored and " +
+					"exceptions are thrown directly within the sending Thread");
+		}
```
garyrussell: Perhaps include the `componentName` in the message?

garyrussell:
And cherry-picked to 5.0.x, 4.3.x

PR #2460, opened 2018-05-31
INT-4474: ConsumerEndpointFactoryBean: Fix phase

artembilan:
JIRA: https://jira.spring.io/browse/INT-4474

* Populate proper `Integer.MIN_VALUE` phase for non-`PollingConsumer`s
in the `ConsumerEndpointFactoryBean`


garyrussell:
Do you really think we should backport this? It's really a behavior change (and nobody is actually complaining about it).

artembilan:
Fine with me.
Fixing JIRA...

PR #2462, opened 2018-05-31
INT-4476: Fall back for ID and Timestamp Headers

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4476

When mapping outbound headers, if the `AmqpHeaders` `ID` and `Timestamp` headers
are not present, fall back to mapping the `MessageHeaders` variants (if present).

Also fix some PDF overflows.


artembilan (approved):
LGTM.
Will merge after fixing compatibility with the latest SF, according its upgrade to Reactor Netty 0.8:

https://build.spring.io/browse/INT-MASTER-1059/
https://github.com/spring-projects/spring-framework/pull/1845

PR #2463, opened 2018-06-01
INT-4387: Fix incorrect contentType mapping

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4387

Don't put a copy in the message properties headers map.


src/reference/asciidoc/whats-new.adoc
```diff
@@ -55,3 +55,6 @@ See <<java-dsl-flows>> for more information.
 
 `ID` and `Timestamp` header mapping changes in the `DefaultAmqpHeaderMapper`.
 See the note near the bottom of <<amqp-message-headers>> for more information.
+
+The `contentType` header is no longer incorrectly mapped as an entry in the general headers map.
```
artembilan: I think this should go to the Migration Guide not as well.

Merging BTW...

garyrussell: Wiki updated.

PR #2469, opened 2018-06-05
INT-4482: AMQP: Fix Double ErrorMessage

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4482

The outer try/catch sends an `ErrorMessage` for all exceptions; it should
only do so for `MessageConversionException`. Integration flow exceptions
will have been already handled by `MessageProducerSupport`.

Also, populate the raw message header consistently - previously it only
was populated for flow exceptions. Although the LEFE contains the raw
message, it should be in the `ErrorMessage` header for consistency.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/StringObjectMapBuilder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/StringObjectMapBuilder.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/StringObjectMapBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+package org.springframework.integration.support;
+
+/**
+ * A map builder creating a map with String keys and values.
+ *
+ * @author Gary Russell
+ *
+ * @since 4.3.17
+ */
+public class StringObjectMapBuilder extends MapBuilder<StringObjectMapBuilder, String, Object> {
```
artembilan: Yeah... We can't do this for the `4.3.x` - the `MapBuilder` has appeared here since `5.0`.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/StringObjectMapBuilder.java
garyrussell: np; the tests don't exist in 4.3 anyway.

artembilan:
Back ported  to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/37c92a128b41803f91052596a34a1f730080df54 after fixing conflicts.

PR #2470, opened 2018-06-06
GH-2467: JdbcLockRegistry should retry on DeadlockLoserDataAccessException

### LINKED ISSUE #2467: JdbcLockRegistry should retry on DeadlockLoserDataAccessException
I am using `LockRegistryLeaderInitiator`  with `JdbcLockRegistry`. It worked fine on 5.6.36 but after the migration to 5.7.19 I noticed flood of "leader granted/revoked" events when more than on instance of my service is running. After some debugging I noticed it is caused by `DeadlockLoserDataAccessException` not handled as "retry" in  `JdbcLockRegistry`.

MySql 5.7.15 introduced setting `innodb_deadlock_detect` (enabled by default, see https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_deadlock_detect).

When this setting is enabled exception will be thrown from MySql JDBC driver when deadlock is detected while accessing table row  (vendorCode=1213, SQLState=40001, message: Deadlock found when trying to get lock; try restarting transaction). It is then wrapped to spring exception `DeadlockLoserDataAccessException`. 

`JdbcLockRegistry` does not handle `DeadlockLoserDataAccessException` with "try again" action (as it does for `CannotSerializeTransactionException`, `TransactionTimedOutException`, `QueryTimeoutException`).

see https://github.com/spring-projects/spring-integration/blob/a2f3dbe5fb558359cc839a81b12a6161cd96cabf/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java#L218-L230

Spring Boot 1.5.13.RELEASE
Spring Integration JDBC 4.3.16.RELEASE
MySQL 5.6.36/5.7.19

brempusz:
Fixes #2467

MySQL 5.7.15 introduced setting `innodb_deadlock_detect` (enabled by
default). As a result MySQL JDBC driver throws
`DeadlockLoserDataAccessException` when deadlock is detected.
`JdbcLockRegistry` doesn't handle it causing lock to be lost.

* Retry `doLock()` on data access deadlock instead of loosing the lock

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -26,15 +26,12 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.springframework.dao.CannotAcquireLockException;
-import org.springframework.dao.CannotSerializeTransactionException;
 import org.springframework.dao.DataAccessResourceFailureException;
-import org.springframework.dao.DeadlockLoserDataAccessException;
-import org.springframework.dao.QueryTimeoutException;
+import org.springframework.dao.TransientDataAccessException;
 import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.integration.support.locks.LockRegistry;
 import org.springframework.integration.util.UUIDConverter;
-import org.springframework.transaction.TransactionTimedOutException;
 import org.springframework.util.Assert;
 
 /**
@@ -115,8 +112,7 @@ public class JdbcLockRegistry implements ExpirableLockRegistry {
 					}
 					break;
 				}
-				catch (CannotSerializeTransactionException | TransactionTimedOutException | QueryTimeoutException |
-						DeadlockLoserDataAccessException e) {
+				catch (TransientDataAccessException e) {
 					// try again
 				}
 				catch (InterruptedException e) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@brempusz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2470)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2470) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@brempusz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2470)!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -113,7 +115,8 @@ public void lock() {
 					}
 					break;
 				}
-				catch (CannotSerializeTransactionException | TransactionTimedOutException | QueryTimeoutException e) {
+				catch (CannotSerializeTransactionException | TransactionTimedOutException | QueryTimeoutException |
```
artembilan: The `CannotSerializeTransactionException` and `DeadlockLoserDataAccessException` are both inheritors of the `PessimisticLockingFailureException`, so maybe we can go the way for their super class?

artembilan (changes requested):
I think we don't need a separate PR in `4.3.x` for such a simple fix.
We will figure out during merging.

Thank you!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
brempusz: After looking at the exception hierarchy - maybe  we can use `TransientDataAccessException`? It does make sense: any transient error should cause retry.

artembilan:
... and cherry-picked to `5.0.x` and `4.3.x`.

@brempusz ,

Looking forward for more!

PR #2472, opened 2018-06-06
GH-2471: MQTT: Fix Thread Leak

### LINKED ISSUE #2471: Reconnection / retry error with Paho MQTT 1.2 leads to ever increasing thread usage
As a follow up to my <a href="https://twitter.com/rotnroll666/status/1003935166484635649">tweet</a>.

Find attached a simple Spring Integration MQTT application ([mqtt-paho-client-threading-bug.zip](https://github.com/spring-projects/spring-integration/files/2075768/mqtt-paho-client-threading-bug.zip)). It's a Spring Boot app, just run

```
./mvnw clean spring-boot:run
```

You'll find the app trying to reach a MQTT broker on `tcp://localhost:45678` (I chose this port to make sure there is no broker answering).

The app will fail, print the MQTT Paho exception and continue.

I'll added a scheduled task that prints the number of live threads. As you see they'll increse by 2 threads after each failed connection retry.

This happens during startup as well as during runtime, for example if the broker goes away in between.

A downgrade to Paho 1.1.1 as suggested in https://github.com/eclipse/paho.mqtt.java/issues/402 respectivly https://github.com/SebaDro/SOS/commit/8b9f09eacdc842655274a9c94a64486067db3f5b helps, for example like this:

```
    <dependency>
            <artifactId>spring-integration-mqtt</artifactId>
            <exclusions>
                <exclusion>
                    <groupId>org.eclipse.paho</groupId>
                    <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
        <dependency>
            <groupId>org.eclipse.paho</groupId>
            <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
            <version>1.1.1</version>
            <scope>compile</scope>
        </dependency>
```

With Paho 1.1.1 the number of threads stays constant.

garyrussell:
Fixes https://github.com/spring-projects/spring-integration/issues/2471

Call `close()` on the client whenever the connection is lost or can't be
established, to release resources in the client.

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


artembilan:
Back-ported to `4.3.x` after resolving some conflicts in test class.

PR #2475, opened 2018-06-11
INT-4485: Don't register bean twice for DSL Specs

artembilan:
JIRA: https://jira.spring.io/browse/INT-4485

The `IntegrationFlowBeanPostProcessor` performs extra bean registration
for the `IntegrationComponentSpec.get()` result.
Essentially it is going to be the same object in the end, but during
bean registration phase we end up with names conflict.

* Remove an explicit `registerComponent()` for the
`IntegrationComponentSpec.get()`
* Modify `CorrelationHandlerTests` for all possible usage for the
`MessageChannelSpec`, which is essentially a `FactoryBean`

**Cherry-pick to 5.0.x**


garyrussell:
And cherry-picked as 08ebc6684ffbf7fffdc0de58ca6ed77c1a3b9042

Issue #2473, opened 2018-06-08
Poller throws java.lang.IllegalArgumentException: Comparison method violates its general contract!


despondency:
The poller seems to be throwing this exception on consuming RSS feed.

`org.springframework.messaging.MessagingException: nested exception is java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at org.springframework.integration.endpoint.AbstractPollingEndpoint$Poller.lambda$run$0(AbstractPollingEndpoint.java:395) ~[spring-integration-core-5.0.5.RELEASE.jar:5.0.5.RELEASE]
	at org.springframework.integration.util.ErrorHandlingTaskExecutor.lambda$execute$0(ErrorHandlingTaskExecutor.java:53) [spring-integration-core-5.0.5.RELEASE.jar:5.0.5.RELEASE]
	at org.springframework.core.task.SyncTaskExecutor.execute(SyncTaskExecutor.java:50) ~[spring-core-5.0.6.RELEASE.jar:5.0.6.RELEASE]
	at org.springframework.integration.util.ErrorHandlingTaskExecutor.execute(ErrorHandlingTaskExecutor.java:51) [spring-integration-core-5.0.5.RELEASE.jar:5.0.5.RELEASE]
	at org.springframework.integration.endpoint.AbstractPollingEndpoint$Poller.run(AbstractPollingEndpoint.java:372) ~[spring-integration-core-5.0.5.RELEASE.jar:5.0.5.RELEASE]
	at org.springframework.scheduling.support.DelegatingErrorHandlingRunnable.run(DelegatingErrorHandlingRunnable.java:54) ~[spring-context-5.0.6.RELEASE.jar:5.0.6.RELEASE]
	at org.springframework.scheduling.concurrent.ReschedulingRunnable.run(ReschedulingRunnable.java:93) ~[spring-context-5.0.6.RELEASE.jar:5.0.6.RELEASE]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_162]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[na:1.8.0_162]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180) ~[na:1.8.0_162]
[...]

PR #2477, opened 2018-06-12
INT-4484: Fix SyndEntryPublishedDateComparator

artembilan:
JIRA: https://jira.spring.io/browse/INT-4484

Fix the `SyndEntryPublishedDateComparator` to move entries without a
date into the end of the sorted list

Fixes spring-projects/spring-integration#2473

### DIFF
--- a/spring-integration-feed/src/main/java/org/springframework/integration/feed/inbound/FeedEntryMessageSource.java
+++ b/spring-integration-feed/src/main/java/org/springframework/integration/feed/inbound/FeedEntryMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


artembilan:
The issue like this might be relevant: https://jira.spring.io/browse/INT-1810...

artembilan:
I believe this fixes spring-projects/spring-integration#2473 as well.
The `DeadlockLoserDataAccessException` is fully not related and the reported doesn't follow our recommendations about `.transactional()` for the poller. Also need to keep in mind that without TX on the poller and with the `fixedRate()` polling we definitely have concurrent INSERTs which may lead to such a dead lock.

garyrussell:
>I believe this fixes spring-projects/spring-integration#2473 as well.

?? I created this JIRA for that issue

garyrussell (changes requested):
No test case?

artembilan:
@garyrussell ,

Anything wrong with this PR? 
It is marked for the back-port, so we can't release `5.0.6` and `4.3.17`.

Thanks

garyrussell:
And cherry-picked as 058ae6f4c040f31617b053af9476f9aa85a02f01 and b776b14b51cc82aff4bf42906875d4546e4d55ce.

PR #2482, opened 2018-06-14
INT-2260: JdbcPollingChA: rename prop to maxRows

artembilan:
JIRA: https://jira.spring.io/browse/INT-2260

Having a feedback about confusing with the `max-rows-per-poll` property
name and its responsibility it would be better do not mention `per-poll`
at all

* Deprecate `max-rows-per-poll` in favor of new `max-rows`
* Some code style polishing, tests improvements
* Docs polishing on the matter

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -44,14 +44,13 @@ public class JdbcOutboundGateway extends AbstractReplyProducingMessageHandler im
 
 	private final JdbcPollingChannelAdapter poller;
 
-	private volatile SqlParameterSourceFactory sqlParameterSourceFactory =
-			new ExpressionEvaluatingSqlParameterSourceFactory();
+	private SqlParameterSourceFactory sqlParameterSourceFactory = new ExpressionEvaluatingSqlParameterSourceFactory();
 
-	private volatile boolean sqlParameterSourceFactorySet;
+	private boolean sqlParameterSourceFactorySet;
 
-	private volatile boolean keysGenerated;
+	private boolean keysGenerated;
 
-	private volatile Integer maxRowsPerPoll;
+	private Integer maxRows;
 
 	public JdbcOutboundGateway(DataSource dataSource, String updateQuery) {
 		this(new JdbcTemplate(dataSource), updateQuery, null);
@@ -66,16 +65,16 @@ public class JdbcOutboundGateway extends AbstractReplyProducingMessageHandler im
 	}
 
 	public JdbcOutboundGateway(JdbcOperations jdbcOperations, String updateQuery, String selectQuery) {
-
 		Assert.notNull(jdbcOperations, "'jdbcOperations' must not be null.");
 
 		if (!StringUtils.hasText(updateQuery) && !StringUtils.hasText(selectQuery)) {
-			throw new IllegalArgumentException("The 'updateQuery' and the 'selectQuery' must not both be null or empty.");
+			throw new IllegalArgumentException(
+					"The 'updateQuery' and the 'selectQuery' must not both be null or empty.");
 		}
 
 		if (StringUtils.hasText(selectQuery)) {
 			this.poller = new JdbcPollingChannelAdapter(jdbcOperations, selectQuery);
-			this.poller.setMaxRowsPerPoll(1);
+			this.poller.setMaxRows(1);
 		}
 		else {
 			this.poller = null;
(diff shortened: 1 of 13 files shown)


artembilan:
@Buzzardo,

FYI.

BTW, not sure how you follow our changes during your work with our docs.
Anyway thank you for your effort and all the feedback your share with us!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
```diff
@@ -98,10 +97,54 @@ public JdbcOutboundGateway(JdbcOperations jdbcOperations, String updateQuery, St
...
+	 * The maximum number of rows to query.
+	 * The value is ultimately set on the underlying {@link JdbcPollingChannelAdapter}.
+	 * If not specified this value will default to {@code 1}.
+	 * This parameter is only applicable if a selectQuery was provided. Null values
+	 * are not permitted.
+	 * @param maxRows the number of rows to select. Must not be null.
+	 * @since 5.1
+	 * @see JdbcPollingChannelAdapter#setMaxRows(int)
+	 */
+	public void setMaxRows(Integer maxRows) {
```
garyrussell: Why not just `int` here? Null check not needed then (but restore the `Assert` in the deprecated method).

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/config/JdbcPollingChannelAdapterParser.java
```diff
@@ -59,24 +61,24 @@ protected BeanMetadataElement parseSource(Element element, ParserContext parserC
...
 		builder.addConstructorArgValue(query);
+
 		IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element, "row-mapper");
 		IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element, "update-sql-parameter-source-factory");
 		IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element, "select-sql-parameter-source");
 		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "max-rows-per-poll");
-		if (update != null) {
-			builder.addPropertyValue("updateSql", update);
-		}
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "max-rows");
```
garyrussell: We should either disallow specifying both, or change the schema docs to explain that if both are supplied, `max-rows` wins.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/MessageGroupQueueTests.java
```diff
@@ -44,7 +43,7 @@
  */
 public class MessageGroupQueueTests {
 
-	@ClassRule
+//	@ClassRule
```
garyrussell: ?? If it's no longer long, then just remove the rule - or is this just left over from your testing?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
```diff
@@ -119,19 +117,29 @@ public void setSelectSqlParameterSource(SqlParameterSource sqlQueryParameterSour
...
-		this.maxRowsPerPoll = maxRows;
+		setMaxRows(maxRows);
+	}
+
+	/**
+	 * The maximum number of rows to query. Default is zero - select all records.
+	 * @param maxRows the max rows to set
+	 * @since 5.1
+	 */
+	public void setMaxRows(int maxRows) {
```
garyrussell: We should clarify here, and in the docs, that this mechanism is crude - it simply discards the excess results - it is generally better to use the DBMS custom SQL to restrict the number of rows returned, if supported (e.g. `LIMIT n` in MySQL).

garyrussell (changes requested):
Just a couple of minor issues.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
artembilan: Well, there is a logic in the `onInit()`:
```
if (this.maxRows != null) {
		Assert.notNull(this.poller, "If you want to set 'maxRows', then you must provide a 'selectQuery'.");
	this.poller.setMaxRows(this.maxRows);
}
```
Should we change this to something like `> 0` then?
Also pay attention to the logic in the ctor:
```
if (StringUtils.hasText(selectQuery)) {
	this.poller = new JdbcPollingChannelAdapter(jdbcOperations, selectQuery);
	this.poller.setMaxRows(1);
}
```
The point is that with the regular poller we have an unlimited request, but with the gateway we are limiting to only the `1` by default. Any custom value is managed by the `Integer` state.

garyrussell: Agreed; I missed that; the current state is fine.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
artembilan: Yeah... You know I found this in the `JdbcTemplate`:
```
	/**
	 * Set the maximum number of rows for this JdbcTemplate. This is important for
	 * processing subsets of large result sets, avoiding to read and hold the entire
	 * result set in the database or in the JDBC driver if we're never interested in
	 * the entire result in the first place (for example, when performing searches
	 * that might return a large number of matches).
	 * <p>Default is -1, indicating to use the JDBC driver's default configuration
	 * (i.e. to not pass a specific max rows setting on to the driver).
	 * <p>Note: As of 4.3, negative values other than -1 will get passed on to the
	 * driver, in sync with {@link #setFetchSize}'s support for special MySQL values.
	 * @see java.sql.Statement#setMaxRows
	 */
	public void setMaxRows(int maxRows) {
```

From here I'm even starting to think that we don't need our property at all.
Since we have that `Statement` level setting there - see `JdbcTemplate.applyStatementSettings(Statement stmt)`.

Plus of course you are right, that it would be much better to deffer the `MaxRows` to the target vendor `SELECT`.

So, WDYT now ?

garyrussell: Do you mean eliminate it altogether and tell them to appropriately configure a `JdbcTemplate` to limit the rows (or use a vendor SQL extension).

Of course, they'd have to use the 4th CTOR (on the ob gateway) for that.

SGTM.

artembilan: Right. Just deprecate the option in favor of such an option on the `JdbcTemplate` to provide.
And also recommend to rely on the  native JDBC vendor limiting support available through SELECT options.

Makes sense ?

garyrussell: yes

artembilan: Hm... You know during fixing that I've changed my mind: the current option is per-Statement and doesn't effect `JdbcTempalate` globally. I think we need to keep the current behavior renaming the property. Also I will consider some really per-Statement configuration for optimization.

The limiting with the `SELECT` is still there for documentation...

garyrussell (approved):

garyrussell:
Merged as 32030c3233e6cf37bc5b0550b600e40cae29dd16 with minor doc polishing.

PR #2485, opened 2018-06-21
INT-4487: Allow to reconfigure ARPMH.adviceChain

artembilan:
JIRA: https://jira.spring.io/browse/INT-4487

Currently the `AbstractReplyProducingMessageHandler` applies advices
only in the `afterPropertiesSet()` phase.
When we use bean references for `MessageHandler`s in Java DSL, the
`.advice()` provided in the flow definition is not applied to the
`AbstractReplyProducingMessageHandler`.

* For consistency with the endpoint configured by the DSL definition,
override and reconfigure advices in the `AbstractReplyProducingMessageHandler`
bean reference during DSL parsing and processing.


garyrussell:
Merged as e6a35c4a6077753ec095dd92b57a3ea6ed8c7f24 after squash and minor doc polishing.

PR #2488, opened 2018-06-22
INT-4452: expire immediately when group timeout<0

artembilan:
JIRA: https://jira.spring.io/browse/INT-4452

This is pretty typical in practice to get a `groupTimeout` to be
evaluated to a negative value: some business decisions, sensitive
calculations and so on.

* Treat any non-positive `groupTimeout` as a signal to expire group
immediately without scheduling.
Only `null` is considered as a signal do nothing for the current message
* Polishing for some tests for better performance - saves some execution
time


garyrussell:
Merged as 13510df575214968cd721b46ceae2b5f258a2d60

artembilan:
Added  Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.0-to-5.1-Migration-Guide

PR #2489, opened 2018-06-25
INT-4491: (S)FTP inbound rotate dirs/servers

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4491

Add Rotating Server/Directory Polling Advice.

**cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
@@ -1,0 +1,215 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.file.remote.session;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.integration.aop.AbstractMessageSourceAdvice;
+import org.springframework.integration.core.MessageSource;
+import org.springframework.integration.file.remote.AbstractRemoteFileStreamingMessageSource;
+import org.springframework.integration.file.remote.synchronizer.AbstractInboundFileSynchronizingMessageSource;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * A smart poller advice that rotates across multiple remote servers/directories.
+ *
+ * @author Gary Russell
+ * @since 5.0.7
+ *
+ */
+public class RotatingServerAdvice extends AbstractMessageSourceAdvice {
+
+	private final RotationPolicy rotationPolicy;
+
+	/**
+	 * Create an instance that rotates to the next server/directory if no message is
+	 * received.
+	 * @param factory the {@link DelegatingSessionFactory}.
+	 * @param keyDirectories a list of {@link KeyDirectory}.
+	 */
+	public RotatingServerAdvice(DelegatingSessionFactory<?> factory, List<KeyDirectory> keyDirectories) {
+		this(factory, keyDirectories, false);
+	}
+
+	/**
+	 * Create an instance that rotates to the next server/directory depending on the fair
+	 * argument.
+	 * @param factory the {@link DelegatingSessionFactory}.
+	 * @param keyDirectories a list of {@link KeyDirectory}.
+	 * @param fair true to rotate on every poll, false to rotate when no message is received.
+	 */
+	public RotatingServerAdvice(DelegatingSessionFactory<?> factory, List<KeyDirectory> keyDirectories, boolean fair) {
+		this(new StandardRotationPolicy(factory, keyDirectories, fair));
+	}
+
+	public RotatingServerAdvice(RotationPolicy rotationPolicy) {
+		this.rotationPolicy = rotationPolicy;
+	}
+
+	@Override
+	public boolean beforeReceive(MessageSource<?> source) {
+		this.rotationPolicy.beforeReceive(source);
+		return true;
+	}
+
+	@Override
+	public Message<?> afterReceive(Message<?> result, MessageSource<?> source) {
+		this.rotationPolicy.afterReceive(result != null, source);
+		return result;
+	}
+
+	/**
+	 * Implementations can reconfigure the message source before and/or after
+	 * a poll.
+	 * @since 5.0.7
+	 *
+	 */
+	public interface RotationPolicy {
+
+		/**
+		 * Invoked before the message source receive() method.
+		 * @param source the message source.
+		 */
+		void beforeReceive(MessageSource<?> source);
+
+		/**
+		 * Invoked after the message source receive() method.
+		 * @param messageReceived true if a message was received.
+		 * @param source the message source.
+		 */
+		void afterReceive(boolean messageReceived, MessageSource<?> source);
+
+	}
+
+	/**
+	 * Standard rotation policy; iterates over key/directory pairs; when the end
+	 * is reached, starts again at the beginning. If the fair option is true
+	 * the rotation occurs on every poll, regardless of result. Otherwise rotation
+	 * occurs when the current pair returns no message.
+	 * @since 5.0.7
+	 *
+	 */
+	private static class StandardRotationPolicy implements RotationPolicy {
+
+		private final Log logger = LogFactory.getLog(getClass());
+
+		private final DelegatingSessionFactory<?> factory;
+
+		private final List<KeyDirectory> keyDirectories = new ArrayList<>();
+
+		private final boolean fair;
+
+		private volatile Iterator<KeyDirectory> iterator;
+
+		private volatile KeyDirectory current;
+
+		private volatile boolean initialized;
+
+		StandardRotationPolicy(DelegatingSessionFactory<?> factory, List<KeyDirectory> keyDirectories,
+				boolean fair) {
+			Assert.notNull(factory, "factory cannot be null");
+			Assert.notNull(keyDirectories, "keyDirectories cannot be null");
+			Assert.isTrue(keyDirectories.size() > 0, "At least one KeyDirectory is required");
+			this.factory = factory;
+			this.keyDirectories.addAll(keyDirectories);
+			this.fair = fair;
+			this.iterator = this.keyDirectories.iterator();
+		}
+
+		@Override
+		public void beforeReceive(MessageSource<?> source) {
+			if (this.fair || !this.initialized) {
+				configureSource(source);
+				this.initialized = true;
+			}
+			if (this.logger.isTraceEnabled()) {
+				this.logger.trace("Next poll is for " + this.current);
+			}
+			this.factory.setThreadKey(this.current.getKey());
+		}
+
+		@Override
+		public void afterReceive(boolean messageReceived, MessageSource<?> source) {
+			if (this.logger.isTraceEnabled()) {
+				this.logger.trace("Poll resulted in message: " + messageReceived);
+			}
+			this.factory.clearThreadKey();
+			if (!this.fair && !messageReceived) {
+				configureSource(source);
+			}
+		}
+
+		private void configureSource(MessageSource<?> source) {
+			Assert.isTrue(source instanceof AbstractInboundFileSynchronizingMessageSource
+					|| source instanceof AbstractRemoteFileStreamingMessageSource,
+				"source must be an AbstractInboundFileSynchronizingMessageSource or a "
+				+ "AbstractRemoteFileStreamingMessageSource");
+			if (!this.iterator.hasNext()) {
+				this.iterator = this.keyDirectories.iterator();
+			}
+			this.current = this.iterator.next();
+			if (source instanceof AbstractRemoteFileStreamingMessageSource) {
+				((AbstractRemoteFileStreamingMessageSource<?>) source).setRemoteDirectory(this.current.getDirectory());
+			}
+			else {
+				((AbstractInboundFileSynchronizingMessageSource<?>) source).getSynchronizer()
+					.setRemoteDirectory(this.current.getDirectory());
+			}
+		}
+
+	}
+
+	/**
+	 * A {@link DelegatingSessionFactory} key/directory pair.
+	 * @since 5.0.7
+	 *
+	 */
+	public static class KeyDirectory {
+
+		private final String key;
+
+		private final String directory;
+
+		public KeyDirectory(String key, String directory) {
+			Assert.notNull(key, "key cannot be null");
+			Assert.notNull(directory, "directory cannot be null");
+			this.key = key;
+			this.directory = directory;
+		}
+
+		public String getKey() {
+			return this.key;
+		}
+
+		public String getDirectory() {
+			return this.directory;
+		}
+
+		@Override
+		public String toString() {
+			return "KeyDirectory [key=" + this.key + ", directory=" + this.directory + "]";
+		}
+
+	}
+
+}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
@@ -16,7 +16,6 @@
 
 package org.springframework.integration.file.remote.session;
 
-import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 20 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
```diff
@@ -0,0 +1,215 @@
...
+	 * argument.
+	 * @param factory the {@link DelegatingSessionFactory}.
+	 * @param keyDirectories a list of {@link KeyDirectory}.
+	 * @param fair true to rotate on every poll, false to rotate when no message is received.
+	 */
+	public RotatingServerAdvice(DelegatingSessionFactory<?> factory, List<KeyDirectory> keyDirectories, boolean fair) {
+		this(new StandardRotationPolicy(factory, keyDirectories, fair));
+	}
+
+	public RotatingServerAdvice(RotationPolicy rotationPolicy) {
```
artembilan: Also some JavaDocs? 
And `Assert.notNull(rotationPolicy, ...)` ?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
```diff
@@ -0,0 +1,215 @@
...
+			if (this.logger.isTraceEnabled()) {
+				this.logger.trace("Next poll is for " + this.current);
+			}
+			this.factory.setThreadKey(this.current.getKey());
+		}
+
+		@Override
+		public void afterReceive(boolean messageReceived, MessageSource<?> source) {
+			if (this.logger.isTraceEnabled()) {
+				this.logger.trace("Poll resulted in message: " + messageReceived);
```
artembilan: Feels like something is missed here else...

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
```diff
@@ -0,0 +1,215 @@
...
+		}
+
+	}
+
+	/**
+	 * A {@link DelegatingSessionFactory} key/directory pair.
+	 * @since 5.0.7
+	 *
+	 */
+	public static class KeyDirectory {
```
artembilan: The `equals()` and `hashCode()` ?

Can't we live with the `Map.Entry` instead BTW ?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -436,9 +440,19 @@ else if (this.logger.isWarnEnabled()) {
...
 
-	private String generateLocalFileName(String remoteFileName) {
+	private String generateLocalFileName(String remoteDirectoryPath, String remoteFileName) {
 		if (this.localFilenameGeneratorExpression != null) {
-			return this.localFilenameGeneratorExpression.getValue(this.evaluationContext, remoteFileName, String.class);
+			if (!(this.localFilenameGeneratorExpression instanceof LiteralExpression) &&
+					this.localFilenameGeneratorExpression.getExpressionString().contains("#remoteDirectory")) {
+				EvaluationContext context = IntegrationContextUtils.getEvaluationContext(this.beanFactory);
+				context.setVariable("remoteDirectory", remoteDirectoryPath);
+				return this.localFilenameGeneratorExpression.getValue(context, remoteFileName,
```
artembilan: Looks like this call can be only once. What we definitely need here is to reconfigure a `context` variable according the `#remoteDirectory` presence. But, yeah... the solution is still cool. 👍

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/SessionFactoryMapBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+package org.springframework.integration.file.remote.session;
+
+import org.springframework.integration.support.MapBuilder;
+
+/**
+ * @author Gary Russell
+ * @since 5.0.7
+ *
+ */
+public class SessionFactoryMapBuilder<T> extends MapBuilder<SessionFactoryMapBuilder<T>, Object, SessionFactory<T>> {
```
artembilan: We need to think about retiring this builder in favor of `Map.of()`: https://stackoverflow.com/questions/46601959/what-is-the-use-of-map-ofentries-instead-of-map-of.

But, yeah... we still can live with that for a while! 😄 

Please, add some simple JavaDoc on the matter though

artembilan (changes requested):
I wonder if this fix is related somehow to the https://jira.spring.io/browse/INT-3205.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/SessionFactoryMapBuilder.java
garyrussell: `Map.of()` is JDK 9 - I don't think we'll be there for a long time.

garyrussell:
>I wonder if this fix is related somehow

It's somewhat related, but not closely, since that is still about outbound SFs based on `Message<?>` - i.e. creating an SF dynamically (host/port etc) based on message properties.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -164,6 +169,7 @@ public void setTemporaryFileSuffix(String temporaryFileSuffix) {
 	 */
 	public void setRemoteDirectory(String remoteDirectory) {
 		this.remoteDirectoryExpression = new LiteralExpression(remoteDirectory);
+		evalueRemoteDirectory();
```
artembilan: Is it really safe to do this?
Or are we going to document that `RotatingServerAdvice` should not be used together with the `fixedRate` or some tight cron?

src/reference/asciidoc/ftp.adoc
```diff
@@ -620,6 +622,89 @@ public class FtpJavaApplication {
...
+The advice configuration consists of a list of `RotatingServerAdvice.KeyDirectory` objects.
+
+.Example
+[source, java]
+----
+@Bean
+public RotatingServerAdvice advice() {
+    List<KeyDirectory> keyDirectories = new ArrayList<>();
+    keyDirectories.add(new KeyDirectory("one", "foo"));
+    keyDirectories.add(new KeyDirectory("one", "bar"));
```
artembilan: Oh! I see why `Map.Entry` is not good for us: the `key` can be the same.
So, well maybe your original proposition was fine. Together with the natural `equals()` and `hashCode()` to rely on.

artembilan (changes requested):
That's all for the current cycle.
Will take a look closer tomorrow.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
garyrussell: This is actually a problem for all smart pollers - will add a warn log and docs.

artembilan: To close the topic I meant this yesterday:
````java
EvaluationContext context = this.evaluationContext;
if (!(this.localFilenameGeneratorExpression instanceof LiteralExpression) &&
					this.localFilenameGeneratorExpression.getExpressionString().contains("#remoteDirectory")) {
			context = IntegrationContextUtils.getEvaluationContext(this.beanFactory);
			context.setVariable("remoteDirectory", remoteDirectoryPath);
}

return this.localFilenameGeneratorExpression.getValue(context , remoteFileName,	String.class);
````

But now you have done fully different solution and it need its own special review.

For example how about to synchronize in the `RotatingServerAdvice` when we reconfigure a `MessageSource`, so only the last one will apply its state? We are about reconfiguring source anyway, so does it really matter which one wins?

**Review by artembilan — CHANGES_REQUESTED** [review]

File: spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java (hunk 6 above)
Comment (garyrussell) [reply spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java]: Well, if we have a fast poller and fair rotation, the first poller might get the directory from the second rotation. Your comment yesterday made me realize that all advices that can mutate the MS should not have a TE.

Comment (artembilan) [reply spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java]: Well, sorry for confusing: I didn't think to broad in that place...
And actually the problem not only with the `TE`. The low `fixedRate` (including similar cron) may cause similar race condition.

We definitely should document (or even just reject otherwise) a limitation of smart advices for only a single thread. Otherwise we indeed can end up with the stale state when one thread deals with the options from another.

Comment (garyrussell) [reply spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java]: >The low fixedRate (including similar cron) may cause similar race condition.

That is not true; without a TE the next poll won't be scheduled until the current one completes; see 

```java
	/**
	 * Determine the next execution time according to the given trigger context.
	 * @param triggerContext context object encapsulating last execution times
	 * and last completion time
	 * @return the next execution time as defined by the trigger,
	 * or {@code null} if the trigger won't fire anymore
	 */
	@Nullable
	Date nextExecutionTime(TriggerContext triggerContext);
```

Even with `fixedRate`, the actual start of the next poll doesn't happen until this one finishes (fixedRate just means it's based on the start time, not the end time). If the run time is > fixedRate then the next one runs immediately.

artembilan: Yeah... I see what you mean - `ReschedulingRunnable`:
```
public void run() {
	Date actualExecutionTime = new Date();
	super.run();
	Date completionTime = new Date();
	synchronized (this.triggerContextMonitor) {
		Assert.state(this.scheduledExecutionTime != null, "No scheduled execution");
		this.triggerContext.update(this.scheduledExecutionTime, actualExecutionTime, completionTime);
		if (!obtainCurrentFuture().isCancelled()) {
			schedule();
		}
	}
}
```

so, we definitely don't run `run();` until the `super.run();` is done.

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/inbound/RotatingServersTests.java
```diff
@@ -0,0 +1,299 @@
...
+		assertThat(f2.exists()).isTrue();
+		File f3 = new File(tmpDir + File.separator + "fiz" + File.separator + "f3");
+		assertThat(f3.exists()).isTrue();
+		assertThat(f1.delete()).isTrue();
+		assertThat(f2.delete()).isTrue();
+		assertThat(f3.delete()).isTrue();
+		assertThat(f1.getParentFile().delete()).isTrue();
+		assertThat(f2.getParentFile().delete()).isTrue();
+		assertThat(f3.getParentFile().delete()).isTrue();
+		assertThat(ctx.getBean("files", QueueChannel.class).getQueueSize()).isEqualTo(3);
```
artembilan: The test fails for me on the regular basis:
```
org.junit.ComparisonFailure: 
Expected :3
Actual   :4
 <Click to see difference>


	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at org.springframework.integration.ftp.inbound.RotatingServersTests.testVariableLocalDir(RotatingServersTests.java:144)
```

src/reference/asciidoc/ftp.adoc
```diff
@@ -620,6 +622,91 @@ public class FtpJavaApplication {
...
+    keyDirectories.add(new KeyDirectory("two", "baz"));
+    keyDirectories.add(new KeyDirectory("two", "qux"));
+    keyDirectories.add(new KeyDirectory("three", "fiz"));
+    keyDirectories.add(new KeyDirectory("three", "buz"));
+    return new RotatingServerAdvice(delegatingSf(), keyDirectories);
+}
+----
+
+This advice will poll directory `foo` on server `one` until no new files exist then move to directory `bar` and then direcctory `baz` on server `two`, etc.
```
artembilan: Typo: `direcctory`

src/reference/asciidoc/ftp.adoc
```diff
@@ -620,6 +622,91 @@ public class FtpJavaApplication {
...
+[source, java]
+----
+@Bean
+public RotatingServerAdvice advice() {
+    return new RotatingServerAdvice(myRotationPolicy());
+}
+----
+
+The `local-filename-generator-expression` attribute (`localFilenameGeneratorExpression` on the synchronizer) can now contain the `#remoteDirectory` variable.
+This allows files retrieved from different directories to be downloaded to similar directorties locally:
```
artembilan: Typo: `directorties`

src/reference/asciidoc/polling-consumer.adoc
```diff
@@ -162,7 +162,15 @@ It enables you to examine and or reconfigure the source at this time. Returning
...
 This method is called after the `receive()` method; again, you can reconfigure the source, or take any action perhaps depending on the result (which can be `null` if there was no message created by the source).
-You can even return a different message!
+You can even return a different message
+
+.Thread safety
+[IMPORTANT]
+====
+You should not configure the poller with a `TaskExecutor` if an advice mutates the `MessageSource`.
+If an advice mutates the source, such mutations are not thread safe and could cause unexpected results, especially with high frequency pollers.
+Consider using a downstream ExecutorChannel instead of adding an executor to the poller if you need to process poll results concurrently.
```
artembilan: An `ExecutorChannel` should be wrapped to the code.

src/reference/asciidoc/sftp.adoc
```diff
@@ -665,6 +667,91 @@ public class SftpJavaApplication {
...
+    keyDirectories.add(new KeyDirectory("one", "bar"));
+    keyDirectories.add(new KeyDirectory("two", "baz"));
+    keyDirectories.add(new KeyDirectory("two", "qux"));
+    keyDirectories.add(new KeyDirectory("three", "fiz"));
+    keyDirectories.add(new KeyDirectory("three", "buz"));
+    return new RotatingServerAdvice(delegatingSf(), keyDirectories);
+}
+----
+
+This advice will poll directory `foo` on server `one` until no new files exist then move to directory `bar` and then direcctory `baz` on server `two`, etc.
```
artembilan: The same typo

src/reference/asciidoc/sftp.adoc
```diff
@@ -665,6 +667,91 @@ public class SftpJavaApplication {
...
+[source, java]
+----
+@Bean
+public RotatingServerAdvice advice() {
+    return new RotatingServerAdvice(myRotationPolicy());
+}
+----
+
+The `local-filename-generator-expression` attribute (`localFilenameGeneratorExpression` on the synchronizer) can now contain the `#remoteDirectory` variable.
+This allows files retrieved from different directories to be downloaded to similar directorties locally:
```
artembilan: DITTO

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/inbound/RotatingServersTests.java
artembilan: Looks like we need to clean up a `tmpDir` before each test, e.g. using a `recursiveDelete()` from the super `RemoteFileTestSupport`

artembilan (changes requested):

artembilan:
Please, consider to rebase to `master`: https://github.com/spring-projects/spring-integration/commit/21fff59112aa7abedd70a33b6a7a4c59453befa3.

Or I'll do that myself if there is nothing more to object about in your latest changes...

artembilan (approved):
LGTM, merging...

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
```diff
@@ -0,0 +1,252 @@
...
+
+		private final boolean fair;
+
+		private volatile Iterator<KeyDirectory> iterator;
+
+		private volatile KeyDirectory current;
+
+		private volatile boolean initialized;
+
+		StandardRotationPolicy(DelegatingSessionFactory<?> factory, List<KeyDirectory> keyDirectories,
```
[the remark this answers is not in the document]
artembilan: Gary, you haven't answered to this concern.
Let me know and I'll fix it during merge.

Thanks!

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -438,11 +448,20 @@ else if (this.logger.isWarnEnabled()) {
...
 	private String generateLocalFileName(String remoteFileName) {
 		if (this.localFilenameGeneratorExpression != null) {
-			return this.localFilenameGeneratorExpression.getValue(this.evaluationContext, remoteFileName, String.class);
+			return this.localFilenameGeneratorExpression.getValue(this.evaluationContext, remoteFileName,
+					String.class);
 		}
 		return remoteFileName;
 	}
 
+	protected void evalueRemoteDirectory() {
```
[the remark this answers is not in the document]
artembilan: This has not been fixed as well.

Proceeding during merge...

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/RotatingServerAdvice.java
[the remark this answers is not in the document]
garyrussell: I had some stashed changes that I failed to unstash for the last commit - pushed.

artembilan:
Cherry-picked to `5.0.x` as https://github.com/spring-projects/spring-integration/commit/9f937f54cd0479ea7d8760c623b3042e814b35a9 after resolving conflicts.

PR #2490, opened 2018-06-26
INT-4488: Require @EnablePublisher

artembilan:
JIRA: https://jira.spring.io/browse/INT-4488

Since `PublisherAnnotationBeanPostProcessor` is quite expensive
(1.5x the CPU used by `MessagingAnnotationPostProcessor` in a simple
application)

* So not register a `PublisherAnnotationBeanPostProcessor` by default.
The `@EnablePublisher` has to be presented on the `@Configuration`
to allow the `@Publisher` AOP
* Change the `<int:annotation-config>` to require a new
`<int:enable-publisher>` sub-element for similar purpose - do not
register `PublisherAnnotationBeanPostProcessor` by default

### DIFF
--- a/src/reference/asciidoc/whats-new.adoc
+++ b/src/reference/asciidoc/whats-new.adoc
@@ -51,7 +51,13 @@ Starting with _version 5.0.5_, generated bean names for the components in an `In
 
 See <<java-dsl-flows>> for more information.
 
-==== AMQP Changes
+==== @Publisher annotation changes
+
+Starting with _version 5.1, the `@Publisher` AOP functionality has to be turned on explicitly via `@EnablePublisher` or via `<int:enable-publisher>` sub-element on the `<int:annotation-config>`.
+
+See <<publisher-annotation>> for more information.
+
+=== AMQP Changes
 
 `ID` and `Timestamp` header mapping changes in the `DefaultAmqpHeaderMapper`.
 See the note near the bottom of <<amqp-message-headers>> for more information.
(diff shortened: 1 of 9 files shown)


src/reference/asciidoc/whats-new.adoc
```diff
@@ -51,15 +51,21 @@ Starting with _version 5.0.5_, generated bean names for the components in an `In
...
+=== AMQP Changes
 
 `ID` and `Timestamp` header mapping changes in the `DefaultAmqpHeaderMapper`.
 See the note near the bottom of <<amqp-message-headers>> for more information.
 
 The `contentType` header is no longer incorrectly mapped as an entry in the general headers map.
 See <<amqp-content-type>> for more information.
 
-==== JDBC Changes
+=== JDBC Changes
```
garyrussell: I don't think these should be reduced to 3x `=` - that is for top-level "New Components" and "General Changes" and there is one more `==== Aggregator Changes` afterwards.

I can change them back during the merge if you agree.

garyrussell: Oh, I see; no you are right; will fix when rebasing.

artembilan: Ah! I see what you mean. 
Well, the `General Changes` name is confusing though...
May be change it to the ``Other Changes` since it goes already after `New Components`  :smile:

Otherwise yes: please, revert it back during merge if it is only concern.

garyrussell:
Merged as 69af967c7869d76ca2be84f2293360d161520fb9 with what's new polishing.

I guess we should add to the migration guide too.

src/reference/asciidoc/whats-new.adoc
artembilan: Ah, right! I definitely meant there `general` as for all the core changes and each other modules have their own top level paragraph.

artembilan:
The Migration Guide note has been added:  https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.0-to-5.1-Migration-Guide#the-publisher-annotation-processing

PR #2493, opened 2018-06-28
Editing

Buzzardo:
I did a full editing pass for the Spring Integration Reference Guide.

I checked and corrected spelling, punctuation, grammar, usage, and voice. I also edited for clarity and conciseness.

I had to make two commits, because people made changes during the month I spent editing. On the second commit, I edited the changes people made during that time.


garyrussell:
Many thanks @Buzzardo ; we will do our best to work through this ASAP.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -89,66 +81,54 @@ public AbstractCorrelatingMessageHandler(MessageGroupProcessor processor, Messag
...
+Only POJO methods could be used for aggregation.
+Now, if the framework detects that the referenced (or inner) bean implements `MessageProcessor`, it is used as the aggregator's output processor.
 
-If you wish to release a collection of objects from a custom `MessageGroupProcessor` as the payload of a message, your
-class should extend `AbstractAggregatingMessageGroupProcessor` and implement `aggregatePayloads()`.
+If you wish to release a collection of objects from a custom `MessageGroupProcessor` as the payload of a message, your class should extend `AbstractAggregatingMessageGroupProcessor` and implement `aggregatePayloads()`.
 
-Also, since _version 4.2_, a `SimpleMessageGroupProcessor` is provided; which simply returns the collection of
-messages from the group, which, as indicated above, causes the released messages to be sent individually.
+Also, since version 4.2, a `SimpleMessageGroupProcessor` is provided.
```
artembilan: Well, @garyrussell selected and insisted on the style to cursive the version whenever it is possible.
What made you to went away from that?
Is that a style applied in other Spring projects?

I don't have objection though, but still... Feels like unnecessary change.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -89,66 +81,54 @@ public AbstractCorrelatingMessageHandler(MessageGroupProcessor processor, Messag
...
+It returns the collection of messages from the group, which, as indicated earlier, causes the released messages to be sent individually.
 
-This allows the aggregator to work as a message barrier where arriving messages are held until the release strategy
-fires, and the group is released, as a sequence of individual messages.
+This lets the aggregator work as a message barrier, where arriving messages are held until the release strategy fires and the group is released as a sequence of individual messages.
 
-===== ReleaseStrategy
+===== `ReleaseStrategy`
```
artembilan: This is going to be a title of the paragraph.
Does it really work properly to wrap it into the code?

src/reference/asciidoc/aggregator.adoc
```diff
@@ -89,66 +81,54 @@ public AbstractCorrelatingMessageHandler(MessageGroupProcessor processor, Messag
...
 
 The `ReleaseStrategy` interface is defined as follows:
 
+====
```
artembilan: Looks like you wrapped almost all the code snippets to these symbols.
What is the point, please ?

src/reference/asciidoc/aggregator.adoc
```diff
@@ -187,53 +168,55 @@ public class MyReleaseStrategy {
...
+You can release partial sequences by using a `MessageGroupStoreReaper` together with `send-partial-result-on-expiry` being set to `true`.
 
 IMPORTANT: To facilitate discarding of late-arriving messages, the aggregator must maintain state about the group after it has been released.
-This can eventually cause out of memory conditions.
-To avoid such situations, you should consider configuring a `MessageGroupStoreReaper` to remove the group metadata; the expiry parameters should be set to expire groups after it is not expected that late messages will arrive.
-For information about configuring a reaper, see <<reaper>>.
+This can eventually cause out-of-memory conditions.
+To avoid such situations, you should consider configuring a `MessageGroupStoreReaper` to remove the group metadata.
+The expiry parameters should be set to expire groups once a point has been reach after after which late messages are not expected to arrive.
+For information about configuring a reaper, see "`<<reaper>>`".
```
artembilan: Is it really going to work properly?
Our point was to have a link to the appropriate chapter.
Why have wrapped it to the code and to double quotes at the same time?

src/reference/asciidoc/aggregator.adoc
```diff
@@ -262,74 +247,75 @@ public interface CorrelationStrategy {
...
-By default, the correlation strategy is a `HeaderAttributeCorrelationStrategy` returning the value of the `CORRELATION_ID` header attribute.
-If you have a custom header name you would like to use for correlation, then simply configure that on an instance of `HeaderAttributeCorrelationStrategy` and provide that as a reference for the Aggregator's correlation-strategy.
+By default, the correlation strategy is a `HeaderAttributeCorrelationStrategy` that returns the value of the `CORRELATION_ID` header attribute.
+If you have a custom header name you would like to use for correlation, you can configure it on an instance of `HeaderAttributeCorrelationStrategy` and provide that as a reference for the aggregator's correlation strategy.
 
-===== LockRegistry
+===== `LockRegistry`
 
-Changes to groups are thread safe; a `LockRegistry` is used to obtain a lock for the resolved correlation id.
+Changes to groups are thread safe. A `LockRegistry` is used to obtain a lock for the resolved correlation ID.
```
artembilan: One sentence per line.
I don't have objection for the `ID`

src/reference/asciidoc/aggregator.adoc
```diff
@@ -598,20 +531,23 @@ public class PojoCorrelationStrategy {
   }
 }
 ----
+====
 
-For example, this aggregator would group numbers by some criterion (in our case the remainder after dividing by 10) and will hold the group until the sum of the numbers provided by the payloads exceeds a certain value.
+The aggregator in the preceding example would group numbers by some criterion (in this case, the remainder after dividing by ten) and hold the group until the sum of the numbers provided by the payloads exceeds a certain value.
 
-NOTE: Wherever it makes sense, the release strategy method, correlation strategy method and the aggregator method can be combined in a single bean (all of them or any two).
+NOTE: Wherever it makes sense to do so, the release strategy method, the correlation strategy method, and the aggregator method can be combined in a single bean. (Actually, all of them or any two of them can be combined.)
```
artembilan: One sentence per line

src/reference/asciidoc/aggregator.adoc
```diff
@@ -622,76 +558,89 @@ public String[] processRelease(List<Message<String>> messages){
...
-If `send-partial-result-on-expiry` is `true`, existing messages in the (partial) `MessageGroup` will be released as a normal aggregator reply Message to the `output-channel`, otherwise it will be discarded.
+The results of forcing the group to complete depends on the `ReleaseStrategy` and the `send-partial-result-on-expiry`.
+First, the release strategy is again consulted to see if a normal release is to be made.
+While the group has not changed, the `ReleaseStrategy` can decide to release the group at this time.
+If the release strategy still does not release the group, it is expired.
+If `send-partial-result-on-expiry` is `true`, existing messages in the (partial) `MessageGroup` are released as a normal aggregator reply message to the `output-channel`.
+Otherwise, it is discarded.
 
-There is a difference between `groupTimeout` behavior and `MessageGroupStoreReaper` (see <<aggregator-config>>).
+There is a difference between `groupTimeout` behavior and `MessageGroupStoreReaper` (see "`<<aggregator>>`").
```
artembilan: This points to fully different paragraph unlike the old one `aggregator-config`.
I think this indeed was the point to link end-user to the particular paragraph, but not the whole chapter

src/reference/asciidoc/aggregator.adoc
```diff
@@ -808,18 +750,21 @@ public interface MessageGroupCallback {
 
 }
 ----
+====
 
-The callback has direct access to the store and the message group so it can manage the persistent state (e.g.
-by removing the group from the store entirely).
+The callback has direct access to the store and the message group so that it can manage the persistent state (for example,
```
artembilan: One sentence per line.
I mean the next line should be merged with this one.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -808,18 +750,21 @@ public interface MessageGroupCallback {
...
+by entirely removing the group from the store).
 
-The `MessageGroupStore` maintains a list of these callbacks which it applies, on demand, to all messages whose timestamp is earlier than a time supplied as a parameter (see the `registerMessageGroupExpiryCallback(..)` and `expireMessageGroups(..)` methods above).
+The `MessageGroupStore` maintains a list of these callbacks, which it applies, on demand, to all messages whose timestamps are earlier than a time supplied as a parameter (see the `registerMessageGroupExpiryCallback(..)` and `expireMessageGroups(..)` methods, described earlier).
+// TODO Need a link here, but there's no clear heading to which to link.
```
artembilan: This is about a `<<reaper>>` paragraph.

artembilan (changes requested):
I think that's enough for now (just an `aggregator.adoc`).
When I get answers to some my concerns, I won't be so picky.

Thank you for the tremendous work, @Buzzardo !

src/reference/asciidoc/aggregator.adoc
garyrussell: I don't recall why we started doing that; I don't mind either way - I certainly don't think it's necessary to change them all back.

artembilan: Yes, I'm fine with that too.
More over I see that SF doesn't use cursive for its version numbers.

Buzzardo: Code should always be in the code font, which is done with  single back ticks in Asciidoc. Otherwise, quotation marks should be curly quotation marks, which is done with single back ticks inside of double quotations marks in Asciidoc. The only use for double quotation marks is when they appear in the code being quoted.

src/reference/asciidoc/aggregator.adoc
Buzzardo: Yes. It renders correctly.

A little clarification: Paragraphs don't get titles. That's a heading.

src/reference/asciidoc/aggregator.adoc
Buzzardo: For code listings with callouts, it groups the callouts in a box with the code listing. Because that applies a certain formatting, we also need to do it all the other code listings. So, yes, I try to do it to all of them. Missing one would be the error.

src/reference/asciidoc/aggregator.adoc
Buzzardo: That combination makes curly double quotation marks. Curly double quotation marks are the standard in the publishing industry. Straight quotation marks are used only in code.

src/reference/asciidoc/aggregator.adoc
Buzzardo: I object to id, because it is actually a word, whereas ID is the accepted abbreviation of "identification". Consequently, the only time I'll let id go by is when it's in code. (Yes, I am super fussy.)

src/reference/asciidoc/aggregator.adoc
Buzzardo: Oops. Missed that one. Thanks.

src/reference/asciidoc/aggregator.adoc
Buzzardo: Cool. Thanks. I'll add the link.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -89,66 +81,54 @@ public AbstractCorrelatingMessageHandler(MessageGroupProcessor processor, Messag
...
 }
 ----
+====
 
-As for actual processing of the message group, the default implementation is the `DefaultAggregatingMessageGroupProcessor`.
-It creates a single Message whose payload is a List of the payloads received for a given group.
-This works well for simple Scatter Gather implementations with either a Splitter, Publish Subscribe Channel, or Recipient List Router upstream.
+As for the actual processing of the message group, the default implementation is the `DefaultAggregatingMessageGroupProcessor`.
+It creates a single `Message` whose payload is a `List` of the payloads received for a given group.
+This works well for simple scatter-gather implementations with a splitter, a publish-subscribe channel, or a recipient list router upstream.
```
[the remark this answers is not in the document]
Buzzardo: The convention is to not add emphasis. Capitalization is a form of emphasis. The structure of the sentence should make clear what's important and what's not. So, yes, I de-capitalize everything that isn't a title (and titles get formatted - italics for works and curly quotation marks for parts of works).

We could add a note to the Code Conventions section indicating that we capitalize pattern names. I would still prefer not to, but I could live with that. 

Possibly interesting aside: In the 18th century (and later in other languages, such as German), the convention was to capitalize all nouns. English moved away from that in the early 1800s (coinciding with the Romantic movement). So now the convention is to capitalize only titles (and that only because the titles are themselves capitalized). Times change.

src/reference/asciidoc/aggregator.adoc
artembilan: I think this is not the place you have been intended to answer, @Buzzardo, but anyway it doesn't matter: we agreed with Gary to stay with your change.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -157,17 +137,16 @@ public interface ReleaseStrategy {
...
+* If the argument is a `java.util.List<T>`, and the parameter type `T` is assignable to `Message`, the whole list of messages accumulated in the group is sent to the method.
+* If the argument is a non-parametrized `java.util.List` or the parameter type is not assignable to `Message`, the method receives the payloads of the accumulated messages.
+* The method must return `true` if the message group is ready for aggregation or false otherwise.
 
-* the method must return true if the message group is ready for aggregation, and false otherwise.
-
-For example:
+The following example shows how to use the `@ReleaseStrategy` annotation for a `List` of type `Message`:
 
 [source,java]
```
artembilan: This code snippet is missed to be wrapped with the `====`.
Or am I missing the point why and when should we wrap them?

Thanks

src/reference/asciidoc/aggregator.adoc
```diff
@@ -249,11 +232,13 @@ Here is an example of how to configure an aggregator with the previous implement
 
 <bean id="releaser" class="SimpleSequenceSizeReleaseStrategy" />
 ----
+====
 
 ===== CorrelationStrategy
```
artembilan: Must be wrapped into code as you do in other headers when it is about a class name.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -262,74 +247,76 @@ public interface CorrelationStrategy {
...
 
-Changes to groups are thread safe; a `LockRegistry` is used to obtain a lock for the resolved correlation id.
+Changes to groups are thread safe.
+A `LockRegistry` is used to obtain a lock for the resolved correlation ID.
 A `DefaultLockRegistry` is used by default (in-memory).
-For synchronizing updates across servers, where a shared `MessageGroupStore` is being used, a shared lock registry
-must be configured.
-See <<aggregator-config>> below for more information.
+For synchronizing updates across servers, where a shared `MessageGroupStore` is being used, you must configure a shared lock registry.
+See "`<<aggregator>>`" for more information.
```
artembilan: This is wrong a bit: the `[[aggregator]]` is a link to the chapter where we are.
The `[[aggregator-config]]` was about a general section for configuration (xml, dsl etc.).
Right now I see with a couple next lines that you broke such a structure and introduced separate sections for `xml`, `dsl` etc. I understand that it may make sense when we come back to the collapsing tabs between configuration styles, but the link here is still wrong: we definitely need to point to the proper configuration position or just mention where to look. Honestly it even much easier here since the configuration part is the next paragraph 😄

src/reference/asciidoc/aggregator.adoc
```diff
@@ -622,76 +560,89 @@ public String[] processRelease(List<Message<String>> messages){
...
+In the preceding example, the root object of the SpEL evaluation context is the `MessageGroup` itself, and you are stating that, as soon as there is a message with payload of `5` in this group, the group should be released.
 
 [[agg-and-group-to]]
 ====== Aggregator and Group Timeout
 
-Starting with _version 4.0_, two new mutually exclusive attributes have been introduced: `group-timeout` and `group-timeout-expression` (see the description above).
-There are some cases where it is needed to emit the aggregator result (or discard the group) after a timeout if the `ReleaseStrategy` doesn't _release_ when the current Message arrives.
-For this purpose the `groupTimeout` option allows scheduling the `MessageGroup` to be forced complete:
+Starting with version 4.0, two new mutually exclusive attributes have been introduced: `group-timeout` and `group-timeout-expression` (see the earlier description).
+// TODO This needs a link, but I don't see a target for it.
```
artembilan: This is about an `[[aggregator-xml]]` section.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -347,214 +334,157 @@ Below you can see an example of an aggregator.
...
-However `MessageGroup` s can alternatively be expired by simply calling `MessageGroupStore.expireMessageGroups(timeout)`.
-That could be accomplished via a Control Bus operation or by simply invoking that method if you have a reference to the `MessageGroupStore` instance.
-Otherwise by itself this attribute has no behavior.
-It only serves as an indicator of what to do (discard or send to the output/reply channel) with Messages that are still in the `MessageGroup` that is about to be expired.
-_Optional_.
-_Default - false_.
-*NOTE:* This attribute is more properly `send-partial-result-on-timeout` because the group may not actually expire if
+Optional.
+By default, it is a volatile in-memory store.
+// TODO How would one change it?
```
artembilan: Well, everything is about beans in Spring.
The sentence like "A reference to a `MessageGroupStore`" should do the trick and make it obvious that you need to configure a bean on the matter and point its bean name here.
Yeah... we can point from here to the `[[message-store]]` chapter though...

src/reference/asciidoc/aggregator.adoc
```diff
@@ -347,214 +334,157 @@ Below you can see an example of an aggregator.
...
+Optional.
+<8> Indicates that expired messages should be aggregated and sent to the 'output-channel' or 'replyChannel' once their containing `MessageGroup` is expired (see https://docs.spring.io/spring-integration/api/org/springframework/integration/store/MessageGroupStore.html#expireMessageGroups-long[`MessageGroupStore.expireMessageGroups(long)`]).
+One way of expiring a `MessageGroup` is by configuring a `MessageGroupStoreReaper`.
+However you can alternatively expire `MessageGroup` by calling `MessageGroupStore.expireMessageGroups(timeout)`.
+You can accomplish that through a Control Bus operation or, if you have a reference to the `MessageGroupStore` instance, by invoking `expireMessageGroups(timeout)`.
+Otherwise, by itself, this attribute does nothing.
+It serves only as an indicator of whether to discard or send to the output or reply channel any messages that are still in the `MessageGroup` that is about to be expired.
+Optional (the default is `false`).
+NOTE: This attribute might more properly be called `send-partial-result-on-timeout`, because the group may not actually expire if
 `expire-groups-upon-timeout` is set to `false`.
```
artembilan: This must be join with the previous line to follow the `sentence per line` rule.

artembilan (changes requested):
Some further review for an aggregator chapter

artembilan:
I merged it to `master` like this https://github.com/spring-projects/spring-integration/commit/f3bf2a68d6ac2ee519ea187b062a2a7640309970 with subsequent polishing like this: https://github.com/spring-projects/spring-integration/commit/9a7ae3dfad8ce7ba5085f41ec38f6ef9c0ee4e22

There are still a lot of warning during Doc building and also I've noticed that  
```
"`<<reaper>>`"
```

Indeed doesn't work in the HTML as many other similar 🤷‍♂️ 

There is also a duplication of author list, but I guess that can be addressed when we get rid of DocBook.

Sorry, for broker Reference Manual on the `master` now, but that is the best we can do for now...

See you next week with more review/polishing!

garyrussell:
@Buzzardo there are some issues with the ToC - compare

current: https://docs.spring.io/spring-integration/reference/html/
snapshot: https://docs.spring.io/spring-integration/docs/5.1.0.BUILD-SNAPSHOT/reference/html/

The latter jumps from "Messaging Endpoints" to "Stream Support".
The former has many other sections in-between (including a couple in Section IV and a bunch from Section V and Section V is now the appendices).

Thanks, Jay.

garyrussell:
Looks like everything starts to go awry below here: https://docs.spring.io/spring-integration/docs/5.1.0.BUILD-SNAPSHOT/reference/html/messaging-endpoints-chapter.html#advised-handler-properties

From then-on, there are many un-rendered headings:

`=== Logging Channel Adapter`

artembilan:
There is this piece of code with the missed wrapping:
```
====
[source, java]
----
String componentName = ((NamedComponent) target).getComponentName();
----

When you implement `MethodInterceptor` directly, you could cast the target object as follows:
```

garyrussell:
This fixes most of the problem (now it stops at XMPP).

```
diff --git a/src/reference/asciidoc/handler-advice.adoc b/src/reference/asciidoc/handler-advice.adoc
index 09f2f9678f..53c2ef6d03 100644
--- a/src/reference/asciidoc/handler-advice.adoc
+++ b/src/reference/asciidoc/handler-advice.adoc
@@ -856,4 +856,6 @@ public IntegrationFlow flow() {
 
 NOTE: The `IdempotentReceiverInterceptor` is designed only for the `MessageHandler.handleMessage(Message<?>)` method.
 Starting with version 4.3.1, it implements `HandleMessageAdvice`, with the `AbstractHandleMessageAdvice` as a base class, for better dissociation.
-See "`<<handle-message-advice>>`" for more information.
+See "<<handle-message-advice>>" for more information.
+
+====
```

(Missing `====`) I guess we put a pair of these around each piece - this one's end was missing.

garyrussell:
Fixed - https://github.com/spring-projects/spring-integration/commit/70db967d4fb3ed9be8846d41b60ca9374b393de2

Note that we also can't have backticks within `<<...>>` - they render as literals instead of links.

garyrussell:
Looks like many of the references have quotes around and backticks in them (and some also have quotes around) - Jay was this intentional for some reason?

```
onfiguration of the `PayloadTypeRouter` is also supported by the namespace provided by Spring Integration (see `<<configuration-namespace>>`), which essentially simplifies configuration by combining the `<router/>` configuration...
```
Rendered as 

>Configuration of the PayloadTypeRouter is also supported by the namespace provided by Spring Integration (see `<<configuration-namespace>>`), which

artembilan:
Right, Gary. This was my question and @Buzzardo 's answer on the matter: https://github.com/spring-projects/spring-integration/pull/2493/files#r199223231

But as experience shows it doesn't work for generate HTML, neither PDF.
Maybe we are missing some upgrades in the ASCIIDoc plugin or anything else?..

Buzzardo:
I apologize. It does indeed turn out that we can't have curly quotes. I didn't think that backticks around links would make them render literally and didn't think to test it. I get the impression that you have already fixed it. If not, I'll be happy to do so.

artembilan:
Thank you, @Buzzardo , for coming back to us!

Sorry, we din't fix that. I thought you have done something similar in other projects and that have worked well, so we could just miss something else to configure, e.g. in the particular Gradle plugin.
And yes, it indeed works with the plain ASCIIDoc viewer, but not in the generated target docs. 😢

src/reference/asciidoc/amqp.adoc
```diff
@@ -1314,7 +1314,11 @@ The `org.springframework.amqp.support.AmqpHeaders` class identifies the default
 * `amqp_clusterId`
 * `amqp_contentEncoding`
 * `amqp_contentLength`
+<<<<<<< HEAD
```
artembilan: Here is some rebase conflict problem.

src/reference/asciidoc/amqp.adoc
```diff
@@ -1341,7 +1345,10 @@ The `org.springframework.amqp.support.AmqpHeaders` class identifies the default
 * `amqp_consumerTag`
 * `amqp_consumerQueue`
 
+<<<<<<< HEAD
```
artembilan: DITTO

src/reference/asciidoc/changes-4.0-4.1.adoc
```diff
@@ -12,50 +12,50 @@ Version 4.1 added a number of new components.
...
 ===== Idempotent Receiver Pattern
 
 We added the idempotent receiver enterprise integration pattern implementation by adding the `<idempotent-receiver>` component in XML or the `IdempotentReceiverInterceptor` and `IdempotentReceiver` annotations for Java configuration.
-See "`<<idempotent-receiver>>`" and the https://docs.spring.io/spring-integration/api/index.html[Javadoc] for more information.
+See <<idempotent-receiver>> and the https://docs.spring.io/spring-integration/api/index.html[Javadoc] for more information.
 
 [[x4.1-BoonJsonObjectMapper]]
 ===== Boon `JsonObjectMapper`
 
 We added T\the Boon `JsonObjectMapper` for the JSON transformers.
```
artembilan: Here is some typo. Probably not yours, but I'd appreciate if you fix it in this PR. 😄 
Thanks

src/reference/asciidoc/changes-4.1-4.2.adoc
```diff
@@ -162,17 +162,17 @@ See "<<tcp-connection-factories>>" for more information.
 ===== TCP Gateway Remote Timeout
 
 The `TcpOutboundGateway` now supports `remote-timeout-expression` as an alternative to the existing `remote-timeout` attribute.
-This allows setting the timeout based on each message.
+This allows setting the timeout based on each message.<<
```
artembilan: Some leftover angle brackets

src/reference/asciidoc/channel.adoc
```diff
@@ -639,13 +639,13 @@ If you provide a resequencer or aggregator downstream from a `PublishSubscribeCh
 Doing so indicates that the channel should set the `sequence-size` and `sequence-number` message headers as well as the correlation ID prior to passing along the messages.
 For example, if there are five subscribers, the `sequence-size` would be set to `5`, and the messages would have `sequence-number` header values ranging from `1` to `5`.
 
-Alongside with the `Executor`, an `ErrorHandler` can be configured as well.
-By default the `PublishSubscribeChannel` uses a `MessagePublishingErrorHandler` implementation to send error to the `MessageChannel` from the `errorChannel` header or a global `errorChannel` instance.
-If an `Executor` is not configured, the `ErrorHandler` is ignored and exceptions are thrown directly to the caller's Thread.
+Along with the `Executor`, you can also configure an `ErrorHandler`.
+By default, the `PublishSubscribeChannel` uses a `MessagePublishingErrorHandler` implementation to send an error to the `MessageChannel` from the `errorChannel` header or from a global `errorChannel` instance.
```
artembilan: Well, actually a "global `errorChannel` instance" is really already a `MessageChannel`, so I think would be better to say:

> or into the global `errorChannel` instance

src/reference/asciidoc/dsl.adoc
```diff
@@ -652,7 +652,11 @@ For example, the `ConsumerEndpointFactoryBean` for the `.transform("Hello "::con
 The `Transformer` implementation bean for that endpoint  has a bean name of `lambdaFlow.o.s.i.transformer.MethodInvokingTransformer#0`.
 (In both cases, `o.s.i` is `org.springframework.integration`, shortened here to fit on the page.)
 These generated bean names are prepended with the flow ID for purposes such as parsing logs or grouping components together in some analysis tool, as well as to avoid a race condition when we concurrently register integration flows at runtime.
+<<<<<<< HEAD
```
artembilan: Rebase conflict again

src/reference/asciidoc/jdbc.adoc
```diff
@@ -831,6 +831,13 @@ Since Spring Integration 3.0.
 Optional.
 ====
 
+<<<<<<< HEAD
```
artembilan: One more conflict

src/reference/asciidoc/message-publishing.adoc
```diff
@@ -16,7 +16,11 @@ Spring Integration provides two approaches: XML configuration and annotation-dri
 
 The annotation-driven approach lets you annotate any method with the `@Publisher` annotation to specify a 'channel' attribute.
 Starting with version 5.1, to switch this functionality on, you must use the `@EnablePublisher` annotation on some `@Configuration` class.
+<<<<<<< HEAD
```
artembilan: Again

src/reference/asciidoc/xml.adoc
```diff
@@ -1,4 +1,4 @@
-[[xml]]
+<<[[xml]]
```
artembilan: I think this brackets are wrong here

artembilan (changes requested):
This is what I've seen so far.

Thank you for your incredible effort!

garyrussell:
@artembilan - we closed this PR; Jay will issue a new PR for today's changes with 2 commits (today's that you reviewed - we squashed his branch and made a new commit comment - and a second with fixes for your review comments).

artembilan:
Yep! Fine with me! Sorry, I’m gone already. Will come back in the evening

PR #2494, opened 2018-06-29
INT-4495: (S)FTP: fix max-fetch with directories

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4495

Previously, `maxFetch` was applied before directories were removed from
the fetch list.

Remove the directories before filtering and applying `maxFetch`.

**cherry-pick to 5.0.x**

**suggest reviewing on GitHub with `?w=1`**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java
@@ -16,16 +16,9 @@
 
 package org.springframework.integration.file.support;
 
-import java.lang.reflect.Array;
 import java.nio.file.FileSystems;
 import java.util.Arrays;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.function.Predicate;
-
-import org.springframework.lang.Nullable;
-import org.springframework.util.Assert;
 
 /**
  * Utilities for operations on Files.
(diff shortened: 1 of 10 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java
```diff
@@ -29,6 +38,53 @@
...
 
+	/**
+	 * Remove entries from the array if the predicate returns true for an element.
+	 * @param fileArray the array.
+	 * @param predicate the predicate.
+	 * @param <F> the file type.
+	 * @return the array (modified or not), or null if all elements are removed.
+	 * @since 5.0.7
+	 */
+	public @Nullable static <F> F[] purgeUnwantedElements(F[] fileArray, Predicate<F> predicate) {
```
artembilan: Can't we do like this instead:
````java
	@Nullable
	public static <F> F[] purgeUnwantedElements(F[] fileArray, Predicate<F> predicate) {
		Object[] files =
				Arrays.stream(fileArray)
						.filter(predicate.negate())
						.toArray();

		if (files.length == 0) {
			return null;
		}
		else {
			return (F[]) files;
		}
	}
````

Or do you really worry about `Stream` instances all those support objects, plus empty array object if that?
However at the same time you do `new LinkedList<>(Arrays.asList(fileArray));`...

File: spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java [hunk 2]
```diff
@@ -29,6 +38,53 @@
...
+	 * Create an array of generic type from a list. Must have at least one entry.
+	 * @param files the list.
+	 * @param <F> the file type.
+	 * @return the array.
+	 * @since 5.0.7
+	 */
+	public static <F> F[] toGenericArray(List<F> files) {
+		Assert.isTrue(files.size() > 0, "file list must have at least one entry");
+		@SuppressWarnings("unchecked")
+		F[] array = (F[]) Array.newInstance(files.get(0).getClass(), files.size());
```
artembilan: How about this instead:
```
return files.toArray((F[]) java.lang.reflect.Array.newInstance(files.get(0).getClass(), files.size()));
```
?

artembilan (changes requested):
Just a couple concerns for utilities.

spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java
garyrussell: Cool; pushing.

artembilan (approved):
Will merge after green Travis.

garyrussell:
No; it won't work.

spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java
artembilan: Hm. OK. 
How about this then:
````java
return Arrays.stream(fileArray)
				.filter(predicate.negate())
				.toArray(size -> (F[]) Array.newInstance(fileArray[0].getClass(), size));
````
?

**Review by artembilan — CHANGES_REQUESTED** [review]

File: spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java [hunk 3]
```diff
@@ -41,9 +42,9 @@
 	 */
 	@SuppressWarnings("unchecked")
 	public static <F> F[] purgeUnwantedElements(F[] fileArray, Predicate<F> predicate) {
-		return (F[]) Arrays.stream(fileArray)
+		return Arrays.stream(fileArray)
 				.filter(predicate.negate())
-				.toArray();
+				.toArray(size -> (F[]) Array.newInstance(fileArray[0].getClass(), size));
```
artembilan: I think we need to add Assert about the length in the beginning of the method.
Or just proceed to the `return` when there are no elements in the original array...

artembilan (changes requested):

artembilan (approved):
Will merge after green Travis.

PR #2495, opened 2018-06-29
INT-4496: (S)FTP: Sort file list earlier

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4496

Sort the file list before applying `maxFetchSize` and filters.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Matcher;
 
@@ -116,6 +117,8 @@ public abstract class AbstractInboundFileSynchronizer<F>
 
 	private BeanFactory beanFactory;
 
+	private Comparator<F> comparator;
+
 	/**
 	 * Create a synchronizer with the {@link SessionFactory} used to acquire {@link Session} instances.
 	 *
@@ -126,6 +129,20 @@ public abstract class AbstractInboundFileSynchronizer<F>
 		this.remoteFileTemplate = new RemoteFileTemplate<F>(sessionFactory);
 	}
 
+
+	protected Comparator<F> getComparator() {
+		return this.comparator;
+	}
+
+	/**
+	 * Set a comparator to sort the retrieved list of F prior to applying
+	 * filters and max fetch size.
+	 * @param comparator the comparator.
+	 * @since 5.1
+	 */
+	public void setComparator(Comparator<F> comparator) {
+		this.comparator = comparator;
+	}
 
 	/**
 	 * @param remoteFileSeparator the remote file separator.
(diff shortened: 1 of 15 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -127,6 +130,20 @@ public AbstractInboundFileSynchronizer(SessionFactory<F> sessionFactory) {
 	}
 
 
+	protected Comparator<F> getComparator() {
+		return this.comparator;
+	}
+
+	/**
+	 * Set a comparator to sort the retrieved list of F prior to applying
```
artembilan: The `F` doesn't sound friendly here.
Can't we say `remote files` instead?

spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java
```diff
@@ -38,19 +40,29 @@
...
-					.toArray(size -> (F[]) Array.newInstance(fileArray[0].getClass(), size));
+			if (comparator == null) {
+				return Arrays.stream(fileArray)
+						.filter(predicate.negate())
+						.toArray(size -> (F[]) Array.newInstance(fileArray[0].getClass(), size));
+			}
+			else {
+				return Arrays.stream(fileArray)
+						.filter(predicate.negate())
+						.sorted(comparator)
```
artembilan: Can't we extract only this part of stream to the `if...else`?
I mean that there are enough copy/pasted code that it would be hard to support in both branches when it becomes much complicated eventually.

Although we can extract it at that point then, not now...

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/dsl/FtpStreamingInboundChannelAdapterSpec.java
```diff
@@ -43,7 +42,7 @@
 				FtpStreamingMessageSource> {
 
 	FtpStreamingInboundChannelAdapterSpec(RemoteFileTemplate<FTPFile> remoteFileTemplate,
-			Comparator<AbstractFileInfo<FTPFile>> comparator) {
+			Comparator<FTPFile> comparator) {
```
artembilan: Update Copyright?

src/reference/asciidoc/whats-new.adoc
```diff
@@ -85,3 +85,8 @@ A `RotatingServerAdvice` is now available to poll multiple servers and/or direct
 See <<ftp-rotating-server-advice>> and <<sftp-rotating-server-advice>> for more information.
 
 Also inbound adapter `localFilenameExpression` s can contain the variable `#remoteDirectory` which contains the remote directory being polled.
+
+The generic type of the comparators, used to sort the fetched file list for the streaming adapters, has changed from `AbstractFileInfo<F>` to simply `<F>`.
```
artembilan: Well, semantically an `<F>` is not equal to `F`, however it definitely feels like you are talking about exactly `F`. Just because your `AbstractFileInfo<F>` is not wrapped to the angle brackets

artembilan (changes requested):
Just a couple concerns.

spring-integration-file/src/main/java/org/springframework/integration/file/support/FileUtils.java
garyrussell: Yeah - my reasoning here was that temporary storage in local variables is avoided this way.

artembilan:
Please, add also a Migration Guide note.

Thanks

garyrussell:
Migration guide updated; also added a Table of Contents: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.0-to-5.1-Migration-Guide

PR #2496, opened 2018-07-05
JDBC Channel message store: Query Provider for SQL server

sundarabalajijk:
Added this file to support channel message store for Azure SQL / SQL server on the go.


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/SqlChannelMessageStoreQueryProvider.java
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.integration.jdbc.store.channel;
+
+/**
+ * @author Sundara Balaji
+ * @since 2.2
+ */
+public class SqlChannelMessageStoreQueryProvider extends AbstractChannelMessageStoreQueryProvider {
```
artembilan: Can we call it `SqlServer...` instead to be more precise?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/SqlChannelMessageStoreQueryProvider.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
```
artembilan: This is a new class, so only a current year.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/SqlChannelMessageStoreQueryProvider.java
```diff
@@ -0,0 +1,60 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.jdbc.store.channel;
+
+/**
+ * @author Sundara Balaji
+ * @since 2.2
```
artembilan: This new class is really introduced in the version `5.1`.

Also, please, provide some short class level JavaDoc to explain what is this about.

artembilan (changes requested):
@sundarabalajijk ,

I have only simple concerns to your fix. If you agree with them, I can fix them during merge.

Also, please, confirm that this provider works as expected against real SQL Server.

Thanks again

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/SqlChannelMessageStoreQueryProvider.java
sundarabalajijk: yes it can be.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/SqlChannelMessageStoreQueryProvider.java
sundarabalajijk: Yep. Thanks.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/SqlChannelMessageStoreQueryProvider.java
sundarabalajijk: Yep. Thanks.

sundarabalajijk:
I agree the fixes you have mentioned.

And yes I confirm this is working. I have tested it with Azure Sql Server.

artembilan:
@sundarabalajijk ,

Looking forward for more!

PR #2499, opened 2018-07-06
Fix ThreadAffinityClientConnectionFactory

garyrussell:
`isSingleUse()` should return `false` by default - the whole point is
to bind a connection to a thread. The TACCF should have its own flag
not return the delegate's.

Also, the `TcpReceivingChanneladapter` calls `closeConnection(connectionId)`.
This should be a no-op for the TACCF since the proper way to close a connection
is to call `releaseConnection()` on the thread to which the connection is bound.

**cherry-pick to 5.0.x**


artembilan:

PR #2502, opened 2018-07-11
INT-4490: DefaultLockRepo: no uppercase format

artembilan:
JIRA https://jira.spring.io/browse/INT-4490

The `DefaultLockRepository.prefix` can be configured with the schema name.
And this one can be case-sensitive.

* Change format hint from the `%S` to `%s` to avoid upper-casing.


garyrussell:
And cherry-picked to 5.0.x, 4.3.x

PR #2504, opened 2018-07-12
INT-4500: The DefaultAmqpHeaderMapper was ignoring the correlationId …

ssinger:
JIRA https://jira.spring.io/browse/INT-4500


artembilan (approved):
Indeed this is a good catch.
The missed test-case is really a shame on me 😢 

Thank you for the fix and test!

Merging with an appropriate polishing.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/0dcbde06426c44ef8f34052c7b3bbc7f885f8a3f and cherry-picked to `5.0.x`.

@ssinger ,

Looking forward for more!

PR #2509, opened 2018-07-23
INT-4502: Add support for new file callback in FileWritingMessageHandler

alturkovic:
JIRA: https://jira.spring.io/browse/INT-4502

When using `APPEND` or `APPEND_NO_FLUSH` and creating a new file, a callback should be called on the file to allow initial setup of the file, such as CSV header writing.


spring-integration-file/src/main/java/org/springframework/integration/file/NewFileCallback.java
```diff
@@ -0,0 +1,30 @@
...
+package org.springframework.integration.file;
+
+import java.io.File;
+
+/**
+ * A callback for newly created files in {@link FileWritingMessageHandler}.
+ *
+ * @author Alen Turkovic
+ */
+public interface NewFileCallback {
```
artembilan: Why can't we just stick with the `java.util.function.Consumer`?
I don't see reason in a new functional interface...

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -539,6 +551,11 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
 					"The destination file already exists at '" + resultFile.getAbsolutePath() + "'.");
 		}
 
+		if (!exists && (FileExistsMode.APPEND.equals(this.fileExistsMode)
+				|| FileExistsMode.APPEND_NO_FLUSH.equals(this.fileExistsMode))) {
+			this.newFileCallback.handle(resultFile);
```
artembilan: This is wrong place to call this callback.
See the code like `resultFile.getParentFile().mkdirs();` below.
You told that you would like to add a header to a new file, but you need to ensure valid file system structure before you open a file for writing.

spring-integration-file/src/main/java/org/springframework/integration/file/NoopNewFileCallback.java
```diff
@@ -0,0 +1,33 @@
...
+package org.springframework.integration.file;
+
+import java.io.File;
+
+/**
+ * A {@link NewFileCallback} which does nothing for new files.
+ *
+ * @author Alen Turkovic
+ */
+public class NoopNewFileCallback implements NewFileCallback {
```
artembilan: I don't see reason in this. The simple `if (this.newFileCallback != null)` is fully enough.

artembilan (changes requested):
New API must have `@since` and good Javadocs.
Also some note in the `whats-new.adoc` and the `file.adoc` would be great, too.

Thank you for you contribution so far!

spring-integration-file/src/main/java/org/springframework/integration/file/NewFileCallback.java
alturkovic: No problem, will change. I was actually thinking of a BiConsumer<File, Message<?>> with the incoming message in case some information (for an example, a CSV header) is written in the message.

spring-integration-file/src/main/java/org/springframework/integration/file/NoopNewFileCallback.java
alturkovic: Will remove.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
alturkovic: Thanks for the correction.

spring-integration-file/src/main/java/org/springframework/integration/file/NewFileCallback.java
artembilan: Yes, that good idea, indeed!

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
artembilan: Well, looking into this one more time, I would say that we have to call this callback much further - in the `handleInputStreamMessage()` and its `WhileLockedProcessor`.
The same should be done in the `handleByteArrayMessage()`. And `handleStringMessage()`...

My- point is that we should allow to modify file only when it is locked by a single process for appending.

We might need to refactor those internal `WhileLockedProcessor` instance to be able to perform this our new `newFileCallback`

src/reference/asciidoc/file.adoc
```diff
@@ -616,6 +616,9 @@ outbound gateway by using the XML namespace.
 
 Starting with version 4.3, you can specify the buffer size to use when writing files.
 
+Starting with version 5.1, you can provide a `newFileCallback` which will be triggered if you are using `FileExistsMode.APPEND` or `FileExistsMode.APPEND_NO_FLUSH` and a new file has to be created.
```
artembilan: We need also to mention what type it is.

src/reference/asciidoc/file.adoc
```diff
@@ -616,6 +616,9 @@ outbound gateway by using the XML namespace.
...
+This callback receives the newly created file and the message which triggered it. This callback could be used to write a CSV header defined in the message header for an example.
```
artembilan: Please, follow "Sentence per line" rule in the docs: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

And I believe there has to be a comma before `for an example`...

artembilan (changes requested):
I hope you have seen my other comments about `WhileLockedProcessor`...

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
alturkovic: I actually did it there the first time and got stuck on the implementation... Since state could have a `BufferedOutputStream` or `BufferedWriter` depending on the input type. Would you mind giving me an advice on how to handle this best?

src/reference/asciidoc/file.adoc
alturkovic: type as in `BiConsumer<File, Message<?>>` ?

artembilan: Correct, but it is not reflected in the Docs. Isn't it?

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
artembilan: Not sure what you mean: you have that `BiConsumer<File, Message<?>>`. That's fully enough to be called from any `WhileLockedProcessor`. 
I wouldn't bother myself with its details: only the point that we can call such a `newFileCallback` from a guarded by the lock block and we definitely have to reuse the lock state.

alturkovic: Oh my bad, I thought I should pass the opened writer, will change.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -159,6 +161,8 @@
 
 	private Set<PosixFilePermission> permissions;
 
+	private volatile BiConsumer<File, Message<?>> newFileCallback;
```
artembilan: We don't use `volatile` for setter's properties.
It make sense only for those properties which can be modified from different thread, e.g. in the `start()`/`stop()` callback.
But not for setters, please.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -443,6 +447,19 @@ public void setChmod(int chmod) {
 		this.permissions = permissions;
 	}
 
+	/**
+	 * Set the callback to use when creating new files. This callback will only be called
+	 * if {@link #fileExistsMode} is {@link FileExistsMode#APPEND} or {@link FileExistsMode#APPEND_NO_FLUSH}
+	 * and new file has to be created. The callback receives the new result file and the message that
+	 * triggered the handler.
+	 * @since 5.1
```
artembilan: `@since` must go in the end of method Javadoc. Well, before `@deprecate` though, but still.
Here is tag order:
```
<property name="tagOrder" value="@param, @return, @throws, @since, @deprecated, @see"/>
```

And no blank lines in the method Javadocs, please: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

src/reference/asciidoc/whats-new.adoc
```diff
@@ -86,6 +87,13 @@ Starting with version 5.1, you must explicitly turn on the `@Publisher` AOP func
 
 See "`<<publisher-annotation>>`" for more information.
 
+[[x5.1-file-writing-message-handler]]
+=== `FileWritingMessageHandler`: New file callback
```
artembilan: This is not a `General Change`, but exactly for `file` module.
So, it must go its own section before or after the mention above in this doc a `FTP and SFTP Changes`.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -773,6 +801,11 @@ private File handleStringMessage(final String content, File originalFile, File t
 
 			@Override
 			protected void whileLocked() throws IOException {
+				if (newFileCallback != null && !fileToWriteTo.exists() && (FileExistsMode.APPEND.equals(fileExistsMode)
```
artembilan: I think this can be factored to some common code, but let it be for now.

I'll take a look into that during merge and polishing...

Let's bring it into the working shape first of all.

So, far I think you need to add a `FileWritingMessageHandler.this.` prefix into this property used in the inner classes.

artembilan (changes requested):
This is the whole failures report from the Travis:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:667:45: Reference to instance variable 'newFileCallback' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:668:49: Reference to instance variable 'newFileCallback' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:756:37: Reference to instance variable 'newFileCallback' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:756:121: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:757:90: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:758:41: Reference to instance variable 'newFileCallback' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:804:37: Reference to instance variable 'newFileCallback' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:804:121: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:805:90: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:806:41: Reference to instance variable 'newFileCallback' needs "FileWritingMessageHandler.this.". [RequireThis]
```

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
alturkovic: I just copied `volatile` since all non-final properties except `permissions` are `volatile` as well, didn't know, will fix.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
alturkovic: My bad

src/reference/asciidoc/whats-new.adoc
alturkovic: Is it okay if I just remove this lines? I already added a section for it: "`FileWritingMessageHandler`: New file callback", I believe I shouldn't have added it to the `General Change` list.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
alturkovic: I have seen `this` codestyle errors, didn't know about `FileWritingMessageHandler.this` syntax, TIL :)

Of course,  some generalization could certainly be made, thanks!

src/reference/asciidoc/whats-new.adoc
artembilan: OK. Never mind.
I'll polish this on merge as well and will share here with you what I mean.

Please, fix the Checkstyle violations and we are good to go with merge!

alturkovic: Sure, appreciated :)

artembilan (changes requested):
You still are missing to fix all the Checkstyle violations:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:755:152: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:756:90: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:803:152: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java:804:90: Reference to instance variable 'fileExistsMode' needs "FileWritingMessageHandler.this.". 
```

Please, consider to run `./gradlew clean check --parallel` before pushing commits.

Although I agree that you can do that only for the `file` module so far: `./gradlew clean :spring-integration-file:check`  😄

alturkovic:
Thanks for the gradle command, I am new to gradle, used to maven, will fix

artembilan (approved):
LGTM, although as we talked there is still need some polishing.

@garyrussell,

would you mind to take a quick look to be sure that we are on the right path?

Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -621,19 +638,20 @@ private File retrieveOriginalFileFromHeader(Message<?> message) {
 		return null;
 	}
 
-	private File handleFileMessage(final File sourceFile, File tempFile, final File resultFile) throws IOException {
+	private File handleFileMessage(final File sourceFile, File tempFile, final File resultFile, final Message<?> requestMessage)
+			throws IOException {
 		if (!FileExistsMode.APPEND.equals(this.fileExistsMode) && this.deleteSourceFiles) {
```
garyrussell: Looks like we missed `APPEND_NO_FLUSH` here (original bug - not related to this PR).

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -645,6 +663,10 @@ private File handleInputStreamMessage(final InputStream sourceFileInputStream, F
 
 				@Override
 				protected void whileLocked() throws IOException {
+					if (FileWritingMessageHandler.this.newFileCallback != null && !fileToWriteTo.exists()) {
+						FileWritingMessageHandler.this.newFileCallback.accept(fileToWriteTo, requestMessage);
```
garyrussell: Since this is called for a raw `InputStream` payload we need `&& append` here.

OTOH, I wonder why we only want to restrict this to be only called when appending. Perhaps we should always call it when a new file is about to be created. It would need some more logic in `handleFileMessage`. i.e. only rename the input file if we are not in an append mode, and want to delete the inbound file, and we **don't** have a `newFileCallback`.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -159,6 +161,8 @@
 
 	private Set<PosixFilePermission> permissions;
 
+	private BiConsumer<File, Message<?>> newFileCallback;
```
garyrussell: Perhaps it should be a `BiFunction<File, Message<?>, File>` with the callback returning the file we will actually write to.

garyrussell (commented):
@alturkovic Thanks for the contribution. Just a couple of comments for consideration.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
artembilan: Good catch!
But looks like it has to be addressed in the separate issue/PR...

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
alturkovic: I am not sure I am following. This PR was supposed to handle writing some content into a newly created file. Why/when would we return a different file from the function? Thanks for the clarification.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
garyrussell: Oops; yes; you are correct; sorry for the noise.

Ok, @artembilan let's leave as-is (append mode only).

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
garyrussell: Let's stick with `BiConsumer` - no need to add complexity.

artembilan: I don't know what scenario may do such a mutation outside of the `FileWritingMessageHandler`, but let it is such a `BiFunction`!
The target writing is going to happen on the resulting file.
On the other hand we are going to to get such a mutated file only once: when a new file is created. All other subsequent appends are going to happen in the original file, evaluated by the `FileWritingMessageHandler`.
So, may be it is not so good idea to let end-user to mutate target file unconditionally.

garyrussell: I was just generalizing - someone might want to change the file location for certain messages (not just write a header). But they can do that with directory expression and filename generator.

artembilan:
OK. Sounds like we have reached consensus 😉 

Let me merge it and polish! In the evening: I need to go soon to see how my doctor is doing...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/9ac0746f1940ff3b9324c6ebc53b9089563ed356 with the polishing like: https://github.com/spring-projects/spring-integration/commit/44b7df94c13c5e93aa36e10c6f2e5a959c360292

Thank you for contribution.
Looking forward for more!

PR #2512, opened 2018-07-23
INT-4503: Fix assertion in HttpMessageHandlerSpec

imitbn:
JIRA: https://jira.spring.io/browse/INT-4503

Custom ErrorHandler can't be used if external RestTemplate is provided

* Replace isClientSet() with !isClientSet() for isTrue assertion in errorHandler method

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
@@ -34,6 +34,7 @@ import org.springframework.web.client.RestTemplate;
  *
  * @author Artem Bilan
  * @author Shiliang Li
+ * @author Oleksii Komlyk
  *
  * @since 5.0
  *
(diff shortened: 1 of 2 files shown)


spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
```diff
@@ -75,8 +75,7 @@ public HttpMessageHandlerSpec requestFactory(ClientHttpRequestFactory requestFac
 	 * @return the spec
 	 */
 	public HttpMessageHandlerSpec errorHandler(ResponseErrorHandler errorHandler) {
-		Assert.isTrue(this.isClientSet(),
-				"the 'errorHandler' must be specified on the provided 'restTemplate'");
+		Assert.isTrue(!isClientSet(), "the 'errorHandler' must be specified on the provided 'restTemplate'");
```
artembilan: Good. Thank you!

Would you mind to modify `HttpDslTests` to cover this change?

Also consider, please, fix `requestFactory()` condition above to use this `isClientSet()` as well.

And add you name to the `@author` list of this class and test one, if you really are going to modify that as well.

artembilan (changes requested):

artembilan (approved):
@imitbn ,

you change has slipped between our fingers because you squashed commits.
Would be great to just add new commits and we will got a notification for new changes.

So far your fix looks good and we are merging it...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/865e2162e520f8be6b26f81472a1ced22e8e1c0f and cherry-picked to `5.0.x`

Looking forward for more!

PR #2513, opened 2018-07-23
INT-4504: A recursive FileListFilter implementation

alturkovic:
https://jira.spring.io/browse/INT-4504

A recursive FileListFilter with a delegate to pass validation to.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CollectFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CollectFileListFilter.java
@@ -16,35 +16,27 @@
 
 package org.springframework.integration.file.filters;
 
-import java.io.File;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+
+import org.springframework.util.Assert;
 
 /**
- * Simple {@link FileListFilter} implementation which runs recursively
- * and passes all found files to the delegate to filter.
+ * The {@link CompositeFileListFilter} extension which collects the result of all filters.
  *
+ * @param <F> The type that will be filtered.
  * @author Alen Turkovic
  * @since 5.1
  */
-public class RecursiveFileListFilter implements FileListFilter<File> {
-
-	private final FileListFilter<File> delegate;
-
-	public RecursiveFileListFilter(final FileListFilter<File> delegate) {
-		this.delegate = delegate;
-	}
+public class CollectFileListFilter<F> extends CompositeFileListFilter<F> {
 
 	@Override
-	public List<File> filterFiles(final File[] files) {
-		return Stream.of(files)
-				.flatMap(file -> {
-					if (file.isDirectory()) {
-						return filterFiles(file.listFiles()).stream();
-					}
-					return this.delegate.filterFiles(new File[]{file}).stream();
-				})
-				.collect(Collectors.toList());
+	public List<F> filterFiles(F[] files) {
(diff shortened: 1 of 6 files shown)


artembilan:
Thank you for contribution!

Although I wonder why existing `RecursiveDirectoryScanner` doesn’t work for you: https://docs.spring.io/spring-integration/docs/5.0.6.RELEASE/reference/html/files.html#file-reading. See `Directory scanning and polling` section

alturkovic:
[Here](https://gist.github.com/alturkovic/47302e08ff61380497c33fbeeb715594) is a Gist demonstrating how this `FileListFilter` could be used to perform complex recursive scans only on certain directories instead of the whole root directory.

This might be also possible with the `RecursiveDirectoryScanner` but I have not managed to do it myself, apologies if this is a duplicated behavior.

artembilan:
>  scans only on certain directories instead of the whole root directory

How is that possible with the `FileReadingMessageSource` which can be configured only with the root directory?
I don't how your solution provides some gain since all the trick in your sample is really done by the `SimplePatternFileListFilter("matching-folder")` as a first member of the wrapping `ChainFileListFilter`. You don't pass `foo` and `matching-a` to your `RecursiveFileListFilter` at all because they don't have a `matching-folder` in their name. Isn't it?

Your goal can be achieved just with really some custom filter to obtain a parent of the current file to pass or not.

On the other hand with the `filterFiles(file.listFiles())` logic you break a `FileListFilter` contract a bit. It is really only for checking files it is requested with, but with that extra `listFiles()` you make the whole directory scanning solution as broken and unaware of new files from sub-directories.
Although that's might be just my vision since I don't feel your feature pretty well.

Thanks for understanding and let's think more!

alturkovic:
> How is that possible with the FileReadingMessageSource which can be configured only with the root directory?

Well, not necessarily, using the `ChainFileListFilter` 'trick'. I needed a more complex file filter for my project and that's why I wrote the `RecursiveFileListFilter`. I do agree though, that it breaks the contract a bit and perhaps I should document this better. 

You are right, my example was bad, I misunderstood what `CompositeFileListFilter` does and I had to add 2 more implementations:

- CollectFileListFilter - collect the result of all registered filters

- DirectoryFileListFilter - non-recursive directory scanning `FileListFilter`

 to do what I want and hopefully this example makes more sense...

Here is a directory tree:
```
/
+- opt/
      +- c.raw
      +- redis.conf
      +- lib/
            +- a.jar
            +- b.jar
            +- backup/
                            +- a-old.jar
+- tmp/
      +- a.txt
      +- b.txt
      +- a.raw
      +- custom/
               +- b.raw 
```

I need to find `redis.conf` in `/opt/`, all `.jar` files in `/opt/lib/`, but not in `/opt/lib/backup/`, and all `.raw` files anywhere in `/tmp`.

[Here ](https://gist.github.com/alturkovic/942ef1f19474570a80b5d244ed4baf9d) is how I would do it using these`FileListFilters`...

WDYT? Too 'hacky'?

garyrussell:
I am not comfortable with this.

As @artembilan says, this is abusing the `FileListFilter` contract in that it adds files rather than filtering them.

Of course, you are at liberty to do what you want, but I don't think it belongs in the framework (at least in this form).

alturkovic:
No problem, appreciate the feedback and sorry for the noise.

artembilan:
Thank you for your effort, @alturkovic !

We are closing this as `Won't Fix` since, as you see, we don't find it as a valuable feature for the Framework.

You still are free to come to us for any other contribution!

PR #2518, opened 2018-07-25
INT-4506: Splitter: close stream and Closeables

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4506

Capture the actual splitter result in the `FunctionIterator and close it if necessary.

**cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
@@ -21,25 +21,67 @@ import java.io.IOException;
 import java.util.Iterator;
 import java.util.function.Function;
 
+import org.springframework.lang.Nullable;
+
 /**
  * An {@link Iterator} implementation to convert each item from the target
  * {@link #iterator} to a new object applying the {@link #function} on {@link #next()}.
  *
  * @author Artem Bilan
  * @author Ruslan Stelmachenko
+ * @author Gary Russell
  * @since 4.1
  */
 public class FunctionIterator<T, V> implements CloseableIterator<V> {
+
+	private final Object root;
 
 	private final Iterator<T> iterator;
 
 	private final Function<? super T, ? extends V> function;
 
+	/**
+	 * Construct an instance with the provided iterable and function.
+	 * @param iterable the iterable.
+	 * @param function the function.
+	 * @deprecated - use {@link #FunctionIterator(Object, Iterable, Function)}
+	 */
+	@Deprecated
 	public FunctionIterator(Iterable<T> iterable, Function<? super T, ? extends V> function) {
-		this(iterable.iterator(), function);
+		this(null, iterable.iterator(), function);
 	}
 
+	/**
+	 * Construct an instance with the provided root object, iterable and function.
+	 * @param root the root object.
+	 * @param iterable the iterable.
+	 * @param function the function.
+	 * @since 5.0.7
+	 */
+	public FunctionIterator(@Nullable Object root, Iterable<T> iterable, Function<? super T, ? extends V> function) {
+		this(null, iterable.iterator(), function);
+	}
+
+	/**
+	 * Construct an instance with the provided iterator and function.
+	 * @param iterator the iterator.
+	 * @param function the function.
+	 * @deprecated - use {@link #FunctionIterator(Object, Iterator, Function)}
+	 */
+	@Deprecated
 	public FunctionIterator(Iterator<T> newIterator, Function<? super T, ? extends V> function) {
+		this(null, newIterator, function);
+	}
+
+	/**
+	 * Construct an instance with the provided root object, iterator and function.
+	 * @param root the root object.
+	 * @param iterator the iterator.
+	 * @param function the function.
+	 * @since 5.0.7
+	 */
+	public FunctionIterator(@Nullable Object root, Iterator<T> newIterator, Function<? super T, ? extends V> function) {
+		this.root = root;
 		this.iterator = newIterator;
 		this.function = function;
 	}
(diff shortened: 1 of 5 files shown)


artembilan:
There are some problems with Javadocs:
```
/home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java:67: error: @param name not found
	 * @param iterator the iterator.
	          ^
/home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java:72: warning: no @param for newIterator
	public FunctionIterator(Iterator<T> newIterator, Function<? super T, ? extends V> function) {
	       ^
/home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java:79: error: @param name not found
	 * @param iterator the iterator.
	          ^
/home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java:83: warning: no @param for newIterator
	public FunctionIterator(@Nullable Object root, Iterator<T> newIterator, Function<? super T, ? extends V> function) {
	       ^
```

See Travis report.

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
```diff
@@ -64,6 +106,27 @@ public void close() throws IOException {
...
+		else if (this.iterator instanceof AutoCloseable) {
+			try {
+				((AutoCloseable) this.iterator).close();
+			}
+			catch (Exception e) {
+				// NOSONAR
+			}
+		}
+		if (!this.root.equals(this.iterator)) {
+			if (this.root instanceof Closeable) {
```
artembilan: See comment in JIRA: https://jira.spring.io/browse/INT-4506?focusedCommentId=161106&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-161106

artembilan (changes requested):

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
```diff
@@ -64,6 +106,11 @@ public void close() throws IOException {
 		if (this.iterator instanceof Closeable) {
 			((Closeable) this.iterator).close();
 		}
+		if (!this.root.equals(this.iterator)) {
+			if (this.root instanceof Closeable) {
```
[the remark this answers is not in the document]
garyrussell: Not enough coffee today 😄

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
```diff
@@ -61,8 +102,23 @@ public V next() {
...
+				// NOSONAR
+			}
+		}
+		if (!this.root.equals(this.iterator)) {
+			if (this.root instanceof AutoCloseable) {
+				try {
+					((AutoCloseable) this.root).close();
+				}
+				catch (Exception e) {
+					// NOSONAR
```
artembilan: I believe we can change a `CloseableIterator` to the `AutoCloseable` now as well...

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
```diff
@@ -61,8 +102,23 @@ public V next() {
...
+		if (this.iterator instanceof AutoCloseable) {
+			try {
+				((AutoCloseable) this.iterator).close();
+			}
+			catch (Exception e) {
+				// NOSONAR
+			}
+		}
+		if (!this.root.equals(this.iterator)) {
+			if (this.root instanceof AutoCloseable) {
```
artembilan: Conditions can be combined to a single one.

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
```diff
@@ -16,30 +16,71 @@
...
 
 	private final Function<? super T, ? extends V> function;
 
+	/**
+	 * Construct an instance with the provided iterable and function.
+	 * @param iterable the iterable.
+	 * @param function the function.
+	 * @deprecated - use {@link #FunctionIterator(Object, Iterable, Function)}
+	 */
+	@Deprecated
```
artembilan: Why is that? What is wrong with the regular ctor without a `root`?
Please, see your change into the `XPathMessageSplitter`. Looks like the `root` is not required and there is no reason to enforce end-users to think about the `root` where it isn't necessary.

Plus, let's consider a better Javadocs what is this `root` about here.

And that's why I really would prefer a wrapping to the `CloseableIterator` instead...


Also it's not visible here, but looks like we now can remove `remove()` method: it is a `default` one on the `Iterator`

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java
garyrussell: My concern is this:

```java
public class Foo implements Iterator, Closeable {

...

}

Since we have no control over what the user returns, I believe we **have** to capture the root object so we can close it when the iteration is complete.

I deprecated the other constructors to discourage their use.

But I can see a better way now; will rework...

**Review by artembilan — APPROVED** [review]
LGTM.
Will merge after green Travis report...

File: spring-integration-core/src/main/java/org/springframework/integration/util/FunctionIterator.java [hunk 6]
```diff
@@ -16,30 +16,68 @@
...
+	}
+
+	/**
+	 * Construct an instance with the provided root object, iterable and function.
+	 * @param closeable an {@link AutoCloseable} to close when iteration is complete.
+	 * @param iterable the iterable.
+	 * @param function the function.
+	 * @since 5.0.7
+	 */
+	public FunctionIterator(@Nullable AutoCloseable closeable, Iterable<T> iterable,
```
artembilan: This `closeable` is not propagated to the target ctor.
Will fix on merge.

garyrussell: Doh - thanks.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/bebde260db912c4a7065657ae1325150ddd3d047 and cherry-picked to `5.0.x`.

PR #2519, opened 2018-07-25
Upgrade to Gradle 4.9; fix RotatingServersTests

garyrussell:
Stop the flow before examining the stream to avoid `ConcurrentModificationException`.

**cherry-pick to 5.0.x**


artembilan:
... and cherry-picked to `5.0.x` after resolving minor conflicts with the wrapper jar and properties.

PR #2521, opened 2018-07-26
Upgrade to the latest Kotlin

artembilan:
* Fix more recursive generics on factory methods
See https://jira.spring.io/browse/INT-4458

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageChannels.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageChannels.java
@@ -16,12 +16,12 @@
 
 package org.springframework.integration.dsl;
 
-import java.util.Queue;
-import java.util.concurrent.Executor;
-
 import org.springframework.integration.store.ChannelMessageStore;
 import org.springframework.integration.store.PriorityCapableChannelMessageStore;
 import org.springframework.messaging.Message;
+
+import java.util.Queue;
+import java.util.concurrent.Executor;
 
 /**
  * @author Artem Bilan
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageChannels.java
```diff
@@ -101,27 +101,24 @@ public static PriorityChannelSpec priority(PriorityCapableChannelMessageStore me
 	}
 
 	public static PriorityChannelSpec priority(String id,
-			PriorityCapableChannelMessageStore messageGroupStore, Object groupId) {
+											   PriorityCapableChannelMessageStore messageGroupStore, Object groupId) {
 		return priority(messageGroupStore, groupId).id(id);
```
garyrussell: Checkstyle doesn't allow this style of indentation; will fix on merge.

artembilan: Huh? Looks like this is an artifact after upgrading to a new IDEA. Sorry

I believe we need to start moving to something like https://github.com/spring-io/spring-javaformat/ ...

Thanks BTW

garyrussell:
Merged as 9d17e09f31ef57f5fe49746f09bc1e7de617315f cherry-picked as 2709bc142a178ee58bc6b4982a98cfa7c61c1f5b after resolving conflicts.

spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageChannels.java
garyrussell: (Fixed import order too).

garyrussell: I don't like the javaformat line limit

>The source formatter intentionally uses a low character count of 90 chars for wrapping. If you’re used to longer lines, this can take some getting used to. Specifically, if you have many nesting levels things can start to look quite bad.

PR #2522, opened 2018-07-26
New AbstractAggregatingMessageGroupProcessor for grouped Messages

alturkovic:
Jira:  https://jira.spring.io/browse/INT-4508

Currently, Spring Integration ships with `SimpleMessageGroupProcessor` which releases Messages separately.

`DefaultAggregatingMessageGroupProcessor` releases an aggregated Message holding the payloads of all Messages.

A new `AbstractAggregatingMessageGroupProcessor` could be useful for releasing a single aggregated Message holding all Messages. This is very useful for conflicting headers since `DefaultAggregatingMessageGroupProcessor` removes all conflicting headers.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/MessageTransformingAggregatingMessageGroupProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/MessageTransformingAggregatingMessageGroupProcessor.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.aggregator;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+import org.springframework.integration.IntegrationMessageHeaderAccessor;
+import org.springframework.integration.store.MessageGroup;
+import org.springframework.integration.support.MessageBuilderFactory;
+import org.springframework.messaging.Message;
+
+/**
+ * This implementation of MessageGroupProcessor will take the messages from the MessageGroup
+ * and pass them on in a single message with a Collection of transformed Messages as a payload.
+ *<p>
+ * By default, this implementation does not transform the payload of the messages but simply removes
+ * {@link IntegrationMessageHeaderAccessor#SEQUENCE_SIZE}, {@link IntegrationMessageHeaderAccessor#SEQUENCE_NUMBER} and
+ * {@link IntegrationMessageHeaderAccessor#CORRELATION_ID}.
+ *
+ * @author Alen Turkovic
+ * @since 5.1
+ */
+public class MessageTransformingAggregatingMessageGroupProcessor extends AbstractAggregatingMessageGroupProcessor {
+
+	private BiFunction<Message<?>, MessageBuilderFactory, Message<?>> projection = (m, factory) -> factory
+			.fromMessage(m)
+			.removeHeaders(
+					IntegrationMessageHeaderAccessor.SEQUENCE_SIZE,
+					IntegrationMessageHeaderAccessor.SEQUENCE_NUMBER,
+					IntegrationMessageHeaderAccessor.CORRELATION_ID)
+			.build();
+
+	/**
+	 * Set the projection function used to convert messages when aggregating them. The function receives the {@link Message} to be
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aggregator/MessageTransformingAggregatingMessageGroupProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+ * {@link IntegrationMessageHeaderAccessor#CORRELATION_ID}.
+ *
+ * @author Alen Turkovic
+ * @since 5.1
+ */
+public class MessageTransformingAggregatingMessageGroupProcessor extends AbstractAggregatingMessageGroupProcessor {
+
+	private BiFunction<Message<?>, MessageBuilderFactory, Message<?>> projection = (m, factory) -> factory
+			.fromMessage(m)
+			.removeHeaders(
```
artembilan: I'm a bit reluctant to introduce some transformation functionality into the correlation handler.
It's really not its responsibility. Plus I'm not sure why would one need to remove all these useful headers if you really are going to return a payload as a `List<Message<?>>`. You are going to iterate over them downstream anyway, so `IntegrationMessageHeaderAccessor.SEQUENCE_NUMBER` would be great indicator for each of them.

I can see only much simple approach to really build single `payload` from the `List<Message<?>>` and that's all:
````java
public class CollectionAggregatingMessageGroupProcessor implement MessageGroupProcessor {
    public Object processMessageGroup(MessageGroup group) {
              return MessageBuilder.withPayload(group.getMessages());
    }
}
````

All other hard transforming logic I would leave for the downstream flow...

File: spring-integration-core/src/test/java/org/springframework/integration/aggregator/MessageTransformingAggregatingMessageGroupProcessorTest.java [hunk 2]
```diff
@@ -0,0 +1,107 @@
...
+import org.springframework.integration.store.MessageGroup;
+import org.springframework.integration.store.SimpleMessageGroup;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+
+/**
+ * @author Alen Turkovic
+ * @since 5.1
+ */
+public class MessageTransformingAggregatingMessageGroupProcessorTest {
```
artembilan: Our pattern is `...Tests`

artembilan (changes requested):
Thank you for contribution, but I'm not sure yet that we are going to merge this...

spring-integration-core/src/main/java/org/springframework/integration/aggregator/MessageTransformingAggregatingMessageGroupProcessor.java
alturkovic: `IntegrationMessageHeaderAccessor.SEQUENCE_NUMBER` might be a mistake, you are right.

Transformation projection might have also been a mistake, initially, I have just removed `IntegrationMessageHeaderAccessor.SEQUENCE_SIZE` and `IntegrationMessageHeaderAccessor.CORRELATION_ID` since they are written in the aggregated message headers, and extracted the projection later which might have been a mistake...

artembilan: Well, with my idea you don't need any aggregation logic at all: all the information we need is present in each message in the payload. Isn't it?

alturkovic: I'm not sure I understand the framework well enough so I am afraid I am not following you, sorry.

```
  @Bean
  public IntegrationFlow out() {
    return IntegrationFlows.from("channel-out")
      // a custom aggregating processor so we keep headers of each message
      // removed duplicated correlation headers
      // by default, we would lose the 'from' header
      .aggregate(a -> a.outputProcessor(new MessageTransformingAggregatingMessageGroupProcessor()))
      .log()
      .get();
  }
```
This is an example of how I used it for testing, the only reason I removed these headers is that they were (as I initially though, but might be wrong) unnecessarily duplicated for each message since they are also written in the aggregated result.

alturkovic: Thanks for the thorough explanation, that helps clarify things a lot :)

I thought these headers were not necessary and added noise, but reading your explanation this is clearly not the case. 

Sorry, please close the PR.

artembilan:
@garyrussell ,

Would you mind to take a look and share your vision?

It's really easy to close PR or merge it, but we really need some pros and cons arguments either way...

Thanks

garyrussell:
I concur with using `.outputProcessor(g -> new GenericMessage<>(g.getMessages()))`

or even `.outputProcessor(g -> new GenericMessage<>(g))` - there may be useful other information in the group that's usable downstream (e.g. `group.getLastModified`). 

You also may want to get messages one-at-a-time lazily (if persistent) rather than fetching them all via `getMessages()`.

garyrussell:
Actually, no; that won't work (with a persistent group) - because it would be removed.

artembilan:
Having the quorum we are closing this as `Invalid` with some simple off-box solution via lambda.

Thank you for your effort @alturkovic anyway !

That's definitely how I like to do things: have some discussion and sharing the doubts. Only this way we can make our product as perfect as possible 😄 

You are still welcome to come back to us with any other contribution.

PR #2526, opened 2018-07-27
INT-4510: Test no memory leak in FluxMessageChannel

artembilan:
JIRA: https://jira.spring.io/browse/INT-4510

Related to https://github.com/reactor/reactor-core/issues/1290

The `Flux.publish()` and subsequent `connect()` doesn't fuse a subscriber
for the hooks flow is interrupted (complete, or error, or disconnect).
In this case the `FluxMessageChannel.publishers` store is not cleared
from the finished publishers

* Add test-case to check the `FluxMessageChannel.publishers` store after
finishing the stream
* Add `hide()` operator with TODO to remove when an appropriate Reactor
version is ready

**Cherry-pick to 5.0.x**


garyrussell:
Cherry-picked as 7e8991a45ea2603f0dfdd2c18b5af77ab2268299 after resolving conflicts.

PR #2528, opened 2018-07-27
Removed Docbook from document production

Buzzardo:
I made the document production part of the build use only Asciidoctor to produce HTML, PDF, and Epub. The goal was to remove Docbook from our processes. I also moved the document-production part of the build (though only for the reference guide) to a separate build file. That file is in src/reference.


src/reference/asciidoc/images/epub-cover.svg
```diff
@@ -6,7 +6,11 @@
   <text x="250" y="192" fill="Black" font-size="72">Reference Guide</text>
   <text x="250" y="420" fill="Black" font-size="36">Mark Fisher, Marius Bogoevici,</text>
   <text x="250" y="466" fill="Black" font-size="36">Iwein Fuld, Jonas Partner,</text>
+<<<<<<< HEAD
```
cppwfs: oops.   Need to remove the merge comments ;-)

Buzzardo: Thanks, Glenn. I fixed it. The aggravating thing is that I searched for HEAD with grep and didn't find it. Sigh.

src/reference/build.gradle
```diff
@@ -0,0 +1,72 @@
+buildscript {
+  dependencies {
+    classpath 'org.asciidoctor:asciidoctorj-epub3:1.5.0-alpha.6'
+    classpath 'org.asciidoctor:asciidoctorj-pdf:1.5.0-alpha.15'
```
artembilan: Aren't there some really GA versions for these plugins?

src/reference/build.gradle
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+plugins {
+  id 'de.undercouch.download' version '3.4.3'
+  id 'org.asciidoctor.convert' version '1.5.3'
+}
+
+apply plugin: 'java'
```
artembilan: Why do we need Java plugin if there is no any Java 🤷‍♂️

src/reference/build.gradle
```diff
@@ -0,0 +1,72 @@
...
+apply plugin: 'org.asciidoctor.convert'
+
+version = '1.0.0-SNAPSHOT'
```
artembilan: What is this version about?

src/reference/build.gradle
```diff
@@ -0,0 +1,72 @@
...
+
+asciidoctorj {
+  version = '1.5.5'
+}
+
+task downloadResources(type: Download) {
+  src 'http://repo.spring.io/libs-snapshot-local/org/springframework/doc-resources/spring-doc-resources/0.1.1-SNAPSHOT/spring-doc-resources-0.1.1-SNAPSHOT.zip'
```
artembilan: What is this about?
Can't we at least extract a variable for the version of this artifact?

src/reference/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-bin.zip
```
artembilan: That's something is wrong. I don't believe that we need a separate Gradle just for docs...

artembilan (changes requested):
Here is some review so far.

build.gradle
```diff
@@ -932,6 +898,8 @@ task depsZip(type: Zip, dependsOn: distZip) { zipTask ->
 	}
 }
 
+apply plugin: 'maven'
```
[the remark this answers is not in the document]
Buzzardo: I got an error about the `artifacts` line if I didn't have it.

build.gradle
```diff
@@ -749,44 +745,14 @@ project("spring-integration-bom") {
...
-			toc2: '',
-			'compat-mode': '',
-			imagesdir: '',
-			stylesdir: "stylesheets/",
-			stylesheet: 'golo.css',
-			'spring-integration-version': "$version",
-			'source-highlighter': 'highlightjs'
-		]
-	]
+task reference(type: GradleBuild) {
```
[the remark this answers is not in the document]
Buzzardo: We could move the Javadoc over there, too.

src/reference/build.gradle
Buzzardo: Not that I can find. I experimented with a number of versions. This the "magic" combination that works. Asciidoctor-epub is especially fragile and requires some serious hoop-jumping (mostly in the attributes) to get to work.

src/reference/build.gradle
Buzzardo: I don't know. I inherited that from the script I borrowed to get started. It may be unnecessary. Or it may be necessary. I'll have to test.

build.gradle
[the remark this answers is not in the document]
artembilan: No, Javadocs should be a part of the regular development. They are done at the same time as regular source code.
Or I just don't understand what you are talking about...

src/reference/build.gradle
Buzzardo: No clue. I inherited that from the script I borrowed to get started. I think it can be removed. I'll have to test.

src/reference/build.gradle
Buzzardo: Putting it in a variable would just add another line.

build.gradle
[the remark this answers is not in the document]
artembilan: Right, that's how reference must be as a sub-module or at least as an import script, but fully not a separate project as it is currently in your PR...

src/reference/gradle/wrapper/gradle-wrapper.properties
Buzzardo: That was auto-generated.

src/reference/build.gradle
artembilan: That sounds creepy. Are we really going to spend a lot of time coming up with the solution when dependencies we rely on are fragile ?

BTW I don't see that you have removed Dookbook artifacts, e.g. `index-docinfo.xml`.

Am I missing anything again?

Buzzardo: I forgot to remove that. Other than the Docbook-specific parts of the build, that file is the only Docbook-specific artifact. I'll zap it.

garyrussell:
Just ran a quick build. A couple of observations:

1. I don't see a copyright 
1. I think we should dump the output into `../../build` rather than `src/reference/build`.
1. I see a bunch of error messages...

```
> Task :asciidoctor
Converting /Users/grussell/Development/spring-integration/src/reference/build/working/index.adoc
Converting /Users/grussell/Development/spring-integration/src/reference/build/working/index.adoc
Converting /Users/grussell/Development/spring-integration/src/reference/build/working/index.adoc
uri:classloader:/gems/rubyzip-1.1.7/lib/zip.rb:35: warning: ObjectSpace impacts performance. See http://wiki.jruby.org/PerformanceTuning#dont-enable-objectspace
asciidoctor: ERROR: index.adoc: line 25: invalid part, must have at least one section (e.g., chapter, appendix, etc.)
asciidoctor: ERROR: index.adoc: line 36: invalid part, must have at least one section (e.g., chapter, appendix, etc.)
asciidoctor: ERROR: index.adoc: line 60: invalid part, must have at least one section (e.g., chapter, appendix, etc.)
asciidoctor: ERROR: index.adoc: line 124: invalid part, must have at least one section (e.g., chapter, appendix, etc.)
asciidoctor: ERROR: index.adoc: line 155: invalid part, must have at least one section (e.g., chapter, appendix, etc.)
conversion missing in epub3 backend for embedded
conversion missing in epub3 backend for embedded
conversion missing in epub3 backend for embedded
...
conversion missing in epub3 backend for embedded
```

garyrussell:
I am a little curious as to why this paragraph is formatted differently, I don't see anything obvious in the text...

screen shot 2018-08-01 at 2 05 23 pm

Buzzardo:
The "invalid part" errors come from having more than one heading one in the document (each book is supposed to have just one - the book title). It has no effect other than producing that error (and italicizing that item in the table of contents). I don't know what drives the other errors. Both kinds of errors come from the Epub generator, which is still in alpha.

That first paragraph being different is part of Asciidoctor's formatting choices. I'm not sure why the folks who make Asciidoctor think it's a good idea. I'll see what I can do about it.

Thanks for testing. I'll be getting back to it this week.

Buzzardo:
It turns out that Asciidoctor applies that different style to the first paragraph only when there is a single paragraph after a level-1 heading (single equal sign). I split the existing single paragraphs into two paragraphs where I could and added a new paragraph in one spot, so that problem is gone.

artembilan (changes requested):
@Buzzardo ,

sorry for late reply, but we really appreciate your effort.

Please, consider to remove all the new Gradle wrapper artifacts and SH scripts.

build.gradle
```diff
@@ -764,42 +760,14 @@ project("spring-integration-bom") {
...
-		toc2: '',
-		'compat-mode': '',
-		imagesdir: '',
-		stylesheet: "${rootProject.projectDir}/src/reference/asciidoc/stylesheets/golo.css",
-		'source-highlighter': 'highlightjs'
-	]
+task reference(type: GradleBuild) {
+  group = 'Documentation'
+	description = 'Generate the reference documentation'
+  buildFile = 'src/reference/build.gradle'
```
artembilan: Looks like you have removed too much from your changes. 
I guess that `'src/reference/build.gradle` still must be included to this change. Otherwise we just can't build the project at all.

artembilan (changes requested):

build.gradle
Buzzardo: I wondered about that but figured you know more than I do. I have put those files back now. :)

artembilan (changes requested):
@garyrussell ,

Would you mind to take a look here?
I'm not sure what am I missing and why do we need this at all.

Thanks

garyrussell:
>I'm not sure what am I missing and why do we need this at all.

Need what? The entire PR?

My understanding id this is to align all Spring documents to the newer style (and eventually we'll have toggles for different configuration options).

artembilan:
> all Spring documents to the newer style

Hm. Doesn't look like. 
The title mention a removal of Docbook conversion. However I don't understand why do we need that new script.

OK. Let's start over! 

@Buzzardo ,

What should I do locally do determine that your change is what we except?
However tell me, please, first of all what we really expect from this change?

And then I'll remove/polish whatever is necessary

Thanks

Buzzardo:
I'm trying to modernize both the document production and the document look and feel.

What I just did for spring-amqp (https://github.com/spring-projects/spring-amqp/pull/889) does it more cleanly. Both my understanding of how to do all this and the tools I can use to do it have improved, so perhaps the best thing to do is reject this PR. Then I can make a new, cleaner PR.

If you really want to have at it, the goals are as follows:

* Import the resources (stylesheet, JavaScript, and static HTML) from spring-doc-resources.
* Use those resources to generate a reference guide that has the modern look and feel (collapsing ToC on the left and the new styles).
* Remove Docbook from the process. We should have no XML or XSL left in the docs directories.

artembilan:
That's nice, @Buzzardo !

Let's finish the PR in Spring AMQP and then we can come back here.

I don't see reason in rejecting this one since you can just rebase and modify it respectively, similar way you did in Spring AMQP.

👍

artembilan:
BTW, it looks like we have a JIRA on the matter, which should be fixed with this PR: https://jira.spring.io/browse/INT-4339

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/7980afef9b035f0b77ada02702ff0ad8b4995179 after some polishing to the build script and docs resources.

A couple concerns remain:

1. PDF doesn't render version and Copyright in the right place.
2. PDF view uses bigger font for code snippets, so even if it doesn't overflow now anymore (phuh!), but the code has breaks to make it hard readable, e.g.:
image

3. The build is too slow: 3:30 for me just to build docs. Although I didn't pay attention how it was before...
4. I wonder how people may understand from where to download a PDF... It is really in the same place as a regular HTML, but you need to know how to jump to that `index.pdf` 😄 Maybe we need to mention that in our HTML version somehow if we don't provide a public link for PDF fie to download?

artembilan:
If you are interested in the result, you can click on the Reference Doc. link for the `5.1.3 SNAPSHOT` version: https://spring.io/projects/spring-integration#learn

PR #2529, opened 2018-07-27
INT-4514: Integration Graph, include dynamic flows

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4514

`getBeansOfType()` with eager init does not currently return late registered singletons.
We should not be eagerly initializing here anyway.

**cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/support/management/graph/IntegrationGraphServerTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/support/management/graph/IntegrationGraphServerTests.java
@@ -16,10 +16,7 @@
 
 package org.springframework.integration.support.management.graph;
 
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.notNullValue;
-import static org.junit.Assert.assertThat;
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/test/java/org/springframework/integration/support/management/graph/IntegrationGraphServerTests.java
```diff
@@ -119,13 +122,26 @@ public void test() throws Exception {
...
+
+	@Test
+	public void testIncludesDynamic() {
+		Graph graph = this.server.getGraph();
+		assertThat(graph.getNodes().size()).isEqualTo(32);
+		IntegrationFlow flow = f -> f.handle(m -> { });
+		IntegrationFlowRegistration reg = this.flowContext.registration(flow).register();
+		graph = this.server.rebuild();
+		assertThat(graph.getNodes().size()).isEqualTo(34);
+		this.flowContext.remove(reg.getId());
```
artembilan: You can just call `reg.destroy()`.
But this is irrelevant.

LGTM, merging...

artembilan:
... and cherry-picked to `5.0.x` after some conflict fixing.

PR #2530, opened 2018-07-27
INT-4513: Clean IntManageConf from removed beans

artembilan:
JIRA: https://jira.spring.io/browse/INT-4513

The `IntegrationManagementConfigurer` keeps metrics beans in its local
stores.
When we add beans at runtime, they are not removed from those caches
after their removal

* Implement `DestructionAwareBeanPostProcessor` in the `IntegrationManagementConfigurer`
and clean up caches according provided bean type and its name
* Improve `ManualFlowTests` and `FtpTests` to be sure that
`IntegrationManagementConfigurer` caches are cleared after destroying
 `IntegrationFlowRegistration`

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
@@ -27,7 +27,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.beans.factory.SmartInitializingSingleton;
-import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.beans.factory.config.DestructionAwareBeanPostProcessor;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.integration.core.MessageSource;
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
```diff
@@ -246,7 +247,8 @@ public void afterSingletonsInstantiated() {
 			this.metricsFactory = new DefaultMetricsFactory();
 		}
 		this.sourceConfigurers.putAll(this.applicationContext.getBeansOfType(MessageSourceMetricsConfigurer.class));
-		Map<String, IntegrationManagement> managed = this.applicationContext.getBeansOfType(IntegrationManagement.class);
+		Map<String, IntegrationManagement> managed = this.applicationContext
+				.getBeansOfType(IntegrationManagement.class);
```
garyrussell: Shouldn't we suppress eager init here too? (and other gBOT calls in this class).

garyrussell (commented):
That's all (and travis checkstyle).

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
artembilan: Well, I believe it isn't going to harm, but at the same time it is idempotent here: all `getBeansOfType()` is called only from the ` afterSingletonsInstantiated()` which is done only once.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/adbbca13fb8e1b4386cbcce9d2e014c4ca73bb88 and cherry-picked to `5.0.x` including subsequent polishing like: https://github.com/spring-projects/spring-integration/commit/169eb7498d93ae48a74293e1f1c71c9e2fc8fab5

PR #2532, opened 2018-07-29
Fix jdbc doc

artembilan:
We need to consider to replace docs for have just released version `5.0.7`.
Everything is broken after this section https://docs.spring.io/spring-integration/docs/5.0.7.RELEASE/reference/html/jdbc.html#jdbc-outbound-gateway


garyrussell:
Broken by this: https://github.com/spring-projects/spring-integration/commit/2126e9b9b5a8a31a5d6182189e235ea24178371d#diff-4d4ec1f14258ba10962491429d46d192

Mismatched block fence.

PR #2533, opened 2018-08-01
Add @NonNullApi where appropriate

garyrussell:
Packages with `@Nullable` annotations need to have `@NonNulApi` at the package level.

There may be classes in these packages that have nullable arguments/returns that are
not currently marked as `@Nullable`. These will be fixed over time.

Also suppress JSR305 warnings when compiling tests.

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisUtils.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisUtils.java
@@ -18,6 +18,7 @@ package org.springframework.integration.redis.util;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Properties;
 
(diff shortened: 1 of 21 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisUtils.java
```diff
@@ -36,9 +37,11 @@
 
 	private static final String VERSION_PROPERTY = "redis_version";
 
+	@SuppressWarnings("serial")
 	private static final Map<RedisOperations<?, ?>, Boolean> unlinkAvailable =
 			new LinkedHashMap<RedisOperations<?, ?>, Boolean>() {
 
+				@SuppressWarnings("rawtypes")
```
artembilan: Can't we fix it with an explicit generic args in the `Map.Entry` definition instead?

But thank you for catching this one!

artembilan (changes requested):
Only one concern

PR #2534, opened 2018-08-02
INT-3364: Add batchUpdate into JdbcMessageHandler

artembilan:
JIRA: https://jira.spring.io/browse/INT-3364

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageHandler.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageHandler.java
```diff
@@ -149,57 +169,99 @@ protected void onInit() throws Exception {
...
 		}
 		else {
-			int updated;
-			if (this.preparedStatementSetter != null) {
-				updated = this.jdbcOperations.getJdbcOperations().update(this.updateSql,
-						ps -> JdbcMessageHandler.this.preparedStatementSetter.setValues(ps, message));
+			if (message.getPayload() instanceof Iterable) {
+				Stream<? extends MutableMessage<?>> messageStream =
+						StreamSupport.stream(((Iterable<?>) message.getPayload()).spliterator(), false)
+								.map(payload -> new MutableMessage<>(payload, message.getHeaders()));
```
garyrussell: How about...

```java
Stream<Message<?>> messageStream =
		StreamSupport.stream(((Iterable<?>) message.getPayload()).spliterator(), false)
			.map(payload -> new Message<Object>() {

				@Override
				public Object getPayload() {
					return payload;
				}

				@Override
				public MessageHeaders getHeaders() {
					return message.getHeaders();
				}
				
			});
```

...to avoid wrapping the headers in a new map for each iteration?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageHandler.java
```diff
@@ -47,41 +52,52 @@
...
  * {@link SqlParameterSourceFactory} abstraction, the default implementation of which wraps the message so that its bean
- * properties can be referred to by name in the query string E.g.
+ * properties can be referred to by name in the query string, e.g.
  *
  * <pre class="code">
  * INSERT INTO FOOS (MESSAGE_ID, PAYLOAD) VALUES (:headers[id], :payload)
  * </pre>
  *
+ * <p>
+ * When message payload is an instance of {@link Iterable}, a
```
garyrussell: When a

src/reference/asciidoc/jdbc.adoc
```diff
@@ -261,7 +261,15 @@ public MessageHandler jdbcMessageHandler(DataSource dataSource) {
...
 From the XML configuration perspective, the `prepared-statement-setter` attribute is available on the `<int-jdbc:outbound-channel-adapter>` component.
-It lets you specify a  `MessagePreparedStatementSetter` bean reference.
+It lets you specify a `MessagePreparedStatementSetter` bean reference.
+
+==== Batch Update
+
+Starting with version 5.1, the `JdbcMessageHandler` performs a `JdbcOperations.batchUpdate()` if the payload of the request message is an `Iterable` instance.
+Each element of the `Iterable` is wrapped to the `MutableMessage` with a copy of headers from the request message.
+In case of regular `SqlParameterSourceFactory`-based configuration these messages are used to build an `SqlParameterSource[]` for an argument used in the mentioned `JdbcOperations.batchUpdate()` function.
+Inb case if `MessagePreparedStatementSetter`, a `BatchPreparedStatementSetter` variant is used to iterate over those messages for each item and provided `MessagePreparedStatementSetter` ise called against them.
```
garyrussell: s/Inb/In the/

and **the** provided

s/ise/is/

garyrussell (changes requested):
A couple of typos, and one question.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageHandler.java
artembilan: Sounds like a good optimization! 👍 

Will address a bit later.

Thanks

garyrussell (approved):

PR #2536, opened 2018-08-06
INT-4480: Use component type for ids in the flow

artembilan:
JIRA: https://jira.spring.io/browse/INT-4480

For better readability of bean names, e.g. from the metrics collectors,
generate them from the component type instead of their fully qualified
class names


garyrussell:
Needs a rebase (and travis failures).

PR #2540, opened 2018-08-15
INT-4519: Improve JMS inbound-c-a receiveTimeout

artembilan:
JIRA: https://jira.spring.io/browse/INT-4519

* Modify `DynamicJmsTemplate` to default for `-1` if we get
`CachingConnectionFactory` and it is with `cacheConsumers`, otherwise
1 second
* Polishing for some JMS XML parsers, in particular remove
`receiveTimeout` population

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/DynamicJmsTemplate.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/DynamicJmsTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,11 @@
 
 package org.springframework.integration.jms;
 
+import javax.jms.ConnectionFactory;
+
+import org.springframework.jms.connection.CachingConnectionFactory;
 import org.springframework.jms.core.JmsTemplate;
+import org.springframework.jms.support.destination.JmsDestinationAccessor;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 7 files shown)


spring-integration-jms/src/main/java/org/springframework/integration/jms/DynamicJmsTemplate.java
```diff
@@ -27,6 +31,28 @@
...
+	@Override
+	public void setConnectionFactory(ConnectionFactory connectionFactory) {
+		super.setConnectionFactory(connectionFactory);
+		if (!this.receiveTimeoutExplicitlySet) {
+			if (connectionFactory instanceof CachingConnectionFactory &&
+					((CachingConnectionFactory) connectionFactory).isCacheConsumers()) {
+				super.setReceiveTimeout(JmsDestinationAccessor.RECEIVE_TIMEOUT_NO_WAIT);
+			}
+			else {
+				super.setReceiveTimeout(1000);
```
garyrussell: Sonar will bark about a "magic" number; should be a constant.

src/reference/asciidoc/jms.adoc
```diff
@@ -63,6 +63,8 @@ If you prefer to have the raw JMS message as the Spring Integration message's pa
 ----
 ====
 
+Starting with version 5.1, a default value of the `receive-timeout` is `-1` (no wait) for the `org.springframework.jms.connection.CachingConnectionFactory` and `cacheConsumers`, otherwise it is 1 second.
```
garyrussell: Conflicts with "cherry-pick" statement in commit comment.

garyrussell (changes requested):

garyrussell:
Suspicious travis test failure.

>org.springframework.integration.jms.dsl.JmsTests > testJmsOutboundInboundFlow FAILED
    java.lang.AssertionError at JmsTests.java:179

artembilan:
Hm... Not sure what's going on...
Can it be an effect of the `JmsTemplate.RECEIVE_TIMEOUT_NO_WAIT` and `CachingConnectionFactory` with the `cacheConsumers` by default?

garyrussell:
I think it's because the outbound flow uses the non-caching CF, so the `vm://` could terminate before the container starts.

The outbound flow should use the `cachingConnectionFactory()` anyway.

garyrussell:
So, not related to this PR, but a test bug anyway.

garyrussell (approved):

PR #2541, opened 2018-08-17
Add Java DSL `.fluxTransform(Function)` operator

artembilan:
For better interoperability with Reactor `Flux` from the
end-user perspective introduce an operator which could
call a target `Function` with integration data wrapped to the `Flux`
and expect a `Publisher<?>` result to continue the flow.
This way end-user just needs to implement a `Function` (or method)
to accept the `Flux` as an input and return some `Publisher`
after the sequence of reactive operators.

Such a new operator also allows a smooth integration with the
Spring Cloud Function, where it is just enough to lookup the function
in the catalog and inject it into this operator

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
@@ -22,7 +22,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Executor;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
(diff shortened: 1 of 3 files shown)


artembilan:
@olegz ,

This is what we talked about how to avoid extra call for the `source.output()` when we mix Spring Integration Java DSL and Spring Cloud Function.
Also pay attention how I use `Mono.subscriberContext()` to propagate a `requestMessage` to be used for building reply message.

Let me know WDYT!

Thank you

/CC @smaldini

garyrussell:
Please update the commit comment with rationale.

artembilan:
Pushed. 
Thank you for the pointer!

olegz:
FWIW, I know Gary already merged it, but as Artem said this came up as he was reviewing one of the PRs for SCSt/SCF integration work, so I can definitely use it now.

PR #2544, opened 2018-08-22
INT-4497: Add rate limiting advice for handler

meherzad:
JIRA: https://jira.spring.io/browse/INT-4497

Add Rate Limiting advice for a handler which supports Rate, MaxDelay

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.handler.advice;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLongArray;
+
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessagingException;
+import org.springframework.util.Assert;
+
+/**
+ * Rate limiter advice, which will ensure call to service is Rate limited within set rate limit.
+ * Request calls breaching the rate limit will wait for time rate limit is not breached.
+ *
+ * @author Meherzad Lahewala
+ * @since 5.1
+ */
+public class RequestHandlerRateLimiterAdvice extends AbstractRequestHandlerAdvice {
+
+	private final AtomicInteger currentIndex;
+	private final long maxDelayInMillis;
+	private final long oneTimeUnitInMillis;
+	private final AtomicLongArray queue;
+	private final int rate;
+
+	/**
+	 * Create an instance of RequestHandlerRateLimiterAdvice with the provided arguments.
+	 *
+	 * @param rate the rate per time unit.
+	 * @param maxDelay the max allowable delay in time units. Use -1 for no max delay.
+	 * @param timeUnit the time unit.
+	 */
+	public RequestHandlerRateLimiterAdvice(int rate, int maxDelay, TimeUnit timeUnit) {
+		Assert.isTrue(rate > 0, "Rate should be a positive integer");
+		Assert.notNull(timeUnit, "Timeunit should be non null");
+		this.currentIndex = new AtomicInteger(0);
+		this.maxDelayInMillis = timeUnit.toMillis(maxDelay);
+		this.oneTimeUnitInMillis = timeUnit.toMillis(1);
+		this.queue = new AtomicLongArray(rate);
+		this.rate = rate;
+	}
+
+	/**
+	 * Create an instance of RequestHandlerRateLimiterAdvice with the provided arguments.
+	 *
+	 * @param rate the rate per time unit.
+	 * @param timeUnit the time unit.
+	 */
+	public RequestHandlerRateLimiterAdvice(int rate, TimeUnit timeUnit) {
+		this(rate, -1, timeUnit);
+	}
+
+	@Override
+	protected Object doInvoke(ExecutionCallback callback, Object target, Message<?> message) throws Exception {
+		delayIfNecessary(target, message);
+		return callback.execute();
+	}
+
+	private void delayIfNecessary(Object target, Message<?> message) throws InterruptedException {
+		long timeToWait = -1;
+		while (timeToWait == -1) {
+			long now = System.currentTimeMillis();
+			int index = this.currentIndex.getAndUpdate(value -> value + 1 < this.rate ? value + 1 : 0);
+			long oldValue = this.queue.get(index);
+			long waitTill;
+			if (now < oldValue) {
+				waitTill = oldValue + this.oneTimeUnitInMillis;
+			}
+			else {
+				long diff = now - oldValue;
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
```diff
@@ -0,0 +1,114 @@
...
+ * @author Meherzad Lahewala
+ * @since 5.1
+ */
+public class RequestHandlerRateLimiterAdvice extends AbstractRequestHandlerAdvice {
+
+	private final AtomicInteger currentIndex;
+	private final long maxDelayInMillis;
+	private final long oneTimeUnitInMillis;
+	private final AtomicLongArray queue;
+	private final int rate;
```
artembilan: This style is not readable. 
Would be better to surround each class member to the blank lines.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Create an instance of RequestHandlerRateLimiterAdvice with the provided arguments.
+	 *
```
artembilan: We prefer do not have blank lines in the method Javadocs: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
```diff
@@ -0,0 +1,114 @@
...
+	 * @param rate the rate per time unit.
+	 * @param maxDelay the max allowable delay in time units. Use -1 for no max delay.
+	 * @param timeUnit the time unit.
+	 */
+	public RequestHandlerRateLimiterAdvice(int rate, int maxDelay, TimeUnit timeUnit) {
```
artembilan: Can't we use a `Duration` instead ?

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
```diff
@@ -0,0 +1,114 @@
...
+		Assert.isTrue(rate > 0, "Rate should be a positive integer");
+		Assert.notNull(timeUnit, "Timeunit should be non null");
+		this.currentIndex = new AtomicInteger(0);
+		this.maxDelayInMillis = timeUnit.toMillis(maxDelay);
+		this.oneTimeUnitInMillis = timeUnit.toMillis(1);
+		this.queue = new AtomicLongArray(rate);
```
artembilan: A `queue` is not a valuable name for the variable, especially when it is an object property.

Please, come up with more self-explanatory name. 

Right, now it isn't clear what is its purpose.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
```diff
@@ -0,0 +1,114 @@
...
+	 * @param rate the rate per time unit.
+	 * @param maxDelay the max allowable delay in time units. Use -1 for no max delay.
+	 * @param timeUnit the time unit.
+	 */
+	public RequestHandlerRateLimiterAdvice(int rate, int maxDelay, TimeUnit timeUnit) {
+		Assert.isTrue(rate > 0, "Rate should be a positive integer");
+		Assert.notNull(timeUnit, "Timeunit should be non null");
+		this.currentIndex = new AtomicInteger(0);
+		this.maxDelayInMillis = timeUnit.toMillis(maxDelay);
+		this.oneTimeUnitInMillis = timeUnit.toMillis(1);
```
artembilan: What is the magic number `1 millisecond`?
Why do we need it?

Well, even if it is important, why can't we initialize it during definition?
What is the point to overhead a ctor body?

The same is about a `currentIndex` initialization here.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
```diff
@@ -0,0 +1,114 @@
...
+			logger.debug(String.format("Sleeping for %s ms", timeToWait));
+		}
+		TimeUnit.MILLISECONDS.sleep(timeToWait);
+	}
+
+
+	/**
+	 * An exception thrown when the max delay is triggered for rate limiter.
+	 */
+	public static final class RateLimiterMaxDelayException extends MessagingException {
```
artembilan: Well, actually exceptions are named after actions: `ResourceNotFound`, `TransactionTimeout` etc.

I'm not sure in the algorithm here yet, so I can't come up with the better name...

`MaxDelayExceededException` ?

The `RateLimiter` is already implied with the wrapping class. 🤷‍♂️

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRateLimiterAdvice.java
```diff
@@ -0,0 +1,114 @@
...
+
+	@Override
+	protected Object doInvoke(ExecutionCallback callback, Object target, Message<?> message) throws Exception {
+		delayIfNecessary(target, message);
+		return callback.execute();
+	}
+
+	private void delayIfNecessary(Object target, Message<?> message) throws InterruptedException {
+		long timeToWait = -1;
+		while (timeToWait == -1) {
```
artembilan: I believe there should be a way for fail fast for a single call or as we already know per `rate`. 
I mean if we still in the `rate` we doin't need to sleep or even calculate `System.currentTimeMillis()` at all.

I need to understand algorithm better, because it looks like I have more questions, than review requires 😢

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -49,11 +49,12 @@ For chains that produce a reply, every child element can be advised.
 [[advice-classes]]
 ==== Provided Advice Classes
 
-In addition to providing the general mechanism to apply AOP advice classes, Spring Integration provides three standard advice classes:
+In addition to providing the general mechanism to apply AOP advice classes, Spring Integration provides four standard advice classes:
```
artembilan: Well,  let's fix it for something like `these standard advice classes`  to cover all the possible future additions 😄

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -459,6 +459,45 @@ public class EerhaApplication {
 ----
 ====
 
+[[rateLimiter-advice]]
+===== Rate Limiter Advice
+
+The Rate Limiter advice allows you to ensure that an endpoints does not get overloaded with requests.
+When the RateLimit is breached the request will go in a blocked state.
+
+A typical use case for this advice might be an external service provider not allowing more than n number of request per minute.
```
artembilan: I believe `n` must be wrapped to code to emphasize it a bit.

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -459,6 +459,45 @@ public class EerhaApplication {
...
+Default value for `maxDelay` is -1 which means no max delay.
+The `timeUnit` property represents the unit of time `java.util.concurrent.TimeUnit`.
+This option must be non null.
+
+The following example configures a RateLimiter advice.
+
+====
+[source,xml]
+----
+<int:service-activator input-channel="input" ref="failer" method="service">
```
artembilan: Can we have a Java sample as well, please?

artembilan (changes requested):
Here is some review.

I may come back with something else after sleeping with this a bit and after investigating what really we would like to do here...

Thank you for a great contribution anyway!

artembilan:
I suggest to revise our own algorithm here in favor of well-known and stable solution in the Resilience4J: http://resilience4j.github.io/resilience4j/#_ratelimiter

In the future we may also reconsider our Circuit Breaker implementation to be based on their solution as well: http://resilience4j.github.io/resilience4j/#_circuitbreaker

Any thoughts?

Thanks

garyrussell:
Sure; makes sense.

artembilan:
Superseded with #2781 

@meherzad , thank you for your effort.
You are welcome to review that PR or any other contribution is always welcome!

PR #2545, opened 2018-08-22
MQTT: completionTimeout should be long

garyrussell:
Incorrectly `int` - inconvenient cast required when using `Duration.toMillis()`.

**cherry-pick to 5.0.x**


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -114,14 +114,26 @@ public MqttPahoMessageDrivenChannelAdapter(String url, String clientId, String..
...
 	/**
 	 * Set the completion timeout for operations. Not settable using the namespace.
-	 * Default 30000 milliseconds.
+	 * Default {@value #DEFAULT_COMPLETION_TIMEOUT} milliseconds.
 	 * @param completionTimeout The timeout.
 	 * @since 4.1
+	 * @deprecated in favor of {@link #setCompletionTimeoutLong(long)}.
 	 */
+	@Deprecated
 	public void setCompletionTimeout(int completionTimeout) {
```
artembilan: Is it really so hard in `5.0.x` to convert `long` to `int`?
I mean let's just change the signature of this method for `5.1` and don't back-port!

artembilan (changes requested):

Issue #2543, opened 2018-08-22
DelayHandler starts to process stored messages before all subsequent integration flows are properly initialized


mr-springtime:
Hi all!

In our application we are defining integration flows with the Spring integration Java DSL. To implement delayed mail sending we are using a Delayer with an underlying JDBC message store. 
When restarting our application we encounter several exceptions and have messages lost due to those exceptions.

Reproduction:

Precondition:
* Spring Integration flow defined with Java DSL containing a JDBC message store based Delayer
* Before starting the server, there need to be several messages in the DB due for immediate processing by the delayer

Issues during Server Startup:
1. When the server starts up, at some point the delayer task is scheduled and begins to produce messages for the flow immediately, while the subsequent channel is not properly initialized. The following exception is shown:

```
2018-08-22 09:08:00.154 ERROR [TheAppName,4c80a4a907a23580,0f1e892a2b06d82b,true] 20472 --- [ask-scheduler-3] o.s.integration.handler.LoggingHandler   : 
org.springframework.messaging.MessageDeliveryException: Dispatcher has no subscribers for channel 'application-1.theFlowContainingADelayer.channel#2'.; 
nested exception is org.springframework.integration.MessageDispatchingException: 
Dispatcher has no subscribers, failedMessage=GenericMessage [payload= .......the message .......]
    at org.springframework.integration.channel.AbstractSubscribableChannel.doSend(AbstractSubscribableChannel.java:77)
    at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:445)
    at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:394)
    at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:181)
    at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:160)
    at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:47)
    at org.springframework.messaging.core.AbstractMessageSendingTemplate.send(AbstractMessageSendingTemplate.java:108)
    at org.springframework.integration.handler.AbstractMessageProducingHandler.sendOutput(AbstractMessageProducingHandler.java:426)
    at org.springframework.integration.handler.AbstractMessageProducingHandler.produceOutput(AbstractMessageProducingHandler.java:336)
[...]

PR #2546, opened 2018-08-23
INT-4527: Delayer Retries and Error Handling

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4527

- Add error handling within the scope of a transaction
- Add `retryDelay` and `maxAttempts`
- Include the delivery attempt header in the `ErrorMessage`
- Add `transactionalRelease()` methods to the DSL

See https://github.com/spring-projects/spring-integration/issues/2543

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/DelayerEndpointSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/DelayerEndpointSpec.java
@@ -27,7 +27,14 @@ import org.springframework.expression.Expression;
 import org.springframework.integration.expression.FunctionExpression;
 import org.springframework.integration.handler.DelayHandler;
 import org.springframework.integration.store.MessageGroupStore;
+import org.springframework.integration.transaction.TransactionInterceptorBuilder;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.support.ErrorMessage;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.interceptor.DefaultTransactionAttribute;
+import org.springframework.transaction.interceptor.TransactionInterceptor;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
```diff
@@ -205,6 +226,71 @@ public void setDelayedAdviceChain(List<Advice> delayedAdviceChain) {
...
+	 * released message fails. If the error flow returns normally, the release is
+	 * complete. If the error flow throws an exception, the release will be re-attempted.
+	 * If there is a transaction advice on the release task, the error flow is called
+	 * within the transaction.
+	 * @param delayedMessageErrorChannel the channel.
+	 * @see #setMaxAttempts(int)
+	 * @see #setRetryDelay(long)
+	 * @since 5.0.8
+	 */
+	public void setDelayedMessageErrorChannel(MessageChannel delayedMessageErrorChannel) {
```
artembilan: Can't we come up with some `RetryInterceptor` which could reuse our `ErrorMessageSendingRecoverer` instead of all these retry configuration details?

This one could be added to the `delayedAdvice()` alongside with the `TransactionalInterceptor`. The same way how we do a `request-handler-advice-chain`

Thanks

spring-integration-core/src/test/java/org/springframework/integration/handler/DelayHandlerTests.java
```diff
@@ -121,19 +125,39 @@ public void noDelayHeaderAndDefaultDelayIsZero() throws Exception {
...
+			if (StaticMessageHeaderAccessor.getDeliveryAttempt(m).get() < 4) {
+				throw new RuntimeException("fail");
+			}
+			latch.countDown();
+			return true;
+		});
+		delayHandler.setOutputChannel((m, t) -> {
+			throw new RuntimeException("fail");
+		});
+		input.send(message);
```
artembilan: Is test not finished?

I don't see any assertions here.

src/reference/asciidoc/delayer.adoc
```diff
@@ -187,3 +189,20 @@ Message<String> delayerReschedulingMessage =
...
+[[delayer-release-failures]]
+==== Release Failures
+
+Starting with version 5.0.8, there are two new properties on the delayer:
+
+- `maxAttempts` (default 5)
+- `retryDelay` (default 1 second)
+
+When a message is released, if the downstream flow fails, the release will be attempted after the `retryDelay`.
+If the `maxAttempts` is reached, the message is discarded (unless the release is transactional, in which case the message will remain in the store, but will no longer be scheduled for release, until the application is restarted).
```
artembilan: Well, we always should not forget that `DelayHandler` is a `@ManagedResource` and has these management operations:
```
public interface DelayHandlerManagement {

	@ManagedAttribute
	int getDelayedMessageCount();

	@ManagedOperation
	void reschedulePersistedMessages();
}
```

So, with the Control Bus command we always can reschedule at runtime.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
garyrussell: No; we can't do that because the retry advice (even with a stateful interceptor) relies on blocking the thread, which we don't want to do; scheduler threads are precious. This mechanism implements delayed retries using the scheduler and doesn't block any threads.

artembilan: OK. Sounds interesting. 
I need to digest it a bit to get used to...

Can we still try to reuse our `ErrorMessagePublisher`  at least?

src/reference/asciidoc/delayer.adoc
garyrussell: Oh, dear; I think there's a problem with that - `releaseMessageAfterDelay()` should only be invoked when this is called after the context refresh. For a managed operation, we should only release past-due messages; otherwise future messages will be duplicated. And, it's not that simple; there will be race conditions.

But, I think resolving this is outside of this PR.

spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
garyrussell: I started looking at it, but it adds quite a bit of complexity.

We should probably revisit it (and the `ErrorMessageStrategy`) to make them easier to reuse for cases like this, but I don't really want to do that as part of this PR.

garyrussell:
Yikes - I installed the spring-formatter plugin for my boot work - it looks like it auto-formatted the delayer javadocs.

Uninstalling.

spring-integration-core/src/main/java/org/springframework/integration/dsl/DelayerEndpointSpec.java
```diff
@@ -97,6 +104,110 @@ public DelayerEndpointSpec delayExpression(String delayExpression) {
...
+	 * @param transactionManager the {@link PlatformTransactionManager} to use.
+	 * @param handleMessageAdvice the flag to indicate the target {@link Advice} type:
+	 * {@code false} - regular {@link TransactionInterceptor}; {@code true} -
+	 * {@link org.springframework.integration.transaction.TransactionHandleMessageAdvice}
+	 * extension.
+	 * @return the spec.
+	 * @since 5.0.8
+	 */
+	public DelayerEndpointSpec transactionalRelease(PlatformTransactionManager transactionManager,
+			boolean handleMessageAdvice) {
```
artembilan: Why do we need this variant of the `TransactionInterceptor` if we are going to proxy only a `DelayHandler.ReleaseMessageHandler` anyway?

spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
```diff
@@ -370,11 +457,75 @@ private void releaseMessageAfterDelay(final Message<?> message, long delay) {
...
+					}
+					rescheduleForRetry(message, identity);
+				}
+			}
+			else {
+				if (this.logger.isDebugEnabled()) {
+					logger.debug("Release flow threw an exception for message: " + message, e);
+				}
+				if (!rescheduleForRetry(message, identity)) {
+					throw e; // there might be an error handler on the scheduler
```
artembilan: Why do we rethrow exception only here?
How about all other `rescheduleForRetry()` usage?

artembilan (changes requested):
A couple concerns.

src/reference/asciidoc/delayer.adoc
garyrussell: Ah, wait; we're protected against this by `removeDelayedMessageFromMessageStore`. We could consider optimizing, though, to avoid double-scheduling future events.

spring-integration-core/src/main/java/org/springframework/integration/dsl/DelayerEndpointSpec.java
garyrussell: Good catch.

spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
garyrussell: This code was added to support "legacy" error channel processing; the new error channel processing is performed within the scope of the transaction (if present); the old one is outside the transaction scope. There are several tests that use the `MessagePublishingErrorHandler` in the scheduler.

The legacy error handling is unchanged in that it is only invoked after the final release attempt. It also cannot influence the retry logic (unlike the new version).

artembilan (approved):
Merging after some polishing.
I believe...

spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
```diff
@@ -370,11 +457,75 @@ private void releaseMessageAfterDelay(final Message<?> message, long delay) {
...
+		this.deliveries.putIfAbsent(identity, new AtomicInteger());
+		try {
+			this.releaseHandler.handleMessage(message);
+			this.deliveries.remove(identity);
+		}
+		catch (Exception e) {
+			if (getErrorChannel() != null) {
+				ErrorMessage errorMessage = new ErrorMessage(e,
+						Collections.singletonMap(IntegrationMessageHeaderAccessor.DELIVERY_ATTEMPT,
+								new AtomicInteger(this.deliveries.get(identity).get() + 1)),
```
artembilan: I believe we just need to initialize an `AtomicInteger` with `1` and reuse it here. 
And, of course, we would need to use a `getAndIncrement()` below instead.
Or do you worry about mutation from the target error process?

What was the point though to make such a header mutable from the first place?
I don't see any mutations throughout the project 🤷‍♂️ 

WDYT?

garyrussell: I don't want to give the error flow the ability to change the hard reference to the actual integer in the map. 

It's an `AtomicInteger` because the header accessor expects it to be.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/075d237c0423eca82d072b29f8a65e99c3d3969d

artembilan:
... and cherry-picked into `5.0.x` as https://github.com/spring-projects/spring-integration/commit/af51c9bf3e50f6d8358cb0a3c7b6d26beea1b5c0 after resolving Docs conflicts.

PR #2548, opened 2018-08-24
INT-4526: Fix Channel Interceptor NonNullApi Call

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4526

- `Message<?>` cannot be null in `postReceive()`.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractPollableChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractPollableChannel.java
@@ -123,7 +123,9 @@ public abstract class AbstractPollableChannel extends AbstractMessageChannel
 				logger.trace("postReceive on channel '" + this + "', message is null");
 			}
 			if (!CollectionUtils.isEmpty(interceptorStack)) {
-				message = interceptorList.postReceive(message, this);
+				if (message != null) {
+					message = interceptorList.postReceive(message, this);
+				}
 				interceptorList.afterReceiveCompletion(message, this, null, interceptorStack);
 			}
 			return message;
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/PollableJmsChannel.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/PollableJmsChannel.java
@@ -95,28 +95,32 @@ public class PollableJmsChannel extends AbstractJmsChannel
 				object = getJmsTemplate().receiveSelectedAndConvert(this.messageSelector);
 			}
 
+			Message<?> message = null;
 			if (object == null) {
 				if (logger.isTraceEnabled()) {
 					logger.trace("postReceive on channel '" + this + "', message is null");
 				}
 				return null;
 			}
-			if (countsEnabled) {
-				getMetrics().afterReceive();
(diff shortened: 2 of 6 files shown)


spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractPollableChannel.java
```diff
@@ -123,7 +123,9 @@ else if (logger.isTraceEnabled()) {
 				logger.trace("postReceive on channel '" + this + "', message is null");
 			}
 			if (!CollectionUtils.isEmpty(interceptorStack)) {
-				message = interceptorList.postReceive(message, this);
+				if (message != null) {
+					message = interceptorList.postReceive(message, this);
```
artembilan: Can't we follow a similar logic in the `PollableAmqpChannel` or `PollableJmsChannel` like this:
```
if (object == null) {
		if (logger.isTraceEnabled()) {
				logger.trace("postReceive on channel '" + this + "', message is null");
		}
		return null;
}
```

I mean instead of check here just return `null` after trace log message.

artembilan (changes requested):
Just only one concern regarding consistency.

I can address it during merge though.
Just need to confirmation how to proceed.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractPollableChannel.java
garyrussell: No because we need to call `afterReceiveCompleted()` - its API allows `null`.

artembilan: So, `PollableAmqpChannel` and `PollableJmsChannel` have to be fixed respectively?
They just `return null;` and definitely don't call that `afterReceiveCompleted()`...

artembilan: OK. So, I'm leaving that for you to fix in the subsequent commits.

I'm not sure that we are going to be good to back-port it, since in the end it will be as a behavior change.

garyrussell: Right - no backport.

spring-integration-jms/src/main/java/org/springframework/integration/jms/PollableJmsChannel.java
```diff
@@ -95,28 +95,32 @@ public long getReceiveErrorCountLong() {
 				object = getJmsTemplate().receiveSelectedAndConvert(this.messageSelector);
 			}
 
+			Message<?> message = null;
 			if (object == null) {
 				if (logger.isTraceEnabled()) {
 					logger.trace("postReceive on channel '" + this + "', message is null");
 				}
 				return null;
```
artembilan: You have missed to remove this call.
Will fix on merge...

artembilan (approved):
Merging with the mentioned fix in the `PollableJmsChannel`...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/e6489abb524c6d0fb0233a9a4a302107698d0580

PR #2549, opened 2018-08-24
Make Headers PREFIX Constants Public

garyrussell:
**cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileHeaders.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileHeaders.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


jmax01:
Great! Thank you.

artembilan (commented):
I believe an `IpHeaders` should be included to this fix as well.

I can fix on merge though...

garyrussell:
Ah; yes; didn't show up in my search because it's not PREFIX.

Yes, please fix on merge.

garyrussell:
(Please change to `PREFIX`).

artembilan:
> (Please change to PREFIX).

We can't see its source code:
```
public final class IpHeaders {

	private static final String IP = "ip_";

	private static final String TCP = IP + "tcp_";
```

Although it's indeed can be like `IP_PREFIX` and `TCP_PREFIX`  😄

garyrussell:
hehe - ok (or leave as-is).

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/5ab7d2493478d4193b64bd26abf6ec300539ed2b and cherry-picked to `5.0.x`

PR #2551, opened 2018-08-29
TCP, SSL, Enable Host Verification by default

garyrussell:
**Backport to 5.0 needs to be off by default; I will issue another PR after review.**


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpNioSSLConnectionSupport.java
```diff
@@ -37,9 +38,25 @@
...
 
-	private volatile SSLContext sslContext;
+	private final SSLContext sslContext;
 
+	private final boolean sslVerifyHost;
+
+	/**
+	 * Create an instance with host verification enabled.
+	 * @param sslContextSupport the ssl context support.
+	 * @since 5.0.8
```
artembilan: I believe this `@since` should go to the other, really new ctor.

If that, I can fix on merge.

garyrussell: Oops; yes; thanks.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/SocketSupportTests.java
```diff
@@ -182,26 +182,26 @@ public void postProcessServerSocket(ServerSocket serverSocket) {
...
 	}
 
-	/*
-$ keytool -genkeypair -alias sitestcertkey -keyalg RSA -validity 36500 -keystore src/test/resources/test.ks
-Enter keystore password: secret
-Re-enter new password: secret
+/*
+$ keytool -genkeypair -alias sitestcertkey -keyalg RSA -validity 36500 -keystore src/test/resources/test.ks -ext san=dns:localhost
+Enter keystore password:
+Re-enter new password:
```
garyrussell: Can you also reinstate `secret` here? Thanks.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/DefaultTcpSSLContextSupport.java
```diff
@@ -69,9 +77,28 @@ public DefaultTcpSSLContextSupport(String keyStore, String trustStore,
 		this.trustStorePassword = trustStorePassword.toCharArray();
 	}
 
+	/**
+	 * Set the key store type. Default JKS.
+	 * @param keyStoreType the type.
+	 * @since 5.0.8
+	 */
+	public void setKeyStoreType(String keyStoreType) {
+		this.keyStoreType = keyStoreType;
```
artembilan: I don't think that this can be as null or even empty: so, I will add `Assert.hasText(keyStoreType, "'keyStoreType' cannot be empty");`, as well as for the `setTrustStoreType()`

garyrussell: Thanks; yes; I meant to add those asserts, but was distracted.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/230d12d425c00ef1e534b6bd750ab1d38e991db1.

I think `5.0.x` fix deserves its own PR, as you mentioned in the comment above.

Thanks

PR #2552, opened 2018-08-29
TCP, SSL, Configure Host Verification

garyrussell:
Make key/trust store types configurable; add a test with host violation.

* Fix some typos and code style in the related classed and docs
* Add asserts for the store type properties


garyrussell:
Separate commit with changes so you don't have to re-review.

artembilan (approved):
That was cool!

Merging...

PR #2555, opened 2018-08-30
Add `IntegrationFlowDefinition.nullChannel()`

artembilan:
When a `NullChannel` is used in the middle of the flow, it may be not so
obvious why our flow is stopped after accidentally added the next
endpoint

* For convenient add a terminal `nullChannel()` operator into the
`IntegrationFlowDefinition`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
@@ -2961,6 +2961,17 @@ public abstract class IntegrationFlowDefinition<B extends IntegrationFlowDefinit
 		return new PublisherIntegrationFlow<>(this.integrationComponents, publisher);
 	}
 
+	/**
+	 * Add a {@value IntegrationContextUtils#NULL_CHANNEL_BEAN_NAME} bean into this flow
+	 * definition as a terminal operator.
+	 * @return The {@link IntegrationFlow} instance based on this definition.
+	 * @since 5.1
+	 */
+	public IntegrationFlow nullChannel() {
+		return channel(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)
+				.get();
+	}
+
 	@SuppressWarnings("unchecked")
 	private <S extends ConsumerEndpointSpec<S, ? extends MessageHandler>> B register(S endpointSpec,
 			Consumer<S> endpointConfigurer) {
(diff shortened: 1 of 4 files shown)


artembilan:
@jmax01,

here ids what we talked about on Gitter recently: https://gitter.im/spring-projects/spring-integration?at=5b86e94c1d3a5711b6b721fe

Nevertheless I'm still reluctant to merge it at all, since no one can stop you to add a `channel("nullChannel")` and continue the flow.
This fix brings some convenience, but it is not a panacea from any possible (and not a fact wrong) solutions.

Let's see what @garyrussell  and others think on the matter!

garyrussell:
@artembilan  I have no issues with merging this; if someone puts a `NullChannel` mid-flow it should be obvious that the flow will stop there.

But the DSL is "your" baby so I think you have the final say for this one.

artembilan:
OK. Feel free to merge.

The problem that it is really not obvious without knowing what is `nullChannel` about in advance.
I though about some preemptive assertion in the `channel()` operator, but this won't follow existing logic in Spring Integration: as long as you can do `@ServiceActivator(inputChannel = "nullChannel")` as well as you can do `<service-activator input-channel="nullChannel">`.

Although we may really add a WARN into the `IntegrationConsumer` impls to say that nothing is going to be pulled from the `NullChannel`, so this endpoint is useless. 🤷‍♂️ 

WDYT?

garyrussell:
Yes; a WARN log makes sense; I am not sure we want to emit it on every poll, though, maybe once an hour?

artembilan:
Hm. Wait. We can just do that in the ctor: `public PollingConsumer(PollableChannel inputChannel, MessageHandler handler) {`

Doh...

Will push a WARN soon...

jmax01:
I don't think `nullChannel()` is a great name for the terminal method, would prefer something that makes it more obvious that this flow terminates at the `nullChannel` maybe `[terminate|complete|end][With|To]NullChannel()`.

Reminds me that I never liked that the `IntegrationFlowDefinition|IntegrationFlowBuilder` build method is `get()` and not `build()`, but that's a another story :wink:

artembilan:
Well, this name describes what we are going to add into the flow and just after that you will realize that you can't add anything else because the return of the method is not an `IntegrationFlowDefinition` anymore.
i don't see reason to overload method names with really an obvious outcome from them. The same with the `logAndReply()`.

The story about `get()` or `build()` is really different and that is really not so explicit from the flow definition. Therefore something like `IntegrationFlows.from()...get();` sounds much straightforward and self-explanatory.
Also need to keep in mind that with the `build()` we suppose to always return a new instance, meanwhile `get()` is idempotent and whenever we call it, we always *get* the same `IntegrationFlow` instance.

jmax01:
RE: `build`:  makes sense.

garyrussell:
Please rebase and resolve conflicts.

artembilan:
Rebased and fixed conflicts in the `PolingConsumer` about `NullChannel`

PR #2557, opened 2018-08-31
Avoid leaking Zookeeper znodes in tryLock

JonathanO:
Creating a EPHEMERAL_SEQUENTIAL node was being used to ensure the
Zookeeper connection had been established. These were never explicitly
removed, which led to tryLock effectively leaking zNodes until the
client was terminated. This in turn lead to an extremely large number of
ephemeral nodes being deleted whenever a long running service
terminated!

This fix replaces the creation of a node with a stat of "/" instead.
This has the desired effect of re-establishing the Zookeeper connection,
but is a read-only operation.


artembilan:
... and cherry-picked to `5.0.x` and `4.3.x`.

@JonathanO , thank you very much for the contribution.
Looking forward for more!

PR #2558, opened 2018-09-05
INT-4525 Configure TCP gateway second chance delay

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4525

Hard-coded 2 second delay is now configurable.

Also fix the logic for creating the SpEL evaluation context.

**cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
@@ -61,19 +61,25 @@ import org.springframework.util.Assert;
 public class TcpOutboundGateway extends AbstractReplyProducingMessageHandler
 		implements TcpSender, TcpListener, Lifecycle {
 
-	private volatile AbstractClientConnectionFactory connectionFactory;
+	private static final int DEFAULT_SECOND_CHANCE_DELAY = 2;
 
-	private volatile boolean isSingleUse;
+	private AbstractClientConnectionFactory connectionFactory;
+
+	private boolean isSingleUse;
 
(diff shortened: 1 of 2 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
```diff
@@ -322,7 +341,7 @@ private AsyncReply(long remoteTimeout) {
 					 * before the reply, on a different thread.
 					 */
 					logger.debug("second chance");
-					this.secondChanceLatch.await(2, TimeUnit.SECONDS); // NOSONAR don't care about result
+					this.secondChanceLatch.await(TcpOutboundGateway.this.secondChanceDelay, TimeUnit.SECONDS);// NOSONAR
```
artembilan: Checkstyle doesn't like that you are missing a whitespace after semicolon 😄 

Will fix on merge

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
```diff
@@ -61,19 +61,25 @@
 public class TcpOutboundGateway extends AbstractReplyProducingMessageHandler
 		implements TcpSender, TcpListener, Lifecycle {
 
-	private volatile AbstractClientConnectionFactory connectionFactory;
+	private static final int DEFAULT_SECOND_CHANCE_DELAY = 2;
```
artembilan: You don't use this constant BTW.
I mean that `secondChanceDelay` is really without those 2 seconds value...
Fixing on merge...

artembilan:
Merged and cherry-picked as https://github.com/spring-projects/spring-integration/commit/6cc4fe530df1d7ee3617bfdb5d2ba628ff568c92

PR #2561, opened 2018-09-07
INT-4317: JMS: dynamic deliverMode and timeToLive

artembilan:
JIRA: https://jira.spring.io/browse/INT-4317

* Add `deliveryModeExpression` and `timeToLiveExpression` properties
to the `JmsSendingMessageHandler` and expose them in the Java DSL and
XML components
* Add `setMapInboundDeliveryMode()` and `setMapInboundExpiration()`
`boolean` properties (default `false`) to the `DefaultJmsHeaderMapper`
for transferring `JMSDeliveryMode` and `JMSExpiration` into appropriate
`JmsHeaders.DELIVERY_MODE` and `JmsHeaders.EXPIRATION` headers
* Upgrade to latest Kotlin and AssertK

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlinVersion = '1.2.60'
+	ext.kotlinVersion = '1.2.61'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlinVersion = '1.2.51'
+	ext.kotlinVersion = '1.2.60'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
@@ -88,7 +88,7 @@ subprojects { subproject ->
 		apacheSshdVersion = '1.7.0'
 		aspectjVersion = '1.9.0'
 		assertjVersion = '3.9.1'
-		assertkVersion = '0.10'
+		assertkVersion = '0.12'
 		boonVersion = '0.34'
 		commonsDbcp2Version = '2.2.0'
 		commonsIoVersion = '2.6'
@@ -178,9 +178,7 @@ subprojects { subproject ->
 		testRuntime "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
 		testRuntime "org.apache.logging.log4j:log4j-jcl:$log4jVersion"
 
-		testCompile("com.willowtreeapps.assertk:assertk:$assertkVersion") {
-				exclude group: 'org.jetbrains.kotlin', module: 'kotlin-reflect'
-		}
+		testCompile("com.willowtreeapps.assertk:assertk-jvm:$assertkVersion")
 
 		testCompile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
 		testRuntime "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsSendingMessageHandler.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsSendingMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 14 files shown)


build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlinVersion = '1.2.51'
+	ext.kotlinVersion = '1.2.60'
```
garyrussell: 61 is out already.

build.gradle
```diff
@@ -178,9 +178,7 @@ subprojects { subproject ->
 		testRuntime "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
 		testRuntime "org.apache.logging.log4j:log4j-jcl:$log4jVersion"
 
-		testCompile("com.willowtreeapps.assertk:assertk:$assertkVersion") {
-				exclude group: 'org.jetbrains.kotlin', module: 'kotlin-reflect'
-		}
+		testCompile("com.willowtreeapps.assertk:assertk-jvm:$assertkVersion")
```
garyrussell: Why is the exclusion no longer needed?

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsSendingMessageHandler.java
```diff
@@ -105,22 +152,42 @@ protected void onInit() {
...
 		try {
-			DynamicJmsTemplateProperties.setPriority(new IntegrationMessageHeaderAccessor(message).getPriority());
-			this.send(destination, objectToSend, messagePostProcessor);
+			send(destination, objectToSend, messagePostProcessor);
 		}
 		finally {
 			DynamicJmsTemplateProperties.clearPriority();
+			DynamicJmsTemplateProperties.clearDeliveryMode();
+			DynamicJmsTemplateProperties.clearTimeToLive();
 		}
```
garyrussell: I wonder if it would be better to have one ThreadLocal with a container?

garyrussell (commented):
Just a couple of comments/questions.

build.gradle
artembilan: According my Gradle report the old version is remapped pretty well:
```
+--- com.willowtreeapps.assertk:assertk-jvm:0.12
|    +--- com.willowtreeapps.assertk:assertk-common:0.12
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib-common:1.2.50 -> 1.2.61
|    |    \--- com.willowtreeapps.opentest4k:opentest4k-common:1.1.0
|    |         \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.2.41 -> 1.2.61
|    +--- org.jetbrains.kotlin:kotlin-stdlib:1.2.50 -> 1.2.61
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib-common:1.2.61
|    |    \--- org.jetbrains:annotations:13.0
|    +--- org.jetbrains.kotlin:kotlin-reflect:1.2.50 -> 1.2.61
|    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.2.61 (*)
|    \--- com.willowtreeapps.opentest4k:opentest4k-jvm:1.1.0
|         +--- com.willowtreeapps.opentest4k:opentest4k-common:1.1.0 (*)
|         +--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.2.41 -> 1.2.61
|         |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.2.61 (*)
|         |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.2.61
|         |         \--- org.jetbrains.kotlin:kotlin-stdlib:1.2.61 (*)
|         \--- org.opentest4j:opentest4j:1.1.0
+--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.2.61 (*)
```

And there is no any issues during running tests against that latest AssertK.

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsSendingMessageHandler.java
artembilan: Not sure what is your concern: we have a container there and it is a `ThreadLocalMap`:
```
    public void set(T value) {
        Thread t = Thread.currentThread();
        ThreadLocalMap map = getMap(t);
        if (map != null)
            map.set(this, value);
        else
            createMap(t, value);
    }
```

So, essentially with your suggestion we are going to have only a single entry in this map, but we would create that extra container anyway.
And at the same time we would lose a flexibility when we deal only with one property. For example a `setPriority()` is used from the `AbstractJmsChannel`, too.

So, what am I missing staying with the standard Java approach?

Thanks

garyrussell: Well, we now have 4 `ThreadLocal`s, 4 `ThreadLocalMap`s.

I was just suggesting one `ThreadLocal<SomeContainer>` and one `DynamicJmsTemplateProperties.clear()` method.

On the other hand, we'd create/remove a new Object for each send, so I suppose the current is better (since the `ThreadLocalMap` is retained).

So, OK, LGTM.

artembilan: Huh? No, there is a single map-per-thread: `ThreadLocalMap map = getMap(t);`. As I said: essentially it is the same container as you suggest.

garyrussell: I didn't mean to say there wasn't a map per thread. My main point was we could replace 

```java
DynamicJmsTemplateProperties.clearPriority();
DynamicJmsTemplateProperties.clearDeliveryMode();
DynamicJmsTemplateProperties.clearTimeToLive();
```

with just one `DynamicJmsTemplateProperties.clear()` method.

But, as I said, I am ok with this.

PR #2562, opened 2018-09-11
File Source: do not deal with dir until start()

artembilan:
StackOverflow: https://stackoverflow.com/questions/52273537/spring-boot-failed-to-load-applicationcontext


garyrussell:
Merged as e4ca8571e0e66e760c1a2705ca753d122131fa81 with minor asciidoc polishing.

PR #2564, opened 2018-09-14
INT-4531: Use Object key in (S)FTP rotating server advice

MichaelF25:
Small fix for https://jira.spring.io/browse/INT-4531.
The `DelegatingSessionFactory` uses plain `Object`s as key types whereas the default `RotatingServerAdvice` does not. This PR just changes the datatype from `String` to `Object`.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
@@ -219,18 +219,18 @@ public class RotatingServerAdvice extends AbstractMessageSourceAdvice {
 	 */
 	public static class KeyDirectory {
 
-		private final String key;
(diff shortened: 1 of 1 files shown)


artembilan:
Please, see my comment in JIRA.
We haven't agreed yet that this API has to be changed.

Thanks

artembilan (approved):
After discussion with @garyrussell we decided to go ahead with your suggestion: it isn't going to break any target logic where `String` is used as a key.

Only what we need in the target project is to recompile it after upgrade to avoid a `NoSuchMethodError` without recompilation.

So, @MichaelF25 ,
thank you for contribution and merging with minor polishing and cherry-picking to `5.0.x`

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/5bfd6971dbf2012b079d58f910650336b634308e and cherry-picked into `5.0.x`

PR #2566, opened 2018-09-18
Use RetrySynchronizationManager

garyrussell:
Use the `RetrySynchronizationManager` instead of a `RetryListener`.

Fix `setAttributesIfNecessary` for gateway conversion errors (this, at least,
should be cherry-picked).


garyrussell:
Will do the same for SIK after review/merge.

Issue #2568, opened 2018-09-20
Assert incorrectly triggered in FileReadingMessageSource


Stexxen:
## Bug report

An Assert is triggered incorrectly in FileReadingMessageSource

    FileReadingMessageSource source = new FileReadingMessageSource();
    source.setScanner(new RecursiveDirectoryScanner());
    source.setLocker(new NioFileLocker());
    source.setDirectory(new File(INBOUND_PATH));
    source.setFilter(new IgnoreHiddenFileListFilter());

    source.afterPropertiesSet();

Generates the following Exception

    java.lang.IllegalStateException: The 'filter' and 'locker' options must be present on the provided external 'scanner': org.springframework.integration.file.RecursiveDirectoryScanner@2a7f1f10
	at org.springframework.util.Assert.state(Assert.java:73)
	at org.springframework.integration.file.FileReadingMessageSource.onInit(FileReadingMessageSource.java:340)
	at org.springframework.integration.context.IntegrationObjectSupport.afterPropertiesSet(IntegrationObjectSupport.java:176)

This is in 5.0.8.RELEASE.

It is triggered from the following code
https://github.com/spring-projects/spring-integration/blob/master/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java#L337-L339

    Assert.state(!(this.scannerExplicitlySet && (this.filter != null || this.locker != null)),
				"The 'filter' and 'locker' options must be present on the provided external 'scanner': "
						+ this.scanner);

I can see what it is trying to do, but I think it should be written as follows

    Assert.state(!(this.scannerExplicitlySet && (this.filter **==** null || this.locker **==** null)),
				"The 'filter' and 'locker' options must be present on the provided external 'scanner': "
						+ this.scanner);

Also the error message is kinda wrong, as it's not testing the filter and locker options on the provided scanner, it's actually testing them on the FileReadingMessageSource object and then passing them to the provided scanner in the lines that follow.

If it would expedite things, I can provide a PR?

artembilan:
No sure what makes you think this way, but the logic is like this:

If you provide an external `DirectoryScanner`, that's your responsibility to set `filter` and `locker` there.
In the `FileReadingMessageSource` we can't mutate external object. This is not how dependency injection works.
If you don't provide an external `DirectoryScanner`, you are free to configure `filter` and `locker` on the `FileReadingMessageSource`. In this case we mutate an internal  `DirectoryScanner` instance.
[...]

PR #2570, opened 2018-09-20
Changed an Assert error message to make the requirements clearer

Stexxen:
Changed an Assert error message to make the requirements when using an external scanner clearer, and added some explanatory comments.

Discussed in #2568

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
@@ -80,6 +80,7 @@ import org.springframework.util.Assert;
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Steven Pearce
  */
 public class FileReadingMessageSource extends AbstractMessageSource<File>
 		implements Lifecycle {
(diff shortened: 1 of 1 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
```diff
@@ -334,8 +334,10 @@ protected void onInit() {
 			this.scanner = new WatchServiceDirectoryScanner();
 		}
 
+		// Check that the local filter and locker options are _NOT_ set if an external scanner has been set.
+		// The external scanner is responsible for the filter and locker options in that case.
 		Assert.state(!(this.scannerExplicitlySet && (this.filter != null || this.locker != null)),
-				"The 'filter' and 'locker' options must be present on the provided external 'scanner': "
+				"When using an external scanner the local 'filter' and 'locker' options should not be used. Instead, set these options on the external DirectoryScanner: "
```
artembilan: I think `local` word here and in the comment above is misleading.
This `FileReadingMessageSource` definitely works only with the local file system.

Also, please, add your name to `@author`

artembilan (changes requested):

artembilan:
@Stexxen ,

Looking forward for more!

PR #2573, opened 2018-09-27
Add `@LongRunningTest` JUnit 5 condition

artembilan:


garyrussell:
Do we really need this? JUnit 5 has `@Tag` for tests which can be excluded and/or included. We should be able to add code the build.gradle to include `@Tag("longRunning")` when the environment is set. https://junit.org/junit5/docs/current/user-guide/#writing-tests-tagging-and-filtering and https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle

artembilan:
OK, but it doesn't help in the IDE. I understand that I can to configure it somehow to reflect the tag behavior via `includeTags` as well, but is this going to be a good practice for all the contributors?

I won't mind to revise the solution to any other possible way in JUnit 5: it is still a new area for me 😄

artembilan:
Gary, let me know if we go this way and I'll add some docs on the matter.
Otherwise let's discuss other possible solution whenever you have a spare time.

Thanks!

garyrussell:
OK (docs, please).

garyrussell:
Merged as 1595462ce704835cc065b3fb4118153a7fcb22d2 after resolving doc conflicts.

PR #2575, opened 2018-10-02
Allow Functions as Service Activators

artembilan:
* Fix `MessagingMethodInvokerHelper` to extract canonical method for the
`Function` and `Consumer` beans
* Demonstrate in test how `Function` and `Consumer` can be configured
with the Messaging annotations

/CC @markpollack , @olegz

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -35,6 +35,8 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.function.Function;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -687,6 +689,20 @@ private boolean contentTypeIsJson(Message<?> message) {
 		final AtomicReference<Class<?>> ambiguousFallbackType = new AtomicReference<>();
 		final AtomicReference<Class<?>> ambiguousFallbackMessageGenericType = new AtomicReference<>();
 		final Class<?> targetClass = getTargetClass(targetObject);
+
+		final String methodName;
+
+		if (methodNameToUse == null) {
```
garyrussell: It's not good practice to mutate parameter arguments. Better to leave the argument final (e.g. `final String methodNameArg`) and then `String methodName = methodNameArg`.

garyrussell (changes requested):

garyrussell (approved):

PR #2588, opened 2018-10-06
INT-4539: Fix Java DSL for prototype beans

artembilan:
JIRA: https://jira.spring.io/browse/INT-4539

The `IntegrationFlowBeanPostProcessor` doesn't check for prototype beans
and just override them in the application context with the singletons

* Since we can't in the DSL understand if provided object is a prototype
or not, we try to check for its bean definition by possible bean name.
Use `NamedComponent` for possible bean name to check.
* Remove `final` from the `IntegrationComponentSpec.get()` since its
result is not visible for CGI proxies when it is declared as a `@Bean`
and subsequent `getObject()` produces a new internal object
* Verify prototype beans with new test in the `IntegrationFlowTests`

**Cherry-pick to 5.0.x**


garyrussell:
cherry-picked as 3a7a14140befab9334036db21123d930b7288c6d after resolving test conflicts.

PR #2589, opened 2018-10-08
INT-4540: Use MessageHeaders in GenericHandler

artembilan:
JIRA: https://jira.spring.io/browse/INT-4540

For better end-user experience use a `MessageHeaders` in the
`GenericHandler` instead of plain `Map<String, Object>`


garyrussell:
Please add to the migration guide too.

artembilan:
Done: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.0-to-5.1-Migration-Guide#generichandler-and-messageheaders.
And pushed the fix for the `IntegrationFlowEventsTests`.

garyrussell:
>The existing lambdas in the Java DSL's .handle() is fully compatible.

But still need recompilation; right?

artembilan:
> But still need recompilation

Correct. Fixed that.

Thanks.

PR #2590, opened 2018-10-08
INT-4541: Fix Reactive MessagingGateway Errors

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4541

Add test case to reproduce.

The gateway correctly sets up the `errorChannel` header so that a downstream
`MPEH` will send exceptions back to the gateway, but the `map()` function
did not check for an error message.

Check for an error message and throw the payload so that the `onErrorResume`
will route to the error channel.


artembilan:
... and back-ported to `5.0.x` as https://github.com/spring-projects/spring-integration/commit/7403e36c1e056ee4ed2f95dc2572e7c2ab765c2d after fixing conflicts in the `WebFluxDslTests`.

PR #2591, opened 2018-10-08
INT-4542: CorrMH: Use popSequenceDetails properly

artembilan:
JIRA: https://jira.spring.io/browse/INT-4542

Perform a `MessageBuilder.popSequenceDetails()` for those
`MessageGroupProcessor` results which are not `Message`
or `Collection<Message>`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -52,6 +52,7 @@ import org.springframework.integration.store.MessageStore;
 import org.springframework.integration.store.SimpleMessageGroup;
 import org.springframework.integration.store.SimpleMessageStore;
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
+import org.springframework.integration.support.MessageBuilder;
 import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.LockRegistry;
 import org.springframework.integration.util.UUIDConverter;
@@ -141,6 +142,8 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	private boolean expireGroupsUponTimeout = true;
 
+	private boolean popSequenceDetails = true;
+
 	private volatile boolean running;
 
 	public AbstractCorrelatingMessageHandler(MessageGroupProcessor processor, MessageGroupStore store,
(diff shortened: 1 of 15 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -653,7 +654,8 @@ protected void forceComplete(MessageGroup group) {
 						 * groups. A longer timeout for empty groups can be enabled by
 						 * setting minimumTimeoutForEmptyGroups.
 						 */
-						removeGroup = lastModifiedNow <= (System.currentTimeMillis() - this.minimumTimeoutForEmptyGroups);
+						removeGroup = lastModifiedNow <= (System
+								.currentTimeMillis() - this.minimumTimeoutForEmptyGroups);
```
garyrussell: This looks odd; move System to the next line too?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -753,10 +755,24 @@ protected void completeGroup(Object correlationKey, MessageGroup group) {
...
+			if (result instanceof AbstractIntegrationMessageBuilder<?>) {
+				messageBuilder = (AbstractIntegrationMessageBuilder<?>) result;
+			}
+			else {
+				messageBuilder = getMessageBuilderFactory()
+						.withPayload(result)
+						.copyHeaders(message.getHeaders());
+			}
+			result = messageBuilder.popSequenceDetails();
+		}
```
garyrussell: Now I look at this, I wonder if it is completely valid to do this unconditionally. Consider

    splitter -> custom-splitter -> custom-aggregator -> aggregator

Perhaps we need a new flag `alwaysPopSequenceDetails` (true in 5.1, false in 5.0).

Or, add some metadata (`seqInfoId`)  to the sequence details and only pop if the splitter and aggregator are configured with the same metadata.

If the custom splitter doesn't push sequence details, we will remove the real ones from the first splitter.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
artembilan: Hm. Some fault of my IDEA code reformatting.
Will fix.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
artembilan: OK. Sounds like we just need to be sure that our aggregation is based on the `sequenceDetails`. Because custom `splitter` may not be based on the `applySequence`.
So, maybe just an extra `sequenceApplied` header? But then how to understand on which level it has not been used?..

Or... I just don't see the whole picture and need to sleep with this...

artembilan: After sleeping with this, I came with the conclusion that your suggestion about `popSequenceDetails(true|false)` is the way to go.
We won't pollute messages with some extra header which may be really out of control and I don't want to tie an aggregator with the splitter.

We may do a condition based on the default `HeaderAttributeCorrelationStrategy(IntegrationMessageHeaderAccessor.CORRELATION_ID)` and `SimpleSequenceSizeReleaseStrategy`, but that still doesn't come as a robust solution since we can use defaults, but populate sequence header some other way, not via splitter.

From here I think we just need to add such a `popSequenceDetails` option to the correlation handler, make it `true` by default (looks like most cases, since we haven't had issues so far), modify a `AbstractAggregatingMessageGroupProcessor` to return a `AbstractIntegrationMessageBuilder` instead of the whole message and do conditional `.popSequenceDetails()` already in the `AbstractCorrelatingMessageHandler`. 

Saying that I don't think that we are good for back-porting to `5.0.x`: looks like this is going to be some breaking change.

Let me know WDYT?
Thanks

garyrussell: Maybe call it simply `popSequence` with the documentation explaining it undoes the nearest upstream `applySequence`?

artembilan: Makes sense. Will fix soon.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -217,6 +221,69 @@ public void setOutputProcessor(MessageGroupProcessor outputProcessor) {
...
+	 */
+	public void setExpireGroupsUponTimeout(boolean expireGroupsUponTimeout) {
+		this.expireGroupsUponTimeout = expireGroupsUponTimeout;
+	}
+
+	/**
+	 * Perform a {@link MessageBuilder#popSequenceDetails()} for output message or not.
+	 * Default to true.
+	 * This option plays an opposite role to the
+	 * {@link org.springframework.integration.splitter.AbstractMessageSplitter#setApplySequence(boolean)}.
```
garyrussell: It's not just splitter - router, pubsub etc; maybe just say "removes the sequence information added by the nearest upstream component with `applySequence=true` (for example splitter).

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
garyrussell: Still weird indentation.

spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
```diff
@@ -531,6 +532,25 @@ public void testExceptionTypeRouteFlow() {
...
+		QueueChannel replyChannel = new QueueChannel();
+		Message<String> request = MessageBuilder.withPayload("this is a test")
+				.setReplyChannel(replyChannel)
+				.build();
+		this.nestedScatterGatherFlowInput.send(request);
+		Message<?> bestQuoteMessage = replyChannel.receive(10000);
+		assertNotNull(bestQuoteMessage);
+		Object payload = bestQuoteMessage.getPayload();
+		assertThat(payload, instanceOf(String.class));
+//		assertThat(((List<?>) payload).size(), greaterThanOrEqualTo(1));
```
garyrussell: Why is this here? The payload is a String.

src/reference/asciidoc/aggregator.adoc
```diff
@@ -103,6 +103,13 @@ This method is invoked for aggregating messages as follows:
 
 NOTE: In the interest of code simplicity and promoting best practices such as low coupling, testability, and others, the preferred way of implementing the aggregation logic is through a POJO and using the XML or annotation support for configuring it in the application.
 
+Starting with version 5.1, after processing message group, an `AbstractCorrelatingMessageHandler` performs a `MessageBuilder.popSequenceDetails()` message headers modification for the proper splitter-aggregator scenario with several nested levels.
+It is done only if the message group release result is not a message or collection of messages.
+In that case a target `MessageGroupProcessor` is responsible for the `MessageBuilder.popSequenceDetails()` call during those messages building.
```
garyrussell: `s/during those messages building/while building those messages/`

src/reference/asciidoc/aggregator.adoc
```diff
@@ -103,6 +103,13 @@ This method is invoked for aggregating messages as follows:
...
+This functionality can be controlled by newly introduced `popSequence` `boolean` property, so the `MessageBuilder.popSequenceDetails()` can be disabled in some scenarios when correlation details have not been populated by the standard splitter.
```
garyrussell: `s/by newly introduced/by a new/`

src/reference/asciidoc/aggregator.adoc
```diff
@@ -103,6 +103,13 @@ This method is invoked for aggregating messages as follows:
...
+This property, essentially, undoes what has been done with the nearest upstream `applySequence = true` in the `AbstractMessageSplitter`.
```
garyrussell: ...by the nearest...

garyrussell (changes requested):
Also needs a migration guide indicating that the sequence is now popped by default for all MGPs.

garyrussell (approved):

garyrussell:
We're no longer cherry-picking; correct?

artembilan:
> We're no longer cherry-picking; correct?

No, we are not. There is a workaround with the custom `MessageGroupProcessor` and explicit `popSequenceDetails()` call there.
Plus an `AbstractAggregatingMessageGroupProcessor` has a breaking change.

Migration Guide note has been added: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.0-to-5.1-Migration-Guide#aggregator-changes

PR #2593, opened 2018-10-09
JMX: Quote object name values if not identifiers

garyrussell:
https://www.oracle.com/technetwork/java/javase/tech/best-practices-jsp-136021.html

The set of characters in a value is also limited. If special characters may
occur, it is recommended that the value be quoted, using ObjectName.quote. If
the value for a given key is sometimes quoted, then it should always be quoted.
By default, if a value is a string (rather than a number, say), then it should
be quoted unless you are sure that it will never contain special characters.

Practically, some special characters are allowed, but we will standardize on
allowed characters in java identifiers.


artembilan:
Uh! That's cool!

Merging...
Should we back-port it as a bug or too much breaking changes?

garyrussell:
I think we can back-port on-demand.

garyrussell:
I changed my mind https://github.com/spring-projects/spring-integration/pull/2594

Added a global property to enable.

PR #2600, opened 2018-10-16
INT-4546: Expose IMAP cancel interval via DSL

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4546

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/dsl/ImapMailInboundChannelAdapterSpec.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/dsl/ImapMailInboundChannelAdapterSpec.java
@@ -65,4 +65,19 @@ public class ImapMailInboundChannelAdapterSpec
 		return this;
 	}
 
+	/**
+	 * How often to recycle the idle task (in case of a silently dropped connection).
+	 * Seconds; default 120 (2 minutes).
+	 * @param interval the interval.
+	 * @return the spec.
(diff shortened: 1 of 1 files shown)


spring-integration-mail/src/main/java/org/springframework/integration/mail/dsl/ImapMailInboundChannelAdapterSpec.java
```diff
@@ -65,4 +65,19 @@ public ImapMailInboundChannelAdapterSpec shouldMarkMessagesAsRead(boolean should
...
+	 * How often to recycle the idle task (in case of a silently dropped connection).
+	 * Seconds; default 120 (2 minutes).
+	 * @param interval the interval.
+	 * @return the spec.
+	 * @see ImapMailReceiver#setCancelIdleInterval(long)
+	 * @since 5.0.10
+	 */
+	public ImapMailInboundChannelAdapterSpec cancelIdleInterval(long interval) {
+		assertReceiver();
+		this.receiver.setCancelIdleInterval(interval);
```
artembilan: How do you think: don't we have anything else missed to be exposed?

garyrussell: These are the only 3 properties on the `ImapMailReceiver` (there are only 3 setters).

artembilan:

PR #2601, opened 2018-10-17
INT-4544: Allow runtime MBeans (un)registered

artembilan:
JIRA: https://jira.spring.io/browse/INT-4544

* Fix `SourcePollingChannelAdapterFactoryBean` to register an
`outputChannelName` into the `SourcePollingChannelAdapter` for late
binding, especially in case of dynamic `IntegrationFlow` registrations.
This must be back-ported to `5.0.x`
* Implement `DestructionAwareBeanPostProcessor` in the
`IntegrationMBeanExporter` for the runtime beans tracking, e.g. via
dynamic `IntegrationFlow` registrations
* Refactor `IntegrationMBeanExporter` for static and dynamic beans
registration and destruction
* Remove unused properties; introduce some new for tracking beans and
their relationship


spring-integration-jmx/src/main/java/org/springframework/integration/monitor/IntegrationMBeanExporter.java
```diff
@@ -300,6 +306,93 @@ public void afterSingletonsInstantiated() {
 
 	}
 
+	private void registerProducer(MessageProducer messageProducer) {
```
garyrussell: Should be below the next method.

garyrussell (changes requested):
That's all after a quick review - will perform a more thorough review tomorrow (after travis).

spring-integration-jmx/src/main/java/org/springframework/integration/monitor/IntegrationMBeanExporter.java
artembilan: It is called from the `afterSingletonsInstantiated()` first time, so this is normal order for the private methods.

garyrussell: But it's also called from `postProcessAfterInitialization()`. I believe the rule is called methods should always be below the last place that calls it.

artembilan: Fine with me, but I've followed this rule so far: https://github.com/spring-projects/spring-framework/wiki/Code-Style#java-source-file-organization

> Note that private or protected methods called from method implementations should be placed immediately below the methods where they're used. In other words if there 3 interface method implementations with 3 private methods (one used from each), then the order of methods should include 1 interface and 1 private method in sequence, not 3 interface and then 3 private methods at the bottom.

garyrussell: Maybe it's just my interpretation then; I guess those guidelines don't specifically address this use case.

artembilan: OK. Fine with me.
Will fix. 😄 

Not sure though, what happened to the `org.springframework.integration.monitor.MessageSourceTests` on Travis...

garyrussell:
Looks like there are still test issues (instance already exists exceptions).

artembilan:
Yeah... Looks like MBean Server is not closed somewhere yet...

spring-integration-jmx/src/main/java/org/springframework/integration/monitor/IntegrationMBeanExporter.java
```diff
@@ -292,12 +296,100 @@ public void afterSingletonsInstantiated() {
...
 		}
 		catch (RuntimeException e) {
 			unregisterBeans();
 			throw e;
 		}
+	}
+
+	@Override
+	public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+		if (this.singletonsInstantiated) {
```
garyrussell: Since we are already at RC, perhaps we should add a `try {...} catch {...}` under here to fail soft (and log an error) instead of failing hard (just in case there are any regression issues) ?

garyrussell (changes requested):
And still travis failure.

artembilan:
OK! Finally it is green 😄 

Sorry for the mess in tests.

Note: When we configure runtime beans, we don't process plain `MessageHandler` and `MessageSource` beans. I would say without endpoints it does not make sense to have MBeans for plain beans.
Plus there is the problem that `postProcessAfterInitialization()` is called for `MessageHandler`/`MessageSource` and their endpoint, but at the moment to parse one of them there is no another yet...

garyrussell (approved):

garyrussell:
> This must be back-ported to 5.0.x

Separate PR?

PR #2602, opened 2018-10-18
INT-4545: Fix RPMHWrapper registration for ``@SA`

artembilan:
JIRA: https://jira.spring.io/browse/INT-4545

Improve the logic in the `AbstractMethodAnnotationPostProcessor` to
check for existing `MessageHandler` and be sure that we need to register
a `ReplyProducingMessageHandlerWrapper` bean

**Cherry-pick to 5.0.x**


garyrussell:
cherry-picked as 0418b7cbdc16ec847a208ac9ea543b082e81610a

PR #2603, opened 2018-10-18
INT-4530: Remove Meters

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4530

- implement `DisposableBean` in all meter-aware components
- remove meters in `destroy()`
- manually destroy annotation components created by `registerSingleton()`
  - changing these beans to register as bean definitions caused some subtle
    initialization side-effects; given we are post-RC, I felt it was too late
    to make such a large change
  - also CPXAC is not a `GenericApplicationContext`
  - hence a bean is registered to destroy them manually

### DIFF
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests.java
@@ -177,7 +177,7 @@ public class AmqpOutboundEndpointTests {
 				.setHeader(AmqpHeaders.CONTENT_TYPE, "application/json")
 				.build();
 		this.ctRequestChannel.send(message);
-		org.springframework.amqp.core.Message m = template.receive();
+		org.springframework.amqp.core.Message m = receive(template);
 		assertNotNull(m);
 		assertEquals("\"hello\"", new String(m.getBody(), "UTF-8"));
 		assertEquals("application/json", m.getMessageProperties().getContentType());
@@ -186,13 +186,24 @@ public class AmqpOutboundEndpointTests {
 		message = MessageBuilder.withPayload("hello")
 				.build();
 		this.ctRequestChannel.send(message);
-		m = template.receive();
+		m = receive(template);
 		assertNotNull(m);
 		assertEquals("hello", new String(m.getBody(), "UTF-8"));
 		assertEquals("text/plain", m.getMessageProperties().getContentType());
 		while (template.receive() != null) {
 			// drain
 		}
+	}
+
+	private org.springframework.amqp.core.Message receive(RabbitTemplate template) throws Exception {
+		int n = 0;
+		org.springframework.amqp.core.Message message = template.receive();
+		while (message == null && n++ < 100) {
+			Thread.sleep(100);
+			message = template.receive();
+		}
+		assertNotNull(message);
+		return message;
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
@@ -51,7 +51,6 @@ import org.springframework.integration.annotation.Poller;
 import org.springframework.integration.channel.DirectChannel;
 import org.springframework.integration.channel.MessagePublishingErrorHandler;
 import org.springframework.integration.config.IntegrationConfigUtils;
-import org.springframework.integration.config.annotation.MessagingAnnotationPostProcessor.Disposables;
 import org.springframework.integration.context.Orderable;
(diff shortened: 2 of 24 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -50,6 +51,7 @@
 import org.springframework.integration.channel.DirectChannel;
 import org.springframework.integration.channel.MessagePublishingErrorHandler;
 import org.springframework.integration.config.IntegrationConfigUtils;
+import org.springframework.integration.config.annotation.MessagingAnnotationPostProcessor.Disposables;
```
artembilan: This is going to cause class tangles: the `MessagingAnnotationPostProcessor` uses all the `AbstractMethodAnnotationPostProcessor` implementations.

Maybe just make that `Disposables` as package protected and refer to it everywhere it is necessary?

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java
```diff
@@ -272,4 +277,30 @@ protected String generateBeanName(String originalBeanName, Method method,
 		return this.postProcessors;
 	}
 
+	@Override
+	public void destroy() throws Exception {
+		this.beanFactory.getBean(Disposables.class).destroy();
```
artembilan: This is not god to call `getBean()` from the `destroy()`.
See some recent cause here: https://stackoverflow.com/questions/52785517/beancreationnotallowedexception-from-predestroy-annotation-from-prometheuspushg

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java
```diff
@@ -272,4 +277,30 @@ protected String generateBeanName(String originalBeanName, Method method,
...
+	}
+
+	public static class Disposables implements DisposableBean {
+
+		private final List<DisposableBean> disposables = new ArrayList<>();
+
+		public void addOne(DisposableBean disposable) {
```
artembilan: Why `addOne` if we don't have `addAll` ?
Just `add` ?

spring-integration-core/src/test/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsTests.java
```diff
@@ -170,6 +169,38 @@ public void testSend() {
...
+		newChannel.destroy();
+		try {
+			registry.get("spring.integration.send")
+				.tag("name", "newChannel")
+				.tag("result", "success")
+				.timer();
+			fail("Expected MeterNotFoundException");
+		}
+		catch (MeterNotFoundException e) {
+			// NOSONAR
```
artembilan: Can't we add some assertion here instead?

artembilan (changes requested):
Pretty cool stuff! 👍 

Just a couple concerns

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java
garyrussell: Right - I have changed the registration so the container will do the destruction.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests.java
```diff
@@ -177,7 +177,7 @@ public void adapterWithContentType() throws Exception {
 				.setHeader(AmqpHeaders.CONTENT_TYPE, "application/json")
 				.build();
 		this.ctRequestChannel.send(message);
-		org.springframework.amqp.core.Message m = template.receive();
+		org.springframework.amqp.core.Message m = receive(template);
```
artembilan: Why `AmqpTemplate.receive(long timeoutMillis)` doesn't work for us here?

artembilan (approved):
Just a question about possible alternative.
Otherwise LGTM.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests.java
garyrussell: Because it's an auto-delete queue; when we have a timeout, we create a temporary consumer; when `receive()` exits, the queue is deleted.

This test doesn't fail often; but it's failed for me a few times when using `--parallel`.

artembilan: Oh! Right. I forgot about a difference between `channel.basicGet()` `TemplateConsumer`.

Thanks!

Merging...

PR #2604, opened 2018-10-19
Extensible MicrometerMetricsCaptor and polishing

garyrussell:
- change `loadCaptor` to return any existing captor
- make builders and facades protected to allow subclassing
- change `AbstractMeter.getMeter()` to use generics
- add test case demonstrating meter name changes
- docs

### DIFF
--- a/src/reference/asciidoc/metrics.adoc
+++ b/src/reference/asciidoc/metrics.adoc
@@ -157,6 +157,9 @@ In addition, there are three `Gauge` Meters:
 * `spring.integration.handlers`: The number of `MessageHandlers` in the application.
 * `spring.integration.sources`: The number of `MessageSources` in the application.
 
+It is possible to customize the names and tags of `Meters` created by integration components by providing a subclass of `MicrometerMetricsCaptor`.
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/metrics.adoc
```diff
@@ -157,6 +157,9 @@ In addition, there are three `Gauge` Meters:
 * `spring.integration.handlers`: The number of `MessageHandlers` in the application.
 * `spring.integration.sources`: The number of `MessageSources` in the application.
 
+It is possible to customize the names and tags of `Meters` created by integration components by providing a subclass of `MicrometerMetricsCaptor`.
+The https://github.com/spring-projects/spring-integration/blob/master/spring-integration-core/src/test/java/org/springframework/integration/support/management/micrometer/MicrometerCustomMetricsTests.java[MicrometerCustomMetricsTests] test case shows a simple example of how to do that.
+You can also further customize the meters by overloading the `build()` methods on builder subclasses.
```
artembilan: Shound't we add this as a "What's New" note as well?

artembilan (changes requested):
Just one Docs concern.

Issue #2605, opened 2018-10-22
Question: Potential bug with sftp session reuse?


davidgoate:
Is there something we are supposed to do to close SFTP sessions? If we leave a service running for ore than a couple of days (anecdotal I know), this is the error we see:

>Caused by: java.io.IOException: inputstream is closed
com.jcraft.jsch.ChannelSftp.fill(ChannelSftp.java:2911) ~[jsch-0.1.54.jar!/:na]
com.jcraft.jsch.ChannelSftp.header(ChannelSftp.java:2935) ~[jsch-0.1.54.jar!/:na]
com.jcraft.jsch.ChannelSftp._put(ChannelSftp.java:583) ~[jsch-0.1.54.jar!/:na]

>org.springframework.integration.sftp.session.SftpSession.write(SftpSession.java:155) ~[spring-integration-sftp-5.0.9.RELEASE.jar!/:5.0.9.RELEASE]

If we restart the service, it'll immediately start "working" again.

My guess as perhaps some sort of connection pool is remaining open, or maybe there is some setting I am omitting to ensure the connections/sessions are handled properly.

# Code:

```java
@Configuration
@RequiredArgsConstructor
@EnableConfigurationProperties(SftpProperties.class)
@IntegrationComponentScan
@EnableIntegration
public class SftpConfiguration {
    private final SftpProperties properties;

    @Bean
    @ConditionalOnMissingBean(name = "sftpSessionFactory")
    public SessionFactory<ChannelSftp.LsEntry> sftpSessionFactory() {
        final DefaultSftpSessionFactory sessionFactory = new DefaultSftpSessionFactory();
        sessionFactory.setHost(properties.getHost());
        sessionFactory.setPort(properties.getPort());
        sessionFactory.setUser(properties.getUser());
        sessionFactory.setPassword(properties.getPassword());
        sessionFactory.setAllowUnknownKeys(true);
        return new CachingSessionFactory<>(sessionFactory);
    }

    @Bean
    @ServiceActivator(inputChannel = "test")
    public MessageHandler handler(@Autowired final SessionFactory<ChannelSftp.LsEntry> sftpSessionFactory) {
        final SftpMessageHandler handler = new SftpMessageHandler(sftpSessionFactory);
        handler.setRemoteDirectoryExpression(new LiteralExpression(properties.getDirectory()));
        handler.setFileNameGenerator(message -> namingService.getName());
        handler.setUseTemporaryFileName(false);
[...]

PR #2606, opened 2018-10-22
GH-2605: (S)FTP test cached sessions

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/2605

### DIFF
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/CachingSessionFactoryTests.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/CachingSessionFactoryTests.java
@@ -124,6 +124,7 @@ public class CachingSessionFactoryTests {
 
 		private volatile boolean open = true;
 
+		@SuppressWarnings("unused")
 		private boolean testCalled;
 
 		private TestSession(String id) {
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/CachingSessionFactoryTests.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/CachingSessionFactoryTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2016 the original author or authors.
+ * Copyright 2013-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/CachingSessionFactoryTests.java
```diff
@@ -61,6 +62,7 @@ public void testCacheAndReset() {
 		assertTrue(sess1.isOpen());
 		sess1 = cache.getSession();
 		assertEquals("session:1", TestUtils.getPropertyValue(sess1, "targetSession.id"));
+		assertTrue((TestUtils.getPropertyValue(sess1, "targetSession.testCalled", Boolean.class)));
```
artembilan: The `TestSession` is a part of this class, so its `private` properties are visible.
And afterwards we don't need that `@SuppressWarnings("unused")` as well.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
```diff
@@ -278,4 +278,20 @@ public ChannelSftp getClientInstance() {
...
 	}
 
+	@Override
+	public boolean test() {
+		return isOpen() && doTest();
+	}
+
+	private boolean doTest() {
+		try {
+			this.channel.stat(this.channel.getHome());
```
artembilan: Why `jschSession.sendKeepAliveMsg()` doesn't work for us?

src/reference/asciidoc/ftp.adoc
```diff
@@ -1412,6 +1413,9 @@ Starting with Spring Integration 3.0, the `CachingConnectionFactory` provides a
 When invoked, all idle sessions are immediately closed and in-use sessions are closed when they are returned to the cache.
 New requests for sessions establish new sessions as necessary.
 
+Starting with version 5.1, the `CachingSessionFactory` has a new poperty `testSession`.
```
artembilan: Typo: p**r**operty

artembilan (changes requested):
Just a couple questions.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/CachingSessionFactoryTests.java
garyrussell: ? 

It's wrapped in a `CachedSession` the `CachedSession.targetSession` is private.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
garyrussell: That command doesn't wait for a reply so we get false positives (e.g. when pulling the network cable).

PR #2608, opened 2018-10-24
INT-4520 Make IntegrationGraphServer customizable

artembilan:
JIRA: https://jira.spring.io/browse/INT-4520

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationGraphServer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationGraphServer.java
@@ -18,12 +18,16 @@ package org.springframework.integration.graph;
 
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
@@ -42,9 +46,9 @@ import org.springframework.integration.router.RecipientListRouter.Recipient;
 import org.springframework.integration.router.RecipientListRouterManagement;
 import org.springframework.integration.support.context.NamedComponent;
 import org.springframework.integration.support.management.MappingMessageRouterManagement;
+import org.springframework.lang.Nullable;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
-import org.springframework.util.StringUtils;
 
 /**
  * Builds the runtime object model graph.
@@ -67,9 +71,15 @@ public class IntegrationGraphServer implements ApplicationContextAware, Applicat
 
 	private String applicationName;
 
+	private Function<NamedComponent, Map<String, Object>> additionalPropertiesCallback;
+
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
 		this.applicationContext = applicationContext; //NOSONAR (sync)
+	}
+
+	protected ApplicationContext getApplicationContext() {
+		return this.applicationContext;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationGraphServer.java
```diff
@@ -99,11 +127,27 @@ public Graph getGraph() {
...
+	 * @see #getGraph()
+	 */
+	public Graph rebuild() {
+		return buildGraph();
+	}
+
+	/**
+	 * Get beans for provided type from the application context.
+	 * This method can be extended for some custom logic, e.g. get beans
+	 * form the parent application context as well.
```
garyrussell: typo `form`

for **the** provided type

spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationNode.java
```diff
@@ -76,6 +81,27 @@ public Stats getStats() {
 		return this.properties.size() == 0 ? null : this.properties;
 	}
 
+	/**
+	 * Add extra property to the node.
+	 * @param name the name for property
+	 * @param value the value of the property
+	 * @since 5.1
+	 */
+	public void addProperty(String name, Object value) {
```
garyrussell: null checks?

src/reference/asciidoc/graph.adoc
```diff
@@ -147,6 +147,25 @@ It is also used in the `links` element to represent a relationship (connection)
 The `input` and `output` attributes are for the `inputChannel` and `outputChannel` properties of the `AbstractEndpoint`, `MessageHandler`, `SourcePollingChannelAdapter`, or `MessageProducerSupport`.
 See the next section for more information.
 
+Starting with version 5.1, the `IntegrationGraphServer` accepts a `Function<NamedComponent, Map<String, Object>> additionalPropertiesCallback` for population additional properties on the `IntegrationNode` for particular `NamedComponent`.
```
garyrussell: for population **of** additional

for **a** particular

src/reference/asciidoc/graph.adoc
```diff
@@ -147,6 +147,25 @@ It is also used in the `links` element to represent a relationship (connection)
...
+For example you can expose a `SmartLifecycle` `autoStartup` and `running` properties into the target graph:
+
```
garyrussell: expose **the** ...

src/reference/asciidoc/whats-new.adoc
```diff
@@ -210,3 +210,9 @@ e.g. `org.springframework.integration:type=MessageChannel,` `name="input#foo.myG
 
 It is now simpler to customize the standard Micrometer meters created by the framework.
 See <<micrometer-integration>> for more information.
+
+[[x51.-integration-graph]]
+=== The Integration Graph Customization
```
garyrussell: remove `The`.

spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationNode.java
```diff
@@ -76,6 +81,27 @@ public Stats getStats() {
 		return this.properties.size() == 0 ? null : this.properties;
```
garyrussell: `@Nullable` on this getter? Or, maybe we should just always return it, even if empty.

Also, now that we have a mechanism to modify the map, maybe we should return an `UnmodifableMap` - if so, it should be a field instead of wrapping the map on every call.

garyrussell (changes requested):
Cool stiuff

spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationNode.java
artembilan: Do you mean we should use a `BiConsumer<NamedComponent, Map<String, Object>>`  instead of that `Function` for additional properties?

garyrussell: ?

No; I just mean that now we have `IN.addProperty()` and `addProperties()` we no longer need to expose a direct reference to properties field.

garyrussell (approved):

PR #2609, opened 2018-10-24
INT-4547: (S)FTP RFOG MPUT with collection payload

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4547


artembilan (commented):
Only one question.

PR #2613, opened 2018-10-25
Fix stylesheet location for HTML Asciidoctor docs

garyrussell:
- also fix `{version}` to `{project-version}`.

(HTML5 doc is in `build/asciidoc/html5`).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -781,7 +781,7 @@ asciidoctor {
 		toc2: '',
 		'compat-mode': '',
 		imagesdir: '',
-		stylesheet: '../../../src/reference/asciidoc/stylesheets/golo.css',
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -781,8 +781,7 @@ asciidoctor {
 		toc2: '',
 		'compat-mode': '',
 		imagesdir: '',
-		stylesdir: "stylesheets/",
-		stylesheet: 'golo.css',
+		stylesheet: '../../../src/reference/asciidoc/stylesheets/golo.css',
```
artembilan: Wait. Can't it be done with something like `${rootProject.projectDir}/src/` ?

artembilan (changes requested):

PR #2617, opened 2018-10-26
Core Sonar fixes

garyrussell:

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
@@ -226,7 +226,7 @@ public abstract class IntegrationObjectSupport implements BeanNameAware, NamedCo
 	}
 
 	public ConversionService getConversionService() {
-		if (this.conversionService == null && this.beanFactory != null) {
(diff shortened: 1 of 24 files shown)


spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
```diff
@@ -226,7 +226,7 @@ protected void setTaskScheduler(TaskScheduler taskScheduler) {
 	}
 
 	public ConversionService getConversionService() {
-		if (this.conversionService == null && this.beanFactory != null) {
+		if (this.conversionService == null && this.beanFactory != null) { // NOSONAR double lock fetching bean, not creating
```
artembilan: I would say we don't need this double check at all and even any synchronization here.
We really just try to obtain a bean, so even if without synchronization we would call it several times, we are going to end up with the same instance.

So, my opinion is like this:
```
if (this.conversionService == null && this.beanFactory != null) {
    this.conversionService = IntegrationUtils.getConversionService(this.beanFactory);
}
```

artembilan (changes requested):
Only one concern.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/2b1976ce6f10b491437e5c0d49fe05921ba53478

PR #2618, opened 2018-10-31
INT-4549: Avoiding Aggregator Deadlocks

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4549

**cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -20,7 +20,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -64,9 +63,6 @@ import org.springframework.scheduling.TaskScheduler;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 
-import reactor.util.function.Tuple2;
-import reactor.util.function.Tuples;
-
 /**
  * Abstract Message handler that holds a buffer of correlated messages in a
  * {@link MessageStore}. This class takes care of correlated groups of messages
@@ -107,10 +103,6 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 	private final Map<UUID, ScheduledFuture<?>> expireGroupScheduledFutures = new ConcurrentHashMap<>();
 
 	private final Set<Object> groupIds = ConcurrentHashMap.newKeySet();
-
-	private final ThreadLocal<Tuple2<Object, Message<?>>> deferredSends = new ThreadLocal<>();
-
-	private final ThreadLocal<List<Message<?>>> deferredDiscards = new ThreadLocal<>();
 
 	private MessageGroupProcessor outputProcessor;
 
@@ -464,7 +456,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		UUID groupIdUuid = UUIDConverter.getUUID(correlationKey);
 		Lock lock = this.lockRegistry.obtain(groupIdUuid.toString());
 
-		boolean maybeOutput = false;
+		boolean noOutput = true;
 		lock.lockInterruptibly();
 		try {
 			ScheduledFuture<?> scheduledFuture = this.expireGroupScheduledFutures.remove(groupIdUuid);
@@ -489,8 +481,8 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 				if (this.releaseStrategy.canRelease(messageGroup)) {
 					Collection<Message<?>> completedMessages = null;
 					try {
-						completedMessages = completeGroup(message, correlationKey, messageGroup);
-						maybeOutput = true;
+						completedMessages = completeGroup(message, correlationKey, messageGroup, lock);
+						noOutput = false;
 					}
 					finally {
 						// Possible clean (implementation dependency) up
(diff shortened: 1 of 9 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -104,6 +108,10 @@
 
 	private final Set<Object> groupIds = ConcurrentHashMap.newKeySet();
 
+	private final ThreadLocal<Tuple2<Object, Message<?>>> deferredSends = new ThreadLocal<>();
+
+	private final ThreadLocal<List<Message<?>>> deferredDiscards = new ThreadLocal<>();
```
artembilan: Would you mind to share some explanation for this `ThreadLocal` trick?
Why simple `if...else` around `sendOutputs(result, message);` to release lock if present as an extra argument for the `completeGroup()` is not enough?

Something similar could be done with the `discardMessage()` as well.

Thanks

src/reference/asciidoc/aggregator.adoc
```diff
@@ -279,6 +279,47 @@ A `LockRegistry` is used to obtain a lock for the resolved correlation ID.
 A `DefaultLockRegistry` is used by default (in-memory).
 For synchronizing updates across servers where a shared `MessageGroupStore` is being used, you must configure a shared lock registry.
 
+[[aggregator-deadlocks]]
```
artembilan: Should we add `What's New` note on the matter?

artembilan (changes requested):
Nothing critical, buy confusing...

Thanks

garyrussell:
Yes, your solution is cleaner, will polish tomorrow; I am thinking we change the default to true in 5.2 and your solution would avoid the TL in the primary path then.

I don't think we need a what's new - it's taken 10 years for someone to hit this.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -479,11 +498,14 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
...
 			}
 			else {
-				discardMessage(message);
+				discardMessage(message, lock);
+				noOutput = false;
 			}
 		}
 		finally {
-			lock.unlock();
+			if (noOutput || !this.releaseLockBeforeSend) {
```
artembilan: I believe this `noOutput` is for the case of error.
How about if we `releaseLockBeforeSend` and unlock in the `completeGroup()` before send, but then send fail?
So, `noOutput` remains `true`, but lock is unlocked already anyway.

Isn't going to be:
```
* @throws IllegalMonitorStateException if the current thread does not
*         hold this lock
```
?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
garyrussell: It's not for the case of error; it's for when we only added a message to a group and didn't produce any output. But, you are right, if sendOutputs fails, we'll try to unlock again.

`RentrantLock` has `isHeldByCurrentThread()` but we can't be assured it will be one of those.

I think the only alternative is a `ThreadLocal`.

artembilan: OK. How about to move `noOutput = false;` to the `final` block:
```
try {
    discardMessage(message, lock);
}
finally {
    noOutput = false;
}

```

garyrussell: I think my latest commit solves it.

artembilan: 👍 

Let me pull it locally for the final review!

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -736,58 +769,67 @@ protected void expireGroup(Object correlationKey, MessageGroup group) {
...
+			if (result instanceof Collection<?>) {
+				verifyResultCollectionConsistsOfMessages((Collection<?>) result);
+				partialSequence = (Collection<Message<?>>) result;
 			}
-			else {
-				messageBuilder = getMessageBuilderFactory()
-						.withPayload(result)
-						.copyHeaders(message.getHeaders());
+
+			if (this.popSequence && partialSequence == null && !(result instanceof Message<?>)) {
```
artembilan: Hm. the `popSequence` is there since `5.1`.
I don't think this change is going to be back-ported to `5.0` so easy...

Maybe as a separate PR?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
garyrussell: Or don't backport since there is a workaround?

artembilan: OK. Let's don't back-port!

Please, proceed then with all the required changes to reflect only `5.1.1` version.

Thanks

src/reference/asciidoc/aggregator.adoc
```diff
@@ -279,6 +279,47 @@ A `LockRegistry` is used to obtain a lock for the resolved correlation ID.
...
+===== Avoiding Deadlocks
+
+As discussed above, when message groups are mutated (messages added or released) a lock is held.
+
+Consider the following flow:
+
+====
+[source]
+----
+...->aggregator1-> ... ->aggregatoor2-> ...
```
aantoniadis: typo

artembilan: Found it. Fixed in my local copy before merge. 

Thank you for review anyway!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/6861a16b95d110650b618582c20ebf4b64bbdc39

PR #2619, opened 2018-11-06
Add keySet() method to MetadataStore and MetadataStoreReaper

dturanski:
JIRA: https://jira.spring.io/browse/INT-4522

* Add `MetadataStoreReaper`
* Add `keySet()` method to `MetadataStore` and all implementations.


spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStoreReaper.java
```diff
@@ -0,0 +1,175 @@
...
+
+	/**
+	 * Remove entries from a {@link MetadataStore} based on a maximum time-to-live.
+	 * @param metadataStore the MetadataStore.
+	 * @param timeToLive the maximum time-to-live.
+	 */
+	public MetadataStoreReaper(MetadataStore metadataStore, Duration timeToLive) {
+		this(metadataStore, value -> {
+			try {
+				long timestamp = Long.valueOf(value);
```
artembilan: I would say that this is too custom solution and should not be as a part of the Framework.

Can we keep this `MetadataStoreReaper` in your target project?
That's just a bad luck that your use-case is based on the fact that we store a `timestamp` in the value, but there are many other scenarios when it is not true. 

So, 👎 from me to merge this into `master`.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
```diff
@@ -16,17 +16,6 @@
 
 package org.springframework.integration.zookeeper.metadata;
 
-import static org.hamcrest.Matchers.containsString;
```
artembilan: This imports reorder change is wrong from this project Checkstyle perspective.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -245,5 +253,10 @@ public String remove(String key) {
 		}
 		return null;
 	}
-
+	@Override
+	public Set<String> keySet() {
+		Set<String> keys = new HashSet<>();
+		keys.addAll(this.jdbcTemplate.queryForList(this.getAllKeysQuery, new Object[]{this.region}, String.class));
```
artembilan: There is this variant to simplify the code here: `queryForList(String sql, Class<T> elementType, @Nullable Object... args)`

spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStoreReaper.java
```diff
@@ -0,0 +1,175 @@
...
+		Assert.notNull(isExpired, "'isExpired predicate' cannot be null");
+		this.metadataStore = metadataStore;
+		this.isExpired = isExpired;
+	}
+
+	@Override
+	public void run() {
+		this.metadataStore.keySet().stream().forEach(k ->
+		{
+			if (this.isExpired.test(this.metadataStore.get(k))) {
```
artembilan: This technique is pretty interesting and I would say that it is a time to add `expireEntries(Predicate<String>)` into the `MetadataStore` contract! 

Although we might need to think about something like `entrySet` from the store at once or as a stream do not overhead the memory. 

On the other hand it should be better to rely on the native *remove-by-filter* functionality in the store, e.g. `DELETE FROM WHERE ...` for JDBC. However in this case we need to supply a converter from String to timestamp in that query per se...

Need to think more...

artembilan (changes requested):
That's my review so far

spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStore.java
```diff
@@ -16,6 +16,9 @@
 
 package org.springframework.integration.metadata;
 
+import java.util.Set;
+import java.util.stream.Stream;
```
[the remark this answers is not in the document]
dturanski: My bad. I ran `clean test install` - doesn't run checkstyle I guess.

spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStoreReaper.java
```diff
@@ -0,0 +1,175 @@
...
+	 * Remove entries from a {@link MetadataStore} based on a maximum time-to-live.
+	 * @param metadataStore the MetadataStore.
+	 * @param timeToLive the maximum time-to-live.
+	 */
+	public MetadataStoreReaper(MetadataStore metadataStore, Duration timeToLive) {
+		this(metadataStore, value -> {
+			try {
+				long timestamp = Long.valueOf(value);
+				Duration age = Duration.between(LocalDateTime.ofInstant(Instant.ofEpochMilli(timestamp),
+								TimeZone.getDefault().toZoneId()), LocalDateTime.now());
```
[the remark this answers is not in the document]
dturanski: Yeah,  the API is still f-ed up.

spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStoreReaper.java
dturanski: IMHO, The reaper is a good thing to have in SI and doesn't hurt anything. People don't have to use it.  If they choose to use it, the value should include date-time info. But to your point, I can will remove this constructor and function and implement it at the app level where it makes sense. The reaper should be a common component though, since it addresses a core problem with the metadataStore.   

A more comprehensive approach I think would require the implementation to support the clean up, similar to the way `MessageGroupReaper` delegates to the `MessageGroupStore`.  All `MetadataStore`s would have to implement the clean up. That's the only way I can see to have a universal solution without violating encapsulation.  

My suggestion is to keep this and refactor for a more comprehensive approach if there is demand.

dturanski: I just saw your comment below:  `... I would say that it is a time to add expireEntries(Predicate<String>) into the MetadataStore contract!`

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
dturanski: This passed checkstyle for me

spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStore.java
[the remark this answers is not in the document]
garyrussell: `clean check` (runs `test`)

or

`clean testAll` (runs `check`) - runs some additional longer-running tests.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
```diff
@@ -49,10 +38,21 @@
...
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+import static org.springframework.integration.test.matcher.EqualsResultMatcher.equalsResult;
+import static org.springframework.integration.test.matcher.EventuallyMatcher.eventually;
```
artembilan: Yeah... So, time to remove `<suppress files="[\\/]test[\\/]" checks="ImportOrder" />` for tests! 😄 

There is an idea to move to the https://github.com/spring-io/spring-javaformat, but it doesn't look like they support imports reorder...

artembilan: OK. So, the Checkstyle for import orders in tests as well has been pushed to master: https://github.com/spring-projects/spring-integration/commit/648aeb8e966c63e49a42ce6cb79cafab56ae73d7

Therefore you don't have choice unless revert all the import order changes 😄

spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStore.java
```diff
@@ -51,11 +55,22 @@
 
 	/**
 	 * Remove a value for the given key from this MetadataStore.
+	 *
```
artembilan: We prefer do not add blank lines in the method Javadocs: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-integration-core/src/main/java/org/springframework/integration/metadata/MetadataStoreReaper.java
artembilan: I still think this is bad idea to enforce people to iterate all the entries in the framewrok's component.

I know that we have something similar in the `MessageGroupStore`, but that is a different story and has some hard-tying with an aggregator logic.

Since here we just do *remove-by-filter* that's definitely fall to the good native `DELETE` support in the target `MetadataStore` implementation.

For some exotic use-cases like yours to remove by the timestamp representation in the `value` we already exposed that `keySet()` for iteration, however I still think that it would be better to do that on the target store directly. We always can convert string to `timestamp` in the `WHERE` clause.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -245,5 +253,10 @@ public String remove(String key) {
 		}
 		return null;
 	}
-
```
artembilan: Every member in the class must be wrapped with blank lines.
Right now you are sticking the next method to the previous.

artembilan (changes requested):

artembilan:
@dturanski ,

gentle ping.

If you would like to have this in the upcoming `5.1.1` (next Monday), then let's consider to resolve all the issues here!

And one more time: I don't want to see that reaper in the `master` ...

@garyrussell , WDYT?

Thanks

garyrussell:
Sorry; I haven't been following this discussion.

Why are we exposing a `keySet()` rather than something like `iterateOverContents(BiPredicate<String, String> removeElement)`; we can then iterate more efficiently over the `entrySet()`.

garyrussell:
Plus the PR title needs editing since there's no longer a reaper.

dturanski:
@garyrussell - Agree. I think there are better ways to do this, but more involved to implement that in every metadata store. As Artem suggested, the most efficient way is probably to ensure every MDS includes a `time_created` field and use repository delete or query by timestamp earlier than.  I'm ok if we close this in favor of something like that. I'm not sure if there are other use cases that would need `keySet()`.

artembilan:
> `iterateOverContents`

I still don't want such an overhead to iterate over all the entries in the store. Even if we won't load all of them into the memory at once, we will get a big impact for round trips to the store during iteration for each entry.

But I might be in a minority, so feel free to introduce such an intermediate solution until we will come up with the proper native `expire()` solution.

garyrussell:
Well we don't have to keep round-tripping for each row; with a `RowCallbackHandler` we can call the `BiPredicate` for each row as the driver handles the batching/cursor and the template hands them to us one-at-a-time.

That said, I agree that a new row with the timestamp would be better and the deletions can then be done with a single update query.

Not sure how that would work for K-V stores like redis and gemfire, though,

artembilan:
Right. That's why I said `ExpirableMetadaStore` just because not all of them cam provide such a functionality.
For Redis we have a request to implement metadata store based on the top-level keys and there is already an expiration functionality.

Meanwhile let's give it a shot with this `BiPredicate` to be consulted during iteration over entries with as native iterator as possible! Like you mentioned this `RowCallbackHandler` in JDBC.
Not sure though that we could be able to delete the row just immediately in the cursor, but that's already implementation details...

PR #2621, opened 2018-11-08
INT-4551: Fail if old Java DSL jar is present

artembilan:
JIRA: https://jira.spring.io/browse/INT-4551

**Cherry-pick to 5.0.x**


garyrussell:
Merged as 2d06169f25f8e3e0875ccd1507d2212545b31b98 with some message polishing and cherry-picked.

PR #2622, opened 2018-11-09
INT-4550: Disallow multi aggregators on same MGS

artembilan:
JIRA: https://jira.spring.io/browse/INT-4550

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
@@ -112,8 +112,8 @@ public abstract class AbstractMessageGroupStore extends AbstractBatchingMessageG
 					this.expiryCallbacks.stream()
 							.anyMatch(UniqueExpiryCallback.class::isInstance);
 
-			if (!uniqueExpiryCallbackPresent && logger.isErrorEnabled()) {
-				logger.error("Only one instance of 'UniqueExpiryCallback' can be registered in the " +
+			if (!uniqueExpiryCallbackPresent && this.logger.isErrorEnabled()) {
+				this.logger.error("Only one instance of 'UniqueExpiryCallback' can be registered in the " +
 						"'MessageGroupStore'. Use a separate 'MessageGroupStore' for each aggregator/resequencer.");
 			}
 		}
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
@@ -27,6 +27,7 @@ import org.springframework.jmx.export.annotation.ManagedAttribute;
 import org.springframework.jmx.export.annotation.ManagedOperation;
 import org.springframework.jmx.export.annotation.ManagedResource;
 import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
 
 /**
  * @author Dave Syer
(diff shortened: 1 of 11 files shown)


artembilan:
OK! There is no other way, unless use some marker interface.
More over we can't hide it in the `AbstractCorrelatingMessageHandler` because we don't want to have a package tangle when we refer to it from the `AbstractMessageGroupStore`.

So, I fully copied/pasted your solution from the #2254 😃 

Enjoy review!

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
```diff
@@ -110,6 +108,16 @@ public void setLazyLoadMessageGroups(boolean lazyLoadMessageGroups) {
...
 	public void registerMessageGroupExpiryCallback(MessageGroupCallback callback) {
+		if (callback instanceof UniqueExpiryCallback) {
+			boolean uniqueExpiryCallbackPresent =
+					this.expiryCallbacks.stream()
+							.anyMatch(UniqueExpiryCallback.class::isInstance);
+
+			Assert.isTrue(!uniqueExpiryCallbackPresent,
+					"Only one instance of 'UniqueExpiryCallback' can be registered in the 'MessageGroupStore'." +
+							" Use separate 'MessageGroupStore' for each aggregator/resequencer.");
+		}
```
garyrussell: Perhaps `logger.error()` in 5.1 and assert in 5.2?

src/reference/asciidoc/aggregator.adoc
```diff
@@ -812,6 +812,13 @@ The callback has direct access to the store and the message group so that it can
 The `MessageGroupStore` maintains a list of these callbacks, which it applies, on demand, to all messages whose timestamps are earlier than a time supplied as a parameter (see the `registerMessageGroupExpiryCallback(..)` and `expireMessageGroups(..)` methods, described earlier).
 For more detail, see <<reaper>>.
 
+IMPORTANT: It is important do not use the same `MessageGroupStore` instance store in different aggregator components, when you are intended to rely on the `expireMessageGroups` functionality.
```
garyrussell: important **not to use** ... **store instance** ... when you **intend** to rely

src/reference/asciidoc/aggregator.adoc
```diff
@@ -812,6 +812,13 @@ The callback has direct access to the store and the message group so that it can
...
+Every `AbstractCorrelatingMessageHandler` registers its own `MessageGroupCallback` based on the `forceComplete()` callback.
+This way each group for expiration may be completed or discarded by wrong aggregator.
```
garyrussell: If the same store is used in multiple components, each expiring group may be ... by **the** wrong ...

src/reference/asciidoc/aggregator.adoc
```diff
@@ -812,6 +812,13 @@ The callback has direct access to the store and the message group so that it can
...
+Starting with version 5.0.10, a `UniqueExpiryCallback` is used from the `AbstractCorrelatingMessageHandler` for registration callback in the `MessageGroupStore`.
```
garyrussell: ...for **the** registration...

src/reference/asciidoc/aggregator.adoc
```diff
@@ -812,6 +812,13 @@ The callback has direct access to the store and the message group so that it can
...
+The `MessageGroupStore`, in turn, checks for presence an instance of this class and fails with an `IllegalArgumentException` if one already present in the callbacks set.
```
garyrussell: The`MessageGroupStore` ensures that only one callback is an instance of `UniqueExpiryCallback`... if one **is** already

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
artembilan: I agree about the `logger.error()`, because this exception is going to be some kind of breaking change in the point releases.
But at the same time I don't agree about `5.2`: in JIRA we talked about something like additional `owner` property on the `MessageGroup` to allow to share a `MessageStore`, but at the same time skip those groups from expiration in the aggregator which were not initiated by this particular bean.

artembilan:
Gary, have you missed to review this after the last push?

Thanks

garyrussell (approved):

garyrussell:
Cherry-picked as d1b30466382d9c11720980475f096c3a47e6e880 after resolving conflicts.

PR #2624, opened 2018-11-11
Fixed typo in ip.adoc

developerworks:
Cancel previous PR of this.


garyrussell:
@developerworks Thanks for the contribution!

In future, please note that it's perfectly ok to add more commits to an open PR; we squash multiple commits into one when merging, after review.

Thanks again.

PR #2633, opened 2018-11-16
INT-4557 Improve query cache in JDBC channel store

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4557

- Avoid hashing the query String on each call.
- Use a concurrent map.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
@@ -476,14 +476,8 @@ public class JdbcChannelMessageStore implements PriorityCapableChannelMessageSto
 	 * @return A transformed query with replacements.
 	 */
 	protected String getQuery(Query queryName, Supplier<String> queryProvider) {
-		String query = this.queryCache.get(queryName);
-
-		if (query == null) {
-			query = StringUtils.replace(queryProvider.get(), "%PREFIX%", this.tablePrefix);
-			this.queryCache.put(queryName, query);
-		}
-
-		return query;
+		return this.queryCache.computeIfAbsent(queryName,
+				k -> StringUtils.replace(queryProvider.get(), "%PREFIX%", this.tablePrefix));
 	}
 
 	/**
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
@@ -17,15 +17,16 @@
 package org.springframework.integration.jdbc.store;
 
 import java.sql.Types;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -448,23 +462,25 @@ public MessageGroup getMessageGroup(Object groupId) {
...
 	 * accesses will be resolved from the cache.
-	 * @param sqlQuery The SQL query to be transformed.
+	 * @param queryName The {@link Query} to be transformed.
+	 * @param queryProvider a supplier to provide the query template.
 	 * @return A transformed query with replacements.
 	 */
-	protected String getQuery(String sqlQuery) {
-		String query = this.queryCache.get(sqlQuery);
+	protected String getQuery(Query queryName, Supplier<String> queryProvider) {
+		String query = this.queryCache.get(queryName);
```
artembilan: Why doesn't `computeIfAbsent()` work for us here?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -409,7 +422,8 @@ public void afterPropertiesSet() throws Exception {
 	@Override
 	public MessageGroup addMessageToGroup(Object groupId, final Message<?> message) {
 		try {
-			this.jdbcTemplate.update(getQuery(this.channelMessageStoreQueryProvider.getCreateMessageQuery()),
+			this.jdbcTemplate.update(getQuery(Query.CREATE_MESSAGE,
+						() -> this.channelMessageStoreQueryProvider.getCreateMessageQuery()),
```
artembilan: What is just wrong with calling the since it just returns a constant anyway?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -448,23 +462,25 @@ public MessageGroup getMessageGroup(Object groupId) {
 	@ManagedAttribute
 	public int getMessageGroupCount() {
 		return this.jdbcTemplate.queryForObject(
-				getQuery("SELECT COUNT(DISTINCT GROUP_KEY) from %PREFIX%CHANNEL_MESSAGE where REGION = ?"),
+				getQuery(Query.COUNT_GROUPS,
+						() -> "SELECT COUNT(DISTINCT GROUP_KEY) from %PREFIX%CHANNEL_MESSAGE where REGION = ?"),
```
artembilan: This supplier is pointless as well: it doesn't do any calculation and just returns a constant.

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
garyrussell: It does work; will fix; I simply changed the existing code.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
garyrussell: Because there is no need to call every time - the supplier is only invoked if the cache does not already contain the key. Even though it only returns a constant, it's an unnecessary method call.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
garyrussell: Simpler than adding a second cache method just for this one case.

PR #2635, opened 2018-11-19
Fix LambdaMessageProcessor for conversion

artembilan:
SO: https://stackoverflow.com/questions/53378821

There is no need to jump into the `MessageConverter` if payload type
is assignable to the target type


tdorzhi:
This seems to be the case with `MappingJackson2MessageConverter`, it doesn't care if payload is assignable to targetType and for some  reason does `mapper.readValue(payload.toString(), targetType)`, which results in JSON mapping error

artembilan:
@tdorzhi ,

Somehow your comment doesn't help and I can't infer an issue from it.

Please, raise a new GH issue (or StackOverflow question) with much more information to determine the problem and some ideas how to fix it.
Otherwise we can't help you and you just confuse us so we can't help others.

Thanks for understanding.

PR #2636, opened 2018-11-19
Log error for class cast exception on lambda

garyrussell:
**cherry-pick to 5.0.x**


artembilan:
Travis failure:
```
org.springframework.integration.dsl.LambdaMessageProcessorTests > testMessageAsArgumentLambda FAILED
    org.springframework.messaging.MessageHandlingException at LambdaMessageProcessorTests.java:82
        Caused by: java.lang.ClassCastException at LambdaMessageProcessorTests.java:82
```

garyrussell:
Doh.

artembilan (changes requested):
Also Oleg mentioned Javadocs improvements.
Would you mind to tackle that as well?

Plus, maybe it makes sense to document it in the Reference Manual as well: https://docs.spring.io/spring-integration/docs/5.0.9.RELEASE/reference/html/java-dsl.html#java-dsl-transformers ?

I know we have similar problem with many other EIP-methods, but let's start from something like this `.transform()`!

garyrussell:
ok; working on JavaDocs and docs.

artembilan:
... and cherry-picked to `5.0.x` after removing `convert()` - was introduced since `5.1`.

PR #2644, opened 2018-11-30
INT-4385: Clarify <jms> namespace in jms.adoc

artembilan:
JIRA: https://jira.spring.io/browse/INT-4385

It is slightly unclear for end-users that `<jms:listener-container/>`
is just a syntax sugar and the real JMS container is created by the
`<jms:listener>` sub-element

* Add a NOTE to the `jms.adoc` to explain such a confuse and what and
how is going to happen if the intention to use a `<jms:listener-container/>`
for Spring Integration channel adapters.


garyrussell:
Merged as 4ade0ceaf48d4baed8a1451db67d87f77a028785 with minor polish.

PR #2647, opened 2018-12-04
INT-4564: ACMH: Fix erroneous error log

garyrussell:
JIRA: https://jira.spring.io/browse/INT-4564

INT-4550 added an error log when multiple ACMHs were registered with the
same store; the error log logic was incorrect and the error was logged
when the first ACMH was registered.


artembilan:
According to commit history this must be back-ported to `5.0.x`: https://github.com/spring-projects/spring-integration/commit/d1b30466382d9c11720980475f096c3a47e6e880

PR #2653, opened 2018-12-06
Sonar Fixes

garyrussell:
Packages `o.s.i.g*`

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundGateway.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundGateway.java
@@ -41,6 +41,7 @@ import org.springframework.integration.amqp.support.DefaultAmqpHeaderMapper;
 import org.springframework.integration.gateway.MessagingGatewaySupport;
 import org.springframework.integration.support.ErrorMessageStrategy;
 import org.springframework.integration.support.ErrorMessageUtils;
+import org.springframework.messaging.MessageChannel;
 import org.springframework.retry.RecoveryCallback;
 import org.springframework.retry.support.RetrySynchronizationManager;
 import org.springframework.retry.support.RetryTemplate;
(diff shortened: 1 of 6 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundGateway.java
```diff
@@ -294,8 +294,9 @@ public void onMessage(final Message message, final Channel channel) throws Excep
 			catch (RuntimeException e) {
 				if (getErrorChannel() != null) {
 					setAttributesIfNecessary(message, null);
-					AmqpInboundGateway.this.messagingTemplate.send(getErrorChannel(), buildErrorMessage(null,
-							new ListenerExecutionFailedException("Message conversion failed", e, message)));
+					AmqpInboundGateway.this.messagingTemplate.send(getErrorChannel(), // NOSONAR not null
```
artembilan: Can't we extract `MessageChannel errorChannel = getErrorChannel();` variable instead?

To be honest I'm not a fan of this kind of `// NOSONAR` comments if we really can fix a smell.

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -716,7 +723,7 @@ protected final ErrorMessage buildErrorMessage(@Nullable Message<?> requestMessa
 	 * @return the attributes.
 	 * @since 4.3.10
 	 */
-	protected AttributeAccessor getErrorMessageAttributes(Message<?> message) {
+	protected AttributeAccessor getErrorMessageAttributes(@Nullable Message<?> message) {
```
artembilan: Maybe it's time to add a `@NonNullApi` into this package?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
garyrussell: I'd rather not add new ones until current smells are cleaned. That said, that fact that this was flagged leads me to believe it is inherited from the parent package.

PR #2661, opened 2018-12-13
RedisLock: Throw exception from unlock on expire

artembilan:
The lock might be expired in target Redis store in between `lock()` and
`unlock()`.
So, throw an `IllegalStateException` when lock is expired during
unlocking.
At the same time the lock lock is unlocked anyway.

**Cherry-pick to 5.0.x**

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -306,7 +306,8 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 			}
 			try {
 				if (!isAcquiredInThisProcess()) {
-					throw new IllegalStateException("Lock was released in the store due to expiration.");
+					throw new IllegalStateException("Lock was released in the store due to expiration. " +
+							"The integrity of data protected by this lock may have been compromised.");
 				}
 
 				if (Thread.currentThread().isInterrupted()) {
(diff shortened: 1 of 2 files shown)


spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -341,21 +341,24 @@ public void testTwoThreadsWrongOneUnlocks() throws Exception {
...
 		Lock lock1 = registry1.obtain("foo");
 		Lock lock2 = registry2.obtain("foo");
 		assertTrue(lock1.tryLock());
 		assertFalse(lock2.tryLock());
+		this.thrown.expect(IllegalStateException.class);
+		this.thrown.expectMessage("Lock was released in the store due to expiration.");
 		waitForExpire("foo");
+		lock1.unlock();
 		assertTrue(lock2.tryLock());
 		assertFalse(lock1.tryLock());
```
garyrussell: These two are redundant now - shouldn't the test for the new logic be in a different test?

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -301,6 +305,10 @@ public void unlock() {
 				return;
 			}
 			try {
+				if (!isAcquiredInThisProcess()) {
+					throw new IllegalStateException("Lock was released in the store due to expiration.");
```
garyrussell: Perhaps we should be more explicit? `...due to expiration. The integrity of data protected by this lock may have been compromised.` ?

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
Cherry-picked as 0199741d29b8dd3eb2884ab26adc8575012b284a

PR #2664, opened 2018-12-17
INT-4569: Disallow beans override in DSL

artembilan:
JIRA: https://jira.spring.io/browse/INT-4569

* Thorw `BeanDefinitionOverrideException` from the
`IntegrationFlowBeanPostProcessor` when it detects existing bean and it
is not the same object we try to register from the DSL
* Document limitations about `prototype` beans
* Some polishing in the DSL chapter of the docs

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
@@ -401,7 +401,7 @@ public class IntegrationFlowBeanPostProcessor
 					}
 				}
 				else {
-					return false;
+					return true;
 				}
 			}
 		}
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
@@ -16,7 +16,6 @@
 
 package org.springframework.integration.dsl.context;
 
-import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -34,10 +33,12 @@ import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.BeanDefinitionCustomizer;
 import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.config.EmbeddedValueResolver;
 import org.springframework.beans.factory.support.AbstractBeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.support.BeanDefinitionOverrideException;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
```diff
@@ -378,17 +379,36 @@ private void invokeBeanInitializationHooks(final String beanName, final Object b
...
+	@SuppressWarnings("unchecked")
+	private boolean noBeanPresentForComponent(Object instance, String parentBeanName) {
 		if (instance instanceof NamedComponent) {
 			String beanName = ((NamedComponent) instance).getComponentName();
 			if (beanName != null) {
-				return !this.beanFactory.containsBean(beanName);
+				if (this.beanFactory.containsBean(beanName)) {
+					BeanDefinition existingBeanDefinition = this.beanFactory.getBeanDefinition(beanName);
+					if (!ConfigurableBeanFactory.SCOPE_PROTOTYPE.equals(existingBeanDefinition.getScope())
+							&& instance != this.beanFactory.getBean(beanName)) {
```
garyrussell: If you fully intend to use `!=` vs `!instance.equals()` we need a NOSONAR.

spring-integration-core/src/test/java/org/springframework/integration/dsl/manualflow/ManualFlowTests.java
```diff
@@ -516,6 +521,17 @@ public void testConcurrentRegistration() throws InterruptedException {
...
+
+	@Test
+	public void testDisabledBeansOverride() {
+		assertThatThrownBy(() -> this.integrationFlowContext
+				.registration(f -> f.channel(c -> c.direct("doNotOverrideChannel")))
+				.register())
+		.isExactlyInstanceOf(BeanCreationException.class)
+		.hasCauseExactlyInstanceOf(BeanDefinitionOverrideException.class)
+		.hasMessageContaining("Invalid bean definition with name 'doNotOverrideChannel'");
+	}
```
garyrussell: Indentation looks wrong.

src/reference/asciidoc/dsl.adoc
```diff
@@ -115,28 +115,37 @@ The endpoints are automatically wired together by using direct channels.
...
 This will fail at runtime with a `ClassCastException` because the lambda doesn't retain the argument type and the framework will attempt to cast the payload to a `Message<?>`.
 
 Instead, use:
 
-====
 [source, java]
 ----
 .(Message.class, m -> newFooFromMessage(m))
 ----
 ====
```
garyrussell: Unbalanced fence removal?

Why remove the fences at all?

src/reference/asciidoc/dsl.adoc
```diff
@@ -115,28 +115,37 @@ The endpoints are automatically wired together by using direct channels.
...
 
+[[java-dsl-class-cast]]
+.Bean Definitions override
+[IMPORTANT]
+====
+The Java DSL can register beans for the object defined in-line in the flow definition, as well as can reuse existing, injected beans.
+In case of the same bean name defined for in-line object and existing bean definition, a `BeanDefinitionOverrideException` is thrown indicating that such a configuration is wrong.
+However when you deal with `prototype` beans, there is no way to detect from the integration flow processor an existing bean definition because every time we call a `prototype` bean from the `BeanFactory` we get a new instance.
+This way a provided instance is used in the `IntegrationFlow` as is without any bean registration and any possible check against existing `prototype` bean definition.
```
garyrussell: Add "Such a bean may not be fully initialized."

(or should we add `BF.initializeBean()` even if we don't register it?

garyrussell (changes requested):
Just a couple of comments/questions.

src/reference/asciidoc/dsl.adoc
artembilan: I removed it because they don't work in the `[IMPORTANT]` block. This piece of code has to be as a part of the one.
Right now it looks like:
default

Not perfect, but still...

garyrussell: Oh; got it; didn't notice the new fence at the top.

src/reference/asciidoc/dsl.adoc
artembilan: Please, see the latest commit. I'm not fully like it because we really do that against an existing bean and only the problem that we have an `id` and bean is `prototype`.
It doesn't apply for `singleton` because we really do have existing bean and reuse it as is since we have passed that `noBeanPresentForComponent()`.

garyrussell (approved):

PR #2667, opened 2018-12-18
Fix code smell in test and syslog modules

artembilan:


artembilan:
@garyrussell ,

Let me know if this is OK way to fix those smell from Sonar.

Yes, I admit that I do some other code polishing like exceeded line length, `this.` on methods and some other simple...

PR #2668, opened 2018-12-18
Sonar fixes

garyrussell:
Critical smells `0.s.i.r*`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMessageRouter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMessageRouter.java
@@ -17,10 +17,12 @@
 package org.springframework.integration.router;
 
 import java.util.Collection;
+import java.util.UUID;
 
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.integration.channel.NullChannel;
+import org.springframework.integration.context.IntegrationObjectSupport;
 import org.springframework.integration.core.MessagingTemplate;
 import org.springframework.integration.handler.AbstractMessageHandler;
 import org.springframework.integration.support.management.IntegrationManagedResource;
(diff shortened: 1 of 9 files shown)


spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMessageRouter.java
```diff
@@ -171,24 +173,20 @@ protected void handleMessageInternal(Message<?> message) {
...
+				if (!this.applySequence) {
+					messageToSend = message;
+				}
+				else {
+					UUID id = message.getHeaders().getId();
+					messageToSend = getMessageBuilderFactory()
 								.fromMessage(message)
-								.pushSequenceDetails(message.getHeaders().getId(), sequenceNumber++, sequenceSize)
-								.build());
+								.pushSequenceDetails(id == null ? IntegrationObjectSupport.generateId() : id,
```
artembilan: The `AbstractMessageRouter.` is an `IntegrationObjectSupport` already: no reason to use explicit method. you can just omit the class. Or Sonar won't like it still?

spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMessageRouter.java
```diff
@@ -204,4 +202,20 @@ else if (this.logger.isDebugEnabled()) {
...
+		try {
+			this.messagingTemplate.send(channel, messageToSend);
+			return true;
+		}
+		catch (MessagingException e) {
+			if (!this.ignoreSendFailures) {
+				throw e;
+			}
+			else if (this.logger.isDebugEnabled()) {
+				this.logger.debug(e);
```
artembilan: I think on reason to wrap into `isDebugEnabled()` since we don't do any logic with argument to log.

artembilan (changes requested):
A couple concerns.

PR #2670, opened 2018-12-19
Sonar fixes

garyrussell:
Critical issues `o.s.i.t*`, `o.s.i.u*`.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/transformer/ClaimCheckInTransformer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transformer/ClaimCheckInTransformer.java
@@ -16,6 +16,8 @@
 
 package org.springframework.integration.transformer;
 
+import java.util.UUID;
+
 import org.springframework.integration.store.MessageStore;
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.messaging.Message;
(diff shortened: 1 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/transformer/ClaimCheckInTransformer.java
```diff
@@ -51,10 +51,10 @@ public String getComponentType() {
 	@Override
 	protected Object doTransform(Message<?> message) throws Exception {
 		Assert.notNull(message, "message must not be null");
-		Object payload = message.getPayload();
-		Assert.notNull(payload, "payload must not be null");
+		Assert.notNull(message.getHeaders().getId(), "ID header must not be null");
 		Message<?> storedMessage = this.messageStore.addMessage(message);
-		AbstractIntegrationMessageBuilder<?> responseBuilder = this.getMessageBuilderFactory().withPayload(storedMessage.getHeaders().getId());
+		AbstractIntegrationMessageBuilder<?> responseBuilder = getMessageBuilderFactory()
+				.withPayload(storedMessage.getHeaders().getId()); // NOSONAR - null checked above
```
artembilan: Why just don't expose a local variable for both `storedMessage.getHeaders().getId()` calls? 
We won't need this `// NOSONAR` and won't spend time for more method calls.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transformer/ClaimCheckInTransformer.java
garyrussell: I did that originally, but decided it wasn't necessary to store it to a local variable since I added the assert above.

While unlikely, we can't necessarily guarantee that the returned id will be same as the source id.

artembilan: Huh? How about my point do not call the same method set again?
Plus an `id` header can't be changed in the message. And I'm not sure that we really should worry for some exotic cases when `id` is changed in between.
If it is, then claim-check is going to be broken.

Plus I would like to stay away from `// NOSONAR`  as much as possible. It makes production code dirty 🤷‍♂️

PR #2673, opened 2018-12-20
Enable ModifierOrderCheck Checkstyle rule

artembilan:
* Fix violations for `static` and `abstract` modifier
* Remove redundant code in the `TcpNioConnection`
* Mark `connectionFactoryName` as `@Nullable` in the `TcpConnectionSupport`
ctor and its inheritors
* Fix some smells according IDEA suggestions in the affected classes
* This should fix some Sonar smells as well


garyrussell:
Travis failures.

PR #2679, opened 2018-12-24
Sonar  Fixes - Javadoc-only imports

garyrussell:
Fix unused imports where the import is for Javadocs only.


artembilan (approved):
I trust you, so merging... 😄

PR #2685, opened 2019-01-09
Sonar fixes

garyrussell:
- remaining hidden fields


spring-integration-core/src/main/java/org/springframework/integration/support/management/ExponentialMovingAverageRate.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2009-2017 the original author or authors.
+ * Copyright 2009-2018 the original author or authors.
```
artembilan: 2019

artembilan (changes requested):
One wrong Copyright update

artembilan:
Also Travis failure

PR #2686, opened 2019-01-09
INT-4571 Made MessageHandlerMethodFactory injectable

olegz:
Made MessageHandlerMethodFactory injectable into MessagingMethodInvokerHelper
Deprecated HandlerMethodArgumentResolversHolder
Added 'integrationMessageHandlerMethodFactory' property to IntegrationContextUtils

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,6 +36,7 @@ import org.springframework.util.Assert;
  * @author Josh Long
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Oleg Zhurakousky
  */
 public abstract class IntegrationContextUtils {
 
@@ -104,6 +105,8 @@ public abstract class IntegrationContextUtils {
 	public static final String LIST_ARGUMENT_RESOLVERS_BEAN_NAME = "integrationListArgumentResolvers";
 
 	public static final String DISPOSABLES_BEAN_NAME = "integrationDisposableAutoCreatedBeans";
+
+	public static final String MESSAGE_HANDLER_FACTORY_BEAN_NAME = "integrationMessageHandlerMethodFactory";
 
 	/**
 	 * @param beanFactory BeanFactory for lookup, must not be null.
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -94,6 +94,7 @@ import org.springframework.messaging.handler.annotation.Header;
 import org.springframework.messaging.handler.annotation.Headers;
 import org.springframework.messaging.handler.annotation.Payload;
 import org.springframework.messaging.handler.annotation.support.DefaultMessageHandlerMethodFactory;
+import org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory;
 import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
 import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
 import org.springframework.messaging.handler.invocation.MethodArgumentResolutionException;
@@ -167,7 +168,7 @@ public class MessagingMethodInvokerHelper<T> extends AbstractExpressionEvaluator
 		SPEL_COMPILERS.put(SpelCompilerMode.MIXED, EXPRESSION_PARSER_MIXED);
 	}
 
-	private final DefaultMessageHandlerMethodFactory messageHandlerMethodFactory =
+	private MessageHandlerMethodFactory messageHandlerMethodFactory =
 			new DefaultMessageHandlerMethodFactory();
 
 	private final Object targetObject;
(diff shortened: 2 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -295,7 +296,14 @@ public void setUseSpelInvoker(boolean useSpelInvoker) {
 	@Override
 	public void setBeanFactory(@NonNull BeanFactory beanFactory) {
 		super.setBeanFactory(beanFactory);
-		this.messageHandlerMethodFactory.setBeanFactory(beanFactory);
+		if (isProvidedMessageHandlerFactoryBean()) {
+			this.messageHandlerMethodFactory = beanFactory.getBean(IntegrationContextUtils.MESSAGE_HANDLER_FACTORY_BEAN_NAME,
+					MessageHandlerMethodFactory.class);
+		}
```
garyrussell: I don't think we can do this here; the beanFactory might not have the bean yet. 

Move to initialize?

garyrussell (changes requested):
Just one issue.

@artembilan You or I can fix this on merge if you find nothing else.

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
olegz: I thought about it and was struggling as well, but i guess you're right.
In all fairness I think this class needs a bit more surgery. Look at line 262 where `messageHandlerMethodFactory` is used which means we will createInvocableHandlerMethod *always* using the default one, yet may still inject one for the actual invocation. But I stop short of trying to make a big change. Perhaps "Consider revisiting..." type issue?

garyrussell:
In future, please use imperative present tense:

**"Make MessageHandlerMethodFactory injectable"**
"Deprecate HandlerMethodArgumentResolversHolder"
"Add 'integrationMessageHandlerMethodFactory' property to IntegrationContextUtils"

See the contribution guidelines and its link to Pro Git: https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -105,6 +106,8 @@
 
 	public static final String DISPOSABLES_BEAN_NAME = "integrationDisposableAutoCreatedBeans";
 
+	public static final String MESSAGE_HANDLER_FACTORY_BEAN_NAME = "integrationMessageHandlerMethodFactory";
```
artembilan: Wait. And where is the bean for this name?
What is the point then to deprecate that `HandlerMethodArgumentResolversHolder` if we don't provide a valuable alternative out-of-the-box?

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
artembilan: Definitely better to move this `getBean()` into the `initialize()` below. This is not a "one-stop-shop" bean and it really may have a lot of dependencies, so calling it from the `setBeanFactory()` is a potential bug.

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
```diff
@@ -107,6 +111,27 @@
...
+	@Test
+	public void testHandlerMethodFactoryInjection() throws Exception {
+		SingleMethodJsonWithSpELMessageWildBean bean = new SingleMethodJsonWithSpELMessageWildBean();
+		MessagingMethodInvokerHelper<?> helper = new MessagingMethodInvokerHelper<>(bean,
+				SingleMethodJsonWithSpELMessageWildBean.class.getDeclaredMethod("foo", Message.class), false);
+		GenericApplicationContext context = new GenericApplicationContext();
+		DefaultMessageHandlerMethodFactory handlerMethodFactory = new DefaultMessageHandlerMethodFactory();
+		context.registerBean(IntegrationContextUtils.MESSAGE_HANDLER_FACTORY_BEAN_NAME,
+				MessageHandlerMethodFactory.class, () -> handlerMethodFactory);
+		context.refresh();
```
artembilan: The `context` has to be closed in the end of the test.

artembilan (changes requested):
Some feedback from me.

I can't insist, but I may merge this only after fixing `setBeanFactory()` concern and good argument about `integrationMessageHandlerMethodFactory` bean definition.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
olegz: Well, we do by allowing injection of `integrationMessageHandlerMethodFactory` (see `HandlerMethodArgumentResolversHolder` deprecation comment)

olegz: What do you mean _good argument about integrationMessageHandlerMethodFactory bean definition._?

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
olegz: I'll move it but keep in mind that it is still broken given what I said above

olegz:
Actually I am withdrawing this PR. The problem I described above (the "always" usage of the default instance of `messageHandlerMethodFactory` in the constructor) is a show stopper, since any override later on does not have any effect since the `handlerMethod` instance already exists. 

As far as the good news that whole setBeanFactory() issue is not really an issue since it forces the creation of the bean as long as its definition is in the bean factory which will always be there (XML or annotation) if defined. Also, I believe it's late enough in the cycle to not worry about forcing the creation of a bean. But in any event this may change with new PR anyway.

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
artembilan: My point is that you just don't register the bean for the `integrationMessageHandlerMethodFactory` as out-of-the-box, default solution.
This way it doesn't sound reasonable to deprecate another feature since we really don't replace it.

artembilan:
> whole setBeanFactory() issue is not really an issue 

That's fully not true.

This callback is called in the application context lifecycle pretty early and if the bean you try to get from there might have (and looks like it is) many other bean dependencies, we are going to end up with the "early bean factory access syndrome" anyway.

PR #2693, opened 2019-01-11
TcpCodecs factory improvement

garyrussell:
- add convenience factory methods to set the max message size while creating the codec

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/serializer/AbstractByteArraySerializer.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/serializer/AbstractByteArraySerializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/serializer/AbstractByteArraySerializer.java
```diff
@@ -39,7 +39,13 @@
...
 		ApplicationEventPublisherAware {
 
-	protected int maxMessageSize = 2048;
+	/**
+	 * The default maximum message size when deserializing.
+	 * @since 5.1.3
+	 */
+	public static final int DEFAULT_MAX_MESSAGE_SIZE = 2048;
+
+	protected int maxMessageSize = DEFAULT_MAX_MESSAGE_SIZE;
```
artembilan: Isn't Sonar going to object that we continue to keep it as a `protected`?
Although I guess the `private` and `protected` getter is going to be as a breaking change in the point release...

artembilan (changes requested):
Only one comment which may deserve a separate issue for the next version.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/serializer/AbstractByteArraySerializer.java
garyrussell: Doh - I meant it to be public - this explains why I couldn't get javadocs `@value` to work.

Fixing...

garyrussell: OK; will fix.

PR #2698, opened 2019-01-14
INT-4574: JDBC: Add NOT NULL for primary keys

artembilan:
JIRA: https://jira.spring.io/browse/INT-4574

Some RDBMS don't allow have nullable columns for primary keys

* Add NOT NULL for all the columns involved in primary keys

**Cherry-pick to 4.3.x**


spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-db2.sql
```diff
@@ -46,12 +46,11 @@ CREATE TABLE INT_CHANNEL_MESSAGE (
...
 	REGION VARCHAR(100) NOT NULL,
-	constraint INT_CHANNEL_MESSAGE_PK primary key (REGION, GROUP_KEY, CREATED_DATE, MESSAGE_SEQUENCE)
+	constraint INT_CHANNEL_MESSAGE_PK primary key (GROUP_KEY, MESSAGE_ID, REGION)
 );
 
-CREATE INDEX INT_CHANNEL_MSG_DELETE_IDX ON INT_CHANNEL_MESSAGE (REGION, GROUP_KEY, MESSAGE_ID);
--- This is only needed if the message group store property 'priorityEnabled' is true
--- CREATE UNIQUE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (REGION, GROUP_KEY, MESSAGE_PRIORITY DESC, CREATED_DATE, MESSAGE_SEQUENCE);
+CREATE INDEX INT_CHANNEL_MSG_DATE_IDX ON INT_CHANNEL_MESSAGE (CREATED_DATE, MESSAGE_SEQUENCE);
+CREATE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (MESSAGE_PRIORITY DESC, CREATED_DATE, MESSAGE_SEQUENCE);
```
garyrussell: Why changes to this table here, but not on master?

garyrussell (commented):
One question.

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-db2.sql
artembilan: Oh! this must not be here. The problem is with `vpp` .We Don't have it on master any more.
Will fix it later here.
We must keep your index here.

artembilan:
Restored all the previous state of scripts with the fix for VPP template.

Let me know if this doesn't back-port to 4.3.x properly.

We don't have VPP on master after this: https://github.com/spring-projects/spring-integration/commit/f970bf03321278c16e26c41bd5fd271bd5379383

garyrussell:
Cherry-pick is not clean because the channel tables are in separate files; please backport; thanks.

artembilan:
Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/7242fc8039cbbb390954018a4e35f2198d9a0357

Issue #2699, opened 2019-01-15
ConcurrentModificationException when two threads execute a method annotated with @Publisher


Cameronjmayfield:
## Bug report

Started when we upgraded to Spring Integration 5.x. Specifically 5.1.1. Example stacktrace:

```
 java.util.ConcurrentModificationException: null
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1134) ~[na:na]
	at org.springframework.integration.aop.MethodAnnotationPublisherMetadataSource.getExpressionsForHeaders(MethodAnnotationPublisherMetadataSource.java:142) ~[spring-integration-core-5.1.1.RELEASE.jar!/:5.1.1.RELEASE]
	at org.springframework.integration.aop.MessagePublishingInterceptor.evaluateHeaders(MessagePublishingInterceptor.java:195) ~[spring-integration-core-5.1.1.RELEASE.jar!/:5.1.1.RELEASE]
	at org.springframework.integration.aop.MessagePublishingInterceptor.publishMessage(MessagePublishingInterceptor.java:162) ~[spring-integration-core-5.1.1.RELEASE.jar!/:5.1.1.RELEASE]
	at org.springframework.integration.aop.MessagePublishingInterceptor.invoke(MessagePublishingInterceptor.java:144) ~[spring-integration-core-5.1.1.RELEASE.jar!/:5.1.1.RELEASE]
```

Looks like it was introduced with this optimization: https://github.com/spring-projects/spring-integration/commit/b1a3ca764c73e5da4bf10eb4b6be1116375ac227

This causes issues for our application on startup. Thoughts on using a `ConcurrentHashMap` instead of a `HashMap`?

artembilan:
@Cameronjmayfield ,

As you see we confirm that this is a problem.
Dare to contribute the fix: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc?

Thank you anyway!

Cameronjmayfield:
@artembilan Here is a pull request: https://github.com/spring-projects/spring-integration/pull/2700

PR #2700, opened 2019-01-16
[GH-2699] Make expressions cache concurrent

Cameronjmayfield:


artembilan (changes requested):
Please, add your name to the `@author` list and update Copyright over there to the current year.

Also would be great to have some test-case on the matter in the `MethodAnnotationPublisherMetadataSourceTests`.

Thank you for the contribution anyway!

Cameronjmayfield:
@artembilan I'm not sure I can reliably reproduce the concurrent modification exception, so I added a test that just checks the instance of the map. Let me know what you think.

spring-integration-core/src/test/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSourceTests.java
```diff
@@ -83,6 +84,13 @@ public void payloadAndHeaders() {
 		assertEquals("#args[2]", headerMap.get("bar").getExpressionString());
 	}
 
+	@Test
+	public void expressionsAreConcurrentHashMap() {
+		assertTrue("Expressions should be concurrent to allow startup", getField(source, "channels") instanceof ConcurrentHashMap);
```
artembilan: There is `assertThat()` and `Matchers.instanceOf()` for better failure report if that.

spring-integration-core/src/test/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSourceTests.java
```diff
@@ -16,25 +16,26 @@
...
 import org.springframework.messaging.handler.annotation.Payload;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static org.junit.Assert.*;
+import static org.springframework.test.util.ReflectionTestUtils.getField;
```
artembilan: Please, no asterisk imports. No, static imports for not allowed methods, like this `getField()`.

Also, please, try to keep the import order we support.

To be sure that you are right, please, run Gradle command: `/gradlew clean :spring-integration-core:check`.

This way you'll get tests passed and also Checkstyle rules checked.

artembilan (changes requested):
This is very close to the merge phase.
Let us know if you can fix all those concerns or we should do that for your on merge.

Thanks

spring-integration-core/src/test/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSourceTests.java
Cameronjmayfield: Great suggestion. Updated.

spring-integration-core/src/test/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSourceTests.java
Cameronjmayfield: Updated the style.

spring-integration-core/src/test/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSourceTests.java
```diff
@@ -83,6 +87,13 @@ public void payloadAndHeaders() {
 		assertEquals("#args[2]", headerMap.get("bar").getExpressionString());
 	}
 
+	@Test
+	public void expressionsAreConcurrentHashMap() {
+		assertThat("Expressions should be concurrent to allow startup", ReflectionTestUtils.getField(source, "channels"), instanceOf(ConcurrentHashMap.class));
+		assertThat("Expressions should be concurrent to allow startup", ReflectionTestUtils.getField(source, "payloadExpressions"), instanceOf(ConcurrentHashMap.class));
+		assertThat("Expressions should be concurrent to allow startup", ReflectionTestUtils.getField(source, "headersExpressions"), instanceOf(ConcurrentHashMap.class));
```
artembilan: One more code style concern and we will go for merge.
We prefer the code line length not more than `120` symbols.

I'm sharing all these rules with you to give your a chance to contribute more, but in much faster manner.
You won't see similar comments from me nay more afterwards! 😄

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSourceTests.java
Cameronjmayfield: I appreciate it. Updated.

artembilan (approved):

artembilan:

@Cameronjmayfield ,

thank you very much for contribution and your patience.
Looking forward for more!

PR #2701, opened 2019-01-16
GH-2695: Add proxy option to @EnablePublisher

### LINKED ISSUE #2695: Can configure @EnablePublisher(proxyTargetClass=true)
## Enhancement

Problem: As discussed in https://stackoverflow.com/questions/54150939/spring-cloud-stream-and-publisher-annotation-compatiblity/54151130, services using `@Publisher` annotation can't be injected using `@Autowire`

Possible solution: using an annotation like `@EnablePublisher(proxyTargetClass=true)`. Other proxy implementations of Spring like `@Cacheable` or `@Async` use this flag.

artembilan:
Fixes spring-projects/spring-integration#2695

To configure a `proxyTargetClass=true` we need declare a
`PublisherAnnotationBeanPostProcessor` bean manually, but that may cause
a confuse when `@EnablePublisher` is still present.
So, target service is proxied twice

* Expose `proxyTargetClass` and `order` into the `@EnablePublisher`
and `<enable-publisher>`
* Refactor `PublisherAnnotationBeanPostProcessor` to extend an
`AbstractBeanFactoryAwareAdvisingPostProcessor` to avoid AOP boilerplate
code altogether
* Add assertion into the `PublisherAnnotationBeanPostProcessor` to be
sure that only one of its instance is present in the application context

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aop/PublisherAnnotationBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aop/PublisherAnnotationBeanPostProcessor.java
@@ -81,7 +81,7 @@ public class PublisherAnnotationBeanPostProcessor extends AbstractBeanFactoryAwa
 		catch (NoUniqueBeanDefinitionException ex) {
 			throw new BeanCreationNotAllowedException(this.name,
 					"Only one 'PublisherAnnotationBeanPostProcessor' bean can be defined in the application context." +
-							" Consider do not use '@EnablePublisher' (or '<int:enable-publisher>') if you declare" +
+							" Do not use '@EnablePublisher' (or '<int:enable-publisher>') if you declare a" +
 							" 'PublisherAnnotationBeanPostProcessor' bean definition manually. " +
 							"Bean names found: " + ex.getBeanNamesFound());
 		}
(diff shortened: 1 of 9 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aop/PublisherAnnotationBeanPostProcessor.java
```diff
@@ -73,63 +59,31 @@ public void setDefaultChannelName(String defaultChannelName) {
...
+			this.beanFactory.getBean(PublisherAnnotationBeanPostProcessor.class);
 		}
-		else {
-			// cannot apply advisor
-			this.nonApplicableCache.add(targetClass);
-			return bean;
+		catch (NoUniqueBeanDefinitionException ex) {
+			throw new BeanCreationNotAllowedException(this.name,
+					"Only one 'PublisherAnnotationBeanPostProcessor' bean can be defined in the application context." +
+							" Consider do not use '@EnablePublisher' (or '<int:enable-publisher>') if you declare" +
```
garyrussell: `Do not use '@EnablePublisher' (or '<int:enable-publisher>') if you declare a`

src/reference/asciidoc/message-publishing.adoc
```diff
@@ -142,6 +142,8 @@ public class IntegrationConfiguration {
 ----
 ====
 
+Starting with version 5.1.3, the `<int:enable-publisher>` component, as well as `@EnablePublisher` annotation are supplied with the `proxy-target-class` and `order` attributes for tuning a `ProxyFactory` configuration.
```
garyrussell: as well as **the** `@EnablePublisher` ...

`s/are supplied with the/have the/`

for tuning **the** `ProxyFactory`

garyrussell (changes requested):
Just a couple of minor polishings...

garyrussell (approved):

PR #2707, opened 2019-01-18
Fix DSL to deal with beanNames for handlers

artembilan:
When consumer endpoint is created by the Framework, the target
`MessageHandler` gets a `componentName` from the `ConsumerEndpointFactoryBean`.
Therefore we can't rely on the `getComponentName()` when we create beans
from Java DSL.

* Introduce `NamedComponent.getBeanName()` contract; make it default
to the `getComponentName()`; implement this method from the
`IntegrationObjectSupport`
* Use this new `getBeanName()` from the `IntegrationFlowBeanPostProcessor`
and `StandardIntegrationFlowContext` for better existing beans checks
and possible reuse existing `MessageHandler` in different endpoints
* Optimize `FixedSubscriberChannel` and implement `getBeanName()` over there
* Implement `getBeanName()` in the `AbstractMessageSource`; use
`ExpressionEvalMap` to evaluate headers expressions
* Implement `getBeanName()` in `Trackable*Metrics` classes as delegation
to the `this.trackable`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/filter/MessageFilter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/filter/MessageFilter.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.filter;
 
+import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.context.Lifecycle;
 import org.springframework.core.convert.ConversionService;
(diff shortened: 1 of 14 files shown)


spring-integration-core/src/main/java/org/springframework/integration/filter/MessageFilter.java
```diff
@@ -110,12 +111,8 @@ public void setDiscardWithinAdvice(boolean discardWithinAdvice) {
 	@Override
 	public MessageChannel getDiscardChannel() {
 		if (this.discardChannelName != null) {
-			synchronized (this) {
-				if (this.discardChannelName != null) {
-					this.discardChannel = getChannelResolver().resolveDestination(this.discardChannelName);
-					this.discardChannelName = null;
-				}
-			}
+			this.discardChannel = getChannelResolver().resolveDestination(this.discardChannelName);
```
garyrussell: Possible NPE here when 2 concurrent threads.

garyrussell (changes requested):

garyrussell (approved):

PR #2709, opened 2019-01-22
* Fix new Sonar issue in the `AbstractMailReceiver`

artembilan:


garyrussell:
This is still a problem

```java
		finally {
			MailTransportUtils.closeFolder(folder, this.shouldDeleteMessages);
			this.folderReadLock.unlock();
		}
```

We won't unlock if the close() fails.

PR #2710, opened 2019-01-22
Implement Lifecycle for IntegrationComponentSpec

artembilan:
Since an `IntegrationComponentSpec` is a `FactoryBean`, all it's target
callbacks and lifecycle is controlled over a `FactoryBean`.

* Add `SmartLifecycle` for the `IntegrationComponentSpec` to delegate
lifecycle hooks to the `target` if necessary
* Refactor `IntegrationComponentSpec` to be an `AbstractFactoryBean`
which is a central place for the `FactoryBean`, `InitializingBean` and
`DisposableBean` interfaces

**Cherry-pick to 5.0.x**

See https://stackoverflow.com/questions/54314755/enpointspec-fails-in-spring-integration-java-dsl

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationComponentSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationComponentSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 the original author or authors.
+ * Copyright 2016-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


garyrussell:
I am sure this will work but Specs are just scaffolding for creating integration components; it really doesn't make much sense to register them as beans.

But I guess there's nothing we can do to stop that, so this is probably the only way to go, but I wanted to air my thoughts before merging in case you have anything to add.

artembilan:
Well, we really can't enforce people do not register them as beans. More over it is there for a while already.
And that is not a way to go in Spring where we definitely connect everything with each other as beans.
Plus this feature is left behind the scene from end user and it becomes just Framework burden to support the proper way of processing.

We may reconsider though in the next version if you feel strong...

garyrussell:
No; it's fine; just wanted to air the issue. Cherry-picked as b4a2de12aa193c870b48d13d8f2f361ab6d21701

PR #2712, opened 2019-01-23
INT-2549: Ignore MBean call reply in op-invoc-c-a

artembilan:
JIRA: https://jira.spring.io/browse/INT-2549

* Add `expectReply` property into the `OperationInvokingMessageHandler`
to align the one-way and request-reply behavior with all other similar
components in Spring Integration
* Ignore an operation invocation result in case of `expectReply == false`
and log warning
* Provide some refactoring into the `OperationInvokingMessageHandler`
to fix Sonar complains about complexity

**Cherry-pick to 5.0.x, 4.3.x**

We may consider this as a breaking change, but my point is to align with similar behavior in may other components, e.g. `FileWritingMessageHandler` or `AmqpOutboundEndpoint`.
Will document change after review.

### DIFF
--- a/spring-integration-jmx/src/main/java/org/springframework/integration/jmx/OperationInvokingMessageHandler.java
+++ b/spring-integration-jmx/src/main/java/org/springframework/integration/jmx/OperationInvokingMessageHandler.java
@@ -59,22 +59,49 @@ import org.springframework.util.ObjectUtils;
  * @author Mark Fisher
  * @author Oleg Zhurakousky
  * @author Gary Russell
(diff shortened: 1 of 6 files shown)


artembilan:
Cherry-picked to `5.0.x` as https://github.com/spring-projects/spring-integration/commit/c723b69f01a61f2e9e6cbe73086e62ecd4cf1d4c and to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/54f97f113cb1f4647406d6e1d632eb28713aab99 after fixing conflicts.

PR #2718, opened 2019-01-24
Some file module Sonar fixes

artembilan:

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractRegexPatternFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractRegexPatternFileListFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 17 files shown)


garyrussell:
There are travis failures.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractRegexPatternFileListFilter.java
```diff
@@ -25,37 +25,46 @@
...
-		implements InitializingBean {
+		implements InitializingBean { // TODO Remove in the next version
 
-	private volatile Pattern pattern;
+	private Pattern pattern;
 
 
 	public AbstractRegexPatternFileListFilter(String pattern) {
-		this.pattern = Pattern.compile(pattern);
+		this(Pattern.compile(pattern));
```
garyrussell: NPE if pattern==null

spring-integration-file/src/main/java/org/springframework/integration/file/filters/LastModifiedFileListFilter.java
```diff
@@ -65,13 +72,22 @@ public LastModifiedFileListFilter(long age) {
...
+	/**
+	 * Set the age that files have to be before being passed by this filter.
+	 * If {@link File#lastModified()} plus age is greater than the current time, the file
+	 * is filtered. The resolution is seconds.
+	 * Defaults to 60 seconds.
+	 * @param age the age
+	 * @since 5.1.3
+	 */
+	public void setAge(Duration age) {
+		setAge(age.get(ChronoUnit.SECONDS));
```
garyrussell: Why not just `age.getSeconds()` ? Simpler, and functionally the same.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
```diff
@@ -618,14 +611,6 @@ private StreamHolder(InputStream stream, String name) {
...
 
-		public InputStream getStream() {
-			return this.stream;
-		}
-
-		public String getName() {
-			return this.name;
-		}
-
 	}
```
garyrussell: `StreamHolder` ctor must be protected to avoid synthetic ctor.

garyrussell (changes requested):
Good work! - just a few minor comments.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/LastModifiedFileListFilter.java
artembilan: Because it is not `public` 😄

garyrussell: `toSeconds()` is private, but...

```java
    public long getSeconds() {
        return seconds;
    }
```

artembilan:
I think that's enough for the current round.
We have a lot of hard logic in those classes. It's almost impossible to simplify complexity over there.

Another point is to work out a rule for ourselves to check Sonar report when we touch some classes and fix smells together with the original reported issue.
That's really too much mind effort to come with the proper refactoring, although I should admit this is really useful task to do anyway! We learn something for good habit, plus in many cases we really fix leaks 😄 

I will push fixes according your review when you done.

Thanks

garyrussell (approved):

PR #2721, opened 2019-01-28
Fix mail lock race condition & Sonar smells

artembilan:
https://build.spring.io/browse/INT-MASTERSPRING40-599/

* The `folderReadLock` might not be re-locked when `openFolder()`
throws an exception
* Fix all the Sonar smells for mail module
* Optimize all the dynamic `Assert` messages to `Supplier`
* Refactor `MailReceiverFactoryBean` to be based on the `AbstractFactoryBean`

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapMailReceiver.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapMailReceiver.java
@@ -253,10 +253,15 @@ public class ImapMailReceiver extends AbstractMailReceiver {
 
 		@Override
 		public void run() {
-			Folder folder = getFolder();
-			logger.debug("Canceling IDLE");
-			if (folder != null) {
-				folder.isOpen(); // resets idle state
+			try {
+				Folder folder = getFolder();
+				logger.debug("Canceling IDLE");
+				if (folder != null) {
+					folder.isOpen(); // resets idle state
+				}
+			}
+			catch (Exception ex) {
+				logger.error("Error during resetting idle state.", ex);
 			}
 		}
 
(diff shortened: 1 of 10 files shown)


spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapMailReceiver.java
```diff
@@ -250,16 +253,13 @@ protected void setAdditionalFlags(Message message) throws MessagingException {
...
-					folder.isOpen(); // resets idle state
-				}
-			}
-			catch (Exception ignore) {
+			Folder folder = getFolder();
+			logger.debug("Canceling IDLE");
+			if (folder != null) {
+				folder.isOpen(); // resets idle state
 			}
 		}
```
garyrussell: I presume this was to avoid log noise. Why change? If your intention is to alert the user of the problem, it would be better to log rather than the task executor printing the stack trace.

spring-integration-mail/src/main/java/org/springframework/integration/mail/MailTransportUtils.java
```diff
@@ -51,26 +52,25 @@ public static void closeService(Service service) {
...
+	//    /**
+	//     * Close the given JavaMail Folder and ignore any thrown exception. This is
+	//     * useful for typical <code>finally</code> blocks in manual JavaMail code.
+	//     *
+	//     * @param folder the JavaMail Folder to close (may be <code>null</code>)
+	//     */
+	//
+	//    public static void closeFolder(Folder folder) {
+	//        closeFolder(folder, false);
+	//    }
```
garyrussell: Why not just delete this?

garyrussell (changes requested):

garyrussell (approved):

PR #2724, opened 2019-01-29
GH-2723: Handle unsupported XML properties

### LINKED ISSUE #2723: Unable to create XsltPayloadTransformer when Saxon HE is in the classpath.
**Affects Version(s):** 4.3.19.RELEASE

---

After upgrading to spring-integration-xml:4.3.19.RELEASE (from 4.3.18.RELEASE), my application failed  to start with the following exception :

```
Error creating bean with name 'XsltPayloadTransformer': Invocation of init method failed;

...

Caused by: java.lang.IllegalArgumentException: Unknown configuration property http://javax.xml.XMLConstants/property/accessExternalDTD
```

The cause of the exception seems to be related with the following changes 

```
- transformerFactory = TransformerFactory.newInstance();
+ transformerFactory = TransformerFactoryUtils.newInstance();
```

The TransformerFactoryUtils.newInstance() force the XMLConstants.ACCESS_EXTERNAL_DTD property which doesn't seems to be supported by Saxon. If I quote answers from Michael Kay about similar subject, I understand that Saxon does not handle Xml Parser related properties.

> Saxon doesn't have control of the XML parsing: we invoke the parser that the application supplies and configures.

As the problem is related to a third party dependency, I provide a [side sample project](https://github.com/stoussaint/sixml-saxon-regression) as a regression test case

artembilan:
Fixes spring-projects/spring-integration#2723

Not all XML components support all the configuration properties.
For example Saxon HE doesn't support `XMLConstants.ACCESS_EXTERNAL_DTD`
and end up with an exception like:
`IllegalArgumentException: Unknown configuration property http://javax.xml.XMLConstants/property/accessExternalDTD`

* Change `XsltPayloadTransformer` to re-use `TransformerFactoryUtils`
from spring-ws as a centralized source of `TransformerFactory`
configuration.
* Wrap `XMLConstants.ACCESS_EXTERNAL_STYLESHEET` to the `try..catch`
and log INFO about not supported property

**Cherry-pick to 5.0.x & 4.3.x**

### DIFF
--- a/spring-integration-xml/src/main/java/org/springframework/integration/xml/transformer/XsltPayloadTransformer.java
+++ b/spring-integration-xml/src/main/java/org/springframework/integration/xml/transformer/XsltPayloadTransformer.java
@@ -49,6 +49,7 @@ import org.springframework.integration.xml.source.SourceFactory;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.PatternMatchUtils;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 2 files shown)


garyrussell:
Cherry-picked to 5.0.x as 5e6506abafe80f7c51de0c5963c6174e0f5cfaad and 4.3.x as 72aaf0b03f013c85b4b2c6261f5377ca92c86b23 after resolving conflicts.

PR #2725, opened 2019-01-30
Make IOS.setTaskScheduler() as public

artembilan:
* Make `IntegrationObjectSupport.setTaskScheduler()` as `public` and
remove all the overrides for visibility
* Fix Sonar smells for all the affected classes

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
@@ -119,7 +119,7 @@ public abstract class AbstractEndpoint extends IntegrationObjectSupport
 	}
 
 	@Override
-	public void destroy() {
+	public void destroy() throws Exception { // NOSONAR TODO: remove throws in 5.2
 		if (this.roleController != null) {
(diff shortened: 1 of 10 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
```diff
@@ -125,7 +119,7 @@ protected void onInit() {
 	}
 
 	@Override
-	public void destroy() throws Exception {
+	public void destroy() {
```
garyrussell: We can't do this - breaking change for anyone that has subclassed and implemented `destroy()`.

I suggest we use `// NOSONAR TODO: remove throws in 5.2`, which removing it from our subclasses.

artembilan: OK. 
Do we remove from our subclasses now as I did in this PR?

Sorry, if I misunderstood you...

artembilan:
Is the latest commit OK?

garyrussell:
LGTM; let's see if Mr. Travis agrees.

artembilan:
OK. It's green finally! 😄

PR #2730, opened 2019-02-01
GH-2727: Ensure JDBC queries are logged

### LINKED ISSUE #2727: Add an ability to log SQL queries executing in JdbcTemplate class provided by JdbcPollingChannelAdapter class when debug is enabled.
**Affects Version(s):** 5.1.2.RELEASE

There is a log message written in `JdbcTemplate` starting with "Executing prepared SQL statement", but it remains not continued with actual SQL query being executed due to the fact that JdbcPollingChannelAdapter does not implement `SqlProvider` interface. 

Are there any significant reasons not to provide this functionality which leads to absence of such an important debug information?

---

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/2727

The `JdbcTemplate` logs a message for the sql to execute when it is
supplied by the `QueryProvider`.

* Refactor `JdbcPollingChannelAdapter` to use new introduced internal
`PreparedStatementCreatorWithMaxRows` with the `QueryProvider`
* Refactor `JdbcMessageHandler` to instantiate a `generatedKeysStatementCreator`
based on the `PreparedStatementCreatorFactory`
* Verify in the `JdbcOutboundGatewayParserTests` that both fixes logs
sql queries properly

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
@@ -29,8 +29,10 @@ import org.springframework.integration.endpoint.AbstractMessageSource;
 import org.springframework.jdbc.core.ColumnMapRowMapper;
 import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.ParameterDisposer;
 import org.springframework.jdbc.core.PreparedStatementCreator;
 import org.springframework.jdbc.core.PreparedStatementCreatorFactory;
+import org.springframework.jdbc.core.PreparedStatementSetter;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.SqlProvider;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcOperations;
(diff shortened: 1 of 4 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
```diff
@@ -200,8 +205,32 @@ protected Object doReceive() {
 	}
 
 	private void executeUpdateQuery(Object obj) {
-		SqlParameterSource updateParameterSource = this.sqlParameterSourceFactory.createParameterSource(obj);
-		this.jdbcOperations.update(this.updateSql, updateParameterSource);
+		this.jdbcOperations.update(this.updateSql, this.sqlParameterSourceFactory.createParameterSource(obj));
+	}
+
+	private static final class PreparedStatementCreatorWithMaxRows implements PreparedStatementCreator, SqlProvider {
```
garyrussell: Shouldn't we implement all the delgate's interfaces?

```java
	private class PreparedStatementCreatorImpl
			implements PreparedStatementCreator, PreparedStatementSetter, SqlProvider, ParameterDisposer {
```

I see code like this in the template...

```java
			finally {
				if (pss instanceof ParameterDisposer) {
					((ParameterDisposer) pss).cleanupParameters();
				}
			}
```

garyrussell (changes requested):

garyrussell (approved):

PR #2737, opened 2019-02-05
GH-2735: Add errorChannel to ScatterGatherHandler

### LINKED ISSUE #2735: Spring Integraton- Issue with Nested Scatter Gather Errors not Propagated
**Affects Version(s):** \<Spring Integration version 5.1.1>

---

We are facing issues with Spring Integration, when nested Scatter Gather is used, success Response is getting propagated to the caller. Error Response is not being sent back to the caller.

Here is the reference project that I am using to replicate the issue.
[ScatterGatherTest.zip](https://github.com/spring-projects/spring-integration/files/2832563/ScatterGatherTest.zip)

This issue is similar to the issue reported here.https://github.com/spring-projects/spring-integration/issues/2731

Thanks for your support in advance.

artembilan:
Fixes spring-projects/spring-integration#2735

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/ScatterGatherSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/ScatterGatherSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.dsl;
 
+import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.scattergather.ScatterGatherHandler;
 import org.springframework.messaging.MessageChannel;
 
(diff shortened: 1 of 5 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/ScatterGatherSpec.java
```diff
@@ -58,4 +59,16 @@ public ScatterGatherSpec gatherTimeout(long gatherTimeout) {
 		return this;
 	}
 
+	/**
+	 * Specify a {@link MessageChannel} bean name for async errors processing.
```
garyrussell: `s/errors/error/` (also in param below and other javadocs).

garyrussell (changes requested):
That's all; but I am not sure I understand why the exception is an entry in the list.

Looking forward to docs.

artembilan:
>  but I am not sure I understand why the exception is an entry in the list.

Sorry, have missed your review somehow...

An exception is an item of the aggregated message because that is what my logic in the error processor does: return `MessageBuilder.withPayload(payload.getCause().getCause())`. See `@ServiceActivator(inputChannel = "scatterGatherErrorChannel")`  in the `RouterTests`. Nothing more. 
The point of the centralized `errorChannel` just because we have several requests from the single scaterrer point. So, it would be better to let all downstream errors to be settled in a single place for centralized error handling.
Of course, you still can have your own error handling in the particular sub-flow of the distribution/auction, but that is a different story, although it might finish with some compensation anyway.

Will fix your concern and explain this behavior in Docs.

Thank you!

artembilan:
Ready for the final review.

Thanks

garyrussell:
Minor doc polishing for your review: https://github.com/garyrussell/spring-integration/commit/9de4791cbf6d131f65f5600a454a6c900c849544

artembilan:
> Minor doc polishing

👍 Thank you!

garyrussell:
Merged as e8aa8618dfbaa155419500df9d1d374a1da1f9a7

PR #2738, opened 2019-02-06
Add default expressions for ExpEvalReqHAdvice

artembilan:
https://stackoverflow.com/questions/54546728/how-to-handle-ftpoutboundadapter-connection-exception-in-spring-integration-flow

When channels are configured for the `ExpressionEvaluatingRequestHandlerAdvice`,
but no expressions, the logic is not performed.
In other words: we can evaluate expressions, when no channels,
but we  don't send messages to channels when no expressions.

* Provide default expressions to be evaluated to `payload`, when only
channels are provided.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ExpressionEvaluatingRequestHandlerAdvice.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ExpressionEvaluatingRequestHandlerAdvice.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ExpressionEvaluatingRequestHandlerAdvice.java
```diff
@@ -36,53 +40,62 @@
...
 
 	/**
 	 * Set the expression to evaluate against the message after a successful
 	 * handler invocation.
+	 * Defaults to {@code payload}, if {@code successChannel} is configured.
 	 * @param onSuccessExpression the SpEL expression.
 	 * @since 5.0
 	 */
 	public void setOnSuccessExpression(Expression onSuccessExpression) {
+		Assert.notNull(onSuccessExpression, "'onSuccessExpression' must not be null");
```
garyrussell: Breaking change; someone might have...

```java
    advice.setFailureExpression(...);
    advice.setSuccessExpression(null);
```

Just for self-documentation purposes.

Maybe 5.2 only? Since the work around is simple?

However, why are these asserts even needed? You retained the null checks in `doInvoke()`.

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -406,6 +406,8 @@ An additional property, called `inputMessage`, contains the original message sen
 A message sent to the `failureChannel` (when the handler throws an exception) is an `ErrorMessage` with a payload of `MessageHandlingExpressionEvaluatingAdviceException`.
 Like all `MessagingException` instances, this payload has `failedMessage` and `cause` properties, as well as an additional property called `evaluationResult`, which contains the result of the expression evaluation.
 
+NOTE: Starting with version 5.1.3, if channels are configured, but expressions are not provided, the default expression is used to evaluate to a `payload` of the message.
+
```
garyrussell: to **the** payload if the message.

garyrussell (changes requested):

artembilan:
Pushed according your advises. I think with the current state it is not a breaking change because previously the channels just haven't had any value. They've been ignored.

garyrussell:
I agree, but it's not necessary to assert not null here; and users might have done it, just for completeness; no need to break such apps.

garyrussell (approved):

PR #2745, opened 2019-02-13
Add `updateCopyrights` Gradle task

artembilan:
To avoid extra manual task to check all the files for actual Copyright
header an `updateCopyrights` is introduced to be performed before
`processResources` and check modified classes via `Git status` command
by the `grgit` plugin.
The `updateCopyrights` task is enabled only locally: when it's not on
Travis or Bamboo - no reason to infer code base when it is not going
to be committed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ buildscript {
 plugins {
 	id 'org.sonarqube' version '2.6.2'
 	id 'org.asciidoctor.convert' version '1.5.9.2'
+	id 'org.ajoberstar.grgit' version '3.0.0'
 }
 
 description = 'Spring Integration'
@@ -25,6 +26,9 @@ ext {
 	linkScmConnection = 'scm:git:git://github.com/spring-projects/spring-integration.git'
 	linkScmDevConnection = 'scm:git:ssh://git@github.com:spring-projects/spring-integration.git'
 	docResourcesVersion = '0.1.0.RELEASE'
+
+	modifiedFiles =
+			files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') || f.name.endsWith('.kt') }
 }
 
 allprojects {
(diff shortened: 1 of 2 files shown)


artembilan:
@garyrussell ,

Please, check locally if this works for your.
Looks like I broke my local Git repository somehow and it doesn't see changes in classes, only `build.gradle` 😢 
Works in other projects though...

garyrussell:
Seems to work ok for me...

```
$ ./gradlew updateCopyrights
...
$ git diff
diff --git a/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java b/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
index 30fad35e20..af33d65b91 100644
--- a/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
+++ b/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2018 the original author or authors.
+ * Copyright 2014-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -66,6 +66,8 @@ public class RedisLockRegistryTests extends RedisAvailableTests {
 
        private final String registryKey2 = UUID.randomUUID().toString();
 
+       int foo;
+
```

artembilan:
Grate! So, let's merge it then! 😄 

Pay attention that it is going to work whenever we run compile for the project or any module.
Although the change is going to be applied to that module and its dependencies only...
It is not going to be applied on CI.
Plus I made a `checkTestConfigs` as incremental to avoid its "hard" work all the time.
Not a big deal, but still good practice for Gradle scripts.

garyrussell:
Merged as daf00154c6b854b6b426eaf5f6108c83719cfaf7

PR #2751, opened 2019-02-14
GH-2749: Deprecate ChannelInterceptorAware

### LINKED ISSUE #2749: Retire ChannelInterceptorAware in favor of org.springframework.messaging.support.InterceptableChannel
Consider to deprecate now with an extension of the `InterceptableChannel` and remove in the next version.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2749

The `org.springframework.messaging.support.InterceptableChannel` provides
exact functionality as `ChannelInterceptorAware`

* Make `ChannelInterceptorAware extends InterceptableChannel`
* Suppress deprecation warning whenever we need to keep backward
compatibility
* Fix all other places to deal with `InterceptableChannel` already


artembilan:
That's just cool to have that `updateCopyrights` task. It saved me a lot of time! 😄

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -243,7 +244,7 @@ public void setMessageConverter(MessageConverter messageConverter) {
 	 * Return a read-only list of the configured interceptors.
 	 */
 	@Override
-	public List<ChannelInterceptor> getChannelInterceptors() {
+	public List<ChannelInterceptor> getInterceptors() {
```
garyrussell: Can't do this; breaking change. Need to deprecate the old method.

src/reference/asciidoc/channel.adoc
```diff
@@ -304,6 +304,7 @@ Also, starting with version 5.1, `ChannelInterceptor.postReceive()` is no longer
 Previously, the method was called.
 If you have an interceptor that relies on the previous behavior, implement `afterReceiveCompleted()` instead, since that method is invoked, regardless of whether a message is received or not.
 
+NOTE: Starting with version 5.1.3, the `ChannelInterceptorAware` is deprecated in favor of `InterceptableChannel` from Spring Messaging module, which it extends now for backward compatibility.
```
garyrussell: from **the** Spring Messaging module ... which it **now extends**

garyrussell (changes requested):
2 issues; one major, one minor.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
artembilan: Unfortunately it clashes with the `InterceptableChannel` API. So, we in chicken-or-egg situation anyway, independently of the version we are going to fix it. Good deal that this method is just `protected`, so visible only for implementors, but not regular end-user code

garyrussell: I've moved it to 5.1.4; let's talk when we have more time.

artembilan:
OK. I think its time to forget such a deprecation for `5.1.x` and just apply the change for the current `5.2`.

I'll add a `Migration Guide` note, when the PR is merged.

Thanks

src/reference/asciidoc/channel.adoc
garyrussell: `s/5.1.3/5.2/`

from **the** Spring Messaging module

garyrussell (changes requested):
Minor nit picks

artembilan:
Added Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.1-to-5.2-Migration-Guide#deprecated-channelinterceptoraware

PR #2758, opened 2019-02-21
Update docs for message routers so that they compile

mamachanko:
See compiling example [here](https://github.com/mamachanko/routetorecipient-docs-update/blob/master/src/main/java/io/github/mamachanko/routetorecipientdocsupdate/RoutetorecipientDocsUpdateApplication.java).


artembilan:
@mamachanko ,

👍 

Issue #2759, opened 2019-02-21
AMQP CorrelationDataWrapper Needs to Delegate New Methods


garyrussell:
**Affects Version(s):**  all

The `AmqpOutboundEndpoint` wraps the user `CorrelationData` in a `CorrelationDataWrapper`.

This wrapper must delegate `getFuture()` and `setReturnedMessage()` methods to the delegate.

Needs a change in Spring AMQP to make `setReturnedMessage()` public (or use reflection).

See https://stackoverflow.com/questions/54713121/spring-integration-publisher-confirms-with-timeout/54799635#comment96403650_54799635

garyrussell:
@artembilan Instead of making `setReturnedMessage()` public, I propose making it protected and with the following...

```java
	protected static final class CorrelationDataWrapper extends CorrelationData {

		private static final Method setReturnedMessage;

		static {
			AtomicReference<Method> method = new AtomicReference<>();
			ReflectionUtils.doWithMethods(CorrelationData.class, m -> {
				m.setAccessible(true);
				method.set(m);
			}, m -> {
				return m.getName().equals("setReturnedMessage");
			});
			setReturnedMessage = method.get();
		}

		private final Object userData;

		private final Message<?> message;

		CorrelationDataWrapper(String id, Object userData, Message<?> message) {
			super(id);
			this.userData = userData;
			this.message = message;
		}

		public Object getUserData() {
			return this.userData;
		}

		public Message<?> getMessage() {
			return this.message;
		}

		@Override
		public SettableListenableFuture<Confirm> getFuture() {
			if (userData instanceof CorrelationData) {
				return ((CorrelationData) userData).getFuture();
			}
			else {
				return super.getFuture();
			}
		}

		@Override
		public org.springframework.amqp.core.Message getReturnedMessage() {
			return super.getReturnedMessage();
		}

		@Override
		public void setReturnedMessage(org.springframework.amqp.core.Message returnedMessage) {
			if (this.userData instanceof CorrelationData && setReturnedMessage != null) {
				try {
					setReturnedMessage.invoke(this.userData, returnedMessage);
				}
				catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
					// empty
				}
			}
			super.setReturnedMessage(returnedMessage);
		}
	}
```

If you agree, I'll push the change to AMQP and issue a PR for this.

artembilan:
[...]

PR #2761, opened 2019-02-21
GH-2759: Fix CorrelationData.future

garyrussell:
Fixes https://github.com/spring-projects/spring-integration/issues/2759

The outbound endpoints wrap user correlation data in a wrapper.
If the user data is a `CorrelationData`, we must delegate methods
involving the `Future<?>` and `returnedMessage` to the user data.

**cherry-pick to 2.1 and switch AMQP to snapshots**

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
@@ -44,6 +44,7 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
+import org.springframework.util.concurrent.SettableListenableFuture;
 
 /**
  * @author Gary Russell
@@ -479,8 +480,10 @@ public abstract class AbstractAmqpOutboundEndpoint extends AbstractReplyProducin
 			if (messageId == null) {
 				messageId = NO_ID;
 			}
-			correlationData = new CorrelationDataWrapper(messageId.toString(),
-					this.correlationDataGenerator.processMessage(requestMessage), requestMessage);
+			Object userData = this.correlationDataGenerator.processMessage(requestMessage);
+			if (userData != null) {
+				correlationData = new CorrelationDataWrapper(messageId.toString(), userData, requestMessage);
+			}
 		}
 		return correlationData;
 	}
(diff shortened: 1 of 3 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
```diff
@@ -604,6 +607,28 @@ public Object getUserData() {
...
+				return ((CorrelationData) this.userData).getFuture();
+			}
+			else {
+				return super.getFuture();
+			}
+		}
+
+		@Override
+		public org.springframework.amqp.core.Message getReturnedMessage() {
+			return super.getReturnedMessage();
```
artembilan: Looks like you are missing here something: since you delegate to the `userData` in the `setReturnedMessage()`, I guess you need to do the same here as well.
Otherwise this override just doesn't make sense.

artembilan (changes requested):
Just one concern, but I think it is critical...

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
garyrussell: No, in the setter I call super too; but you are correct, this should be removed.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
```diff
@@ -479,8 +480,10 @@ protected CorrelationData generateCorrelationData(Message<?> requestMessage) {
 			if (messageId == null) {
 				messageId = NO_ID;
 			}
-			correlationData = new CorrelationDataWrapper(messageId.toString(),
-					this.correlationDataGenerator.processMessage(requestMessage), requestMessage);
+			Object userData = this.correlationDataGenerator.processMessage(requestMessage);
+			if (userData != null) {
+				correlationData = new CorrelationDataWrapper(messageId.toString(), userData, requestMessage);
```
artembilan: Looks like previously we have had a `CorrelationDataWrapper` instance independently of the `userData`, but now you restrict it only if that one is not null.

Also I see that you are protected against `null` in the `CorrelationDataWrapper` any way.

So, what is the motivation to not do that any more?

artembilan (changes requested):

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AbstractAmqpOutboundEndpoint.java
garyrussell: Because we get an NPE in `handleConfirm()` if the expression resolves to `null` (bug - try to create a message with a `null` payload).

I should have mentioned it in the commit comment and, I suppose, this part should be back-ported.

artembilan: Should we emit WARN  then in case of `correlationDataGenerator` resolves to `null`?

garyrussell: I decided to go with DEBUG (there is a corresponding DEBUG in `handleConfirm`).

Someone might want to use this as a feature ("I don't care about a confirm for this message").

artembilan:
Cherry-picked to `5.1.x` as https://github.com/spring-projects/spring-integration/commit/cb1b1c81f1b8bb1c09b39d19b14c2a463c8dbbdd after fixing conflicts.

Do we need to back-port further down to `4.3.x`? At least some part of your change, e.g. that NPE protection

garyrussell:
Yes, but just the NPE protection - the future was added in S-A 2.1.

I can create a new PR if you prefer.

artembilan:
I'm OK to back-port that part.

Thanks for confirmation!

artembilan:
Back-ported NPE fix to `5.0.x` as https://github.com/spring-projects/spring-integration/commit/8f1a1c1777c983e5ef56ef945eea32cfb082b242 and cherry-picked to `4.3.x`

PR #2764, opened 2019-02-22
INT-4559: JDBC Gateway: return List for maxRows>1

artembilan:
JIRA: https://jira.spring.io/browse/INT-4559

* Return a single item from result only if
`maxRows == null || maxRows == 1`
* Remove deprecated `max-rows-per-poll` from the
`JdbcOutboundGateway.java` and its XSD and parsers


garyrussell:
We'll need a migration guide entry for this.

PR #2766, opened 2019-02-26
GH-2753: Remove Guava dependency

### LINKED ISSUE #2753: Consider to remove Guava dependency used only as optional in the StoredProcExecutor for caching
Looks like the provided caching capabilities (including stats) are just an overhead for such an internal logic of the Framework.
A simple `LinkedHashMap` with plain eviction policy is fully enough to achieve a functionality of the `StoredProcExecutor` component.
And that is used already by default.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2753

* Remove Guava dependency and its minor functionality from the
`StoredProcExecutor`
* Remove `@ManagedResource` and its operations/attributes from
`StoredProcExecutor` since they are not relevant any more
* Remove tests related to JMX and Guava
* Refactor all other tests in the affected classes
* Some code polishing in the `StoredProcExecutor`


artembilan:
I will add Migration Guide note if we agree with this change.

Thanks

PR #2768, opened 2019-02-27
 Fix new Sonar smells

artembilan:
* Fix NPE around `MetricsCaptor` in channels
* Fix potential leaks with `MeterFacade`  in `PollableChannel` implementations
* Some refactoring for `MessagingMethodInvokerHelper`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
@@ -184,9 +184,11 @@ public final class IntegrationUtils {
 		RuntimeException runtimeException = (ex instanceof RuntimeException)
 				? (RuntimeException) ex
 				: new MessageHandlingException(message, text.get(), ex);
+
 		if (!(ex instanceof MessagingException) ||
 				((MessagingException) ex).getFailedMessage() == null) {
-			runtimeException = new MessageHandlingException(message, text.get(), ex);
+			runtimeException = new MessageHandlingException(message, text.get(),
+					(ex instanceof IllegalStateException && ex.getCause() != null) ? ex.getCause() : ex);
 		}
 		return runtimeException;
 	}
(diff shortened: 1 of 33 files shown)


artembilan:
Gary, pay attention, please, what I've done for the `MessagingMethodInvokerHelper`, especially exception handling.
And with that fix I had to do something for the `IntegrationUtils.wrapInHandlingExceptionIfNecessary()` to unwrap an `IllegalStateException`.

Let me know if that is good!

Also my opinion: I don't like how Sonar react to our `catch()` and re-throws or `if...else` in `catch` blocks. Looks like it doesn't see the whole picture and doesn't suggest to some proper solution.
I realized that we can trick it with extra methods where we do the "hard" logic to process an exception 😄 

Thanks

artembilan:
Plus forgot to say: we don't remove `MeterFacade` in some places, e.g. `PollableAmqpChannel`.
I will issue a separate PR for `5.1` to back-port that.

spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
```diff
@@ -184,9 +184,11 @@ public static RuntimeException wrapInHandlingExceptionIfNecessary(Message<?> mes
 		RuntimeException runtimeException = (ex instanceof RuntimeException)
 				? (RuntimeException) ex
 				: new MessageHandlingException(message, text.get(), ex);
+
 		if (!(ex instanceof MessagingException) ||
 				((MessagingException) ex).getFailedMessage() == null) {
-			runtimeException = new MessageHandlingException(message, text.get(), ex);
+			runtimeException = new MessageHandlingException(message, text.get(),
+					(ex instanceof IllegalStateException && ex.getCause() != null) ? ex.getCause() : ex);
```
artembilan: I think this might be a bit of abuse and we might need to introduce our own exception like `MessagingMethodInvokerException` to be able to let end-user `IllegalStateException` to bubble as is.

Of course if you agree with the change to the `MessagingMethodInvokerHelper`...

PR #2769, opened 2019-02-27
Defer Messaging annotations process

artembilan:
The `AbstractMethodAnnotationPostProcessor` and its implementations
have a `beanFactory.getBean()` call for the `@Bean` methods with
Messaging annotations.
This is done, actually, from the
`MessagingAnnotationPostProcessor.postProcessAfterInitialization()`
which might be still too early in some scenarios, like Spring Cloud Feign
with its child application contexts being initialized from the
`FeignClientFactoryBean`, causing a `BeanCurrentlyInCreationException`

See https://stackoverflow.com/questions/54887963/beancurrentlyincreationexception-when-using-spring-integration-with-spring-cloud

* Implement a `SmartInitializingSingleton` for the `MessagingAnnotationPostProcessor`
and gather `Runnable` wrappers for newly introduced `postProcessMethodAndRegisterEndpointIfAny()`
to be called later in the `afterSingletonsInstantiated()` when context is
still in the initialization phase.
All runtime-registered beans are going to be processed normally from the
regular `postProcessAfterInitialization()`


artembilan:
I'll fix test-cases when we agree with the solution.
Mostly the failing tests are about not calling newly introduced `afterSingletonsInstantiated()` in the `MessagingAnnotationPostProcessor`.

garyrussell:
Please re-push with corrected commit comment `s/deffer/defer/`

artembilan:
OK. I will. Thanks.
When you say the solution is good and we even can back-port it to `5.1.x`

garyrussell:
The approach looks good.

artembilan:
Re-pushed after rebasing, fixing tests and commit message.

artembilan:
I think we should not back-port this: too much changes and might be some kind of breaking change for some use-cases...

Issue #2770, opened 2019-02-28
AbstractMessageSource - check for an empty headerExpressions Map before evaluating


garyrussell:
See https://github.com/spring-cloud/spring-cloud-stream/issues/1621

[This change](https://github.com/spring-projects/spring-integration/commit/d37e6562d8259fb760a5c5a0e3c4f03396673e33#diff-fe353c7d6f819b243595a127bc5916a5) means we always create an evaluation context, even if we don't need one.

PR #2772, opened 2019-02-28
GH-2770: AbstractMessageSource EvaluationContext

garyrussell:
Fixes https://github.com/spring-projects/spring-integration/issues/2770

Only create an evaluation context for header expressions if we actually
have header expressions.


artembilan (approved):
Merging with some polishing to have `headerExpressions` as `null` by default

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4365eaef6decf64a8667921e8fe56c25e6186404 and cherry-picked to `5.1.x`

Issue #2776, opened 2019-03-01
Possible race in FtpStreamingMessageSourceTests


garyrussell:
There seems to be an issue when the adapter is stopped and the subsequent restart doesn't get the remaining file(s).

Issue #2777, opened 2019-03-01
Defer remote file filtering logic until a file is processed


garyrussell:
The fix for https://github.com/spring-projects/spring-integration/issues/2776 will work by reversing the filtered files on exceptions and when the adapter is stopped.

 But this will not be a complete solution because unprocessed files can still be filtered if the server loses power.

Moving the filter logic until as late as possible will mitigate the extent of this problem.

PR #2778, opened 2019-03-01
GH-2776: Fix Streaming Remote File MessageSource

garyrussell:
Fixes https://github.com/spring-projects/spring-integration/issues/2776
Also see https://github.com/spring-projects/spring-integration/issues/2777

- reset the filter for the current file if the fetch fails
- implement `Lifecycle` and clear the `toBeReceived` queue and corresponding
  filter entries

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
@@ -178,7 +178,7 @@ public abstract class AbstractRemoteFileStreamingMessageSource<F>
 
 	@Override
 	protected Object doReceive() {
-		Assert.state(this.running.get(), getComponentName() + " is not running");
+		Assert.state(this.running.get(), () -> getComponentName() + " is not running");
 		AbstractFileInfo<F> file = poll();
 		if (file != null) {
 			try {
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 the original author or authors.
+ * Copyright 2016-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -149,23 +153,53 @@ public final void onInit() {
...
+					return getMessageBuilderFactory()
+							.withPayload(session.readRaw(remotePath))
+							.setHeader(IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE, session)
+							.setHeader(FileHeaders.REMOTE_DIRECTORY, file.getRemoteDirectory())
+							.setHeader(FileHeaders.REMOTE_FILE, file.getFilename())
+							.setHeader(FileHeaders.REMOTE_FILE_INFO,
+									this.fileInfoJson ? file.toJson() : file);
+				}
+				catch (IOException e) {
+					throw new MessagingException("IOException when retrieving " + remotePath, e);
```
artembilan: Can we change this to the `UncheckedIOException` instead? There is really nothing to do with the messaging...

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -176,6 +210,16 @@ protected Object doReceive(int maxFetchSize) {
 		return doReceive();
 	}
 
+	private void resetFilterIfnecessary(AbstractFileInfo<F> file) {
```
artembilan: I guess `necessary` must be capitalized to have a proper camel case style.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/StoredProcJavaConfigTests.java
```diff
@@ -77,7 +77,8 @@
 	public void test() {
 		Message<?> received = fooChannel.receive(10000);
 		assertThat(received).isNotNull();
-		Collection<?> primes = (Collection<?>) received.getPayload();
+		@SuppressWarnings("unchecked")
+		Collection<Integer> primes = (Collection<Integer>) received.getPayload();
 		assertThat(primes).containsExactly(2, 3, 5, 7);
```
artembilan: We don't fail here even right now even my IDEA doesn't like wildcards as well:

image

I wonder what is wrong then...

artembilan (changes requested):
Just a couple comments.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/StoredProcJavaConfigTests.java
garyrussell: A couple of days ago, I think I tweaked a setting in STS to be more strict, because it just started complaining about this.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -149,23 +153,53 @@ public final void onInit() {
...
+	}
+
+	@Override
+	public boolean isRunning() {
+		return this.running.get();
+	}
+
 	@Override
 	protected Object doReceive() {
+		Assert.state(this.running.get(), getComponentName() + " is not running");
```
artembilan: Use `Supplier` for the message since it is not a constant string.

I'll fix it for `4.3.x` during cherry-picking

artembilan (changes requested):
And a couple failed tests on Travis:
```
org.springframework.integration.file.remote.StreamingInboundTests > testFilterReversedOnBadFetch FAILED
    java.lang.AssertionError at StreamingInboundTests.java:222
org.springframework.integration.file.remote.StreamingInboundTests > testExceptionOnFetch FAILED
    java.lang.AssertionError at StreamingInboundTests.java:151
```

I guess because we don't throw `MessagingException` anymore...

artembilan (approved):

artembilan:
Cherry-picked to `5.1.x` after resolving conflicts as: https://github.com/spring-projects/spring-integration/commit/648760bb43ac3673b543404ecea6c87fd4fc76a8
Cherry-picked to `5.0.x` after further conflicts resolution as: https://github.com/spring-projects/spring-integration/commit/34febd443b586838a971d5d91b3f6a69448c34ef
And to `4.3.x` after more conflicts as: https://github.com/spring-projects/spring-integration/commit/45789a99b850245ef0a79990f4fa51892d411b36

PR #2779, opened 2019-03-01
GH-2777: Remote File Filter Improvements

### LINKED ISSUE #2777: Defer remote file filtering logic until a file is processed

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/2777

If the filter supports it, defer filtering until the last possible moment.
Then, the worst case scenario after a catestrophic failure (e.g. power loss),
would be that at most one file will be incorrectly filtered on restart.

**Do not merge - needs more test coverage and fix one broken test case**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractFileListFilter.java
@@ -26,6 +26,7 @@ import java.util.List;
  *
  * @author Mark Fisher
  * @author Iwein Fuld
+ * @author Gary Russell
  */
 public abstract class AbstractFileListFilter<F> implements FileListFilter<F> {
 
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractFileListFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -42,11 +42,17 @@ public abstract class AbstractFileListFilter<F> implements FileListFilter<F> {
 		return accepted;
 	}
 
+	@Override
+	public boolean supportsSingleFileFiltering() {
+		return true;
+	}
+
 	/**
 	 * Subclasses must implement this method.
 	 * @param file The file.
 	 * @return true if the file passes the filter.
 	 */
+	@Override
 	public abstract boolean accept(F file);
 
 }
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractPersistentAcceptOnceFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractPersistentAcceptOnceFileListFilter.java
@@ -173,7 +173,7 @@ public abstract class AbstractPersistentAcceptOnceFileListFilter<F> extends Abst
 			try {
 				this.flushableStore.flush();
 			}
-			catch (IOException e) {
+			catch (@SuppressWarnings("unused") IOException e) {
 				// store's responsibility to log
 			}
 		}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
@@ -63,7 +63,7 @@ public class CompositeFileListFilter<F>
 
 	public CompositeFileListFilter(Collection<? extends FileListFilter<F>> fileFilters) {
 		this.fileFilters = new LinkedHashSet<>(fileFilters);
-		this.allSupportAccept = fileFilters.stream().allMatch(f -> f.supportsSingleFileFiltering());
+		this.allSupportAccept = fileFilters.stream().allMatch(FileListFilter::supportsSingleFileFiltering);
 	}
 
 
@@ -115,7 +115,7 @@ public class CompositeFileListFilter<F>
 			}
 		}
 		this.fileFilters.addAll(filtersToAdd);
-		this.allSupportAccept &= filtersToAdd.stream().allMatch(f -> f.supportsSingleFileFiltering());
+		this.allSupportAccept &= filtersToAdd.stream().allMatch(FileListFilter::supportsSingleFileFiltering);
 		return this;
 	}
 
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
@@ -63,7 +63,7 @@ public class CompositeFileListFilter<F>
 
 	public CompositeFileListFilter(Collection<? extends FileListFilter<F>> fileFilters) {
 		this.fileFilters = new LinkedHashSet<>(fileFilters);
-		this.allSupportAccept = fileFilters.stream().allMatch(FileListFilter::supportsSingleFileFiltering);
+		this.allSupportAccept = fileFilters.stream().allMatch(FileListFilter<F>::supportsSingleFileFiltering);
 	}
 
 
@@ -86,8 +86,9 @@ public class CompositeFileListFilter<F>
 	 * @return this CompositeFileFilter instance with the added filters
 	 * @see #addFilters(Collection)
 	 */
-	@SuppressWarnings("unchecked")
-	public CompositeFileListFilter<F> addFilters(FileListFilter<F>... filters) {
+	@SafeVarargs
+	@SuppressWarnings("varargs")
+	public final CompositeFileListFilter<F> addFilters(FileListFilter<F>... filters) {
 		List<FileListFilter<F>> asList = Arrays.asList(filters);
 		return addFilters(asList);
 	}
@@ -115,7 +116,7 @@ public class CompositeFileListFilter<F>
 			}
 		}
 		this.fileFilters.addAll(filtersToAdd);
-		this.allSupportAccept &= filtersToAdd.stream().allMatch(FileListFilter::supportsSingleFileFiltering);
+		this.allSupportAccept &= filtersToAdd.stream().allMatch(FileListFilter<F>::supportsSingleFileFiltering);
 		return this;
 	}
 
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 
 import org.springframework.beans.factory.InitializingBean;
@@ -53,6 +54,8 @@ public class CompositeFileListFilter<F>
 
 	private Consumer<F> discardCallback;
 
+	private boolean allSupportAccept = true;
+
 
 	public CompositeFileListFilter() {
 		this.fileFilters = new LinkedHashSet<>();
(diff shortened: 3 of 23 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
```diff
@@ -53,13 +54,16 @@
...
+	private boolean allSupportAccept = true;
+
 
 	public CompositeFileListFilter() {
 		this.fileFilters = new LinkedHashSet<>();
 	}
 
 	public CompositeFileListFilter(Collection<? extends FileListFilter<F>> fileFilters) {
 		this.fileFilters = new LinkedHashSet<>(fileFilters);
+		fileFilters.stream().allMatch(f -> f.supportsSingleFileFiltering());
```
artembilan: Are you missing `this.allSupportAccept &=` here?
What is the point of `allMatch()` if we don't use its result?

spring-integration-file/src/main/java/org/springframework/integration/file/filters/FileListFilter.java
```diff
@@ -38,4 +39,26 @@
...
+	/**
+	 * Filter a single file.
+	 * @param file the file.
+	 * @return true if the file passes the filter, false to filter.
+	 * @since 5.2
+	 */
+	default boolean accept(F file) {
+		@SuppressWarnings("unchecked")
+		F[] fileArray = (F[]) Array.newInstance(file.getClass(), 1);
+		fileArray[0] = file;
```
artembilan: How about `F[] fileArray = (F[]) Stream.of(file).toArray();` instead?

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractFileListFilter.java
```diff
@@ -42,11 +42,17 @@
...
+		return true;
+	}
+
 	/**
 	 * Subclasses must implement this method.
 	 * @param file The file.
 	 * @return true if the file passes the filter.
 	 */
+	@Override
 	public abstract boolean accept(F file);
```
artembilan: Can we remove this `abstract` now since you have that `default` one in the super interface already?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -180,27 +185,33 @@ public boolean isRunning() {
 	protected Object doReceive() {
 		Assert.state(this.running.get(), () -> getComponentName() + " is not running");
 		AbstractFileInfo<F> file = poll();
-		if (file != null) {
-			try {
-				String remotePath = remotePath(file);
-				Session<?> session = this.remoteFileTemplate.getSession();
+		while (file != null) {
```
artembilan: I think this is something wrong for this component contract.
We can't constantly poll from the source until `null`.
The point of the `receive()` is to get file or nothing.

artembilan (changes requested):
Some review

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
garyrussell: Below - if we are successful, the return ends the loop.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
garyrussell: Doh.

garyrussell: Doh - that's exactly why I said more test coverage is needed 😄

spring-integration-file/src/main/java/org/springframework/integration/file/filters/FileListFilter.java
garyrussell:
```
Exception in thread "main" java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.String;
```

artembilan: Hm. So, the `Array.newInstance` is still need to be used, but `Stream` can remain yet 😀

garyrussell: Actually, it can be replaced with an `UnsupportedOperationException`, or `IllegalStateException`. I added this before adding the `supportsSingleFileFiltering`; any filter that returns true there must implement this method.

artembilan: Ha! Good catch!
Yes, let's go with the `UnsupportedOperationException` since `supportsSingleFileFiltering()` returns `false`!

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractPersistentAcceptOnceFileListFilter.java
```diff
@@ -173,7 +173,7 @@ protected void flushIfNeeded() {
 			try {
 				this.flushableStore.flush();
 			}
-			catch (IOException e) {
+			catch (@SuppressWarnings("unused") IOException e) {
 				// store's responsibility to log
```
artembilan: Can't we just re-throw as an `UncheckedIOException` ?

spring-integration-file/src/main/java/org/springframework/integration/file/filters/FileListFilter.java
```diff
@@ -46,16 +45,15 @@
...
+				"Filters that return true in supportsSingleFileFiltering() must implement this method");
 	}
 
 	/**
 	 * Indicates that this filter supports filtering a single file.
 	 * Default false.
 	 * @return true to allow external calls to {@link #accept(Object)}.
+	 * @since 5.2
 	 */
 	default boolean supportsSingleFileFiltering() {
```
artembilan: Can we also mention in JavaDoc for this method that `accept()` must be implemented in case of `true` ?

do you mention that in Docs though?

Thanks

spring-integration-ftp/src/test/resources/log4j2-test.xml
```diff
@@ -8,7 +8,7 @@
 	<Loggers>
 		<Logger name="org.springframework" level="warn"/>
 		<Logger name="org.springframework.integration" level="warn"/>
-		<Logger name="org.springframework.integration.file" level="warn"/>
+		<Logger name="org.springframework.integration.file" level="trace"/>
```
artembilan: I think we can revert this already altogether.

artembilan (changes requested):
Just a couple concerns.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractPersistentAcceptOnceFileListFilter.java
garyrussell: I don't think we want to throw an exception here.

spring-integration-file/src/test/resources/log4j2-test.xml
```diff
@@ -6,8 +6,8 @@
 		</Console>
 	</Appenders>
 	<Loggers>
-		<Logger name="org.springframework.integration" level="warn"/>
-		<Logger name="org.springframework.integration.file" level="warn"/>
+		<Logger name="org.springframework.integration" level="trace"/>
+		<Logger name="org.springframework.integration.file" level="trace"/>
```
artembilan: Do we really need to live with `trace` in this module?

artembilan (changes requested):
So, LGTM so far.
Pulling locally for the final review.
Meanwhile, tell me, please, if you are going to add something on the matter into Docs.

Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -162,11 +162,16 @@ public void start() {
 	@Override
 	public void stop() {
 		if (this.running.compareAndSet(true, false)) {
-			// remove unprocessed files from the queue (and filter)
-			AbstractFileInfo<F> file = this.toBeReceived.poll();
-			while (file != null) {
-				resetFilterIfNecessary(file);
-				file = this.toBeReceived.poll();
+			if (this.filter == null || this.filter.supportsSingleFileFiltering()) {
```
artembilan: Please, consider to fix unused import for `org.springframework.beans.factory.InitializingBean` in this class.
See the latest Sonar report.

Thanks

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/filters/FileListFilter.java
```diff
@@ -38,4 +38,29 @@
...
 
+	/**
+	 * Filter a single file; only called externally if {@link #supportsSingleFileFiltering()}
+	 * returns true.
+	 * @param file the file.
+	 * @return true if the file passes the filter, false to filter.
+	 * @since 5.2
+	 * @see #supportsSingleFileFiltering()
+	 */
+	default boolean accept(F file) {
```
artembilan: Your name to the `@author` list.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/LastModifiedFileListFilter.java
```diff
@@ -124,4 +124,21 @@ else if (this.discardCallback != null) {
 		return list;
 	}
 
+	@Override
+	public boolean accept(File file) {
+		if (file.lastModified() / ONE_SECOND + this.age <= System.currentTimeMillis()) {
```
artembilan: See `filterFiles()` logic: you are missing to divide `System.currentTimeMillis()` to the `ONE_SECOND`

artembilan (changes requested):
That's all after local review.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/CompositeFileListFilter.java
```diff
@@ -83,7 +88,8 @@ public void close() throws IOException {
 	 */
 	@SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
garyrussell: In STS, it suppresses "Type safety: Potential heap pollution via varargs parameter filters"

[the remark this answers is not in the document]
artembilan: Oh! Can we use `@SafeVarargs` instead then?

spring-integration-file/src/main/java/org/springframework/integration/file/filters/LastModifiedFileListFilter.java
```diff
@@ -105,16 +105,15 @@ public long getAge() {
...
+		this.discardCallback = discardCallbackToSet;
 	}
 
 	@Override
 	public List<File> filterFiles(File[] files) {
 		List<File> list = new ArrayList<>();
-		long now = System.currentTimeMillis() / ONE_SECOND;
 		for (File file : files) {
-			if (file.lastModified() / ONE_SECOND + this.age <= now) {
+			if (file.lastModified() / ONE_SECOND + this.age <= System.currentTimeMillis() / ONE_SECOND) {
```
artembilan: Well, if we decide to go this way, than it would be better to reuse `accept(File)` instead of logic duplication.

artembilan (changes requested):

src/reference/asciidoc/file.adoc
```diff
@@ -1114,3 +1114,33 @@ public class FileSplitterApplication {
 }
 ----
 ====
+
+[[remote-persistent-flf]]
+=== Remote Persistent File List Filters
+
+Inbound and streaming inbound remote file channel adapters (`FTP`, `SFTP`, and other technologies) are configured with correponding implementations of `AbstractPersistentFileListFilter` by default, configured with an in-memory `MetadataStore`.
```
artembilan: Typo: `corresponding`

Will fix on merge if all minor issues.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/931df86274c116f806b059f98a375b38dbd0fd77

PR #2786, opened 2019-03-05
Fix MMIHelper for reinitialization

artembilan:
If we provide a custom `MessageHandlerMethodFactory`, the
`MessagingMethodInvokerHelper` obtains its bean and reinitialize a
`handlerMethod`, but it is done only for single, explicit
and provided method.
In case of `Function` or `Consumer` we use local names for methods to
extract, but this is not populated to properties of the
`MessagingMethodInvokerHelper`

* Change an internal `MessagingMethodInvokerHelper` logic to recreate
`InvocableHandlerMethod` instances based on the `MessageHandlerMethodFactory`
bean for all the methods scanned on the target.
* Populate a `handlerMethodsList` for the purpose above even if we have
only one candidate
* Fix `AggregatorParserTests` to reflect the current logic around
`handlerMethodsList`
* Prove that new logic works well with a custom `MessageHandlerMethodFactory`
bean in the `MessagingAnnotationsWithBeanAnnotationTests`

**Cherry-pick to 5.1.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -271,6 +271,54 @@ public class MessagingMethodInvokerHelper extends AbstractExpressionEvaluator im
 		this.jsonObjectMapper = mapper;
 	}
 
+	private MessagingMethodInvokerHelper(Object targetObject, Class<? extends Annotation> annotationType,
+			String methodName, Class<?> expectedType, boolean canProcessMessageList) {
+
+		this.annotationType = annotationType;
+		this.methodName = methodName;
+		this.canProcessMessageList = canProcessMessageList;
+		Assert.notNull(targetObject, "targetObject must not be null");
+		if (expectedType != null) {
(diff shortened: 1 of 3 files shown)


artembilan:
Note: this might not back-port clearly to previous version because of refactoring according Sonar reports.
So, let me know if a separate PR is needed.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -359,7 +360,7 @@ private MessagingMethodInvokerHelper(Object targetObject, Class<? extends Annota
 		}
```
garyrussell: This CTOR should be moved up, it is out of place - not really related to this issue, though.

garyrussell (commented):
No issues from the review; looks good. However, I have turned on hidden variable warning in STS and this class has a lot of them and they might be flagged as new smells when Sonar runs after this is merged.

artembilan:
I would prefer to resolve those smells when I have already a report from Sonar.

artembilan:
Back-ported to `5.1.x` as https://github.com/spring-projects/spring-integration/commit/a42411f2d6487d5116e2fe092e1116ea7399964f after fixing conflict.

PR #2790, opened 2019-03-06
INT-3459: Log exceptions in case of failOver

artembilan:
JIRA: https://jira.spring.io/browse/INT-3459

When `UnicastingDispatcher` is configured with `failOver` (true by default),
it loses exceptions it caught with previous handler when the next one
processes message properly

* Add INFO logging for exceptions which are caught before going to fail
over to the next handler

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
@@ -19,6 +19,7 @@ package org.springframework.integration.dispatcher;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.Executor;
 
 import org.springframework.integration.MessageDispatchingException;
@@ -133,11 +134,11 @@ public class UnicastingDispatcher extends AbstractDispatcher {
 			return true;
 		}
 		boolean success = false;
-		Iterator<MessageHandler> handlerIterator = this.getHandlerIterator(message);
+		Iterator<MessageHandler> handlerIterator = getHandlerIterator(message);
 		if (!handlerIterator.hasNext()) {
 			throw new MessageDispatchingException(message, "Dispatcher has no subscribers");
 		}
-		List<RuntimeException> exceptions = new ArrayList<>();
+		List<RuntimeException> exceptions = null;
 		while (!success && handlerIterator.hasNext()) {
 			MessageHandler handler = handlerIterator.next();
 			try {
@@ -148,11 +149,13 @@ public class UnicastingDispatcher extends AbstractDispatcher {
 				RuntimeException runtimeException =
 						IntegrationUtils.wrapInDeliveryExceptionIfNecessary(message,
 								() -> "Dispatcher failed to deliver Message", ex);
+				if (exceptions == null) {
+					exceptions = new ArrayList<>();
+				}
 				exceptions.add(runtimeException);
 				boolean isLast = !handlerIterator.hasNext();
-				if (!isLast && this.failover && this.logger.isInfoEnabled()) {
-					this.logger.info("An exception thrown from the '" + handler + "' for the '" + message + "'. " +
-							"Will be in a failover for the next subscriber.", ex);
+				if (!isLast && this.failover) {
+					logExceptionBeforeFailOver(ex, handler, message);
 				}
 				handleExceptions(exceptions, message, isLast);
 			}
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
```diff
@@ -137,18 +137,24 @@ private boolean doDispatch(Message<?> message) {
...
 				handler.handleMessage(message);
 				success = true; // we have a winner.
 			}
-			catch (Exception e) {
-				RuntimeException runtimeException = IntegrationUtils.wrapInDeliveryExceptionIfNecessary(message,
-						() -> "Dispatcher failed to deliver Message", e);
+			catch (Exception ex) {
+				RuntimeException runtimeException =
+						IntegrationUtils.wrapInDeliveryExceptionIfNecessary(message,
+								() -> "Dispatcher failed to deliver Message", ex);
```
garyrussell: Why this line wrap? It was previously 112.

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
```diff
@@ -137,18 +137,24 @@ private boolean doDispatch(Message<?> message) {
...
 				exceptions.add(runtimeException);
-				this.handleExceptions(exceptions, message, !handlerIterator.hasNext());
+				boolean isLast = !handlerIterator.hasNext();
+				if (!isLast && this.failover && this.logger.isInfoEnabled()) {
+					this.logger.info("An exception thrown from the '" + handler + "' for the '" + message + "'. " +
+							"Will be in a failover for the next subscriber.", ex);
```
garyrussell: Sorry for my language, but the two `the` should be removed. 

```java
this.logger.info("An exception was thrown by '" + handler + "' for while handling '" + message + "'. " +
							"Failing over to the next subscriber.", ex);
```

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
```diff
@@ -137,18 +137,24 @@ private boolean doDispatch(Message<?> message) {
 		if (!handlerIterator.hasNext()) {
 			throw new MessageDispatchingException(message, "Dispatcher has no subscribers");
 		}
-		List<RuntimeException> exceptions = new ArrayList<RuntimeException>();
+		List<RuntimeException> exceptions = new ArrayList<>();
```
garyrussell: We should create this array lazily because it's only used in error situations. 99.9999% not used.

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
garyrussell: Perhaps only include full stack trace at DEBUG and `ex.getMessage()` at INFO?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
artembilan: Well, my own preferences: I find it is much readable when variable initialization is multi-line. So, I start initialization block fully from a new line.

garyrussell: OK

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/UnicastingDispatcher.java
```diff
@@ -166,10 +169,22 @@ private boolean doDispatch(Message<?> message) {
...
+	}
+
+	private void logExceptionBeforeFailOver(Exception ex, MessageHandler handler, Message<?> message) {
+		if (this.logger.isInfoEnabled()) {
+			this.logger.info("An exception was thrown by '" + handler + "' while handling '" + message + "': " +
+					ex.getMessage() + ". Failing over to the next subscriber.");
+		}
+		else if (this.logger.isDebugEnabled()) {
+			this.logger.debug("An exception was thrown by '" + handler + "' while handling '" + message +
+					"'. Failing over to the next subscriber.", ex);
```
garyrussell: Debug has to be detected first otherwise we'll always take the INFO branch.

garyrussell (changes requested):

PR #2794, opened 2019-03-07
Avoid throws Exception where possible - Phase III

garyrussell:


spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -120,14 +122,15 @@ public String putIfAbsent(String key, String value) {
...
+			catch (@SuppressWarnings(UNUSED) KeeperException.NodeExistsException e) {
 				// so the data actually exists, we can read it
 				try {
 					byte[] bytes = this.client.getData().forPath(getPath(key));
 					return IntegrationUtils.bytesToString(bytes, this.encoding);
 				}
 				catch (Exception exceptionDuringGet) {
-					throw new ZookeeperMetadataStoreException("Exception while reading node with key '" + key + "':", e);
+					throw new ZookeeperMetadataStoreException("Exception while reading node with key '" + key + "':",
+							exceptionDuringGet);
```
garyrussell: This is a bug, needs backporting (wrong exception).

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -300,10 +299,16 @@ public Boolean call() throws Exception {
...
+			catch (@SuppressWarnings("unused") TimeoutException e) {
+				if (future != null) {
+					future.cancel(true);
+				}
 				return false;
 			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				throw e;
+			}
```
garyrussell: These two are bugs (need backporting).

garyrussell (commented):
This also fixes a couple of bugs in zookeeper that will need back-porting.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
artembilan: Can you push these fixes by yourself to those branches?
Or do you want me to extract them somehow from such a big change? 😄 

Thanks

garyrussell: No; I'll do it after review; I just didn't want to forget.

PR #2798, opened 2019-03-09
GH-2690: Add MessageHandlerMethodFactory beans

### LINKED ISSUE #2690: Revise the  MessagingMethodInvokerHelper complexity and consider to add out-of-the-box beans for the MessageHandlerMethodFactory
See more discussion in the https://github.com/spring-projects/spring-integration/pull/2688

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2690

For more end-user flexibility and for simplicity of the `MessagingMethodInvokerHelper`
logic, add `IntegrationContextUtils.MESSAGE_HANDLER_FACTORY_BEAN_NAME` and
`IntegrationContextUtils.LIST_MESSAGE_HANDLER_FACTORY_BEAN_NAME`
infrastructure beans

* Use mentioned beans in the `MessagingMethodInvokerHelper` if any,
otherwise fallback to the locally configured `MessageHandlerMethodFactory`,
although this is necessary only for our non-Spring-based unit tests
* Relax `BeanFactory` requirement for the `EvaluationContext` since it
is going to be there in the Spring-based environment, but might not be
available in the plain unit tests
* Provide some optimization and refactoring for the `MessagingMethodInvokerHelper`
* Remove deprecated `HandlerMethodArgumentResolversHolder` and
`ARGUMENT_RESOLVERS_BEAN_NAME` & `LIST_ARGUMENT_RESOLVERS_BEAN_NAME`
bean definitions and constants for their names


garyrussell:
I haven't reviewed yet but the reason we log a WARN with a missing BF is because we had several components in the framework which failed to propagate the BF to the evaluator.

The warn helps us find any new errors of that kind.

I am not so keen on disabling it.

artembilan:
> The warn helps us find any new errors of that kind.

Oh! I missed that. Sure, will revert.
Thanks for the pointer!

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -943,27 +881,20 @@ private void populateHandlerMethod(Map<Class<?>, HandlerMethod> candidateMethods
...
 			try {
 				// Can't use real class because of package tangle
 				if ("org.springframework.integration.gateway.RequestReplyExchanger".equals(iface.getName())) {
-					frameworkMethod =
-							ClassUtils.getMostSpecificMethod(
-									targetClass.getMethod("exchange", Message.class),
-									this.targetObject.getClass());
-					if (LOGGER.isDebugEnabled()) {
-						LOGGER.debug(this.targetObject.getClass() +
-								": Ambiguous fallback methods; using RequestReplyExchanger.exchange()");
```
garyrussell: Why do we lose this debug log?

spring-integration-core/src/test/java/org/springframework/integration/transformer/MethodInvokingTransformerTests.java
```diff
@@ -37,10 +38,10 @@
  * @author Mark Fisher
  * @author Artem Bilan
  */
-public class MethodInvokingTransformerTests {
+class MethodInvokingTransformerTests {
 
 	@Test
-	public void simplePayloadConfiguredWithMethodReference() throws Exception {
+	void simplePayloadConfiguredWithMethodReference() throws Exception {
```
garyrussell: Why change from `public`?

garyrussell (commented):
Just a couple of questions. Not sure what's up with Travis; I tried re-running.

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
artembilan: I don’t understand why we need it at all even with the current master. Looks like we don’t do any fall backs, but just choose a first one for the `exchange` and that’s all. My feelings that it is a left over after some radical change here. So, since it doesn’t bring any value , remove it altogether!

spring-integration-core/src/test/java/org/springframework/integration/transformer/MethodInvokingTransformerTests.java
artembilan: Because it is now JUnit 5 test and that one doesn’t require `public` on methods. My point is that eventually we have to come fully only to JUnit 6 or already 6. The 4th one might just go EOL some day...

artembilan:
The problem with Travis is fully related to the story about unavailability if our repo. So, we might need to hold merge of this off until resolution of that problem.

spring-integration-core/src/test/java/org/springframework/integration/transformer/MethodInvokingTransformerTests.java
garyrussell: Just because public is not required, I don't see why it's worth the effort to change it all over the place; seems like unnecessary work to me.

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
garyrussell: ok

artembilan:
So, the Travis build is green.

PR #2805, opened 2019-03-13
INT-2277: Register ChResolver & ErrHandler beans

artembilan:
JIRA: https://jira.spring.io/browse/INT-2277

To avoid extra objects at runtime and reuse a central configuration,
register `BeanFactoryChannelResolver` and `MessagePublishingErrorHandler`
bean via `DefaultConfiguringBeanFactoryPostProcessor`

* Use those beans whenever it is necessary via `IntegrationContextUtils`
factory methods against provided `BeanFactory`
* To avoid changes in the non-managed test, those new factories fall
back to a new instance if there is no an appropriate bean in the `beanFactory`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aop/MessagePublishingInterceptor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aop/MessagePublishingInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ import org.springframework.core.LocalVariableTableParameterNameDiscoverer;
 import org.springframework.core.ParameterNameDiscoverer;
 import org.springframework.expression.Expression;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
+import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.core.MessagingTemplate;
 import org.springframework.integration.expression.ExpressionEvalMap;
 import org.springframework.integration.expression.ExpressionUtils;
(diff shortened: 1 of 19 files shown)


garyrussell:
Hmmm - no travis for some reason - I checked and all is enabled. I wonder if they suppress a build because there are conflicts? Perhaps rebase and see?

artembilan:
OK. Now it's back after rebasing.
Let's see how it feels!

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -212,4 +220,37 @@ public static Properties getIntegrationProperties(BeanFactory beanFactory) {
 		return properties;
 	}
 
+	/**
+	 * Obtain a {@link DestinationResolver} registered with the
+	 * {@value #CHANNEL_RESOLVER_BEAN_NAME} bean name.
+	 * @param beanFactory BeanFactory for lookup, must not be null.
```
garyrussell: Assertion?

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -212,4 +220,37 @@ public static Properties getIntegrationProperties(BeanFactory beanFactory) {
...
+		}
+		return beanFactory.getBean(CHANNEL_RESOLVER_BEAN_NAME, DestinationResolver.class);
+	}
+
+	/**
+	 * Obtain an {@link ErrorHandler} registered with the
+	 * {@value #MESSAGE_PUBLISHING_ERROR_HANDLER_BEAN_NAME} bean name.
+	 * By default resolves to the {@link org.springframework.integration.channel.MessagePublishingErrorHandler}
+	 * with the {@value #CHANNEL_RESOLVER_BEAN_NAME} {@link DestinationResolver} bean.
+	 * @param beanFactory BeanFactory for lookup, must not be null.
```
garyrussell: ditto

garyrussell (commented):
Cool. Just a couple of questions.

spring-integration-core/src/main/java/org/springframework/integration/aop/MessagePublishingInterceptor.java
```diff
@@ -59,29 +60,31 @@
...
 		Assert.notNull(metadataSource, "metadataSource must not be null");
 		this.metadataSource = metadataSource;
 	}
 
-
+	/**
+	 * @param metadataSource the {@link PublisherMetadataSource} to use.
+	 * @deprecated since 5.2 in favor constructor argument.
+	 */
+	@Deprecated
```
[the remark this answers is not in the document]
artembilan: Well, I'm not sure what is that *tooling*, but I guess we have already much more components with similar ctor-based initialization and no one has complained so far.

Therefore I believe we are safe to remove it now.
Well, at least deprecate since it is a mandatory ctor option anyway, but someone may still would like to call a setter. Although that is already not so good to mutate this component at runtime.

PR #2807, opened 2019-03-13
GH-2806: Add generics support to HTTP Inbound

### LINKED ISSUE #2806: Support generic types in the HttpRequestHandlingEndpointSupport
See SO for more info: https://stackoverflow.com/questions/55142118/httprequesthandlingmessaginggateway-json-array-payload

Also consider to support the same functionality in the `JsonToObjectTransformer`.

The general problem that some `HttpMessageConverter` implementations are `GenericHttpMessageConverter`, e.g. `AbstractJackson2HttpMessageConverter` with of `Type` which can be extracted from already supported `ResolvableType` for the `requestPayloadType`.

The same `ResolvableType` should be applied for the `JsonToObjectTransformer`.

Since it is an omission with the current implementation and there is no easy way to workaround it we definitely need to consider to back port it to `5.1.x`.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2806

**Cherry-pick to 5.1.x**

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/inbound/BaseHttpInboundEndpoint.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/inbound/BaseHttpInboundEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
+ * Copyright 2017-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


artembilan:
I will add `JsonToObjectTransformer` changes tomorrow.
Or that can already be fixed in the separate issue/PR...

artembilan:
@garyrussell ,

any chances to merge this, so i will bring Sonar smells fix for these classes in the separate PR.

A `JsonToObjectTransformer` change definitely should be done in the separate PR and already might be without back-porting.
We have there a type mapping against `JsonHeaders` as a workaround.

Thanks

PR #2819, opened 2019-03-18
Support generics in the `JsonToObjectTransformer`

artembilan:
* Deprecate `JsonObjectMapperAdapter` if favor of `default` methods in
the `JsonObjectMapper`
* Introduce `JsonObjectMapper.fromJson(Object, ResolvableType)` to
support generics during deserialization
* Add `JsonHeaders.RESOLVABLE_TYPE` header handling for the
`ResolvableType` management
* Add `ResolvableType` argument into the `JsonToObjectTransformer`
* Change the `JsonToObjectTransformer` logic to consult request message
headers first
* Add `ResolvableType`-based factory method into the `Transformers`
* Document the change

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/json/JsonToObjectTransformer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/json/JsonToObjectTransformer.java
@@ -20,12 +20,16 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 
 import org.springframework.beans.factory.BeanClassLoaderAware;
+import org.springframework.core.ResolvableType;
 import org.springframework.integration.mapping.support.JsonHeaders;
-import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.integration.support.json.JsonObjectMapper;
 import org.springframework.integration.support.json.JsonObjectMapperProvider;
 import org.springframework.integration.transformer.AbstractTransformer;
+import org.springframework.lang.Nullable;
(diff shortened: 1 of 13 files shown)


spring-integration-core/src/main/java/org/springframework/integration/json/JsonToObjectTransformer.java
```diff
@@ -83,21 +112,72 @@ public String getComponentType() {
...
 		}
 		catch (IOException e) {
 			throw new UncheckedIOException(e);
 		}
+
+		if (removeHeaders) {
+			return getMessageBuilderFactory()
+					.withPayload(result)
+					.copyHeaders(headers)
+					.removeHeaders(JsonHeaders.HEADERS.toArray(new String[3]))
```
garyrussell: I thought `[0]` is preferred now, not new, I know, but might as well since we're changing.

https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size

garyrussell (changes requested):
Just one minor issue.

spring-integration-core/src/main/java/org/springframework/integration/json/JsonToObjectTransformer.java
artembilan: Hm. Looks like my IDEA after some update has missed this warning 😢

Pushed the fix for `0`. More over it is right now `4` anyway 😄

garyrussell (approved):

PR #2823, opened 2019-03-19
Short-circuit methods for lambdas from annotation

artembilan:
When we have a `@ServiceActivator` or any other messaging annotations
on `Function` or `Consumer` `@Bean`s, there is a restriction when we
can't use lambdas because of target method argument type erasure in Java.

* Use a `@Bean` method return to determine the target function argument
type and wrap the call into the `LambdaMessageProcessor`.

In this case we call the target method directly after possible payload
conversion according expected generic type for `Function` or `Consumer`.
There is just no reason to go a `MessagingMethodInvokerHelper` route
for this lambda variants

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
@@ -22,6 +22,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
 
 import org.aopalliance.aop.Advice;
 import org.apache.commons.logging.Log;
@@ -40,6 +42,7 @@ import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.support.BeanDefinitionValidationException;
 import org.springframework.context.annotation.Bean;
 import org.springframework.core.GenericTypeResolver;
+import org.springframework.core.ResolvableType;
 import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.core.annotation.Order;
@@ -61,10 +64,13 @@ import org.springframework.integration.endpoint.ReactiveStreamsConsumer;
 import org.springframework.integration.endpoint.SourcePollingChannelAdapter;
 import org.springframework.integration.handler.AbstractMessageProducingHandler;
 import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.integration.handler.LambdaMessageProcessor;
+import org.springframework.integration.handler.MessageProcessor;
 import org.springframework.integration.handler.ReplyProducingMessageHandlerWrapper;
 import org.springframework.integration.handler.advice.HandleMessageAdvice;
 import org.springframework.integration.router.AbstractMessageRouter;
 import org.springframework.integration.scheduling.PollerMetadata;
+import org.springframework.integration.util.ClassUtils;
 import org.springframework.integration.util.MessagingAnnotationUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.MessageChannel;
@@ -77,7 +83,6 @@ import org.springframework.scheduling.Trigger;
 import org.springframework.scheduling.support.CronTrigger;
 import org.springframework.scheduling.support.PeriodicTrigger;
 import org.springframework.util.Assert;
-import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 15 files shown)


artembilan:
If the solution is OK, I'll go ahead with other Messaging annotations.
Also need to check with Kotlin Lambdas and, of course, really check if the target classs is lambda...

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/ServiceActivatorAnnotationPostProcessor.java
```diff
@@ -63,7 +67,17 @@ protected MessageHandler createHandler(Object bean, Method method, List<Annotati
...
 				}
 				else {
-					serviceActivator = new ServiceActivatingHandler(target);
+					ResolvableType methodReturnType = ResolvableType.forMethodReturnType(method);
+					if (Function.class.isAssignableFrom(methodReturnType.toClass()) ||
+							Consumer.class.isAssignableFrom(methodReturnType.toClass())) {
+
+						Class<?> expectedPayloadType = methodReturnType.getGeneric(0).toClass();
+						serviceActivator = new ServiceActivatingHandler(
+								new LambdaMessageProcessor(target, expectedPayloadType));
```
garyrussell: I think you need to check for wildcard generic types here.

garyrussell (dismissed):
Nice! Just one comment.

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/ServiceActivatorAnnotationPostProcessor.java
artembilan: Not sure what you mean, but with the `methodReturnType.getGeneric(0).toClass()` it returns an `Object` class for a `Consumer<?>`. So, I think we are good.

artembilan:
Ready for review.

Thanks

garyrussell:
```
> Task :spring-integration-core:checkstyleMain
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/ServiceActivatorAnnotationPostProcessor.java:23:8: Unused import - java.util.function.Consumer. [UnusedImports]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/ServiceActivatorAnnotationPostProcessor.java:24:8: Unused import - java.util.function.Function. [UnusedImports]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/ServiceActivatorAnnotationPostProcessor.java:28:8: Unused import - org.springframework.core.ResolvableType. [UnusedImports]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/ServiceActivatorAnnotationPostProcessor.java:32:8: Unused import - org.springframework.integration.handler.LambdaMessageProcessor. [UnusedImports]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/handler/ExpressionCommandMessageProcessor.java:55:17: Reference to instance variable 'methodFilter' needs "this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/util/ClassUtils.java:20:8: Unused import - java.util.HashMap. [UnusedImports]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/util/ClassUtils.java:21:8: Unused import - java.util.Map. [UnusedImports]
> Task :spring-integration-core:checkstyleMain FAILED
```

artembilan:
Pushed the fix.

Thank you for pointing that out!

There is an `AnnotationBeanUtils` deprecated in SF, but that should be considered as a separate issue...

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -510,9 +515,26 @@ protected void checkMessageHandlerAttributes(String handlerBeanName, List<Annota
 		}
 	}
 
+	protected boolean resolveAttributeToBoolean(String requiresReply) {
+		return Boolean.parseBoolean(this.beanFactory.resolveEmbeddedValue(requiresReply));
```
garyrussell: Sounds like a generic method name, but specific parameter name. `String attribute` ?

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlowDefinition.java
```diff
@@ -3118,7 +3122,8 @@ protected StandardIntegrationFlow get() {
 		if (this.integrationFlow == null) {
 			if (this.currentMessageChannel instanceof FixedSubscriberChannelPrototype) {
 				throw new BeanCreationException("The 'currentMessageChannel' (" + this.currentMessageChannel
-						+ ") is a prototype for 'FixedSubscriberChannel' which can't be created without 'MessageHandler' "
+						+ ") is a prototype for 'FixedSubscriberChannel' which can't be created without " +
+						"'MessageHandler' "
```
garyrussell: without **a** 'MessageHandler'.

garyrussell (changes requested):
Just a couple of minor issues.

garyrussell:
```
org.springframework.integration.jmx.config.OperationInvokingOutboundGatewayTests > gatewayWithPrimitiveArgs FAILED
    org.springframework.messaging.MessagingException at OperationInvokingOutboundGatewayTests.java:103
```

PR #2837, opened 2019-03-22
Optimize performance when many new connections arrive

xiexin36:
When many new connections arrive at once, we should accept as soon as possible.  accept connections in a for loop until no new connection is ready

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioServerConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioServerConnectionFactory.java
@@ -205,7 +205,7 @@ public class TcpNioServerConnectionFactory extends AbstractServerConnectionFacto
 		// accept connections in a for loop until no new connection is ready
 		for (;;) {
 			SocketChannel channel = server.accept();
-			if(channel == null) {
+			if (channel == null) {
 				return;
 			}
 			if (isShuttingDown()) {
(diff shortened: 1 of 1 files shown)


garyrussell:
Thanks for the contribution, but why not just increase the backlog?

```java
	/**
	 * The number of sockets in the connection backlog. Default 5;
	 * increase if you expect high connection rates.
	 * @param backlog The backlog to set.
	 */
	public void setBacklog(int backlog) {
````

Also, we don't accept code changes without test cases that exercise the new code.

xiexin36:
@garyrussell Yes, I also tried to set the backlog, even to 20, and I also set the nioHarvestInterval bigger than default, but it did not work.

xiexin36:
> 	 * The number of sockets in the connection backlog. Default 5;
> 	 * increase if you expect high connection rates.
> 	 * @param backlog The backlog to set.
> 	 */
> 	public void setBacklog(int backlog) {
> ```
> 
> Also, we don't accept code changes without test cases that exercise the new code.

@garyrussell  With the current test ConnectionFactoryTests.testObtainConnectionIdsNio, the new code can be converage, I have run the test.

garyrussell:
OK; thanks for the contribution.

garyrussell:
@xiexin36 Thanks for the contribution, but I have some concerns about read starvation, so I made this optional (but true by default).

Replaced by https://github.com/spring-projects/spring-integration/pull/2865

xiexin36:
> @xiexin36 Thanks for the contribution, but I have some concerns about read starvation, so I made this optional (but true by default).
> 
> Replaced by #2865

@garyrussell Great job !  And when the 5.1.4 version will be release?

garyrussell:
Currently scheduled for next Tuesday (April 2).

PR #2865, opened 2019-03-28
GH-2837: TCP NIO Prioritize accept

garyrussell:
See https://github.com/spring-projects/spring-integration/pull/2837

Give priority to accepting new connections over reading data from
existing connections.
Add a flag to allow users to revert to the previous behavior of
giving reads priority.

**cherry-pick to 5.1.x**


artembilan:
Cherry-picked to `5.1.x` as https://github.com/spring-projects/spring-integration/commit/6f3fd0961f3d6d0bbd8f8e5e645e12f2b4458f01 after fixing conflicts.

Issue #2868, opened 2019-03-29
RedisUtils.isUnlinkAvailable bug


sojw:
**Affects Version(s):** \5.1.3

---

hi. there

we are upgrading to springboot 2.1.3.

since then, NE have occurred in the redis lock part of spring-integration.
we looked for the cause.

- https://github.com/spring-projects/spring-integration/blob/d21e8f03499c93939b0a618ce96a026eef940ca2/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisUtils.java#L64

we have modified version of our redis server info and returned it.
ex) "111.111.111.111:7000.redis_version" -> "4.0.9"

If we handle this redis_version in the isUnlinkAvailable function, an NE occurs.
I think we need to fix this part.
If it does not fit in the pattern, try to return it as the default value by try / catch.

currently, we have downgraded to 5.1.2 and are running normally.

lonnyDong:
“111.111.111.111:7000.redis_version" -> "4.0.9 , the point is  info.getProperty(VERSION_PROPERTY) can not get redis_version properly，cause the  key has changed. info.getProperty(VERSION_PROPERTY)  returns NULL

PR #2870, opened 2019-03-29
Fix `RedisUtils.isUnlinkAvailable()` for NPE

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2868

* Check if `redis_version` property is returned from the `info`
before parsing

**Cherry-pick to 5.1.x**

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisUtils.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisUtils.java
@@ -23,6 +23,7 @@ import java.util.Properties;
 
 import org.springframework.data.redis.core.RedisCallback;
 import org.springframework.data.redis.core.RedisOperations;
+import org.springframework.util.StringUtils;
 
 /**
  * A set of utility methods for common Redis functions.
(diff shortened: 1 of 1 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisUtils.java
```diff
@@ -61,8 +62,14 @@ public static boolean isUnlinkAvailable(RedisOperations<?, ?> redisOperations) {
...
 			if (info != null) {
-				int majorVersion = Integer.parseInt(info.getProperty(VERSION_PROPERTY).split("\\.")[0]);
-				return majorVersion >= 4;
+				String version = info.getProperty(VERSION_PROPERTY);
+				if (StringUtils.hasText(version)) {
+					int majorVersion = Integer.parseInt(version.split("\\.")[0]);
+					return majorVersion >= 4;
+				}
+				else {
+					return false;
```
garyrussell: I wonder if we should fall back to a `spring.integration.properties`?

Clearly, their version supports the operation.

artembilan: Huh? What is the difference between `spring.integration.properties` or `redis_version` property?
Technically they are the same, only the problem that you abusing a system renaming such an important property.
I mean the same way you definitely can change the property in the `spring.integration.properties` file and also may break the system

Our goal here is to use faster delete on Redis, if it is not available by some reason we fall back to the regular blocking delete.
So, if you rename  `redis_version`, you just make your own life harder 😄 

That's why I'd stay with the `false` as a fallback.

garyrussell: OK; makes sense; thanks.

garyrussell (approved):

garyrussell:
Cherry-picked to 5.1.x as 659613c4c97785b1fd76c702f7df5be2cc4255e6

PR #2877, opened 2019-04-01
GH-2873: Preserve mapping order in the router

### LINKED ISSUE #2873: Exception match order in ErrorMessageExceptionTypeRouter
ErrorMessageExceptionTypeRouter uses ConcurrentHashMap internally, and this does not guarantee the order.
Therefore, it is not possible to flexibly route Exception with hierarchy.

link : https://github.com/spring-projects/spring-integration/blob/v5.1.3.RELEASE/spring-integration-core/src/main/java/org/springframework/integration/router/ErrorMessageExceptionTypeRouter.java#L49

For example, when routing NullPointerException and RuntimeException, NullPointerException may be routed to the channel of RuntimeException.
This is the routing that are not intended.

```
<int:exception-type-router>
  <int:mapping exception-type="java.lang.NullPointerException" channel="channelA" />
  <int:mapping exception-type="java.lang.RuntimeException" channel="channelB" />
</int:exception-type-router>
```

The following measures are required to solve this problem.
- Hold mapping information in the form of guaranteeing access order

```
// example
Map<String,Object> classNameMappings = Collections.synchronizedMap(new LinkedHashMap<String,Object>());
Map<String,Object> classNameMappings =new LinkedHashMap<String,Object>();
```
- Determine the mapping in the defined order

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2873

Sometime it is important to map to most specific exception instead of
its super class.

* Use `LinkedHashMap` for mapping keys in the
`ErrorMessageExceptionTypeRouter`, as well as in its
`AbstractMappingMessageRouter` superclass.
Since we don't do that internal map modification, there is no reason to
worry about concurrent access: we just replace an internal instance
atomically with a new `LinkedHashMap` every time we modify a mapping
for router

**Cherry-pick to 5.1.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
@@ -16,7 +16,7 @@
 
 package org.springframework.integration.dsl;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 import org.springframework.core.convert.ConversionService;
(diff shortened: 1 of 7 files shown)


artembilan:
Should be OK now.
Thanks

PR #2879, opened 2019-04-02
Fix `If-Unmodified-Since` header mapping

artembilan:
https://build.spring.io/browse/INT-MASTERSPRING40-664
https://build.spring.io/browse/INT-FATS5IC-833

Fix `DefaultHttpHeaderMapper` to populate an `If-Unmodified-Since`
request header with the same formatter as it is in the `HttpHeaders` in
Spring Web

**Cherry-pick to 5.1.x & 5.0.x**


artembilan:
It didn't fail yesterday because the day before had been `31`. The current format fails only with single number for the day of month.
With this fix we fully align our formatting for this kind of dates with what has been fixed recently in Spring Web.

garyrussell:
Cherry-picked after resolving conflicts.

PR #2881, opened 2019-04-04
Fix complexity for DefaultHttpHeaderMapper

artembilan:
* Deprecate `DefaultHttpHeaderMapper` constants which are fully
Spring Web `HttpHeaders` constants
* Introduce lowercase constants for `getHttpHeader()` switch
* Reuse `HttpHeaders` API as much as possible
* Simplify logic in some methods to break them to smaller methods


artembilan:
Do not merge yet. 
I think we should avoid Java Streams in critical paths as much as possible. 
It is really going to be some GC impact on every message headers mapping.

PR #2883, opened 2019-04-04
GH-2765: Add discardChannel for splitter

### LINKED ISSUE #2765: Add discard channel support for splitters
When encountering empty collections, splitter should be able to send the result to a discard channel. Currently, when encountering an empty collection, the splitter ends the flow. Some use-cases may rely on a custom split function which may returns empty collections. These use-cases should be able to define a discard channel so they can proceed with a possible compensation flow.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2765

When encountering empty collections, splitter should be able to send
the result to a discard channel.
Currently, when encountering an empty collection,
the splitter ends the flow.
Some use-cases may rely on a custom split function which may returns
empty collections.
These use-cases should be able to define a discard channel
so they can proceed with a possible compensation flow.

* Add `discardChannel` option to the `AbstractMessageSplitter`
* Delegate `discardChannel` population from everywhere it is possible:
DSL, XML, `AbstractMessageSplitter` extension like `FileSplitter` etc.
* Fix `FileSplitterTests` for broken charset
* Document new feature; fix some typos and out-dated code sample

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
@@ -96,12 +96,20 @@ public abstract class AbstractMessageSplitter extends AbstractReplyProducingMess
 
 	@Override
 	public MessageChannel getDiscardChannel() {
-		String channelName = this.discardChannelName;
-		if (channelName != null) {
-			this.discardChannel = getChannelResolver().resolveDestination(channelName);
-			this.discardChannelName = null;
+		if (this.discardChannel == null) {
+			String channelName = this.discardChannelName;
+			if (channelName != null) {
+				this.discardChannel = getChannelResolver().resolveDestination(channelName);
+				this.discardChannelName = null;
+			}
 		}
 		return this.discardChannel;
+	}
+
+	@Override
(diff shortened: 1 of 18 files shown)


spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
```diff
@@ -59,10 +68,46 @@ public void setApplySequence(boolean applySequence) {
...
+	 * @since 5.2
+	 */
+	public void setDiscardChannelName(String discardChannelName) {
+		Assert.hasText(discardChannelName, "'discardChannelName' must not be empty");
+		this.discardChannelName = discardChannelName;
+	}
+
+	@Override
+	public MessageChannel getDiscardChannel() {
+		String channelName = this.discardChannelName;
```
garyrussell: We should exit early if `this.discardChannel` is not null, no need to store this local variable in that case.

garyrussell: Also, if both are set the channel should win.

garyrussell (changes requested):
Just one small issue.

PR #2887, opened 2019-04-08
INT-4299: Add AbstractMailReceiver.autoCloseFolder

artembilan:
JIRA: https://jira.spring.io/browse/INT-4299

* Optimize `AbstractMessageSource` do not evaluate headers when polling
result is `null`
* Optimize `MailReceivingMessageSource` do not wrap polling result to
the message: the `AbstractMessageSource` will do that later

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.mail;
 
 import java.io.ByteArrayOutputStream;
+import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Date;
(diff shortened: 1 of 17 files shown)


artembilan:
If this is OK, I'll add some docs on the matter.

Thanks

garyrussell (changes requested):
Just minor javadoc polishing.

LGTM - awaiting docs.

spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
```diff
@@ -357,7 +378,9 @@ private Folder obtainFolderInstance() throws MessagingException {
 				return convertMessagesIfNecessary(searchAndFilterMessages());
 			}
 			finally {
-				MailTransportUtils.closeFolder(this.folder, this.shouldDeleteMessages);
+				if (this.autoCloseFolder) {
+					closeFolder();
```
garyrussell: Need to close here if no messages were retrieved (needs a test),

artembilan: Why? Why do we need to close the folder at all?
I don't see any comment in the Internet requiring to close the folder after each fetch.
😕

garyrussell: ok

garyrussell (approved):

PR #2888, opened 2019-04-08
Fix race condition around `sharedJschSession`

artembilan:
https://build.spring.io/browse/INT-MASTER-1481

The `SftpSession.connect()` may lead to race condition when we try to
open the `channel`, but `JschSession` is closed already.
It may happen in cases when we have `DefaultSftpSessionFactory`
configured for the `isSharedSession` and that shared session may be
closed by another thread before we reach the `channel.connect()`,
because we already have left the `this.sharedSessionLock` blocking path

**Cherry-pick to 5.1.x**


garyrussell:
Cherry-picked as ad8d8367b9060c488f2db838a800dbff5d2442e2

Issue #2803, opened 2019-03-12
Deprecate MetricsFactory and friends


garyrussell:

artembilan:
@garyrussell ,

Does it mean not only to deprecate `MetricsFactory` and everything around, but also remove their usage from our components?
For example in this code from the `AbstractMessageChannel`:
```
if (countsAreEnabled) {
				metricsContext = metrics.beforeSend();
				if (this.metricsCaptor != null) {
					sample = this.metricsCaptor.start();
				}
				sent = doSend(message, timeout);
				if (sample != null) {
					sample.stop(sendTimer(sent));
				}
				metrics.afterSend(metricsContext, sent);
				metricsProcessed = true;
}
```
only `metricsCaptor` should be left?

Thanks

garyrussell:
I meant deprecate the user APIs (ChannelMetrics etc) but still retain the stats; then pull everything out in a later release.

artembilan:
Yeah... I still have no idea how to proceed. 😢 

Let's circulate in on Monday!

PR #2889, opened 2019-04-08
GH-2803: Deprecate legacy metrics

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/2803

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
@@ -34,10 +34,8 @@ import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.support.management.AbstractMessageChannelMetrics;
 import org.springframework.integration.support.management.AbstractMessageHandlerMetrics;
 import org.springframework.integration.support.management.ConfigurableMetricsAware;
-import org.springframework.integration.support.management.DefaultMetricsFactory;
 import org.springframework.integration.support.management.IntegrationManagement;
 import org.springframework.integration.support.management.IntegrationManagement.ManagementOverrides;
-import org.springframework.integration.support.management.MessageSourceMetricsConfigurer;
 import org.springframework.integration.support.management.PollableChannelManagement;
 import org.springframework.integration.support.management.metrics.MetricsCaptor;
 import org.springframework.integration.support.management.micrometer.MicrometerMetricsCaptor;
@@ -78,7 +76,9 @@ public class IntegrationManagementConfigurer
 	private final Map<String, org.springframework.integration.support.management.MessageSourceMetrics>
 		sourcesByName = new HashMap<>();
 
-	private final Map<String, MessageSourceMetricsConfigurer> sourceConfigurers = new HashMap<>();
+	private final Map<String,
+		org.springframework.integration.support.management.MessageSourceMetricsConfigurer>
+			sourceConfigurers = new HashMap<>();
 
 	private ApplicationContext applicationContext;
 
@@ -115,10 +115,14 @@ public class IntegrationManagementConfigurer
 	/**
 	 * Set a metrics factory.
 	 * Has a precedence over {@link #metricsFactoryBeanName}.
-	 * Defaults to {@link DefaultMetricsFactory}.
+	 * Defaults to {@link org.springframework.integration.support.management.DefaultMetricsFactory}.
+	 * @deprecated in favor of dimensional metrics via
+	 * {@link org.springframework.integration.support.management.metrics.MeterFacade}.
+	 * Built-in metrics will be removed in a future release.
 	 * @param metricsFactory the factory.
 	 * @since 4.2
 	 */
+	@Deprecated
 	public void setMetricsFactory(org.springframework.integration.support.management.MetricsFactory metricsFactory) {
 		this.metricsFactory = metricsFactory;
 	}
@@ -127,8 +131,12 @@ public class IntegrationManagementConfigurer
 	 * Set a metrics factory bean name.
 	 * Is used if {@link #metricsFactory} isn't specified.
 	 * @param metricsFactory the factory.
+	 * @deprecated in favor of dimensional metrics via
+	 * {@link org.springframework.integration.support.management.metrics.MeterFacade}.
+	 * Built-in metrics will be removed in a future release.
 	 * @since 4.2
 	 */
+	@Deprecated
 	public void setMetricsFactoryBeanName(String metricsFactory) {
 		this.metricsFactoryBeanName = metricsFactory;
 	}
@@ -164,8 +172,12 @@ public class IntegrationManagementConfigurer
 	 * 'foo' in {@link #setEnabledCountsPatterns(String[]) enabledCountsPatterns}. For
 	 * components that match multiple patterns, the first pattern wins. Enabling stats at
 	 * runtime also enables counts.
+	 * @deprecated in favor of dimensional metrics via
+	 * {@link org.springframework.integration.support.management.metrics.MeterFacade}.
+	 * Built-in metrics will be removed in a future release.
 	 * @param enabledStatsPatterns the patterns.
 	 */
+	@Deprecated
 	public void setEnabledStatsPatterns(String[] enabledStatsPatterns) {
 		Assert.notEmpty(enabledStatsPatterns, "enabledStatsPatterns must not be empty");
 		this.enabledStatsPatterns = Arrays.copyOf(enabledStatsPatterns, enabledStatsPatterns.length);
@@ -188,11 +200,23 @@ public class IntegrationManagementConfigurer
 	 * Set whether managed components maintain message statistics by default.
 	 * Defaults to false, unless an Integration MBean Exporter is configured.
 	 * @param defaultStatsEnabled true to enable.
+	 * @deprecated in favor of dimensional metrics via
+	 * {@link org.springframework.integration.support.management.metrics.MeterFacade}.
+	 * Built-in metrics will be removed in a future release.
 	 */
+	@Deprecated
 	public void setDefaultStatsEnabled(Boolean defaultStatsEnabled) {
 		this.defaultStatsEnabled = defaultStatsEnabled;
 	}
 
+	/**
+	 * Return true if stats are enabled by default.
+	 * @return the stats enabled.
+	 * @deprecated in favor of dimensional metrics via
+	 * {@link org.springframework.integration.support.management.metrics.MeterFacade}.
+	 * Built-in metrics will be removed in a future release.
+	 */
+	@Deprecated
 	public Boolean getDefaultStatsEnabled() {
 		return this.defaultStatsEnabled;
 	}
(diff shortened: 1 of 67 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/management/MessageChannelMetrics.java
```diff
@@ -20,14 +20,18 @@
...
 
 /**
- * Interface for all message channel monitors containing accessors for various useful metrics that are generic for all
- * channel types.
+ * Interface for all message channel monitors containing accessors for various useful
+ * metrics that are generic for all channel types.
  *
  * @author Dave Syer
  * @author Gary Russell
+ * @deprecated in favor of dimensional metrics via
```
artembilan: You know I have a thought do not deprecate `*Metrics` classes, but make them to delegate to the meters from the Micrometer. I think it is still good to have actual metrics exposed from the components per se.
See `HistogramSupport.takeSnapshot()`. Looks like those `count`, `total`, `max` are usable for us.

I haven't figure out how to properly deprecate the `MetricsFactory` not removing too much test classes.
And also I didn't understand if that make sense to deprecate `ExponentialMovingAverage*` family...

I'll take a look tomorrow closer what you have so far.

Thanks for the effort!

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/management/MessageChannelMetrics.java
garyrussell: Yeah, that's why I added the new super interfaces, so we can still expose counts via the meters. If there are other interesting metrics we can add, we can put them there; either in a future 5.2 milestone, or when we pull the legacy metrics.

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
```diff
@@ -90,7 +90,7 @@
 
 	private Boolean defaultStatsEnabled = false;
 
-	private MetricsFactory metricsFactory;
+	private org.springframework.integration.support.management.MetricsFactory metricsFactory;
```
artembilan: The `setMetricsFactory`, `setMetricsFactoryBeanName` and I guess `setEnabledStatsPatterns` and `setDefaultStatsEnabled` must be deprecated in this class as well.

The same about `stats` and factory I guess is applied for the `@EnableIntegrationManagement`

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
```diff
@@ -72,11 +69,14 @@
...
 
-	private final Map<String, MessageHandlerMetrics> handlersByName = new HashMap<>();
+	private final Map<String, org.springframework.integration.support.management.MessageHandlerMetrics>
+		handlersByName = new HashMap<>();
 
-	private final Map<String, MessageSourceMetrics> sourcesByName = new HashMap<>();
+	private final Map<String, org.springframework.integration.support.management.MessageSourceMetrics>
+		sourcesByName = new HashMap<>();
 
 	private final Map<String, MessageSourceMetricsConfigurer> sourceConfigurers = new HashMap<>();
```
artembilan: Why `MessageSourceMetricsConfigurer` is not deprecated?

spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
```diff
@@ -42,8 +42,7 @@
  * @since 5.0
  */
 public final class RouterSpec<K, R extends AbstractMappingMessageRouter>
-		extends AbstractRouterSpec<RouterSpec<K, R>, R>
-		implements ComponentsRegistration {
+		extends AbstractRouterSpec<RouterSpec<K, R>, R> {
```
artembilan: How do you catch this kind of issues?

Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
```diff
@@ -37,11 +37,7 @@
 import org.springframework.integration.support.management.DefaultMetricsFactory;
```
artembilan: Why `DefaultMetricsFactory` is not deprecated?
Well, actually all the `MetricsFactory` implementations.

artembilan (changes requested):
Just a couple questions.

I guess whatever we have missed, can be caught up in the next iteration for the feature...

spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
garyrussell: I enabled an eclipse compiler error/warning option:

```
Description	Resource	Path	Location	Type
Redundant superinterface ComponentsRegistration for the type RouterSpec<K,R>, already defined by EndpointSpec<RouterSpec<K,R>,ConsumerEndpointFactoryBean,R>	RouterSpec.java	/spring-integration-core/src/main/java/org/springframework/integration/dsl	line 46	Java Problem
```

spring-integration-core/src/main/java/org/springframework/integration/support/management/BaseHandlerMetrics.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.management;
+
+import org.springframework.jmx.export.annotation.ManagedMetric;
+import org.springframework.jmx.support.MetricType;
+
+/**
+ * @author Gary Russell
```
artembilan: DITTO about JavaDocs

spring-integration-core/src/main/java/org/springframework/integration/support/management/BaseSourceMetrics.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.management;
+
+/**
+ * @author Gary Russell
```
artembilan: Same here

spring-integration-jmx/src/test/java/org/springframework/integration/jmx/config/MBeanExporterParserTests.java
```diff
@@ -125,13 +127,13 @@ private void checkCustomized(MessageChannelMetrics metrics) {
...
 
-	private void checkCustomized(MessageHandlerMetrics metrics) {
+	private void checkCustomized(org.springframework.integration.support.management.MessageHandlerMetrics metrics) {
 		assertThat(TestUtils.getPropertyValue(metrics, "handlerMetrics.duration.window")).isEqualTo(20);
 		assertThat(TestUtils.getPropertyValue(metrics, "handlerMetrics.duration.factor", Double.class))
 				.isCloseTo(1000000., Offset.offset(.01));
 	}
 
-	public static class CustomMetrics implements MetricsFactory {
+	public static class CustomMetrics implements org.springframework.integration.support.management.MetricsFactory {
```
artembilan: Can we consider do not do this in the tests any more?
Maybe it ever is going to be a fact to remove this test altogether since it covers a deprecated functionality...

spring-integration-jmx/src/test/java/org/springframework/integration/jmx/configuration/EnableMBeanExportTests.java
```diff
@@ -142,9 +141,10 @@ public QueueChannel output() {
 			return new QueueChannel();
 		}
 
+		@SuppressWarnings("deprecation")
 		@Bean
-		public MetricsFactory myMetricsFactory() {
-			return new DefaultMetricsFactory();
+		public org.springframework.integration.support.management.MetricsFactory myMetricsFactory() {
```
artembilan: Can we consider to make this test without `myMetricsFactory`?
It is deprecated anyway, so maybe no reason to assert against at all...

spring-integration-jmx/src/test/java/org/springframework/integration/monitor/ChannelIntegrationTests.java
```diff
@@ -85,7 +85,8 @@ public void testMessageChannelStatistics() throws Exception {
 
 		assertThat(messageChannelsMonitor.getChannelMetrics(intermediateChannelName)).isSameAs(intermediate);
 
-		MessageHandlerMetrics handlerMetrics = messageChannelsMonitor.getHandlerMetrics("bridge");
+		org.springframework.integration.support.management.MessageHandlerMetrics handlerMetrics = messageChannelsMonitor
+				.getHandlerMetrics("bridge");
```
artembilan: Can't we get here a new abstraction instead?

spring-integration-jmx/src/test/java/org/springframework/integration/monitor/MonitorTests.java
```diff
@@ -87,7 +87,9 @@ public void testStats() throws InterruptedException {
 			Object result = invocation.callRealMethod();
 			afterSendLatch.countDown();
 			return result;
-		}).when(channelMetrics).afterSend(Mockito.any(MetricsContext.class), Mockito.eq(Boolean.TRUE));
+		}).when(channelMetrics)
+			.afterSend(Mockito.any(org.springframework.integration.support.management.MetricsContext.class),
+					Mockito.eq(Boolean.TRUE));
```
artembilan: I wonder if we don't need this test any more since `MetricsContext` is deprecated...

artembilan (changes requested):
Gary, your thoughts about deprecating `Statistics` and `ExponentialMovingAverage` family and remove their test, of course 😄

garyrussell:
>Gary, your thoughts about deprecating Statistics and ExponentialMovingAverage family and remove their test, of course 😄

They will naturally go away when we remove the metrics, do you really feel we need to deprecate them as well?

spring-integration-core/src/main/java/org/springframework/integration/support/management/BaseChannelMetrics.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.management;
+
+import org.springframework.jmx.export.annotation.ManagedMetric;
+import org.springframework.jmx.support.MetricType;
+
+/**
+ * @author Gary Russell
```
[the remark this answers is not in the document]
garyrussell: Hmmm - I wonder why checkstyle didn't catch this.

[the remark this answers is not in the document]
artembilan: We just don't that rule enabled:
```
<!-- 		<module name="JavadocStyle"> -->
		<!-- 			<property name="checkEmptyJavadoc" value="true"/> -->
		<!-- 		</module> -->
```

There are really a lot of our classes without JavaDocs.
Although it may treat even this `@author` presence as not empty...

artembilan:
>  do you really feel we need to deprecate them as well?

How can we then remove them if we haven't deprecated them before?

I understand that they are going to be out of use, but presence of classes without usage doesn't look for me reasonable...

I'm OK to revise it later though...

spring-integration-core/src/main/java/org/springframework/integration/support/management/BaseChannelMetrics.java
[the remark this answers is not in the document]
garyrussell: Ah, right; (no, I have seen checkstyle report a missing JD, even with `@author, @since` in projects with this rule enabled).

garyrussell:
Good review; thanks for nit-picking!

Issue #2890, opened 2019-04-09
Explain the need for an initial bridge in a subflow starting with a channel


garyrussell:
See https://stackoverflow.com/questions/55591704/whats-the-purpose-of-automatically-added-bridge-component-after-router

artembilan:
See my comments in that SO question. 
Maximum I would consider this as a documentation improvement for better end-user experience.

garyrussell:
I am ok with closing this as "Working as Designed".

These bridges are very lightweight. Perhaps another option would be to add an option `suppressSyntheticBridges` on the `IntegrationGraphServer` (if the bridge input channel has a name with the  the form `.*\.#0`.

garyrussell:
Converted to "documentation"

PR #2891, opened 2019-04-09
GH-2890: Avoid unnecessary bridge [PoC Only]

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/2890

Seems to work, but the test needs some work.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.dsl;
 
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -25,7 +26,11 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.springframework.context.SmartLifecycle;
+import org.springframework.integration.channel.DirectChannel;
+import org.springframework.integration.dsl.support.MessageChannelReference;
+import org.springframework.integration.handler.BridgeHandler;
 import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
 
 /**
  * The standard implementation of the {@link IntegrationFlow} interface instantiated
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
```diff
@@ -74,8 +79,31 @@
...
+		}
+		if (iterator.hasNext()) {
+			two = iterator.next();
+		}
+		if (iterator.hasNext()) {
+			three = iterator.next();
+		}
+		// We have at least 4 components and the first three are a synthesized bridge
+		if (iterator.hasNext() && three != null
+				&& one instanceof DirectChannel && ((DirectChannel) one).getBeanName() == null
```
artembilan: I think this is some kind of abusing. I can just do `.channel(new DirectChannel())` and you will remove it in this logic.

spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
```diff
@@ -74,8 +79,31 @@
...
+				&& two instanceof GenericEndpointSpec
+				&& ((GenericEndpointSpec<? extends MessageHandler>) two).getHandler() instanceof BridgeHandler
```
artembilan: I may intentionally want a bridge between channels. Even if they are `MessageChannelReference` instances, I may add `ChannelInterceptor` to them eventually at runtime.

spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
```diff
@@ -899,6 +911,32 @@ public IntegrationFlow propagateErrorFromGatherer(TaskExecutor taskExecutor) {
 					.get();
 		}
 
+		@Bean
+		public IntegrationFlow routerWithSubflowStartsWithAChannel() {
+			return IntegrationFlows.from(MessageChannels.direct())
+					.route("'foo'", mapping -> mapping
+							.subFlowMapping("foo",  subflowStartsWithAChannel()))
```
artembilan: This doesn't reflect the use-case in the SO question.
The real point is that it is used as a *lambda*, so we don't have a `StandardIntegrationFlow` instance there yet...

artembilan (changes requested):
I'm not sure in your solution yet...

We may introduce something like `InnerDirectChannel` or `SubFlowStartChannel`, which could really be an indicator to remove eventually when we have sub-flow lambda resolved to the `StandardIntegrationFlow`. Although I might be still wrong and there is just no way to figure out it yet...

spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
garyrussell: This is still a lambda (`IntegrationFlow`).

artembilan: We have a test config like this:
```
@Bean
		public IntegrationFlow upperCase() {
			return f -> f
					.<String>handle((p, h) -> p.toUpperCase());
		}

		@Bean
		public IntegrationFlow routeSubflowToReplyChannelFlow() {
			return f -> f
					.<Boolean>route("true", m -> m
							.subFlowMapping(true, upperCase())
					);
		}

```

And over some AOP magic I ended up in the `StandardIntegrationFlow.getInputChannel()` which returns for me an `upperCase.input` `DirectChannel`.
That's all just because we call a *bean method* which is going to resolve to a bean and we have there `IntegrationFlowLifecycleAdvice` wrapping a lambda and delegate to the created internally `StandardIntegrationFlow`.

I understand that we will still have there a `bridge` if you declare a `channel()` in the beginning of that lambda, but that's already all by design.
If you would like to avoid that, please, use `IntegrationFlows.from([CHANNEL_NAME])`.

So, I still insist that we need to end up with the documentation improvement only.

PR #2898, opened 2019-04-18
Add support for Kotlin JSR223 scripts

artembilan:
* Add required Kotlin dependencies into the `spring-integration-scripting`
module
* Introduce `KotlinScriptExecutor` to interact with the
`KotlinJsr223JvmLocalScriptEngineFactory` directly since there is no
`META-INF/services/javax.script.ScriptEngineFactory` file in the Kotlin
* Also set an `idea.use.native.fs.for.win` system property to `false` in
this class to disable check for native support on Windows.
(Might be removed in future Kotlin versions)
* Move `ScriptParser.getLanguageFromFileExtension()` logic into the
`ScriptExecutorFactory.deriveLanguageFromFileExtension()` since the same
one must be applied in the `DslScriptExecutingMessageProcessor`, too.
* Modify tests to reflect Kotlin support
* Fix some test scripts to their official extensions


spring-integration-scripting/src/main/java/org/springframework/integration/scripting/jsr223/ScriptExecutorFactory.java
```diff
@@ -16,13 +16,21 @@
...
 /**
+ * The scripting configuration utilities.
+ *
  * @author David Turanski
+ * @author Artem Bilan
+ *
  * @since 2.1
  */
-public abstract class ScriptExecutorFactory {
+public final class ScriptExecutorFactory {
```
garyrussell: Needs a `private` CTOR.

garyrussell (changes requested):
Just one small issue.

spring-integration-scripting/src/main/java/org/springframework/integration/scripting/jsr223/ScriptExecutorFactory.java
```diff
@@ -31,7 +39,32 @@ public static ScriptExecutor getScriptExecutor(String language) {
...
+		if (extension.equals("kts")) {
+			return "kotlin";
+		}
+		ScriptEngineManager engineManager = new ScriptEngineManager();
+		ScriptEngine engine = engineManager.getEngineByExtension(extension);
+		Assert.state(engine != null, () -> "No suitable scripting engine found for extension '" + extension + "'");
+		return engine.getFactory().getLanguageName();
+	}
+
+	private ScriptExecutorFactory() {
```
artembilan: > Needs a private CTOR

Doesn't this work?

garyrussell: Sorry; missed that 😦

garyrussell: But it needs `super()` to prevent Sonar from complaining.

artembilan: OK. Will add. Can we tweak Sonar do not complain on the matter?
There is definitely a `super()` in the target byte code, so it's complaining is pointless.

Or do I miss something else ?

garyrussell: Adding super is a no-op - the compiler generates one anyway.

I don't think we want to disable empty block detection.

garyrussell (approved):

artembilan:
> garyrussell approved these changes 34 minutes ago 

Are you going to merge this? Or have you let it to me to squash and rebase?

Thanks

garyrussell:
I was waiting for our friend Travis - merging now.

PR #2899, opened 2019-04-19
Support non-Object types for WebFlux requests

artembilan:
* Add the support for a `Publisher`, `Resource` and `MultiValueMap`
into the `WebFluxRequestExecutingMessageHandler`
* Along side with the `WebFluxRequestExecutingMessageHandler.setPublisherElementType`
and `WebFluxRequestExecutingMessageHandler.setPublisherElementTypeExpression`,
add XSD support for the `publisher-element-type(-expression)`, which
is used for the element type when request body is a `Publisher`
* Polishing for `AbstractHttpRequestExecutingMessageHandler`


garyrussell:
Merged as f8f69c912993b791ece194b0e5c39f4e9ca1c02d with minor doc polishing.

PR #2900, opened 2019-04-24
Add caught exception as suppressed in SftpSession

heffebaycay:
The `org.springframework.integration.sftp.session.SftpSession#rename` method catches an instance of `SftpException` but does not actually use it.

This makes is harder to troubleshoot issues as this exception cannot be accessed by the callers.

This commit adds this exception as "suppressed" to the exceptions that can be thrown by this method.

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
@@ -208,14 +208,18 @@ public class SftpSession implements Session<LsEntry> {
 				}
 			}
 			catch (IOException ioex) {
-				throw new NestedIOException("Failed to delete file " + pathTo, ioex);
+				NestedIOException nioex = new NestedIOException("Failed to delete file " + pathTo, ioex);
+				nioex.addSuppressed(sftpex);
+				throw nioex;
 			}
 			try {
 				// attempt to rename again
 				this.channel.rename(pathFrom, pathTo);
 			}
 			catch (SftpException sftpex2) {
-				throw new NestedIOException("failed to rename from " + pathFrom + " to " + pathTo, sftpex2);
+				NestedIOException nioex2 = new NestedIOException("failed to rename from " + pathFrom + " to " + pathTo, sftpex2);
+				nioex2.addSuppressed(sftpex);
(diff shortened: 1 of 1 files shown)


garyrussell:
Why is this necessary? The root exception is added as the `cause`.

From the javadoc...

```java
     * <p>Note that when one exception {@linkplain
     * #initCause(Throwable) causes} another exception, the first
     * exception is usually caught and then the second exception is
     * thrown in response.  In other words, there is a causal
     * connection between the two exceptions.
     *
     * In contrast, there are situations where two independent
     * exceptions can be thrown in sibling code blocks, in particular
     * in the {@code try} block of a {@code try}-with-resources
     * statement and the compiler-generated {@code finally} block
     * which closes the resource.
     *
     * In these situations, only one of the thrown exceptions can be
     * propagated.  In the {@code try}-with-resources statement, when
     * there are two such exceptions, the exception originating from
     * the {@code try} block is propagated and the exception from the
     * {@code finally} block is added to the list of exceptions
     * suppressed by the exception from the {@code try} block.  As an
     * exception unwinds the stack, it can accumulate multiple
     * suppressed exceptions.
```

So suppressed exceptions are not used for causal relationships.

heffebaycay:
@garyrussell Correct me if I'm wrong, but the exception I'm interested in, `sftpex` is actually swallowed and not added as the root cause.

If this makes more sense, I could also update the code to actually throw `sftpex` with either `nioex` or `nioex2` as suppressed exceptions. The rationale being that if this method throws an exception, it is ultimately because of `sftpex`.
However, this would be problematic for retrocompatibility.

garyrussell:
Oh, my apologies, I should have looked at the code; I didn't read your patch closely enough; you are correct; suppressed exceptions is the right approach here.

Sorry for the noise.

garyrussell:
However, reading the javadocs again, I believe the `sftpex` should be the `cause` with the secondary exception being suppressed (to be consistent with the try-with-resources behavior).

garyrussell:
@heffebaycay Thanks for the contribution; I decided to fix this, along with some others, in https://github.com/spring-projects/spring-integration/pull/2911, but reversing the cause/suppressed so that the cause is the primary exception and the suppressed exception is the secondary.

Thanks again.

PR #2902, opened 2019-04-25
Add initial support for RSockets

artembilan:
* Add `spring-integration-rsocket` module and respective dependencies
* Implement `RSocketOutboundGateway` based on the Spring Messaging
`RSocketRequester`.
This component supports dynamic RSocket properties via expressions
against request message.
to handle `Publisher` for requests, it must be present in the request
message `payload` instead of `FluxMessageChannel` upstream, since the
last one just flattens events to be handled in the `MessageHandler` one
by one.
The result `Mono` is subscribed downstream in the `FluxMessageChannel`
or directly by the `AbstractReplyProducingMessageHandler`.
If result is a `Flux` it is just wrapped into the `Mono` to be processed
downstream by end-user code.
The point is that these request/replies are volatile and live in the
particular context meanwhile a `FluxMessageChannel` is long living
publisher in the application context boundaries.
* The `RSocketOutboundGatewayIntegrationTests` is an adapted copy of
`RSocketClientToServerIntegrationTests` from Spring Messaging
* Add `doOnError()` into the `Flux` created in the
`AbstractMessageProducingHandler` for `Publisher` replies


artembilan:
@garyrussell ,

You can merge if it is OK: other features and Docs can be added in the upcoming PRs.
Or let me know and I'll continue in this branch with the rest of stuff

Thanks

build.gradle
```diff
@@ -596,6 +598,18 @@ project('spring-integration-rmi') {
 	}
 }
 
+project('spring-integration-rsocket') {
+	description = 'Spring Integration RSockets Support'
```
garyrussell: It appears this should be singular `RSocket`.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGateway.java
```diff
@@ -0,0 +1,303 @@
...
+	public void setExpectedResponseTypeExpression(Expression expectedResponseTypeExpression) {
+		this.expectedResponseTypeExpression = expectedResponseTypeExpression;
+	}
+
+
+	@Override
+	protected void doInit() {
+		super.doInit();
+		RSocketRequester.Builder builder = RSocketRequester.builder();
+		if (this.factoryConfigurer != null) {
```
garyrussell: It's probably cleaner to default to a no-op configurers and assert not null in setters.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGateway.java
```diff
@@ -0,0 +1,303 @@
...
+/**
+ * An Outbound Messaging Gateway for RSocket client requests.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.2
+ *
+ * @see RSocketRequester
+ */
+public class RSocketOutboundGateway extends AbstractReplyProducingMessageHandler {
```
garyrussell: Shouldn't this be `RSocketMessageHandler` since it supports `fireAndForget` (adapter semantics)?

spring-integration-rsocket/src/test/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGatewayIntegrationTests.java
```diff
@@ -0,0 +1,437 @@
...
+/**
+ * @author Artem Bilan
+ *
+ * @since 5.2
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class RSocketOutboundGatewayIntegrationTests {
+
+	private static final int PORT = SocketUtils.findAvailableTcpPort();
```
garyrussell: Is there no support for port 0 so the OS allocates the port?

garyrussell (changes requested):
Good work! Just a few issues.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGateway.java
artembilan: I thought about that, too, but there is already a `RSocketMessageHandler` in Spring Messaging and its point is about a *server* side, so we don't have a choice and choose a different name.

Re. `fireAndForget` (a-la one-way): it is really doesn't fit to reactive model. We always need to return something (in this case a `Mono<Void>`) to let end-user to decide when to subscribe to it and, essentially, perform a request.
So, even if it is empty, it still going to be a reply from this gateway. The `FluxMessageChannel` downstream will subscribe to that `Mono` or `async` mode will do an eager subscription otherwise.

In summary: we always return something from here, even if it is just about `fireAndForget`

garyrussell: ok

spring-integration-rsocket/src/test/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGatewayIntegrationTests.java
artembilan: I just can't swallow code complexity in Netty, so it isn't clear if it is there.
@rstoyanchev just uses a static `7000` port on the matter 😢 

/CC @smaldini

garyrussell: It's just that we went through a process of converting all our TCP module tests to use port 0 so the OS chooses the port. These "findAvailablePort" schemes, while they work most of the time, we occasionally saw failures (BindException) on CI builds. A hard-wired port # doesn't sound good either.

artembilan: OK. I have figured out how to use `0` for port.

See latest commit.

Thanks

spring-integration-rsocket/src/test/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGatewayIntegrationTests.java
```diff
@@ -74,7 +74,7 @@
 @DirtiesContext
 public class RSocketOutboundGatewayIntegrationTests {
 
-	private static final int PORT = SocketUtils.findAvailableTcpPort();
+	private static int PORT;
```
garyrussell: This should no longer be static.

artembilan: ??? It is initialized from the `static @BeforeAll` and it won't be available in the `ClientConfig` below which has to be a `static` as well.

garyrussell: Sorry; I meant it should no longer be upper case.

Issue #2908, opened 2019-04-29
Attribute `name` in @Header not taken into account


chengchen:
**Affects Version(s):**
v5.1.4.RELEASE

## Bug report

When we annotate some parameters with `@Header` while publishing some messages, the attribute `name` is not taken into account. Please refer to the test case that I have modified (without the fix, it's going to break).

PR #2909, opened 2019-04-29
make sure the attribute  is taken into account in @Header annotation

chengchen:
Hello,

This PR is related to the issue #2908 that I have submitted.
It's pretty trivial fix and I have run the whole test suite.

Please take a look, many thanks!

Chengchen

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSource.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aop/MethodAnnotationPublisherMetadataSource.java
@@ -204,8 +204,10 @@ public class MethodAnnotationPublisherMetadataSource implements PublisherMetadat
 	@SuppressWarnings("unchecked")
 	private <T> T getAnnotationValue(Annotation annotation, String attributeName, Class<T> expectedType) {
 		T value = null;
-		Object valueAsObject = (attributeName == null) ? AnnotationUtils.getValue(annotation)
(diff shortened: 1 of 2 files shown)


artembilan:
That's indeed a good catch!

Looks like this is only a place where we deal with `Method.getParameterAnnotations()` directly and that standard Java API returns for raw annotations, not those synthesized by Spring like we have in case of `@Payload` in the `getExpressionForPayload()`.

I think we are good to merge this and back-port to `5.1.x`, but at the same time we need to think about using a new `MergedAnnotations` API in the current `master`.
Separate issue though.

@chengchen ,

Thank you for the contribution!

Would you mind also to add your name to the `@author` list for all the affected classes?

artembilan:
I've just realized that we have the same issue with the `@Payload` on the code:
```
Assert.state("".equals(AnnotationUtils.getValue(currentAnnotation)),
	"@Payload on a parameter for a @Publisher method may not contain an expression");
```

so, when I configure (modify existing code in test, essentially) like this:
```
@Publisher
	public void methodWithVoidReturnAndParameterPayloadAnnotation(@Payload(expression = "foo") String payload) {
	}
```
It doesn't fail on that assertion. Just because the the `@AliasFor` is not resolved here again for the same raw Java API issue.

Would you mind to address that as well?

Thanks

chengchen:
@artembilan Definitely, willing to help!

artembilan:
@chengchen ,

chengchen:
@artembilan thanks for the high reactivity. I am glad to contribute to high quality well managed projects!

PR #1359, opened 2015-01-16
INT-3601: Use `ChannelResolver` instead of BF

artembilan:
JIRA: https://jira.spring.io/browse/INT-3601
- Move `setChannelResolver(DestinationResolver<MessageChannel> channelResolver)` to the `IntegrationObjectSupport`
- Introduce `IntegrationObjectSupport#getChannelResolver()`
- Change all `IntegrationObjectSupport` inheritors to use `getChannelResolver()` instead of direct `beanFactory` usage
- Fix `ServiceActivatorEndpointTests` do not fall

**Cherry-pick to 4.1.x**

PR #2910, opened 2019-04-29
Sonar complexity issues

garyrussell:


spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
```diff
@@ -210,6 +210,14 @@ public void afterPropertiesSet() {
 			}
 		}
 
+		adviceChain();
+		if (this.channelResolver == null) {
+			this.channelResolver = new BeanFactoryMessageChannelDestinationResolver(this.beanFactory);
```
artembilan: Why is this not a `BeanFactoryChannelResolver`?
Although why don't you use a `ChannelResolverUtils.getChannelResolver()` instead any way?

artembilan (changes requested):
Looks like only one issue.

spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
garyrussell: Hehe - you tell me: https://github.com/spring-projects/spring-integration/pull/1359

artembilan: Oh! Looks like that one was an omission... 😄 

Or an idea was that this kind of `DestinationResolver` doesn't look into the `HeaderChannelRegistry` since we really don't have any headers interaction at this point...

Let's move it to the `ChannelResolverUtils.getChannelResolver()` anyway!

I don't think it is going to be some kind of an overhead that we still rely here on our global `BeanFactoryChannelResolver` bean

garyrussell: Yikes, I pushed it to master instead of to this PR 😦

artembilan: OK. So, rebase then.
That's why I try to be more specific in every commit message.
`Polishing - PR comment` is not so informative 😉

PR #2915, opened 2019-04-30
Synchronize RemoteFileUtils.makeDirectories()

garyrussell:
https://stackoverflow.com/questions/55918648

Avoid the race with concurrent threads.

**cherry-pick to 5.x.x, 4.3.x**


artembilan:
... and cherry-picked to all the mentioned versions.

PR #2917, opened 2019-05-01
Sonar fixes

garyrussell:
* exposing internal objects
* boolean complexity
* names, methods with _
* missing `default` in switch
* lost stack trace
* unnecessary null check before instanceof
* unused parameter

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/FilterFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/FilterFactoryBean.java
@@ -132,9 +132,13 @@ public class FilterFactoryBean extends AbstractStandardMessageHandlerFactoryBean
 	@Override
 	protected boolean canBeUsedDirect(AbstractMessageProducingHandler handler) {
 		return handler instanceof MessageFilter
-				|| (!(handler instanceof MessageSelector)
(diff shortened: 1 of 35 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/FilterFactoryBean.java
```diff
@@ -132,9 +132,13 @@ protected void postProcessReplyProducer(AbstractMessageProducingHandler handler)
 	@Override
 	protected boolean canBeUsedDirect(AbstractMessageProducingHandler handler) {
 		return handler instanceof MessageFilter
-				|| (!(handler instanceof MessageSelector)
-				&& this.discardChannel == null && this.throwExceptionOnRejection == null
-				&& this.discardWithinAdvice == null);
+				|| (!(handler instanceof MessageSelector) && nofilterAttributesProvided());
+	}
+
+	private boolean nofilterAttributesProvided() {
```
artembilan:
```suggestion
	private boolean noFilterAttributesProvided() {
```
?

artembilan (changes requested):
Only one naming concern.
Although we can merge with that as well: it's just a `private` method.

Let's wait for green Travis then!

PR #2921, opened 2019-05-02
Move SF dependencies to the latest BUILD-SNAPSHOT

artembilan:
To avoid class loader issues it is better to align Spring Data
dependencies with SF version we use explicitly.
This way we don't have a `ClassLoader` issue when a newly introduced
`TransactionManager` might not be visible from the previous versions

* Fix code style warnings in the `TransactionSynchronizationFactoryBean`
* Use `JavaUtils` in the `TransactionSynchronizationFactoryBean` to
decrease code complexity


artembilan:
We need to consider to merge this ASAP.
This is getting as a blocker already: https://build.spring.io/browse/INT-MASTERSPRING40-695/

Thanks

PR #2923, opened 2019-05-08
Add `RSocketInboundGateway`; refactoring

artembilan:
* Extract an `AbstractRSocketConnector` for common client and server connectors logic
* Introduce an `IntegrationRSocketAcceptor` and `IntegrationRSocket` for the
mapping and handling logic of RSockets and messages in between
* Introduce an `IntegrationRSocketEndpoint` marker interface for Inbound Gateway mappings
* Add `RSocketInboundGateway` implementation, which is called from the
`IntegrationRSocketAcceptor` by the `IntegrationRSocketEndpoint` mapping
* Add `RSocketConnectedEvent` to emit when the client is connected to the server.
It does not make sense in Spring Integration to delegate such a logic into the
`RSocketInboundGateway`

### DIFF
--- a/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/IntegrationRSocket.java
+++ b/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/IntegrationRSocket.java
@@ -1,0 +1,238 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.rsocket;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
+
+import org.reactivestreams.Publisher;
+
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.core.io.buffer.DataBuffer;
+import org.springframework.core.io.buffer.DataBufferFactory;
+import org.springframework.core.io.buffer.DataBufferUtils;
+import org.springframework.core.io.buffer.NettyDataBuffer;
+import org.springframework.core.io.buffer.NettyDataBufferFactory;
+import org.springframework.lang.Nullable;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.handler.DestinationPatternsMessageCondition;
+import org.springframework.messaging.handler.invocation.reactive.HandlerMethodReturnValueHandler;
+import org.springframework.messaging.rsocket.RSocketPayloadReturnValueHandler;
+import org.springframework.messaging.rsocket.RSocketRequester;
+import org.springframework.messaging.rsocket.RSocketRequesterMethodArgumentResolver;
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.messaging.support.MessageHeaderAccessor;
+import org.springframework.util.Assert;
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
+import org.springframework.util.StringUtils;
+
+import io.netty.buffer.ByteBuf;
+import io.rsocket.AbstractRSocket;
+import io.rsocket.ConnectionSetupPayload;
+import io.rsocket.Payload;
+import io.rsocket.RSocket;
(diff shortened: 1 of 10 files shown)


artembilan:
This is not fully ready yet.

At least tests are needed to cover an introduced functionality.

The server inbound side and docs might be deferred into a separate PR...

Right now I just want to share what I have so far to gather any possible feedback.

Thanks

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/IntegrationRSocketAcceptor.java
```diff
@@ -0,0 +1,132 @@
...
+	public void addEndpoint(IntegrationRSocketEndpoint endpoint) {
+		registerHandlerMethod(endpoint, HANDLE_MESSAGE_METHOD,
+				new CompositeMessageCondition(
+						new DestinationPatternsMessageCondition(endpoint.getPath(), getPathMatcher())));
+	}
+
+	@Override
+	public Mono<RSocket> accept(ConnectionSetupPayload setupPayload, RSocket sendingRSocket) {
+		IntegrationRSocket rsocket = createRSocket(sendingRSocket);
+		rsocket.setApplicationEventPublisher(this.applicationEventPublisher);
```
garyrussell: null check or no-op default implementation.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/IntegrationRSocket.java
```diff
@@ -0,0 +1,238 @@
...
+ * Also this class doesn't delegate a {@link #handleConnectionSetupPayload}
+ * into the target handler {@link Function} and emits a {@link RSocketConnectedEvent} instead.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.2
+ *
+ * @see org.springframework.messaging.rsocket.MessagingRSocket
+ */
+class IntegrationRSocket extends AbstractRSocket {
```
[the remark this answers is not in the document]
garyrussell: Oh, I see it's not a bean.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/IntegrationRSocket.java
```diff
@@ -0,0 +1,238 @@
...
+		}
+		// frameDecoder does not apply to connectionSetupPayload
+		// so retain here since handle expects it..
+		payload.retain();
+		String destination = getDestination(payload);
+		DataBuffer dataBuffer = retainDataAndReleasePayload(payload);
+		int refCount = refCount(dataBuffer);
+		return Mono
+				.defer(() -> {
+					this.applicationEventPublisher.publishEvent(
```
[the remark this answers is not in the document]
garyrussell: See comment at caller.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	public DataBuffer getData() {
+		return this.data;
+	}
+
+	public RSocketRequester getRequester() {
+		return this.requester;
+	}
+
```
garyrussell: I find `toString()` useful on events.

garyrussell (changes requested):
A few comments from first review pass.

artembilan:
Pushed some further implements and implementation.
See commit comment for details.

Thanks

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
```diff
@@ -61,4 +63,13 @@ public RSocketRequester getRequester() {
 		return this.requester;
 	}
 
+	@Override
+	public String toString() {
+		return "RSocketConnectedEvent{" +
+				"destination='" + this.destination + '\'' +
+				", data=" + Arrays.toString(this.data.asByteBuffer().array()) +
```
garyrussell: I don't think rendering a `byte[]` as a bunch of ints is useful, maybe omit this (or just include the length, perhaps).

garyrussell (changes requested):
Just one issue from this pass.

artembilan:
OK. Now I think it is very close for merging.

Pay attention, please, how `@Transformer` looks in the `RSocketInboundGatewayIntegrationTests.CommonConfig`. It is nice to have it without subscription - everything is done by the RSocket protocol.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
```diff
@@ -67,7 +65,7 @@ public RSocketRequester getRequester() {
 	public String toString() {
 		return "RSocketConnectedEvent{" +
 				"destination='" + this.destination + '\'' +
-				", data=" + Arrays.toString(this.data.asByteBuffer().array()) +
+				", data length=" + this.data.asByteBuffer().array().length +
```
garyrussell: This still seems to be a lot of overhead in a `toString()`.

Why do we need this `DataBuffer` in the event?

PR #2934, opened 2019-05-16
RSocket XML Config support

artembilan:


artembilan:
I can add Java DSL and Docs into this PR or separately.

Thanks

PR #2938, opened 2019-05-25
Fix TcpNioConnection timedOut set when close.

xiexin36:


artembilan:
Sounds like a good catch!

Merging and cherry-picking to all the supported and effected versions.

Thank you!

garyrussell:
I was going to add a test for this - if we had had a test, we would have caught it before now.

PR #2946, opened 2019-05-30
Build: check external asciidoc links

garyrussell:


build.gradle
```diff
@@ -855,8 +855,44 @@ task prepareAsciidocBuild(type: Sync) {
 	into "$buildDir/asciidoc"
 }
 
-task('makePDF', type: org.asciidoctor.gradle.AsciidoctorTask) {
+task checkAsciidocLinks {
 	dependsOn prepareAsciidocBuild
+	inputs.dir("$buildDir/asciidoc")
```
artembilan: Why do you do this in `buildDir`, but not in the `src/reference/asciidoc/` ?

build.gradle
```diff
@@ -855,8 +855,44 @@ task prepareAsciidocBuild(type: Sync) {
...
+					if (!anchors.contains(anchor)) {
+						throw new InvalidUserDataException("Anchor '" + anchor
+							+ "' not found in '" + file.name
+							+ "', if in another file, it needs to be qualified with './fileName.adoc#'")
+					}
+				}
+				else {
+					if (!hasComma) {
+						throw new InvalidUserDataException("External anchor '" + anchor
+							+ "' should have a textual description (,...)")
```
artembilan: Shouldn't we include a `file.name` in this exception as well?

artembilan (changes requested):
A couple concerns.

build.gradle
garyrussell: Why does it matter? The `prepareAsciidocBuild` copies them there and all other doc tasks work on that dir.

artembilan: Got it. Thanks!

build.gradle
garyrussell: Good catch

PR #2951, opened 2019-06-04
Reinstate DSL File Transformer Factory

garyrussell:


spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransformers.java
```diff
@@ -0,0 +1,52 @@
...
+ * @since 5.2
+ *
+ */
+public final class FileTransformers {
+
+	private FileTransformers() {
+		super();
+	}
+
+	public static FileToStringTransformer fileToString() {
```
artembilan: Can't we have these simple factory methods just directly in the `Files` class?
And make that `final`, too 😉

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/Files.java
```diff
@@ -226,4 +233,37 @@ public static FileToByteArrayTransformer toByteArrayTransformer(boolean deleteFi
 		return fileToByteArrayTransformer;
 	}
 
+	/**
+	 * Create a {@link FileToStringTransformer} with {@link Charset#defaultCharset()}.
+	 * @return the transformer.
+	 * @since 5.2
+	 */
+	public static FileToStringTransformer fileToString() {
```
artembilan: Wait (and sorry), looks like we have there already here:
```
public static FileToStringTransformer toStringTransformer(String charset) {
...
public static FileToByteArrayTransformer toByteArrayTransformer() {
```

😢

artembilan (changes requested):
Time to close your PR altogether? 😉

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/Files.java
garyrussell: Doh

PR #2961, opened 2019-06-11
GH-2958: Also Add to Common Header in AMHEMS

### LINKED ISSUE #2958: Add generic source_message header
To allow streams app to always have access to the raw message regardless of whether or not retry is enabled.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/2958

The previous commit added the raw source message to a common header
`IntegrationMessageHeaderAccessor.SOURCE_DATA` so that the spring-cloud-stream
can construct an error message containing the data when using functions.

However, we also need to add it to the common header when the `ErrorMessage`
is created by Spring Integration for the normal message channel binder.

This allows the DLQ processing to be agnostic as to who created the
error message.


artembilan:
... and cherry-picked to `5.1.x`

PR #2963, opened 2019-06-11
[2962] Mismatch between mechanism to check if nullChannel bean exists and mechanism to fetch that bean

### LINKED ISSUE #2962: Mismatch between mechanism to check if nullChannel bean exists and mechanism to fetch that bean
**Found it in**  5.1.X Though I'm sure it affects other versions as well.

## Bug report
consider 

> registerNullChannel method in org.springframework.integration.config.DefaultConfiguringBeanFactoryPostProcessor

```
	private void registerNullChannel() {
		if (this.beanFactory.containsBean(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)) {
			BeanDefinition nullChannelDefinition = null;
			if (this.beanFactory.containsBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)) {
				nullChannelDefinition =
						this.beanFactory.getBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME);
			}
			else {
				BeanDefinitionRegistry parentBeanFactory =
						(BeanDefinitionRegistry) this.beanFactory.getParentBeanFactory();
				if (parentBeanFactory != null) {
					nullChannelDefinition =
							parentBeanFactory.getBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME);
				}
			}

			if (nullChannelDefinition != null &&
					!NullChannel.class.getName().equals(nullChannelDefinition.getBeanClassName())) {
				throw new IllegalStateException("The bean name '" + IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME
						+ "' is reserved.");
			}
		}
		else {
			this.registry.registerBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME,
					new RootBeanDefinition(NullChannel.class));
		}
	}
```
So what it does is first ask the beanFactory if the bean exists - **this.beanFactory.containsBean**

If you look at the code for this method you'll notice that it goes all the way up the spring context heirarchy looking for the method.

And if this method returns true, the bean is requested.

When it does not find it in the current bean factory it only checks the immediate parent factory for the bean - it does not roll all the way up the heirarchy in the way that the containsBean method does.

So when your nullChannel bean is in not in the immediate parent (e.g. it's in the parent of the parent) - this method will fail.

I'll try see if I can create a test case - though I suspect it might be simpler for me to just create a pull request that rolls up the heirarchy looking for the factory that does contain the bean. So call containsBeanDefinition on each parent until a true is returned and then call getBeanDefinition on that beanfactory.

michaelwiles:
…2962

Pull request to resolve https://github.com/spring-projects/spring-integration/issues/2962

I did it very defensively so with lots of type checking etc as I can't be sure of the parent factories. I think this is all necessary as we're having to fetch the bean definition from the outside.

Also apologies if the formatting is not correct. I did try to use the eclipse formatter spec but the section of code I changed may not be formatted correctly.

Also see some of my inline comments.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
@@ -32,6 +32,8 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanClassLoaderAware;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.HierarchicalBeanFactory;
 import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.BeanDefinitionHolder;
@@ -172,18 +174,21 @@ class DefaultConfiguringBeanFactoryPostProcessor
 	private void registerNullChannel() {
 		if (this.beanFactory.containsBean(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)) {
 			BeanDefinition nullChannelDefinition = null;
-			if (this.beanFactory.containsBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)) {
-				nullChannelDefinition =
-						this.beanFactory.getBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME);
-			}
-			else {
-				BeanDefinitionRegistry parentBeanFactory =
-						(BeanDefinitionRegistry) this.beanFactory.getParentBeanFactory();
-				if (parentBeanFactory != null) {
-					nullChannelDefinition =
-							parentBeanFactory.getBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME);
+			BeanFactory beanFactory = this.beanFactory;
+			do {
+				if (beanFactory instanceof ConfigurableListableBeanFactory) {
+					ConfigurableListableBeanFactory listable = (ConfigurableListableBeanFactory) beanFactory;
+					if (listable.containsBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)) {
+						nullChannelDefinition = listable
+								.getBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME);
+					}
 				}
-			}
+				// don't think it will hurt to always find the parent
+				if (beanFactory instanceof HierarchicalBeanFactory) {
+					beanFactory = ((HierarchicalBeanFactory) beanFactory).getParentBeanFactory();
+				}
+				// will definitely be found as containsBean returned true - but also want to be defensive in case of NPE
+			} while (nullChannelDefinition == null && beanFactory != null); // not sure if beanFactroy not null is necessary
 
 			if (nullChannelDefinition != null &&
 					!NullChannel.class.getName().equals(nullChannelDefinition.getBeanClassName())) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
@@ -85,6 +85,7 @@ import org.springframework.util.ClassUtils;
  * @author Oleg Zhurakousky
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Michael Wiles
  *
  * @see IntegrationContextUtils
  */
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@michaelwiles Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2963)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2963) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@michaelwiles Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=2963)!

michaelwiles:
Added a test

spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
```diff
@@ -172,18 +174,21 @@ private void registerMessagePublishingErrorHandler() {
...
+			do {
+				if (beanFactory instanceof ConfigurableListableBeanFactory) {
+					ConfigurableListableBeanFactory listable = (ConfigurableListableBeanFactory) beanFactory;
+					if (listable.containsBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)) {
+						nullChannelDefinition = listable
+								.getBeanDefinition(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME);
+					}
 				}
-			}
+				// don't think it will hurt to always find the parent
```
artembilan: I'm not sure that we really are going to accept so internal thoughts as comments.
Why can't we really live without these comments at all?

spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
```diff
@@ -172,18 +174,21 @@ private void registerMessagePublishingErrorHandler() {
...
+				if (beanFactory instanceof HierarchicalBeanFactory) {
+					beanFactory = ((HierarchicalBeanFactory) beanFactory).getParentBeanFactory();
+				}
+				// will definitely be found as containsBean returned true - but also want to be defensive in case of NPE
+			} while (nullChannelDefinition == null && beanFactory != null); // not sure if beanFactroy not null is necessary
```
artembilan: The `while` must be on a new line. 
I think we indeed don't need to check for `beanFactory != null`, since you are right that `containsBean(IntegrationContextUtils.NULL_CHANNEL_BEAN_NAME)` does the trick for us do not fail with the infinite loop. 
Although if Checkstyle or Sonar Qube will require it I would just live with an extra check in this condition.

spring-integration-core/src/test/java/org/springframework/integration/configuration/EnableIntegrationTests.java
```diff
@@ -470,6 +470,28 @@ public void testMessagingGateway() throws InterruptedException {
...
+		parent.refresh();
+
+		AnnotationConfigApplicationContext child;
+		child = new AnnotationConfigApplicationContext();
+		child.register(ChildConfiguration.class);
+		child.setParent(parent);
+		child.refresh();
+
+		parent.close();
+		child.close();
```
artembilan: This test doesn't verify anything. It is just not a test.
This is indeed not enough just to say it should not fail. It even won't fail for some other reasons we may make changes.
So, if we add a test it must verify something.

Why don't check that child context doesn't create a new `NullChannel` just because there is one in the parent?
More over I think such a test won't fail even in the current state of code. just because your original concern that we don't go deep into the hierarchy. So, a test must have at least 3 "relatives" in the parent-child relationship.

artembilan (changes requested):
Here is some my review.

Thanks

spring-integration-core/src/test/java/org/springframework/integration/configuration/EnableIntegrationTests.java
michaelwiles: Fair point though there are 3 levels as the default level coming from the test itself is one of the levels.
This test will definitely fail on the current code.

But I'll see what verify I can add - check that a null channel is not added to the child context would be something.

spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
michaelwiles: Agreed and no problem

michaelwiles: I guess I was just explaining my thinking

spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
michaelwiles: Yes, I'll ensure that the coding style is adhered to. Finding meeting coding guidelines quite challenging

spring-integration-core/src/test/java/org/springframework/integration/configuration/EnableIntegrationTests.java
michaelwiles: Assertions added as suggested. This does make a lot of sense - provies the null channel is being found in each case by the parent context.

spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
michaelwiles: comments and null check on bean factory removed.

artembilan (changes requested):
Please, add your name to `@author` list for all affected classes.

That's all - otherwise LGTM

michaelwiles:
should I rebase into 1 commit?

artembilan:
No, we do that on merge.
Thanks

michaelwiles:
Cool. This was my first pull request btw

artembilan:
... and cherry-picked to `5.1.x` as https://github.com/spring-projects/spring-integration/commit/3668d9f1d77a077e7d0d30862485a3411c895ca0 after fixing conflicts in the test in regards of missed yet AssertJ support.

@michaelwiles ,

PR #2954, opened 2019-06-06
INT-3073: Add FluxAggregatorMessageHandler

artembilan:
JIRA: https://jira.spring.io/browse/INT-3073

PR #2964, opened 2019-06-12
Fix SplitterMH for primitive arrays cast

artembilan:
The plain cast like `(Object[]) array` doesn't work for array of
primitives - fails with `ClassCastException`

* Use `ObjectUtils.toObjectArray()` instead which does a smart job to
determine a wrapper for primitive and cast into that array of types
afterwards


artembilan:
The fix for `master` is included in the #2954

PR #2965, opened 2019-06-14
Fix new tangle introduced by Control Bus changes

garyrussell:
- tangle between `expression` and `endpoint`
- move `Pausable` to `core`


garyrussell:
Heads-up @olegz

PR #2972, opened 2019-06-21
Fix HttpHeaderMapper to support MimeType

artembilan:
It turns out that `contentType` header (as well as `Accept`) may be
converted into `MimeType` upstream before the message is sent into HTTP
Outbound Channel Adapter meanwhile the logic is only to support String
and MediaType

* Change the `DefaultHttpHeaderMapper` to support more generic `MimeType`
for the `Content-Type` and `Accept` HTTP headers which can simply be
converted to the expected `MediaType`

**Cherry-pick to 5.1.x & 5.0.x**


garyrussell:
Looks like more SF RSocket changes (travis).

artembilan:
> more SF RSocket changes

Yep! Looking into that... 😢

artembilan:
Rebased to the `master` after fixing RSocket module.

garyrussell:
Please backport; severe conflicts (at least superficially).

artembilan:
Back-ported to `5.1.x` as https://github.com/spring-projects/spring-integration/commit/68dfbbe3f0e46194eefffede41d5dae5acc25dee

PR #2973, opened 2019-06-21
Fix `SmartLifecycle.stop(Runnable)` usage

artembilan:
We always have to call `callback` in the `SmartLifecycle.stop(Runnable)`
implementation independently of component state
* Fix `StandardIntegrationFlow.stop(Runnable)` for a logic when we
don't have any `this.lifecycles`
* Remove those `stop(Runnable)` which are fully equivalent of the
`default` on in the `SmartLifecycle`
* Some other simple polishing for the affected classes, e.g.
`isSingleton()` is `default` with `true` in the `InitializingBean`

**Cherry-pick to 5.1.x**


garyrussell:
Travis (CheckStyle)

garyrussell:
Cherry-picked as eea5089d692545cb58d876f846bf4f336f9d92c2 after resolving conflicts.

Issue #2974, opened 2019-06-24
TcpNetConnection: Read exception - Socket is closed


membersound:
I'm creating a spring-integration client that connects to a server socket. The socket is a singleUse connection.

Problem: the server that I connect to forcibly closes the connection after my client received the server response. Therefore I'm getting lots of errors as follows.
I have to add that those errors do not occur all the time, only occasionally. But still often enough.

```
2019-06-24 12:15:25,106 ERROR o.s.i.i.t.c.TcpNetConnection: Read exception my.socket.host:10136:42132:e89e5c73-4987-428e-834d-61a85a739036 SocketException:Socket is closed
2019-06-24 12:16:14,026 ERROR o.s.i.i.t.c.TcpNetConnection: Read exception my.socket.host:10136:42148:07957d9e-9191-49d1-b8ad-96acf85ed6ef SocketException:Socket is closed
2019-06-24 12:16:24,736 ERROR o.s.i.i.t.c.TcpNetConnection: Read exception my.socket.host:10136:42158:53b879c5-289c-478e-b1b5-97a9cbc956fa SocketException:Socket is closed
2019-06-24 12:17:17,207 ERROR o.s.i.i.t.c.TcpNetConnection: Read exception my.socket.host:10136:42202:864c413e-7cac-4fc4-9c0e-79cbc98e8645 SocketException:Socket is closed
```

This is my simple socket config:
```
    @Bean
    public AbstractClientConnectionFactory clientFactory() throws Exception {
        TcpConnectionFactoryFactoryBean fact = new TcpConnectionFactoryFactoryBean();
        fact.setType("client");
        fact.setHost(host);
        fact.setPort(port);
        fact.setUsingNio(false);
	fact.setSingleUse(true);
        fact.setDeserializer(new MyDeserializer());
        fact.afterPropertiesSet();
        return (AbstractClientConnectionFactory) fact.getObject();
    }

      static class MyDeserializer implements Deserializer<String> {
		@Override
		public String deserialize(InputStream inputStream) throws IOException {
			try (BufferedReader br = new BufferedReader(new InputStreamReader(inputStream))) {
				return br.lines().collect(Collectors.joining("\n"));
			}
		}
	}
```

Suggestion: could you add a feature that should NOT log any error if socket is forcibly closed by the remote server? Or could there be a workaround?

```
    @Bean
    @ServiceActivator(inputChannel = "clientChannel")
[...]

PR #2976, opened 2019-06-25
GH-2974: Fix race in TcpNetConnection.getPayload() [51x]

garyrussell:
Fixes https://github.com/spring-projects/spring-integration/issues/2974

There is a race in that we could get a `SocketException` in `inputStream`.
Since this is between payloads, it needs to be thrown as a
`SoftEndOfStreamException`.

Also fix unnecessary `this.` in `MessageHistoryConfigurer.java`.

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNetConnectionTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNetConnectionTests.java
@@ -16,6 +16,8 @@
 
 package org.springframework.integration.ip.tcp.connection;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.mockito.Mockito.doAnswer;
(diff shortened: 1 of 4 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNetConnectionTests.java
```diff
@@ -137,4 +147,30 @@ public void transferHeaders() throws Exception {
...
+		server.registerListener(message -> {
+			return false;
+		});
+		server.afterPropertiesSet();
+		server.start();
+		assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
+		Socket socket = SocketFactory.getDefault().createSocket("localhost", port.get());
+		TcpNetConnection connection = new TcpNetConnection(socket, false, false, publisher, "socketClosedNextRead");
+		socket.close();
+		assertThatThrownBy(() -> connection.getPayload())
```
artembilan: This is not going to back-port smoothly: there is no AssertJ in the requested versions.
I'll figure during cherry-picking...

artembilan:
Cherry-picked into `5.0.x` as https://github.com/spring-projects/spring-integration/commit/93cd669b761fa48b264261c376132e0cf9bf1ea1 after fixing conflicts.
Cherry-picked into `4.3.x` as https://github.com/spring-projects/spring-integration/commit/2e7c5d82f26eb85c3f2dc1adf31538272f3b9cc5 with more conflicts resolutions.

PR #2979, opened 2019-06-28
Fix Redis components for JDK deserialization

artembilan:
It turns out that `JdkSerializationRedisSerializer` by default is
based on the default Java class loader which may lead into
`ClassCastException` downstream after deserialization

* Make all the `JdkSerializationRedisSerializer` usage (default)
in Redis components based on the BeanFactory `ClassLoader`
* Fix tests to call `setBeanClassLoader()`
* Fix Mark Fisher's name in the `MultipartFileReader` :-)

**Cherry-pick to 5.1.x, 5.0.x & 4.3.x after restoring diamonds**

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/multipart/MultipartFileReader.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/multipart/MultipartFileReader.java
@@ -23,7 +23,8 @@ import org.springframework.web.multipart.MultipartFile;
 /**
  * Strategy for reading {@link MultipartFile} content.
  *
- * @author mark Fisher
+ * @author Mark Fisher
+ *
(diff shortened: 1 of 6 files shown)


garyrussell:
It doesn't cherry-pick to 5.1.x cleanly because of INT-2277.

artembilan:
OK. Let me do that on my side.

Thanks

artembilan:
Cherry-picked to `5.1.x` and `5.0.x` as https://github.com/spring-projects/spring-integration/commit/c64dcc10d898d6d38c24a1576e2674677d4f971e after fixing conflicts.
Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/tree/4.3.x after restoring diamonds changes.

PR #2981, opened 2019-07-01
Fix race condition around Redis key

artembilan:
https://build.spring.io/browse/INT-FATS5IC-922/
https://build.spring.io/browse/INT-MJATS41-1764/

The `RedisQueueMessageDrivenEndpointTests` may be called from different
CI plans against the same Redis instance.

So, clean up keys before and after every unit test

**Cherry-pick until 4.3.x**

### DIFF
--- a/spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/RedisQueueMessageDrivenEndpointTests.java
+++ b/spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/RedisQueueMessageDrivenEndpointTests.java
@@ -30,6 +30,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/RedisQueueMessageDrivenEndpointTests.java
```diff
@@ -78,24 +81,38 @@
 @DirtiesContext
 public class RedisQueueMessageDrivenEndpointTests extends RedisAvailableTests {
 
+	public static final String TEST_QUEUE = "testQueue";
```
garyrussell: Don't we need to randomize this to avoid collisions?

artembilan: Yeah... I thought about that, but looks like the problem is more about not cleaned key in Redis.
If you insist, I can do that now or let's deffer when we fail again! 😄

garyrussell: No; it's fine.

artembilan:
Back-ported down to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/cdc8ac0d6cdd7424c5ce4a8a641778328cd4013c after fixing the logic to slightly different direction. Looks like `delete()` is an async, so we can't use the same key in between.

PR #2982, opened 2019-07-02
INT-2480: Add aggregate headers strategy

artembilan:
JIRA: https://jira.spring.io/browse/INT-2480

* Introduce `headers-function` option into the `aggregator` for merging
and computing headers for the output message based on the completed
group
* Implement a `DefaultAggregateHeadersFunction` and use it in the
`AbstractAggregatingMessageGroupProcessor` for default behavior with
possible injection for any other implementation
* Add `DelegatingMessageGroupProcessor` to wrap any other
`MessageGroupProcessor` implementations with possible usage of the
`headersFunction` if result is not a `Message` or `MessageBuilder`
* Make `AbstractCorrelatingMessageHandler.getOutputProcessor()` as
`public` rto give access to this option from the `AggregatorSpec` to
be able to inject a `headersFunction` in Java DSL configuration
* Add `AbstractIntegrationMessageBuilder.getHeader()` to get access to
some underlying header avoiding extra `Map` in case of `getHeaders()`
* Change a logic in the `AbstractMessageProducingHandler.produceOutput()`
to consult a `reply` for the `replyChannel` as well `routingSlip` header
if the `reply` is a `Message` or `MessageBuilder`
* Introduce a `AbstractMessageProducingHandler.messageBuilderForReply()`
and use it in `AbstractMessageSplitter` to avoid duplication
* Validate a new functionality in tests
* Fix `FileOutboundGatewayParserTests` to rely on the `TemporaryFolder`
to clean up test files after using

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/DefaultAggregateHeadersFunction.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/DefaultAggregateHeadersFunction.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.aggregator;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.integration.IntegrationMessageHeaderAccessor;
+import org.springframework.integration.store.MessageGroup;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+
+/**
+ * @author Artem Bilan
+ *
+ * @since 5.2
+ */
+public class DefaultAggregateHeadersFunction implements Function<MessageGroup, Map<String, Object>> {
+
+	private static final Log LOGGER = LogFactory.getLog(DefaultAggregateHeadersFunction.class);
+
+	@Override
+	public Map<String, Object> apply(MessageGroup messageGroup) {
+		Map<String, Object> aggregatedHeaders = new HashMap<>();
+		Set<String> conflictKeys = doAggregateHeaders(messageGroup, aggregatedHeaders);
+		for (String keyToRemove : conflictKeys) {
+			if (LOGGER.isDebugEnabled()) {
+				LOGGER.debug("Excluding header '" + keyToRemove + "' upon aggregation due to conflict(s) "
+						+ "in MessageGroup with correlation key: " + messageGroup.getGroupId());
+			}
+			aggregatedHeaders.remove(keyToRemove);
+		}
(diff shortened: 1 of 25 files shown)


artembilan:
If the solution is OK. I'll come back with Docs.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/aggregator/DefaultAggregateHeadersFunction.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.integration.IntegrationMessageHeaderAccessor;
+import org.springframework.integration.store.MessageGroup;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+
+/**
+ * @author Artem Bilan
```
garyrussell: javadocs

spring-integration-core/src/main/java/org/springframework/integration/aggregator/DefaultAggregateHeadersFunction.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @since 5.2
+ */
+public class DefaultAggregateHeadersFunction implements Function<MessageGroup, Map<String, Object>> {
```
garyrussell: I wonder if it would be better to introduce `AggregateHeadersFunction extends ...` ?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/DelegatingMessageGroupProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+		this.beanFactory = beanFactory;
+		if (this.delegate instanceof BeanFactoryAware) {
+			((BeanFactoryAware) this.delegate).setBeanFactory(beanFactory);
+		}
+	}
+
+	@Override
+	public Object processMessageGroup(MessageGroup group) {
+		Object result = this.delegate.processMessageGroup(group);
+		if (!(result instanceof Message<?>) && !(result instanceof AbstractIntegrationMessageBuilder)) {
```
garyrussell: What about `o.s.messaging.support.MessageBuilder`?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/aggregator/DefaultAggregateHeadersFunction.java
artembilan: For what reason, Gary?
Essentially we would need to introduce such `Function` extensions everywhere we use `Fucntion`s. It is not done in other Frameworks, like Reactor, so I don't think this is going to be confusing for end-users to figure out that strategy is just `Function<MessageGroup, Map<String, Object>>`.

Well, eventually we might need to think about removing many other functional interface strategies in favor of `java.util.function`...

spring-integration-core/src/main/java/org/springframework/integration/aggregator/DelegatingMessageGroupProcessor.java
artembilan: That one is out of our control and doesn't provide access to the headers.
As a workaround you can call its `build()` and have a `Message` already here. 
We try to deal with `AbstractIntegrationMessageBuilder` in our code because it is a result for `MessageBuilderFactory` and we may have some gain not creating message every time we want just add some headers internally.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/DefaultAggregateHeadersFunction.java
garyrussell: OK

artembilan:
Let me know if I can go ahead with Docs!

garyrussell:
Yes; please do; LGTM

garyrussell (approved):

garyrussell:
Merged as 5a1846cfe502989f043ba353372ebcb30940cabe with minor doc polishing.

PR #2984, opened 2019-07-08
Add simple Avro transformers

garyrussell:
Unsophisticated Avro transformers for `SpecificRecord` implementations.


garyrussell:
Docs to follow after review.

spring-integration-core/src/main/java/org/springframework/integration/dsl/Avro.java
```diff
@@ -0,0 +1,45 @@
...
+ * @since 5.2
+ *
+ */
+public final class Avro {
+
+	public static SimpleToAvroTransformer toAvro() {
+		return new SimpleToAvroTransformer();
+	}
+
+	public static SimpleFromAvroTransformer fromAvro(Class<? extends SpecificRecord> type) {
```
artembilan: Can't these two go to the `Transformers` factory?

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleFromAvroTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+	public SimpleFromAvroTransformer(Class<? extends SpecificRecord> type) {
+		this.type = type;
+	}
+
+	@Override
+	protected Object doTransform(Message<?> message) {
+		Assert.state(message.getPayload() instanceof byte[], "Payload must be a byte[]");
+		DatumReader<?> reader = new SpecificDatumReader<>(this.type);
+		try {
+			return reader.read(null, this.decoderFactory.binaryDecoder((byte[]) message.getPayload(), null));
```
artembilan: Isn't there some `ClassLoader` configuration in this Avro API?
Fully unclear how a class is loaded for instantiation...
Also, we might need to consider some whitelisting. Otherwise we will bump typical deserialization vulnerability.

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleToAvroTransformer.java
```diff
@@ -0,0 +1,62 @@
...
+ * @since 5.2
+ *
+ */
+public class SimpleToAvroTransformer extends AbstractTransformer {
+
+	private final EncoderFactory encoderFactory = new EncoderFactory();
+
+	@Override
+	protected Object doTransform(Message<?> message) {
+		Assert.state(message.getPayload() instanceof SpecificRecordBase,
```
artembilan: Isn't `SpecificRecord` enough here?
At least that is what you are asking on the next line for casting and that is what is your contract for the deserialization.

artembilan (changes requested):
Some review. Although I admit that I don't know what is Avro at all...

spring-integration-core/src/main/java/org/springframework/integration/dsl/Avro.java
garyrussell: I did that originally but figured it would mean a non-optional dependency on Avro because of the `SpecificRecord` reference.

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleToAvroTransformer.java
garyrussell: Yes; missed this one; I originally had `SpecificRecordBase` util I realized `getSchema()` is on the `SpecificRecord`.

spring-integration-core/src/main/java/org/springframework/integration/dsl/Avro.java
artembilan: OK. Can we then just not have this factory at all?
Doesn't look like it brings too much value over simple direct usage of those transformers.
Or do you have something else in your mind in regards to Avro features?

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleFromAvroTransformer.java
garyrussell: I will look regarding the class loader; I don't believe we need while listing because the developer provides us with the type here (from which the schema is obtained). So that is, in effect the white list.

We don't deserialize arbitrary types here.

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleToAvroTransformer.java
artembilan: Doh. Indeed! You ask for the `type`. So, both my concerns are eliminated.
Never mind 😢

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleFromAvroTransformer.java
garyrussell: The type's class loader is used (`type.getClassLoader()`) so whatever classloader was used to load the constructor arg is used to create the instance.

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleFromAvroTransformer.java
```diff
@@ -37,18 +43,94 @@
...
+	private final Class<? extends SpecificRecord> defaultType;
+
+	private final Map<String, Class<? extends SpecificRecord>> typeMappings = new HashMap<>();
 
 	private final DecoderFactory decoderFactory = new DecoderFactory();
 
-	public SimpleFromAvroTransformer(Class<? extends SpecificRecord> type) {
-		this.type = type;
+	private Expression typeIdExpression =
+			EXPRESSION_PARSER.parseExpression("headers['" + AvroHeaders.TYPE_ID + "']");
```
artembilan: I think `FunctionExpression` is much efficient here.

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleFromAvroTransformer.java
```diff
@@ -37,18 +43,94 @@
...
+		this.defaultType = defaultType;
+	}
+
+	/**
+	 * Set type mappings.
+	 * @param typesToMap the types to map.
+	 * @return the transformer.
+	 * @see #typeIdExpression
+	 */
+	public SimpleFromAvroTransformer typeMappings(Map<String, Class<? extends SpecificRecord>> typesToMap) {
```
artembilan: I wonder why you don't honor XML configuration any more...
It really would be great to have them as `set...`, so end-user still simply can use this class for the `<bean>` definition.

On the other hand why do we need this mapping at all? Why a simple expression/function execution against a message is not enough to get an expected type for deserialization? At least that is exactly what we do in the HTTP/WebFlux/RSocket and many other places.

spring-integration-core/src/main/java/org/springframework/integration/transformer/support/AvroHeaders.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 the original author or authors.
+ * Copyright 2013-2019 the original author or authors.
```
artembilan: Why `2013` ?

src/reference/asciidoc/transformer.adoc
```diff
@@ -453,6 +453,30 @@ Starting with version 5.2, the `JsonToObjectTransformer` can be configured with
...
 The `ObjectToJsonTransformer` now also populates a `JsonHeaders.RESOLVABLE_TYPE` header based on the request message payload for any possible downstream scenarios.
 
+[[Avro-transformers]]
+===== Apache Avro Transformers
+
+Version 5.2 added simple transformers to transform to/from Apache Avro.
+
+They are unsophisticated in that there is no schema registry; the transformers simply use the schema embedded in the `SpecificRecord` implementation generated from the Avro schema.
+
+Messages sent to the `SimpleToAvroTransformer` must have a payload that implements `SpecificRecord`; the transformer can handle multiple types.
```
artembilan: Can't this `SimpleToAvroTransformer` populate a serialized type into the headers?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleFromAvroTransformer.java
```diff
@@ -0,0 +1,161 @@
...
+ * @since 5.2
+ *
+ */
+public class SimpleFromAvroTransformer extends AbstractTransformer implements BeanClassLoaderAware {
+
+	private final Class<? extends SpecificRecord> defaultType;
+
+	private final DecoderFactory decoderFactory = new DecoderFactory();
+
+	private final Map<String, Class<? extends SpecificRecord>> typeCache = new ConcurrentHashMap<>();
```
artembilan: Shouldn't this be based on some eviction policy?
On the other hand isn't there a cache already on `ClassLoader` level, so any subsequent `ClassUtils.forName()` doesn't have impact?

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleToAvroTransformer.java
```diff
@@ -0,0 +1,65 @@
...
+		try {
+			writer.write(specific, encoder);
+			encoder.flush();
+		}
+		catch (IOException e) {
+			throw new UncheckedIOException(e);
+		}
+		return getMessageBuilderFactory().withPayload(out.toByteArray())
+				.copyHeaders(message.getHeaders())
+				.setHeader(AvroHeaders.TYPE, specific.getClass().getName())
```
artembilan: Why can't we just send a `Class<?>` as is?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleToAvroTransformer.java
garyrussell: Because if it goes to, say, RabbitMQ, he will do a `toString()` on it and it won't be the simple class name (`class ...`).

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleFromAvroTransformer.java
garyrussell: Good point

spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleToAvroTransformer.java
artembilan: Well, shouldn't we fix the mapping there so when it sees a `Class<?>` it maps it into a `getName()` instead of `toString()`?

I just don't think that it is really this component responsibility to deal with others omissions...

artembilan (approved):
LGTM

Do you have anything else to add to this PR?

garyrussell:
No; I think it is done; thanks for the good suggestions.

PR #2991, opened 2019-07-18
Add type mapping to SimpleToAvroTransformer

garyrussell:


spring-integration-core/src/main/java/org/springframework/integration/transformer/SimpleToAvroTransformer.java
```diff
@@ -41,6 +45,64 @@
...
+		assertExpressionNotNull(expression);
+		this.typeIdExpression = expression;
+	}
+
+	/**
+	 * Set the expression to evaluate against the message to determine the value
+	 * for the {@link AvroHeaders#TYPE} header.
+	 * @param expression the expression.
+	 */
+	public void setTypeExpression(String expression) {
```
artembilan: This is not good for Java Beans introspector (used in Spring XML parsers). It just won't be able to determine which setter to use for incoming `String` value from the XML tag attribute.
That's why in other similar places such a setter is called with a `String` suffix.

artembilan (changes requested):
May you also can resolve this @jmax01 's request as well: https://github.com/spring-projects/spring-integration/commit/9a2b75bae3b7670995256481e145cbb78186780c#r34360543 ?

Or let's wait for his feedback and possible contribution!

garyrussell:
We're still debating the name to use so lets treat that as a separate PR.

PR #2997, opened 2019-07-22
Upgrade dependencies, including Gradle

artembilan:
* `exclude group: 'org.springframework'` from all non-core Spring
dependencies to avoid CLASSPATH conflicts
* Remove suppress from from some modules which don't report compilation
warnings any more
* Fix deprecation warnings reported after upgrades

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlinVersion = '1.3.31'
+	ext.kotlinVersion = '1.3.40'
 	repositories {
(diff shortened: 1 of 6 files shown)


artembilan:
Related to https://build.spring.io/browse/INT-MASTERSPRING40-774/

artembilan:
Ignore, please. `Flux.transform()` deprecation warning. It is going to be restored soon enough: https://github.com/reactor/reactor-core/pull/1815

PR #3001, opened 2019-07-23
Fix TimeoutCountSequenceSizeReleaseStrategy (#2999)

### LINKED ISSUE #2999: TimeoutCountSequenceSizeReleaseStrategy does not respect timeout
**Affects Version(s): 5.1.1.RELEASE +** 

## Bug report

`TimeoutCountSequenceSizeReleaseStrategy` appears to have a bug where it will not respect the timeout when there are multiple messages in the MessageGroup.

The bug appears to have been introduced in 5.1.1.RELEASE when a null check was added.  The change causes the code to default to a timeout of `Long.MAX_VALUE` if any of the timestamp are null.  Unfortunately it introduced this bug where if any message in the group has a timestamp larger than the first it also triggers this default timeout.

The behaviour is demonstrated in this simple example.  Using spring integration 5.1.0.RELEASE it will print out a message with payload `[hello, world]`.  In 5.1.1.RELEASE and later versions it will not print out any message as the release strategy will not release the messages even though the timeout has been reached.

``` java
package com.example;

import org.springframework.boot.SpringApplication;
import org.springframework.context.ConfigurableApplicationContext;
import org.springframework.integration.aggregator.TimeoutCountSequenceSizeReleaseStrategy;
import org.springframework.integration.annotation.Gateway;
import org.springframework.integration.annotation.MessagingGateway;

@SpringBootApplication
@Configuration
public class TimeoutCountSequenceBugApplication {

	public static void main(String[] args) throws InterruptedException {
		ConfigurableApplicationContext context = SpringApplication.run(TimeoutCountSequenceBugApplication.class, args);
		Example gateway = context.getBean(Example.class);

		gateway.submit("hello");
		Thread.sleep(2000);
		gateway.submit("world");
	}

	@MessagingGateway
	public interface Example {

		@Gateway(requestChannel = "example.input")
		void submit(String s);
	}

	@Bean
	IntegrationFlow example() {
		return f -> f
			.aggregate(
				agg -> agg
				.correlationStrategy(s -> "")
				.releaseStrategy(new TimeoutCountSequenceSizeReleaseStrategy(Integer.MAX_VALUE, 1000))
			)
			.handle(msg -> System.out.println(msg));
	}

}
```

The offending code appears to be this method within the `TimeoutCountSequenceSizeReleaseStrategy` class

``` java
	private long findEarliestTimestamp(MessageGroup messages) {
		long result = Long.MAX_VALUE;
		for (Message<?> message : messages.getMessages()) {
			Long timestamp = message.getHeaders().getTimestamp();
			if (timestamp != null && timestamp < result) {
				result = timestamp;
			}
			else {
				return Long.MAX_VALUE; // can't release based on time if there is no timestamp
			}
		}
		return result;
	}
```

Simply removing the else block should solve the issue.  It would cause the code to default to `Long.MAX_VALUE` only if all messages do not have a timestamp header, rather than any of them, and return the earliest non-null timestamp otherwise.

puhlenbruck:
Fixes #2999

Remove else block from findEarliestTimestamp causing
the method to return with Long.MAX_VALUE if the condition was ever
false.

If accepted this change should also be included in 5.1.x I think, but I'm unclear on the process for that.

The added test fails before the fix but passes after.

Please let me know if anything else is needed.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/TimeoutCountSequenceSizeReleaseStrategy.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/TimeoutCountSequenceSizeReleaseStrategy.java
@@ -31,6 +31,7 @@ import org.springframework.messaging.Message;
  *
  * @author Dave Syer
  * @author Gary Russell
+ * @author Peter Uhlenbruck
  *
  * @since 2.0
  */
@@ -79,9 +80,6 @@ public class TimeoutCountSequenceSizeReleaseStrategy implements ReleaseStrategy 
 			Long timestamp = message.getHeaders().getTimestamp();
 			if (timestamp != null && timestamp < result) {
 				result = timestamp;
-			}
-			else {
-				return Long.MAX_VALUE; // can't release based on time if there is no timestamp
 			}
 		}
 		return result;
--- a/spring-integration-core/src/test/java/org/springframework/integration/aggregator/TimeoutCountSequenceSizeReleaseStrategyTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/aggregator/TimeoutCountSequenceSizeReleaseStrategyTests.java
@@ -27,6 +27,7 @@ import org.springframework.messaging.Message;
 /**
  * @author Dave Syer
  * @author Artme Bilan
+ * @author Peter Uhlenbruck
  */
 public class TimeoutCountSequenceSizeReleaseStrategyTests {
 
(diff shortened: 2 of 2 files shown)


garyrussell:
Thanks; we will cherry pick back to 5.1.x after merging.

artembilan:
@garyrussell ,

Would you mind to cherry-pick it? I'm in a middle of something else in SI, so can't switch between branches.
Thanks

garyrussell:
Sure

garyrussell:
Cherry-picked as 9ab1fc821fe5950c7029e0e47d87359a44e2217b

PR #3006, opened 2019-07-24
GH-3004: Fix MMIH argument resolution

### LINKED ISSUE #3004: InvocableHandlerMethod fails to resolve Object payload after resolving Map payload
Hello,

I faced an issue using 5.2.0.M2 release that didn't appear in 5.1.6

When having two GenericHandler implementation sequentially handling the message if the first handler is a subtype of Map the second fails with ClassCastException due to attempted use of cached MapArgumentResolver which was used for the first invocation

I nailed the code facing the issue to a small program below.

I'm not 100% certain about that but in 5.1.6 the HandlerMethodArgumentResolverComposite didn't seem to have anything at all for the second invocation after the first one

```

@SpringBootApplication
public class DemoApplication {

        SpringApplication.run(DemoApplication.class, args).getBean("inputChannel", DirectChannel.class)
                .send(new GenericMessage<>(Collections.singletonMap("key", "value")));
    }

    @Bean
    public DirectChannel inputChannel() {
        return MessageChannels.direct().get();
    }

    @Bean
    public IntegrationFlow sampleFlow() {
        return IntegrationFlows.from(inputChannel())
                .handle(new SampleMapHandler())
                .handle(new SampleHandler())
                .handle(message -> {
                }).get();
    }

    public static class SampleHandler implements GenericHandler<String> {

        @Override
        public String handle(String stringPayload, MessageHeaders messageHeaders) {
            return stringPayload + " World!";
        }
    }

    public static class SampleMapHandler implements GenericHandler<Map<String, String>> {

        @Override
        public String handle(Map<String, String> mapPayload, MessageHeaders messageHeaders) {
            return "Hello " + mapPayload.get("key");
        }
    }
}
```

garyrussell:
Fixes https://github.com/spring-projects/spring-integration/issues/3004

Change the `DefaultMessageHandlerMethodFactory` beans to prototype scope.

See https://github.com/spring-projects/spring-framework/issues/23352

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
@@ -38,6 +38,7 @@ import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.BeanDefinitionHolder;
 import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.config.PropertiesFactoryBean;
 import org.springframework.beans.factory.config.RuntimeBeanReference;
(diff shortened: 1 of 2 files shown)


garyrussell:
@olegz I think we will need to make this bean in `BinderFactoryAutoConfiguration` a prototype in spring-cloud-stream, until the SPR issue is resolved.

```java
	@Bean(IntegrationContextUtils.MESSAGE_HANDLER_FACTORY_BEAN_NAME)
	public static MessageHandlerMethodFactory messageHandlerMethodFactory(
```

olegz:
I did and also raised the issue to remove it once SPR is addressed https://github.com/spring-cloud/spring-cloud-stream/issues/1772

PR #3011, opened 2019-07-29
INT-1926: Option to disallow arbitrary routing

garyrussell:
JIRA: https://jira.spring.io/browse/INT-1926

Add an option to mapping routers to disable falling back to the
channel key as the channel name.


artembilan:
I think you mean this JIRA ticket instead: https://jira.spring.io/browse/INT-1926

garyrussell:
Doh; thanks; fixed.

PR #3012, opened 2019-07-29
Fix nullability issue in queuechannel

augur:
As per`@Nullable` annotation description, 

> Methods override should repeat parent {@code @Nullable} annotations unless they behave differently.

`QueueChannel` inherits from `AbstractPollableChannel`, which has annotation `@Nullable` on `doReceive`, but `QueueChannel` lacked it. It presents a problem in some cases, see discussion here:
https://discuss.kotlinlang.org/t/override-java-method-return-type-nullability/13633

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/QueueChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/QueueChannel.java
@@ -110,6 +110,7 @@ public class QueueChannel extends AbstractPollableChannel implements QueueChanne
 	}
 
 	@Override
+	@Nullable
 	protected Message<?> doReceive(long timeout) {
 		try {
 			if (timeout > 0) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/QueueChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/QueueChannel.java
@@ -167,7 +167,7 @@ public class QueueChannel extends AbstractPollableChannel implements QueueChanne
 			((BlockingQueue<Message<?>>) this.queue).drainTo(clearedMessages);
 		}
 		else {
-			Message<?> message = null;
+			Message<?> message;
 			while ((message = this.queue.poll()) != null) {
 				clearedMessages.add(message);
 			}


garyrussell:
I am not sure I understand - seems to be a bug in the compiler.

Are you saying the `@Nullable` is not inherited by the subclass?

I thought you only had to add `@Nullable` on the subclass if the super class has `@NonNull` and the subclass overrides the nullability.

augur:
> I thought you only had to add `@Nullable` on the subclass if the super class has `@NonNull` and the subclass overrides the nullability.

Me too, but then I read description of this annotation: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/lang/Nullable.html

> Methods override should repeat parent @Nullable annotations unless they behave differently.

Thus, if I understand correct, even if super class has this annotation, sub class also must have it, otherwise it would have no effect.

garyrussell:
Hmmm - this works fine for me...

```java
package com.example.demo;

import org.springframework.lang.Nullable;

public class Foo {

    @Nullable
    public Object getOne() {
        return "one";
    }

}

package com.example.demo;

public class Bar extends Foo {

    @Override
    public Object getOne() {
        return null;
    }

}
```

```kotlin
package kot

import com.example.demo.Bar

class Baz: Bar() {

    override fun getOne(): Any? {
        return "two"
    }

}

package kot

fun main() {
    val bar: Baz = Baz()
    val one: Any? = bar.one
    println(one)
}
```

What's the difference between this and your case?

augur:
Have you tested with `-Xjsr305=strict` compiler flag? This is default flag in Spring Initializr generated projects (https://github.com/spring-io/initializr/issues/591)
More info about flag here: https://kotlinlang.org/docs/reference/java-interop.html

garyrussell:
Interesting, with that I get the error when subclassing `QueueChannel` but there are no complaints about my Foo/Bar/Baz friends.

I'll go ahead and merge this, but I fear there are many other places in the framework that will need to be addressed 😦

garyrussell:
Cherry-picked as e5fadc3f863c7e6678b552cef9d46921588afcff to 5.1.x

artembilan:
See `@Inherited` JavaDocs:
```
* <p>Note that this meta-annotation type has no effect if the annotated
 * type is used to annotate anything other than a class.  Note also
 * that this meta-annotation only causes annotations to be inherited
 * from superclasses; annotations on implemented interfaces have no
 * effect.
```

So, this `@Nullable` on super method has no effect on the overridden method.

garyrussell:
No; I didn't mean `@Inherited` I meant that the Kotlin compiler would walk up the super classes to find the first annotation (like we do with Spring).

It does seem to silence Sonar, though.

garyrussell:
Surprisingly, we only have 12 classes with abstract methods marked with `@Nullable`.

It won't catch them all but fixing those shouldn't be too bad; I'' start tomorrow.

Of course, some of them have many, many subclasses.

PR #3023, opened 2019-08-08
AMQP OB Adapter - option to wait for confirms

garyrussell:
- add an option to block the caller until a confirm is received

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpoint.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpoint.java
@@ -16,10 +16,12 @@
 
 package org.springframework.integration.amqp.outbound;
 
+import java.time.Duration;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import org.springframework.amqp.AmqpException;
 import org.springframework.amqp.core.AmqpTemplate;
 import org.springframework.amqp.rabbit.connection.CorrelationData;
 import org.springframework.amqp.rabbit.connection.CorrelationData.Confirm;
@@ -32,7 +34,6 @@ import org.springframework.integration.MessageTimeoutException;
 import org.springframework.integration.amqp.support.MappingUtils;
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.messaging.Message;
-import org.springframework.messaging.MessageHandlingException;
 import org.springframework.util.Assert;
 
 /**
@@ -48,6 +49,8 @@ import org.springframework.util.Assert;
 public class AmqpOutboundEndpoint extends AbstractAmqpOutboundEndpoint
 		implements ConfirmCallback, ReturnCallback {
 
+	private static final Duration DEFAULT_CONFIRM_TIMEOUT = Duration.ofSeconds(5);
+
 	private final AmqpTemplate amqpTemplate;
 
 	private final RabbitTemplate rabbitTemplate;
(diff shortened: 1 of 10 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpoint.java
```diff
@@ -101,14 +124,38 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
...
+			}
 			return null;
 		}
 	}
 
+	private void waitForConfirm(Message<?> requestMessage, CorrelationData correlationData) {
+		try {
+			Confirm confirm = correlationData.getFuture().get(getConfirmTimeout().toMillis(), TimeUnit.MILLISECONDS);
+			if (!confirm.isAck()) {
+				throw new MessageHandlingException(requestMessage, "Negative publisher confirm received: " + confirm);
```
artembilan: I have fought recently about a `BeanDefinition` context in our exceptions. I think it is not going to happen in the super class automatically because you throw already `MessageHandlingException`. 
Please, consider to include `this` for presenting a `BeanDefinition` context in the exception message for better end-user experience. Or just throw non-MessageHandlingExceptions!

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests2.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.test.context.junit.jupiter.DisabledIf;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * @author Gary Russell
+ * @since 5.2
+ *
+ */
+@SpringJUnitConfig
+@RabbitAvailable(queues = "testConfirmOk")
```
artembilan: Don't we need to have a `@DirtiesContext` as well?
At least to destroy `ConnectionFactory` properly without leaking connection to RabbitMQ in cached context.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests2.java
```diff
@@ -0,0 +1,126 @@
...
+	@Test
+	void testWithReturn(@Autowired IntegrationFlow flow) {
+		assertThatThrownBy(() -> flow.getInputChannel()
+				.send(new GenericMessage<>("test", Collections.singletonMap("rk", "junkjunk"))))
+						.isInstanceOf(MessageHandlingException.class)
+						.hasMessageStartingWith("Message was returned by the broker");
+	}
+
+	@Test
+	@DisabledIf("#{systemEnvironment['TRAVIS'] ?: false}") // needs RabbitMQ 3.7
```
artembilan: Is it really a thing? I mean doesn't it really come with not RabbitMQ 3.7 yet?
Can't we adjust the test somehow not having such a condition?

Thanks

artembilan (changes requested):

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests2.java
garyrussell: Well, we have to run travis on trusty for jdk reasons, and it still has 3.6.x (I checked the version in the travis log.

I don't see how we can adjust the test because it needs a 3.7 feature; I'd rather have the test work properly on our CI.

artembilan: OK. Let's live with this then! Eventually we will come back to clean code, when Travis supports 3.7.x already.

PR #3025, opened 2019-08-12
JUnit 5 Log Level Adjuster

garyrussell:


spring-integration-test-support/src/main/java/org/springframework/integration/test/condition/LogLevels.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
```
artembilan: I don't think `2017` should be here, even if some logic has existed before in other classes.
Also I think we need to mention how do they combine together (if any) when we have a class and method level annotations.

Would be also great to mention this new feature in Docs!

spring-integration-test-support/src/main/java/org/springframework/integration/test/condition/LogLevels.java
```diff
@@ -0,0 +1,58 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Test classes annotated with this will change logging levels between tests.
```
artembilan: Looks like we can apply this for test methods as well.
Can we have that note in this JavaDocs please, too?
Thanks

artembilan (changes requested):

artembilan:
LGTM, but how about some “what’s new” entry and some short paragraph in the testing chapter? Thanks

garyrussell:
I don't really consider this for end user use, just for use in the framework tests; we've never documented it before.

PR #3028, opened 2019-08-13
Implement RotatingServerAdvice enhancements

### LINKED ISSUE #3027: Enhancements to RotatingServerAdvice
## Enhancement

dturanski:
Resolves #3027

Create `AbstractStandardRotationPolicy` subclasses implement `onRotation(MessageSource source)`.

Pull dependent inner classes from RotatingServerAdvice to break circular dependencies.


spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotationPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+	void afterReceive(boolean messageReceived, MessageSource<?> source);
+
+
+	/**
+	 *
+	 * @return the current {@link KeyDirectory}
+	 */
+	KeyDirectory getCurrent();
+
+}
```
garyrussell: Missing newline.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotationPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+	 * Invoked after the message source receive() method.
+	 * @param messageReceived true if a message was received.
+	 * @param source the message source.
+	 */
+	void afterReceive(boolean messageReceived, MessageSource<?> source);
+
+
+	/**
+	 *
+	 * @return the current {@link KeyDirectory}
```
garyrussell: I don't think this will pass checkstyle. Description needed.

garyrussell (changes requested):

garyrussell:
@dturanski Please rebase and re-push; I have fixed the .travis.yml for this branch.

garyrussell:
Merged as 03b6316d30d1c87c92dbfa00bb54ad0b3ceb2427 with minor formatting fixes.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
```diff
@@ -88,167 +84,28 @@ public boolean beforeReceive(MessageSource<?> source) {
 		return result;
 	}
 
-	/**
-	 * Implementations can reconfigure the message source before and/or after
-	 * a poll.
-	 */
-	public interface RotationPolicy {
```
artembilan: We have to revert this change. It is fully blown breaking change in the point version.
I'm OK with a decoupling in the current `5.2` (preparing some polishing PR), but I'm against any breaking changes for public API in the point release.

Does it make sense, @garyrussell , @dturanski ?

Thanks

dturanski: Technically you're right. But I think it would be ok if we default it. Then it won't break anything. I'd be surprised if anyone else has actually implemented this interface.

artembilan: We just can't guess from here. This interface is `public` in this place, therefore moving it somewhere else is a breaking change.
The first rule: don't break any public API in the point release!

dturanski: I see, because I moved it from the inner class.  Yes that is a breaking change.

artembilan: Good. So, @garyrussell , your turn now and then we will make a final decision!

garyrussell:
Yes; I had a brain fart yesterday.

But, we do need to satisfy David's extensibility needs.

artembilan:
> But, we do need to satisfy David's extensibility needs.

Right, so we leave it fixed in `5.2` and also see my polishing in the #3032

artembilan:
Reverted: https://github.com/spring-projects/spring-integration/commit/f7032c7127cc790bc974f62cea33bbce0b1bb147

PR #3029, opened 2019-08-13
Implement enhancements to RotatingServerAdvice

### LINKED ISSUE #3027: Enhancements to RotatingServerAdvice

dturanski:
Resolves #3027, see also #3028 5.1.x backport


garyrussell:
See checkstyle errors on travis.

Use `./gradlew check` to run checkstyle locally.

garyrussell:
Merged as ba017686fb3d83077fab108de7866ba9358220e5 with minor formatting fixes.

PR #3032, opened 2019-08-15
GH-3027: Decouple RotationPolicy

### LINKED ISSUE #3027: Enhancements to RotatingServerAdvice

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3027

This is a polishing for PR https://github.com/spring-projects/spring-integration/pull/3029

* Move `KeyDirectory` in to the `RotationPolicy` since it looks like fully
coupled concept with that abstraction and the class name is so bad for
the public API
* Remove `AbstractStandardRotationPolicy` in favor of a `StandardRotationPolicy`
extendability
* Change `ftp.adoc` and `sftp.adoc` to reflect a new API reality
* Mantion these changes in the `whats-new.adoc`


dturanski:
Thanks.  LGTM;

PR #3033, opened 2019-08-15
GH-3027: Rotator enhancements 5.1.x

dturanski:
Backport for #3027 

- Move MessageSource specific logic to 'onRotation()'
- Remove assertions
- Make getCurrent() public

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
@@ -171,7 +171,7 @@ public class RotatingServerAdvice extends AbstractMessageSourceAdvice {
 			return this.fair;
 		}
 
-		protected KeyDirectory getCurrent() {
+		public KeyDirectory getCurrent() {
 			return this.current;
 		}
 
@@ -201,18 +201,18 @@ public class RotatingServerAdvice extends AbstractMessageSourceAdvice {
 		}
 
 		protected void configureSource(MessageSource<?> source) {
-			Assert.isTrue(source instanceof AbstractInboundFileSynchronizingMessageSource
-							|| source instanceof AbstractRemoteFileStreamingMessageSource,
-					"source must be an AbstractInboundFileSynchronizingMessageSource or a "
-							+ "AbstractRemoteFileStreamingMessageSource");
 			if (!this.iterator.hasNext()) {
 				this.iterator = this.keyDirectories.iterator();
 			}
 			this.current = this.iterator.next();
+			onRotation(source);
+		}
+
+		protected void onRotation(MessageSource<?> source) {
 			if (source instanceof AbstractRemoteFileStreamingMessageSource) {
 				((AbstractRemoteFileStreamingMessageSource<?>) source).setRemoteDirectory(this.current.getDirectory());
 			}
-			else {
+			else if (source instanceof AbstractInboundFileSynchronizingMessageSource) {
 				((AbstractInboundFileSynchronizingMessageSource<?>) source).getSynchronizer()
 						.setRemoteDirectory(this.current.getDirectory());
 			}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
@@ -208,6 +208,15 @@ public class RotatingServerAdvice extends AbstractMessageSourceAdvice {
 			onRotation(source);
 		}
 
+		/**
+		 * Update the state of the {@link MessageSource} after the server is rotated, if necessary.
+		 * The default implementation updates the remote directory for known MessageSource implementations that require it,
+		 * specifically, instances of {@link AbstractRemoteFileStreamingMessageSource}, and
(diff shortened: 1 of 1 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
```diff
@@ -201,18 +201,18 @@ public void afterReceive(boolean messageReceived, MessageSource<?> source) {
...
-							+ "AbstractRemoteFileStreamingMessageSource");
 			if (!this.iterator.hasNext()) {
 				this.iterator = this.keyDirectories.iterator();
 			}
 			this.current = this.iterator.next();
+			onRotation(source);
+		}
+
+		protected void onRotation(MessageSource<?> source) {
 			if (source instanceof AbstractRemoteFileStreamingMessageSource) {
```
artembilan: So, how does your `onRotation()` overridden version look now?

I mean it feels like not fully gained logic when we just exit from the method silently if none condition satisfies.
We need to say something in the `else`, like `UnsupportedOperationException`. But from here it looks like full equivalent to the current `Assert.isTrue()`...

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotatingServerAdvice.java
dturanski: Currently, I don't need to override anything in`StandardRotationPolicy`.  I'm not sure why we can't exit silently.  For example, I'm applying this to a List only source, I don't need to update it's state at all.  If someone wants to use this with a different MessageSource, they can override `onRotation()` if they need to.  I don't see much advantage to throwing an exception if they forget to do it. They'll have to figure it out anyway.

artembilan: Hm... @garyrussell , WDYT?

I'm a bit reluctant to leave our public API without any signal what is going on...

dturanski: @artembilan  - If this method throws an exception because the MessageSource is not one of the types it supports then I am forced to subclass `StandardRotationPolicy` only to :

```
@Override
    protected void onRotation(MessageSource<?> source) {
       if(source instanceof AbstractInboundFileSynchronizingMessageSource
                        || source instanceof AbstractRemoteFileStreamingMessageSource) {
           super.onRotation(source);
       }
    }
```
or 

```
@Override
    protected void onRotation(MessageSource<?> source) {
       try {
           super.onRotation(source);
       } catch (IlegalArgumentException e) {
       }
    }
```
That's a work around for a leaky abstraction.  Technically it is a behavior change that in theory could break some existing code that relies on this exception being thrown but highly unlikely.  I tend to be more practical than purist.

 I'm sure there is a better design for all this, possibly using generics instead of checking types and casting,  but we are constrained by backward compatibility for 5.1.   We can revisit this for 5.2.

artembilan: I think that is right design to throw an exception from the default one and let you to override it with that simple logic you need. So, non-standard behavior is posible with an inheritance.
Not sure if we are good to revise it in the future, but as you noticed this is still a breaking change for `5.1`. so, please, consider to reinstate that `Assert.isTrue()`, please. And let's merge it finally!

Anyway, @garyrussell 's opinion would be valuable, too.

Thanks

dturanski: Why is it the right design?  Previously we could not reuse any of this for the the list only sftp source. I can respect the breaking change argument on some level, but the assertion should never have been there in the first place.

artembilan: @garyrussell , gentle ping. 
Thanks

garyrussell: How about a compromise...

```
else {
    customOnRotation(source);
}

...

protected void customOnRotation(MessageSource<?> source) {
    this.logger.error("Rotation not supported for " + source.getClass());
}
```

artembilan: Hm. The name isn't so obvious. How about `onRotationForCustomSource`. Although I still think we need to throw an exception from there to avoid breaking change.

I think David's use-case is a bit of non-standard and that is slightly wrong to pursue it from the standard implementation, meanwhile any extensions are possible.

garyrussell: >Although I still think we need to throw an exception from there to avoid breaking change.

I disagree; I think an ERROR log is sufficient.

>`onRotationForCustomSource`

Fine with me.

garyrussell: Javadocs LGTM.

artembilan (approved):
LGTM!

@garyrussell ?

garyrussell:
I already said LGTM.

artembilan:
Great! Let's wait for green Travis report and merge!

@dturanski , feel free to provide similar JavaDoc and method body fix on `master`.
Thanks

PR #3039, opened 2019-08-19
GH-3010 Add idleBetweenTries for JdbcLockRegistry

### LINKED ISSUE #3010: Ability to configure sleep Time in JdbcLockRegistry
In clustered environment, the Non leader nodes tries for lock too frequently (100ms).
Can we enhance it to be configurable property either from constructor or property value in 
_org.springframework.integration.jdbc.lock.JdbcLockRegistry_

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3010

Sometimes `100` milliseconds interval is too often to try to obtain a
lock with `UPDATE/INSERT` queries

**Cherry-pick to 5.1.x**


garyrussell:
cherry-picked to 5.1.x as c9e0cc0e0261ecc869da1cfb9ca0b3e0fcf0af19

PR #3044, opened 2019-08-23
GH-3043: Add FileHeaders.REMOTE_HOST header

### LINKED ISSUE #3043: Add (S)FTP source server information to messages
See https://stackoverflow.com/questions/57606912/spring-integration-java-dsl-sftp-how-to-get-remote-sftp-server-information-in-ha/57609407#57609407

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3043

* Populate a `FileHeaders.REMOTE_HOST` from the
`AbstractRemoteFileStreamingMessageSource` and "get"-based commands
in the `AbstractRemoteFileOutboundGateway`
* Extract the value from the a `Session.getHost()` contract
* The `AbstractInboundFileSynchronizingMessageSource` cannot be
addressed with this because the real message is already based on the
locally stored file
* Adjust some affected tests according our code style requirements

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileHeaders.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileHeaders.java
@@ -55,6 +55,6 @@ public abstract class FileHeaders {
 	/**
 	 * A remote host/port the file has been polled from
 	 */
-	public static final String REMOTE_HOST = PREFIX + "remoteHost";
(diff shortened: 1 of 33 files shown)


artembilan:
Will add Docs after review.
Thanks

garyrussell:
>real message is already based on the locally stored file

How about an (optional) `MetadataStore` to hold the information?

garyrussell (changes requested):

garyrussell (approved):

PR #3046, opened 2019-08-28
GH-3038: Delegate more RemoteFileTempalte options

### LINKED ISSUE #3038: Improve AbstractRemoteFileOutboundGateway
@Bean
		return IntegrationFlows.from(UploadGateway.class)
				.handle(Sftp.outboundGateway(template(), "put", "headers['file_name']"))
				.get();
	}

	@Bean
	public SftpRemoteFileTemplate template() {
		SftpRemoteFileTemplate template = new SftpRemoteFileTemplate(sessionFactory());
		template.setRemoteDirectoryExpression(new LiteralExpression("foo/test"));
		return template;
	}

Works fine.

But there is no way to configure the `remoteDirectoryExpression` if you use this CTOR

    .handle(Sftp.outboundGateway(sessionFactory(), "put", "headers['file_name']"))

>Caused by: java.lang.IllegalArgumentException: 'remoteDirectoryExpression' is required

We should either deprecate those CTORs or add support to the ARFOG/DSL to set the RDE.

Also, we should add `...ExpressionString()` variants for the `RemoteFIleTemplate` properties.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3038

* Add more delegating setters into the `AbstractRemoteFileOutboundGateway`
for its `RemoteFileTemplate` property
* Restrict externally provided `RemoteFileTemplate` from modifications
in the `AbstractRemoteFileOutboundGateway`: those options must be
configured on that external `RemoteFileTemplate`
* Expose new options in the Java DSL specs

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
@@ -25,6 +25,7 @@ import org.springframework.expression.Expression;
 import org.springframework.integration.dsl.ComponentsRegistration;
 import org.springframework.integration.dsl.MessageHandlerSpec;
 import org.springframework.integration.expression.FunctionExpression;
+import org.springframework.integration.file.FileNameGenerator;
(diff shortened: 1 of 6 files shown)


garyrussell (commented):
Good stuff! Just one question.

PR #3049, opened 2019-08-30
GH-3041: Deprecate RedisUtils.isUnlinkAvailable()

### LINKED ISSUE #3041: Introduction of unlink with lock broke integrations with Twemproxy
## Bug report / Enhancement

We use spring-integration-redis to use twemproxy or T-Redis. (T-Redis is a sharding mechanism for Redis). The library worked great until the introduction of the following diff. 

https://github.com/spring-projects/spring-integration/commit/88c7646ed9eaea2fe755485e1676ad130eb391a9#diff-befc3689f434cc94881bf3d1818f9191

Once this diff was released it causes an exception to be thrown when you issue an lock.unlock() call. This happens because twemproxy does not support the call `info server`.

The last working version was 5.0.1. All of the more recent versions are broken. 

Is this something you support / are willing to fix?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3041

Some Redis clients/servers don't allow to perform an `INFO` command,
therefore we are not able to determine if we can perform `UNLINK` or
not.

* Deprecate `RedisUtils.isUnlinkAvailable()` as not reliable source of
through; use trial with fallback algorithm in the target logic around
`UNLINK` command calls.

**Cherry-pick to 5.1.x**

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
@@ -26,7 +26,6 @@ import org.springframework.data.redis.serializer.JdkSerializationRedisSerializer
 import org.springframework.data.redis.serializer.RedisSerializer;
 import org.springframework.data.redis.serializer.SerializationException;
 import org.springframework.data.redis.serializer.StringRedisSerializer;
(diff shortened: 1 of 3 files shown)


garyrussell:
Merged as 3dc0fb5980 with minor English polishing; cherry-picked as f7e0848bfc51bb220b0d6fb5e6764190364120b4

PR #3050, opened 2019-08-30
GH-3047: Add method and args headers into gateway

### LINKED ISSUE #3047: Support header setting on gateways that initiate a DSL flow
```
@MessagingGateway(defaultHeaders = @GatewayHeader(name = "method", expression = "#gatewayMethod.name"))
interface Gate {

	void method1(String foo);

	void method2(String foo);

}
```

Doesn't work - 

>Caused by: java.lang.IllegalStateException: send is not supported, because no request channel has been configured

We need some mechanism to set headers so we can do something like...

	@Bean
		return IntegrationFlows.from(Gate.class)
				.route(...)
				.get();
	}

See https://stackoverflow.com/questions/57709118/what-is-analog-of-intgateway-xml-tag-in-java-dsl-in-spring-integration/57710401#57710401

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3047

In some use-cases we can't use `headerExpressions` for messaging gateway
interface, e.g. when we use third-party class and don't use XML, but
Java DSL.

* Always add `gatewayMethod` and `gatewayArgs` headers to messages
produced by the gateway calls.
This information may be useful for downstream routing or analyzes
* Demonstrate a routing approach in the Docs with Java DSL.
Expose convenient methods to access those headers in the
`StaticMessageHeaderAccessor`
* Improve `GatewayProxyFactoryBean` to determine the return type of the
method call from the interface generic types, when the `serviceInterface`
is a `java.util.function.Function`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
@@ -123,7 +123,7 @@ public final class StaticMessageHeaderAccessor {
 	 */
 	@Nullable
 	public static Method getGatewayMethod(Message<?> message) {
-		return message.getHeaders().get("gatewayMethod", Method.class);
+		return message.getHeaders().get(IntegrationMessageHeaderAccessor.GATEWAY_METHOD, Method.class);
 	}
 
 	/**
@@ -135,7 +135,7 @@ public final class StaticMessageHeaderAccessor {
 	 */
 	@Nullable
 	public static Object[] getGatewayArguments(Message<?> message) {
-		return message.getHeaders().get("gatewayArgs", Object[].class);
+		return message.getHeaders().get(IntegrationMessageHeaderAccessor.GATEWAY_ARGS, Object[].class);
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
@@ -17,7 +17,6 @@
 package org.springframework.integration;
 
 import java.io.Closeable;
-import java.lang.reflect.Method;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -112,30 +111,6 @@ public final class StaticMessageHeaderAccessor {
 	@Nullable
 	public static <T> T getSourceData(Message<?> message) {
 		return (T) message.getHeaders().get(IntegrationMessageHeaderAccessor.SOURCE_DATA);
-	}
-
-	/**
-	 * Return the method called from the messaging gateway interface or {@code null}
-	 * from the headers in the provided message.
-	 * @param message the message to extract gateway method header.
-	 * @return the method called from the messaging gateway interface.
-	 * @since 5.2
-	 */
-	@Nullable
-	public static Method getGatewayMethod(Message<?> message) {
-		return message.getHeaders().get(IntegrationMessageHeaderAccessor.GATEWAY_METHOD, Method.class);
-	}
-
-	/**
-	 * Return the the messaging gateway method invocation arguments array or {@code null}
-	 * from the headers in the provided message.
-	 * @param message the message to extract gateway method arguments header.
-	 * @return the gateway method  invocation arguments.
-	 * @since 5.2
-	 */
-	@Nullable
-	public static Object[] getGatewayArguments(Message<?> message) {
-		return message.getHeaders().get(IntegrationMessageHeaderAccessor.GATEWAY_ARGS, Object[].class);
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
@@ -17,6 +17,7 @@
 package org.springframework.integration;
 
 import java.io.Closeable;
+import java.lang.reflect.Method;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -111,6 +112,30 @@ public final class StaticMessageHeaderAccessor {
 	@Nullable
 	public static <T> T getSourceData(Message<?> message) {
 		return (T) message.getHeaders().get(IntegrationMessageHeaderAccessor.SOURCE_DATA);
+	}
+
+	/**
+	 * Return the method called from the messaging gateway interface or {@code null}
+	 * from the headers in the provided message.
+	 * @param message the message to extract gateway method header.
+	 * @return the method called from the messaging gateway interface.
+	 * @since 5.2
+	 */
+	@Nullable
+	public static Method getGatewayMethod(Message<?> message) {
+		return message.getHeaders().get("gatewayMethod", Method.class);
+	}
+
+	/**
+	 * Return the the messaging gateway method invocation arguments array or {@code null}
+	 * from the headers in the provided message.
+	 * @param message the message to extract gateway method arguments header.
+	 * @return the gateway method  invocation arguments.
+	 * @since 5.2
+	 */
+	@Nullable
+	public static Object[] getGatewayArguments(Message<?> message) {
+		return message.getHeaders().get("gatewayArgs", Object[].class);
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingGatewayRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingGatewayRegistrar.java
@@ -84,7 +84,6 @@ public class MessagingGatewayRegistrar implements ImportBeanDefinitionRegistrar 
 		String errorChannel = (String) gatewayAttributes.get("errorChannel");
 		String asyncExecutor = (String) gatewayAttributes.get("asyncExecutor");
 		String mapper = (String) gatewayAttributes.get("mapper");
-		String mapInternalHeaders = (String) gatewayAttributes.get("mapInternalHeaders");
 
 		boolean hasMapper = StringUtils.hasText(mapper);
 		boolean hasDefaultPayloadExpression = StringUtils.hasText(defaultPayloadExpression);
@@ -95,19 +94,18 @@ public class MessagingGatewayRegistrar implements ImportBeanDefinitionRegistrar 
 		Assert.state(!hasMapper || !hasDefaultHeaders,
 				"'defaultHeaders' are not allowed when a 'mapper' is provided");
 
-		boolean hasMapInternalHeaders = !ObjectUtils.isEmpty(mapInternalHeaders);
-		Assert.state(!hasMapper || !hasMapInternalHeaders,
-				"'mapInternalHeaders' are not allowed when a 'mapper' is provided");
-
 		BeanDefinitionBuilder gatewayProxyBuilder =
 				BeanDefinitionBuilder.genericBeanDefinition(GatewayProxyFactoryBean.class);
 
-		if (hasDefaultHeaders || hasDefaultPayloadExpression || hasMapInternalHeaders) {
+		if (hasDefaultHeaders || hasDefaultPayloadExpression) {
 			BeanDefinitionBuilder methodMetadataBuilder =
 					BeanDefinitionBuilder.genericBeanDefinition(GatewayMethodMetadata.class);
 
 			if (hasDefaultPayloadExpression) {
-				methodMetadataBuilder.addPropertyValue("payloadExpression", defaultPayloadExpression);
+				methodMetadataBuilder.addPropertyValue("payloadExpression",
+						BeanDefinitionBuilder.genericBeanDefinition(ExpressionFactoryBean.class)
+								.addConstructorArgValue(defaultPayloadExpression)
+								.getBeanDefinition());
 			}
 
 			if (hasDefaultHeaders) {
@@ -130,10 +128,6 @@ public class MessagingGatewayRegistrar implements ImportBeanDefinitionRegistrar 
 					headerExpressions.put((String) header.get("name"), expressionDef);
 				}
 				methodMetadataBuilder.addPropertyValue("headerExpressions", headerExpressions);
-			}
-
-			if (hasMapInternalHeaders) {
-				methodMetadataBuilder.addPropertyValue("mapInternalHeaders", mapInternalHeaders);
 			}
 
 			gatewayProxyBuilder.addPropertyValue("globalMethodMetadata", methodMetadataBuilder.getBeanDefinition());
@@ -190,7 +184,9 @@ public class MessagingGatewayRegistrar implements ImportBeanDefinitionRegistrar 
 	 * @param importingClassMetadata The importing class metadata
 	 * @return The captured values.
 	 */
-	private List<MultiValueMap<String, Object>> captureMetaAnnotationValues(AnnotationMetadata importingClassMetadata) {
+	private  static List<MultiValueMap<String, Object>> captureMetaAnnotationValues(
+			AnnotationMetadata importingClassMetadata) {
+
 		Set<String> directAnnotations = importingClassMetadata.getAnnotationTypes();
 		List<MultiValueMap<String, Object>> valuesHierarchy = new ArrayList<>();
 		// Need to grab the values now; see SPR-11710
@@ -212,8 +208,9 @@ public class MessagingGatewayRegistrar implements ImportBeanDefinitionRegistrar 
 	 * @param valuesHierarchy The values hierarchy in order.
 	 * @param annotationAttributes The current attribute values.
 	 */
-	private void replaceEmptyOverrides(List<MultiValueMap<String, Object>> valuesHierarchy,
+	private static void replaceEmptyOverrides(List<MultiValueMap<String, Object>> valuesHierarchy,
 			Map<String, Object> annotationAttributes) {
+
 		for (Entry<String, Object> entry : annotationAttributes.entrySet()) {
 			Object value = entry.getValue();
 			if (!MessagingAnnotationUtils.hasValue(value)) {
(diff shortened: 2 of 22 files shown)


spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
```diff
@@ -113,4 +114,28 @@ public static AcknowledgmentCallback getAcknowledgmentCallback(Message<?> messag
...
+	/**
+	 * Return the method called from the messaging gateway interface or {@code null}
+	 * from the headers in the provided message.
+	 * @param message the message to extract gateway method header.
+	 * @return the method called from the messaging gateway interface.
+	 * @since 5.2
+	 */
+	@Nullable
+	public static Method getGatewayMethod(Message<?> message) {
+		return message.getHeaders().get("gatewayMethod", Method.class);
```
garyrussell: Constant for these two headers?

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -484,13 +486,17 @@ protected Object doInvoke(MethodInvocation invocation, boolean runningOnCallerTh
...
 		if (!this.initialized) {
-			this.afterPropertiesSet();
+			afterPropertiesSet();
 		}
 		Method method = invocation.getMethod();
 		MethodInvocationGateway gateway = this.gatewayMap.get(method);
 		Class<?> returnType = method.getReturnType();
-		boolean shouldReturnMessage = Message.class.isAssignableFrom(returnType)
-				|| hasReturnParameterizedWithMessage(method, runningOnCallerThread);
+		if (gateway.isReturnTypeMessage == null) {
```
garyrussell: Package protected to avoid synthetic access method; but why is this a field there anyway - it's only used here.

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -821,7 +828,7 @@ protected void doStop() {
 		}
 	}
 
-	private static boolean hasReturnParameterizedWithMessage(Method method, boolean runningOnCallerThread) {
```
garyrussell: Any reason why?

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodInboundMessageMapper.java
```diff
@@ -298,6 +297,10 @@ private static Expression parsePayloadExpression(Method method) {
 					headersToMap != null
 							? new HashMap<>(headersToMap)
 							: new HashMap<>();
+
+			headersToPopulate.put("gatewayMethod", holder.getMethod());
+			headersToPopulate.put("gatewayArgs", holder.getArgs());
```
garyrussell: We should probably add a mechanism to suppress these so the previous behavior can be retained if needed, especially for "classic" (non-DSL) gateways.

e.g. integration property `spring.integration.gateway.suppress.default.headers.interfaces=`

src/reference/asciidoc/gateway.adoc
```diff
@@ -175,6 +175,33 @@ However, default headers do NOT override any `@Header` annotations in the servic
 
 The gateway now also supports a `default-payload-expression`, which is applied for all methods (unless overridden).
 
+Starting with version 5.2, the invoked gateway `Method` is always presented as a `gatewayMethod` message header, alongside with the `gatewayArgs` header for the `Object[]` with actually method invocation arguments.
```
garyrussell: s/presented/added/
s/alongside.../along with the `gatewayArgs` header with the value being the actual method invocation arguments in an `Object[]`./

src/reference/asciidoc/gateway.adoc
```diff
@@ -175,6 +175,33 @@ However, default headers do NOT override any `@Header` annotations in the servic
...
+This information is useful, when we can't declare messaging gateway annotations on the third party interface or XML is not appropriate configuration model.
```
garyrussell: on a third party ... and XML is not an appropriate

garyrussell (changes requested):
Some things to consider.

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
artembilan: Correct. I just don't see reason to have an overhead parsing generics every single method call.
So, this is lazy loaded, but cached static state for the method. It is going to be definitely the same for each method invocation.

Do you mean that `private` field access is always wrapped to some generated setter call?

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
artembilan: Yeah... I wanted to move `isReturnTypeMessage` logic into the `MethodInvocationGateway`, but looks like `runningOnCallerThread` is an obstacle, so I've decided to go the lazy-load approach.
Will reinstate this `static` in the next commit.

Thanks for catching this!

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodInboundMessageMapper.java
artembilan: ??? This is just about two extra header in the message. There is no any harm in logic. More over you are going to simplify many places in your config with those extra configurations for `method` header and so on.
It is already there, so config is simpler ever if that is non-DSL one.
This is what we agreed in private discussion. 
In other places we populate source-based headers unconditionally, so what is the problem here?
If you don't need then downsrteam, you can just `filter` them!

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
garyrussell: Ah - makes sense, yes.

Yes, with a private field, the compiler has to generate synthetic static getters and setters; with package protected it uses simple GETFIELD and PUTFIELD operators.

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodInboundMessageMapper.java
garyrussell: My real concern is adding overhead to serialization with large arguments where the serialization may be so far downstream that these arguments are no longer relevant. Of course `Method` is not `Serializable` so that won't be an issue.

Another option would be to set a boolean on the GPFB in `IntegrationFlows.from(Class<?> serviceInterface, String beanName)` so we don't get these for "legacy" gateways.

artembilan: OK. How about a `boolean mapInternalHeaders` option on a `MessagingGateway` and `<gateway>` with `false` by default?
It is indeed going to be `true` for `IntegrationFlows.from(Class<?> serviceInterface, String beanName)`, since this is only the place where we may not have access to annotation.

artembilan:
Noticed funny thing. We don't need this:
```
private StandardEvaluationContext createMethodInvocationEvaluationContext(Object[] arguments) {
	StandardEvaluationContext context = ExpressionUtils.createStandardEvaluationContext(this.beanFactory);
	context.setVariable("args", arguments);
	context.setVariable("gatewayMethod", this.method);
	return context;
}
```

The `MethodArgsHolder` can be just propagated as a `rootObject`. So, we are going to have a single `EvaluationContext`. Better for performance again!
But I guess we need to stay with "per-call" until `5.3`: even if we deprecate those SpEL variabled, we can't remove then now for backward compatibility. 😢

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodInboundMessageMapper.java
garyrussell: Sounds good.

garyrussell:
>`MethodArgsHolder`

Yes, that would be a good improvement.

spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
```diff
@@ -113,4 +114,28 @@ public static AcknowledgmentCallback getAcknowledgmentCallback(Message<?> messag
...
+	 * Return the method called from the messaging gateway interface or {@code null}
+	 * from the headers in the provided message.
+	 * @param message the message to extract gateway method header.
+	 * @return the method called from the messaging gateway interface.
+	 * @since 5.2
+	 */
+	@Nullable
+	public static Method getGatewayMethod(Message<?> message) {
+		return message.getHeaders().get(IntegrationMessageHeaderAccessor.GATEWAY_METHOD, Method.class);
+	}
```
olegz: I am wondering how important it is to store reference to the actual method? Reading the docs it appears that the intention is to _know the method name_. Plus having reference to a method exposes the possibility that someone may actually invoke it and consequences and so on. . .

garyrussell:
One more thought... 

     .from(Class<?> serviceInterface,  Gatewayheaderspec headerSpec)

Where the headerSpec supplies header definitions like we do with XML and annotations?

I think that would be more consistent.

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodMetadata.java
```diff
@@ -60,7 +65,9 @@ public void setPayloadExpression(String payloadExpression) {
 	}
 
 	public void setHeaderExpressions(Map<String, Expression> headerExpressions) {
-		this.headerExpressions = headerExpressions;
+		Assert.notNull(headerExpressions, "'headerExpressions' must not be null");
```
olegz: Do we really care about it being null or should we simply `clear()` and skip `putAll` if null?

spring-integration-core/src/main/resources/org/springframework/integration/config/spring-integration-5.2.xsd
```diff
@@ -821,14 +821,26 @@
 					</xsd:appinfo>
 				</xsd:annotation>
 			</xsd:attribute>
+			<xsd:attribute name="map-internal-headers" default="false">
+				<xsd:annotation>
+					<xsd:documentation>
+						The flag to control the mapping of internal headers (e.g. 'gatewayMethod', 'gatewayArgs')
+						They are not mapped by default to avoid performance overhead in the downstream processing and/or
+						(de)serialization, when 'gatewayArgs' might be as big objects.
```
olegz: `... when 'gatewayArgs' might represent big objects.`

spring-integration-core/src/main/java/org/springframework/integration/StaticMessageHeaderAccessor.java
garyrussell: The problem there is you can have overloaded methods with the same name and different return types. This was a problem we fixed long ago with the #gatewayMethod SpEL variable.I had similar concerns back then but we decided the benefits outweigh the risks.

artembilan:
OK. After debating with myself a little bit I agree with Gary:

>  .from(Class<?> serviceInterface,  Gatewayheaderspec headerSpec)

So, let's take headers back, but allow end-users to configure them manually!
Even if original idea was to add a "source of data" into headers like we do in (S)FTP, HTTP, AMQP, Kafka etc., but since nobody is happy to have all the `args` in headers unconditionally, we can stay with the current behavior, but also expand it into DSL as well.

Thank you all for the feedback!
Stay tuned and I hope the solution will be there for tomorrow's release, so you will be able to review it one more time in the morning 😄

artembilan:
So, here you are a new vision based on the `GatewayProxySpec` instead of an implicit, suspicious headers.

I'll polish Docs after your review.

Thank you!

spring-integration-core/src/main/java/org/springframework/integration/config/MessagingGatewayRegistrar.java
```diff
@@ -181,9 +184,11 @@ else if (StringUtils.hasText(asyncExecutor)) {
 	 * @param importingClassMetadata The importing class metadata
 	 * @return The captured values.
 	 */
-	private List<MultiValueMap<String, Object>> captureMetaAnnotationValues(AnnotationMetadata importingClassMetadata) {
+	private  static List<MultiValueMap<String, Object>> captureMetaAnnotationValues(
```
garyrussell: Double space

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodMetadata.java
```diff
@@ -30,37 +31,42 @@
...
+	private String replyTimeout;
 
-	public String getPayloadExpression() {
+	@Nullable
+	public Expression getPayloadExpression() {
 		return this.payloadExpression;
 	}
 
-	public void setPayloadExpression(String payloadExpression) {
+	public void setPayloadExpression(Expression payloadExpression) {
```
garyrussell: Breaking change?

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -563,7 +572,8 @@ private Object sendOrSendAndReceive(MethodInvocation invocation, MethodInvocatio
 		return response;
 	}
 
-	private void rethrowExceptionCauseIfPossible(Throwable originalException, Method method) throws Throwable { // NOSONAR
+	private void rethrowExceptionCauseIfPossible(Throwable originalException, Method method)
+			throws Throwable { // NOSONAR
 		Class<?>[] exceptionTypes = method.getExceptionTypes();
```
garyrussell: Newline before

garyrussell (changes requested):
Minor issues, except possible breaking change.

garyrussell:
Perhaps we should drop the `default` prefix since there are no method-level overrides in this context; or, add a TODO if we think we might add method level overrides in the future if the community requests it.

artembilan:
> I may drag in other people for some clarifications.

So, already then we rename them back with `default` 😄

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMethodMetadata.java
artembilan: I realized that, but this is about a new `5.2` generation and there is a little chance that this class is used somewhere outside of the framework.
Plus a workaround is easy as well.
Also we can't have overloaded version with `String` because XML parser with its Java Beans introspection is going to fail.

garyrussell: Ok. Agree.

garyrussell:
I don't see that text you are quoting so I don't understand your comment. I am on my phone.

artembilan:
Sorry, I didn't notice that I hadn't copy/pasted your comment.
My point is that I agree to rename them now without `default`.

Stay tuned. Will come back soon with commit...

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -842,11 +853,24 @@ private static boolean hasReturnParameterizedWithMessage(Method method, boolean
 		return false;
 	}
 
+	private boolean hasReturnMessageTypeOnFunction(Method method) {
+		if (Function.class.isAssignableFrom(this.serviceInterface) && "apply".equals(method.getName())) {
```
olegz: Wondering if NOT `&& "apply".equals(method.getName())` should really trigger the exception instead of simply returning `false`? I mean that would be some gross miss-configuration, right?

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -842,11 +853,24 @@ private static boolean hasReturnParameterizedWithMessage(Method method, boolean
...
+			Class<?> returnType =
+					ResolvableType.forClass(Function.class, this.serviceInterface)
+							.getGeneric(1)
+							.getRawClass();
+			return returnType != null && Message.class.isAssignableFrom(returnType);
```
olegz: Again, how would it be possible to have `returnType` null on the Function.apply()?

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
artembilan: Your interface may `extends Function`, but have more methods declared.
Why is that so invalid?

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
artembilan: This is syntax sugar to make our IDE and Sonar happy.
this is just because `ResolvableType.getRawClass()` is `@Nullable`, so sending `null` as argument into `isAssignableFrom()` is bad practice.

artembilan:
Pushed Docs and removed `default` from the `GatewayProxySpec`.

Thank you for your review!

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
olegz: What I am saying is that at that moment it is really not a Function. What I mean is that you may extend and have some additional methods, but in the Functional world the only method I know is `apply` and that is my entry point if I down-cast everything to Function.
It may not be a problem in your context, but i wanted to bring it up to ensure you thought of that.

olegz:
+1 to merge!

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
artembilan: OK. Thanks for your feedback, but for extra safety I would stay with this extra protection.

garyrussell (approved):

PR #3058, opened 2019-09-20
Add failure counter to AbstractMessageSource

garyrussell:


spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractMessageSource.java
```diff
@@ -234,9 +253,8 @@ private void incrementReceiveCounter() {
 
 	@Override
 	public void destroy() {
-		if (this.receiveCounter != null) {
-			this.receiveCounter.remove();
-		}
+		this.meters.forEach(MeterFacade::remove);
+		this.meters.clear();
```
artembilan: Hm. Looks like we are missing to clear in other places, e.g. `AbstractMessageHandler` 

Would you mind to address that as well with this your PR?

Thanks

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractMessageSource.java
garyrussell: Done; but not critical, I think, since the bean is being destroyed anyway 😄

PR #3059, opened 2019-09-20
Document RSocket support

artembilan:
* Fix some typos in `webflux.adoc` and `dsl.adoc`
* Make `FunctionsTests.kt` more Kotlin friendly
* Improve RSocket components and tests for them,
especially rely on the default `RSocketStrategies` in the
target SF RSocket components
* Rework `ServerRSocketConnector.setClientRSocketKeyStrategy` to use the whole `MessageHeaders` for consultation.
It turns out that just destination is not enough since it can be used from different clients
* Make the key based on the provided `setupData` by default
* Include all the `MessageHeaders` into the `RSocketConnectedEvent`

### DIFF
--- a/src/reference/asciidoc/rsocket.adoc
+++ b/src/reference/asciidoc/rsocket.adoc
@@ -24,14 +24,15 @@ compile "org.springframework.integration:spring-integration-rsocket:{project-ver
 ====
 
 This module is available starting version 5.2 and is fully based on the Spring Messaging foundation with its RSocket components implementation, such as `RSocketRequester`, `RSocketMessageHandler` and `RSocketStrategies`.
+See https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html#rsocket[Spring Framework RSocket Support] for more information about RSocket protocol, terminology and components.
 
 Before starting an integration flow processing via channel adapters we need to establish RSocket connection between server and client.
 For this purpose Spring Integration RSocket support provides a `ServerRSocketConnector` and `ClientRSocketConnector` implementations of the `AbstractRSocketConnector`.
 
 The `ServerRSocketConnector` exposes a listener on the host and port according to provided `io.rsocket.transport.ServerTransport` for accepting connections from clients.
 An internal `RSocketFactory.ServerRSocketFactory` instance can be customized with the `setFactoryConfigurer()`, as well as other options can be configured, e.g. `RSocketStrategies` and `MimeType` for payload data and headers metadata.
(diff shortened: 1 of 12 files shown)


src/reference/asciidoc/rsocket.adoc
```diff
@@ -0,0 +1,327 @@
...
+[source, java]
+----
+@Bean
+public RSocketStrategies rsocketStrategies() {
+    return RSocketStrategies.builder()
+        .decoder(StringDecoder.allMimeTypes())
+        .encoder(CharSequenceEncoder.allMimeTypes())
+        .dataBufferFactory(new NettyDataBufferFactory(PooledByteBufAllocator.DEFAULT))
+        .build();
+}
```
rstoyanchev: Note that these are `RSocketStrategies` defaults now to register basic codecs and use pooled buffers. So the above is equivalent to `RSocketStrategies.create()` or `RSocketStrategies.builder.build()`.

src/reference/asciidoc/rsocket.adoc
```diff
@@ -0,0 +1,327 @@
...
+====
+[source, java]
+----
+@Bean
+public RSocketStrategies rsocketStrategies() {
+    return RSocketStrategies.builder()
+        .decoder(StringDecoder.allMimeTypes())
+        .encoder(CharSequenceEncoder.allMimeTypes())
+        .dataBufferFactory(new NettyDataBufferFactory(PooledByteBufAllocator.DEFAULT))
+        .build();
```
rstoyanchev: Same comment as above.

src/reference/asciidoc/rsocket.adoc
artembilan: Yeah... I've noticed that during refactoring in the main code, but forgot to change this snippet to something what is not default.
Thank you for the pointer, @rstoyanchev !

artembilan:
Pushed some changes according SF Docs: https://docs.spring.io/spring/docs/5.2.0.BUILD-SNAPSHOT/spring-framework-reference/web-reactive.html#rsocket-annot-responders-server

The link is pointing now to the `current` which is based on the latest GA, so broken at the moment, but should be OK whenever SF 5.2 is GA.

Thanks

garyrussell:
Minor doc polishing here https://github.com/garyrussell/spring-integration/commit/8d80280f6d1fbfc99775037a443696eb413f6d37 for your review.

artembilan:
Thank you, Gary, for polishing!
You can merge it with that.

garyrussell:
Merged as b12657c599dde967f3b05cde0c2421791080f3f2

PR #3060, opened 2019-09-23
Initial Kotlin DSL

artembilan:
Kotlin is good with reifying generics types into the target `Class`
arguments.

* Introduce an `IntegrationFlowBuilderExtensions.kt` and implement inline
functions for reifying types for the `IntegrationFlowBuilder.convert()`
operator
* Add `KotlinDslTests.kt` to demonstrate a new Kotlin API
* Move Kotlin dependencies to `compile` scope

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -170,6 +170,13 @@ subprojects { subproject ->
 		targetCompatibility = 1.8
 	}
 
+	compileKotlin {
+		kotlinOptions {
+			jvmTarget = '1.8'
+			freeCompilerArgs = ["-Xjsr305=strict"]
+			allWarningsAsErrors = true
+		}
+	}
 	compileTestKotlin {
 		kotlinOptions {
 			freeCompilerArgs = ['-Xjsr305=strict']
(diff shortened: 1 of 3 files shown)


artembilan:
We can merge this or can keep it until the next version.
Definitely there are many other methods in the `IntegrationFlowBuilder` to reify.

Also we still need to think what to do with the ` IntegrationFlows.from(CharacterStreamReadingMessageSource.stdin()) { e -> e.poller(Pollers.fixedDelay(1000)) }`. Kotlin can't inline for `static` methods.
Although we can introduce something like `integrationFlow {}` singleton function to cover all the incompatibility.

Also from here I start to think that we don't need that `IntegrationFlowDefinition` and just have everything in the `IntegrationFlowBuilder`. 
No reason to support intermediate object just to hide a `get()` 😄 

It would make Kotlin API to support much shorter.

Need to think more, though...

garyrussell:
Perhaps we should create it as an extension for 5.2 since it sounds like it might need a faster release cadence than core.

artembilan:
>  as an extension

Well, these Kotlin additions for our classes must be in the same package as original classes.
Are you OK with such an inconsistency in the separate dependency?

artembilan:
Closed in favor of: https://github.com/spring-projects/spring-integration-extensions/issues/218

PR #3068, opened 2019-09-27
GH-3067: Use default NullChannel instance if poss.

### LINKED ISSUE #3067: Revise new NullChannel() usage in favor of `nullChannel` bean
The problem with a `new NullChannel()` that we won't get a consistent metrics on the `nullChannel` bean.

Also consider to change `successTimer` in the `NullChannel` into the `sendCounter` since we don't care about timing on the `nullChannel`, but rather how many messages have been discarded.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3067

Also add missing receive counter.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -30,18 +30,14 @@ import java.util.concurrent.locks.Lock;
 import org.aopalliance.aop.Advice;
 
 import org.springframework.aop.framework.ProxyFactory;
-import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
-import org.springframework.beans.factory.SmartInitializingSingleton;
-import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.context.Lifecycle;
 import org.springframework.expression.EvaluationContext;
 import org.springframework.expression.Expression;
 import org.springframework.integration.IntegrationMessageHeaderAccessor;
-import org.springframework.integration.channel.NullChannel;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.expression.ExpressionUtils;
 import org.springframework.integration.handler.AbstractMessageProducingHandler;
@@ -95,8 +91,7 @@ import org.springframework.util.CollectionUtils;
  * @since 2.0
  */
 public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageProducingHandler
-		implements DiscardingMessageHandler, ApplicationEventPublisherAware, Lifecycle,
-				SmartInitializingSingleton {
+		implements DiscardingMessageHandler, ApplicationEventPublisherAware, Lifecycle {
 
 	private final Comparator<Message<?>> sequenceNumberComparator = new MessageSequenceComparator();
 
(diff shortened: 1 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -357,6 +358,22 @@ protected void onInit() {
 		this.forceReleaseProcessor = createGroupTimeoutProcessor();
 	}
 
+	@Override
+	public void afterSingletonsInstantiated() {
```
artembilan: Let's stay away from `afterSingletonsInstantiated()` because it is not called from dynamic integration flows.

In other places we use `getDiscardChannel()` to resolve it lazily.

spring-integration-core/src/main/java/org/springframework/integration/transaction/ExpressionEvaluatingTransactionSynchronizationProcessor.java
```diff
@@ -116,17 +115,23 @@ protected void onInit() {
 
 	@Override
 	public void processBeforeCommit(IntegrationResourceHolder holder) {
-		doProcess(holder, this.beforeCommitExpression, this.beforeCommitChannel, "beforeCommit");
+		if (this.beforeCommitChannel != null) {
```
artembilan: I don't this this is right solution.

We don't need to to build and send a message only when there is no particular channel.

artembilan (changes requested):
Please, fix a couple Checkstyle violations:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/test/java/org/springframework/integration/endpoint/PseudoTransactionalMessageSourceTests.java:99: Line has leading space characters; indentation should be performed with tabs only. [RegexpSinglelineJava]

[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/test/java/org/springframework/integration/endpoint/PseudoTransactionalMessageSourceTests.java:100: Line has leading space characters; indentation should be performed with tabs only. [RegexpSinglelineJava]
```

See Travis report.

spring-integration-core/src/main/java/org/springframework/integration/transaction/ExpressionEvaluatingTransactionSynchronizationProcessor.java
garyrussell: Sorry, I don't understand what you mean.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
garyrussell: I had a bunch of failing tests when it was in `onInit()` - I will take another look.

spring-integration-core/src/main/java/org/springframework/integration/transaction/ExpressionEvaluatingTransactionSynchronizationProcessor.java
garyrussell: Oh, I get it; sorry.

artembilan: Sorry, I meant: "I don't think".
My point was to call that `doProcess()` method don't build a reply message and don't send anything if there is no channel.
Not sure why would one build a message just to send it into a `NullChannel`...

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
artembilan: No, that must not be in the `onInit()`. We need to introduce a `getDiscardChannel()`  with a similar lazy-init logic like we have in the `getOutputChannel()`

PR #3073, opened 2019-10-04
GH-3072: j.u.f.Suppliers for inbound queries

### LINKED ISSUE #3072: Feature Enhancement: JdbcPollingChannelAdapter to have configurable SQL
The Spring Integration JdbcPollingChannelAdapter jdbc:inbound-channel-adapter allows you to use a SqlParameterSource to change the values in the selectQuery WHERE clause. However there is no mechanism to change the actual SQL. The adapter's constructor takes the selectQuery:String which can not be overridden or updated.

My requirement is such that a different SELECT (& UPDATE) is required based on the mode the application is running in. And the mode will change at runtime.

Gary Russell suggested a workaround, but also to open a feature suggestion:

> You can use a regular POJO inbound channel adapter with a JdbcTemplate as a work around.
> 
> Feel free to open a new Feature Suggestion.  Something like:
> 
> public JdbcPollingChannelAdapter(JdbcOperations jdbcOperations, Supplier<String> selectQuery) {

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3072

Allows the queries to change between polls.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
@@ -21,6 +21,7 @@ import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.util.List;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import javax.sql.DataSource;
 
@@ -56,7 +57,7 @@ public class JdbcPollingChannelAdapter extends AbstractMessageSource<Object> {
 
 	private final NamedParameterJdbcOperations jdbcOperations;
 
-	private final String selectQuery;
+	private final Supplier<String> selectQuery;
 
 	private RowMapper<?> rowMapper;
 
@@ -64,7 +65,7 @@ public class JdbcPollingChannelAdapter extends AbstractMessageSource<Object> {
 
 	private boolean updatePerRow = false;
 
-	private String updateSql;
+	private Supplier<String> updateSql;
 
 	private SqlParameterSourceFactory sqlParameterSourceFactory = new ExpressionEvaluatingSqlParameterSourceFactory();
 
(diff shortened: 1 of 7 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
```diff
@@ -82,14 +83,38 @@ public JdbcPollingChannelAdapter(DataSource dataSource, String selectQuery) {
 		this(new JdbcTemplate(dataSource), selectQuery);
 	}
 
+	/**
+	 * Constructor taking {@link DataSource} from which the DB Connection can be
+	 * obtained and the select query to execute to retrieve new rows.
+	 * @param dataSource Must not be null
+	 * @param selectQuery Supplier for the selectQuery query to execute
+	 * @since 5.3.1
```
artembilan: `5.2.1` then, but I'm a bit reluctant to do all this `Supplier`-based stuff.
Why then don't allow to configure a `DataSource` the same way?

What would be a problem to have just a simple setter for this `selectQuery` instead?
So, you still be able to change a query at runtime, but that is going to be in more cleaner way: really like a Control Bus command...

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
```diff
@@ -185,29 +220,32 @@ protected Object doReceive() {
 			payload = null;
 		}
 		if (payload != null && this.updateSql != null) {
-			if (this.updatePerRow) {
-				for (Object row : payload) {
-					executeUpdateQuery(row);
+			String sqlForUpdate = this.updateSql.get();
+			if (this.updateSql != null) {
```
artembilan: This is wrong already according a ` this.updateSql.get()` call before.

artembilan (changes requested):

garyrussell:
Closing in favor of a simple setter.

PR #3074, opened 2019-10-04
GH-3072: Allow JDBC inbound query changes

### LINKED ISSUE #3072: Feature Enhancement: JdbcPollingChannelAdapter to have configurable SQL

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3072

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
@@ -56,21 +56,21 @@ public class JdbcPollingChannelAdapter extends AbstractMessageSource<Object> {
 
 	private final NamedParameterJdbcOperations jdbcOperations;
 
-	private final String selectQuery;
-
 	private RowMapper<?> rowMapper;
 
 	private SqlParameterSource sqlQueryParameterSource;
 
 	private boolean updatePerRow = false;
 
-	private String updateSql;
-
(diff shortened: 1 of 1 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
```diff
@@ -116,6 +116,15 @@ public void setRowMapper(@Nullable RowMapper<?> rowMapper) {
 		}
 	}
 
+	/**
+	 * Set the select query.
+	 * @param selectQuery the query.
+	 * @since 5.2.1
+	 */
+	public void setSelectQuery(String selectQuery) {
+		this.selectQuery = selectQuery;
```
artembilan: `Assert.hasText(selectQuery, "'selectQuery' must be specified.");` still must be here: we can't have this component without `SELECT`

artembilan (changes requested):

PR #3075, opened 2019-10-05
Refinement for gateway Mono processing

artembilan:
* Introduce a `MessagingGatewaySupport.MonoReplyChannel` instead of
`FutureReplyChannel` for better on demand handling and reusing a
`Mono` returned from the target handler
* Refactor `GatewayProxyFactoryBean` to identify a target return type
(including generics for `Function`) during initialization
* Handle a `Mono` return type via
`MessagingGatewaySupport.sendAndReceiveMessageReactive()`
* Some `@Nullable` in the `GatewayProxyFactoryBean` and `ExpressionUtils`
* Add `MonoFunction` test-case into the `FunctionsTests.kt`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
@@ -17,8 +17,6 @@
 package org.springframework.integration.gateway;
 
 import java.lang.reflect.Method;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
 import java.lang.reflect.UndeclaredThrowableException;
 import java.util.Collections;
 import java.util.HashMap;
@@ -247,7 +245,6 @@ public class GatewayProxyFactoryBean extends AbstractEndpoint
 	/**
 	 * Set the default timeout value for sending request messages. If not explicitly
 	 * configured with an annotation, or on a method element, this value will be used.
-	 *
 	 * @param defaultRequestTimeout the timeout value in milliseconds
 	 */
 	public void setDefaultRequestTimeout(Long defaultRequestTimeout) {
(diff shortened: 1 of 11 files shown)


spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -424,7 +416,7 @@ protected void onInit() {
 
 	@Override
 	public Class<?> getObjectType() {
-		return (this.serviceInterface != null ? this.serviceInterface : null);
+		return this.serviceInterface;
```
garyrussell: Why? NPE if `afterPropertiesSet()` is not called.

garyrussell (commented):
Seems like a big change for a point release.

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
artembilan: Well, I think we can just remove a setter initialization since there is no any dependency injection overhead with just a `Class<?>` ctor.
Plus that `RequestReplyExchanger.class` could be used from a default ctor. And we don't need that `determineServiceInterface()` logic in the `afterPropertiesSet()`.

I'm OK deferring this to `5.3` and I really regret that it didn't make it into `5.2`, but on the other hand this is fully an internal change and brings some benefit for end-user code.
See that new `verify Mono gateway` test. It doesn't work without some workaround on the end-user side.
We could treat this as a bug, though, because a downstream `Mono` from end-user code is not processed properly in the current version.
But again: I'm OK to move this to `5.3` if you wish.

garyrussell: I am ok with merging it, as long as you are confident there are no breaking changes; I suppose we have good test coverage on gateways.

I would say go ahead an initialize the field to RRE and we can get rid of that `determineServiceInterface()` call in aPS.

PR #3077, opened 2019-10-11
GH-3076: Add file existence check

### LINKED ISSUE #3076: FileChannelCache.tryLockFor creates empty files
Hi all

`FileChannelCache.tryLockFor(File)` actually creates an empty file when trying to get a lock on a non-existent file.

It seems to be caused by line 56 `FileChannel newChannel = new RandomAccessFile(fileToLock, "rw").getChannel();` which opens the file using "rw" mode (create the file if it does not exist)

Tested on **v5.1.5.RELEASE**, seems to still be the same on **v5.2.0.RELEASE**.

It may be better to either :
- throw an exception to indicate the file does not exist
- return a specific value to indicate the file does not exist
- or just document that behavior

To reproduce, just call :

`FileChannelCache.tryLockFor(new File("/a/file/that/does/not/exists.txt"));`

Thanks!

EmmanuelRoux:
Fixes spring-projects/spring-integration#3076

* Add file existence check: throw a FileNotFoundException when trying to get a lock for a non-existent file

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/locking/FileChannelCache.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/locking/FileChannelCache.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.file.locking;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
@@ -30,6 +31,7 @@ import java.util.concurrent.ConcurrentMap;
  *
  * @author Iwein Fuld
  * @author Gary Russell
+ * @author Emmanuel Roux
  * @since 2.0
  */
 final class FileChannelCache {
@@ -43,13 +45,18 @@ final class FileChannelCache {
 
 	/**
 	 * Try to get a lock for this file while guaranteeing that the same channel will be used for all file locks in this
-	 * VM. If the lock could not be acquired this method will return <code>null</code>.
+	 * VM. If the lock could not be acquired this method will return <code>null</code>. If the file does not exist this
+	 * method will throw a {@link FileNotFoundException}.
 	 * <p>
 	 * Locks acquired through this method should be passed back to #closeChannelFor to prevent memory leaks.
 	 * <p>
 	 * Thread safe.
 	 */
 	public static FileLock tryLockFor(File fileToLock) throws IOException {
+		if (!fileToLock.exists()) {
+			throw new FileNotFoundException("Unable to get lock for non-existent file: " + fileToLock.getAbsolutePath());
+		}
+
 		FileChannel channel = channelCache.get(fileToLock);
 		if (channel == null) {
 			@SuppressWarnings("resource")
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/locking/FileChannelCache.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/locking/FileChannelCache.java
@@ -17,7 +17,6 @@
 package org.springframework.integration.file.locking;
 
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
@@ -25,6 +24,8 @@ import java.nio.channels.FileLock;
 import java.nio.channels.OverlappingFileLockException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+
+import org.springframework.lang.Nullable;
 
 /**
  * Static cache of FileLocks that can be used to ensure that only a single lock is used inside this ClassLoader.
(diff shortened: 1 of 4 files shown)


artembilan (changes requested):
@EmmanuelRoux ,

Thank you for contribution. Looks good, but any chances to have a unit test covering such a behavior?

EmmanuelRoux:
@artembilan Maybe we should discuss about modifying `NioFileLocker` too, as it's the only usage of `FileChannelCache`

With my first suggestion, the `FileNotFoundException` will be wrapped in a `MessagingException` by `NioFileLocker` (line 54).

I suggest `NioFileLocker.lock(File)` should instead return **false** in that case. That would result in both:
* `DefaultDirectoryScanner.tryClaim(File)` also returning **false**
* `FileReadingMessageSource.doReceive()` skipping non-existent (or removed) scanned files being kept in internal `toBeReceived` cache if `scanEachPoll` is **false** (the default)

artembilan:
Sounds like a plan, @EmmanuelRoux !

Although I'd like to see some unit tests anyway.

Thank you for your effort!

spring-integration-file/src/main/java/org/springframework/integration/file/locking/FileChannelCache.java
```diff
@@ -43,13 +45,18 @@ private FileChannelCache() {
...
+	 * VM. If the lock could not be acquired this method will return <code>null</code>. If the file does not exist this
+	 * method will throw a {@link FileNotFoundException}.
 	 * <p>
 	 * Locks acquired through this method should be passed back to #closeChannelFor to prevent memory leaks.
 	 * <p>
 	 * Thread safe.
 	 */
 	public static FileLock tryLockFor(File fileToLock) throws IOException {
+		if (!fileToLock.exists()) {
+			throw new FileNotFoundException("Unable to get lock for non-existent file: " + fileToLock.getAbsolutePath());
```
artembilan: Wouldn't return `null` instead of this exception enough for the method contract?
As you see in the `NioFileLocker`, you just suppress such an exception any way.

So, let's make a `return null;` here, add a `@Nullable` for method return and so on!

spring-integration-file/src/test/java/org/springframework/integration/file/locking/FileChannelCacheTests.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
```
artembilan: Well, we stay with a convention to have a first year when the class has been created.

spring-integration-file/src/test/java/org/springframework/integration/file/locking/FileChannelCacheTests.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/**
+ * @author Emmanuel Roux
```
artembilan: Since this class is new, it has to be marked with the version it was introduced: `@since 4.3.22`

spring-integration-file/src/test/java/org/springframework/integration/file/locking/FileChannelCacheTests.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class FileChannelCacheTests {
+
+	private File workdir;
+
+	@Rule
+	public TemporaryFolder temp = new TemporaryFolder() {
+
+		@Override
+		public void create() throws IOException {
```
artembilan: I don't see reason in such extra logic.
It looks like `temp.newFile("test0");` is fully enough against a default `public TemporaryFolder temp = new TemporaryFolder();`

Although I would insist on JUnit 5 already in the tests.
I'll back-port it to JUnit 4 for version `4.3.x` on merge

spring-integration-file/src/test/java/org/springframework/integration/file/locking/FileChannelCacheTests.java
```diff
@@ -0,0 +1,61 @@
...
+			super.create();
+			workdir = temp.newFolder(FileChannelCacheTests.class.getSimpleName());
+		}
+	};
+
+	@Test(expected = FileNotFoundException.class)
```
artembilan: It is much cleaner to use an `Assertions.assertThatExceptionOfType()` in the code instead.
Although when we agree about `null` instead, this will go away anyway.

spring-integration-file/src/test/java/org/springframework/integration/file/locking/NioFileLockerTests.java
```diff
@@ -76,4 +77,20 @@ public void fileNotListedWhenLockedByOtherFilter() throws IOException {
...
+		locker.unlock(testFile);
+	}
+
+	@Test
+	public void fileNotLockedWhenNotExists() throws IOException {
+		NioFileLocker locker = new NioFileLocker();
+		File testFile = new File(workdir, "test3");
+		assertThat(locker.lock(testFile)).isFalse();
+		assertThat(testFile.exists()).isFalse();
+	}
```
artembilan: Every member in the class should be wrapped with blank lines, even the last method before the end of class.

artembilan (changes requested):
Please, find my review.
It might be slight nit-picking, but we develop here Open Source and our code should be in high quality 😉 

Also, consider in the future not squashing commits for PR. First of all we get a notification about new commits. Secondly it is much easier to review a historical flow. We squash them during merge anyway.

Thanks for undertstanding

spring-integration-file/src/main/java/org/springframework/integration/file/locking/FileChannelCache.java
EmmanuelRoux: That's what I suggested as an alternative in #3076. In my mind, the semantics were different between _a lock that failed because the file was already locked (or any other filesystem-related reason)_, and _an attempt to lock a file that does not exist_.

Anyway, I will push a new version returning `null` 😉

artembilan: Well, if there is nothing to lock, then we definitely not locked. Seems for me that's the point of that contract.
So, let's live with `null`  for non-existing file as well!

spring-integration-file/src/test/java/org/springframework/integration/file/locking/FileChannelCacheTests.java
EmmanuelRoux: Reason is that `TemporaryFolder.newFile(String)` calls `File.createNewFile()` which is not what is desired for test case of a non-existent file

artembilan: Well, you can do then `this.temp.getRoot()` for a `new File()`.
My point is: do not over-engineer a `TemporaryFolder` if we really can rely on its default public API.
With JUnit 5 we definitely don't have a luxury to extend extensions on the fly.

EmmanuelRoux: You're right, let's go for that

artembilan:
... and cherry-picked to `5.1.x`.

Back-ported to `4.3.x` after removing `@Nullable` and AssertJ constructions.

Will clean up code a little bit and fix AssertJ on `5.1.x` branch as well.

Thank you very much for contribution; looking forward for more!

Issue #3026, opened 2019-08-13
can't chmod using ftp outbound Adapter


joaquinjsb:
**Affects Version(s):** 5.2.0.M4

---
## Question

Hello guys, I'm having problems to use chmod with an outbound adapter,  how can this be solved?

I gave a look around the code class **FileTransferringMessageHandler,** seems to be hardcoded
![image](https://user-images.githubusercontent.com/2885322/62949456-7edc3800-bde6-11e9-8764-dcb417201ff0.png)

my dsl

```java

  return flow -> flow
            .handle(Ftp.outboundGateway(FTPServers.PC_LOCAL.getFactory(), AbstractRemoteFileOutboundGateway.Command.MGET, "payload")
                .fileExistsMode(FileExistsMode.REPLACE)
                .filterFunction(ftpFile -> ftpFile.getName().contains("SCN") || ftpFile.getName().contains("REP") || ftpFile.getName().contains("EL"))
                .localDirectory(new File("/localTmp")))
            .log(Message::getPayload)
            .transform(GenericMessage.class, GenericMessage::getPayload)
            .split()
            .handle(Ftp.outboundAdapter(FTPServers.PC_LOCAL.getFactory(), FileExistsMode.IGNORE)
                .useTemporaryFileName(false)
                .chmod(666)
                .fileNameExpression("headers['" + FileHeaders.REMOTE_FILE + "']")
                .remoteDirectory("/dati/recpv"), e -> e.advice(afterFlow()));
    }
```

garyrussell:
GitHub issues are not for asking questions. Did you not read the instructions?

>## Question

>Please ask questions about how to use something, or to understand why something isn't
working as you expect it to, on
>[Stack Overflow using the `spring-integration` tag](https://stackoverflow.com/questions/tagged/spring-integration).

The FTP protocol does not support changing the file permissions, only SFTP.

joaquinjsb:
Hello,
I'm using FTP protocol, I'm pretty sure that you can chmod using FTP (at least as far as I know, you can do it with Filezilla), I tried too with ftp command line client on ubuntu and seems to be possible,  it's should be an enhancement?

garyrussell:
I should have googled it when I implemented it for SFTP; there is no direct support in the Apache `FTPClient` - but it appears it can be done with a site command:

https://stackoverflow.com/questions/12741938/how-can-i-change-permissions-of-a-file-on-a-ftp-server-using-apache-commons-net

[...]

PR #3080, opened 2019-10-15
GH-3026: (S)FTPMessageHandler chmod functionality

joaquinjsb:
Fixes spring-projects/spring-integration#3026

Hello,
this should address the GH-3026 issue, I have to do the coverage test but I'm having a problem with the gradle checkstyle.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -331,6 +331,7 @@ subprojects { subproject ->
 
 	checkstyle {
 		configFile = file("$rootDir/src/checkstyle/checkstyle.xml")
+		configDir = file("$rootDir/src/checkstyle/")
 		toolVersion = '8.24'
 	}
 
--- a/build.gradle
+++ b/build.gradle
@@ -331,7 +331,6 @@ subprojects { subproject ->
 
 	checkstyle {
 		configFile = file("$rootDir/src/checkstyle/checkstyle.xml")
-		configDir = file("$rootDir/src/checkstyle/")
 		toolVersion = '8.24'
 	}
 
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
@@ -16,11 +16,6 @@
 
 package org.springframework.integration.file.dsl;
 
-import java.nio.charset.Charset;
-import java.util.Collections;
-import java.util.Map;
-import java.util.function.Function;
-
 import org.springframework.expression.common.LiteralExpression;
 import org.springframework.integration.dsl.ComponentsRegistration;
 import org.springframework.integration.dsl.MessageHandlerSpec;
@@ -33,6 +28,11 @@ import org.springframework.integration.file.remote.session.SessionFactory;
 import org.springframework.integration.file.support.FileExistsMode;
 import org.springframework.messaging.Message;
 import org.springframework.util.Assert;
+
+import java.nio.charset.Charset;
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.Function;
 
 /**
  * The {@link MessageHandlerSpec} for the {@link FileTransferringMessageHandler}.
@@ -51,6 +51,9 @@ public abstract class FileTransferringMessageHandlerSpec<F, S extends FileTransf
 	private FileNameGenerator fileNameGenerator;
 
 	private DefaultFileNameGenerator defaultFileNameGenerator;
+
+	protected FileTransferringMessageHandlerSpec() {
+	}
 
 	protected FileTransferringMessageHandlerSpec(SessionFactory<F> sessionFactory) {
 		this.target = new FileTransferringMessageHandler<>(sessionFactory);
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
@@ -16,6 +16,11 @@
 
 package org.springframework.integration.file.dsl;
 
+import java.nio.charset.Charset;
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.Function;
+
 import org.springframework.expression.common.LiteralExpression;
 import org.springframework.integration.dsl.ComponentsRegistration;
 import org.springframework.integration.dsl.MessageHandlerSpec;
@@ -28,11 +33,6 @@ import org.springframework.integration.file.remote.session.SessionFactory;
 import org.springframework.integration.file.support.FileExistsMode;
 import org.springframework.messaging.Message;
 import org.springframework.util.Assert;
-
-import java.nio.charset.Charset;
-import java.util.Collections;
-import java.util.Map;
-import java.util.function.Function;
 
 /**
  * The {@link MessageHandlerSpec} for the {@link FileTransferringMessageHandler}.
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
@@ -52,6 +52,7 @@ public abstract class FileTransferringMessageHandlerSpec<F, S extends FileTransf
 
 	private DefaultFileNameGenerator defaultFileNameGenerator;
 
+	// TODO: 16/10/2019 this should be refactored using generics in next release (breaking change), see PR-3080.
 	protected FileTransferringMessageHandlerSpec() {
 	}
 
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
@@ -64,7 +64,7 @@ public class FtpMessageHandler extends FileTransferringMessageHandler<FTPFile> {
 	@Override
 	protected void doChmod(RemoteFileTemplate<FTPFile> remoteFileTemplate, final String path, final int chmod) {
 		remoteFileTemplate.executeWithClient((ClientCallbackWithoutResult<FTPClient>) client -> {
-			String chModCommand = "chmod " + Integer.toOctalString(chmod) + " " + path;
+			String chModCommand = "chmod " + chmod + " " + path;
 			try {
 				client.sendSiteCommand(chModCommand);
 			}
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
@@ -48,13 +48,10 @@ public class FtpMessageHandler extends FileTransferringMessageHandler<FTPFile> {
 		super(remoteFileTemplate);
 	}
 
-	public FtpMessageHandler(RemoteFileTemplate<FTPFile> remoteFileTemplate, FileExistsMode mode) {
-		super(remoteFileTemplate, mode);
+	public FtpMessageHandler(FtpRemoteFileTemplate template, FileExistsMode fileExistsMode) {
+		super(template, fileExistsMode);
 	}
 
-	public FtpMessageHandler(RemoteFileTemplate<FTPFile> remoteFileTemplate) {
-		super(remoteFileTemplate);
-	}
 
 	@Override
 	public boolean isChmodCapable() {
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
@@ -52,6 +52,10 @@ public class FtpMessageHandler extends FileTransferringMessageHandler<FTPFile> {
 		super(remoteFileTemplate, mode);
 	}
 
+	public FtpMessageHandler(RemoteFileTemplate<FTPFile> remoteFileTemplate) {
+		super(remoteFileTemplate);
+	}
+
 	@Override
 	public boolean isChmodCapable() {
 		return true;
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
@@ -48,8 +48,8 @@ public class FtpMessageHandler extends FileTransferringMessageHandler<FTPFile> {
 		super(remoteFileTemplate);
 	}
 
-	public FtpMessageHandler(FtpRemoteFileTemplate template, FileExistsMode fileExistsMode) {
-		super(template, fileExistsMode);
+	public FtpMessageHandler(RemoteFileTemplate<FTPFile> remoteFileTemplate, FileExistsMode mode) {
+		super(remoteFileTemplate, mode);
 	}
 
 
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
@@ -64,7 +64,7 @@ public class FtpMessageHandler extends FileTransferringMessageHandler<FTPFile> {
 	@Override
 	protected void doChmod(RemoteFileTemplate<FTPFile> remoteFileTemplate, final String path, final int chmod) {
 		remoteFileTemplate.executeWithClient((ClientCallbackWithoutResult<FTPClient>) client -> {
-			String chModCommand = "chmod " + chmod + " " + path;
+			String chModCommand = "chmod " + Integer.toOctalString(chmod) + " " + path;
 			try {
 				client.sendSiteCommand(chModCommand);
 			}
--- a/spring-integration-ftp/src/test/java/org/springframework/integration/ftp/dsl/FtpTests.java
+++ b/spring-integration-ftp/src/test/java/org/springframework/integration/ftp/dsl/FtpTests.java
@@ -198,7 +198,7 @@ public class FtpTests extends FtpTestSupport {
 				.handle(Ftp.outboundAdapter(sessionFactory(), FileExistsMode.FAIL)
 						.useTemporaryFileName(false)
 						.fileNameExpression("headers['" + FileHeaders.FILENAME + "']")
-						.chmod(644)
+						.chmod(0644)
 						.remoteDirectory("ftpTarget"));
 		IntegrationFlowRegistration registration = this.flowContext.registration(flow).register();
 		String fileName = "foo.file";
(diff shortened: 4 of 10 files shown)


pivotal-issuemaster:
@joaquinjsb Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3080)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3080) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@joaquinjsb Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3080)!

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
```diff
@@ -34,6 +29,11 @@
 import org.springframework.messaging.Message;
 import org.springframework.util.Assert;
 
+import java.nio.charset.Charset;
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.Function;
```
artembilan: You can't reorder imports. That's where Checkstyle fails for you.
Please, consider do not do auto-formatting in your IDE before pushing commits.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/outbound/SftpMessageHandler.java
```diff
@@ -66,6 +66,14 @@ public SftpMessageHandler(SessionFactory<LsEntry> sessionFactory) {
 		this(new SftpRemoteFileTemplate(sessionFactory));
 	}
 
+	public SftpMessageHandler(RemoteFileTemplate<LsEntry> remoteFileTemplate) {
```
artembilan: We can't introduce these new ctors based on the super `RemoteFileTemplate`.
The specs must be restricted only to their specific `SftpRemoteFileTemplate` or `FtpRemoteFileTemplate`.

It's kinda breaking change, but we can introduce new ctors an factory methods with deprecation of those based on just a `RemoteFileTemplate`

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java
```diff
@@ -52,6 +52,9 @@
 
 	private DefaultFileNameGenerator defaultFileNameGenerator;
 
+	protected FileTransferringMessageHandlerSpec() {
```
artembilan: I would consider extra generic argument on this class like:
```
FileTransferringMessageHandlerSpec<F, H extends FileTransferringMessageHandler<F>, S extends FileTransferringMessageHandlerSpec<F, H, S>>
		extends MessageHandlerSpec<S, H>
```

So, all the implementations would follow requirements.
But this indeed going to be a breaking change.
So, let's leave your empty ctor solution for time being with `//TODO` to fix in the next version via generics!

artembilan (changes requested):
Please, push some tests as well to be sure that chmod works and also add your name to `@author` list for all the changed classes.

Thanks

build.gradle
```diff
@@ -331,6 +331,7 @@ subprojects { subproject ->
 
 	checkstyle {
 		configFile = file("$rootDir/src/checkstyle/checkstyle.xml")
+		configDir = file("$rootDir/src/checkstyle/")
```
artembilan: Hm. Doesn't this mean that you fully bypass Checkstyle?
Your import order is wrong and it definitely has to fail.

Please, consider to revert Checkstyle changes and fix your code according requirements of our checks.
Thanks

artembilan (changes requested):

artembilan:
Anyway it failed for you again on Travis:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java:32: Wrong order for 'java.nio.charset.Charset' import. [ImportOrder]
```

joaquinjsb:
> Anyway it failed for you again on Travis:
> 
> ```
> [ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java:32: Wrong order for 'java.nio.charset.Charset' import. [ImportOrder]
> ```

I'm adding the check style plugin for intelliJ, should solve the issue!

artembilan:
> I'm adding the check style plugin for intelliJ

I don't think so. I don't know if that plugin can resolve a config from our project.
You should use a Gradle command to check your changes: `/gradlew clean check --parallel`

joaquinjsb:
> > I'm adding the check style plugin for intelliJ
> 
> I don't think so. I don't know if that plugin can resolve a config from our project.
> You should use a Gradle command to check your changes: `/gradlew clean check --parallel`

looks like that at the beginning IntelliJ completely ignored the file, that's the reason of the ordered imports, and there was another problem with the tests 

> Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: cannot initialize module SuppressionFilter - Unable to find: src/checkstyle/checkstyle-suppressions.xml
	at com.puppycrawl.tools.checkstyle.Checker.setupChild(Checker.java:477)
	at com.puppycrawl.tools.checkstyle.api.AutomaticBean.configure(AutomaticBean.java:198)
	at com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask.createRootModule(CheckstyleAntTask.java:412)
	... 213 more
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to find: src/checkstyle/checkstyle-suppressions.xml

looks like it couldn'f find the file on the module. RegexpHeader

that's why I changed this with the parameter

```<!-- Root Checks -->
	<module name="RegexpHeader">
		<property name="headerFile" value="src/checkstyle/checkstyle-header.txt"/>
		<property name="fileExtensions" value="java"/>
	</module>```

File: spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java (hunk 3 above)
Comment (joaquinjsb) [reply spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java]: I'll add the TODO

**Comment by artembilan** [comment]
> looks like it couldn'f find the file on the module

Let's don't discuss IDE support for Checkstyle in this issue!

If you really have some valuable solution for that side, it should be addressed in the separate issue to preserve a scope.

Thanks

**Comment by artembilan** [comment]
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/dsl/SftpMessageHandlerSpec.java:20: Wrong order for 'org.springframework.integration.file.dsl.FileTransferringMessageHandlerSpec' import. [ImportOrder]
```

Please, try do not reorder imports.

File: build.gradle (hunk 4 above)
Comment (joaquinjsb) [reply build.gradle]: I removed this code, modified locally the xml file.

File: spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java (hunk 1 above)
Comment (joaquinjsb) [reply spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java]: fixed this issue.

File: spring-integration-sftp/src/main/java/org/springframework/integration/sftp/outbound/SftpMessageHandler.java (hunk 2 above)
Comment (joaquinjsb) [reply spring-integration-sftp/src/main/java/org/springframework/integration/sftp/outbound/SftpMessageHandler.java]: so changing this to

```
SftpMessageHandlerSpec(RemoteFileTemplate<ChannelSftp.LsEntry> remoteFileTemplate) {
		this.target = new SftpMessageHandler(remoteFileTemplate.getSessionFactory());
	}

	SftpMessageHandlerSpec(RemoteFileTemplate<ChannelSftp.LsEntry> remoteFileTemplate, FileExistsMode fileExistsMode) {
		this.target = new SftpMessageHandler(new SftpRemoteFileTemplate(remoteFileTemplate.getSessionFactory()), fileExistsMode);
	}
``` 

and removing ctors makes sense for you?

**Comment by joaquinjsb** [comment]
the tests are failing now, looks like the chmod isn't setting the right permissions, I thought it was a Windows fault, looks like octal mode is wrong?

File: spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java (hunk 3 above)
Comment (joaquinjsb) [reply spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileTransferringMessageHandlerSpec.java]: it's ok now

File: spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java [hunk 5]
```diff
@@ -60,7 +64,7 @@ public boolean isChmodCapable() {
 	@Override
 	protected void doChmod(RemoteFileTemplate<FTPFile> remoteFileTemplate, final String path, final int chmod) {
 		remoteFileTemplate.executeWithClient((ClientCallbackWithoutResult<FTPClient>) client -> {
-			String chModCommand = "chmod " + Integer.toOctalString(chmod) + " " + path;
+			String chModCommand = "chmod " + chmod + " " + path;
```
artembilan: What made you to do this radical architectural change?

I think that is really a point to perform `chmod` command with *octal*.

artembilan (changes requested):

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
joaquinjsb: Already reverted!

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/dsl/FtpTests.java
```diff
@@ -192,6 +192,36 @@ public void testFtpOutboundFlow() {
 		registration.destroy();
 	}
 
+	@Test
+	public void testFtpOutboundFlowWithChmod() {
+		IntegrationFlow flow = f -> f
+				.handle(Ftp.outboundAdapter(sessionFactory(), FileExistsMode.FAIL)
+						.useTemporaryFileName(false)
+						.fileNameExpression("headers['" + FileHeaders.FILENAME + "']")
+						.chmod(644)
```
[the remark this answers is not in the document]
joaquinjsb: Yes, but is is done in the client with Integer.toOctalString(), which should return  1204

```
@Override
	protected void doChmod(RemoteFileTemplate<FTPFile> remoteFileTemplate, final String path, final int chmod) {
		remoteFileTemplate.executeWithClient((ClientCallbackWithoutResult<FTPClient>) client -> {
			String chModCommand = "chmod " + Integer.toOctalString(chmod) + " " + path;
			try {
				client.sendSiteCommand(chModCommand);
			}
			catch (IOException e) {
				throw new UncheckedIOException("Failed to execute '" + chModCommand  + "'", e);
			}
		});
	}
```

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
```diff
@@ -52,6 +52,10 @@ public FtpMessageHandler(RemoteFileTemplate<FTPFile> remoteFileTemplate, FileExi
 		super(remoteFileTemplate, mode);
 	}
 
+	public FtpMessageHandler(RemoteFileTemplate<FTPFile> remoteFileTemplate) {
```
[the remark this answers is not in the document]
artembilan: Well, that's right, but this one is an `FtpMessageHandler` and you have introduced wrong ctor exactly here.

Please, revise an FTP part the same way you have in SFTP.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/outbound/SftpMessageHandler.java
joaquinjsb: should be ok now, thanks for the feeback!

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/dsl/FtpTests.java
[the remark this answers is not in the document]
artembilan: Not sure why you said that it must return `1204`, because this is what I have in my test:
```
Integer.toOctalString(0644)
...
Result: 644
```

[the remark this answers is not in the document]
joaquinjsb: I just made some more tests, then I'm right, give a look to my debug session

image

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
```diff
@@ -48,10 +48,11 @@ public FtpMessageHandler(FtpRemoteFileTemplate remoteFileTemplate) {
 		super(remoteFileTemplate);
 	}
 
-	public FtpMessageHandler(RemoteFileTemplate<FTPFile> remoteFileTemplate, FileExistsMode mode) {
-		super(remoteFileTemplate, mode);
+	public FtpMessageHandler(FtpRemoteFileTemplate template, FileExistsMode fileExistsMode) {
```
artembilan: That's not correct to make such a breaking change.

We need to introduce a new ctor (to align with an `SftpMessageHandler`) and deprecate this one.

artembilan (changes requested):
Please, address the latest concern then we will be ready for another big change/review round.

I'll pull this locally to take a look closer.

Thanks so far!

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/dsl/FtpTests.java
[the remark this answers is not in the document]
joaquinjsb: > > looks like the permission is not being set properly
> 
> Maybe you expect permissions which are not really a result of the provided octal?

I just made a test on Linux, chmod 0644 test.file, and the result was this
-rw-r--r--   1 root root     0 Oct 16 17:30 test.file

so that translated to the code should make sense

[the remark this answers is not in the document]
artembilan: Shouldn't that be `chmod 644 test.file` on Linux instead?
That is exactly what we get with the `"chmod " + Integer.toOctalString(chmod) + " " + path;`

The point is that such a number must be *octal* and only the way to restrict it on Java side is to provide a value with a leading `0`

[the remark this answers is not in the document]
joaquinjsb: > Shouldn't that be `chmod 644 test.file` on Linux instead?
> That is exactly what we get with the `"chmod " + Integer.toOctalString(chmod) + " " + path;`
> 
> The point is that such a number must be _octal_ and only the way to restrict it on Java side is to provide a value with a leading `0`

wether i do it with linux (0)644, the result is always the same (-rw-r--r--), that's the weird thing, is maybe the fake internal client ? I'll try with a local filezilla thing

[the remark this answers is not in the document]
joaquinjsb: looks like it's the implementation, 
I'll update you when I'm done with filezilla
https://mina.apache.org/ftpserver-project/ftpserver_site_commands.html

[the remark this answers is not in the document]
artembilan: Ah! You mean an embedded MINA FTP Server we use for testing doesn't implement `chmod` site command properly?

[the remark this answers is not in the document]
joaquinjsb: I think so, even though I would expect an exception that's it's not implement, or maybe it just bypasses it?
I'll let you know by the way after the test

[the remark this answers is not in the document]
joaquinjsb: Yes, 
I can confirm, the problem is the built-in server, we'll have to skip the test or what other options do we have?
image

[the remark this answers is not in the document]
artembilan: Well, we can just make the test against mocks. 
See sample in the `FtpServerOutboundTests.testInt3088MPutNotRecursive()`.

It really not so important if command has reached the server or not. Our goal is to ensure that your DSL fix brings for us a proper chmod command in the end.

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/outbound/FtpMessageHandler.java
joaquinjsb: I restored it back as i was before

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/dsl/FtpTests.java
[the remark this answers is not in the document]
joaquinjsb: @artembilan  turning back to this point, so is enough as it is with the .chmod on the flow ? the mockito should be sending the site command right ?

image

artembilan:
Gentle ping.
Any chances that you will come back to us with the fix by the end of this week?
We have release plans next week though...

It's not a problem if you can't make it: we will take your work so far and will adjust it for the merge respectively.

Thanks for understanding!

joaquinjsb:
Hey,
I was on holidays, tomorrow I'll be commiting!

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/dsl/FtpTests.java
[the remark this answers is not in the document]
joaquinjsb: yes, that was what I tried, I commited, let me know!

joaquinjsb:
> Hello, @joaquinjsb !
> 
> I see you have added some changes, but I still don't see a proper mock test.
> 
> Are you going to implement it yet?
> Or what are your plans left for this PR?
> 
> Thanks

what else can do I add?  I mean, adding the "chmod" option to the flow is now passing, before you got an Exception, how else can I test this ?

artembilan (approved):
Great!
I'll pull it locally for one more review round and will let you know how it is today later.

Thanks

joaquinjsb:
thank you!

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/dsl/SftpTests.java
```diff
@@ -146,6 +146,33 @@ public void testSftpOutboundFlow() {
 		registration.destroy();
 	}
 
+
+	@Test
+	public void testSftpOutboundFlowWithChmod() {
```
artembilan: I `@Ignore` this test because it doesn't work as expected on Windows: the `rw` permissions are assigned for all principals 😢

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/7aea76c6c45261cfa5fb62b4ec521769341e6912.

Cherry-picked to `5.1.x` as https://github.com/spring-projects/spring-integration/commit/35b964cc79ead8c11ed2f062a3a40a94ad391933 after fixing conflicts in tests with AssertJ.

So, @joaquinjsb ,
thank you very much for your contribution; looking forward for more!

For example: https://github.com/spring-projects/spring-integration/issues/3092 😄

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/dsl/SftpTests.java
joaquinjsb: Yes, I got the same error on windows :( looks like it works only on Linux

joaquinjsb:
Looking forward for 3092 :D

PR #3082, opened 2019-10-15
GH-3079: Use getMostSpecificMethod for SpEL calls

### LINKED ISSUE #3079: SpEL invocation for functional interfaces cannot find the method for signature
See more info in the https://stackoverflow.com/questions/58259113/how-transform-string-in-subflow-route-using-spring-integration-dsl.

When we can't process an *empty payload* via `PayloadMethodArgumentResolver`, we fallback to SpEL execution and here it try to find a method in the interface implementation, but cannot, because there is no an `abstract` method in the list of target methods.

Looks like an assumption about:
```
TRANSFORMER_TRANSFORM_METHOD =
				ReflectionUtils.findMethod(genericTransformerClass, "transform", (Class<?>[]) null);
```
to be used in the `FixedMethodFilter` for SpEL invoker is wrong and we need to resolve non-lambda methods other way.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3079

It turns out that we need to use a *most specific* method for the target
to be called reflectively from the SpEL

NOTE: We drop a *Method not found* error in case of wrong method source
since now we find a right source via `ClassUtils.getMostSpecificMethod()`

**Cherry-pick to 5.1.x**

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/dsl/flows/IntegrationFlowTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/dsl/flows/IntegrationFlowTests.java
@@ -74,6 +74,7 @@ import org.springframework.integration.store.MessageStore;
 import org.springframework.integration.store.SimpleMessageStore;
 import org.springframework.integration.support.MessageBuilder;
 import org.springframework.integration.support.MutableMessageBuilder;
+import org.springframework.integration.transformer.GenericTransformer;
 import org.springframework.integration.transformer.PayloadSerializingTransformer;
 import org.springframework.integration.util.NoBeansOverrideAnnotationConfigContextLoader;
 import org.springframework.messaging.Message;
@@ -362,10 +363,10 @@ public class IntegrationFlowTests {
 		assertThat(out).isNotNull();
 		assertThat(out.getPayload()).isEqualTo("bar");
 
-		this.tappedChannel5.send(new GenericMessage<>("foo"));
+		this.tappedChannel5.send(new GenericMessage<>(""));
 		out = this.wireTapSubflowResult.receive(10000);
 		assertThat(out).isNotNull();
-		assertThat(out.getPayload()).isEqualTo("FOO");
+		assertThat(out.getPayload()).isEqualTo("");
 	}
 
 	@Autowired
@@ -700,7 +701,15 @@ public class IntegrationFlowTests {
 		public IntegrationFlow wireTapFlow5() {
 			return f -> f
 					.wireTap(sf -> sf
-							.<String, String>transform(String::toUpperCase)
+							.transform(// Must not be lambda for SpEL fallback behavior on empty payload
+									new GenericTransformer<String, String>() {
+
+										@Override
+										public String transform(String source) {
+											return source.toUpperCase();
+										}
+
+									})
 							.channel(MessageChannels.queue("wireTapSubflowResult")))
 					.channel("nullChannel");
 		}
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/test/java/org/springframework/integration/dsl/flows/IntegrationFlowTests.java
```diff
@@ -700,7 +701,15 @@ public IntegrationFlow wireTapFlow4() {
 		public IntegrationFlow wireTapFlow5() {
 			return f -> f
 					.wireTap(sf -> sf
-							.<String, String>transform(String::toUpperCase)
+							.transform(// Must not be lambda for SpEL fallback behavior on empty payload
```
garyrussell: I am not sure why the payload resolver doesn't allow an empty string but I suppose it's rare so falling back to SpEL is ok.

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
```diff
@@ -449,18 +448,6 @@ public void testProcessMessageCheckedException() throws Exception {
...
-		Method method = TestErrorService.class.getMethod("checked", String.class);
-		MethodInvokingMessageProcessor processor = new MethodInvokingMessageProcessor(service, method);
-		processor.setUseSpelInvoker(true);
-		processor.setBeanFactory(mock(BeanFactory.class));
-		assertThatExceptionOfType(MessageHandlingException.class)
-				.isThrownBy(() -> processor.processMessage(new GenericMessage<>("foo")))
-				.withCauseInstanceOf(SpelEvaluationException.class);
-	}
-
 	@Test
```
garyrussell: Hmmm - I don't think this is right - there is no relationship between the methods on these two classes - it looks like `getMostSpecificMethod()` doesn't check that the method is in the class hierarchy - seems like a bug to me. I think this test should still pass.

Consider ignoring instead and perhaps open an issue in SF to see whether they consider the behavior to be correct?

garyrussell (commented):
LGTM; but see comment regarding removed test.

spring-integration-core/src/test/java/org/springframework/integration/dsl/flows/IntegrationFlowTests.java
artembilan: I treat this like a note and not as some confusion or advice to change something.
Correct?

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
artembilan: OK. Makes sense.
It bumped me as well: why `getMostSpecificMethod()` doesn't care about the class of source.

I'll revert the test change with an appropriate `@Ignore`.

We still don't pass for Kotlin variant, but that's also out of SI scope...

garyrussell: Looks like it's intentional...

```java
return targetClass.getMethod(method.getName(), method.getParameterTypes());
```

But it seems odd to me.

spring-integration-core/src/test/java/org/springframework/integration/dsl/flows/IntegrationFlowTests.java
garyrussell: Yes, it's fine; I was just making a comment; since the problem only arises with an empty payload.

artembilan: Well, it also is there when we use a `@UseSpelInvoker`.

I believe I had to come up with the `@Transformer` against that `GenericTransformer` impl test-case, but it doesn't matter so far: I don't think it makes sense to use  `@Transformer` and don't rely on powerful POJO method call support.

spring-integration-core/src/test/java/org/springframework/integration/handler/MethodInvokingMessageProcessorTests.java
artembilan: Here it is: https://github.com/spring-projects/spring-framework/issues/23824

garyrussell:
Cherry-picked as 80ed25ebf308a28d383c119d480061d54873985a after resolving conflicts.

PR #3084, opened 2019-10-16
GH-3083: Support @Header with dotted literals

### LINKED ISSUE #3083: @Header("'foo.bar'") looks for a header ''foo' with property bar'
See https://stackoverflow.com/questions/58403246/header-annotation-with-in-header-name-such-as-headerfoo-bar-doesnt-m/58414528#58414528

When `@Header("'foo.bar'")` is used, we end up with:

`#target.handleMessage(payload, #requiredHeader(headers, ''foo').bar')`

garyrussell:
Fixes https://github.com/spring-projects/spring-integration/issues/3083

`@Header("foo.bar")` means extract property `bar` from header `foo`.

Support `@Header("'foo.bar'")`, meaning get the value of header `foo.bar`.

**cherry-pick to 5.1.x, 4.3.x**


artembilan (approved):
I wish we haven't had this SpEL support in `@Header` from the beginning at all...
It is too complicated and can be achieved with much cleaner code before or after header extraction.

But you fixed it - LGTM!

artembilan:
... and cherry-picked to `5.1.x` and `4.3.x` after fixing conflicts.

gabrielbauman:
Thank you for this. We hit it pretty hard and it took us awhile to realize SpEL was going on.

PR #3093, opened 2019-10-30
Introduce `IntegrationPattern` abstraction

artembilan:
* Add `IntegrationPattern` contract to implement on the target components
which represent particular EIP
* Add `IntegrationPatternType` with an internal `IntegrationPatternCategory`
to return from the component implementing `IntegrationPattern`
* Parse `IntegrationPatternType` in the `IntegrationNode` for potential
use on the UI for drawing a particular icon

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
@@ -21,7 +21,7 @@ package org.springframework.integration;
  *
  * @author Artem Bilan
  *
- * @since 5.2.1
+ * @since 5.2.2
  *
  * @see IntegrationPatternType
  * @see <a href="https://www.enterpriseintegrationpatterns.com/patterns/messaging">EIP official site</a>
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
@@ -21,7 +21,7 @@ package org.springframework.integration;
  *
  * @author Artem Bilan
  *
- * @since 5.2.2
+ * @since 5.3
  *
  * @see IntegrationPatternType
  * @see <a href="https://www.enterpriseintegrationpatterns.com/patterns/messaging">EIP official site</a>
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
@@ -17,9 +17,7 @@
 package org.springframework.integration;
 
 /**
- * A marker interface indicating that component implements some
- * Enterprise Integration Pattern.
- *
+ * Indicates that a component implements some Enterprise Integration Pattern.
  *
  * @author Artem Bilan
  *
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration;
+
+/**
+ * A marker interface indicating that component implements some
+ * Enterprise Integration Pattern.
+ *
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.2.1
+ *
+ * @see IntegrationPatternType
+ * @see <a href="https://www.enterpriseintegrationpatterns.com/patterns/messaging">EIP official site</a>
+ */
+public interface IntegrationPattern {
+
+	/**
+	 * Return a pattern type this component implements.
+	 * @return the {@link IntegrationPatternType} this component implements.
+	 */
+	IntegrationPatternType getIntegrationPatternType();
+
+}
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
@@ -54,6 +54,8 @@ public enum IntegrationPatternType {
 
 	inbound_gateway(IntegrationPatternCategory.messaging_endpoint),
 
+	gateway(IntegrationPatternCategory.messaging_endpoint),
+
 	splitter(IntegrationPatternCategory.message_routing),
 
 	transformer(IntegrationPatternCategory.message_transformation),
@@ -119,6 +121,7 @@ public enum IntegrationPatternType {
 				inbound_channel_adapter,
 				outbound_gateway,
 				inbound_gateway,
+				gateway,
 				bridge),
 
 		message_routing(
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
@@ -22,7 +22,7 @@ import java.util.stream.Collectors;
 
 /**
  * The Enterprise Integration Pattern types.
- * Used to indicate which pattern the target component implements.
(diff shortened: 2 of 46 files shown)


spring-integration-core/src/main/java/org/springframework/integration/IntegrationPattern.java
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration;
+
+/**
+ * A marker interface indicating that component implements some
```
garyrussell: a component

garyrussell: Not strictly a "marker" - they are usually empty.

How about "Indicates that a component ..." ?

garyrussell (changes requested):
Maybe I am missing something but I don't see enough information about how the UI will choose an icon.

artembilan:
see an `IntegrationNode` changes. There is already an `integrationPatternType` property.
Why this cannot be mapped to icon on the UI?
I fully don't understand the reason to mix UI attributes in our code.
That is really not our model responsibility to dictate the icon.
An `integrationPatternType`  alway can simply be mapped to some icon on UI as one-to-one relationship.

What am I missing?

Thanks

BoykoAlex:
Great news! Adding @kdvolder as well.

garyrussell:
It's not clear what the category is for; if it's not exposed on the graph.

garyrussell:
My confusion is AMH exposes itself as service activator whereas it can be that, an outbound gateway, etc, etc

artembilan:
Added more pattern implementations.

TODO:

1. Inbound Gateways
2. Condition Outbound gateway/channel adapter

artembilan:
From my point of view it is ready for merge for UI project.
We may consider to add more IEP indicators in other components, like `MessageStore`.
We don't have a place to add a hook for `Idempotent Receiver` (although just an `IdempotentReceiverInterceptor` could have that) and for `Routing Slip` which is just a header in the message.
All of those could be done on demand.

Anyway before merging I would like to discuss why should we treat a `@MessagingGateway` differently from  any other `Inbound Gateway` implementation. It just does exactly the same: wraps an inbound data into a `Message` and sends it into a channel.
So, from the UI perspective it really looks the same. Not sure why should we extract it into a separate integration pattern and also introduce a new UI icon for it...

Thanks

artembilan:
However I see that there enough attention is paid for that pattern in the book: https://www.enterpriseintegrationpatterns.com/patterns/messaging/MessagingGateway.html ...

artembilan:
OK. I've changed my mind: will show a `@MessagingGateway` as a `gateway` pattern...

spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
```diff
@@ -0,0 +1,161 @@
...
+				service_activator,
+				outbound_channel_adapter,
+				inbound_channel_adapter,
+				outbound_gateway,
+				inbound_gateway,
+				gateway,
+				bridge),
+
+		message_routing(
+				splitter,
```
garyrussell: This is tricky - a splitter is not really routing in any way, but maybe it can't be considered a transformer either, although it does transform one message into many.

Maybe it needs its own category?

spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
```diff
@@ -0,0 +1,161 @@
...
+				filter,
+				aggregator,
```
garyrussell: ditto

maybe splitter. aggregator, scatter_gather, resequencer in a new category?

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractReplyProducingMessageHandler.java
```diff
@@ -91,6 +92,11 @@ protected ClassLoader getBeanClassLoader() {
 		return this.beanClassLoader;
 	}
 
+	@Override
+	public IntegrationPatternType getIntegrationPatternType() {
+		return IntegrationPatternType.outbound_gateway;
```
garyrussell: This is not intuitive; I would expect the default to be service_activator; I am guessing you did this to avoid having to patch every outbound gateway, which is ok, but perhaps a comment here to explain why this is the default might be useful.

garyrussell (commented):
Good work; just a few comments for your consideration - also perhaps `IntegrationGraphServerTests` needs to be enhanced to check the graph is populated properly?

spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
artembilan: I placed it into a category where it belongs in EIP: https://www.enterpriseintegrationpatterns.com/patterns/messaging/Sequencer.html

> Messaging Patterns » Message Routing » Splitter

spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
artembilan: DITTO

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractReplyProducingMessageHandler.java
artembilan: Right. I did an opposite: whenever it is not an `outbound gateway`, I did an explicit type: `service activator`, `transformer`, `splitter` etc.

artembilan:
> also perhaps `IntegrationGraphServerTests` needs to be enhanced to check the graph is populated properly?

What do you mean, Gary?

I did some changes over there to be sure that new attributes are populated.
Perhaps not all of them, of course, but still...

Thanks

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractReplyProducingMessageHandler.java
garyrussell: Right, but I think it warrants a comment (not java doc, just `/* ... */`.

garyrussell:
Sorry; I somehow skipped over the test.

spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
garyrussell: OK

artembilan:
Hm.. I've looked into a docs for this and it turns out that we definitely need bump graph version - `private static final float GRAPH_VERSION = 1.2f;`, because of these new attributes.
So, I'm finishing the task anyway, but again: it sounds more like a feature for next minor version - `5.3`.

Any other thoughts?

garyrussell:
No more thoughts; we should get this to the tools team sooner rather than later, though.

artembilan:
OK. I've this change to version `5.3`. 
Will update *What's New* when the respective branch is ready.

artembilan:
OK. So, this is rebased onto `5.3-WIP`. Ready for final review and merge.
Therefore this new feature is going to be available in the published `BUILD-SNAPSHOT` for version `5.3.0`.

Thanks

src/reference/asciidoc/whats-new.adoc
```diff
@@ -15,6 +15,12 @@ If you are interested in more details, see the Issue Tracker tickets that were r
 [[x5.3-new-components]]
 === New Components
 
+[[x5.2-rsocket-support]]
+==== Integration Pattern
```
garyrussell: 5.3 - will fix on merge after travis is green.

artembilan: Doh... And not `rsocket-support` 😄 

Thanks

garyrussell: Also not rsocket - fixed.

src/reference/asciidoc/graph.adoc
```diff
@@ -251,6 +257,9 @@ The preceding gateway produces nodes similar to the following:
 You can use this  `IntegrationNode` hierarchy for parsing the graph model on the client side as well as to understand the general Spring Integration runtime behavior.
 See also <<./overview.adoc#programming-tips,Programming Tips and Tricks>> for more information.
 
+The version 5.3 has introduced an `IntegrationPattern` abstraction and all out-of-the-box components, which belongs to particular EI pattern implements this abstraction and provides an `IntegrationPatternType` enum value.
+This information can be useful for some categorizing logic in the target application or being exposed into the graph node it can be used on UI to determine how to draw the provided component.
+
 === Integration Graph Controller
```
garyrussell: Merged as 71224bc1c6738c1f4eb877c0d3d4763c94b0eb02 with minor doc polishing and added new elements to the graph example above this paragraph.

PR #3094, opened 2019-10-30
GH-3090: Add `logout() to `FtpSession.close()`

### LINKED ISSUE #3090: FTP Session not correctly closed
**Affects Version(s):** \ 4.3.20-RELEASE

Hi, we're using spring-integration for uploading FTP files to a server of ours.
Due to some unrelated problem we took a TCP dump on the FTP traffic and we could identify an issue in spring-integration which I'd like to report here.

Basically the FTP session is not closed at all but just the connection is closed. See

https://github.com/spring-projects/spring-integration/blob/3cd751a8f3656a19deabc8eaa7ef2e96319153c7/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java#L146

This causes the following interaction
spring-integration

I would suggest to add method call logout() in the Session's close() implementation.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3090

Without `logout()` the FTP session is not closed at all,
but just the connection is closed.
Some FTP servers close those sessions eventually anyway, but some just
leak with resources.

**Cherry-pick to 5.1.x & 4.3.x**

### DIFF
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
@@ -154,6 +154,7 @@ public class FtpSession implements Session<FTPFile> {
 			if (this.readingRaw.get() && !finalizeRaw() && LOGGER.isWarnEnabled()) {
 				LOGGER.warn("Finalize on readRaw() returned false for " + this);
 			}
+			this.client.logout();
 			this.client.disconnect();
 		}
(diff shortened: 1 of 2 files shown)


artembilan:
Cherry-pick only the first commit.

garyrussell:
Cherry-picked first commit to 5.1.x as 315fafdaf2e4caa94fae1aa5314ad1c336caf9fe and to 4.3.x as 127f683952d2f4e0eaa47b189d577e7190c37186

PR #3099, opened 2019-11-01
Add back-pressure tests for SubscribableChannelPublisherAdapter

bsideup:
As discovered while pairing with @olegz, `SubscribableChannel` reactive adapter does not respect back-pressure.

There are two problems, actually:
1. The requests are ignored due to the use of `FluxSink.OverflowStrategy.IGNORE`
2. Even if you change it to `FluxSink.OverflowStrategy.BUFFER`, the buffer will be unbounded and eventually lead to OOM (see https://github.com/reactor/reactor-core/issues/1936)

It is highly recommended to use `EmitterProcessor` instead of `Flux.create` in the adapter, so that you push until the buffer is full, and it will block the `SubscribableChannel`'s callback when calling `onNext` until the queue gets more capacity to handle new messages.

See [this gist](https://gist.github.com/bsideup/2a75b8f95096ef74d05546e8cef6a749) if you want to play with `Flux.create` vs `EmitterProcessor`.

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/channel/MessageChannelReactiveUtilsTest.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/channel/MessageChannelReactiveUtilsTest.java
@@ -16,6 +16,10 @@
 
 package org.springframework.integration.channel;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.time.Duration;
+
 import org.junit.Test;
 
 import org.springframework.messaging.SubscribableChannel;
@@ -26,10 +30,6 @@ import reactor.core.Disposables;
 import reactor.core.scheduler.Schedulers;
 import reactor.test.StepVerifier;
 import reactor.util.concurrent.Queues;
-
-import java.time.Duration;
-
-import static org.assertj.core.api.Assertions.assertThat;
 
 public class MessageChannelReactiveUtilsTest {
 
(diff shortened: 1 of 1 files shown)


olegz:
Linking to https://github.com/spring-cloud/spring-cloud-stream/issues/1835 as an originating issue

artembilan:
I'm working now on the fix in Spring Integration.
PR - soon.

Thank you for tests, @bsideup ! 😉 

I think when it is done in Spring Integration, we don't need these tests and any fixes here in SCSt - only upgrade to the proper SI version.

bsideup:
@artembilan I would highly recommend to keep these tests (and cover every new publisher you create) to avoid regressions :)

Although this line made me think that you confused the repos :D
> we don't need these tests and any fixes here in SCSt - only upgrade to the proper SI version.

BTW feel free to merge them into your WIP branch where you're working on migrating to `EmitterProcessor` 👍

artembilan:
> that you confused the repos

Doh. You made a change exactly in Spring Integration. 
:shame: 

Will merge this and re-use in my work-in-progress 😉

bsideup:
@artembilan 👍 be aware that I messed up the checkstyle config tho 😂 

BTW how do you run it locally? I get "Unable to create Root Module" error in Gradle o_O

artembilan:
Not sure what is going on for you locally. We just use `gradlew clean check`.
Since you did change only in a core module, it is enough for you to run against that one: `gradlew clean :spring-integration-core:check`

artembilan:
No worries, I've just merged it and will fix Checkstyle locally if that.

Thanks

bsideup:
@artembilan thank you! 👍 

I think the last commit fixed everything, and the Travis has reported a red status due to the actual problem (to fix one, you would use `.BUFFER`, and other can be fixed by migrating to `EmitterProcessor`)

PR #3104, opened 2019-11-06
Use EmitterProcessor in the FluxMessageChannel

artembilan:
The `EmitterProcessor` has a good logic to block upstream producer
when its downstream subscriber cannot keep up with overproducing.

* Rework `FluxMessageChannel` logic to rely on the `EmitterProcessor`
instead of `Flux.create()`
* Cancel `FluxMessageChannel` internal subscriptions in the `destroy()`
* Fix `ReactiveStreamsTests.testFluxTransform()` for the splitter's
delimiter
* Ensure in the `FluxMessageChannelTests.testFluxMessageChannel`
that we can have several concurrent subscribers to the
`FluxMessageChannel`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
@@ -16,7 +16,7 @@
 
 package org.springframework.integration.channel;
 
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
@@ -27,7 +27,8 @@ import org.springframework.util.Assert;
 import reactor.core.publisher.EmitterProcessor;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.FluxSink;
-import reactor.core.publisher.Mono;
+import reactor.core.publisher.ReplayProcessor;
+import reactor.core.scheduler.Schedulers;
 
 /**
  * The {@link AbstractMessageChannel} implementation for the
@@ -35,28 +36,30 @@ import reactor.core.publisher.Mono;
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Sergei Egorov
  *
  * @since 5.0
  */
 public class FluxMessageChannel extends AbstractMessageChannel
 		implements Publisher<Message<?>>, ReactiveStreamsSubscribableChannel {
 
-	private final AtomicInteger subscribed = new AtomicInteger();
+	private final EmitterProcessor<Message<?>> processor;
 
-	private final EmitterProcessor<Message<?>> flux;
+	private final FluxSink<Message<?>> sink;
 
-	private final EmitterProcessor<Integer> subscriptionDelay = EmitterProcessor.create(false);
+	private final AtomicBoolean subscribed = new AtomicBoolean();
 
-	private FluxSink<Message<?>> sink;
+	private final ReplayProcessor<Boolean> subscribedSignal = ReplayProcessor.create(1);
 
 	public FluxMessageChannel() {
-		this.flux = EmitterProcessor.create(1, false);
-		this.sink = this.flux.sink();
+		this.processor = EmitterProcessor.create(1, false);
+		this.sink = this.processor.sink(FluxSink.OverflowStrategy.BUFFER);
+		this.subscribedSignal.doOnNext(this.subscribed::set).subscribe();
 	}
 
 	@Override
 	protected boolean doSend(Message<?> message, long timeout) {
-		Assert.state(this.subscribed.get() > 0,
+		Assert.state(this.subscribed.get(),
 				() -> "The [" + this + "] doesn't have subscribers to accept messages");
 		this.sink.next(message);
 		return true;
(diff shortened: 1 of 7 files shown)


artembilan:
@bsideup ,

Would you mind to take a look into this fix, please?

Thanks

bsideup (changes requested):
(here was a comment but I missed an important detail so I am updating it :D)

spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
```diff
@@ -47,15 +49,15 @@
 
 	private final Map<Publisher<? extends Message<?>>, ConnectableFlux<?>> publishers = new ConcurrentHashMap<>();
 
-	private final Flux<Message<?>> flux;
+	private final Map<ConnectableFlux<?>, Disposable> disposables = new ConcurrentHashMap<>();
```
bsideup: Consider using `Disposables.composite()`

spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
```diff
@@ -70,11 +72,12 @@ protected boolean doSend(Message<?> message, long timeout) {
 	public void subscribe(Subscriber<? super Message<?>> subscriber) {
 		this.subscribers.add(subscriber);
 
-		this.flux.doOnCancel(() -> this.subscribers.remove(subscriber))
+		this.flux.doFinally((signal) -> this.subscribers.remove(subscriber))
 				.retry()
 				.subscribe(subscriber);
 
-		this.publishers.values().forEach(ConnectableFlux::connect);
+		this.publishers.values()
```
bsideup: could you please explain the use of `ConnectableFlux` here?

spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
```diff
@@ -84,14 +87,22 @@ public void subscribeTo(Publisher<? extends Message<?>> publisher) {
...
-			connectableFlux.connect();
+			connectableFlux.connect((disposable) -> this.disposables.put(connectableFlux, disposable));
 		}
 	}
 
+	@Override
+	public void destroy() {
+		super.destroy();
+
+		this.subscribers.forEach(Subscriber::onComplete);
```
bsideup:
```suggestion
		this.flux.onComplete();
```

spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
artembilan: I need `ConnectableFlux` to subscribe to provideded publishers in lazy-manner: if this `FluxMessageChannel` is subscribed already, I do `connect()` - otherwise I `connect()` to all the provided publishers in its `subscribe()`.
My point is to be able to propagate events from upstream publishers into subscribers of this `FluxMessageChannel` honoring an internal logic of the `MessageChannel` implementation.

I couldn't find any other way in Reactor how to achieve a lazy subscription task unless this `ConnectableFlux`.

And with that it looks like I can't use `Disposables.composite()` since I don't have a way to remove one from that composite when such an upstream published is finished.

I'm OK to change the logic if you can help with that.

Thanks

P.S. Pushed `flux.onComplete()` BTW and removed reference to subscribers since it is not necessary any more.
👍

artembilan:
I think I found a solution with the `delaySubscription(Publisher<U> subscriptionDelay)`.

Stay tuned! Will come back to you today 😄

artembilan:
So, here you are a fix based on the `delaySubscription()` and some additional internal `EmitterProcessor` to handle external subscriptions to this `FluxMessageChannel`.

spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
```diff
@@ -43,55 +41,53 @@
...
-		}
+		Flux.from(publisher)
+				.handle((message, sink) -> sink.next(send(message)))
+				.onErrorContinue((throwable, event) ->
+						logger.warn("Error during processing event: " + event, throwable))
+				.delaySubscription(
+						Mono.fromSupplier(this.subscribed::get)
+								.filter((subscribers) -> subscribers > 0)
+								.switchIfEmpty(this.subscriptionDelay.next()))
+				.subscribe();
```
artembilan: I think the proper way would be to have a `subscribeTo()` to return `Mono<Void>` instead to let that caller to subscribe or compose further its reactive flow.
This way we would defer an emission from this source publisher as late as possible.

But I'd like to revise that stuff in a separate PR.

Thanks

artembilan:
@bsideup , gentle ping. 😄

artembilan:
OK! I have minimized the change for this PR as much as possible - 2 files. 😄 

Looking forward for your feedback!

artembilan:
OK. I pushed what we have discussed with some workaround for subscription race condition.

Thank you for your feedback, @bsideup, looking forward for more!

artembilan:
@bsideup ,

so, I have rebased to `master` and decreased a change as much as possible. Again 😄 

Would be great to have some feedback before release this Wednesday.

Thanks

artembilan:
Pushed the fix as we discussed in private chat.

Thanks

bsideup (approved):

spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
```diff
@@ -16,82 +16,81 @@
...
-
-		this.flux.doOnCancel(() -> this.subscribers.remove(subscriber))
-				.retry()
+		this.processor
+				.doFinally((s) -> this.subscribedSignal.onNext(this.processor.hasDownstreams()))
 				.subscribe(subscriber);
-
-		this.publishers.values().forEach(ConnectableFlux::connect);
+		if (this.processor.hasDownstreams()) {
+			this.subscribedSignal.onNext(true);
```
bsideup: probably `this.subscribedSignal.onNext(this.processor.hasDownstreams())` would be better

artembilan: He-he! True. Will fix soon.

artembilan:
@garyrussell ,

We're done here with @bsideup . So, let us know if you are OK and we merge (or you).

Thanks

Issue #3105, opened 2019-11-07
Spring integration stream SFTP: pattern filter not working


lukasGemela:
Hi guys, 

as follow up on this issue

https://stackoverflow.com/questions/58752135/spring-integration-stream-sftp-pattern-filter-not-working/58753207#58753207

 I'm raising this ticket.
I have used Spring integration 5.2.0

artembilan:
@lukasGemela ,

thank you for checking that functionality and sorry that we brought to you some pain with those unexpected files 😄 

I think as a workaround you can inject your own:
```
private CompositeFileListFilter<LsEntry> composeFilters(FileListFilter<LsEntry> fileListFilter) {
		CompositeFileListFilter<LsEntry> compositeFileListFilter = new CompositeFileListFilter<>();
		compositeFileListFilter.addFilters(fileListFilter,
				new SftpPersistentAcceptOnceFileListFilter(new SimpleMetadataStore(), "sftpStreamingMessageSource"));
		return compositeFileListFilter;
	}
```
with overridden `supportsSingleFileFiltering()` to return always `false`, so all those files are going to be filtered just after a fetch, not during processing each of them.

On the other hand let us know if you are OK with contribution: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc

I guess what we need is a `continue;` in the block:
```
if (this.toBeReceived.size() > 0) { // don't re-fetch already filtered files
	file = poll();
}
```

Thank you again!

lukasGemela:
@artembilan Hi, thank you for a workaround!

`continue;` in that block makes sense I think, maybe also break in 

```
else {
     file = null;
}
```

instead of assignment

artembilan:
Yes, that makes sense too.

So, would you mind to raise a PR and we will continue discussion in the code already?

lukasGemela:
@artembilan sure PR on the way

PR #3106, opened 2019-11-07
GH-3105: Spring integration stream SFTP: pattern filter not wrking

lukasGemela:
https://github.com/spring-projects/spring-integration/issues/3105

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
@@ -49,6 +49,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Lukas Gemela
  *
  * @since 4.3
  *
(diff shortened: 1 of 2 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -190,9 +190,10 @@ protected Object doReceive() {
 
 					if (this.toBeReceived.size() > 0) { // don't re-fetch already filtered files
 						file = poll();
+						continue;
 					}
 					else {
-						file = null;
+						break;
```
artembilan: Grate!

I think we don't need an `if (file != null)` any more after this.

artembilan (changes requested):
Please, add your name to the `@author` list.
And also it would be great to have some test to cover this functionality.

I guess if we would have had such a test before we wouldn't have this bug at all 😢 

Thank you for your contribution anyway!

spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
lukasGemela: Sure thing will do tomorrow

lukasGemela: @artembilan test added

artembilan (approved):
Grate!

Pulling locally for final review and merge after code clean up.

👍

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/1820c05c26ad625d449d49b867a72f8d1cf97571.

@lukasGemela ,

PR #3108, opened 2019-11-12
GH-3107: Add errorOnTimeout for Inbound Gateways

### LINKED ISSUE #3107: Add a reply timeout hook to TCP inbound gateway
Allow the user to determine the action after a reply times out.

https://stackoverflow.com/questions/58796500/tcpinboundgateway-server-configuration-on-timeout-a-custom-message-to-be-sent

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3107

The `MessagingGatewaySupport` has an `errorOnTimeout` option to throw
a `MessageTimeoutException` when downstream reply doesn't come back in
time for configured reply timeout

* Expose an `errorOnTimeout` option as a `TcpInboundGateway` ctor
property
* Add new factory methods into a `Tcp` factory for Java DSL
* Ensure a property works as expected in the `IpIntegrationTests`
* Document a new option

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
@@ -49,32 +49,54 @@ import org.springframework.util.Assert;
  * inbound / outbound channel adapters should be used.
  *
  * @author Gary Russell
+ * @author Artem Bilan
+ *
  * @since 2.0
  */
 public class TcpInboundGateway extends MessagingGatewaySupport implements
 		TcpListener, TcpSender, ClientModeCapable, OrderlyShutdownCapable {
 
-	private volatile AbstractServerConnectionFactory serverConnectionFactory;
+	/**
+	 * A default retry interval in milliseconds - {@value #DEFAULT_RETRY_INTERVAL}.
+	 */
+	public static final long DEFAULT_RETRY_INTERVAL = 60000;
 
-	private volatile AbstractClientConnectionFactory clientConnectionFactory;
+	private final Map<String, TcpConnection> connections = new ConcurrentHashMap<>();
 
-	private final Map<String, TcpConnection> connections = new ConcurrentHashMap<String, TcpConnection>();
+	private final AtomicInteger activeCount = new AtomicInteger();
 
-	private volatile boolean isClientMode;
+	private AbstractServerConnectionFactory serverConnectionFactory;
 
-	private volatile boolean isSingleUse;
+	private AbstractClientConnectionFactory clientConnectionFactory;
 
-	private volatile long retryInterval = 60000;
+	private boolean isClientMode;
 
-	private volatile ScheduledFuture<?> scheduledFuture;
+	private boolean isSingleUse;
 
-	private volatile ClientModeConnectionManager clientModeConnectionManager;
+	private long retryInterval = DEFAULT_RETRY_INTERVAL;
 
 	private volatile boolean active;
 
+	private volatile ClientModeConnectionManager clientModeConnectionManager;
+
+	private volatile ScheduledFuture<?> scheduledFuture;
+
 	private volatile boolean shuttingDown;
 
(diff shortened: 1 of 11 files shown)


artembilan:
I wonder if we should just go ahead and do the same for all the Inbound Gateway implementations...

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
```diff
@@ -117,7 +139,7 @@ public boolean onMessage(Message<?> message) {
 	}
 
 	private boolean doOnMessage(Message<?> message) {
-		Message<?> reply = this.sendAndReceiveMessage(message);
+		Message<?> reply = sendAndReceiveMessage(message);
```
garyrussell: Need to catch and swallow `MessageTimeoutException` to retain existing behavior when no EC or error flow returns `null`.

garyrussell (changes requested):
Yes, we should probably do it for all in gates - this PR? Or open another "backlog" issue?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
artembilan: Well, what we have here is fully a new behavior which doesn't break an existing one by default because all the existing `TcpInboundGateways` are created without that new `errorOnTimeout` ctor.
Also see an explanation of that option in the super class:
```
  If no error channel is
	 * configured, the {@link MessageTimeoutException} will be thrown.
```

So, what you are asking is fully redundant since it is related only to new option and we definitely should throw that `MessageTimeoutException` according contact. And again: we do that only if `errorOnTimeout` is turned on, which is not by default.

Does it make sense?

artembilan:
> do it for all in gates - this PR

Yes, I can play with that in this PR: doesn't look like we are in hurry to release this fix.
Although I'd follow your review preferences.
If that easier for you to merge this and go ahead with other in other ticket, I can do that as well.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
garyrussell: Good point; I was concerned about generating ERROR logs in the caller to `onMessage` but, as you say, they have to opt-in by setting the boolean, in which case an error log is appropriate anyway.

garyrussell:
I don't mind if you add more here.

artembilan:
OK. So, pushed a general solution to configure any inbound gateway for `errorOnTimeout`.

src/reference/asciidoc/endpoint-summary.adoc
```diff
@@ -223,3 +223,6 @@ The `<int:outbound-channel-adapter>` element lets you send data to a `void` meth
 As discussed in <<./gateway.adoc#gateway,Messaging Gateways>>, the `<int:gateway>` element lets any Java program invoke a messaging flow.
 Each of these works without requiring any source-level dependencies on Spring Integration.
 The equivalent of an outbound gateway in this context is using a service activator (see <<./service-activator.adoc#service-activator,Service Activator>>) to invoke a method that returns an `Object` of some kind.
+
+Starting with version `5.2.2`, all the inbound gateway can be configured with an `errorOnTimeout` boolean flag to throw a `MessageTimeoutException` when downstream reply doesn't come back during reply timeout.
```
garyrussell: `s/gateway/gateways/`

...when **the** downstream flow doesn't return a reply during the reply timeout.
The timer is not started until the thread returns the gateway so usually is only useful when the downstream flow is asynchronous.

garyrussell (changes requested):
LGTM with doc polish. Let's see if Travis agrees 😄

garyrussell (approved):

garyrussell:
Merged as 72f7c723920d0e10f5d6ac38f4231b61eabe52ec with a bit more polishing.

PR #3109, opened 2019-11-14
INT-3494: Resolve dir for writing as a Resource

artembilan:
JIRA: https://jira.spring.io/browse/INT-3494

The expression for local directory can be resolved into a `Resource`
or resource location.
* Fix `ExpressionUtils.expressionToFile()` to support `Resource` and
also use `ResourceUtils.getFile(path)` when expression result is a string
* Modify tests to ensure that resource is resolved properly
* Upgrade affected tests to JUnit 5
* Mention an new functionality in docs

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionUtils.java
@@ -165,12 +165,12 @@ public final class ExpressionUtils {
 	 * @param expression the expression.
 	 * @param evaluationContext the evaluation context.
 	 * @param message the message (if available).
-	 * @param name the name of the result of the evaluation.
+	 * @param propertyName the property name the expression is evaluated for.
 	 * @return the File.
 	 * @since 5.0
 	 */
 	public static File expressionToFile(Expression expression, EvaluationContext evaluationContext,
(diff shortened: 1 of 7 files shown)


spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionUtils.java
```diff
@@ -168,29 +172,49 @@ private static EvaluationContext createEvaluationContext(@Nullable ConversionSer
 	public static File expressionToFile(Expression expression, EvaluationContext evaluationContext,
 			@Nullable Message<?> message, String name) {
```
garyrussell: We should take this opportunity to improve the javadocs for the `name` parameter.

`@param name the name of the result of the evaluation.`

`@param name the name of the property we are evaluating an expression for` ?

garyrussell (changes requested):
I think, in future, unless there are only one or two tests, we should perform JUnit5 conversion in a separate PR (or at least a separate commit) to make it easier to review.

artembilan:
> or at least a separate commit

Yeah... That's my flaw: I can't stay away if I see that code should be refactored 😄 

Will work on that habit for future tasks!

garyrussell (approved):

PR #3115, opened 2019-11-22
GH-3114: Honor SpEL contract in ExpressionEvalMap

### LINKED ISSUE #3114: NULL HttpEntity in headerFunction in HTTP inbound gateway
When trying to upgrade from version 5.1.8 to 5.2.1, our HTTP inbound gateway started throwing NPEs when evaluating header functions. Our code is quite basic:
```
IntegrationFlows.from(
                        inboundGateway( gatewayPath )
                                .headerFunction(SOME_HEADER, ent -> getSomeHeaderValue(ent ) )  // ent is null
```

Seems to be related to the call `FunctionExpression.getValue(...)` which ignores `EvaluationContext.getRootObject()`

Do you need more info on this?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3114

The contract of SpEL with its
`getValue(EvaluationContext context, @Nullable Object rootObject)` is
that we need to deal with provided `rootObject` even if it is `null`
and don't consult with `context.getRootObject()`

* Fix `ExpressionEvalMap` to have an internal `rootExplicitlySet`
to indicate that `root` explicitly provided by consumer, even if it is null.
According this flag call respective `Expression.getValue()`
* Add `@Nullable` to methods and their arguments into `ExpressionEvalMap`
& `FunctionExpression` to honor `Expression` contracts
* Populate an `HttpEntity` explicitly into `ExpressionEvalMap` from the
`HttpRequestHandlingEndpointSupport` and `WebFluxInboundEndpoint` for
full picture

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
@@ -47,13 +47,13 @@ import org.springframework.util.Assert;
  * <pre class="code">
  * {@code
  *ExpressionEvalMap evalMap = ExpressionEvalMap
- *	.from(expressions)
- *	.usingCallback(new EvaluationCallback() {
- *		Object evaluate(Expression expression) {
- *			// return some expression evaluation
- *        }
+ *  .from(expressions)
+ *  .usingCallback(new EvaluationCallback() {
+ *        Object evaluate(Expression expression) {
+ *	            // return some expression evaluation
+ *         }
  *    })
- *	.build();
+ *    .build();
  *}
  * </pre>
  * <p>
--- a/spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
@@ -51,8 +51,8 @@ import org.springframework.util.Assert;
  *	.usingCallback(new EvaluationCallback() {
  *		Object evaluate(Expression expression) {
  *			// return some expression evaluation
- *		}
- *	})
+ *        }
+ *    })
  *	.build();
  *}
  * </pre>
@@ -84,6 +84,7 @@ public final class ExpressionEvalMap extends AbstractMap<String, Object> {
 	 * from {@link #original} and returns the result of evaluation using {@link #evaluationCallback}.
 	 */
 	@Override
+	@Nullable
 	public Object get(Object key) {
 		Object value = this.original.get(key);
 		if (value != null) {
@@ -106,9 +107,9 @@ public final class ExpressionEvalMap extends AbstractMap<String, Object> {
 
 	@Override
 	public Set<Map.Entry<String, Object>> entrySet() {
-		return this.original.entrySet()
+		return this.original.keySet()
 				.stream()
-				.map(e -> new SimpleImmutableEntry<>(e.getKey(), get(e.getKey())))
+				.map((key) -> new SimpleImmutableEntry<>(key, get(key)))
 				.collect(Collectors.toSet());
 	}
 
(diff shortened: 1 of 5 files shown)


spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
```diff
@@ -51,8 +51,8 @@
  *	.usingCallback(new EvaluationCallback() {
  *		Object evaluate(Expression expression) {
  *			// return some expression evaluation
- *		}
- *	})
+ *        }
+ *    })
  *	.build();
```
garyrussell: Change from tabs to spaces has messed up alignment.

spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
```diff
@@ -206,22 +207,35 @@ public static ExpressionEvalMapBuilder from(Map<String, ?> expressions) {
...
+		@Nullable
 		private final Class<?> returnType;
 
-		public ComponentsEvaluationCallback(EvaluationContext context, Object root, Class<?> returnType) {
+		public ComponentsEvaluationCallback(@Nullable EvaluationContext context, @Nullable Object root,
+				boolean rootExplicitlySet, @Nullable Class<?> returnType) {
+
 			this.context = context;
 			this.root = root;
+			this.rootExplicitlySet = rootExplicitlySet;
```
garyrussell: It's not clear what the purpose of this is; we still get an NPE if the root has explicitly been set to `null`.

>is that we need to deal with provided rootObject even if it is `null`

When I changed the HRHES to `.withRoot(null)` BOOM!!

garyrussell (changes requested):
See comments; need a test with explicit `null` root.

spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
artembilan: That's correct: it is exactly what SpEL contract does in the `getValue(EvaluationContext context, @Nullable Object rootObject)`.
If you explicitly provide a root as `null`, then it is used as is, without any consultation with `context`.

So, we fix an NPE on our level because we really have a root, but we still don't break SpEL contract when root is explicitly provided as `null`

spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
artembilan: Doesn't look like, but I've changed everything into spaces. I don't think tabs are good for JavaDocs.

spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
garyrussell: So you are saying the NPE is ok,  because it's in "user" code (in this case, the lambda) so the user is responsible for a null check?

spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
garyrussell: Right, but you had mixed tabs and spaces here. The indentation is still wrong (extra space before `}`). Will fix on merge.

spring-integration-core/src/main/java/org/springframework/integration/expression/ExpressionEvalMap.java
artembilan: That's right if (s)he provides a `null` explicitly, but in our case we just miss the proper root to propagate.
See all the `Expression.getValue()` with `root` - it is `@Nullable`. Therefore without checking for `null` it is really going to lead into NPE.
But in our case we provide all the stuff for expression environment, therefore it is safe not check because we guarantee now a non-null root.
See changes in those HTTP components.

garyrussell:
Merged as 516ecbcf8b

PR #3116, opened 2019-12-03
Add Reactive Channel Mono subscribeToUpstream()

artembilan:
It would be better to let  `ReactiveStreamsSubscribableChannel` to
return some Reactor type for its `subscribeTo` instead of plain `void`.
This way we can compose reactive streams for deferred subscription
instead of intermediate subscribe inside a `FluxMessageChannel`

* Introduce a `Mono<Void> subscribeToUpstream(Publisher)` into the
`ReactiveStreamsSubscribableChannel` as a `default` method to avoid
breaking changes and call existing deprecated now
`void subscribeTo(Publisher)`
* Fix `FluxMessageChannel` to implement a new `subscribeToUpstream`
and call this one from the deprecated `subscribeTo`
* Fix all the `subscribeTo` usage in favor of newly introduced
`subscribeToUpstream`.

The fix doesn't bring too much subscription gain in the framework
(unless `MessagingGatewaySupport.sendAndReceiveMessageReactive()`
changes), but can be useful in target application when reactive streams
composition is important even if we place a `FluxMessageChannel` in
between

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/ReactiveStreamsSubscribableChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/ReactiveStreamsSubscribableChannel.java
@@ -20,7 +20,12 @@ import org.reactivestreams.Publisher;
 
 import org.springframework.messaging.Message;
 
+import reactor.core.publisher.Mono;
+
 /**
+ * Reactive Streams specific message channel for composing upstream {@link Publisher}s
+ * into an on-demand processing in this channel implementation.
+ *
  * @author Artem Bilan
  * @author Gary Russell
  *
@@ -28,6 +33,26 @@ import org.springframework.messaging.Message;
  */
 public interface ReactiveStreamsSubscribableChannel {
 
+	/**
+	 * Subscribe to the provided {@link Publisher}.
+	 * @param publisher the {@link Publisher} to subscribe to.
+	 * @deprecated in favor of {@link #subscribeToUpstream}.
+	 * Framework doesn't call this method internally any more.
+	 */
+	@Deprecated
 	void subscribeTo(Publisher<? extends Message<?>> publisher);
+
+	/**
+	 * Compose with the provided upstream {@link Publisher} for processing its events in this channel.
+	 * It is recommended to implement this method for possible reactive streams composition.
+	 * @param upstreamPublisher the upstream {@link Publisher} to subscribe to.
+	 * @return the {@link Mono} for reactive streams composition.
+	 * @since 5.2.2
+	 */
+	@SuppressWarnings("deprecagtion")
+	default Mono<Void> subscribeToUpstream(Publisher<? extends Message<?>> upstreamPublisher) {
+		subscribeTo(upstreamPublisher);
+		return Mono.empty();
+	}
 
 }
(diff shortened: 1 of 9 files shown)


artembilan:
Gary, this is more for your review in regards of breaking changes.
There is no too many changes in the framework unless we move a `subscribe()` internaly to the `Publisher` outside of `FluxMessageChannel`.

I admit that this is a "hard" change for contract, but I'd rather made it right now as a deprecated one and remove it in the next `5.3` version, than deprecate it there and wait more time for the next release.

Anyway let me know if I need to ask Reactor team for review as well.
I guess we have some time before tomorrow's release 😄 

Thanks

garyrussell:
LGTM, but I am surprised there are no tests.

You should have the Reactor team have a look.

artembilan:
Well, as I said: only internal changes - moving `subscribe()` outside of `FluxMessageChannel`.
This way is it enough for me to be sure that with this change all the existing tests pass.

It's hard right now for me to come up with the test which would fail before, but pass now...
But yeah, for more coverage it would be better to have one.
I'll sleep with that a little bit 😄

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -653,11 +652,12 @@ private void sendMessageForReactiveFlow(MessageChannel requestChannel, Message<?
 							? requestChannel.send(requestMessage, sendTimeout)
 							: requestChannel.send(requestMessage);
 
-			if (!sent) {
-				throw new MessageDeliveryException(requestMessage,
-						"Failed to send message to channel '" + requestChannel +
-								"' within timeout: " + sendTimeout);
-			}
+			return Mono.just(sent)
```
bsideup: please use standard `if (!sent) { return Mono.error(...); } else { return Mono.empty(); }`

artembilan:
Thanks, @bsideup , for review!

Anything else should be fixed here, please?
Or we are good to go with merging already?

spring-integration-core/src/main/java/org/springframework/integration/channel/ReactiveStreamsSubscribableChannel.java
```diff
@@ -20,14 +20,39 @@
...
+	/**
+	 * Compose with the provided upstream {@link Publisher} for processing its events in this channel.
+	 * It is recommended to implement this method for possible reactive streams composition.
+	 * @param upstreamPublisher the upstream {@link Publisher} to subscribe to.
+	 * @return the {@link Mono} for reactive streams composition.
+	 * @since 5.2.2
+	 */
+	@SuppressWarnings("deprecagtion")
+	default Mono<Void> subscribeToUpstream(Publisher<? extends Message<?>> upstreamPublisher) {
+		subscribeTo(upstreamPublisher);
```
bsideup: I would highly recommend to subscribe only when the returned Mono is subscribed, otherwise you're creating a hot publisher

artembilan:
Closing as redundant (for now) and a feature without a justification the reason to expose such an API.
In most cases it turns out that this `subscribeTo` contract becomes an event source for reactive stream, so direct `subscribe()` in the `FluxMessageChannel` is fully OK.
Plus we have a single place of lifecycle control for those sources (which still need to be improved) instead of spreading such a logic throughout the framework.

Thanks, @bsideup , for deep review! 😄

PR #3117, opened 2019-12-06
Refactor AbstractMessageHandler and implements MongoDb MessageHandler

dturanski:
Addresses https://jira.spring.io/browse/INT-4568 by introducing `AbstractReactiveMessageHandler` in core and  implementing`ReactiveMongoDbStoringMessageHandler` .


spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractReactiveMessageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+
+	@Override
+	public Mono<Void> handleMessage(Message<?> message) {
+		Assert.notNull(message, "Message must not be null");
+		Assert.notNull(message.getPayload(), "Message payload must not be null"); // NOSONAR - false positive
+		if (this.loggingEnabled && this.logger.isDebugEnabled()) {
+			this.logger.debug(this + " received message: " + message);
+		}
+		//TODO: Can't capture metrics the same way in this context since returning a Mono<>
+		org.springframework.integration.support.management.MetricsContext start = null;
```
artembilan: Let's do without metrics for now!
They are going to be removed anyway in favor of Micrometer. And I think there is something in Reactor to support them over context: https://projectreactor.io/docs/core/release/reference/#metrics

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractReactiveMessageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+		if (this.countsEnabled && this.metricsCaptor != null) {
+			sample = this.metricsCaptor.start();
+		}
+		try {
+			if (this.shouldTrack) {
+				message = MessageHistory.write(message, this, getMessageBuilderFactory());
+			}
+			return handleMessageInternal(message);
+		}
+		catch (Exception e) {
```
artembilan: I think there is no exception thrown possible because of `Mono<Void>`, so all the error must go in the reactive stream.
Therefore maximum we can do here is like `doOnError()` on the returned `Mono`

spring-integration-core/src/test/java/org/springframework/integration/handler/ReactiveMessageHandlerTests.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.springframework.integration.channel.QueueChannel;
+import org.springframework.integration.history.MessageHistory;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.support.GenericMessage;
+
+import reactor.core.publisher.Mono;
+
+/**
+ * @author David Turanski
```
artembilan: I prefer `@since` in test classes as well: much easier to understand where a regression happened.

spring-integration-core/src/test/java/org/springframework/integration/handler/ReactiveMessageHandlerTests.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class ReactiveMessageHandlerTests {
+	private AtomicBoolean handled = new AtomicBoolean();
+	private QueueChannel output = new QueueChannel();
```
artembilan: Every member in the class has to surrounded with blank lines.
That's also related to the blank line in the end of class 😄

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/inbound/ReactiveMongoDbMessageSource.java
```diff
@@ -0,0 +1,243 @@
...
+ * <p>
+ * When expectSingleResult is true, the {@link ReactiveMongoOperations#findOne(Query, Class)} is
+ * used instead, and the message payload will be the single object returned from the
+ * query.
+ *
+ * @author David Turanski
+ *
+ * @since 5.2.2
+ */
+public class ReactiveMongoDbMessageSource extends AbstractMessageSource<Object> {
```
artembilan: I think this one should go away for now. 
We need to think more how reactive sources should process a reactive types they get from underlying drivers...

artembilan (changes requested):
Just some rough review, but should be enough for now 😄

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractReactiveMessageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @since 5.3
+ */
+public abstract class AbstractReactiveMessageHandler extends AbstractBaseMessageHandler
+		implements ReactiveMessageHandler {
+
+	@Override
+	public Mono<Void> handleMessage(Message<?> message) {
+		Assert.notNull(message, "Message must not be null");
+		Assert.notNull(message.getPayload(), "Message payload must not be null"); // NOSONAR - false positive
```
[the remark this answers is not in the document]
dturanski: Should be removed in `AbstractMessageHandler` also

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/outbound/MongoDbStoringMessageHandlerTests.java
```diff
@@ -49,29 +50,38 @@
  */
 public class MongoDbStoringMessageHandlerTests extends MongoDbAvailableTests {
```
artembilan: Yeah... I know we need to come up with an extension for JUnit 5, or just try to rely on the embedded MongoDB.
Different story, though...
Meanwhile let's stick with JUnit 4 here for a while!

artembilan (changes requested):

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/ReactiveMongoDbStoringMessageHandler.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	@Override
+	public String getComponentType() {
+		return "mongo:reactive-outbound-channel-adapter";
+	}
+
+	@Override
+	protected void onInit() {
+		this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(this.getBeanFactory());
```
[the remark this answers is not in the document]
dturanski: BTW - All the java doc, and `onInit()` copied from the blocking version

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/ReactiveMongoDbStoringMessageHandler.java
```diff
@@ -0,0 +1,129 @@
...
+		if (this.mongoTemplate == null) {
+			this.mongoTemplate = new ReactiveMongoTemplate(this.mongoDbFactory, this.mongoConverter);
+		}
+		this.initialized = true;
+	}
+
+	@Override
+	protected Mono<Void> handleMessageInternal(Message<?> message) {
+		Assert.isTrue(this.initialized, "This class is not yet initialized. Invoke its afterPropertiesSet() method");
+		String collectionName = this.collectionNameExpression.getValue(this.evaluationContext, message, String.class);
```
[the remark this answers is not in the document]
artembilan: Right, because we don't do "polishing" for the code just because we notices some inconsistency.
Whenever we touch the code in the class, we try to refactor its content according our current code style requirements.
Gary insist to do that in the separate commit (if do that at all), but I'm OK to review the whole change at once.

So, yeah, you can leave code style for me: I'll try to fix it during merge having all your logic on board already.

artembilan (commented):
Let me know if have anything else in mind about for this PR!
Or is it good to go even if we are going to change something else lately.

Thanks

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/outbound/MongoDbStoringMessageHandlerTests.java
dturanski: I'll have a PR in a minute

dturanski:
I think I have addressed all your comments. LMK if there is anything else

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractBaseMessageHandler.java
```diff
@@ -0,0 +1,280 @@
...
+ * @author Oleg Zhurakousky
+ * @author Gary Russell
+ * @author Artem Bilan
+ * @author Amit Sadafule
+ * @author David Turanski
+ *
+ */
+@SuppressWarnings("deprecation")
+@IntegrationManagedResource
+public abstract class AbstractBaseMessageHandler extends IntegrationObjectSupport
```
artembilan: Well, looks like this one is not a `MessageHandler` at all.
Can we call it then something like `MessageHandlerBase` ?
I kind don't like `AbstractBase` combination and it is a bit awkward when the name of the class doesn't reflect it reality. I mean since it doesn't implement `MH`, it cannot be `AbstractMH`

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/ReactiveMongoDbStoringMessageHandler.java
```diff
@@ -0,0 +1,131 @@
...
+
+	private volatile ReactiveMongoDatabaseFactory mongoDbFactory;
+
+	private volatile MongoConverter mongoConverter;
+
+	private volatile StandardEvaluationContext evaluationContext;
+
+	private volatile Expression collectionNameExpression = new LiteralExpression("data");
+
+	private boolean initialized = false;
```
artembilan: No, you didn't get me.
I asked exactly about opposite. All other properties must go without a `volatile` and only this one has to have it, since only this one is definitely accessed from different threads. All others are populated from a single thread and never (roughly) changed.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/outbound/ReactiveMongoDbStoringMessageHandlerTests.java
```diff
@@ -0,0 +1,176 @@
...
+
+import reactor.core.publisher.Mono;
+
+/**
+ * @author Amol Nayak
+ * @author Oleg Zhurakousky
+ * @author Gary Russell
+ * @author David Turanski
+ *
+ * @since 5.2.2
```
artembilan: `5.3` ?

src/reference/asciidoc/whats-new.adoc
```diff
@@ -23,3 +23,8 @@ See its JavaDocs and <<./graph.adoc#integration-graph,Integration Graph>> for mo
 
 [[x5.3-general]]
 === General Changes
+=======
+[[x5.3-general]]
+=== General Changes
```
artembilan: Don't you find this as a duplication of already existing text?

artembilan (changes requested):
Here is some review for the latest changes.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractBaseMessageHandler.java
dturanski: Often we name things like this `MessageHandlerSupport`

src/reference/asciidoc/whats-new.adoc
dturanski: Had to manually merge it. Ooops

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractBaseMessageHandler.java
artembilan: Oh! That's even better, but I hope you agree with me that it's not an `MH` 😄 

Thanks

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/ReactiveMongoDbStoringMessageHandler.java
dturanski: `evaluationContext` and `collectionNameExpression` are also accessed from different threads

artembilan: That's true, but the point is how they are changes. It is definitely a nature of SI components that they are accessed from different threads. Only a difference how they are changed at runtime.
I mean independently of access we set them only once - during application start up.

dturanski: I think SI is the only Spring project that prefers to use volatile. I'm not sure that we ever encountered an actual race condition where it made a difference but I always put it in SI classes for legacy reasons. You're the lead now so you can change it.

artembilan (approved):
OK. Will pull it locally for final review and possible merge after clean up.

Thanks

artembilan:
Merged into `5.3-WIP` as https://github.com/spring-projects/spring-integration/commit/f9fa7c3f9b05144f98f045d130d5ba676247b39f after some code clean up.

Thank you @dturanski , looking forward for more!

Issue #3118, opened 2019-12-06
GatewayProxyFactoryBean should not proxy default methods in the interface


artembilan:
```
@Autowired
	private Function<Object, Message<?>> functionGateay;

	@Test
	void testHeadersFromFunctionGateway() {
		Object payload = this.functionGateay
				.andThen(message -> {
					assertThat(message.getHeaders()).containsKeys("gatewayMethod", "gatewayArgs");
					return message.getPayload();
				})
				.apply("testPayload");

		assertThat(payload).isEqualTo("testPayload");
	}
```
So, we definitely would like to use `default` methods in the interface as well, but it turns out they are proxied as well leading to unexpected behavior when we don't call a real `default` method, but send a message downstream instead.

On the other hand it is not so easy to exclude those methods from proxing and their normal logic:
```
			NameMatchMethodPointcutAdvisor gatewayAdvisor = new NameMatchMethodPointcutAdvisor(this);

			Method[] methods =
					ReflectionUtils.getUniqueDeclaredMethods(this.serviceInterface,
							method -> Modifier.isAbstract(method.getModifiers()));
			for (Method method : methods) {
				MethodInvocationGateway gateway = createGatewayForMethod(method);
				this.gatewayMap.put(method, gateway);
				gatewayAdvisor.addMethodName(method.getName());
			}

			ProxyFactory gatewayProxyFactory =
					new ProxyFactory(this.serviceInterface, AdvisedSupport.EMPTY_TARGET_SOURCE);
			gatewayProxyFactory.addAdvisor(gatewayAdvisor);
			this.serviceProxy = gatewayProxyFactory.getProxy(this.beanClassLoader);
```
When a `default` method in called we have a `target` for reflection as `null` leading to NPE.

Spring Data has something like this `DefaultMethodInvokingMethodInterceptor` - need to investigate what is the beast.

artembilan:
The `DefaultMethodInvokingMethodInterceptor` doesn't work for `Function.class` proxing use-cases:
```
Caused by: java.lang.IllegalArgumentException: illegal lookupClass: interface java.util.function.Function
	at java.lang.invoke.MethodHandleStatics.newIllegalArgumentException(MethodHandleStatics.java:139)
	at java.lang.invoke.MethodHandles$Lookup.checkUnprivilegedlookupClass(MethodHandles.java:686)
	at java.lang.invoke.MethodHandles$Lookup.<init>(MethodHandles.java:604)
```

[...]

PR #3119, opened 2019-12-11
GH-3118: MessagingGW: Don't proxy default methods

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3118

The `GatewayProxyFactoryBean` proxies all the methods in the provided interface,
including `default` and `static`.
This is not what is expected from end-users.

* Proxy only `abstract` methods from the provided interface
* Introduce a `DefaultMethodInvokingMethodInterceptor` to call `default`
method on the interface using a `MethodHandle` approach for those methods
calling

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ if (System.getenv('TRAVIS') || System.getenv('bamboo_buildKey')) {
 
 	nohttp {
 		whitelistFile = file('src/nohttp/whitelist.lines')
-		source.include '**/src/**'
+		source.include '**/src/**'GatewayDslTests
 		source.exclude '**/*.gif', '**/*.jpg', '**/*.png', '**/*.svg', '**/*.ks'
 	}
 }
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ if (System.getenv('TRAVIS') || System.getenv('bamboo_buildKey')) {
 
 	nohttp {
 		whitelistFile = file('src/nohttp/whitelist.lines')
-		source.include '**/src/**'GatewayDslTests
+		source.include '**/src/**'
 		source.exclude '**/*.gif', '**/*.jpg', '**/*.png', '**/*.svg', '**/*.ks'
 	}
 }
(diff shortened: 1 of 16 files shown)


garyrussell:
While this generally makes sense; it's a severely breaking change and, at least, it needs to be an option at the class or method level.

Consider

```java
public interfact Foo {

    String bar(String baz);

    default String qux(String fiz) {
        throw new UnsupportedOperationException();
    }

}
```

A user might reasonably expect to be able to override `qux` with an SI flow.

Furthermore, someone might already be using it that way.

artembilan:
>  it's a severely breaking change

Oh! I didn't think it that way. "How an unusual behavior can be OK?!" 😄 

So, `proxyDeafultMethods` on the `<gateway>`  and `@MessagingGateway` be it!
Or if you would like to proxy only some specific one, you can use a `@Gateway` explicitly on that method.

Is this OK?

garyrussell:
Yes; that would be sufficient (and false by default is ok too).

Do we really need all that `MethodHandle` complexity; we can't use normal reflection?

artembilan:
>  we can't use normal reflection?

No, because there is no `target` for method call in case of proxy against just an interface.
I saw somewhere only a solution with creating a class as an implementation on the fly and then use it as a target for that default method call.

The `MethodHandle` complexity here is with the same reason Spring Data has: proxy an interface without any class implementation.
More over: pay attention to the docs change - we still are not good with JDK interfaces anyway.
Here is some article on the matter: https://rmannibucau.wordpress.com/2014/03/27/java-8-default-interface-methods-and-jdk-dynamic-proxies/
And SO on the matter: https://stackoverflow.com/questions/26206614/java8-dynamic-proxy-and-default-methods

artembilan:
>  it needs to be an option at the class or method level.

After more thinking I treat this issue as a `5.3` theme: no one complained so far about `default` methods, plus it really may feel as an expected behavior to have those methods proxied as well.
So, changing behavior (even if it is kinda bug fix) should go to the next version.

artembilan:
OK. Introduced a `proxyDefaultMethods` options and incorporated the change against `5.3-WIP` branch.

Thanks

Issue #3120, opened 2019-12-12
Identify a pulling Messaging Gateway method as an Outbound Channel Adapter EI pattern.


artembilan:
See https://docs.spring.io/spring-integration/reference/html/messaging-endpoints.html#gateway-calling-no-argument-methods

> When invoking methods on a Gateway interface that do not have any arguments, the default behavior is to receive a `Message` from a `PollableChannel`.

In this case a Messaging Gateway components is going to be present in the end of flow.

artembilan:
Fixed via: https://github.com/spring-projects/spring-integration/commit/bc8e77527c9cad84fb95a9e66faa30fe383c7048

PR #3122, opened 2019-12-12
GH-3120: Identify polling MG as outbound-CA

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3120

The messaging gateway method without arguments works as a
polling channel adapter.
In this case a Messaging Gateway component is present in the end of flow.
Therefore we treat it as an `outbound_channel_adapter`


garyrussell:
LGTM; awaiting travis' approval 😄

artembilan:
It is green already. 😄

PR #3124, opened 2019-12-16
Handle possible null management overrides

jdpgrailsdev:
Adds handling for cases where the `ManagementOverrides` retrieved from a bean during auto configuration may be `null`.  I ran into this an a Spring Boot integration test that was loading configuration from my application with the `EnableAutoConfiguration` annotation.  Removing the auto configuration annotation resolved the NPE's in the configurer, but that may not always be possible for tests, especially those that rely on Spring configuration classes with auto configuration enabled that are loaded from a provided library JAR.  This PR attempts to handle that case by returning an empty/un-configured `ManagementOverrides` in the case where the call to the `getOverrides()` method on the bean returns null.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
@@ -278,7 +278,7 @@ public class IntegrationManagementConfigurer
 				.getBeansOfType(IntegrationManagement.class);
 		for (Entry<String, IntegrationManagement> entry : managed.entrySet()) {
 			IntegrationManagement bean = entry.getValue();
-			if (!bean.getOverrides().loggingConfigured) {
+			if (!getOverrides(bean).loggingConfigured) {
 				bean.setLoggingEnabled(this.defaultLoggingEnabled);
 			}
 			String name = entry.getKey();
@@ -292,7 +292,7 @@ public class IntegrationManagementConfigurer
 				.getBeansOfType(IntegrationManagement.class);
 		for (Entry<String, IntegrationManagement> entry : managed.entrySet()) {
 			IntegrationManagement bean = entry.getValue();
-			if (!bean.getOverrides().loggingConfigured) {
+			if (!getOverrides(bean).loggingConfigured) {
 				bean.setLoggingEnabled(this.defaultLoggingEnabled);
 			}
 			bean.registerMetricsCaptor(this.metricsCaptor);
@@ -364,7 +364,7 @@ public class IntegrationManagementConfigurer
 			metrics = this.metricsFactory.createChannelMetrics(name);
 		}
 		Assert.state(metrics != null, "'metrics' must not be null");
-		ManagementOverrides overrides = bean.getOverrides();
+		ManagementOverrides overrides = getOverrides(bean);
 		Boolean enabled = PatternMatchUtils.smartMatch(name, this.enabledCountsPatterns);
 		if (enabled != null) {
 			bean.setCountsEnabled(enabled);
@@ -396,7 +396,7 @@ public class IntegrationManagementConfigurer
 			org.springframework.integration.support.management.MessageHandlerMetrics bean) {
 		AbstractMessageHandlerMetrics metrics = this.metricsFactory.createHandlerMetrics(name);
 		Assert.state(metrics != null, "'metrics' must not be null");
-		ManagementOverrides overrides = bean.getOverrides();
+		ManagementOverrides overrides = getOverrides(bean);
 		Boolean enabled = PatternMatchUtils.smartMatch(name, this.enabledCountsPatterns);
 		if (enabled != null) {
 			bean.setCountsEnabled(enabled);
@@ -432,7 +432,7 @@ public class IntegrationManagementConfigurer
 			bean.setCountsEnabled(enabled);
 		}
 		else {
-			if (!bean.getOverrides().countsConfigured) {
+			if (!getOverrides(bean).countsConfigured) {
 				bean.setCountsEnabled(this.defaultCountsEnabled);
 			}
 		}
@@ -504,6 +504,10 @@ public class IntegrationManagementConfigurer
 			LOGGER.debug("No source found for (" + name + ")");
 		}
 		return null;
+	}
+
+	private ManagementOverrides getOverrides(final IntegrationManagement bean) {
+		return bean.getOverrides() != null ? bean.getOverrides() : new ManagementOverrides();
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
@@ -54,6 +54,7 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Meherzad Lahewala
+ * @author Jonathan Pearlin
  *
  * @since 4.2
  *
@@ -506,7 +507,7 @@ public class IntegrationManagementConfigurer
 		return null;
 	}
 
-	private ManagementOverrides getOverrides(final IntegrationManagement bean) {
+	private static ManagementOverrides getOverrides(IntegrationManagement bean) {
 		return bean.getOverrides() != null ? bean.getOverrides() : new ManagementOverrides();
 	}
 


artembilan:
@jdpgrailsdev , thank you for bringing this up!

May we see an exception on the matter?
I wonder how that is possible that `managementOverrides` is `null` from the bean we try to call.

What I see in our code it is always like this `private final ManagementOverrides managementOverrides = new ManagementOverrides();`. 
So, cannot be `null` in Spring Integration components...

Thanks

garyrussell:
Probably a CGLib proxy. It is a final method.

artembilan:
Ah! Great! SO, maybe let's remove that `final` instead?

garyrussell:
No; it's not the final property, but I don't see a `final getOverrides()` anywhere, so I am not sure how it can be `null`.

jdpgrailsdev:
@artembilan I should be able to re-create the stack trace by adding back the `@EnableAutoConfiguration` annotation in my tests.  Let me know if seeing that is still helpful.

jdpgrailsdev:
@artembilan 
Here is the stack trace when `@EnableAutoConfiguration` to one of the configurations imported by the integration test configuration:

```
java.lang.IllegalStateException: Failed to load ApplicationContext
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:132)
	at org.springframework.test.context.support.DefaultTestContext.getApplicationContext(DefaultTestContext.java:123)
	at org.springframework.test.context.web.ServletTestExecutionListener.setUpRequestContextIfNecessary(ServletTestExecutionListener.java:190)
	at org.springframework.test.context.web.ServletTestExecutionListener.prepareTestInstance(ServletTestExecutionListener.java:132)
	at org.springframework.test.context.TestContextManager.prepareTestInstance(TestContextManager.java:244)
	at org.spockframework.spring.SpringTestContextManager.prepareTestInstance(SpringTestContextManager.java:56)
	at org.spockframework.spring.SpringInterceptor.interceptInitializerMethod(SpringInterceptor.java:43)
	at org.spockframework.runtime.extension.AbstractMethodInterceptor.intercept(AbstractMethodInterceptor.java:24)
	at org.spockframework.runtime.extension.MethodInvocation.proceed(MethodInvocation.java:97)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:110)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:38)
	at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:62)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.processTestClass(TestWorker.java:118)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:182)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:164)
	at org.gradle.internal.remote.internal.hub.MessageHub$Handler.run(MessageHub.java:412)
	at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
	at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:48)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:56)
	at java.lang.Thread.run(Thread.java:748)
	at org.springframework.integration.config.IntegrationManagementConfigurer.afterSingletonsInstantiated(IntegrationManagementConfigurer.java:281)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:896)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:878)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:550)
	at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:747)
	at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:397)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:315)
	at org.springframework.boot.test.context.SpringBootContextLoader.loadContext(SpringBootContextLoader.java:125)
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContextInternal(DefaultCacheAwareContextLoaderDelegate.java:99)
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:124)
	... 29 more
```

I get a slightly different NPE at another location in the configurer if I included `micrometer-core` on the classpath, but it is the same issue (trying to dereference the `getOverrides` method).  Here is the basic setup for the test:

```
@Configuration
@EnableAutoConfiguration
public class IntegrationConfiguration {
```

Imported by the test configuration:

```
@Configuration
@Import([IntegrationConfiguration])
class IntegrationTestConfiguration { 
```

And then used by the integration test:

```
@ActiveProfiles('test')
@SpringBootTest(classes=[IntegrationTestConfiguration])
class IntegrationSpec extends Specification {
```

This test uses version 5.2.2.RELEASE of spring/spring-integration and 2.2.2.RELEASE of Spring Boot.

artembilan:
OK! Since a stack trace doesn't say us what bean is guilty, would it be possible for your to share with us a simple Spring Boot app to reproduce an issue and play with it to dig out the cause?
Thanks

jdpgrailsdev:
@artembilan Sure thing.  Might take me a few days to get that spun up.  Will update this PR once I have it pushed up to GitHub.

garyrussell:
Thanks; we'd like to understand how it's happening rather than just putting a band-aid over it.

jdpgrailsdev:
@artembilan @garyrussell 
I am still trying to reproduce this in a separate project.  I have narrowed it down to a bean that is component scanned, extends `AbstractMessageHandler` and implements `DisposableBean`.  The bean also has a few methods annotated with `@Scheduled` in addition to the required methods from the message handler parent.  I confirmed that the `@EnableScheduling` annotation is on the application's configuration, in addition to the `@EnableAutoConfiguration`.  I have recreated the class in another project, but it does not seem to run into the same issue.  I suspect that it may also potentially have to do with classpath differences (the test project has very few dependencies relative to the project where I can reproduce this issue consistently).  I was able to remote debug the integration tests in the original project where I discovered this issue and I can confirm that the `managementOverrides` are in fact `null` on the bean in question.  Any thoughts/hints on how to narrow this down now that I know what the bean is (e.g. where to put a breakpoint in some of the Spring code) would be greatly appreciated.

jdpgrailsdev:
> @artembilan @garyrussell
> I am still trying to reproduce this in a separate project. I have narrowed it down to a bean that is component scanned, extends `AbstractMessageHandler` and implements `DisposableBean`. The bean also has a few methods annotated with `@Scheduled` in addition to the required methods from the message handler parent. I confirmed that the `@EnableScheduling` annotation is on the application's configuration, in addition to the `@EnableAutoConfiguration`. I have recreated the class in another project, but it does not seem to run into the same issue. I suspect that it may also potentially have to do with classpath differences (the test project has very few dependencies relative to the project where I can reproduce this issue consistently). I was able to remote debug the integration tests in the original project where I discovered this issue and I can confirm that the `managementOverrides` are in fact `null` on the bean in question. Any thoughts/hints on how to narrow this down now that I know what the bean is (e.g. where to put a breakpoint in some of the Spring code) would be greatly appreciated.

I posted this too soon.  I figured out what is causing it.  The bean in question is mocked via Mockito for the integration test.  I have re-created it using the sample Spring Boot project:  https://github.com/jdpgrailsdev/gs-spring-boot/tree/npe-management-overrides.  Simply run `./gradlew intTest` from the `initial` directory to see it fail.  I believe that the mock was set it to avoid component scanning in the test.  Not sure why we decided to mock it instead of just declaring the bean in our test configuration.  However, it does not seem unreasonable to me to be able to mock these beans for integration testing, especially if you want to avoid using a bean that talks to an external service.

garyrussell:
@jdpgrailsdev Thanks for persevering.

With this code...

```java
@SpringBootApplication
@EnableScheduling
public class Sigh3124Application {

		SpringApplication.run(Sigh3124Application.class, args);
	}

}

@Component
class Foo extends AbstractMessageHandler {

	Foo() {
		super();
		System.out.println("ctor");
	}

	@Override
	protected void handleMessageInternal(Message<?> message) {

	}

	@Scheduled(fixedDelay = 5000)
	public void sched() {
		System.out.println("in sched()");
	}

}
```

I was able to set a breakpoint on the `super()` and "step into" the instantiation of the super class and the first step initialized that variable, Without the `super()` I was able to set a watch point on the field and, again, see it stop there. Since it is null, that probably won't trigger for you but if you can add a `super()` to the subclass CTOR it might be worth examining what happens there.

...time passes...

Is it possible that you are mocking the message handler class? I can reproduce the NPE with this test...

```java
@SpringJUnitConfig
class Sigh3124ApplicationTests {

	@Test
	void contextLoads() {
	}

	@Configuration
	@EnableIntegrationManagement
	public static class Config {

		@Bean
		public Foo foo() {
			return mock(Foo.class);
		}

	}

}
```

It's generally dangerous to mock classes (rather than interfaces) for reasons like this because Mockito doesn't perform normal instantiation.

While your fix will work around this, I am not sure you won't see other problems too.

If you are not mocking the class, I am out of ideas.

garyrussell:
Ha - our posts crossed.

garyrussell:
I concur that your fix will avoid this, but I can't guarantee there won't be other issues; it's generally not good practice to mock classes - can you not simply mock `MessageHandler`? 

If you are injecting your custom handler by type somewhere, it would be better to have it implement an interface and inject it that way.

jdpgrailsdev:
> I concur that your fix will avoid this, but I can't guarantee there won't be other issues; it's generally not good practice to mock classes - can you not simply mock `MessageHandler`?
> 
> If you are injecting your custom handler by type somewhere, it would be better to have it implement an interface and inject it that way.

Let me change the integration tests in the real application to mock the interface instead of the class.  However, I suspect it may break other code that is expecting the actual type to be wired in (autowire by type after all).

garyrussell:
But I am ok with applying this fix; let's see if @artembilan agrees when he's back.

artembilan:
Right. My experience said that it is really bad practice to mock classes not interfaces. 
It is really unpredictable what to expect from real method calls. 
When we mock only interfaces you definitely have a full control over method calls with stubbing.
Yes, you can stub for classes mocks, but as you see you not always do that for all the methods in the target class.

Well, even if we pursue here an attempt to satisfy some test with that scary class mocking, we definitely may have some `IntegrationManagement` implementation with `getOverrides()` returning `null`.
So, let's accept this fix and move Mockito issue to some other place!

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
```diff
@@ -506,4 +506,8 @@ private void registerComponentGauges() {
 		return null;
 	}
 
+	private ManagementOverrides getOverrides(final IntegrationManagement bean) {
```
artembilan: Should be `static`. No `final` for arg, please.
Your name to the `@author` list.

Thank you!

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
jdpgrailsdev: @artembilan Addressed your comments.

artembilan:
@jdpgrailsdev ,

PR #3128, opened 2019-12-18
Shorten internal component names for graph

artembilan:
The long `IntegrationContextUtils.ERROR_LOGGER_BEAN_NAME` name
is used for `errorChannel` `LoggingHandler`'s endpoint.
It is too long for endpoint node in the graph meanwhile the
`errorChannel` is good everywhere

* Change an `IntegrationGraphServer` logic to shorten internal
component names before drawing a graph
* Introduce `IntegrationUtils.obtainComponentName()` for
shortening an internal name if any
* Use `IntegrationConfigUtils.BASE_PACKAGE` in the `IntegrationMBeanExporter`
instead of its own static property
* Remove `IntegrationConfigUtils.FACTORY_BEAN_OBJECT_TYPE` in favor
of introduced in SF `FactoryBean.OBJECT_TYPE_ATTRIBUTE`
* Rework affected test classes to JUnit 5
* Remove usage of deprecated options in the `@EnableIntegrationManagement`
* Document names shortening feature for graph

The current picture has a not good label:

image

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
@@ -24,8 +24,10 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.core.convert.ConversionService;
+import org.springframework.integration.config.IntegrationConfigUtils;
 import org.springframework.integration.support.DefaultMessageBuilderFactory;
 import org.springframework.integration.support.MessageBuilderFactory;
+import org.springframework.integration.support.context.NamedComponent;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHandlingException;
@@ -191,6 +193,20 @@ public final class IntegrationUtils {
 					(ex instanceof IllegalStateException && ex.getCause() != null) ? ex.getCause() : ex);
 		}
 		return runtimeException;
+	}
+
+	/**
+	 * Obtain a component name from the provided {@link NamedComponent}.
+	 * @param component the {@link NamedComponent} source for component name.
+	 * @return the component name
+	 * @since 5.3
+	 */
+	public static String obtainComponentName(NamedComponent component) {
+		String name = component.getComponentName();
+		if (name.startsWith('_' + IntegrationConfigUtils.BASE_PACKAGE)) {
+			name = name.substring(('_' + IntegrationConfigUtils.BASE_PACKAGE).length() + 1);
+		}
+		return name;
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
@@ -28,6 +28,7 @@ import org.springframework.integration.config.IntegrationConfigUtils;
 import org.springframework.integration.support.DefaultMessageBuilderFactory;
 import org.springframework.integration.support.MessageBuilderFactory;
 import org.springframework.integration.support.context.NamedComponent;
+import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHandlingException;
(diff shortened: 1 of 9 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
```diff
@@ -233,7 +233,7 @@ private void registerErrorChannel() {
 
 			BeanComponentDefinition componentDefinition =
 					new BeanComponentDefinition(loggingEndpointBuilder.getBeanDefinition(),
-							IntegrationContextUtils.ERROR_LOGGER_BEAN_NAME);
+							"errorLogger", new String[] { IntegrationContextUtils.ERROR_LOGGER_BEAN_NAME });
```
garyrussell: Hmm... maybe too generic a name - likely to collide with a user bean.

`default_errorLogger`
`integration_errorLogger`
`si_errorLogger`
`_si_errorLogger`
...

might be better choices.

artembilan: Well, my point is to align with the `errorChannel` indicating that this default logger is definitely tied with that global default channel.
Yes, it could collide with something in the target application, that why I'm targeting this for `5.3`, so, we have a room for some breaking changes with an appropriate Migration Guide.

garyrussell:
I understand your goal; but my point is that people will definitely use `errorChannel` in their configuration - particularly on gateways where they just want the error logged.

However, I doubt that anyone will explicitly reference the default error logger - hence its obscure name.

I just think it might be an unnecessary burden to break an app that already has a bean `errorLogger` that might not even be related to the Spring Integratin part of the application, especially when this change is really just to have a shorter name in the graph.

spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
garyrussell: I understand your goal; but my point is that people will definitely use `errorChannel` in their configuration - particularly on gateways where they just want the error logged.

However, I doubt that anyone will explicitly reference the default error logger - hence its obscure name.

I just think it might be an unnecessary burden to break an app that already has a bean `errorLogger` that might not even be related to the Spring Integratin part of the application, especially when this change is really just to have a shorter name in the graph.

artembilan:
>  especially when this change is really just to have a shorter name in the graph.

Yes, you asked me to provide short names for those endpoints and that is where I was stuck 😄 

And yeah, I did look at this like something what is not really a bean for end-user.
So, yeah. the name like `default_integration_errorLogger` should be good.

@kdvolder , tell us, please, how long the node name can be?

Thanks

artembilan:
@garyrussell ,

What if we do something similar for graph like we have in JMX:
```
private String getInternalComponentName(String name) {
	return name.substring(('_' + SI_PACKAGE).length() + 1);
}
```

This way we don't need to change a bean name for logger, but we will shorten it for UI like we do for JMX.

garyrussell:
Yes, that would work for me, but that code should be improved to check `name.charAt(0) == '_'` before substringing.

artembilan:
Yes, of course, this one is covered in JMX like this:
```
if (managedName != null && managedName.startsWith('_' + SI_PACKAGE)) {
	managedName = getInternalComponentName(managedName);
	managedType = "internal";
}
```

spring-integration-core/src/main/java/org/springframework/integration/support/utils/IntegrationUtils.java
```diff
@@ -193,4 +195,18 @@ public static RuntimeException wrapInHandlingExceptionIfNecessary(Message<?> mes
...
 
+	/**
+	 * Obtain a component name from the provided {@link NamedComponent}.
+	 * @param component the {@link NamedComponent} source for component name.
+	 * @return the component name
+	 * @since 5.3
+	 */
+	public static String obtainComponentName(NamedComponent component) {
+		String name = component.getComponentName();
+		if (name.startsWith('_' + IntegrationConfigUtils.BASE_PACKAGE)) {
```
garyrussell: This would be more efficient as `if (name.charAt(0) == '_' && name.startsWith(...`.

Not a big deal with JMX but we're building the graph more often than during initialization.

garyrussell (changes requested):
Just one

PR #3137, opened 2020-01-08
Infrastructure for ReactiveMessageHandler

artembilan:
We have now a `ReactiveMongoDbStoringMessageHandler` which implements
a `ReactiveMessageHandler`, but not a `MessageHandler` for possible
deferred subscriptions to the returned Reactor type

We don't have a proper application context processing for this
new type of message handlers

* Change a  `ConsumerEndpointFactoryBean` to apply an `MH` and `RMH`
as possible types for handler
* Introduce a `ReactiveMessageHandlerAdapter` to wrap an `RMH`
into a `MH` for synchronous calls in the regular consumer endpoints
* Wrap an `RMH` into a `ReactiveMessageHandlerAdapter` for regular
endpoints and unwrap for `ReactiveStreamsConsumer`
* Add `RMH`-based ctor into `ReactiveStreamsConsumer` for target
reactive streams composition (`flatMap()` on the `RMH`)
* Remove a `DelegatingSubscriber` from the `ReactiveStreamsConsumer`
in favor of direct calls from the `doOnSubscribe()`, `doOnComplete()`
& `doOnNext()`
* Add an `onErrorContinue()` to handle per-message errors, but don't
cancel the whole source `Publisher`
* Use `Disposable` from the `subscribe()` to cancel in the `stop()`
- recommended way in Reactor
* Use `onErrorContinue()` in the `FluxMessageChannel` instead of
`try..catch` in the `doOnNext()` - for possible `onErrorStop()`
in the provided upstream `Publisher`
* Handle `RMH` in the `ServiceActivatorFactoryBean` as a direct handler
as well with wrapping into `ReactiveMessageHandlerAdapter` for return.
The `ConsumerEndpointFactoryBean` extracts an `RMH` from the adapter
for the `ReactiveStreamsConsumer` anyway
* Add XML parsing test for `ReactiveMongoDbStoringMessageHandler`
* Add `log4j-slf4j-impl` for all the test runtime since `slf4j-api`
comes as a transitive dependency from many places

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
@@ -42,12 +42,14 @@ import org.springframework.integration.endpoint.EventDrivenConsumer;
 import org.springframework.integration.endpoint.PollingConsumer;
 import org.springframework.integration.endpoint.ReactiveStreamsConsumer;
 import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.integration.handler.ReactiveMessageHandlerAdapter;
 import org.springframework.integration.handler.advice.HandleMessageAdvice;
 import org.springframework.integration.scheduling.PollerMetadata;
 import org.springframework.integration.support.channel.ChannelResolverUtils;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.messaging.PollableChannel;
+import org.springframework.messaging.ReactiveMessageHandler;
 import org.springframework.messaging.SubscribableChannel;
 import org.springframework.messaging.core.DestinationResolver;
 import org.springframework.scheduling.TaskScheduler;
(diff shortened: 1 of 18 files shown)


artembilan:
TODO:

* Java DSL
* Annotations
* Docs

Although I can do them all in the separate PR(s).

@dturanski , this is what I meant about processing `ReactiveMessageHandler` in the flow.

Thanks

artembilan:
This one depends on #3139 . So, after merging that one I'll update here.

But I definitely would like to have this merged for upcoming release.

Thanks

garyrussell:
I will look at it today, but you might want to request a review from the Reactor team since my Reactor skills are lacking.

Minor checkstyle issue reported by Travis.

artembilan:
> Minor checkstyle issue reported by Travis.

Will fix after merging MongoDB revert.

artembilan:
@bsideup ,

any chances that you can sanctify this PR?

Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
```diff
@@ -112,11 +114,17 @@
 
 	private volatile boolean initialized;
 
-	public void setHandler(MessageHandler handler) {
-		Assert.notNull(handler, "handler must not be null");
+	public void setHandler(Object handler) {
+		Assert.state(handler instanceof MessageHandler || handler instanceof ReactiveMessageHandler,
```
garyrussell: Why ISE instead of IAE? This is a setter. `Assert.isTrue()` ?

spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
```diff
@@ -210,7 +218,12 @@ public void afterPropertiesSet() {
 			}
 		}
 
-		adviceChain();
+		if (!(this.handler instanceof ReactiveMessageHandlerAdapter)) {
+			adviceChain();
+		}
+		else {
+			LOGGER.warn("the advice chain cannot be applied for 'ReactiveMessageHandler'");
```
garyrussell: `s/for/to a/`

garyrussell (changes requested):
Minor issues; otherwise LGTM

artembilan:
Please, find the changes for Java DSL in the latest commit.
Since no feedback from Reactor team, I guess we are good to go and can revise in the future if that.

Thanks

Issue #2804, opened 2019-03-12
Eliminate MicrometerMetricsCaptor


garyrussell:

SyedMuhammadSufyian:
@garyrussell , I would like to work on this issue

garyrussell:
Thanks for the offer but this is slated for a future release - 5.3 at the earliest. That's why it's on the backlog milestone, not 5.2.x.

artembilan:
Some relevant work: https://github.com/spring-projects/spring-integration/pull/3357

PR #3140, opened 2020-01-09
Remove deprecations from previous versions

artembilan:
* Remove Boon dependency and its usage
* Remove overloaded methods from the `IntegrationFlowDefinition`
- we can simply rely now on the super class
* Remove (or rework) deprecated entities in the docs
* Fix tests for removed deprecated APIs
* Rework affected tests to JUnit 5


artembilan:
If that is really necessary I'll add `Migration Guide` page for this version.

I haven't done anything for deprecated metrics, because have no idea how to proceed.
I think those removals are related somehow to this: https://github.com/spring-projects/spring-integration/issues/2804

Thanks

garyrussell:
Yes; I will work on that starting tomorrow - maybe won't make M1, though.

garyrussell:
Removing legacy metrics is independent of #2804.

garyrussell:
@artembilan 

>I haven't done anything for deprecated metrics, because have no idea how to proceed.

I looked at this some time ago; but it turns out there are some classes that will have to be removed that are NOT deprecated.

I will create a PR to deprecate all of those.

PR #3144, opened 2020-01-15
Fix for SimplePool.setPoolSize() #3143

### LINKED ISSUE #3143: SimplePool setPoolSize incorrectly updates internal fields
##  Bug report

When new pool size is less then existing, setPoolSize() method breaks consistency.
This happens when CachingSessionFactory is constructed using the most "convenient" constructor `CachingSessionFactory(SessionFactory<F> sessionFactory)`. In this case initial poolSize=Integer.MAX_VALUE. And so when later we are trying to setup some other value, the pool object becomes broken.

sergebg:
Original implementation exits from the loop if "available" collection is empty. That is not correct, because the number of allocated items can be less then pool size. And so it's OK if "item==null". It's just required to perform cleanup if "item!=null". Otherwise it's required to complete the loop until "delta!=0". 

Resolves https://github.com/spring-projects/spring-integration/issues/3143

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
@@ -36,6 +36,7 @@ import org.springframework.util.Assert;
  * demand up to the limit.
  *
  * @author Gary Russell
+ * @author Sergey Bogatyrev
  * @since 2.2
  *
  */
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
@@ -102,11 +102,9 @@ public class SimplePool<T> implements Pool<T> {
 					break;
 				}
 				T item = this.available.poll();
-				if (item == null) {
-					this.permits.release();
-					break;
+				if (item != null) {
+					doRemoveItem(item);
 				}
-				doRemoveItem(item);
 				this.poolSize.decrementAndGet();
 				delta++;
 			}
--- a/spring-integration-core/src/test/java/org/springframework/integration/util/SimplePoolTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/util/SimplePoolTests.java
@@ -16,20 +16,23 @@
 
 package org.springframework.integration.util;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
-import org.junit.Test;
-import org.springframework.integration.test.util.TestUtils;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.fail;
+import org.junit.Test;
+
+import org.springframework.integration.test.util.TestUtils;
 
 /**
  * @author Gary Russell
+ * @author Sergey Bogatyrev
  * @since 2.2
  *
  */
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@sergebg Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3144)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3144) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sergebg Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3144)!

spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
```diff
@@ -102,11 +102,9 @@ public synchronized void setPoolSize(int poolSize) {
 					break;
 				}
 				T item = this.available.poll();
-				if (item == null) {
-					this.permits.release();
```
artembilan: You have stopped to release permits.
Why is that, please?

spring-integration-core/src/test/java/org/springframework/integration/util/SimplePoolTests.java
```diff
@@ -16,18 +16,18 @@
...
 import java.util.Set;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.junit.Test;
-
 import org.springframework.integration.test.util.TestUtils;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
```
artembilan: Please, do not rearrange imports order.
Consider to run `gradlew check --parallel` task to be sure that your changes follows our Checkstyle rules.
Also that will say you if your fix is valid against all the existing tests in the project.

artembilan (changes requested):
Thank you for contribution!

I'm not fully understand your fix.
Would you mind to explain it a bit different way?

Also consider to add your name to the `@author` list.

spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
sergebg: In this branch we need  to call only `acquire()`. Original intent was to revert acquire if item==null. But this is absolutely incorrect. For example, if we have a new instance of pool, then "available" collection is empty. But setPoolSize() method should correctly update all structures. After calling `setPoolSize(5)` everything should contain `5`. (permits==5, poolSize==5, and targetPoolSize==5). Of course, I am describing the situation when no objects were allocated. Please, check added test cases.

spring-integration-core/src/test/java/org/springframework/integration/util/SimplePoolTests.java
sergebg: sorry, I tried to keep imports untouched. I will revert this

spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
sergebg: I have added additional checks to test that "permits" field is correctly updated

artembilan:
@garyrussell , would you mind to take a look here?
Looks like you are more familiar with the code involved.

Thank you!

garyrussell:
@sergebg Good catch, thanks.

Some suggested test polishing here: https://github.com/garyrussell/spring-integration/commit/3521f2a028266da6e043450be9bf9fe86396b206

@artembilan I can squash/merge after review of that commit.

artembilan (approved):
LGTM after @garyrussell 's polishing.

Thank you both!

garyrussell:
Doing a local full build now; will merge after success.

garyrussell:
Merged as dba307adf62e703b8f5df9bc23330eb99ecf1a01

@sergebg Thanks for the contribution!

garyrussell:
Cherry-picked to 5.2.x as 3850c7ee0ccec27fc23c4af64d05a4882a032274

Issue #3152, opened 2020-01-23
Scatter Gather is getting hung when Nested Scatter Gathers are used


sirimamilla:
[5.2.3.log](https://github.com/spring-projects/spring-integration/files/4102774/5.2.3.log)
[5.1.3.log](https://github.com/spring-projects/spring-integration/files/4102780/5.1.3.log)

**Affects Version(s):** \<Spring Integration version>
5.1.7 and above
5.2.3
---
Hi,

Scatter Gather is failing when nested scatter Gathers are used with latest versions of Spring Integration. we tried with 5.1.7, 5.1.9 and 5.2.3.  Same is working fine in 5.1.6 and below.
Here is the sample project
https://github.com/sirimamilla/ScatterGatherTest

Test case ScatterGatherTestSuccess runs indefinitely in this project.

when analyzed, scatter gather is using origialReplyChannel and originalErrorChannel Headers to preserve the previous original reply and error channels.

Inner scatterGather is responding correctly to the outer scatter gather. But outer scatter gather is failing to respond as it is finding its own gatherchannel in original replychannel header and sending response to gather channel and waiting indefinitely to continue. 

This issue is blocking our version upgrade from spring boot 2.1.3 to 2.2.4. Please assist.

Regards,
Jayadev

sirimamilla:
Hi,

On Further analysis of issue, in version 5.1.3, reply channels are handled in handleRequestMessage. Reply channels are directly taken from the request message headers. Also no additional headers are added to the request sent to the subflows.

@Override
	protected Object handleRequestMessage(Message<?> requestMessage) {
		PollableChannel gatherResultChannel = new QueueChannel();

		Object gatherResultChannelName = this.replyChannelRegistry.channelToChannelName(gatherResultChannel);

		Message<?> scatterMessage = getMessageBuilderFactory()
				.fromMessage(requestMessage)
				.setHeader(GATHER_RESULT_CHANNEL, gatherResultChannelName)
				.setReplyChannel(this.gatherChannel)
				.build();

		this.messagingTemplate.send(this.scatterChannel, scatterMessage);

		Message<?> gatherResult = gatherResultChannel.receive(this.gatherTimeout);
		if (gatherResult != null) {
			return getMessageBuilderFactory()
					.fromMessage(gatherResult)
					.removeHeader(GATHER_RESULT_CHANNEL)
[...]

PR #3153, opened 2020-01-24
Fix for Scatter Gather getting hung when Nested Scatter Gathers are used

sirimamilla:
Github Issue: https://github.com/spring-projects/spring-integration/issues/3152

Issue with flow going in hung state when nested scatter gather is used in a flow.

Cause:
In 5.2.3 version, ReplyChannel and ErrorChannel are copied to headers ORIGINAL_REPLY_CHANNEL, ORIGINAL_ERROR_CHANNEL. This process is not taking care of any ORIGINAL_REPLY_CHANNEL, ORIGINAL_ERROR_CHANNEL already available in request headers. if these headers are already present in the headers, they will be overridden without being preserved for the reply path.

This is causing issue in the nested ScatterGather situations.

@Override
		MessageHeaders requestMessageHeaders = requestMessage.getHeaders();
		PollableChannel gatherResultChannel = new QueueChannel();

		Message<?> scatterMessage =
				getMessageBuilderFactory()
						.fromMessage(requestMessage)
						.setHeader(GATHER_RESULT_CHANNEL, gatherResultChannel)
						.setHeader(ORIGINAL_REPLY_CHANNEL, requestMessageHeaders.getReplyChannel())
						.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.getErrorChannel())
						.setReplyChannel(this.gatherChannel)
						.setErrorChannelName(this.errorChannelName)
						.build();

		this.messagingTemplate.send(this.scatterChannel, scatterMessage);

		return gatherResultChannel.receive(this.gatherTimeout);
	}

	private Message<?> enhanceScatterReplyMessage(Message<?> message) {
		MessageHeaders headers = message.getHeaders();
		return getMessageBuilderFactory()
				.fromMessage(message)
				.setHeader(MessageHeaders.REPLY_CHANNEL, headers.get(ORIGINAL_REPLY_CHANNEL))
				.setHeader(MessageHeaders.ERROR_CHANNEL, headers.get(ORIGINAL_ERROR_CHANNEL))
				.removeHeaders(ORIGINAL_REPLY_CHANNEL, ORIGINAL_ERROR_CHANNEL)
				.build();
	}

Possible Solution:
Enhance the reply before responding in handleRequestMessage method. this will let the flow continue in success scenarios

@Override
		MessageHeaders requestMessageHeaders = requestMessage.getHeaders();
		PollableChannel gatherResultChannel = new QueueChannel();

		Message<?> scatterMessage =
				getMessageBuilderFactory()
						.fromMessage(requestMessage)
						.setHeader(GATHER_RESULT_CHANNEL, gatherResultChannel)
						.setHeader(ORIGINAL_REPLY_CHANNEL, requestMessageHeaders.getReplyChannel())
						.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.getErrorChannel())
						.setReplyChannel(this.gatherChannel)
						.setErrorChannelName(this.errorChannelName)
						.build();

		this.messagingTemplate.send(this.scatterChannel, scatterMessage);

		Message<?> gatherResult = gatherResultChannel.receive(this.gatherTimeout);
		if (gatherResult != null && (requestMessageHeaders.containsKey(ORIGINAL_REPLY_CHANNEL)
				|| requestMessageHeaders.containsKey(ORIGINAL_ERROR_CHANNEL)
				|| requestMessageHeaders.containsKey(GATHER_RESULT_CHANNEL))) {
			return getMessageBuilderFactory()
					.fromMessage(gatherResult)
					.removeHeader(GATHER_RESULT_CHANNEL)
					.setHeader(GATHER_RESULT_CHANNEL, requestMessageHeaders.get(GATHER_RESULT_CHANNEL))
					.setHeader(ORIGINAL_REPLY_CHANNEL, requestMessageHeaders.get(ORIGINAL_REPLY_CHANNEL))
					.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.get(ORIGINAL_ERROR_CHANNEL))
					.setHeader(MessageHeaders.REPLY_CHANNEL, requestMessage.getHeaders().getReplyChannel())
					.build();
		}

		return gatherResult;
	}

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 the original author or authors.
+ * Copyright 2014-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -190,7 +190,21 @@ public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler i
 
 		this.messagingTemplate.send(this.scatterChannel, scatterMessage);
 
-		return gatherResultChannel.receive(this.gatherTimeout);
+		Message<?> gatherResult = gatherResultChannel.receive(this.gatherTimeout);
+		if (gatherResult != null && (requestMessageHeaders.containsKey(ORIGINAL_REPLY_CHANNEL)
+				|| requestMessageHeaders.containsKey(ORIGINAL_ERROR_CHANNEL)
+				|| requestMessageHeaders.containsKey(GATHER_RESULT_CHANNEL))) {
+			return getMessageBuilderFactory()
+					.fromMessage(gatherResult)
+					.removeHeader(GATHER_RESULT_CHANNEL)
+					.setHeader(GATHER_RESULT_CHANNEL, requestMessageHeaders.get(GATHER_RESULT_CHANNEL))
+					.setHeader(ORIGINAL_REPLY_CHANNEL, requestMessageHeaders.get(ORIGINAL_REPLY_CHANNEL))
+					.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.get(ORIGINAL_ERROR_CHANNEL))
+					.setHeader(MessageHeaders.REPLY_CHANNEL, requestMessage.getHeaders().getReplyChannel())
+					.build();
+		}
+
+		return gatherResult;
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
@@ -38,7 +38,6 @@ import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.PollableChannel;
 import org.springframework.messaging.SubscribableChannel;
 import org.springframework.messaging.support.ChannelInterceptor;
-import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.messaging.support.InterceptableChannel;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
@@ -56,6 +55,8 @@ import org.springframework.util.ClassUtils;
 public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler implements Lifecycle {
 
 	private static final String GATHER_RESULT_CHANNEL = "gatherResultChannel";
+
+	private static final String ORIGINAL_ERROR_CHANNEL = "originalErrorChannel";
 
 	private final MessageChannel scatterChannel;
 
@@ -165,7 +166,8 @@ public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler i
 		MessageHeaders headers = message.getHeaders();
 		return getMessageBuilderFactory()
 				.fromMessage(message)
-				.setHeader(MessageHeaders.ERROR_CHANNEL, headers.get(GATHER_RESULT_CHANNEL))
+				.setHeader(MessageHeaders.ERROR_CHANNEL, headers.get(ORIGINAL_ERROR_CHANNEL))
+				.removeHeaders(ORIGINAL_ERROR_CHANNEL)
 				.build();
 	}
 
@@ -178,6 +180,7 @@ public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler i
 				getMessageBuilderFactory()
 						.fromMessage(requestMessage)
 						.setHeader(GATHER_RESULT_CHANNEL, gatherResultChannel)
+						.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.getErrorChannel())
 						.setReplyChannel(this.gatherChannel)
 						.setErrorChannelName(this.errorChannelName)
 						.build();
@@ -185,14 +188,11 @@ public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler i
 		this.messagingTemplate.send(this.scatterChannel, scatterMessage);
 
 		Message<?> gatherResult = gatherResultChannel.receive(this.gatherTimeout);
-
 		if (gatherResult != null) {
-			if (gatherResult instanceof ErrorMessage) {
-				throw (RuntimeException) gatherResult.getPayload();
-			}
 			return getMessageBuilderFactory()
 					.fromMessage(gatherResult)
 					.setHeader(GATHER_RESULT_CHANNEL, requestMessageHeaders.get(GATHER_RESULT_CHANNEL))
+					.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.get(ORIGINAL_ERROR_CHANNEL))
 					.setHeader(MessageHeaders.REPLY_CHANNEL, requestMessage.getHeaders().getReplyChannel())
 					.setHeader(MessageHeaders.ERROR_CHANNEL, requestMessage.getHeaders().getErrorChannel())
 					.build();
--- a/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
@@ -38,6 +38,7 @@ import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.PollableChannel;
 import org.springframework.messaging.SubscribableChannel;
 import org.springframework.messaging.support.ChannelInterceptor;
+import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.messaging.support.InterceptableChannel;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
@@ -55,10 +56,6 @@ import org.springframework.util.ClassUtils;
 public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler implements Lifecycle {
 
 	private static final String GATHER_RESULT_CHANNEL = "gatherResultChannel";
-
-	private static final String ORIGINAL_REPLY_CHANNEL = "originalReplyChannel";
-
-	private static final String ORIGINAL_ERROR_CHANNEL = "originalErrorChannel";
 
 	private final MessageChannel scatterChannel;
 
@@ -168,9 +165,7 @@ public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler i
 		MessageHeaders headers = message.getHeaders();
 		return getMessageBuilderFactory()
 				.fromMessage(message)
-				.setHeader(MessageHeaders.REPLY_CHANNEL, headers.get(ORIGINAL_REPLY_CHANNEL))
-				.setHeader(MessageHeaders.ERROR_CHANNEL, headers.get(ORIGINAL_ERROR_CHANNEL))
-				.removeHeaders(ORIGINAL_REPLY_CHANNEL, ORIGINAL_ERROR_CHANNEL)
+				.setHeader(MessageHeaders.ERROR_CHANNEL, headers.get(GATHER_RESULT_CHANNEL))
 				.build();
 	}
 
@@ -183,8 +178,6 @@ public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler i
 				getMessageBuilderFactory()
 						.fromMessage(requestMessage)
 						.setHeader(GATHER_RESULT_CHANNEL, gatherResultChannel)
-						.setHeader(ORIGINAL_REPLY_CHANNEL, requestMessageHeaders.getReplyChannel())
-						.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.getErrorChannel())
 						.setReplyChannel(this.gatherChannel)
 						.setErrorChannelName(this.errorChannelName)
 						.build();
@@ -192,16 +185,16 @@ public class ScatterGatherHandler extends AbstractReplyProducingMessageHandler i
 		this.messagingTemplate.send(this.scatterChannel, scatterMessage);
 
 		Message<?> gatherResult = gatherResultChannel.receive(this.gatherTimeout);
-		if (gatherResult != null && (requestMessageHeaders.containsKey(ORIGINAL_REPLY_CHANNEL)
-				|| requestMessageHeaders.containsKey(ORIGINAL_ERROR_CHANNEL)
-				|| requestMessageHeaders.containsKey(GATHER_RESULT_CHANNEL))) {
+
+		if (gatherResult != null) {
+			if (gatherResult instanceof ErrorMessage) {
+				throw (RuntimeException) gatherResult.getPayload();
+			}
 			return getMessageBuilderFactory()
 					.fromMessage(gatherResult)
-					.removeHeader(GATHER_RESULT_CHANNEL)
 					.setHeader(GATHER_RESULT_CHANNEL, requestMessageHeaders.get(GATHER_RESULT_CHANNEL))
-					.setHeader(ORIGINAL_REPLY_CHANNEL, requestMessageHeaders.get(ORIGINAL_REPLY_CHANNEL))
-					.setHeader(ORIGINAL_ERROR_CHANNEL, requestMessageHeaders.get(ORIGINAL_ERROR_CHANNEL))
 					.setHeader(MessageHeaders.REPLY_CHANNEL, requestMessage.getHeaders().getReplyChannel())
+					.setHeader(MessageHeaders.ERROR_CHANNEL, requestMessage.getHeaders().getErrorChannel())
 					.build();
 		}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
@@ -48,6 +48,7 @@ import org.springframework.util.ClassUtils;
  *
  * @author Artem Bilan
  * @author Abdul Zaheer
+ * @author Jayadev Sirimamilla
  *
  * @since 4.1
  */
--- a/spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
@@ -926,10 +926,10 @@ public class RouterTests {
 			return flow -> flow.scatterGather(s -> s.applySequence(true)
 									.recipientFlow(inflow -> inflow
 											.scatterGather(s1 -> s1.applySequence(true)
-															.recipientFlow(IntegrationFlowDefinition::bridge)
-													, g -> g.outputProcessor(MessageGroup::getOne)
-											))
-							, g -> g.outputProcessor(MessageGroup::getOne)
+															.recipientFlow(IntegrationFlowDefinition::bridge),
+													g -> g.outputProcessor(MessageGroup::getOne)
+											)),
+					g -> g.outputProcessor(MessageGroup::getOne)
 					);
 		}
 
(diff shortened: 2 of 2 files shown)


artembilan:
Any tests, please, to confirm an issue?
Does it fail in the previous `5.1.x` generation?

Thanks. I’ll take a look tomorrow

sirimamilla:
Hi @artembilan,

5.1.7 and above versions are also failing. 
I have tested with the scattergather test cases i have given in https://github.com/sirimamilla/ScatterGatherTest

artembilan:
> I have tested with the scattergather test cases i have given in

That's great and I'll take a look into your sample because I still can't make a sense from what you're saying.
But the point is that we can't accept a contribution without a reasoning test case. So, I can revert the change in the `main` code and that new test is going to fail for me or something else.
But just a change in the behavior is fully not good. You may fix your own use-case, but may break some other.

So, consider to add something testable into a `RouterTests`.

Thank you!

sirimamilla:
HI @artembilan,

I have ran all you existing test cases from spring integration core, they are not impacted due to this change.

artembilan:
> I have ran all you existing test cases from spring integration core

That's good sign, but we still can't accept the change without tests. Or some reason why tests can't be presented for this fix.

Since I haven't had from you anything to play with and really be able to reproduce I still doubt that we need to fix something in the framework.

sirimamilla:
Hi @artembilan,

I write my testcases using spring integration dsl, same is present in below project.
But the coding style in for scatter gather tests is configuration xml. I will add test cases to the project to ensure future enhancements will not face this problem.

https://github.com/sirimamilla/ScatterGatherTest

artembilan:
> But the coding style in for scatter gather tests is configuration xml

I'm not sure what you mean since I'm asking you to add exactly Java DSL test-case into an existing `org.springframework.integration.dsl.routers.RouterTests`.

So, that indeed is going to be an `IntegrationFlow` and unit test against it.
It has to fail before the fix and has to pass with the fix.
That's an idea how to proceed.

Actually I indeed have just realized that we don't have any `scatter-gather -> scatter-gather` test-cases...

Thank you for minimizing the code in the sample app! 
Looking into that...

sirimamilla:
Hi @artembilan ,

I have added a test case that will fail before code fix is done.

org.springframework.integration.dsl.routers.RouterTests > testNestedScatterGatherSuccess FAILED
    java.lang.AssertionError at RouterTests.java:609
18 tests completed, 1 failed

After the test fix is applied, testcase succeeds.

BUILD SUCCESSFUL in 18s
15 actionable tasks: 6 executed, 9 up-to-date
1:37:06 AM: Tasks execution finished ':spring-integration-core:cleanTest :spring-integration-core:test --tests "org.springframework.integration.dsl.routers.RouterTests"'.

artembilan:
So, I have minimized your config to this:
```
@Bean
public IntegrationFlow scatterGatherFlow() {
	return flow -> flow
			.scatterGather(s -> s
							.applySequence(true)
							.recipientFlow(f -> f
									.scatterGather(scatterer -> scatterer
											.applySequence(true)
											.recipientFlow(IntegrationFlowDefinition::bridge))),
					g -> g.outputProcessor(MessageGroup::getOne));
}
```
and it really works on version `5.1.6` and started to fail since version `5.1.7`. I mean it really hangs waiting for reply from something downstream.

Please, consider to apply such a simply config for your unit test here in your fix and I'll take a look meanwhile what made it working before.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
```diff
@@ -190,7 +190,21 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
 
 		this.messagingTemplate.send(this.scatterChannel, scatterMessage);
 
-		return gatherResultChannel.receive(this.gatherTimeout);
+		Message<?> gatherResult = gatherResultChannel.receive(this.gatherTimeout);
+		if (gatherResult != null && (requestMessageHeaders.containsKey(ORIGINAL_REPLY_CHANNEL)
```
artembilan: I think the fix is much simpler:
```
if (gatherResult != null) {
    return gatherResult.getPayload();
}
```

and `AbstractReplyProducingMessageHandler` super class will take care about proper request message headers copying. I mean that your logic is fully the same what would happen by default.
So, the problem that we return fro this `handleRequestMessage()` the whole message with some headers.
And those headers are not overridden in the super class.

spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
```diff
@@ -898,6 +917,30 @@ public IntegrationFlow propagateErrorFromGatherer(TaskExecutor taskExecutor) {
 							sg -> sg.gatherTimeout(100))
 					.get();
 		}
+		@Bean
+		public IntegrationFlow scatterGatherflow() {
```
artembilan: So, when you change this flow definition to the simpler one I provided in the comments here in the PR, I'm going to accept it and merge.

artembilan (changes requested):
@sirimamilla ,

Thank you for hard work on the issue!

Please, add also your name to the `@author` list of the affected classes.

I really appreciate your effort!

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
sirimamilla: Hi @artembilan ,

AbstractReplyProducingMessageHandler will not have the context of any new headers introduced by the subprocesses.

if (gatherResult != null) {
    return gatherResult.getPayload();
}

Losing of headers produced by downstream is going to be a problem. where in if developer sets some additional infromation required by some complex flow in headers those can be lost.

Since we are forced to introduce the logic in handleRequestMessage, can we also remove the addition of the ORIGINAL_REPLY_CHANNEL, ORIGINAL_ERROR_CHANNEL & GATHER_RESULT_CHANNEL in the request message to declutter the headers in all subflows and simplify the entire logic?

If you agree to that I can handle these changes as well.

spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
sirimamilla: Hi @artembilan,

I have made the requested changes and pushed the changes to GIT.

I will work on simplifying the scatter gather by removing additional headers in the messages and taking the same from requestMessage context in handleRequestMessage.

This will take some time as I will have to review the impact one more time.

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
artembilan: Yeah... I see your point. I have missed the fact that we return headers produced by the sub-flow. Even if we won't need them in the end, it doesn't mean that we can't consult those headers in the `gatherer` process.
So, yes, please, go ahead with your idea: we should restore request headers, but can keep all other new headers.

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
```diff
@@ -182,15 +178,27 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
...
 						.build();
 
 		this.messagingTemplate.send(this.scatterChannel, scatterMessage);
 
-		return gatherResultChannel.receive(this.gatherTimeout);
+		Message<?> gatherResult = gatherResultChannel.receive(this.gatherTimeout);
+
+		if (gatherResult != null) {
+			if (gatherResult instanceof ErrorMessage) {
+				throw (RuntimeException) gatherResult.getPayload();
```
artembilan: I don't think this is a framework responsibility to make such an unconditional decision.
In many other places we really just produce a returned message as is without any assumption about its content or type.
It would be much better to let end-user to make a logic from such a returned message.

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
```diff
@@ -182,15 +178,27 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
...
+			}
+			return getMessageBuilderFactory()
+					.fromMessage(gatherResult)
+					.setHeader(GATHER_RESULT_CHANNEL, requestMessageHeaders.get(GATHER_RESULT_CHANNEL))
+					.setHeader(MessageHeaders.REPLY_CHANNEL, requestMessage.getHeaders().getReplyChannel())
+					.setHeader(MessageHeaders.ERROR_CHANNEL, requestMessage.getHeaders().getErrorChannel())
```
artembilan: I would say we just need to remove these headers and let a super class to populate them!

Yeah, now I see your point about those `ORIGINAL_` headers... Good catch!

artembilan (changes requested):
Please, see my latest review.
I think we are very close to the merge! 😄

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
sirimamilla: Hi @artembilan,

When we remove the replyChannel and errorChannel Header mapping, super class is not populating the reply and error channel. This is causing the scatterGather to fail in the nested ScatterGather scenario.

org.springframework.integration.dsl.routers.RouterTests > testNestedScatterGather FAILED
    java.lang.AssertionError at RouterTests.java:545
18 tests completed, 1 failed


Also, we need to keep the headers as the headers populated by Output processor needs to be preserved. This is very common for application to populate the information in headers and later reference them in downstream processing.

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
sirimamilla: HI @artembilan ,

Good Morning!!

I understand it is not framework responsibility to handle functional scenario's. But here it is behaviour of the ScatterGather that needs to be consistent. when outputProcessor throws an exception, ScatterGather needs to propagate it to proper Error Channel.

When output processor throws exception, the exception will be wrapped in Error message, it will be propagated to ScatterGatherResultChannel. This message will be propagated to next channel in the FlowDefinition.

Please have a look at the flow in the testcase propagateErrorFromGatherer.

`@Bean
		public IntegrationFlow propagateErrorFromGatherer(TaskExecutor taskExecutor) {
			return IntegrationFlows.from(Function.class)
					.scatterGather(s -> s
									.applySequence(true)
									.recipientFlow(subFlow -> subFlow
											.channel(c -> c.executor(taskExecutor))
											.transform(p -> "foo")),
							g -> g
									.outputProcessor(group -> {
										throw new RuntimeException("intentional");
									}),
							sg -> sg.gatherTimeout(100))
					.transform(m -> "This should not be executed, results must have been propagated to Error Channel")
					.get();
		}`

When `.transform(m -> "This should not be executed, results must have been propagated to Error Channel")` was not added, test case passed with success as it was able to get the Expected error response from the Error Channel.

But same test case failed when the .transform is added. This behavior needs to be corrected by handling the Exception. Similar checks are performed in the class MessagingGatewaySupport.java. when reply message is received, it is checked for instance of ErrorMessage or Throwable class

`	@Nullable // NOSONAR
	private Object doSendAndReceive(Object object, boolean shouldConvert) {
		initializeIfNecessary();
		Assert.notNull(object, "request must not be null");
		MessageChannel channel = getRequestChannel();
		if (channel == null) {
			throw new MessagingException("No request channel available. Cannot send request message.");
		}

		registerReplyMessageCorrelatorIfNecessary();

		Object reply;
		Message<?> requestMessage = null;
		try {
			if (this.countsEnabled) {
				this.messageCount.incrementAndGet();
			}
			if (shouldConvert) {
				reply = this.messagingTemplate.convertSendAndReceive(channel, object, Object.class,
						this.historyWritingPostProcessor);
			}
			else {
				requestMessage = (object instanceof Message<?>)
						? (Message<?>) object : this.requestMapper.toMessage(object);
				Assert.state(requestMessage != null, () -> "request mapper resulted in no message for " + object);
				requestMessage = this.historyWritingPostProcessor.postProcessMessage(requestMessage);
				reply = this.messagingTemplate.sendAndReceive(channel, requestMessage);
			}

			if (reply == null && this.errorOnTimeout) {
				throwMessageTimeoutException(object, "No reply received within timeout");
			}
		}
		catch (Exception ex) {
				logger.debug("failure occurred in gateway sendAndReceive: " + ex.getMessage());
			}
			reply = ex;
		}

		if (reply instanceof Throwable || reply instanceof ErrorMessage) {
			Throwable error =
					reply instanceof ErrorMessage
							? ((ErrorMessage) reply).getPayload()
							: (Throwable) reply;
			return handleSendAndReceiveError(object, requestMessage, error, shouldConvert);
		}
		return reply;
	}
`

artembilan: Well, this is definitely not a `gateway` here. We check there for `ErrorMessage` because a high-level contract is a plain old Java method invocation, so we need to honor its signature.

Here we talk about internals of Spring Integration which deal with messages. I fully not agree to re-throw that exception. If you return an `ErrorMessage` from the gatherer, it is already up to your downstream flow to process it this or other way. The logic has to be a simple as possible. We should not mix concerns in one component. We have hidden already the maximum, but that's not our responsibility to make a decision about returned message. We only need to correct headers nothing more. Even if it is not an `ErrorMessage`, you still can return an `Exception` in the `payload`. The flow you are mentioned really has to be mentioned with a `handle()` or `filter()` to decide what to do with that exception. That's not what we want to do in the `Scatter-Gather` hard logic.

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
```diff
@@ -167,9 +165,7 @@ else if (this.gatherChannel instanceof ReactiveStreamsSubscribableChannel) {
 		MessageHeaders headers = message.getHeaders();
 		return getMessageBuilderFactory()
 				.fromMessage(message)
-				.setHeader(MessageHeaders.REPLY_CHANNEL, headers.get(ORIGINAL_REPLY_CHANNEL))
-				.setHeader(MessageHeaders.ERROR_CHANNEL, headers.get(ORIGINAL_ERROR_CHANNEL))
-				.removeHeaders(ORIGINAL_REPLY_CHANNEL, ORIGINAL_ERROR_CHANNEL)
+				.setHeader(MessageHeaders.ERROR_CHANNEL, headers.get(GATHER_RESULT_CHANNEL))
```
artembilan: I find this as a wrong decision. We have that `errorChannelName` to process exceptions in the scatterer sub-flows and gatherer.

artembilan (changes requested):
Please, stop mixing concerns. 
Every component in the framework must perform its own purpose.
That's really not what EIP dictates to us how to compose integration flows.

Thanks for understanding.

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
sirimamilla: @artembilan,
errorChannelName  is meant to handle errors in the subflows. The result will flow back to Gather Channel.
If there is an error the Gatherer and if we propagate the error back to errorChannelName, it will cause the Scatter Gather to hang for ever if any exception thrown in outputProcessor.

spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
sirimamilla: Hi @artembilan 

If you see the implemetation of the AbstractCorrelatingMessageHandler (gather handler), when the outprocessor throws exception this will be wrapped in the error message and sent to Error Channel.

Now if we don't throw error in the handleRequestMessage, message will flow to the next channel in the flow instead of being raised as exception.

sirimamilla:
Hi @artembilan,

If we closely see all 2 issues raised by you are connected to each other.

Design Choice 1:(Original Solution before this PR)
If we set OriginalErrorChannel as the errorChannel(as was the case earlier) there is no need of handling  Error condition as Gather will throw response directly on parent error queue.single scatter Gather works fine. 

			
This solution fails in case of nested scatter gather as the GatherChannel and the Reply Channels in the error message are not corrected, message will deliver to wrong queue.

Design Choice 2:
Set GatherResult Channel as Error Channel, Do Explicit checking for ErrorMessage and throw it so the response will be received by the parent error chanel along with the right message channels set by super class

This solution will work in the nested ScatterGather situatuion.

Design Choice 3:
Set GatherResult Channel as Error Channel, just pass on the error and success result to next channel in the flow. Developer will have to handle throwing of exception if need arises.

My Choice:
I would prefer to go with Design choice 2 as it ensures the behavior of the Scatter Gather is inline with all other patterns.

artembilan:
1. We should not abuse an `errorChannel` to return result. That is not for what that `gatherResultChannel` has been designed.
2. We definitely need to send an error into a correct `errorChannel`. When we have thread shifting in the sub-flow, the gatherer is going to be called exactly on that channel. I need to play with this, but it looks like a `errorChannelName`  is in use here.
3. We definitely should not make a decision about errors in this components. This is exactly not its responsibility. 
4. Your change is not aligning Scatter Gather with other components because they also don't do any assumption about errors. But seems for me it is not related to the original issue for this change. We have a problem with nested Scatter-Gather, but you have started mixing something about errors which could be address in other issues. 

What I think we are missing is just a fact of not restored `ORIGINAL_REPLY_CHANNEL` & `ORIGINAL_ERROR_CHANNEL`.

So, let's fix the nested scatter-gather issue first and then we will see if need to do anything else!

artembilan (changes requested):
Please, don't mix error processing into this issue.

sirimamilla:
I am restoring the OriginalErrorChannel Header back as it was before this change and remove the additional Error Handling. For Error Handling, I will raise separate issue tomorrow with supporting test cases.

artembilan:
I found the reason.
We have missed the fact to re-instantiate a `GATHER_RESULT_CHANNEL` header. Since super classes uses a `filterAndCopyHeadersIfAbsent()` logic it doesn't restore an original `GATHER_RESULT_CHANNEL` for us.

So, I'm merging this with your tests and a working fix in the `ScatterGatherHandler`.

Thank you for your contribution; looking forward for more!

sirimamilla:
> I found the reason.
> We have missed the fact to re-instantiate a `GATHER_RESULT_CHANNEL` header. Since super classes uses a `filterAndCopyHeadersIfAbsent()` logic it doesn't restore an original `GATHER_RESULT_CHANNEL` for us.
> 
> So, I'm merging this with your tests and a working fix in the `ScatterGatherHandler`.
> 
> Thank you for your contribution; looking forward for more!

I have reverted the error related changes after your comment. If you want me to put back the error handling changes, I will be happy to do so

artembilan:
Thank you! I'm looking into your latest changes locally now.


>  If you want me to put back the error handling changes

As I said: I'm against any error handling inside this component. the gatherer should throw an exception as is or push it into an original error channel. The returned `ErrorMessage` should be pushed down stream as is.

I would say the logic has to be covered with simple generic Service Activator: if we wouldn't have a Scatter-Gather implementation, we would be able to implement with a custom code which would be called from the Service Activator.
Any error handling branches should be delegated to dedicated components on the matter.

sirimamilla:
> Thank you for your contribution; looking forward for more!

I would like to contribute on any fixes identified, earlier, we had issue with compliance, so couldn't do the checkins to GITHUB. Now, I am allowed to contribute in my personal time.

artembilan:
Merged as: https://github.com/spring-projects/spring-integration/commit/c2c44c257cace1e9a4aab8548369f411b57fafe1

Cherry-picked to `5.1.x` as: https://github.com/spring-projects/spring-integration/commit/35fd383399bd1b0a69c8ebc36b7f10512611a852

Cherry-picked to `master` as: https://github.com/spring-projects/spring-integration/commit/35fd383399bd1b0a69c8ebc36b7f10512611a852

@sirimamilla ,

thank you very much for contribution and so thorough. You definitely made to doubt in some places and re-think our solutions. 😄 

Looking forward for more!

PR #3159, opened 2020-01-27
Fix for Issue with AbstractCorrelatingMessageHandler in handling sequences when Message is returned from outputProcessor

sirimamilla:
In AbstractCorrelatingMessageHandler, while processing the outputProcessor result, sequences are not being popped when result is an instance of Message. 

An additional condition is added to handle such sequences.

			else if (this.popSequence && partialSequence == null && result instanceof Message && (ObjectUtils.nullSafeEquals(((Message) result).getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS), message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS)))) {
				result = getMessageBuilderFactory()
						.fromMessage((Message) result).popSequenceDetails();
			}
		}

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlinVersion = '1.3.60'
+	ext.kotlinVersion = '1.3.61'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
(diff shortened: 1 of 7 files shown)


artembilan:
Please, consider to pull latest `master`, rebase your branch to that one and then re-push for PR.

Issue #3157, opened 2020-01-27
Issue with AbstractCorrelatingMessageHandler in handling sequences if used in nested ScatterGather sequence


sirimamilla:
**Affects Version(s):** 5.2.x, 5.1.x
---
If the output processor in ScatterGather returns a message, Scatter Gather is hanging indefinitely. 
This issue is present with all versions of Spring Integration.

This can be tested with SI version 5.1.3

Please find the test case here in the example git project.
https://github.com/sirimamilla/ScatterGatherTest

sirimamilla:
I have the fix for the issue, once the issue is accepted, I will open pull request for the same. Currently I am not able to raise the pull request as it is dependent of fix for https://github.com/spring-projects/spring-integration/issues/3152

This fix needs to be done along with or after Pull Request https://github.com/spring-projects/spring-integration/pull/3153. To test the fix, issue 3152 needs to be resolved.

artembilan:
@sirimamilla ,

tell me, please, how this issue is different from that one you are mentioning in the opened PR.

Thanks

sirimamilla:
Hi @artembilan,

The issue mentioned in PR is a new bug that is introduced in 5.1.7 and above and is related to Original_ Header handling.

This issue is related to Sequence Handling that has one unhandled scenario. If we agree that this fix can be merged along with PR #3153. I will push this fix to PR #3153

artembilan:
No, let's hold of from complicating that PR.
I'm about for final review and merging that one.

We can come back to this afterwards.

So, stay tuned!

sirimamilla:
I have not pushed the changes to Test application, so you were not able to see the difference. you can review test application now.

PR #3160, opened 2020-01-27
Fix for Issue with AbstractCorrelatingMessageHandler in handling sequences when Message is returned from outputProcessor 

sirimamilla:
In AbstractCorrelatingMessageHandler, while processing the outputProcessor result, sequences are not being popped when result is an instance of Message.

https://github.com/spring-projects/spring-integration/issues/3157

An additional condition is added to handle such sequences.

			else if (this.popSequence && partialSequence == null && result instanceof Message && (ObjectUtils.nullSafeEquals(((Message) result).getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS), message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS)))) {
				result = getMessageBuilderFactory()
						.fromMessage((Message) result).popSequenceDetails();
			}

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,6 +59,7 @@ import org.springframework.messaging.MessageHandlingException;
 import org.springframework.messaging.core.DestinationResolutionException;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
+import org.springframework.util.ObjectUtils;
 
 /**
  * Abstract Message handler that holds a buffer of correlated messages in a
@@ -89,6 +90,7 @@ import org.springframework.util.CollectionUtils;
  * @author David Liu
  * @author Enrique Rodriguez
  * @author Meherzad Lahewala
+ * @author Jayadev Sirimamilla
  *
  * @since 2.0
  */
@@ -841,6 +843,10 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 							.copyHeaders(message.getHeaders());
 				}
 				result = messageBuilder.popSequenceDetails();
+			}
+			else if (this.popSequence && partialSequence == null && result instanceof Message && (ObjectUtils.nullSafeEquals(((Message) result).getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS), message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS)))) {
+				result = getMessageBuilderFactory()
+						.fromMessage((Message) result).popSequenceDetails();
 			}
 		}
 		finally {
--- a/spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/dsl/routers/RouterTests.java
@@ -613,6 +613,20 @@ public class RouterTests {
 
 	}
 
+	@Test(timeout = 11000)
+	public void testNestedScatterGatherSequenceTest() {
+		PollableChannel replyChannel = new QueueChannel();
+		this.scatterGatherInSubFlowChannel.send(
+				MessageBuilder.withPayload("sequencetest")
+						.setReplyChannel(replyChannel)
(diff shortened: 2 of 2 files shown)


artembilan:
Sounds like your fix is in the conflict of what we claim in docs already: https://docs.spring.io/spring-integration/docs/5.2.3.RELEASE/reference/html/message-routing.html#aggregator-api

> Starting with version 5.1, after processing message group, an `AbstractCorrelatingMessageHandler` performs a `MessageBuilder.popSequenceDetails()` message headers modification for the proper splitter-aggregator scenario with several nested levels. It is done only if the message group release result is not a message or collection of messages. In that case a target `MessageGroupProcessor` is responsible for the `MessageBuilder.popSequenceDetails()` call while building those messages.

See the fix here: https://github.com/spring-projects/spring-integration/pull/2591

We can't make an assumption what you have done with the `Message` returned from the `outputProcessor`.

So, any further thoughts ?

sirimamilla:
HI @artembilan ,

SeqenceDetails are pushed by Splitter/ScatterGather when applySequence is set to true by the framework. Ideally the responsibiity of poping the sequence should be with the framework.

popSequence is the field which decides whether to PopSequence or not. In this case default value of the field is true.

In case popsequence should not be done this flag should be set to false then the condition will not become true anyways.

artembilan:
It works exactly as you explain, but only for the case when your output processor doesn't return a `Message<?>`.
We did that intentionally because in our opinion the whole `Message` is end-user decision and we definitely should not impact that decision.

However I see your point and having a `popSequence` flag we are really conditional here and leave the final decision what to mutate or not really up to end-user.

@garyrussell ,

WDYT, can we relax our rule on the matter and let the returned message to be modified for that `popSequenceDetails()` ?
In the end it is just target project and it might be even abuse to make a strong resolution for end-user data in the framework.

Thanks

garyrussell:
I think it's ok given that the patch compares the sequence details headers - i.e. if they are unchanged from the original it's ok to pop them.

We will need a doc change, though, to indicate the pop will occur with a message result when the PSDs are unchanged.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -842,6 +844,10 @@ protected void completeGroup(Object correlationKey, MessageGroup group, Lock loc
 				}
 				result = messageBuilder.popSequenceDetails();
 			}
+			else if (this.popSequence && partialSequence == null && result instanceof Message && (ObjectUtils.nullSafeEquals(((Message) result).getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS), message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_DETAILS)))) {
```
garyrussell: This line's too long (> 120).

Also, the if/elses can be restructured to avoid double checks on `this.popSequence && partialSequence == null && result instanceof Message`

artembilan:
>  if they are unchanged from the original it's ok to pop them.

Do you mean we should compare that header from the returned message with the one in the request message?
Then (if it is possible) it doesn't matter what is the returned result and we even might not need that `popSequence` flag at all...

I consider this as a breaking change therefore we can apply it only for the current `5.3` generation.

If we agree with all the suggestion, @sirimamilla , consider to re-push your PR against `master`: we definitely are not going to introduce a breaking change in the minor release.

Thank you!

garyrussell:
He's already comparing the sequence details (unless I am missing something).

artembilan:
> He's already comparing the sequence details

OK. I didn't look into a change because was in doubt to accept it at all.
Now since we have agreed on the change, let's proceed with review, but I'd like to see it only against `master`.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
sirimamilla: I will restructure the If else, I kept it like this to ensure minimum changes for ease of review.

sirimamilla:
Hi @artembilan,
This is not a breaking change as we are ensuring to compare sequence before poping them.

If you agree, I will update the doc and push the change again tonight.

artembilan:
It is a breaking change since your test fails exactly according a documented behavior. Someone may rely on the current behavior. What you suggest is really convenient, but it cannot be a change in behavior in minor release. I must upgrade my project into the latest minor version without any changes in the project. Imaging a situation when I don’t pop sequence in my output processor and I know that aggregator doesn’t do that for me according docs. I do that somewhere later. But with your change it is going to brake and my code will pop again to lead into unexpected behavior in my project. 

Does it sound convincing?

garyrussell:
Plus there is a work-around for previous releases - pop the sequence in your custom processor.

sirimamilla:
Closing Pull request. I have raised a new Pull request for master.

PR #3162, opened 2020-01-30
GH-3154: Support `UriBuilderFactory.EncodingMode`

### LINKED ISSUE #3154: URI Is Being Generated With URI Variables Before RestTemplate Does It
https://github.com/spring-projects/spring-integration/blob/cf01b7c0e5007f7da1094230ddcc08b17acac6af/spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java#L301

Spring Integration is generating the entire URI with the declared uri variables, and then passing the generated URI to the RestTemplate exchange method defined in HttpRequestExecutingMessageHandler.exchange()

Here is a truncated declaration of the outbound gateway with the LANGUAGE query parameter templated
```
<int-http:outbound-gateway
                url="/properties/availability?language={LANGUAGE}"                
                rest-template="restTemplate"
        >            
            <int-http:uri-variable name="LANGUAGE" expression="headers['LANGUAGE']"/>            
        </int-http:outbound-gateway>
```

This is causing issues in Micrometer which auto instruments the RestTemplate, and is creating a new Timer for every request as the path/query parameters differ. See https://github.com/micrometer-metrics/micrometer/issues/98 for the issue.

From the above we would expect metrics to be recorded like

`httpClientRequests{clientName="test",method="GET",status="200",uri="/properties/availability&language={LANGUAGE}",quantile="0.5"}`

However what we are seeing is

`httpClientRequests{clientName="test",method="GET",status="200",uri="/properties/availability&language=en-US",quantile="0.5"}
httpClientRequests{clientName="test",method="GET",status="200",uri="/properties/availability&language=de-DE",quantile="0.5"}
httpClientRequests{clientName="test",method="GET",status="200",uri="/properties/availability&language=ko-KR",quantile="0.5"}
........
`

I believe the correct behavior should be that the URI and the URI variables should be passed to the HttpRequestExecutingMessageHandler.exchange() method, so that RestTemplate can map the variables to the URI. This would correctly create the Micrometer timers for the RestTemplate

The call currently is

`httpResponse = this.restTemplate.exchange(uri, httpMethod, httpRequest,
						(ParameterizedTypeReference<?>) expectedResponseType);

`

but it should be 

`httpResponse = this.restTemplate.exchange(uri, httpMethod, httpRequest,
						(ParameterizedTypeReference<?>) expectedResponseType, uriVariables);
`

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3154

Spring Framework now provides a `DefaultUriBuilderFactory.EncodingMode`
for encoding URIs in the `RestTemplate` before and after uri template
enrichment with uri variables.
Therefore `encodeUri` and manual uri variables substitution is not necessary
in Spring Integration HTTP components

* Deprecate `AbstractHttpRequestExecutingMessageHandler.encodeUri` in favor of
`DefaultUriBuilderFactory.EncodingMode` and respective configuration
on the `RestTemplate` in HTTP module and `WebClient` in WebFlux module

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,6 @@
 package org.springframework.integration.http.outbound;
 
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
@@ -27,7 +26,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.function.Supplier;
 
 import javax.xml.transform.Source;
 
@@ -56,14 +54,13 @@ import org.springframework.integration.support.AbstractIntegrationMessageBuilder
 import org.springframework.integration.support.MessageBuilderFactory;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
-import org.springframework.messaging.MessageHandlingException;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 import org.springframework.util.StringUtils;
-import org.springframework.web.util.UriComponents;
+import org.springframework.web.util.DefaultUriBuilderFactory;
 import org.springframework.web.util.UriComponentsBuilder;
 
 /**
@@ -85,6 +82,8 @@ public abstract class AbstractHttpRequestExecutingMessageHandler extends Abstrac
 	private static final List<HttpMethod> NO_BODY_HTTP_METHODS =
 			Arrays.asList(HttpMethod.GET, HttpMethod.HEAD, HttpMethod.TRACE);
 
+	DefaultUriBuilderFactory uriFactory = new DefaultUriBuilderFactory();
+
 	private final Map<String, Expression> uriVariableExpressions = new HashMap<>();
 
 	private final Expression uriExpression;
@@ -94,8 +93,6 @@ public abstract class AbstractHttpRequestExecutingMessageHandler extends Abstrac
 	private SimpleEvaluationContext simpleEvaluationContext;
 
 	private boolean trustedSpel;
-
-	private boolean encodeUri = true;
 
 	private Expression httpMethodExpression = new ValueExpression<>(HttpMethod.POST);
 
@@ -118,6 +115,7 @@ public abstract class AbstractHttpRequestExecutingMessageHandler extends Abstrac
 	public AbstractHttpRequestExecutingMessageHandler(Expression uriExpression) {
 		Assert.notNull(uriExpression, "URI Expression is required");
 		this.uriExpression = uriExpression;
+		this.uriFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.TEMPLATE_AND_VALUES);
 	}
 
 	/**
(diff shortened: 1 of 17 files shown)


artembilan:
TODO:

* The same support for WebFlux module
* Docs
* Migration Guide
* Clean up affected test cases for code style and JUnit 5

Will address them lately, but you can review now.

Thanks

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
```diff
@@ -118,6 +115,7 @@
 	public AbstractHttpRequestExecutingMessageHandler(Expression uriExpression) {
 		Assert.notNull(uriExpression, "URI Expression is required");
 		this.uriExpression = uriExpression;
+		this.uriFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.TEMPLATE_AND_VALUES);
```
garyrussell: Why is this needed? This is the default.

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
```diff
@@ -85,6 +82,8 @@
 	private static final List<HttpMethod> NO_BODY_HTTP_METHODS =
 			Arrays.asList(HttpMethod.GET, HttpMethod.HEAD, HttpMethod.TRACE);
 
+	DefaultUriBuilderFactory uriFactory = new DefaultUriBuilderFactory();
```
garyrussell: Why package visibility?

Why not `UriBuilderFactory`?

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
```diff
@@ -127,9 +125,25 @@ public AbstractHttpRequestExecutingMessageHandler(Expression uriExpression) {
...
+						: DefaultUriBuilderFactory.EncodingMode.NONE);
+	}
+
+	/**
+	 * Set the encoding mode to use.
+	 * By default this is set to {@link DefaultUriBuilderFactory.EncodingMode#TEMPLATE_AND_VALUES}.
+	 * @param encodingMode the mode to use for uri encoding
+	 * @since 5.3
+	 */
+	public void setEncodingMode(DefaultUriBuilderFactory.EncodingMode encodingMode) {
```
garyrussell: Why not `setUriBuilderFactory` ?

garyrussell (changes requested):
A few issues/questions.

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
artembilan: The `UriBuilderFactory` doesn't have a `setEncodingMode()`

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
artembilan: Right. I'm going to document that for more complex scenarios it would be better to configure an external `RestTemplate` and already use its `setUriTemplateHandler()`.

I would even go with all the `RestTempalte` options as an external reference. I'm not a fun of mutating internal objects.
But that would be too drastic even in the point release...

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
garyrussell: and package?

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
garyrussell: Ok, makes sense; proceed with the remaining work 😄

spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundChannelAdapterParserTests.java
```diff
@@ -192,13 +200,15 @@ public void withUrlAndTemplate() {
...
+		HttpRequestExecutingMessageHandler handler = (HttpRequestExecutingMessageHandler) endpointAccessor
+				.getPropertyValue("handler");
 		DirectFieldAccessor handlerAccessor = new DirectFieldAccessor(handler);
 		assertThat(handlerAccessor.getPropertyValue("expectReply")).isEqualTo(false);
 		assertThat(endpointAccessor.getPropertyValue("inputChannel"))
 				.isEqualTo(this.applicationContext.getBean("requests"));
 		assertThat(handlerAccessor.getPropertyValue("outputChannel")).isNull();
-		DirectFieldAccessor templateAccessor = new DirectFieldAccessor(handlerAccessor.getPropertyValue("restTemplate"));
+		DirectFieldAccessor templateAccessor = new DirectFieldAccessor(handlerAccessor.getPropertyValue("restTemplate"
+		));
```
garyrussell: Unusual formatting.

spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundChannelAdapterParserTests.java
```diff
@@ -224,13 +234,15 @@ public void withUrlExpression() {
...
```
garyrussell: Here too.

spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundChannelAdapterParserTests.java
```diff
@@ -257,13 +269,15 @@ public void withUrlExpressionAndTemplate() {
...
```
garyrussell: And here

garyrussell (changes requested):

garyrussell:
> Task :spring-integration-http:checkstyleTest FAILED
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundChannelAdapterParserTests.java:22:8: Unused import - java.nio.charset.Charset. [UnusedImports]

garyrussell (approved):

PR #3167, opened 2020-02-05
GH-3155: Add support for Java DSL extensions

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3155

Provide an `IntegrationFlowExtension` for possible custom EI-operators
in the target project use-cases.

PR #3170, opened 2020-02-06
AMQP: Multiple Sends within an OB Channel Adapter

garyrussell:
In preparation for: https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/209

When `multiSend` is true and multiple messages are sent as the payload
of a message, each message is sent within the `invoke()` method of the
`RabbitTemplate`.

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpBaseOutboundEndpointSpec.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpBaseOutboundEndpointSpec.java
@@ -312,7 +312,6 @@ AmqpBaseOutboundEndpointSpec<S extends AmqpBaseOutboundEndpointSpec<S, E>, E ext
 	}
 
 	/**
-	/**
 	 * Set a timeout after which a nack will be synthesized if no publisher confirm has
 	 * been received within that time. Missing confirms will be checked every 50% of this
 	 * value so the synthesized nack will be sent between 1x and 1.5x this timeout.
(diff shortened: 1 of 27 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpOutboundChannelAdapterSpec.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Spec for an outbound AMQP channel adapter
+ *
+ * @author Gary Russell
+ * @since 5.3
+ *
+ */
+public class AmqpOutboundChannelAdapterSpec extends AmqpOutboundEndpointSpec<AmqpOutboundChannelAdapterSpec> {
+
+	AmqpOutboundChannelAdapterSpec(AmqpTemplate amqpTemplate) {
```
artembilan: Gary, can you make all the ctors and `final` props in these AMQP specs as `protected` for possible extension?
So, I won't include them into my extensions PR #3167 to avoid conflicts between our two fixes.

Thanks

artembilan (changes requested):

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpBaseOutboundEndpointSpec.java
```diff
@@ -311,4 +311,17 @@ public S errorMessageStrategy(ErrorMessageStrategy errorMessageStrategy) {
...
 
+	/**
+	 * Set a timeout after which a nack will be synthesized if no publisher confirm has
+	 * been received within that time. Missing confirms will be checked every 50% of this
+	 * value so the synthesized nack will be sent between 1x and 1.5x this timeout.
+	 * @param timeout the approximate timeout.
+	 * @return the spec.
+	 * @since 5.3
+	 */
+	public S confirmTimeout(long timeout) {
```
artembilan: Right, this is one of those I'd like to cover with DSL extensions and all those `protected` ctors 😄

artembilan (approved):

Issue #3169, opened 2020-02-06
String key should be Object, cause removeSessionFactory and getSessionFactory works with Object


kezhevatov:
https://github.com/spring-projects/spring-integration/blob/509e8237e8aa374990f07d288da083eb2229bdf5/spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/DefaultSessionFactoryLocator.java#L57

artembilan:
Wow! That's good catch!

Indeed this is a bug: all other `DefaultSessionFactoryLocator` contracts are based on the `Object`, so this `addSessionFactory` has to be on `Object` as well.

As a workaround you can fully rely on the `DefaultSessionFactoryLocator(Map<Object, SessionFactory<F>> factories)` constructor.

Feel free to provide a contribution: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc 

Thanks!

artembilan:
I think for binary backward compatibility we need to deprecate an existing method and introduce a new one.
The old one will be removed in the next version.

PR #3171, opened 2020-02-06
spring-projectsGH-3169: Fix DefaultSessionFactoryLocator addSessionFactory key's type from String to Objec

kezhevatov:
spring-projectsGH-3169: Fix DefaultSessionFactoryLocator addSessionFactory key's type from String to Object

All other DefaultSessionFactoryLocator contracts are based on the Object, so this addSessionFactory has to be on Object as well.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/DefaultSessionFactoryLocator.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/DefaultSessionFactoryLocator.java
@@ -53,8 +53,20 @@ public class DefaultSessionFactoryLocator<F> implements SessionFactoryLocator<F>
 	 * Add a session factory.
 	 * @param key the lookup key.
 	 * @param factory the factory.
+	 * @deprecated since 5.3 in favor of {@link #addSessionFactory}
 	 */
+	@Deprecated
 	public void addSessionFactory(String key, SessionFactory<F> factory) {
+		addSessionFactory((Object) key, factory);
+	}
+
+	/**
+	 * Add a session factory.
+	 * @param key the lookup key.
+	 * @param factory the factory.
+	 * @since 5.3
+	 */
+	public void addSessionFactory(Object key, SessionFactory<F> factory) {
 		this.factories.put(key, factory);
 	}
 


spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/DefaultSessionFactoryLocator.java
```diff
@@ -54,7 +54,7 @@ public DefaultSessionFactoryLocator(Map<Object, SessionFactory<F>> factories, Se
 	 * @param key the lookup key.
 	 * @param factory the factory.
 	 */
-	public void addSessionFactory(String key, SessionFactory<F> factory) {
+	public void addSessionFactory(Object key, SessionFactory<F> factory) {
```
artembilan: How about this my concern: https://github.com/spring-projects/spring-integration/issues/3169#issuecomment-583058147 ?

Whenever we are going to back-port it we should keep backward compatibility, so an upgrade for the target project will be smooth. Even if we don't back-port it into a previous version, we still need to keep binary compatibility for current `5.3` version: Spring Cloud is going to support both Spring Boot 2.2 and 2.3 generations.

Therefore my concern about deprecating the current method and introduce a new one like you have done.

Thanks

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/DefaultSessionFactoryLocator.java
kezhevatov: Done. Thanks for quick answer

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/DefaultSessionFactoryLocator.java
```diff
@@ -53,11 +53,23 @@ public DefaultSessionFactoryLocator(Map<Object, SessionFactory<F>> factories, Se
 	 * Add a session factory.
 	 * @param key the lookup key.
 	 * @param factory the factory.
+	 * @deprecated since 5.3 in favor of {@link #addSessionFactory}
 	 */
+	@Deprecated
 	public void addSessionFactory(String key, SessionFactory<F> factory) {
 		this.factories.put(key, factory);
```
artembilan: I have an opinion that deprecated API should call a new one for better coverage.
This way when we need to make a change in the target logic, a deprecated API won't be affected because it is meant to be for removal eventually anyway. 😄

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/DefaultSessionFactoryLocator.java
kezhevatov: Yep, it make sense. Done

artembilan (approved):
Merging...

I'll add your name to the class author list and do some clean up locally...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/edf84a393d4dd8f5db8de0d0fdba581be86efa50 and cherry-picked to `5.2.x`.

@kezhevatov ,

thank you for the contribution; looking forward for more!

For the future, please, consider to test the project against your changes: you have missed to fix a deprecated method usage in the `DelegatingSessionFactoryTests`.

Plus it would be great to make a good commit message (not a PR description, but commit): https://chris.beams.io/posts/git-commit/

PR #3173, opened 2020-02-07
GH-3172: Support consumer-side batching

### LINKED ISSUE #3172: Add support for AMQP Consumer-side batch creation
See https://github.com/spring-projects/spring-amqp/issues/1032

Add a `ChannelAwareBatchMessageListener` and map `List<Message>` to `List<?>`.

In preparation for https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/209

In order to get efficiency on the producer side we need to enable batching on the consumer side, even when the upstream producer doesn't send messages in a batch.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3172

When the listener container supports creating batches of consumed
messages, present the batch as the message payload - either a
`List<Message<?>>` or `List<SomePayload>`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,7 +90,7 @@ ext {
 	rsocketVersion = '1.0.0-RC6'
 	servletApiVersion = '4.0.1'
 	smackVersion = '4.3.4'
-	springAmqpVersion = project.hasProperty('springAmqpVersion') ? project.springAmqpVersion : '2.2.3.RELEASE'
+	springAmqpVersion = project.hasProperty('springAmqpVersion') ? project.springAmqpVersion : '2.2.4.BUILD-SNAPSHOT'
 	springDataVersion = project.hasProperty('springDataVersion') ? project.springDataVersion : 'Neumann-BUILD-SNAPSHOT'
 	springSecurityVersion = project.hasProperty('springSecurityVersion') ? project.springSecurityVersion : '5.3.0.RC1'
 	springRetryVersion = '1.2.5.RELEASE'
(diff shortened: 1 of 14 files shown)


build.gradle
```diff
@@ -90,7 +90,7 @@ ext {
 	rsocketVersion = '1.0.0-RC6'
 	servletApiVersion = '4.0.1'
 	smackVersion = '4.3.4'
-	springAmqpVersion = project.hasProperty('springAmqpVersion') ? project.springAmqpVersion : '2.2.3.RELEASE'
+	springAmqpVersion = project.hasProperty('springAmqpVersion') ? project.springAmqpVersion : '2.2.4.BUILD-SNAPSHOT'
```
artembilan: Or you will move Spring AMQP 2.2.4 release to February 12th: https://github.com/spring-projects/spring-integration/milestone/23.

Or this PR will be placed on hold until February 26th.

But we can't release the next Spring Integration milestone without proper dependencies.

Thanks for understanding.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/EndpointUtils.java
```diff
@@ -50,9 +52,28 @@ public static ListenerExecutionFailedException errorMessagePayload(Message messa
...
 	}
 
+	/**
+	 * Return an {@link ListenerExecutionFailedException} or a {@link ManualAckListenerExecutionFailedException}
+	 * depending on whether isManualAck is false or true.
+	 * @param messages the failed messages.
+	 * @param channel the channel.
+	 * @param isManualAck true if the container uses manual acknowledgment.
+	 * @param ex the exception.
+	 * @return the exception.
```
artembilan: `@since 5.3`? Because of new public API

artembilan (changes requested):
LGTM so far. 
I'll take a look closed locally when we agree about release concern.

build.gradle
garyrussell: I moved the release to 2/12.

PR #3178, opened 2020-02-12
GH-2818: DSL support for -ws module

### LINKED ISSUE #2818: Java DSL for web services
We are starting a new project with spring Integration.
We found that it is recommended to use Java DSL to configure channels.Well, there are lots of first class integration for `Amqp`, `Feed`, `Jms`, `Files`, `(S)Ftp`, `Http`, `JPA`, `MongoDb`, `TCP/UDP`, `Mail`.
We need to integrate with `web services` (  `inboud` and `out bound`)

Is there any planning for implementing WebService DSL. 

Thanks !

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/2818

Docs to follow tomorrow.

### DIFF
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
@@ -172,9 +172,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 	 */
 	@Deprecated
 	public void setMessageSender(WebServiceMessageSender messageSender) {
-		Assert.state(!this.webServiceTemplateExplicitlySet,
-				() -> "'messageSender' must be specified on the provided: " + this.webServiceTemplate);
-		this.webServiceTemplate.setMessageSender(messageSender);
+		setMessageSenders(messageSender);
 	}
 
 	public void setMessageSenders(WebServiceMessageSender... messageSenders) {
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -165,6 +165,12 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 		this.webServiceTemplate.setFaultMessageResolver(faultMessageResolver);
 	}
 
+	/**
+	 * Specify the {@link WebServiceMessageSender} to use.
+	 * @param messageSender the sender.
+	 * @deprecated in favor of {@link #setMessageSenders(WebServiceMessageSender...)}
+	 */
+	@Deprecated
 	public void setMessageSender(WebServiceMessageSender messageSender) {
 		Assert.state(!this.webServiceTemplateExplicitlySet,
 				() -> "'messageSender' must be specified on the provided: " + this.webServiceTemplate);
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsInboundGatewaySpec.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsInboundGatewaySpec.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.ws.dsl;
+
+import org.springframework.integration.dsl.MessagingGatewaySpec;
+import org.springframework.integration.ws.AbstractWebServiceInboundGateway;
+import org.springframework.integration.ws.SoapHeaderMapper;
+
+/**
+ * Base {@link MessagingGatewaySpec} for web services.
+ *
+ * @param <S> the target {@link BaseWsInboundGatewaySpec} implementation type.
+ * @param <E> the target {@link AbstractWebServiceInboundGateway} implementation type.
+ *
+ * @author Gary Russell
+ * @since 5.3
+ *
+ */
+public class BaseWsInboundGatewaySpec<
+		S extends BaseWsInboundGatewaySpec<S, E>, E extends AbstractWebServiceInboundGateway>
+	extends MessagingGatewaySpec<S, E> {
+
+	/**
+	 * Construct an instance for the gateway.
+	 * @param gateway the gateway.
+	 */
+	protected BaseWsInboundGatewaySpec(E gateway) {
+		super(gateway);
+	}
+
+	/**
+	 * Configure the header mapper.
+	 * @param headerMapper the mapper.
+	 * @return the spec.
+	 */
+	public S headerMapper(SoapHeaderMapper headerMapper) {
+		this.target.setHeaderMapper(headerMapper);
+		return _this();
+	}
+
+}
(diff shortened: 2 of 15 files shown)


spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
```diff
@@ -165,6 +165,12 @@ public void setFaultMessageResolver(FaultMessageResolver faultMessageResolver) {
...
 	}
 
+	/**
+	 * Specify the {@link WebServiceMessageSender} to use.
+	 * @param messageSender the sender.
+	 * @deprecated in favor of {@link #setMessageSenders(WebServiceMessageSender...)}
+	 */
+	@Deprecated
 	public void setMessageSender(WebServiceMessageSender messageSender) {
 		Assert.state(!this.webServiceTemplateExplicitlySet,
```
artembilan: Would you mind to change logic then in favor of calling a replacing API instead?
Then it is going to be much easier to support possible changes in the target logic afterward.

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * Specify whether the URI should be encoded after any <code>uriVariables</code>
+	 * are expanded and before sending the request. The default value is <code>true</code>.
+	 * @param encodeUri true if the URI should be encoded.
+	 * @return the spec.
+	 * @see org.springframework.web.util.UriComponentsBuilder
+	 */
+	public S encodeUri(boolean encodeUri) {
```
artembilan: Hm. Looks like we need to consider to support an `EncodingMode` instead. Similar to what we have now in HTTP and WebFlux.

Different issue though

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsInboundGatewaySpec.java
```diff
@@ -0,0 +1,75 @@
...
+	protected MarshallingWsInboundGatewaySpec(Marshaller marshaller) {
+		super(new MarshallingWebServiceInboundGateway(marshaller));
+	}
+
+	/**
+	 * Construct an instance with the provided arguments.
+	 * @param marshaller the marshaller.
+	 * @param unmarshaller the unmarshaller.
+	 */
+	protected MarshallingWsInboundGatewaySpec(Marshaller marshaller, Unmarshaller unmarshaller) {
```
artembilan: I wouldn't expose two of these ctors since it feels much natural just to call those configuration options...

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsOutboundGatewaySpec.java
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * Construct an instance with the provided arguments.
+	 * @param uri the URI.
+	 * @param marshaller the marshaller.
+	 * @param unmarshaller the unmarshaller.
+	 * @param messageFactory the message factory.
+	 */
+	protected MarshallingWsOutboundGatewaySpec(String uri,
```
artembilan: I think we need to consider to have all of these ctors spreaded as options and build the final `MarshallingWebServiceOutboundGateway` in the `doGet()`.
If possible, of course...

Otherwise it doesn't look too useful to have such a DSL spec based only on ctors

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/SimpleWsOutboundGatewaySpec.java
```diff
@@ -0,0 +1,109 @@
...
+		this(uri, sourceExtractor, null);
+	}
+
+	/**
+	 * Construct an instance with the provided arguments.
+	 * @param uri the URI.
+	 * @param sourceExtractor the source extractor.
+	 * @param messageFactory the message factory.
+	 */
+	protected SimpleWsOutboundGatewaySpec(String uri,
```
artembilan: Same here.

We may live with what you have so far, but I wonder if we really can make it more natural builder style way...

artembilan (changes requested):
Some review to think about.

Thanks

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
garyrussell: True (different issue).

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsOutboundGatewaySpec.java
garyrussell: Yeah - will rework - seemed a little clunky to me; I think a CTOR that takes a WST would be useful and return a different spec in that case (one that doesn't offer the properties that are disallowed with an explicit WST; there are many).

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
artembilan: https://github.com/spring-projects/spring-integration/issues/3180

artembilan:
@garyrussell ,

any updates? Or do you want me to take over this from you?

Thanks

garyrussell:
? 

I pushed rework a couple of days ago - didn't you get a notification?

artembilan:
>  garyrussell force-pushed the garyrussell:GH-2818 branch from 2a96e81 to d4bdbc0 

Seems for me when we "force-pushed", there is no notification. GH flaw? 😉 

OK. Sorry for noise: I should check PR before bothering you.
Will review soon.

Thanks

garyrussell:
I know force-push for a rebase doesn't notify, but I assumed force push with additional commits would - but I guess not.

In future, I'll try to remember to push the new commit; then force-push after rebasing.

spring-integration-ws/src/test/java/org/springframework/integration/ws/dsl/WsDslTests.java
```diff
@@ -0,0 +1,215 @@
...
+import org.springframework.ws.client.support.destination.DestinationProvider;
+import org.springframework.ws.client.support.interceptor.ClientInterceptor;
+import org.springframework.ws.transport.WebServiceMessageSender;
+
+/**
+ * @author Gary Russell
+ * @since 5.3
+ *
+ */
+public class WsDslTests {
```
artembilan: Is it expected that we don't have any integration test with this new DSL used from an `IntegrationFlow`?

artembilan (changes requested):
Some non-critical review.

I can address them all on merge (including an integration test)

Thanks

P.S. Although there has to be some Docs on the matter! 😄

spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
garyrussell: Sorry, I missed this; but, that said, I don't understand it.

spring-integration-ws/src/test/java/org/springframework/integration/ws/dsl/WsDslTests.java
garyrussell: I am not sure why an integration test is necessary since these tests verify the adapters are populated properly. But I can add one if you insist.

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsInboundGatewaySpec.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	protected E doGet() {
+		return assemble(create());
+	}
+
+	protected abstract E create();
+
+	protected E assemble(E gateway) {
```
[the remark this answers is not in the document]
garyrussell: Just for consistency with the outbound spec, in case common components are added later.

spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
artembilan: I mean this:
```
public void setMessageSender(WebServiceMessageSender messageSender) {
   setMessageSenders(messageSender);
}
```

So, we don't need to have that `Assert` in this method any more. The logic from a deprecated method is fully delegated to replacing API.

artembilan (changes requested):
OK. One more comment and code style. Plus docs and we are good to go.
Everything else is in agreement silently.

spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
garyrussell: Doh; ok.

garyrussell:
Force pushed with additional commit after rebase for doc conflicts.

PR #3186, opened 2020-02-19
Fixed broken links in Whats New doc

tedyoung:
Links used `.adoc/section` instead of `.adoc#section` causing links to not work.

Fixed English 'effects' to 'affects'.


artembilan:
Good catch! Thank you!

PR #3188, opened 2020-02-20
Timeout added to WebFluxMessageHandlerSpec.java & WebFluxRequestExecutingMessageHandler.java

### LINKED ISSUE #3183: Add hooks for reactive reply customizations: timeout(), retry(), tag() etc.
**Affects Version(s):** 5.2.3

---
## Enhancement
Consider adding Timeout to Webflux.outboundGateway in DSL. 

Mono & Flex Supports timeout. If the timeout is exposed in exposed in the Outbound Gateway, it will simplify the configuration of services in a fluent API style.

Happy to contribute if accepted for enhancement.

sirimamilla:
Timeout added to WebFluxMessageHandlerSpec.java & WebFluxRequestExecutingMessageHandler.java #3183 

Repushed from GH-3183


artembilan:
Thanks, but see the last sentence in the CONTRIBUTION guideline:

> Please, follow Chris Beams' recommendations in regards to the good commit message

Having so many contribution from you would benefit from a good commit message formatting.

😄

sirimamilla:
> > Please, follow Chris Beams' recommendations in regards to the good commit message
> 
> Having so many contribution from you would benefit from a good commit message formatting.

I will have a look at this and follow this. I have little time to contribute for this, so created a PR. Will look at your review comments tomorrow. Good Night.

Hope we can agree on this Feature Enhancement. I am working on adding transform method as well as you suggested, I would prefer to raise a separate PR for transform.

Also, I am thinking of right usecase for adding it in the test cases for the transform message.

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/dsl/WebFluxMessageHandlerSpec.java
```diff
@@ -144,6 +145,41 @@ public WebFluxMessageHandlerSpec publisherElementTypeExpression(Expression publi
...
 	}
 
+	/**
+	 * Specify the timeout value for receiving the response form the server.
+	 * If the response is not received with in timeout value, will result in Timeout Exception
+	 * @param timeoutInMillis the timeout in milliseconds
+	 * @return the spec
+	 * @since 5.3
+	 */
+	public WebFluxMessageHandlerSpec timeout(long timeoutInMillis) {
```
artembilan: That's bad name. There are many aspects in Spring Integration components and a lot of different timeout options.
Would be great to be more specific here. How about a `requestTimeout` ?

Also: I wouldn't expose this `long` option. It is not too hard to have that `Duration.ofMillis()` in the target application. 
Plus with Spring Boot configuration properties we can provide an option in the duration style and we'll get exactly `Duration` instance in the POJO

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
```diff
@@ -198,6 +202,26 @@ public void setPublisherElementTypeExpression(Expression publisherElementTypeExp
 		this.publisherElementTypeExpression = publisherElementTypeExpression;
 	}
 
+	/**
+	 * Specify the timeout value for receiving the response form the server.
+	 * If the response is not received with in timeout value, will result in Timeout Exception
+	 * @param timeout accepts {@link java.time.Duration}
+	 * @since 5.3
+	 */
+	public void setTimeout(Duration timeout) {
```
artembilan: I understand that this is simple one, but would be better to not have it since a `Function` can simple cover the static value as well.

So, please, remove

spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
```diff
@@ -198,6 +202,26 @@ public void setPublisherElementTypeExpression(Expression publisherElementTypeExp
...
+		this.timeoutFunction = m -> timeout;
+	}
+
+	/**
+	 * @param timeoutFunction accepts {@link Function}. The function accepts a Message as input and returns Duration.
+	 * Function is evaluated on each request basis and the {@link Duration} is applied to Mono
+	 * If the response is not received with in timeout value, will result in Timeout Exception
+	 * @since 5.3
+	 */
+	public void setTimeoutFunction(Function<Message<?>, Duration> timeoutFunction) {
```
artembilan: Well, we prefer an `Expression` variant instead of function. This way a timeout could be configured in the XML an easy way.
There is a `FunctionExpression` to cover this case. And exactly Java DSL should expose a function-based option. Of course, alongside with an expression. See `publisherElementTypeFunction` in the modified by you `WebFluxMessageHandlerSpec`

spring-integration-webflux/src/test/java/org/springframework/integration/webflux/dsl/WebFluxDslTests.java
```diff
@@ -145,6 +163,82 @@ public void setup() {
...
+				.build();
+
+		new DirectFieldAccessor(this.webFluxFlowWithTimeout)
+				.setPropertyValue("webClient", webClient);
+
+		QueueChannel replyChannel = new QueueChannel();
+
+		Message<String> testMessage =
+				MessageBuilder.withPayload("test")
+						.setReplyChannel(replyChannel)
```
artembilan: I think we shouldn't use a `setReplyChannel` for this kind of test. Since you use the same instance for both reply and error responses, I was confused first about exception assertion in the end. I was expected from "error channel", but I saw a `replyChannel`.  So, your test-case mislead me to investigate how do we process exceptions like that timeout 😢

src/reference/asciidoc/whats-new.adoc
```diff
@@ -68,3 +68,9 @@ See <<./amqp.adoc#amqp-inbound-channel-adapter,AMQP Inbound Channel Adapter>>
 The `encodeUri` property on the `AbstractHttpRequestExecutingMessageHandler` has been deprecated in favor of newly introduced `encodingMode`.
 See `DefaultUriBuilderFactory.EncodingMode` JavaDocs and <<./http.adoc#http-uri-encoding,Controlling URI Encoding>> for more information.
 This also affects `WebFluxRequestExecutingMessageHandler`, respective Java DSL and XML configuration.
+
+[[x5.3-WebFlux]]
+=== WebFlux Changes
+
+The `timeoutFunction` property is added to `WebFluxRequestExecutingMessageHandler`.
+Java DSL accessor which accepts long Timeout, Duration Timeout & Timeout Function is provided
```
artembilan: The good sentence ends with period. 
I don't think that word "timeout" deserves to be capitalized. 
I would even say if your don't talk about code, then even `function` word doesn't deserve to be capitalized. 

Would be great to have here a brief description for the reason of this new option. See something already existed in this doc.
Then we definitely need a link to the target docs.
And what is very important we need a description in the final docs - `webflux.adoc`.

artembilan (changes requested):
Please, consider to expose this new `request-timeout` attribute in the XML configuration. See `spring-integration-webflux.xsd` and respective `WebFluxOutboundChannelAdapterParser`.
Then you can find a `WebFluxOutboundChannelAdapterParserTests` to verify that new option is applied properly.

You can also add your name to all the affected classes.

Thank you!

sirimamilla:
> Please, consider to expose this new `request-timeout` attribute in the XML configuration. See `spring-integration-webflux.xsd` and respective `WebFluxOutboundChannelAdapterParser`.

request-timeout is already present in the spring-integration-webflux.xsd. So cannot be used for our purpose. Can you suggest a better name for the field.

For now I am using http-request-timeout.

artembilan:
I don't see changes and your comment in the PR, but if you talk about `request-timeout` on the `inbound-gateway`, that's fully different story and should not confuse with alike name on the `outbound-gateway`.
So, I still insists to call it a `requestTimeout` everywhere what is based on the `WebFluxRequestExecutingMessageHandler`.

sirimamilla:
Hi @artembilan ,

I was busy with office activities,  I am almost done with my changes. I will push changes today

artembilan:
@sirimamilla ,

you know after watching yesterday [Josh Long](https://joshlong.com/)'s [Reactive Revolution](https://www.youtube.com/watch?v=4-vEW8Ck_6g) presentation I have to take my decision back. There are indeed much more control and aspect operators in the Reactor `Flux` & `Mono`. 
That is your use-case to go with a simple `timeout`. Other people would like to do a `retry()`, `onErrorResume()` or even a Circuit Breaker: https://spring.io/projects/spring-cloud-circuitbreaker

So, for me it sounds more like a general solution for any `MessageHandler` which produces a reactive type as a result. Therefore I think about something like `ReactiveRequestHandlerAdvice` to apply for any endpoint with reactive capabilities.
That's definitely not only a WebFlux and not only a timeout concern to not have some general solution for anything reactive-based what can have problems connecting to remote part.

So, starting from here it is 👎 from me for this PR: a general solution with a `ReactiveRequestHandlerAdvice` and injected `BiFunction` into it for any possible returned reactive type customization will cover this `timeout` concern as well.

sirimamilla:
Hi @artembilan,
Your solution of providing an advice seems more generic. I am ok with that solution as it can solve more problems

artembilan:
Closed in favor of #3197 

@sirimamilla , you are feel free to provide any feedback over there.

PR #3190, opened 2020-02-20
GH-3182: Properly reset bean in the MockIntCtx

### LINKED ISSUE #3182: MockIntegrationContext.reset() does not behave as expected
**Affects Version(s):** 5.2.3.RELEASE using spring-boot 2.2.4.RELEASE

I am developing an application containing multiple IntegrationFlows passing messages to each other using DirectChannels.

I have created several small tests which are testing a single IntegrationFlow each. I have used MockIntegrationContext.substituteMessageHandlerFor() to capture the message once it is passed to the "output channel" of a flow.

However even though I'm calling MockIntegrationContext.reset() in an @AfterEach method, the substituted MockMessageHandlers seem to stick around until the ApplicationContext is torn down.

I have created a minimal gist demonstrating the behavior: https://gist.github.com/hgarus/0ec9714f1d381bd8404c08ddc88f55e3

The assertEquals in line 80 fails because the ArgumentCaptor receives the second message. Even though I called `reset()` beforehand.

I did some debugging and it seems the substitution only works, when the IntegrationConsumer is stopped before the replacement MessageHandler is assigned to the it - which is done in `substituteMessageHandlerFor()` (with the default `autoStartup = true`), but not in `reset()`.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3182

The `MockIntegrationContext.resetBeans()` doesn't really restore a state
of endpoint beans: the handle/source is applied only after start for that endpoint

* Stop endpoints in the `MockIntegrationContext.resetBeans()` is they are running currently
* Start them back only if their `autoStartup == true`
* Clear only those beans which names are provided for the `resetBeans()`

**Cherry-pick to 5.2.x**

### DIFF
--- a/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
+++ b/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.context.Lifecycle;
+import org.springframework.context.SmartLifecycle;
 import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.endpoint.IntegrationConsumer;
@@ -97,6 +98,11 @@ public class MockIntegrationContext implements BeanFactoryAware {
 				.forEach(e -> {
 					Object endpoint = this.beanFactory.getBean(e.getKey());
 					DirectFieldAccessor directFieldAccessor = new DirectFieldAccessor(endpoint);
+					SmartLifecycle lifecycle = null;
+					if (endpoint instanceof SmartLifecycle && ((SmartLifecycle) endpoint).isRunning()) {
+						lifecycle = (SmartLifecycle) endpoint;
+						lifecycle.stop();
+					}
 					if (endpoint instanceof SourcePollingChannelAdapter) {
 						directFieldAccessor.setPropertyValue("source", e.getValue());
 					}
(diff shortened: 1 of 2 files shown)


spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -108,9 +114,19 @@ else if (endpoint instanceof ReactiveStreamsConsumer) {
...
 						directFieldAccessor.setPropertyValue(HANDLER, e.getValue());
 					}
+					if (lifecycle != null && lifecycle.isAutoStartup()) {
+						lifecycle.start();
+					}
 				});
 
-		this.beans.clear();
+		if (!ObjectUtils.isEmpty(beanNames)) {
+			for (String name : beanNames) {
```
garyrussell: Don't we need a test for this "partial" reset use case?

garyrussell (commented):
Just one question/comment.

garyrussell:
Cherry-picked as 4e4e8db17410a7127e1416517a0ff25919f13211

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -108,9 +114,19 @@ else if (endpoint instanceof ReactiveStreamsConsumer) {
 					else if (endpoint instanceof IntegrationConsumer) {
 						directFieldAccessor.setPropertyValue(HANDLER, e.getValue());
 					}
+					if (lifecycle != null && lifecycle.isAutoStartup()) {
```
hgarus: Doesn't this have potential to break existing tests?

Maybe this just matters for constructed examples. But I could imagine something like this:
```
@BeforeClass
void init() {
  nonAutostartupHandler.start();
}

@After 
void cleanup() {
  mockIntegrationContext.reset()
}

@Test
void t1() {
  // do something with handler
}

@Test 
void t2() {
  // expect handler to be still running
}
```

Admittedly the `@BeforeClass` is somewhat artificial. But I could see a  handler being manually started as part of context initialization.

I'm sorry for not contributing a fix myself, I had a pretty busy week at work.

artembilan: `nonAutostartupHandler.start();` - you definitely start it manually. The framework can't assume when and how you have started it or going to.
Only the way for the framework to start it when you have that `autoStartup = true`.

I admit that it may break some peculiar test, but at the same time the test is wrong because it has been relied on bug which we are fixing here.

On the other hand I don't worry too much about breaking changes for tests. That's not that part of project which may make your customer unhappy.

PR #3197, opened 2020-02-27
GH-3183: Add ReactiveRequestHandlerAdvice

### LINKED ISSUE #3183: Add hooks for reactive reply customizations: timeout(), retry(), tag() etc.

artembilan:
Fixes #3183

* Introduce a `ReactiveRequestHandlerAdvice` with a `BiFunction<Message<?>, Mono<?>, Publisher<?>>`
logic to apply a `Mono.transform()` operator for a returned from the handler `Mono` reply
* Fix `WebFluxRequestExecutingMessageHandler` to return a `Mono.then()` instead of an explicit subscription -
it happens downstream anyway during reply producing with a proper error handling, too
* Demonstrate `ReactiveRequestHandlerAdvice` in the `RSocketDslTests` - without `retry()` it fails
* Add `ConsumerEndpointSpec.customizeMonoReply()` for convenience
* Document `ReactiveRequestHandlerAdvice` feature

### DIFF
--- a/src/reference/asciidoc/handler-advice.adoc
+++ b/src/reference/asciidoc/handler-advice.adoc
@@ -56,6 +56,7 @@ In addition to providing the general mechanism to apply AOP advice classes, Spri
 * `ExpressionEvaluatingRequestHandlerAdvice` (described in <<expression-advice>>)
 * `RateLimiterRequestHandlerAdvice` (described in <<rate-limiter-advice>>)
 * `CacheRequestHandlerAdvice` (described in <<cache-advice>>)
+* `ReactiveRequestHandlerAdvice` (described in <<reactive-advice>>)
 
 [[retry-advice]]
 ===== Retry Advice
@@ -514,6 +515,7 @@ This configuration functionality is similar to Spring Framework's `@CacheConfig`
 If a `CacheManager` is not provided, a single bean is resolved by default from the `BeanFactory` in the `CacheAspectSupport`.
 
 The following example configures two advices with different set of caching operations:
+
 ====
 [source, java]
 ----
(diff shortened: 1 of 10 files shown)


artembilan:
If the solution is good, I'll go ahead with WebFlux reworking.

This feature opens for us a possible to apply a `TransactionInterceptor` with a `ReactiveTransactionManager` onto our reactive-replying components. 

I may admit that this could be kinda breaking change though since that modified `RSocketOutboundGateway` is not an `AbstractReplyProducingMessageHandler` anymore...

sirimamilla:
HI @artembilan,

This will add a lot of value. As it gives a lot of scope for customization and unleashes the power of Reactive framework.

garyrussell:
>If the solution is good, I'll go ahead with WebFlux reworking.

LGTM

artembilan:
I'll come back to this after merging #3201 .
They are kinda depend and complement each other.

artembilan:
I have changed my mind: now `ReactiveRequestHandlerAdvice` checks a type of the return from the `invocation.proceed()`.
So, we don't need extra abstraction to enforce reactive behavior for target `MessageHandler`.

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -549,6 +551,26 @@ public Message<?> service(Message<?> message) {
 ----
 ====
 
+[[reactive-advice]]
+==== Reactive Advice
+
+Starting with version 5.3, a `ReactiveRequestHandlerAdvice` can be used for request message handlers producing a `Mono` replies.
+A `BiFunction<Message<?>, Mono<?>, Publisher<?>>` has to be provided for this advice and it is called from the `Mono.transform()` operator on a reply produced by the intercepted `handleRequestMessage()` method implementation.
+Typically such a `Mono` customization is necessary when we would like to control network fluctuations via `timeout()`, `retry()` and similar support operators.
+For example when we can an HTTP request over WebFlux client, we could use bellow configuration to not wait for response more than 5 seconds:
```
garyrussell:
```suggestion
For example when we can an HTTP request over WebFlux client, we could use below configuration to not wait for response more than 5 seconds:
```

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -549,6 +551,26 @@ public Message<?> service(Message<?> message) {
...
+
+====
+[source, java]
+----
+.handle(WebFlux.outboundGateway("https://somehost/"),
+                       e -> e.customizeMonoReply((message, mono) -> mono.timeout(Duration.ofSeconds(5))));
+----
+====
+
+The `message` argument is a request message for such a message handler and can be used to determine a request-scope attributes.
```
garyrussell:
```suggestion
The `message` argument is the request message for the message handler and can be used to determine request-scope attributes.
```

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -549,6 +551,26 @@ public Message<?> service(Message<?> message) {
...
+The `mono` argument is a result of this message handler `handleRequestMessage()` method implementation.
```
garyrussell:
```suggestion
The `mono` argument is the result of this message handler's `handleRequestMessage()` method implementation.
```

garyrussell (changes requested):

garyrussell (approved):

PR #3201, opened 2020-02-28
Add Reactive TX Manager support

artembilan:
* Change `TransactionInterceptorBuilder` and `TransactionHandleMessageAdvice` to reply
on a generic `TransactionManager`, so we can configure reactive one as well
* Change a `<transactional>` XML element to support a generic `TransactionManager` reference,
so we can configure reactive one as well
* Support `adviceChain` configuration for the `ReactiveMessageHandler` in the `ConsumerEndpointFactoryBean`
* Attemp to apply reactive transaction for the Reactive MongoDB channel adapter

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
@@ -203,10 +203,9 @@ public class ConsumerEndpointFactoryBean
 			this.handler = adviceChain(this.handler);
 		}
 		else if (!CollectionUtils.isEmpty(this.adviceChain)) {
-			ReactiveMessageHandler reactiveMessageHandler =
-					((ReactiveMessageHandlerAdapter) this.handler).getDelegate();
-			reactiveMessageHandler = adviceChain(reactiveMessageHandler);
-			this.handler = new ReactiveMessageHandlerAdapter(reactiveMessageHandler);
+			this.handler =
+					new ReactiveMessageHandlerAdapter(
+							adviceChain(((ReactiveMessageHandlerAdapter) this.handler).getDelegate()));
 		}
 		if (this.channelResolver == null) {
 			this.channelResolver = ChannelResolverUtils.getChannelResolver(this.beanFactory);
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
@@ -200,10 +200,13 @@ public class ConsumerEndpointFactoryBean
 		}
 
 		if (!(this.handler instanceof ReactiveMessageHandlerAdapter)) {
-			adviceChain();
+			this.handler = adviceChain(this.handler);
 		}
 		else if (!CollectionUtils.isEmpty(this.adviceChain)) {
-			LOGGER.warn("the advice chain cannot be applied to a 'ReactiveMessageHandler'");
+			ReactiveMessageHandler reactiveMessageHandler =
+					((ReactiveMessageHandlerAdapter) this.handler).getDelegate();
+			reactiveMessageHandler = adviceChain(reactiveMessageHandler);
+			this.handler = new ReactiveMessageHandlerAdapter(reactiveMessageHandler);
 		}
 		if (this.channelResolver == null) {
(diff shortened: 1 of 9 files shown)


artembilan:
The feature is ready, but I'd like to have a real test against reactive transaction.
Need consultation with Spring Data team to understand what else has to be provided for test configuration.

Thanks

artembilan:
I decided to go ahead without testing MongoDB reactive transaction. The support is too involved: we need the latest MongoDB server with replica enabled. Plus we need include the replica key from the server into a connection string.
Essentially it is not this project responsibility to have such a hard integration testing.

spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
```diff
@@ -200,10 +200,13 @@ public void afterPropertiesSet() {
...
 		if (!(this.handler instanceof ReactiveMessageHandlerAdapter)) {
-			adviceChain();
+			this.handler = adviceChain(this.handler);
 		}
 		else if (!CollectionUtils.isEmpty(this.adviceChain)) {
-			LOGGER.warn("the advice chain cannot be applied to a 'ReactiveMessageHandler'");
+			ReactiveMessageHandler reactiveMessageHandler =
+					((ReactiveMessageHandlerAdapter) this.handler).getDelegate();
+			reactiveMessageHandler = adviceChain(reactiveMessageHandler);
+			this.handler = new ReactiveMessageHandlerAdapter(reactiveMessageHandler);
```
garyrussell:
```suggestion
        this.handler = new ReactiveMessageHandlerAdapter(adviceChain(((ReactiveMessageHandlerAdapter) this.handler).getDelegate())));
```

spring-integration-core/src/main/java/org/springframework/integration/transaction/TransactionHandleMessageAdvice.java
```diff
@@ -46,12 +46,16 @@
 	public TransactionHandleMessageAdvice() {
 	}
 
-	public TransactionHandleMessageAdvice(PlatformTransactionManager ptm, Properties attributes) {
-		super(ptm, attributes);
+	public TransactionHandleMessageAdvice(TransactionManager transactionManager, Properties transactionAttributes) {
+		setTransactionManager(transactionManager);
+		setTransactionAttributes(transactionAttributes);
 	}
```
garyrussell: Need `final` overrides of these (or a new method taking both) to call from a CTOR - even though Spring is in violation of that rule in the super class.

garyrussell:
Sorry hit single comment instead of start review; still reviewing.

spring-integration-core/src/main/java/org/springframework/integration/config/ConsumerEndpointFactoryBean.java
garyrussell:
```suggestion
        this.handler = new ReactiveMessageHandlerAdapter(adviceChain(((ReactiveMessageHandlerAdapter) this.handler).getDelegate())));
```

src/reference/asciidoc/whats-new.adoc
```diff
@@ -57,6 +57,10 @@ A new `publishSubscribeChannel()` operator, based on the `BroadcastCapableChanne
 This fluent API has its advantage when we configure sub-flows as pub-sub subscribers for broker-backed channels like `SubscribableJmsChannel`, `SubscribableRedisChannel` etc.
 See <<./dsl.adoc#java-dsl-subflows,Sub-flows support>> for more information.
 
+The transactional support in Spring Integration now also includes options to configure a `ReactiveTransactionManager` if a `MessageSource` or `MessageHandler` implementation produces a reactive type for payload to send.
```
garyrussell:
```suggestion
Transactional support in Spring Integration now also includes options to configure a `ReactiveTransactionManager` if a `MessageSource` or `MessageHandler` implementation produces a reactive type for payload to send.
```

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transaction/TransactionHandleMessageAdvice.java
artembilan: Well, I think we can live with this for a while.
See here: https://github.com/spring-projects/spring-framework/issues/24612
When that is fixed, we can revert this change to rely on super ctors.

garyrussell (approved):

PR #3203, opened 2020-02-29
[GH-3202] Fix locale-dependent tests

### LINKED ISSUE #3202: Failing tests due to translated errors
Observed with commit id: e2b63bed7c1dbd0d9b0c08482baa70a67eaa4e08

14 tests fail for `gradlew build` on my machine (`org.springframework.integration.config.xml.ChainElementsFailureTests` and `org.springframework.integration.file.config.ChainElementsTests`) as they assert translated error messages:

```
org.opentest4j.AssertionFailedError: 
Expecting:
 <"cvc-complex-type.3.2.2: Attribut 'input-channel' darf nicht in Element 'int:aggregator' vorkommen.">
to be equal to:
 <"cvc-complex-type.3.2.2: Attribute 'input-channel' is not allowed to appear in element 'int:aggregator'.">
but was not.
...
```

The tests should set the default locale of the VM prior to execution.

grubeninspekteur:
Fixes https://github.com/spring-projects/spring-integration/issues/3202

Sets the locale for the problematic tests to en-US.

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/config/xml/ChainElementsFailureTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/config/xml/ChainElementsFailureTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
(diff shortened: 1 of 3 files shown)


artembilan:
Good catch! Thank you! Looking forward for more!

Issue #3204, opened 2020-02-29
Add .interceptor() to flow DSL


grubeninspekteur:
## Enhancement
Consider introducing a `.interceptor(ChannelInterceptor...)` method to the flow builder for adding interceptors to the current (possibly implicit) channel. At the moment the best way to do this is

```java
.channel(
    MessageChannels
        .directChannel()
        .interceptor(...)
  )
```

One could argue that this is a clear separation of concerns, however, `wireTap` has also been lifted up to the flow builder.

Example usage: Adding values to the Mapped Diagnostic Context (MDC) for incoming messages and cleaning up in `ChannelInterceptor#afterSendCompletion`.

PR #3205, opened 2020-02-29
GH-3204: Add channel interceptor support to flow definition

grubeninspekteur:
Adds an `intercept()` builder method to register one or more
channel interceptors at the current flow position.

Issue #3204

* add `intercept(ChannelInterceptor...)` to `BaseIntegrationFlowDefinition`
* refactor to reuse interceptable channel creation from `wireTap`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
@@ -90,6 +90,7 @@ import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.support.ChannelInterceptor;
 import org.springframework.messaging.support.InterceptableChannel;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
@@ -109,6 +110,7 @@ import reactor.util.function.Tuple2;
  * @author Artem Bilan
  * @author Gary Russell
  * @author Gabriele Del Prete
+ * @author Tim Feuerbach
  *
  * @since 5.2.1
  *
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
```diff
@@ -179,6 +180,21 @@ protected MessageChannel getCurrentMessageChannel() {
 		return this.currentMessageChannel;
 	}
 
+	/**
+	 * Returns the current channel if it is an {@link InterceptableChannel}, otherwise registers a new implicit
+	 * {@link DirectChannel} in the flow and returns that one.
+	 * @return the current channel after the operation
+	 */
+	protected MessageChannel makeCurrentMessageChannelInterceptable() {
```
artembilan: I think we need to return exactly an `InterceptableChannel` from here.
And let's call it like `currentInterceptableChannel()`! A "make" is confusing a little bit...

spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
```diff
@@ -2829,6 +2840,26 @@ public B trigger(MessageTriggerAction triggerAction,
 		return handle(new ServiceActivatingHandler(triggerAction, "trigger"), endpointConfigurer);
 	}
 
+	/**
+	 * Adds one or more {@link ChannelInterceptor} implementations
```
artembilan: Method JavaDocs must be imperative style: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
```diff
@@ -2829,6 +2840,26 @@ public B trigger(MessageTriggerAction triggerAction,
...
+	 * to the current {@link #currentMessageChannel}, in the given order, after any interceptors already registered.
+	 * @param interceptorArray one or more {@link ChannelInterceptor}s.
+	 * @return the current {@link BaseIntegrationFlowDefinition}.
+	 * @throws IllegalArgumentException if no interceptors or at least one null value are provided
+	 * @since 5.3
+	 */
+	public B interceptor(ChannelInterceptor... interceptorArray) {
```
artembilan: Well, our style to make methods on this builder as verbs, in other words commands: do this or that.
Look at this as EIP-operators!

Therefore if we add this feature here, it has to be named as `intercept()`.

artembilan (changes requested):
Please, add your name to the `@author` list. Document a new feature in the `dsl.adoc`. And ink to that from the `whats-new.adoc`.

Thank you for contribution so far!

grubeninspekteur:
Thank you for your review! I made the requested changes and also relaxed the `Assert.notEmpty()` requirement for the interceptor array to `Assert.notNull()`.

artembilan:
@grubeninspekteur ,

PR #3208, opened 2020-03-04
GH-3207: RSocket inbound: decode each flux item

### LINKED ISSUE #3207: Rsocket InboundGateway message handling should be consistent with regular @Controller/@MessageMapping
The original question is posted on StackOverflow, check [here](https://stackoverflow.com/questions/60464223/spring-integraton-rsocket-and-spring-rsocket-interaction-issues).

My code sample is [here](https://github.com/hantsy/rsocket-sample/tree/master/integration).

Currently, I have to add a new line delimiter to every item in the `Flux` in the client and let server side know it should be a `Flux`.

```java
Flux.just("a\n", "b\n", "c\n", "d\n")
```

Currently, I have not tried other interaction modes Like regular `@Controller/@RequestMapping`, RSocketInboundGateway should accept the following payloads.

* (empty)
* T
* Mono&lt;T>
* Flux&lt;T>

The following explanation is quoted from the answer from @artembilan:

>Currently the server side of Spring Integration doesn't treat an incoming Flux as a stream of independent payloads. So, we try to connect all of them into a single value. The new line delimiter is an indicator that we expect independent values. Spring Messaging on its side does exactly opposite: it checks for multi-value expected type and decode every element in the incoming Flux in its map() instead of an attempt for the whole Publisher decoding.

>It's going to be kinda breaking change, but possibly need to consider to fix RSocketInboundGateway logic to be consistent with regular @MessageMapping for RSocket support.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3207

Previously an incoming RSocket Publisher has been decoded as a single unit
leading to extra work on the client side, e.g. a delimiter has to be provided
to treat each payload item as independent

* To have a consistency with Spring Messaging and its `PayloadMethodArgumentResolver`
change an `RSocketInboundGateway` to process inbound payloads as `Flux` and decode
each item independently.
* Change `RSocketDslTests` to remove delimiters and make it consistent with the regular
`RSocketRequester` client

### DIFF
--- a/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/config/RSocketInboundGatewayParser.java
+++ b/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/config/RSocketInboundGatewayParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 the original author or authors.
+ * Copyright 2019-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,7 +37,6 @@ public class RSocketInboundGatewayParser extends AbstractInboundGatewayParser {
 
 	private static final List<String> NON_ELIGIBLE_ATTRIBUTES =
 			Arrays.asList("path",
-					"interaction-models",
 					"rsocket-strategies",
 					"rsocket-connector",
 					"request-element-type");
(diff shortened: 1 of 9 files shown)


garyrussell (changes requested):
Needs a what's new doc entry.

Doesn't this need an option to revert to previous behavior?

garyrussell (approved):

PR #3212, opened 2020-03-10
GH-3211: Add DefSftpSF.setKnownHosts(Resource)

### LINKED ISSUE #3211: DefaultSftpSessionFactory file resolving inconsistency
Using Spring Boot 2.2.4 with Spring Integration SFTP 5.2.3

`org.springframework.integration.sftp.session.DefaultSftpSessionFactory`

Actual Behavior:
Property `privateKey` expects the prefix `file:` since it is resolved by Spring.
Property `knownHosts` does not work with `file:` prefix since it is passed directly to JCraft.

Expected Behavior:
Both properties `privateKey` and `knownHosts` work with `file:` prefix.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3211

* Add `DefaultSftpSessionFactory.setKnownHosts(Resource)` to
allow to configure externally any resource for file with known_hosts content
* Deprecate an existing method in favor of new one
* The new method makes it aligned with the `setPrivateKey(Resource)`
* Fix tests do not use a deprecated method any more

**Cherry-pick to 5.2.x**

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
@@ -193,11 +193,11 @@ public class DefaultSftpSessionFactory implements SessionFactory<LsEntry>, Share
 	 * false (default).</b>
 	 * @param knownHosts The known hosts.
 	 * @see JSch#setKnownHosts(String)
-	 * @deprecated since 5.2.5 in favor of {@link #setKnownHosts(Resource)}
+	 * @deprecated since 5.2.5 in favor of {@link #setKnownHostsResource(Resource)}
 	 */
 	@Deprecated
 	public void setKnownHosts(String knownHosts) {
-		setKnownHosts(new FileSystemResource(knownHosts));
+		setKnownHostsResource(new FileSystemResource(knownHosts));
 	}
 
 	/**
@@ -207,7 +207,7 @@ public class DefaultSftpSessionFactory implements SessionFactory<LsEntry>, Share
 	 * @see JSch#setKnownHosts(InputStream)
 	 * @since 5.2.5
 	 */
-	public void setKnownHosts(Resource knownHosts) {
+	public void setKnownHostsResource(Resource knownHosts) {
 		this.knownHosts = knownHosts;
 	}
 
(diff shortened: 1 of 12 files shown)


spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
```diff
@@ -191,8 +193,21 @@ public void setPassword(String password) {
...
+	}
+
+	/**
+	 * Specifies the filename that will be used for a host key repository.
+	 * The file has the same format as OpenSSH's known_hosts file.
+	 * @param knownHosts the resource for known hosts.
+	 * @see JSch#setKnownHosts(InputStream)
+	 * @since 5.2.5
+	 */
+	public void setKnownHosts(Resource knownHosts) {
```
garyrussell: Won't this cause a problem with `<bean/>` config? (ambiguous setter) `setKnownHostsResource()` ?

garyrussell (changes requested):

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
artembilan: Looks like Juergen has fixed something in SF. See Travis 😄

garyrussell: If I recall correctly, failures were random; also it looks like you changed all the tests to return a `Resource` using SpEL.

Perhaps the problem only occurs when a PropertyEditor is needed (although I remember most of the problems were with String Vs. Expression .

I will defer to you, though, if you think it's safe.

artembilan: Well, I can rename it to `setKnownHostsResource` to have it really very safe...

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
```diff
@@ -191,8 +193,21 @@ public void setPassword(String password) {
 	 * false (default).</b>
 	 * @param knownHosts The known hosts.
 	 * @see JSch#setKnownHosts(String)
+	 * @deprecated since 5.2.5 in favor of {@link #setKnownHostsResource(Resource)}
 	 */
+	@Deprecated
```
garyrussell: Should we really deprecate? This is easier if you are using Java Config and want to point to a simple file name.

Not related to this PR but perhaps we should consider adding a `SftpSessionFactorySpec` ?

artembilan: Then we need also to add a string-based for `privateKey`.

More over it is better to resolve any resources via `@Value Resource` then inject.

garyrussell (approved):

PR #3214, opened 2020-03-10
3213: Provide a mechanism to customize the Candidate Id…

berngp:
This commits adds a mechanism to customize the identifier, and therefore
data, that is used for a _Leadership Candidate_. This can be used to
facilitate triage since a custom provider can have additional information, e.g. identifier of the node/instance of the application.

image

### DIFF
--- a/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/config/CandidateIdProvider.java
+++ b/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/config/CandidateIdProvider.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.zookeeper.config;
+
+/**
+ * Provider that can be used to customize the identifier used by the
+ * {@link org.springframework.integration.leader.Candidate Leadership Candidate}.
+ * This can be used to customize the data that is stored in Zookeeper for the specific zknode.
+ *
+ * @since 5.2.5
+ */
+@FunctionalInterface
+public interface CandidateIdProvider {
+	String generate();
+}
--- a/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/config/LeaderInitiatorFactoryBean.java
+++ b/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/config/LeaderInitiatorFactoryBean.java
@@ -38,13 +38,12 @@ import org.springframework.util.Assert;
  *
  * @author Gary Russell
  * @author Artem Bilan
- *
  * @since 4.2
  */
 public class LeaderInitiatorFactoryBean
 		implements FactoryBean<LeaderInitiator>, SmartLifecycle, InitializingBean, ApplicationEventPublisherAware {
 
-	private static final CandidateIdProvider DEFAULT_CANDIDATE_ID_PROVIDER = () -> UUID.randomUUID().toString();
+	private static final CandidateProvider DEFAULT_CANDIDATE_ID_PROVIDER = (role) -> new DefaultCandidate(UUID.randomUUID().toString(), role);
 
 	private CuratorFramework client;
 
@@ -62,15 +61,15 @@ public class LeaderInitiatorFactoryBean
 
 	private LeaderEventPublisher leaderEventPublisher;
 
-	private final CandidateIdProvider candidateIdProvider;
+	private final CandidateProvider candidateProvider;
 
 	public LeaderInitiatorFactoryBean() {
-		this.candidateIdProvider = DEFAULT_CANDIDATE_ID_PROVIDER;
+		this.candidateProvider = DEFAULT_CANDIDATE_ID_PROVIDER;
 	}
 
-	public LeaderInitiatorFactoryBean(@NonNull CandidateIdProvider candidateIdProvider) {
-		Assert.notNull(candidateIdProvider, "The 'candidateIdProvider' must not be null.");
-		this.candidateIdProvider = candidateIdProvider;
+	public LeaderInitiatorFactoryBean(@NonNull CandidateProvider candidateProvider) {
+		Assert.notNull(candidateProvider, "The 'candidateProvider' must not be null.");
+		this.candidateProvider = candidateProvider;
 	}
 
 	public LeaderInitiatorFactoryBean setClient(CuratorFramework client) {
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@berngp Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3214)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3214) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@berngp Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration&pullRequestId=3214)!

garyrussell (changes requested):
Looks reasonable to me; a couple of comments.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/config/CandidateIdProvider.java
```diff
@@ -0,0 +1,13 @@
+package org.springframework.integration.zookeeper.config;
+
+/**
+ * Provider that can be used to customize the identifier used by the
+ * {@link org.springframework.integration.leader.Candidate Leadership Candidate}.
+ * This can be used to customize the data that is stored in Zookeeper for the specific zknode.
+ *
+ * @since ?
```
garyrussell: 5.3, but this is benign, so cherry-picking to 5.2.5 is probably ok.

artembilan:
We need to be careful with such a change. See `Candidate` API:

```
	 * @return a unique ID for this candidate; no other candidate for
	 *         leader election should return the same id
	 */
	String getId();
```

We need to document it respectively.
Also I don’t see a reason in extra abstraction for that Id: you can’t make me to return the same value all the time. So, if we still go an external configuration option, it has to be a setter for a string value. Seems for me we have a similar solution in the `DefaultLockRepository‘ in JDBC module.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/config/LeaderInitiatorFactoryBean.java
```diff
@@ -72,10 +84,24 @@ public LeaderInitiatorFactoryBean setPath(String path) {
 	}
 
 	public LeaderInitiatorFactoryBean setRole(String role) {
-		this.candidate = new DefaultCandidate(UUID.randomUUID().toString(), role);
+		this.candidate = newCandidate(role);
+		Assert.state(this.candidate != null,
+				String.format("The %s requires a candidate; newCandidate(role) returned none.", getClass().getName()));
 		return this;
```
garyrussell: You should use the override method that takes a supplied to avoid the `String.format()` call unless the assertion failed... `() -> String.format(...)`.

garyrussell:
Agreed; we either need to document carefully that it must be unique (or add a UUID to whatever the user code returns).

artembilan:
> or add a UUID to whatever the user code returns

I guess the main point of this issue is exactly to avoid something random, unreadable. Therefore a request to let end-user to provide meaningful identifier.

But again: I'm against some extra `Provider` abstraction.
The simple string injection should be fully enough. 
JavaDocs and docs for this option should state clearly that it will become end-user responsibility to ensure that provided identifier is really unique per cluster.

garyrussell:
>The simple string injection should be fully enough.

Good point; that you meant that wasn't clear from your previous comment; sorry.

artembilan:
> wasn't clear from your previous comment

Sorry, commented at night from the phone. 😄 

I saw a lot of activities in this PR, so decide to stop it while it hasn't gone too far in the polishing whatever could not make it to the master.

artembilan:
It turns out that we might even don't need any changes.
The `org.springframework.integration.zookeeper.leader.LeaderInitiator` has a ctor like:
```
public LeaderInitiator(CuratorFramework client, Candidate candidate) {
```

so, we can provide a `DefaultCandidate` with respective external `id`:
```
public DefaultCandidate(String id, String role) {
```

It might be the case that `LeaderInitiatorFactoryBean` needs some changes to accept externally configured `candidate`.
But that's all what we need to do. Even we don't need docs because we have that `DefaultCandidate` ctor already for while.

I would really consider to close this PR in favor of something much simpler and portable to the previous version: we definitely need to release `5.3 M3` today...

Thanks for understanding

berngp:
Thanks for the feedback @artembilan. I don't think there is any underlying intention to have this changes as part of `5.3 M3`. I can easily refactor this PR to address your observations. I would rather favor a design we are comfortable with. In general what this PR is attempts is to help users associate the zknode with the specific app/node instance that owns it.

Some thoughts...
1. The `LeaderInitiatorFactoryBean` implicitly creates the client once the `role` is set. If we provide an externally built `client` then the _setting for the role_ will not take effect unless we establish that the client will have to be mutable on its _role field_.  Wouldn't this add yet more complexity?
2. I would agree that the method `protected Candidate newCandidate(String role) ` is not needed. The initial intention was to provide an _escape_ hatch to a different implementation of a `Candidate` that can hold additional information. Since there is no easy way to serialize such information as part of the as part of Curator's `LeaderSelector` this is not very valuable, at least yet.
3. Instead of providing a `CandidateIdProvider` as a `LeaderInitiatorFactoryBean` ctor arg we could define a `DefaultCandidateBuilder` or a `DefaultCandidateProvider` where we state the characteristics of uniqueness for the id.

*Note* I am not sure we should change how the `LeaderInitiatorFactoryBean` interacts with the `LeaderInitiator`. As of now it leverages the already built `candidate` as follows.

```
this.leaderInitiator = new LeaderInitiator(this.client, this.candidate, this.path);
```

artembilan:
@berngp ,

I think to close all the gaps we should go the way that `LeaderInitiatorFactoryBean` should have a `setCandidate(Candidate)` and that `setRole()` has to be deprecated.

I don't see a reason in any new abstractions just to let end-user to configure some `id`. The `DefaultCandidate` with its ctor is fully enough for us.

berngp:
@artembilan I am not sure I feel comfortable on deprecating `setRole` since I think this will affect folks that depend on this for XML configuration or the `IntegrationFlow` DSL. I updated the PR with a `CandidateProvider` as the only interface/abstraction added by this PR. If this is still not sufficient feel free to close the PR. I can always address this on my side.

artembilan:
One more time:

1. If we don't take into account `LeaderInitiatorFactoryBean`, we are good with existing API on the `LeaderInitiator` and `DefaultCandidate`. They both allow you to achieve an original requirement without any changes to the framework.
2. An abstraction like you try to push with the `CandidateProvider` doesn't make sense because we already have `Candidate` and its `DefaultCandidate` implementation with all the requirements you have stated in the issue. It is really much easier for end-users to rely on `new` operator for self-containing entities, rather than try to figure out what is that `provider` abstraction is for. Plus "provider" as a leverage does not make sense here because we are going to call it only once in the ctor. You are free to have such an abstraction in your project and call it for that `LeaderInitiator` ctor, but it fully doesn't bring value for the framework.
3. There is nothing wrong or bad with an XML to provide an extra `<bean>` for the `candidate` to cover an improved behavior around that deprecated `setRole()`.
4. I don't see any value in your concern in regards Java DSL. Fully non-relevant argument... Whatever choice we do for Java configuration, it is very easy to accomplish with IDE code auto-completion.

Thanks

berngp:
I'll close this PR and let you folks decide what is better for the framework. If you provide a `setCandidate(Candidate)` and deprecate `setRole()` that will work for me.

PR #3218, opened 2020-03-17
Add Kotlin wrappers for router specs

artembilan:
To avoid casting and extra logic logic in the end-user code,
it is better to provide Kotlin-specific API to let end-users to
do whatever is really dictated by API and don't think about specific types to cast

### DIFF
--- a/spring-integration-core/src/main/kotlin/org/springframework/integration/dsl/KotlinIntegrationFlowDefinition.kt
+++ b/spring-integration-core/src/main/kotlin/org/springframework/integration/dsl/KotlinIntegrationFlowDefinition.kt
@@ -985,7 +985,7 @@ class KotlinIntegrationFlowDefinition(@PublishedApi internal val delegate: Integ
 	}
 
 	/**
-	 * Populate a [KotlinRecipientListRouterSpec] to the current integration flow position
+	 * Populate a [ScatterGatherHandler] to the current integration flow position
 	 * based on the provided [KotlinRecipientListRouterSpec] for scattering function
 	 * and [AggregatorSpec] for gathering function.
 	 */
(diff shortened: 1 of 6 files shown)


spring-integration-core/src/main/kotlin/org/springframework/integration/dsl/KotlinIntegrationFlowDefinition.kt
```diff
@@ -971,31 +977,33 @@ class KotlinIntegrationFlowDefinition(@PublishedApi internal val delegate: Integ
...
-	fun scatterGather(scatterer: RecipientListRouterSpec.() -> Unit) {
-		this.delegate.scatterGather(scatterer)
+	fun scatterGather(scatterer: KotlinRecipientListRouterSpec.() -> Unit) {
+		this.delegate.scatterGather(Consumer { scatterer(KotlinRecipientListRouterSpec(it)) })
 	}
 
 	/**
-	 * Populate a [ScatterGatherHandler] to the current integration flow position
-	 * based on the provided [RecipientListRouterSpec] for scattering function
+	 * Populate a [KotlinRecipientListRouterSpec] to the current integration flow position
```
garyrussell: This looks suspicious. The function is `scatterGather` and the variant below does not have this change.

garyrussell (commented):
LGTM; just one comment/question.

spring-integration-core/src/main/kotlin/org/springframework/integration/dsl/KotlinIntegrationFlowDefinition.kt
artembilan: Doh...

Will fix soon

PR #3220, opened 2020-03-17
GH-3215: MQTT Event for failed connection outbound

### LINKED ISSUE #3215: Publishing MqttConnectionFailedEvent on reconnection in MqttPahoMessageHandler
We use MQTT server with Keycloak requiring us to generate a token with Keycloak before we start a connection between our spring boot application and MQTT. On occasions for whatever reason that connection gets dropped, current behaviour means it will try to reconnect when the next message is published. When this occurs the existing connection details are used which potentially contains an expired token.

There doesn't seem to be to way that we can alter the connection options or at least have some flexibility when the connection is lost. My proposal includes something that has already been implemented in the MqttPahoMessageDrivenChannelAdapter class where a MqttConnectionFailedEvent is published.  The same would be published on the producer side in the MqttPahoMessageHandler class enabling us to have some sort of flexibility in dealing with lost connection issues.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3215

**cherry-pick to 5.2.x**


artembilan (changes requested):
Can we document this feature?
It feels like a simple note would be really use-full for use-cases like in the issue.

Thanks

artembilan:
... and cherry-picked to `5.2.x`

PR #3221, opened 2020-03-17
GH-3181: MQTT: Support MANUAL Acks

### LINKED ISSUE #3181: MQTT should support `setManualAcks`
**Affects Version(s):** all

## Enhancement

MQTT is designed to provide reliable commuication. When broker send message using QoS>0, the client should do something before sending ack to ensure the message is really processed. 

For example, i created a thread pool to process message in parallel.

```java
    @Bean
    public ThreadPoolTaskExecutor mqttMessageExecutor() {
        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
        executor.setCorePoolSize(2);
        executor.setMaxPoolSize(10);
        executor.setQueueCapacity(100);
        executor.setThreadNamePrefix("mqtt_");
        executor.setRejectedExecutionHandler(new ThreadPoolExecutor.CallerRunsPolicy());
        executor.setWaitForTasksToCompleteOnShutdown(true);
        return executor;
    }
    @Bean
    public MessageChannel mqttInboundChannel(ThreadPoolTaskExecutor mqttMessageExecutor) {
        return new ExecutorChannel(mqttMessageExecutor);
    }
```

If the service shutdown abnormally when processing message, all message in queue will be lost. So, ack should be sent after `MessageHandler` handle the message. Or, using original `setManualAcks`, client can send ack manually.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3181


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -51,6 +54,25 @@
...
+
+		/**
+		 * Client will auto ack when integration flow returns.
+		 */
+		AUTO,
+
+		/**
+		 * User must ack via the {@link IntegrationMessageHeaderAccessor#ACKNOWLEDGMENT_CALLBACK} header.
+		 */
+		MANUAL
```
artembilan: Can't we just live with similar to the client option like `manualAcks(boolean)` ?

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -379,4 +420,46 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
...
+			if (this.ackClient != null || !this.ackClient.equals(MqttPahoMessageDrivenChannelAdapter.this.client)) {
+				try {
+					this.ackClient.messageArrivedComplete(this.id, this.qos);
+				}
+				catch (MqttException e) {
+					throw new IllegalStateException(e);
+				}
+			}
+			else {
+				throw new IllegalStateException("Client has changed");
```
artembilan: I wonder why do we care about this.
We are just have her self-containing `AcknowledgmentImpl`  with the client it has been created.
If that client becomes invalid at the moment of `acknowledge()` we have nothing to do anyway.

What am I missing?

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/support/Acknowledgment.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+public interface Acknowledgment {
+
+	/**
+	 * Complete the message arrival.
+	 * @throws IllegalStateException if the client has been disconnected since this
+	 * was created.
+	 */
+	void acknowledge();
```
artembilan: `@FunctionalInterface` ?

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/support/MqttHeaderAccessor.java
```diff
@@ -0,0 +1,94 @@
...
+		return message.getHeaders().get(MqttHeaders.DUPLICATE, Boolean.class);
+	}
+
+	/**
+	 * Return the acknowledgment callback header.
+	 * @param message the message.
+	 * @return the header.
+	 */
+	public static Acknowledgment acknowledgment(Message<?> message) {
+		return message.getHeaders().get(IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK, Acknowledgment.class);
```
artembilan: This is kinda not compatible with what we have in the `StaticMessageHeaderAccessor.getAcknowledgmentCallback()`.

Shouldn't we have a singe place of truce even if that `AcknowledgmentCallback` is not going to be in a full power for MQTT as it was designed?

Thanks

src/reference/asciidoc/mqtt.adoc
```diff
@@ -145,6 +147,23 @@ A new application context reverts to the configured settings.
 
 Changing the topics while the adapter is stopped (or disconnected from the broker) takes effect the next time a connection is established.
 
+[[mqtt-ack-mode]]
+==== Manual Acks
+
+Starting with version 5.3, you can set the `ackMode` property.
+Often used to asynchronously acknowledge delivery.
+When set to `MANUAL`, an `Acknowledgment` header (`IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACKß`) is added to the message.
```
artembilan: What is that `ß` symbol? How have you managed to type it at all? 😄

src/reference/asciidoc/whats-new.adoc
```diff
@@ -116,3 +116,9 @@ See <<./rsocket.adoc#rsocket-inbound,RSocket Inbound Gateway>> for more informat
 
 A `LeaderInitiatorFactoryBean` (as well as its XML `<int-zk:leader-listener>`) exposes a `candidate` option for more control over a `Candidate` configuration.
 See <<./zookeeper.adoc#zk-leadership,Leadership event handling>> for more information.
+
+[[x5.3-mqtt]]
+=== MQTT Changes
+
+The inbound channel adapter can now be configured to provide user control over when a message is acknowledged as being delivered.
```
artembilan: Can we also mention here recently added `MqttConnectionFailedEvent` for the `MqttPahoMessageHandler` ?

artembilan (changes requested):
The biggest my concern here is that `Acknowledgment` new contract which brings a conflicts with existing `AcknowledgmentCallback`.
May be the last one could extend the simple  `Acknowledgment` with a `default` method impl for `acknowledge()` ?

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/support/MqttHeaderAccessor.java
garyrussell: Good point.

artembilan: I fact I see that SIK implements an `org.springframework.kafka.support.Acknowledgment` for full compatibility.
So, probably we really need to introduce that simple `Acknowledgment` here and have it extended from the `AcknowledgmentCallback`.
Then we may introduce a new `Acknowledgment ack = StaticMessageHeaderAccessor.getAcknowledgment()` for end-user convenience.

spring-integration-core/src/main/java/org/springframework/integration/acks/SimpleAcknowledgment.java
```diff
@@ -0,0 +1,36 @@
...
+ * @author Gary Russell
+ * @since 5.3
+ *
+ */
+@FunctionalInterface
+public interface SimpleAcknowledgment {
+
+	/**
+	 * Complete the message arrival.
+	 * @throws IllegalStateException if the client has been disconnected since this
```
artembilan: I think this one is not relavant to the generic contract...

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
artembilan: How that is possible that `this.ackClient` can be a `null`?
Why do we need this `if..else` logic here at all?

spring-integration-mqtt/src/main/resources/org/springframework/integration/mqtt/config/spring-integration-mqtt.xsd
```diff
@@ -89,6 +89,14 @@
 					</xsd:documentation>
 				</xsd:annotation>
 			</xsd:attribute>
+			<xsd:attribute name="manual-acks" type="xsd:boolean" default="false">
```
artembilan: In other places we do like this:
```
<xsd:simpleType>
	<xsd:union memberTypes="xsd:boolean xsd:string"/>
</xsd:simpleType>
```
to avoid validation errors in the IDE when we resolve a value against properties placeholder.
Search for some `boolean` exactly in this `spring-integration-mqtt.xsd`

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -379,4 +399,46 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
 	public void deliveryComplete(IMqttDeliveryToken token) {
 	}
 
+	/**
+	 * Used to complete message arrival when {@link AckMode#MANUAL}.
+	 *
+	 * @since 5.3
+	 */
+	private static class AcknowledgmentImpl implements SimpleAcknowledgment {
```
artembilan: I wonder if this internal one should fully implement a `AcknowledgmentCallback` fro compatibility with a `AcknowledgmentCallback getAcknowledgmentCallback(Message<?> message)`  contract...

Any objection do not do that?

Thanks

artembilan (changes requested):
A review for the current code state.

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
garyrussell: It's just for safety. I don't know the client internals and whether `connectionLost` can be called concurrently with `messageArrived` so there could be a race.

spring-integration-mqtt/src/main/resources/org/springframework/integration/mqtt/config/spring-integration-mqtt.xsd
garyrussell: I thought about this, but it is unlikely that someone would set this via a property - the code either manually acks or not.

garyrussell: But I can change it if you insist.

artembilan: No, I'm OK with your explanation.
More over it doesn't look like people worry about XML config these days too much 😉

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
garyrussell: That is intended for message source implementations. We can't support its other features here.

artembilan:
Are you familiar with issue in docs:
```
Execution failed for task ':checkAsciidocLinks'.

> [*** Anchor reference errors found:, 

  Anchor 'mqtt-events' not found in 'whats-new.adoc', if in another file, it needs to be qualified with './fileName.adoc#']
```
See Travis report.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
artembilan: OK. We an give it a shot.
But let's document it in that docs properly, that this MQTT ack is not an `AcknowledgmentCallback`.

Also I wonder if we should say something on the matter in the `polling-consumer.adoc#deferred-acks-message-source` ...

garyrussell: I will add some more docs...

artembilan (approved):
👍 
Will merge after Travis confirmation.

PR #3225, opened 2020-03-18
GH- 3223: Build `ResolvableType` after mapping

### LINKED ISSUE #3223: Resolving issue with resolvableType/TypeId and AMQP (ClassNotFoundException)
**Affects Version(s):** 5.2.4 (maybe 5.2.x)

**DEMO:** https://github.com/knalli/poc-sprint-int-resolve-type-error/tree/e7def89d406a5cb19276a636cda85afdc15619ea

---
Since updating to Integration 5.2 (via Spring Boot 2.2), I am noticing an issue regarding `json__resolvableType` and `json___TypeId__`.

There are some issues around this context (I'm already searched here): #3096, #3163

Small overview about my setup (and this is also included in [my poc/demo here](https://github.com/knalli/poc-sprint-int-resolve-type-error)): different Spring Boot applications sending basically JSON-serialized objects via RabbitMQ (AMQP) using Spring Integration (DSL) ecspecially `Transformer.toJSON()` and `Transformer.fromJSON()`.

Until Spring Boot 2.1, the same code as in the demo (without the filter for excluding both header fields) worked. However, after updating to the 2.2 (and thereby Spring Integration 5.2.?), I was faced with `org.springframework.integration.transformer.MessageTransformationException: failed to transform message; nested exception is java.lang.IllegalStateException: java.lang.ClassNotFoundException` produced by the `JsonToObjectTransformer.getClassForValue`. Yes, the sending object type is not available under the same canonical class name. Why should it?

Because lacking time, I found the workaround by adding a filter for `json__resolvableType` and `json___TypeId__` on the incoming side:

https://github.com/knalli/poc-sprint-int-resolve-type-error/blob/e7def89d406a5cb19276a636cda85afdc15619ea/incoming/src/main/java/de/knallisworld/poc/incoming/IncomingMain.java#L86-L87

And then I just forgot about this again, I'm sorry.

Now, with the update to Spring Boot 2.2.5 and thereby Spring Integration 5.2.4, there is a new warning logging even with my workaround filter excluding this header:

```
2020-03-18 11:23:50.387  WARN 21821 --- [erContainer#0-1] o.s.i.a.support.DefaultAmqpHeaderMapper  : error occurred while mapping from AMQP properties to MessageHeaders

java.lang.IllegalStateException: java.lang.ClassNotFoundException: …

Caused by: …
  at JsonHeaders.getClassForValue
```

You can replay this with my poc:
* It is a multi module maven project so each module (incoming, outgoing) have different class paths.
* Start a RabbitMQ instance (e.g. via docker) or configure an existing one via the properties.
* Start IncomingMain
* Start OutgoingMain (which generates one message)

On the receiving side, you will get the error logged out. 

If you downgrade Spring Integration via `incoming/pom.xml` to 5.2.3, there is not logger warning anymore.

---
So, there are now the following things:

1. I have no idea how I can exclude this on the receiving side: The headers will be parsed before any filter is done (which explains why my workaround does not help any more).
2. Additionally, I don't know how to exclude them on the sending side: I tried naiv overwriting this headers with an enricher, but seems not to be the way..
3. IMHO, I don't like that the default seems to load any class provided in this fields. Should this be an opt-in maybe?
4. Even it is "only a warning", it is basically an internal exception so we cannot be sure all headers are parsed, right? So disabling the logger won't help in the long run.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3223

The `json__TypeId__` header may have a value which cannot be resolved to the `Class` in the current classpath
So, skip `ResolvableType` building logic until we really sure that end-user wants to map JSON headers
* WARN a build exception that we can't load a class for the `json__TypeId__` when we try to build a `ResolvableType`
in the  `DefaultAmqpHeaderMapper`
* Document the negation feature for JSON headers

**Cherry-pick to 5.2.x**

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/DefaultAmqpHeaderMapper.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/support/DefaultAmqpHeaderMapper.java
@@ -151,24 +151,11 @@ public class DefaultAmqpHeaderMapper extends AbstractHeaderMapper<MessagePropert
 					headers.put(jsonHeader, value);
 				}
 			}
-
-			createJsonResolvableTypHeaderInAny(headers);
 		}
 		catch (Exception e) {
-			if (logger.isWarnEnabled()) {
-				logger.warn("error occurred while mapping from AMQP properties to MessageHeaders", e);
-			}
+			this.logger.warn("error occurred while mapping from AMQP properties to MessageHeaders", e);
 		}
 		return headers;
-	}
-
-	private void createJsonResolvableTypHeaderInAny(Map<String, Object> headers) {
-		Object typeIdHeader = headers.get(JsonHeaders.TYPE_ID);
-		if (typeIdHeader != null) {
-			headers.put(JsonHeaders.RESOLVABLE_TYPE,
-					JsonHeaders.buildResolvableType(getClassLoader(), typeIdHeader,
-							headers.get(JsonHeaders.CONTENT_TYPE_ID), headers.get(JsonHeaders.KEY_TYPE_ID)));
-		}
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


garyrussell:
Cherry-picked as 9a6849740f65464df30948a5e94c7b2e841da9ed

Issue #3226, opened 2020-03-19
Implement Channel Adapters for Redis Streams


akuma8:
It's a reminder of that JIRA issue: https://jira.spring.io/browse/INT-4567
I currently need this feature in my project and I don't see how to implement it easily. 
I can contribute with a little guidance since I never contributed to SI before.
Thanks

artembilan:
See `StreamOperations` in Spring Data Redis and respective documentation: https://docs.spring.io/spring-data/redis/docs/2.3.0.M4/reference/html/#redis.streams

Also see https://spring.io/projects/spring-kafka for the Spring API around Apache Kafka. Don't confuse yourself with Kafka Streams though, because it looks like Redis Streams is more like regular Kafka Consumer/Producer.
There is also https://github.com/spring-projects/spring-integration-kafka project - Spring Integration channel adapter implementations around  Spring Kafka abstractions.

So, the goal of this issue is to have channel adapter implementations around Spring Data `StreamOperations` API.

Does it make sense?

Thank you for willing to contribute the feature!

akuma8:
@artembilan Thank you for these details. I take a look about the implementation, I'll let you know if I need other information.

PR #3227, opened 2020-03-24
WIP : Channel Adapters for Redis Streams

akuma8:
#3226 
The PR is in WIP because I need information for improvements.


spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisStreamInboundChannelAdapter.java
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    @Override
+    public String getComponentType() {
+        return "redis:stream-inbound-channel-adapter";
+    }
+
+    private Message<?> convertMessage( Record<Object, Object> message ) {
+        return this.messageConverter.toMessage( Record.of( message ), null );
+    }
```
akuma8: @artembilan How to improve this converter? 
Currently I send back a `Record` to the output channel, it would be ideal to extract the paylod and headers and construct a real `Message`  before publishing it.

spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStreamOutboundChannelAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+    public String getComponentType() {
+        return "redis:stream-outbound-channel-adapter";
+    }
+
+    @Override
+    protected void handleMessageInternal( Message<?> message ) {
+        ObjectRecord<String, Message<?>> record = StreamRecords
+                .<String, Message<?>> objectBacked( message )
+                .withStreamKey( streamName );
+        this.streamOperations.add( record );
```
akuma8: @artembilan How can I improve this code?
Currently I don't make any difference about incoming data type. I also publish all the message as is because I want to keep its headers.
Also I have some troubles about how well converting and reading data. 
Say I have that class:
```
public class SimpleUser implements Serializable {
    private final String nom;
    private final String prenom;
    private final String username;
}
```
When publishing a message with a paylod of type `SimpleUser ` I am unable to get it from the inbound part (see my comment above). In the inbound side, I would like to extract it before sending it to the output channel.

artembilan (changes requested):
I hope you find my superficial review sufficient and we are good to continue a work here.

Thank you for your effort!

artembilan:
@akuma8 ,

please, don't squash commits when you push them to the PR: it is going to be much easier for us to review when change history is present.
Also it would be great to comment the changes in commit: kinda what has been done and the reason.
It also helps us to review faster.

See more info about good commits in Chris Beams' article: https://chris.beams.io/posts/git-commit/

Thanks for understanding.

P.S. for now it is OK because PR is small yet, but it is going to grow I believe.

akuma8:
@artembilan Thank you for the remark, I will be careful in the future. It's a personal habit.
I let you a comment because I am stuck, please let me know when you will read it.

artembilan:
>  please let me know when you will read it.

I'm in review process. 
Stay tuned. BTW not reason to be in hurry: we are OK to release it whenever it is ready.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisStreamInboundChannelAdapter.java
```diff
@@ -0,0 +1,155 @@
...
+    protected void doStart() {
+        super.doStart();
+        createGroup();
+        this.container.start();
+    }
+
+    @Override
+    protected void doStop() {
+        super.doStop();
+        this.container.stop();
```
artembilan: I think `container` has to be stopped first.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisStreamInboundChannelAdapter.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public void setConsumerName( @Nullable String consumerName ) {
+        this.consumerName = consumerName;
+    }
+
+    public void setCreateGroupeIfNotExist( boolean createGroupeIfNotExist ) {
+        this.createGroupeIfNotExist = createGroupeIfNotExist;
+    }
+
+    public void setReadOffset( ReadOffset readOffset ) {
```
artembilan: Setters/getters have to go after ctors.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisStreamInboundChannelAdapter.java
```diff
@@ -0,0 +1,155 @@
...
+        this.readOffset = readOffset;
+    }
+}
```
artembilan: Every class member has to be wrapped with blank lines.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisStreamInboundChannelAdapter.java
```diff
@@ -0,0 +1,155 @@
+package org.springframework.integration.redis.inbound;
```
artembilan: Copyright has to be present in source files. In classes to be precise.

spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStreamMessageHandler.java
```diff
@@ -0,0 +1,90 @@
...
+
+	private volatile RedisSerializer<?> serializer = StringRedisSerializer.UTF_8;
+
+	private volatile Expression streamKeyExpression;
+
+	private volatile boolean extractPayload = true;
+
+	public RedisStreamMessageHandler( String streamKey, RedisConnectionFactory connectionFactory ) {
+		Assert.notNull( streamKey, "'streamName' must not be null" );
+		Assert.notNull( connectionFactory, "'connectionFactory' must not be null" );
```
artembilan: We don't need this because `template.afterPropertiesSet();` ensures such an assertion.

spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStreamMessageHandler.java
```diff
@@ -0,0 +1,90 @@
...
+		this.serializer = serializer;
+	}
+
+	public void setStreamKey( String streamKey ) {
+		Assert.hasText( streamKey, "'streamKey' must not be an empty string." );
+		this.streamKeyExpression = new LiteralExpression( streamKey );
+	}
+
+	public void setStreamKeyExpression( Expression streamKeyExpression ) {
+		this.streamKeyExpression = streamKeyExpression;
```
artembilan: This cannot be `null`. So, an appropriate assert.

Also I think this is really an important option, so has to be in the ctor.
I'm OK with an existing ctor for the `String streamKey`, but then we need to have a new one based on the ` Expression streamKeyExpression`.
Therefore `final` for expression property and not setters at all.

spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStreamMessageHandler.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public class RedisStreamMessageHandler extends AbstractMessageHandler {
+
+	private final StreamOperations streamOperations;
+
+	private volatile RedisSerializer<?> serializer = StringRedisSerializer.UTF_8;
+
+	private volatile Expression streamKeyExpression;
+
+	private volatile boolean extractPayload = true;
```
artembilan: As I said before: no `volatile` for properties which are not changed at run time.

spring-integration-redis/src/main/java/org/springframework/integration/redis/outbound/RedisStreamMessageHandler.java
```diff
@@ -0,0 +1,90 @@
...
+	@Override
+	protected void onInit() {
+		super.onInit();
+		Assert.notNull( streamKeyExpression, "'streamKeyExpression' must not be null" );
+	}
+
+	@Override
+	protected void handleMessageInternal( Message<?> message ) {
+
+		String streamKey = this.streamKeyExpression.getExpressionString();
```
artembilan: That's something awkward.
You probably need to make yourself familiar with an `Expression.getValue()`.
See other `MessageHandler` implementations in this Redis module to determine how we evaluate similar SpELs.

artembilan (changes requested):
You know I have just discovered for myself a `ReactiveStreamOperations` in Spring Data Redis.
I my experience it is better to develop Reactive component and let it to be used in the blocking scenarios.

Can we revise your solution in favor of `ReactiveStreamOperations`?
See `ReactiveMongoDbStoringMessageHandler` for example.

The `RedisStreamInboundChannelAdapter` could also be based on the `ReactiveStreamOperations.read()`. The returned `Flux` could be damped into a `ReactiveStreamsSubscribableChannel` if configured that way or as regular `send()` otherwise.

Let me know WDYT?

akuma8 (commented):
> Can we revise your solution in favor of `ReactiveStreamOperations`?

Sure, I saw `ReactiveStreamOperations` in the doc but I thought it only be relevant to use it in a full reactive context which is not our case but I will revise my first solution if this one is better.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisStreamInboundChannelAdapter.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private volatile StreamMessageListenerContainer                                       container;
+
+    private volatile StreamMessageListenerContainer.StreamMessageListenerContainerOptions containerOptions       = StreamMessageListenerContainer.StreamMessageListenerContainerOptions
+            .builder()
+            .serializer( StringRedisSerializer.UTF_8 )
+            .pollTimeout( Duration.ZERO )
+            .executor( new SimpleAsyncTaskExecutor() )
+            .batchSize( 1 )
+            .objectMapper( new ObjectHashMapper() )
```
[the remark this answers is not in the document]
akuma8: I simply didn't see it, my mistake.

About formatting, I will do it at the end, when we will have something that works as expected.

artembilan:
Hi, @akuma8 !

Thank you for update! I'll review your solution and try to figure problems the next week.
Meanwhile the answer to your question: yes, I want an *outbound* part as reactive as well. You need to implement a `ReactiveMessageHandler` over there. See a `ReactiveMongoDbStoringMessageHandler` for example.

Right now I'm working on some reactive streams functionality for the `MessageProducerSupport`. I somehow believe your inbound channel adapter would benefit from that.
Let's see the next week how it is going!

artembilan:
I have just raised a PR https://github.com/spring-projects/spring-integration/pull/3240 about reactive Message Producer.
I have over there an implementation for MongoDb change stream.
I think that one should be a sample how this Redis Streams has to be implemented on the inbound side.
Let's hope it will be merged next week, so you would have some foundation to rely on!

artembilan (changes requested):
@akuma8 ,

tell, me, please, how that would be hard for you to divide this PR into two.
One for already working functionality on the outbound side.
And then we can come back to be inbound one when PR #3240 is merged.
This way we won't have everything blocked.

Thank for understanding.

akuma8:
Hi @artembilan
Since Monday I am sick, it’s not the Covid-19 😁.
It’s a good idea to split this PR in 2, I’ll do that next week, I hope getting better.

artembilan:
Stay safe, @akuma8 , and get well soon!

No need to be in a hurry: looks like no pressure for the feature yet. So, we are good to bring it whenever it is ready.

When you come back from your sick time off, please, divide this into two PRs - inbound & outbound.
Please, make sure you contribute some tests for this new channel adapters.

At a glance `MessageHandler` part is OK - should be merged shortly after an appropriate PR.

I haven't looked into an inbound part because wanted to be sure that `MessageProducerSupport.subscribeToPublisher(Publisher<Message<?>>)` approach is correct.

Now it is merged to master and you can start looking into the `MongoDbChangeStreamMessageProducer` to make your inbound part similar way.

Fingers crossed for your health!

artembilan (changes requested):
Hi @akuma8 !

Any chances to see an update from you on the matter?

What we have so far and what we still need to do to have that Redis Stream Inbound Channel Adapter?

Thanks

akuma8:
Hi @artembilan,
I am a bit busy, I started looking at the inbound part, I hope opening a first PR next week. 
I also opened that issue about [reading List from a Redis Stream](https://jira.spring.io/browse/DATAREDIS-1175)
I'll let you know. 
Thanks

artembilan:
Hi @akuma8 !

No problem with your load!

Just let me know that you can make it before September or should I pull this task from you and fix myself!
Great to see the fix in Spring Data Redis. I think it's time reinstate your test for list. Nothing more we need to do on the matter.
See a related PR in Spring Data: they have some tests to investigate.

akuma8:
@artembilan,
> ... you can make it before September...

Yes, I think so. If we keep doing it like we did for the outbound side I can make it before September.
I am also looking for a way to test the fix from Spring Data Redis, if you can include the related version, please let me know.

artembilan:
>  if you can include the related version, please let me know.

Just did: https://github.com/spring-projects/spring-integration/commit/ebec500abd85d10b9eb584a8794f1abba4acf0f6

Upgraded to Spring Data `2020.0.0-SNAPSHOT`

akuma8:
Hi @artembilan,
I have some questions about implementation choices.
Redis Stream Api provides 2 ways to read data, we can read as a simple client with the `XREAD` command or as a client belonging to a Consumer Group: `XREADGROUP`. Should we provide one inbound implementation including the 2 options or devide them in 2 different implementations? 
Thanks a lot

artembilan:
Yes, if that is possible, I'd prefer to have a one inbound component with a `consumerGroup` property (optional).
And yes, it would be better to have it as reactive. I think the `MongoDbChangeStreamMessageProducer` is a good sample to borrow ideas.

Thanks

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisStreamInboundChannelAdapter.java
```diff
@@ -0,0 +1,171 @@
...
+		if ( StringUtils.isEmpty( this.consumerGroupName ) ) {
+			this.receiver
+					.receive( offset )
+					.subscribe( record -> sendMessage( convertMessage( ( Record<Object, Object> ) record ) ) );
+		}
+		else {
+			Assert.hasText( consumerName, "'consumerName' must be set" );
+			Consumer consumer = Consumer.from( this.consumerGroupName, this.consumerName );
+			this.receiver
+					.receiveAutoAck( consumer, offset )
```
artembilan, line 142: Looks like we also need to think about manual acknowledgement:
```
 * Every record must be acknowledged using
	 * {@link org.springframework.data.redis.connection.ReactiveStreamCommands#xAck(ByteBuffer, String, String...)}
```

See `receive()` API for the `StreamReceiver`.
This has to be populated as a `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK`.
And also see `SimpleAcknowledgment`  abstraction to wrap the call to that `xAck`. for the particular record returned.

Saying that I'd expect an `autoAck(true|false)` option on this channel adapter.

artembilan (changes requested):
This is good progress, but definitely there is still some work to do: tests, docs, JavaDocs etc.

Thanks

akuma8:
Hi @artembilan,
Sorry but I forgot to tell you to not review this, I just rebased master branch. 
I planned to delete this PR and opening a new one.

artembilan:
OK. So, I'm closing this one to avoid confusion.

Thanks for confirmation!

PR #3232, opened 2020-03-31
GH-3230: MMIH: Fix Pausable/Lifecycle Methods

### LINKED ISSUE #3230: ServiceActivator methods not discovered if named "start", "stop" or "isRunning"
**Affects Version(s):** 5.1.0

## Bug report
If a service activator has methods named "start", "stop" or "isRunning", these cannot be automatically discovered by SI. The activating method must be specified explicitly.

`Target object of type [...] has no eligible methods for handling Messages.`

This happens due to these methods being filtered in [MessagingMethodInvokerHelper.java#L827](https://github.com/spring-projects/spring-integration/blob/a6fb891a817dacc1af9fa238fa0797623ea65416/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java#L827)

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3230

Previously, methods with the same name as `Lifecycle` methods were unconditionally ignored when no method name provided.

- Use logic similar to `ControlBusMethodFilter` to explicitly compare the candidate Method
  to those on `Paused` and `Lifecycle` and only ignore those.
- Explicitly disallow these methods when named - previously the check was only performed
  if no method name was supplied.

**Consideration for backporting:**

**Behavior change, test `lifecycleOnlyExplicitMethod()` fails on 5.2.x (method explicitly requested works).**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -818,19 +818,24 @@ public class MessagingMethodInvokerHelper extends AbstractExpressionEvaluator im
 	}
 
 	private boolean isMethodEligible(Method methodToProcess) {
-		return (!(methodToProcess.isBridge() || // NOSONAR boolean complexity
+		return !(methodToProcess.isBridge() || // NOSONAR boolean complexity
 				isMethodDefinedOnObjectClass(methodToProcess) ||
 				methodToProcess.getDeclaringClass().equals(Proxy.class) ||
 				(this.requiresReply && void.class.equals(methodToProcess.getReturnType())) ||
-				(this.methodName != null && !this.methodName.equals(methodToProcess.getName()))))
-			&& !isPausableMethod(methodToProcess);
+				(this.methodName != null && !this.methodName.equals(methodToProcess.getName())) ||
+				(this.methodName == null && isPausableMethod(methodToProcess)));
 	}
 
 	private boolean isPausableMethod(Method pausableMethod) {
 		Class<?> declaringClass = pausableMethod.getDeclaringClass();
-		return (Pausable.class.isAssignableFrom(declaringClass) || Lifecycle.class.isAssignableFrom(declaringClass))
+		boolean pausable = (Pausable.class.isAssignableFrom(declaringClass)
+					|| Lifecycle.class.isAssignableFrom(declaringClass))
 				&& ReflectionUtils.findMethod(Pausable.class, pausableMethod.getName(),
 						pausableMethod.getParameterTypes()) != null;
+		if (pausable && this.logger.isTraceEnabled()) {
+			this.logger.trace(pausableMethod + " is not considered a candidate method unless explicitly requested");
+		}
+		return pausable;
 	}
 
 	private void populateHandlerMethod(Map<Class<?>, HandlerMethod> candidateMethods,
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -818,14 +818,19 @@ private HandlerMethod obtainHandlerMethodIfAny(Method methodToProcess) {
...
-				(this.methodName == null &&
-						ObjectUtils.containsElement(new String[] { "start", "stop", "isRunning" },
-								methodToProcess.getName())));
+				(this.methodName != null && !this.methodName.equals(methodToProcess.getName()))))
+			&& !isPausableMethod(methodToProcess);
+	}
+
+	private boolean isPausableMethod(Method pausableMethod) {
+		Class<?> declaringClass = pausableMethod.getDeclaringClass();
+		return (Pausable.class.isAssignableFrom(declaringClass) || Lifecycle.class.isAssignableFrom(declaringClass))
```
artembilan: Any chances that we log special info that `Pausable` methods are not considered as candidates?

artembilan (changes requested):
We probably need to document also that `Pausable` are not selected as candidates.
It is going to work only if you provide a name explicitly.

Thanks

garyrussell:
>We probably need to document also that Pausable are not selected as candidates.
It is going to work only if you provide a name explicitly.

??

If they were candidates, wouldn't they be ambiguous because they have similar signatures?

artembilan (changes requested):
> wouldn't they be ambiguous because they have similar signatures?

No, I mean skip them from scanning at all. Exactly what we do right now.

Only concern left for this PR is some doc note to let end-users to know that those methods are not selected by default and has to be used explicitly if there is such a requirement.

garyrussell:
>Only concern left for this PR is some doc note to let end-users to know that those methods are not selected by default and has to be used explicitly if there is such a requirement.

Sorry; I still don't understand what you mean; if someone really wants to call one of the Lifecycle methods they would specify it's name. Why would it be necessary to tell them anything? Why would they expect us to select one at random?

artembilan:
> Why would they expect us to select one at random?

OK. Maybe my language is bad these days. Let me quote a reporter:

> If a service activator has methods named "start", "stop" or "isRunning", these cannot be automatically discovered by SI. The activating method must be specified explicitly.

So, I want to avoid this confusing part: if we don't specify a method name we should not expect `Pausable` methods to be selected as service-activator candidates.
I mean that we never explained such a design decision in our docs.

garyrussell:
But that's no longer the case after this fix - see the test `(String) null`.

artembilan:
Right. That's what I mean. Your `LifecycleBean` is not OK for candidates since its `start()/stop()` comes from the `Pausable/Lifecycle`.
But `LifeCycleWithCustomStart` is good because its `start()` doesn't belong to `Pausable/Lifecycle`.

I believed that the main concern of the issue that `start()` from `Lifecycle` is not selected automatically and there is no a note on the matter.

Even if current behavior is slightly different than before we should say something about `Pausable` methods auto-discovery.

I'm fine merging this as is though: no concerns from the code perspective.

garyrussell:
No; the original problem was that the user happened to create his POJO method, something like `start(Message<?>)`. And the old code ignored it. Now we don't ignore it.

>We just happened to name a method of an endpoint POJO "start". The error message was not helpful in finding the cause; If these method names are not allowed, maybe we should document it in Service Activator?

artembilan:
> And the old code ignored it. Now we don't ignore it.

OK. That's good. I fully agree with the logic we have so far.
Just wander you don't want to document this change... 😄 

That's all.

(I'll merge it is after your next reply)

P.S. Sorry for wasting your time probably for nothing...

garyrussell:
I think that adding...

"You can now name POJO methods with the same names as Lifecycle methods and they will be detected (if not ambiguous). Previously, such methods were not considered as candidates."

...is not necessary. But I will add it if you want.

artembilan (approved):
> is not necessary.

OK. Convinced.  😄 

Merging...

artembilan:
... and cherry-picked to `5.2.x`

PR #3233, opened 2020-03-31
Add more Kotlin DSL for sub-flows

artembilan:
* Introduce `KotlinSplitterEndpointSpec`, `KotlinEnricherSpec` &
`KotlinFilterEndpointSpec` to avoid an explicit use of `IntegrationFlow` usage
from Kotlin DSL
* Fix `AbstractKotlinRouterSpec` to extend `ConsumerEndpointSpec` for access
to more endpoint options as it is with Java DSL for similar specs
* Introduce a `KotlinIntegrationFlowDefinition.publishSubscribe()`
for a `BroadcastCapableChannel` and `vararg` of `KotlinIntegrationFlowDefinition` builders.
This allows us to not introduce a `BroadcastPublishSubscribeSpec` wrapper for Kotlin
and let us to remove `publishSubscribeChannel()` DSL methods for `PublishSubscribeSpec`
since it is covered with the `publishSubscribe(PublishSubscribeChannel())` and
respective set of sub-flows in Kotlin DSL as subscribers
* Use new Kotlin specs in the `KotlinIntegrationFlowDefinition` instead of their
Java variants

### DIFF
--- a/spring-integration-core/src/main/kotlin/org/springframework/integration/dsl/AbstractKotlinRouterSpec.kt
+++ b/spring-integration-core/src/main/kotlin/org/springframework/integration/dsl/AbstractKotlinRouterSpec.kt
@@ -29,7 +29,8 @@ import org.springframework.messaging.MessageChannel
  * @since 5.3
  */
 abstract class AbstractKotlinRouterSpec<S : AbstractRouterSpec<S, R>, R : AbstractMessageRouter>(
-		open val delegate: AbstractRouterSpec<S, R>) {
+		open val delegate: AbstractRouterSpec<S, R>)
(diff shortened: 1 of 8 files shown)


garyrussell (changes requested):
Any doc updates for this?

Also, the index/ToC needs a reference to the Kotlin DSL chapter - only reachable via the What's New.

artembilan:
> Any doc updates for this?

No need. It is just transparent from the `integrationFlow()` API. Those new Kotlin classes are really not intended for end-user use: you just end up with imposed code flow in the current typing context.

I'll fix an index. Thanks. One day I remembered that, but another forgot to fix. 😄

PR #3234, opened 2020-03-31
GH-3219: Add `HandleMessageAdviceAdapter`

### LINKED ISSUE #3219: Consider to implement HandleMessageAdviceAdapter implements HandleMessageAdvice
Right now we can advice only a `handleRequestMessage()` for the particular `MessageHandler` in the flow.
To make the whole sub-flow advised we need to implement a `HandleMessageAdvice` for the particular functionality, like retry or transactions.

Something simple like:
```
public class HandleMessageAdviceAdapter implements HandleMessageAdvice {
	
	private final MethodInterceptor delegate;

	public HandleMessageAdviceAdapter(MethodInterceptor delegate) {
		this.delegate = delegate;
	}


	@Override
	public Object invoke(MethodInvocation invocation) throws Throwable {
		return this.delegate.invoke(invocation);
	}

}
```
in the Framework should simplify end-user requirements for sub-flow proxying from the target configuration.

See for more info: https://stackoverflow.com/questions/60718734/how-to-retry-an-entire-poller-flow-in-spring-integration

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3219

### DIFF
--- a/src/reference/asciidoc/handler-advice.adoc
+++ b/src/reference/asciidoc/handler-advice.adoc
@@ -685,7 +685,8 @@ In the case of a `MessageHandler` that does not return a response, the advice ch
 
 Starting with version 5.3, the `HandleMessageAdviceAdapter` is present to let apply any existing `MethodInterceptor` for the `MessageHandler.handleMessage()` and, therefore, whole sub-flow.
 For example a `RetryOperationsInterceptor` could be applied for the whole sub-flow starting from some endpoint, which is not possible by default because consumer endpoint applies advices only for the `AbstractReplyProducingMessageHandler.RequestHandler.handleRequestMessage()`.
-See `HandleMessageAdviceAdapter` JavaDocs for more information.
+Starting with version 5.3, the `HandleMessageAdviceAdapter` is provided to apply any `MethodInterceptor` for the `MessageHandler.handleMessage()` method and, therefore, the whole sub-flow.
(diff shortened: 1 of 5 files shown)


src/reference/asciidoc/handler-advice.adoc
```diff
@@ -683,6 +683,10 @@ Note that, in that case, the entire downstream flow is within the transaction sc
 
 In the case of a `MessageHandler` that does not return a response, the advice chain order is retained.
 
+Starting with version 5.3, the `HandleMessageAdviceAdapter` is present to let apply any existing `MethodInterceptor` for the `MessageHandler.handleMessage()` and, therefore, whole sub-flow.
+For example a `RetryOperationsInterceptor` could be applied for the whole sub-flow starting from some endpoint, which is not possible by default because consumer endpoint applies advices only for the `AbstractReplyProducingMessageHandler.RequestHandler.handleRequestMessage()`.
+See `HandleMessageAdviceAdapter` JavaDocs for more information.
```
garyrussell, line 688:
```suggestion
Starting with version 5.3, the `HandleMessageAdviceAdapter` is provided to apply any `MethodInterceptor` for the `MessageHandler.handleMessage()` method and, therefore, the whole sub-flow.
For example, a `RetryOperationsInterceptor` could be applied to the whole sub-flow starting from some endpoint; this is not possible, by default, because the consumer endpoint applies advices only to the `AbstractReplyProducingMessageHandler.RequestHandler.handleRequestMessage()`.
```

garyrussell (changes requested):

garyrussell (approved):

PR #3236, opened 2020-04-01
GH-2736: Async Mode for TcpOutboundGateway

### LINKED ISSUE #2736: Support setAsync() on TcpOutboundGateway
Support asynchronous replies in the `TcpOutboundGateway` to avoid blocking the calling thread to wait for the reply.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/2736

Support asynchronous request/reply.

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2001-2019 the original author or authors.
+ * Copyright 2001-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.integration.ip.tcp;
 
 import java.io.IOException;
+import java.util.Date;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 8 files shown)


garyrussell:
** DO NOT MERGE ** (XML support and flaky test).

spring-integration-ip/src/main/resources/org/springframework/integration/ip/config/spring-integration-ip.xsd
```diff
@@ -488,6 +488,16 @@
 					</xsd:documentation>
 				</xsd:annotation>
 			</xsd:attribute>
+			<xsd:attribute name="async" default="false">
+				<xsd:annotation>
+					<xsd:documentation>
+						Set to true for async request/reply - see reference manual.
```
artembilan, line 494: Would you mind to fix an `expresssion` typo in this file meanwhile?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
```diff
@@ -172,7 +196,9 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
 			throw new MessageHandlingException(requestMessage, "Interrupted in the [" + this + ']', e);
 		}
 		finally {
-			cleanUp(haveSemaphore, connection, connectionId);
+			if (!isAsync()) {
```
artembilan, line 199: Can we obtain `isAsync()` as local variable only once in the begging of method?
Kinda to avoid a race condition when it could be changed in between at runtime...

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/TcpOutboundGatewayTests.java
```diff
@@ -98,8 +105,8 @@
 	public static LongRunningIntegrationTest longTests = new LongRunningIntegrationTest();
 
 	@Rule
-	public Log4j2LevelAdjuster adjuster = Log4j2LevelAdjuster.trace();
-
+	public Log4j2LevelAdjuster adjuster = Log4j2LevelAdjuster.trace()
```
artembilan, line 108: Can we make this test class as JUnit 5 since we are here already?

artembilan (changes requested):
Just a couple suggestions.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/TcpOutboundGatewayTests.java
```diff
@@ -1011,7 +1002,7 @@ public void postSend(Message<?> message, MessageChannel channel, boolean sent) {
 	}
 
 	@Test
-	public void testAsyncTimeout() throws Exception {
```
artembilan: Just for info: IDEA doesn't mark such a `public` modifier as redundant. 
So, we don't need to bother ourselves to remove them any more 😄 

(If it doesn't annoy you, of course)

artembilan (approved):
Merging...

PR #3242, opened 2020-04-06
GH-3241: MetadataStoreSelector - compare old/new

### LINKED ISSUE #3241: IdempotentReceiverInterceptor Improvement
Enhance the `MetadataStoreSelector` to provide an option to evaluate an expression against the current value (if present) instead of simply `putIfAbsent`.

Consider https://stackoverflow.com/questions/61040057/spring-integration-process-files-line-by-line-persisting-read-state/61046158#61046158

Such an enhancement would allow a single MetadataStore entry with the key being the file and the value being the line number. The new option could compare the stored value to to the current line number and skip the line if it has already been processed.

Currently, we'd need a key containing the file name and line number.

garyrussell:
Resolves: https://github.com/spring-projects/spring-integration/issues/3241

Docs and XML to follow.


artembilan (changes requested):
LGTM so far!

You can go a head with the rest of changes.

Thanks

PR #3246, opened 2020-04-09
[DO NOT MERGE YET] GH-3103: Introduce CloudEvents transformers

### LINKED ISSUE #3103: Consider to support CloudEvents: HeaderMappers, Transformers, MessageConverters etc.
See more info about Java SDK: https://github.com/cloudevents/sdk-java

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3103

* Add an `io.cloudevents:cloudevents-api` optional dependency
* Introduce a `HeaderMapper` and `Marshallers` in the `support.cloudevents`
to marshal `CloudEvent` instances
* Introduce a `ToCloudEventTransformer` to build a `CloudEvent` instance
from a `Message` and optional marshaling logic if necessary.
Such a transformer could be used as a general purpose CE protocol binder
before sending a result message into the target protocol channel adapter

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
@@ -34,12 +34,7 @@ import io.cloudevents.v1.ContextAttributes;
  *
  * @since 5.3
  */
-public class HeaderMapper {
-
-	/**
-	 * Cloud event headers prefix as a {@value HEADER_PREFIX}.
-	 */
-	public static final String HEADER_PREFIX = "ce_";
+public final class HeaderMapper {
 
 	/**
 	 * Following the signature of {@link io.cloudevents.fun.FormatHeaderMapper}
@@ -59,7 +54,7 @@ public class HeaderMapper {
 										&& !ContextAttributes.datacontenttype.name().equals(attribute.getKey()))
 						.map(header ->
 								new AbstractMap.SimpleEntry<>(
-										HEADER_PREFIX + header.getKey().toLowerCase(Locale.US),
+										CloudEventHeaders.PREFIX + header.getKey().toLowerCase(Locale.US),
 										header.getValue()))
 						.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
@@ -43,8 +43,8 @@ public final class HeaderMapper {
 	 * @return The map of headers
 	 */
 	public static Map<String, String> map(Map<String, String> attributes, Map<String, String> extensions) {
-		Assert.notNull(attributes, "'attributes' must noy be null");
-		Assert.notNull(extensions, "'extensions' must noy be null");
+		Assert.notNull(attributes, "'attributes' must not be null");
+		Assert.notNull(extensions, "'extensions' must not be null");
 
 		Map<String, String> result =
 				attributes.entrySet()
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.cloudevents;
+
+import java.util.AbstractMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.util.Assert;
+
+import io.cloudevents.v1.ContextAttributes;
+
+/**
+ * A Cloud Event header mapper.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.3
+ */
+public class HeaderMapper {
+
+	/**
+	 * Cloud event headers prefix as a {@value HEADER_PREFIX}.
+	 */
+	public static final String HEADER_PREFIX = "ce_";
+
+	/**
+	 * Following the signature of {@link io.cloudevents.fun.FormatHeaderMapper}
+	 * @param attributes The map of attributes
+	 * @param extensions The map of extensions
+	 * @return The map of headers
+	 */
+	public static Map<String, String> map(Map<String, String> attributes, Map<String, String> extensions) {
+		Assert.notNull(attributes, "'attributes' must noy be null");
+		Assert.notNull(extensions, "'extensions' must noy be null");
+
+		Map<String, String> result =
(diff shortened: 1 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
```diff
@@ -0,0 +1,85 @@
...
+ * @author Artem Bilan
+ *
+ * @since 5.3
+ */
+public class HeaderMapper {
+
+	/**
+	 * Cloud event headers prefix as a {@value HEADER_PREFIX}.
+	 */
+	public static final String HEADER_PREFIX = "ce_";
```
n3wscott, line 42: `ce-`

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/Marshallers.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @return a builder to provide the {@link io.cloudevents.CloudEvent} and marshal as JSON
+	 * @see BinaryMarshaller
+	 */
+	public static <T> EventStep<AttributesImpl, T, byte[], String> binary() {
+		return BinaryMarshaller.<AttributesImpl, T, byte[], String>builder()
+				.map(AttributesImpl::marshal)
+				.map(Accessor::extensionsOf)
+				.map(ExtensionFormat::marshal)
+				.map(HeaderMapper::map)
+				.map(Json::binaryMarshal)
```
n3wscott, line 57: the payload may not be json

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
artembilan: Thanks for feedback!

However we talk here about a general approach for any possible protocol supported here in Spring Integration. I see you use proposed prefix for HTTP, but ours for Kafka. BTW Any other headers mapping in Spring Integration uses an underscore

artembilan: Thanks for feedback!

However we talk here about a general approach for any possible protocol supported here in Spring Integration. I see you use proposed prefix for HTTP, but ours for Kafka. BTW Any other headers mapping in Spring Integration uses an underscore

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/Marshallers.java
artembilan: But that’s exactly what you do for Kafka and HTTP. What am I missing?

artembilan: But that’s exactly what you do for Kafka and HTTP. What am I missing?

spring-integration-core/src/test/java/org/springframework/integration/transformer/ToCloudEventTransformerTests.java
```diff
@@ -0,0 +1,114 @@
...
+
+	@Test
+	void testBinary() {
+		ToCloudEventTransformer transformer =
+				new ToCloudEventTransformer(SOURCE, ToCloudEventTransformer.Result.BINARY);
+		transformer.setSubjectExpression(new LiteralExpression("some_subject"));
+		GenericMessage<String> message = new GenericMessage<>("test");
+		Message<?> result = transformer.transform(message);
+		assertThat(result.getHeaders())
+				.containsEntry("ce_type", String.class.getName())
```
n3wscott, line 80: note these all need to be `ce-*` not `ce_` if this is for http.

n3wscott:
Rather than implement this yourself, you should come help fix https://github.com/cloudevents/sdk-java

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/Marshallers.java
n3wscott: We only assume json if it is structured mode, and we pre-compute data into a byte array before attempting to create a structured message. 

for binary, we have a set of supported marshalers based on media type.

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
n3wscott: A generalized approach is not going to work if you also want to interoperate with other spec implementors, like the goland sdk over http will reject your messages as invalid

spring-integration-core/src/test/java/org/springframework/integration/transformer/ToCloudEventTransformerTests.java
artembilan: Well, one more time. What we do here is a general framework approach when you would like to build a `CloudEvent` and  carry it from component to component as our `Message<?>` content.
This is not about protocol binding per se.
Of course when we come with this content to the HTTP level we will remap these `ce_` to `ce-`.

I have designed this after this library general approach for target protocol header mapping, e.g. https://github.com/spring-projects/spring-integration/blob/master/spring-integration-http/src/main/java/org/springframework/integration/http/HttpHeaders.java#L26, https://github.com/spring-projects/spring-integration/blob/master/spring-integration-redis/src/main/java/org/springframework/integration/redis/support/RedisHeaders.java#L34 and so on.
Plus I have looked to CE Kafka protocol binder and this is that prefix: https://github.com/cloudevents/sdk-java/blob/master/kafka/src/main/java/io/cloudevents/v1/kafka/HeaderMapper.java#L25

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
artembilan: It won't because we will re-map our headers according agreed HTTP protocol binder.

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/Marshallers.java
artembilan: OK. I think we can give it a chance to be configurable.
That's an improvement we can do.

Thanks.

artembilan:
> Rather than implement this yourself, you should come help fix 

I'm not sure what you suggest us to fix over there since here we try to come up with some framework-specific tool which should help end-users to deal with CE well-know Spring Integration way.

Thank you for all your feedback!

Now we have something else to think about 😄

n3wscott:
ok it sounds like I misunderstood this, you are not implementing for HTTP, but for some custom Message<> type and you are free to use any method to add prefix (or no prefix if you don't need them).

In that case you might consider writing a small spec with this PR to allow others to also implement the same choices you are making for Message<>. We had to do this for PubSub to be spec compliant: https://github.com/google/knative-gcp/blob/master/docs/spec/pubsub-protocol-binding.md

And this now lets pubsub be linked from the CloudEvents spec: https://github.com/cloudevents/spec/blob/master/proprietary-specs.md

It would be pretty neat to have Spring in that list too!

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/HeaderMapper.java
```diff
@@ -0,0 +1,80 @@
...
+
+	/**
+	 * Following the signature of {@link io.cloudevents.fun.FormatHeaderMapper}
+	 * @param attributes The map of attributes
+	 * @param extensions The map of extensions
+	 * @return The map of headers
+	 */
+	public static Map<String, String> map(Map<String, String> attributes, Map<String, String> extensions) {
+		Assert.notNull(attributes, "'attributes' must noy be null");
+		Assert.notNull(extensions, "'extensions' must noy be null");
```
olegz, line 47: check spelling 'noy'

spring-integration-core/src/main/java/org/springframework/integration/support/cloudevents/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Provides classes supporting for Cloud Events.
```
olegz, line 2: 'supporting' to 'support'

artembilan:
Superseded with: https://github.com/spring-projects/spring-integration/pull/10448

PR #3248, opened 2020-04-13
GH-3247: Fix `SftpSession.exists` for error code

### LINKED ISSUE #3247: SftpSession.exists() does not throw exception
`SftpSession.exists()` method catches `SftpException` but ignores it. Why not throw a `NestedIOException` like other methods do?

In my code I use `exists` to check whether the remote file exists or not before I call `remove` to delete the file. In case when the logic inside `exists` throws an `IOException` (Pipe closed), it returns `false`, indicating the remote file does not exists.

Thanks,
Gary

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3247

When there is no path on the SFTP server, a `ChannelSftp.SSH_FX_NO_SUCH_FILE`
error is returned in the thrown `SftpException`.

* Fix `SftpSession.exists()` to check for the `SSH_FX_NO_SUCH_FILE` to
return `false` and re-throw an exception otherwise
* Add mock test for `SftpSession.exists()`
* Add `org.mockito.AdditionalMatchers` to `checkstyle.xml` exclusions

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


garyrussell (changes requested):
LGTM, but I think we need a test against the mina server, e.g. in `SftpServerOutboundTests` to verify we actually get that result from JSch.

garyrussell:
Probably not related but

```
> Task :spring-integration-sftp:test
SftpStreamingMessageSourceTests > testMaxFetchLambdaFilter FAILED
    org.springframework.messaging.MessagingException at SftpStreamingMessageSourceTests.java:160
        Caused by: org.springframework.integration.util.PoolItemNotAvailableException at SftpStreamingMessageSourceTests.java:160
            Caused by: java.lang.IllegalStateException at SftpStreamingMessageSourceTests.java:160
                Caused by: java.lang.IllegalStateException at SftpStreamingMessageSourceTests.java:160
                    Caused by: com.jcraft.jsch.JSchException at SftpStreamingMessageSourceTests.java:160
```

garyrussell (approved):

artembilan:
@garyrussell ,

It isn't related, indeed, but I can't figure out such a sporadic failure. Something is broken sometimes on Mina.
So, I have restarted the build and now it is green.

Feel free to merge!

Thanks

garyrussell:
Hmmm - I just cherry-picked this to 5.2 but while resolving conflicts in 5.1, I changed the exception to UncheckedIOException to avoid a breaking change. I will push that to the other branches as well.

Issue #3249, opened 2020-04-17
Deadlock when using SFTP CachingSessionFactory with FileExistsMode.FAIL


javierestevez:
**Affects 5.2.5**

It looks like there is a deadlock when using a `CachingSessionFactory` and `FileExistsMode.FAIL`. When `FileExistsMode.FAIL` is used, then we first check if the file exists in the server by calling `RemoteFileTemplate#exists` which in turn calls `RemoteFileTemplate#execute`. The latter requests a new session to the session factory instead of using the session already obtained (`RemoteFileTemplate#activeTemplateCallbacks` size always evaluated to 0 during my tests). This can end up in a deadlock situation if there are as many (or less) cached sessions as the number of concurrent threads (possibly with more too).

Below is a test I used to reproduce this locally using 2 concurrent threads and 2 cached sessions. I omitted some details for brevity. Hopefully it is enough to reproduce it on your side.

    private final Runnable runnable = () -> {
        UUID uuid = UUID.randomUUID();
        String filename = uuid + ".tst";
        LOGGER.info("Uploading file " + filename + " to SFTP...");

        testSftpGateway.send(DIRECTORY_NAME, filename, "Some bytes".getBytes());
        LOGGER.info("... Uploaded " + filename + "!");
    };

    @Test
    public void test() {
        withConfiguredSftpServer(sftpServer -> {
            ExecutorService executorService = Executors.newFixedThreadPool(2);
            for (int i = 0; i < 100; i++) {
                executorService.execute(runnable);
            }
            executorService.shutdown();
            executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS);
            LOGGER.info("Finished all threads");
        });
    }
    @MessagingGateway
    interface TestSftpGateway {
        @Gateway(requestChannel = OUTBOUND_CHANEL)
        void send(@Header(FileHeaders.REMOTE_DIRECTORY) String directoryName,
                  @Header(FileHeaders.FILENAME) String fileName,
                  @Payload byte[] payload);
    }

    @Configuration
    @IntegrationComponentScan
    @EnableIntegration
    static class TestConfiguration {

        @Bean
        public SessionFactory<ChannelSftp.LsEntry> sftpSessionFactory() {
            DefaultSftpSessionFactory factory = new DefaultSftpSessionFactory(true);
            factory.setHost("localhost");
            factory.setPort(22);
            factory.setUser("foo");
            factory.setPassword("pass");
[...]

PR #3250, opened 2020-04-19
GH-3249: Fix RemoteFileTemplate dead lock in send

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/3249

When the `CachingSessionFactory` is configured with small enough pool
and it is very likely that dead lock may happen when `RemoteFileTemplate.send()`
is used.
The problem happens when we reach the `RemoteFileTemplate.exists()` call
which is done from the internal method called from already pulled from cache
`Session`

* Fix `RemoteFileTemplate` to use a `session.exists()` instead on the provided
into the method `Session`
* Demonstrate the problem in the `SftpRemoteFileTemplateTests.testNoDeadLockOnSend()`

**Cherry-pick to 5.2.x, 5.1.x & 4.3.x**


artembilan:
Sorry, I have merged the fix into the `master` by mistake: https://github.com/spring-projects/spring-integration/commit/5ba1b9321f7de2737820241ba7bdc4fc4bd5a550

We can revert it or fix after review this PR for `5.2.x`.

It might not cherry-pick clean into `5.1.x` and definitely is not going to be compatible with `4.3.x`.

So, I'll do back-port when this is reviewed and merged.

Thanks

garyrussell (approved):

garyrussell:
Cherry-picked to 5.2.x; back ported to 5.1.x.

garyrussell:
And 4.3.x.

PR #3252, opened 2020-04-21
Add remaining deprecations for legacy metrics

garyrussell:
- add a module to suppress legacy metrics from the Integration Graph
- move graph tests package to match src/main


artembilan:
Missed some deprecation suppression (or fix), Gary:

```
/home/travis/build/spring-projects/spring-integration/spring-integration-core/src/test/java/org/springframework/integration/support/management/IntegrationManagementConfigurerTests.java:98: warning: [deprecation] DefaultMessageChannelMetrics in org.springframework.integration.support.management has been deprecated

				.isInstanceOf(DefaultMessageChannelMetrics.class);
```

Plus there is some Checkstyle violation.

Please, see Travis report.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -175,12 +174,14 @@ public void setLoggingEnabled(boolean loggingEnabled) {
 		this.managementOverrides.loggingConfigured = true;
 	}
 
-	protected AbstractMessageChannelMetrics getMetrics() {
+	protected org.springframework.integration.support.management.AbstractMessageChannelMetrics getMetrics() {
```
artembilan, line 177: Shouldn't methods like this be deprecated as well?
Kinda what is going to happen to them when we remove classes for their signature?..

spring-integration-core/src/main/java/org/springframework/integration/config/AggregatorFactoryBean.java
```diff
@@ -120,7 +120,8 @@ public void setOutputChannelName(String outputChannelName) {
 		this.outputChannelName = outputChannelName;
 	}
 
-	public void setMetrics(AbstractMessageHandlerMetrics metrics) {
+	@SuppressWarnings("deprecation")
+	public void setMetrics(org.springframework.integration.support.management.AbstractMessageHandlerMetrics metrics) {
```
artembilan, line 124: Maybe deprecated, too?

spring-integration-jmx/src/test/java/org/springframework/integration/jmx/config/MBeanExporterParserTests.java
```diff
@@ -133,8 +129,10 @@ private void checkCustomized(org.springframework.integration.support.management.
 	public static class CustomMetrics implements org.springframework.integration.support.management.MetricsFactory {
 
 		@Override
-		public AbstractMessageChannelMetrics createChannelMetrics(String name) {
-			return new DefaultMessageChannelMetrics(name,
+		public org.springframework.integration.support.management.AbstractMessageChannelMetrics
```
artembilan, line 132: Same here: can we just remove all the tests code with those deprecated classes and methods?

spring-integration-core/src/test/java/org/springframework/integration/graph/IntegrationGraphServerTests.java
```diff
@@ -152,10 +150,17 @@ void test() throws Exception {
...
+		String routerJson = jsonArray.toJSONString();
+		assertThat(routerJson).contains("\"deprecated\":\"stats are deprecated");
+
+
 		this.server.rebuild();
 		graph = this.server.getGraph();
 		baos = new ByteArrayOutputStream();
+		objectMapper = new ObjectMapper();
 		objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
+		objectMapper.registerModule(new GraphLegacyStatsNullModule());
```
artembilan, line 163: How about this instead:

```
mapper.registerModule(new SimpleModule().addSerializer(IntegrationNode.Stats.class, NullSerializer.instance));
```
?

So, we don't need to have any custom classes in the framework.

Or `MixInAnnotations`...

Or let them decide what is the best Jackson option to exclude the property:

https://stackoverflow.com/questions/7421474/how-can-i-tell-jackson-to-ignore-a-property-for-which-i-dont-have-control-over

spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationGraphServer.java
```diff
@@ -66,7 +66,7 @@
  */
 public class IntegrationGraphServer implements ApplicationContextAware, ApplicationListener<ContextRefreshedEvent> {
 
-	private static final float GRAPH_VERSION = 1.2f;
+	private static final float GRAPH_VERSION = 1.3f;
```
artembilan, line 69: I wonder what have been changed in the `Graph` hierarchy to make a decision to bump the version...

artembilan (changes requested):
That's really cool! Even if I have some concerns yet, I find this as really valuable change.

Any thought what (and when) we are going to do with the `MetricsCaptor` and brothers?
I thought we also don't need them any more if we fully rely on Micrometer...

Thanks

spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationGraphServer.java
garyrussell: Just because there is a possibility of `"stats": null` which wasn't possible before.

spring-integration-core/src/main/java/org/springframework/integration/config/AggregatorFactoryBean.java
garyrussell: Ugh - yes - there are lots of these 😦

spring-integration-core/src/main/java/org/springframework/integration/graph/IntegrationGraphServer.java
artembilan: I still think it is not a framework matter. We definitely haven't changed our model.
It is out of our control how Jackson is going to be configured in the target application which indeed may affect the JSON is produced outside...

garyrussell:
The `MetricsCaptor` can be removed after all this is pulled out.

artembilan:
👍 

> The MetricsCaptor can be removed after all this is pulled out.

I hope you talk about this one: https://github.com/spring-projects/spring-integration/issues/2804

garyrussell:
Correct.

artembilan (approved):
Merging after some polishing...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/c2babe4ba2227eed6bc34a9c72e68ac715cdf633

PR #3260, opened 2020-04-23
GH-3256: Support Testing TCP Connections

### LINKED ISSUE #3256: Create an interface Custom Health Check to validate the Connection health and manage connection
**Expected Behavior**

TCP Connection Factory is not having a Mechanism enabled through framework to perform Custom Health Check on a fixed interval and close the connection based on custom health check

**Current Behavior**

Currently the Connection Factory only establish the connection with the defined server and port and there is no option to validate the custom health check and close the connection based on custom health check.

https://stackoverflow.com/questions/61373546/with-tcp-failoverconnectionfactory-is-there-a-way-to-do-custom-health-check-and

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3256

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractClientConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractClientConnectionFactory.java
@@ -16,10 +16,13 @@
 
 package org.springframework.integration.ip.tcp.connection;
 
+import java.io.IOException;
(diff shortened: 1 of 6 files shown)


artembilan (approved):
I'm OK with the solution, but any comments about reporter request with the " Custom Health Check on a fixed interval "?

Thanks

garyrussell:
Hmmm - I didn't notice that - I was going on his original request; that can be done with a small SI flow; I'll add a comment to the issue.

PR #3261, opened 2020-04-23
GH-3253: Scan BF hierarchy for BeanDefinition

### LINKED ISSUE #3253: Spring Integration Java DSL stops working when used in a Spring Batch modular configuration
**Describe the bug**
Spring Integration using the DSL produces a `NoSuchBeanDefinitionException:` No bean named `'javaDslInChildContextFlow.channel#0' available`, when the flow is configured within a child `ApplicationContext` configured in a modular Spring Batch setup.

The problem appears between Spring Integration releases `5.1.1` and `5.1.2` and is still present in the latest release.

A secondary observation is that the child `ApplicationContext` requires `@EnableIntegration` on it's `@Configuration` class, whereas previously it would work when specified on just the `@SpringBootApplication` class.

**To Reproduce**
Add an `IntegrationFlow` with the Java DSL to a Spring Batch child `ApplicationContext` while using a version of Spring Integration `5.1.2` or greater.

**Expected behavior**
The Integration flow is created/instantiated correctly.

**Sample**

Here's a link to a [minimal, reproducible sample](https://github.com/garethpowell/spring-integration-batch-problem).

There's a branch calling `working-version` that shows the same code working with the Spring Integration version downgraded to `5.1.1`

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3253

The `IntegrationFlowBeanPostProcessor` uses a `containsBean()`
and then `getBeanDefinition()` to be sure that we don't override already
existing bean even if it is created in the parent(s) context.
The problem that `containsBean()` check the hierarchy, but `getBeanDefinition()`
doesn't.
So, we fail with `NoSuchBeanDefinitionException` if bean exists in the parent ctx

* Introduce an utility `IntegrationContextUtils.getBeanDefinition()` to
scan `BeanFactory` recursively for `BeanDefinition` for the requested `name`
* Use this tool in the `IntegrationFlowBeanPostProcessor` logic

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,9 @@ package org.springframework.integration.context;
 import java.util.Properties;
 
 import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.expression.spel.support.SimpleEvaluationContext;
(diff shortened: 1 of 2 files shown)


artembilan:
The fix has been tested in the provided sample application.

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -212,4 +215,26 @@ public static Properties getIntegrationProperties(BeanFactory beanFactory) {
...
+	 * @throws NoSuchBeanDefinitionException if a {@link BeanDefinition} is not found
+	 * @since 5.1.10
+	 */
+	public static BeanDefinition getBeanDefinition(String name, ConfigurableListableBeanFactory beanFactory) {
+		try {
+			return beanFactory.getBeanDefinition(name);
+		}
+		catch (NoSuchBeanDefinitionException ex) {
+			BeanFactory parentBeanFactory = beanFactory.getParentBeanFactory();
+			if (parentBeanFactory instanceof ConfigurableListableBeanFactory) {
```
garyrussell, line 233: Why just one parent? Don't we need to recurse up the BF tree?

artembilan: Because the next `return getBeanDefinition(name, (ConfigurableListableBeanFactory) parentBeanFactory);` call is indeed recursive.

Not sure what am I missing...

garyrussell: Doh - not enough coffee.

garyrussell:
Cherry-picked to 5.2.x and to 5.1.x after resolving conflicts.

PR #3264, opened 2020-04-24
Upgrade to latest version of Asciidoctor

Buzzardo:
Upgrade to latest version of Asciidoctor because we were several
versions behind and were steadily incurring more technical debt.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,9 +16,9 @@ plugins {
 	id "io.spring.dependency-management" version '1.0.9.RELEASE'
 	id 'com.jfrog.artifactory' version '4.15.1' apply false
 	id 'org.jetbrains.dokka' version '0.10.1'
-	id 'org.asciidoctor.jvm.pdf' version '3.1.0'
-	id 'org.asciidoctor.jvm.gems' version '3.1.0'
-	id 'org.asciidoctor.jvm.convert' version '3.1.0'
+	id 'org.asciidoctor.jvm.pdf'
+	id 'org.asciidoctor.jvm.gems'
+	id 'org.asciidoctor.jvm.convert'
 }
 
 if (System.getenv('TRAVIS') || System.getenv('bamboo_buildKey')) {
--- a/build.gradle
+++ b/build.gradle
@@ -16,9 +16,9 @@ plugins {
 	id "io.spring.dependency-management" version '1.0.9.RELEASE'
 	id 'com.jfrog.artifactory' version '4.15.1' apply false
 	id 'org.jetbrains.dokka' version '0.10.1'
-	id 'org.asciidoctor.jvm.pdf'
-	id 'org.asciidoctor.jvm.gems'
-	id 'org.asciidoctor.jvm.convert'
+	id 'org.asciidoctor.jvm.pdf' version '3.1.0'
+	id 'org.asciidoctor.jvm.gems' version '3.1.0'
+	id 'org.asciidoctor.jvm.convert' version '3.1.0'
 }
 
 if (System.getenv('TRAVIS') || System.getenv('bamboo_buildKey')) {
--- a/build.gradle
+++ b/build.gradle
@@ -2,8 +2,6 @@ buildscript {
 	ext.kotlinVersion = '1.3.71'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
-		jcenter()
-		mavenCentral()
 	}
 	dependencies {
 		classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -926,7 +926,8 @@ asciidoctorPdf {
 		'sectnums': '',
 		'toc': '',
 		'source-highlighter' : 'coderay',
-		revnumber: project.version
+		revnumber: project.version,
+		'project-version': project.version
 	}
 }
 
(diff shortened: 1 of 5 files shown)


src/reference/asciidoc/index-header.adoc
```diff
@@ -3,7 +3,7 @@
 Mark Fisher; Marius Bogoevici; Iwein Fuld; Jonas Partner; Oleg Zhurakousky; Gary Russell; Dave Syer; Josh Long; David Turanski; Gunnar Hillert; Artem Bilan; Amol Nayak; Jay Bryant
 
 ifdef::backend-html5[]
-*{project-version}*
```
artembilan, line 6: So, looks like `project-version` has gone.

Please, revise all our docs then to replace this one with a new `revnumber`.

For example right now now I see this:
```
<version>{project-version}</version>
```
Or this `compile "org.springframework.integration:spring-integration-amqp:{project-version}"`

Has to be an actual project version we built these docs against.

artembilan (changes requested):
Just only one version placeholder concern! 👍

build.gradle
```diff
@@ -2,22 +2,25 @@ buildscript {
 	ext.kotlinVersion = '1.3.71'
 	repositories {
 		maven { url 'https://repo.spring.io/plugins-release' }
+		jcenter()
+		mavenCentral()
```
artembilan, line 6: Sorry, forgot to ask: do we really need these extra repos?

Buzzardo: Yes, we do. I tried it without them, but it doesn't work. Those repos are where the asciidoctor and asciidoctorPdf tasks come from, so the build breaks without them. I share your pain here: It seems goofy to add those.

src/reference/asciidoc/index-header.adoc
Buzzardo: I added a project-version attribute back into the asciidoctor and asciidoctorPdf tasks. That fixed it.

build.gradle
```diff
@@ -6,18 +6,19 @@ buildscript {
...
 	id 'org.sonarqube' version '2.8'
-	id 'org.asciidoctor.convert' version '1.6.1'
 	id 'io.spring.nohttp' version '0.0.4.RELEASE' apply false
 	id 'org.ajoberstar.grgit' version '4.0.1'
 	id "io.spring.dependency-management" version '1.0.9.RELEASE'
 	id 'com.jfrog.artifactory' version '4.15.1' apply false
 	id 'org.jetbrains.dokka' version '0.10.1'
+	id 'org.asciidoctor.jvm.pdf' version '3.1.0'
+	id 'org.asciidoctor.jvm.gems' version '3.1.0'
+	id 'org.asciidoctor.jvm.convert' version '3.1.0'
```
artembilan, line 21: Can we extract this as a `asciidoctorVersion`?
Alongside with that `ext.kotlinVersion = '1.3.71'` above.

Thanks

artembilan (changes requested):

artembilan (changes requested):
@Buzzardo , while we are here, would you mind to fix a `mongodb.adoc` for the `reactive-streams.adoc/reactive-message-handler` typo? Has to be `reactive-streams.adoc#reactive-message-handler`.

Nice to have this upgrade: it has caught this issue! 😄

build.gradle
Buzzardo: I don't think the asciidoctor folks are trying to keep them in sync. I think it's a coincidence that they are the same just now. We can do it, but we may end up with three variables down the road. I'll go ahead and do it, though, since it seems to be Spring's usual way of doing things.

gradle.properties
```diff
@@ -2,3 +2,4 @@ version=5.3.0.BUILD-SNAPSHOT
 org.gradle.jvmargs=-Xmx1536m -Dfile.encoding=UTF-8
 org.gradle.caching=true
 org.gradle.parallel=true
+asciidoctorVersion=3.1.0
```
artembilan, line 5: No, I kinda agree with you about different versions in the future.

Also I don't like the change spreaded to several files which are fully not related to docs.
Can we, please, revert this commit?

artembilan (changes requested):

src/reference/asciidoc/mongodb.adoc
```diff
@@ -548,7 +548,7 @@ Starting with version 5.3, the `ReactiveMongoDbStoringMessageHandler` and `React
 They are based on the `ReactiveMongoOperations` from Spring Data and requires a `org.mongodb:mongodb-driver-reactivestreams` dependency.
 
 The `ReactiveMongoDbStoringMessageHandler` is an implementation of the `ReactiveMessageHandler` which is supported natively in the framework when reactive streams composition is involved in the integration flow definition.
-See more information in the <<./reactive-streams.adoc/reactive-message-handler,ReactiveMessageHandler>>.
+See more information in the <<./reactive-streams.adoc/#reactive-message-handler,ReactiveMessageHandler>>.
```
artembilan, line 551: I think this is still wrong.
Must be:
```suggestion
See more information in the <<./reactive-streams.adoc#reactive-message-handler,ReactiveMessageHandler>>.
```

gradle.properties
Buzzardo: I didn't like it rolling into three files, either. I'll revert it.

artembilan (changes requested):

src/reference/asciidoc/mongodb.adoc
Buzzardo: Got it this time.

PR #3265, opened 2020-04-28
Introduce a `ReceiveMessageAdvice`

artembilan:
* Deprecate an `AbstractMessageSourceAdvice` in favor of
`default` method in the `MessageSourceMutator`
* Move a `applyReceiveOnlyAdviceChain()` logic into the `AbstractPollingEndpoint`:
now both `PollingConsumer` and `SourcePollingChannelAdapter` can use
`ReceiveMessageAdvice`
* Introduce a `SimpleActiveIdleReceiveMessageAdvice` based already
on the `ReceiveMessageAdvice` and deprecate a `SimpleActiveIdleMessageSourceAdvice`
which is fully replaceable with newly introduced `SimpleActiveIdleReceiveMessageAdvice`
* Add `@SuppressWarnings("deprecation")` for those out-of-the-box `ReceiveMessageAdvice`
implementation which still use an `AbstractMessageSourceAdvice` for
backward compatibility
* Document a new feature and give the `MessageSourceMutator` a new meaning

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aop/CompoundTriggerAdvice.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aop/CompoundTriggerAdvice.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package org.springframework.integration.aop;
 
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.util.CompoundTrigger;
+import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.scheduling.Trigger;
 import org.springframework.util.Assert;
@@ -35,7 +36,10 @@ import org.springframework.util.Assert;
  * @since 4.3
  *
  */
-public class CompoundTriggerAdvice extends AbstractMessageSourceAdvice {
+@SuppressWarnings("deprecation")
+public class CompoundTriggerAdvice
+		extends AbstractMessageSourceAdvice
+		implements ReceiveMessageAdvice {
 
 	private final CompoundTrigger compoundTrigger;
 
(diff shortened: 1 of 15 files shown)


src/reference/asciidoc/polling-consumer.adoc
```diff
@@ -145,26 +145,29 @@ These "`around advice`" methods do not have access to any context for the poll -
 This is fine for requirements such as making a task transactional or skipping a poll due to some external condition, as discussed earlier.
 What if we wish to take some action depending on the result of the `receive` part of the poll or if we want to adjust the poller depending on conditions? For those instances, Spring Integration offers "`Smart`" Polling.
 
+[[smart-polling]]
 ===== "`Smart`" Polling
 
-Version 4.2 introduced the `AbstractMessageSourceAdvice`.
-Any `Advice` objects in the `advice-chain` that subclass this class are applied only to the receive operation.
+Version 5.3 introduced the `ReceiveMessageAdvice` interface.
+(An `AbstractMessageSourceAdvice` has been deprecated in favor of `default` methods in the `MessageSourceMutator`.)
```
garyrussell, line 152: (The

spring-integration-core/src/main/java/org/springframework/integration/aop/CompoundTriggerAdvice.java
```diff
@@ -47,8 +51,21 @@ public CompoundTriggerAdvice(CompoundTrigger compoundTrigger, Trigger overrideTr
...
+	 */
 	@Override
+	@Deprecated
 	public Message<?> afterReceive(Message<?> result, MessageSource<?> source) {
+		return afterReceive(result, (Object) source);
+	}
+
+	@Override
+	@Nullable
+	public Message<?> afterReceive(@Nullable Message<?> result, Object source) {
```
garyrussell, line 68: Wouldn't something like `Advisable<T>` be better than `Object`?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/aop/CompoundTriggerAdvice.java
artembilan: Well, in most cases it is really about both `MessageSource` and `PollableChannel`, so even with generic argument it still going to be an `Object` 😄 

I thought about a `Supplier` for both those interfaces, but this is not what we are going to use with the `receive(timeout)`.
Therefore I'm OK to stick with an `Object` for a `source` argument.

src/reference/asciidoc/polling-consumer.adoc
artembilan, line 152:
```suggestion
(The `AbstractMessageSourceAdvice` has been deprecated in favor of `default` methods in the `MessageSourceMutator`.)
```

PR #3267, opened 2020-04-28
GH-3266: Don't override props in AnnGWProxyFB

### LINKED ISSUE #3266: The AnnotationGatewayProxyFactoryBean overrides provided properties into default values in its onInit() implementation
See for more info: https://github.com/spring-cloud/spring-cloud-stream/issues/1948

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3266

It turns out that `AnnotationGatewayProxyFactoryBean.onInit()` implementation
parses a `@MessagingGateway` attributes ignoring possible properties
population by setters.
This way a Java DSL `GatewayProxySpec` becomes useless since all
its options are overridden by default values from a synthesized
`@MessagingGateway`.
Also it is inconsistency when we declare an `AnnotationGatewayProxyFactoryBean`
as regular bean, but then called setters are ignored

* Add `protected` getters into `GatewayProxyFactoryBean` for all
the properties which can be overridden by annotation attributes
* Fix `AnnotationGatewayProxyFactoryBean` to consult with those getters
before populating a property with value from the annotation

**Cherry-pick to 5.2.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/AnnotationGatewayProxyFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/AnnotationGatewayProxyFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


artembilan:
Well, let's back-port this to `5.1.x` as well! The problem with `AnnotationGatewayProxyFactoryBean` is still there even if it is not used in DSL.
(I guess the test won't be cherry-picked clearly - so maybe no back-porting for it 😄 )

garyrussell (commented):
Dumb question, perhaps - why is the DSL using the annotation FB instead, given that all the properties it sets is on the super class?

The annotation subclass should only be used for the, errr, annotation.

I don't see why this needs to be back-ported.

artembilan:
> why is the DSL using the annotation FB instead, given that all the properties it sets is on the super class?

Hm. I think the idea was to let to mark a provided interface with a `@MessagingGateway` and override only those properties in the flow which cannot be expressed through annotations.
I mean for readability it looks better to have it like this in the DSL `IntegrationFlows.from(MyInterface.class)` and separate gateway configuration into a `@MessagingGateway` on that interface.

I might agree with you in the past, but now when the cat is out of the bag, I don't think it is too bad to let end-user to combine annotation with some explicit setters invocation.

It has to be back-ported because currently the `AnnotationGatewayProxyFactoryBean` behavior is definitely not what users could expect.

Does it make sense?

Thanks

garyrussell:
>Does it make sense?

Your comment about the DSL does, yes, but I still don't see why a back-port is needed.

Why would someone add an `AnnotationGatewayProxyFactoryBean` as a bean?

artembilan:
> Why would someone add an AnnotationGatewayProxyFactoryBean as a bean?

OK. Let's just stop then with `5.2.x` back-porting when we introduced this Java DSL feature!

garyrussell:
Cherry-picked to 5.2.x after resolving conficts.

PR #3269, opened 2020-04-29
BoundRabbitChannelAdvice - reject invalid config

garyrussell:
Don't allow a `waitForConfirmsTimeout` if the factory is not configured
for simple publisher confirmations.

Otherwise, a runtime error will occur.

**cherry-pick to 5.2.x, 5.1.x**


artembilan:
... and cherry-picked to `5.2.x` & `5.1.x`

PR #3274, opened 2020-05-11
GH-3271: Close session on error in stream source

### LINKED ISSUE #3271: Unreleased cached sessions when receiving file fails
**Affects 5.2.5.RELEASE**

I have an inbound flow that reads files from a SFTP server using a `CachingSessionFactory` with size 1. I had a very strange situation where files were removed from the SFTP server after the polling occurred but before the message was built in `AbstractRemoteFileStreamingMessageSource#doReceive`. The call to `session.readRaw(remotePath)` failed with a `SftpException: No such file or directory` from `jsch` wrapped in a NestedIOException by `SftpSession`, which was caught and rethrown.

As a result, the session that was acquired to read the payload was not released back to the pool, and further attempts to obtain a session from the pool failed with:
```
org.springframework.messaging.MessagingException: nested exception is org.springframework.integration.util.PoolItemNotAvailableException: Timed out while waiting to acquire a pool entry.
```

I believe this is a bug and the session should be returned back to the pool in case of failure.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3271

When exception happens at `.withPayload(session.readRaw(remotePath))`
in the `AbstractRemoteFileStreamingMessageSource` we don't close session.
The resource leaking happens in the caching session factory

* Add `session.close();` into the `catch (IOException e) {`
in the `AbstractRemoteFileStreamingMessageSource.doReceive()`
to clean up resources properly

**Cherry-pick to 5.2.x, 5.1.x & 4.3.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
@@ -216,6 +216,7 @@ public abstract class AbstractRemoteFileStreamingMessageSource<F>
 									this.fileInfoJson ? file.toJson() : file);
 				}
 				catch (IOException e) {
+					session.close();
 					throw new UncheckedIOException("IOException when retrieving " + remotePath, e);
 				}
 			}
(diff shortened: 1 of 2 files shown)


garyrussell:
cherry-picked to 5.2.x, 5.1.x, back ported to 4.3.x

PR #3277, opened 2020-05-13
Fix XML examples in handler-advice reference doc

ykartsev:
1. Add proper terminating tags for `request-handler-advice-chain`.
2. Replace `int:request-handler-advice-chain` with `int-http:request-handler-advice-chain` when used inside of an `int-http:outbound-gateway`.
3. Replace `<bean ref="*" />` with `<ref bean="*" />`
4. Replace `</int:int:service-activator>` with `</int:service-activator>`


artembilan:
Nice catches, @ykartsev !

Thank you very much; looking forward for more!

PR #3278, opened 2020-05-13
Fix RESOLVABLE_TYPE header population

artembilan:
* We should not build a RESOLVABLE_TYPE header when we map requests.
Only for replies. See: https://github.com/spring-projects/spring-integration-samples/issues/277
* We should log `ClassNotFoundException` only at debug level - to noise with info or warn

**Cherry-pick to 5.2.x**


garyrussell:
Cherry-picked as 7cfab7b9a874e73929929cbec3bb691f273ea288

PR #3279, opened 2020-05-13
GH-3276: reactive inbound: Fix `onErrorResume`

### LINKED ISSUE #3276: Reactive channel adapter not using error channel
**Description**
I was trying to use `WebFlux.inboundGateway()` and use a custom error channel with `.errorChannel(customErrorChannel())`. I need this to prepare custom error response. But could not make the error channel to work and was getting default framework response for errors.

**Code**
I tried like the following
```
@Bean
public IntegrationFlow mainFlow() {
    return IntegrationFlows.from(WebFlux.inboundGateway(path)
            .errorChannel(customErrorChannel())
        )
        .transform(p -> {
            if(<some validation fails>)
                throw new RuntimeException("Error!");
            return "Ok Response";
        })
        .get();
}

@Bean
public PublishSubscribeChannel customErrorChannel() {
    return MessageChannels.publishSubscribe().get();
}

@Bean
public IntegrationFlow errorFlow() {
    return IntegrationFlows.from("customErrorChannel")
        .transform(p -> {
            return "Error Response";
        })
        .get();
}
```

**Link**
SO thread: https://stackoverflow.com/questions/61718691/how-to-use-custom-error-channel-to-produce-custom-error-response

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3276

The `onErrorResume` for the `MessagingGatewaySupport.doSendAndReceiveMessageReactive()`
was in wrong place: only for the `buildReplyMono` which works only
when an outbound flow is fully based on reactive channels.
With a regular direct channel we can get an exception from the
`sendMessageForReactiveFlow` which is not covered with the mentioned
`onErrorResume` for the error handling on the configured `errorChannel`

Cherry-pick to `5.2.x & 5.1.x`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
@@ -643,7 +643,8 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 			sendMessageForReactiveFlow(requestChannel, requestMessage);
 
 			return buildReplyMono(requestMessage, replyChan.replyMono, error, originalReplyChannelHeader,
-					originalErrorChannelHeader);
+					originalErrorChannelHeader)
+					.onErrorResume(t -> error ? Mono.error(t) : handleSendError(requestMessage, t));
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


garyrussell:
Cherry-picked to 5.2.x; conflicts on 5.1.x; please review and backport.

PR #3285, opened 2020-05-19
GH-3282: Fix JdbcMetadataStore for DuplicateKeyEx

### LINKED ISSUE #3282: DuplicateKeyException raised when using Spring JdbcMetadataStore (Oracle)
As discussed via [StackOverFlow](https://stackoverflow.com/questions/61747197/duplicatekeyexception-raised-when-using-spring-jdbcmetadatastore-oracle) website. It looks like `JdbcMetadataStore.putIfAbsent()` should be improved to catch `SQLIntegrityConstraintViolationException` as a fact of data presence and return the current value.

Issue:

My current configuration of `file:inbound-channel-adapter` is working fine. I'm running the application in multiple servers and only one is actually processing the file. Database table `INT_METADATA_STORE` also updated successfully. Issue I'm facing is that one of the servers are still trying to insert the record and below exception is raised:

```
org.springframework.messaging.MessagingException: nested exception is org.springframework.dao.DuplicateKeyException: PreparedStatementCallback; SQL [INSERT INTO INT_METADATA_STORE(METADATA_KEY, METADATA_VALUE, REGION) SELECT ?, ?, ? FROM INT_METADATA_STORE WHERE METADATA_KEY=? AND REGION=? HAVING COUNT(*)=0]; ORA-00001: unique constraint (SMS_OWNER.INT_METADATA_STORE_PK) violated; nested exception is java.sql.SQLIntegrityConstraintViolationException: ORA-00001: unique constraint (SMS_OWNER.INT_METADATA_STORE_PK) violated 
```

I've tried different flavors of isolation and no luck. 

```
<int-file:inbound-channel-adapter directory="file:/tmp/input/" prevent-duplicates="true" filter="compositeFileListFilter">
  <int:poller max-messages-per-poll="1" cron="*/10 * * * * *">
    <int:transactional transaction-manager="transactionManager" isolation="READ_COMMITTED" timeout="5" />
  </int:poller>
</int-file:inbound-channel-adapter>
```

and

```
@Configuration
@EnableTransactionManagement
public class MetadataStoreConfiguration {

    @Value("${input.file.pattern:(DUMMY)(_).*\\.(xml)}")
    private String pattern;

    @Bean
    @Qualifier("fileSystemPersistentAcceptOnceFileListFilter")
    public FileSystemPersistentAcceptOnceFileListFilter fileSystemPersistentAcceptOnceFileListFilter(final DataSource dataSource) {
        return new FileSystemPersistentAcceptOnceFileListFilter(metadataStore(dataSource),"");
    }

    @Bean
    @Qualifier("metadataStore")
    public JdbcMetadataStore metadataStore(final DataSource dataSource) {
        JdbcMetadataStore metadataStore = new JdbcMetadataStore(dataSource);
        return metadataStore;
    }

    @Bean 
    public CompositeFileListFilter<File> compositeFileListFilter(FileSystemPersistentAcceptOnceFileListFilter fileSystemPersistentAcceptOnceFileListFilter) {       
        CompositeFileListFilter<File> filter = new CompositeFileListFilter<>(Arrays.asList(fileSystemPersistentAcceptOnceFileListFilter, new RegexPatternFileListFilter(pattern)));
        return filter;
    }

    @Bean
    @Primary
    public PlatformTransactionManager transactionManager(final DataSource dataSource) {
        return new DataSourceTransactionManager(dataSource);
    }

}
```

I'm using Spring Boot 2.2.4.RELEASE, Java 8 and Oracle as database.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/3282

The `INSERT INTO ... SELECT ... FROM ... HAVING` may fail with
`DuplicateKeyException` in between transactions.
* Catch `DuplicateKeyException` from in the `tryToPutIfAbsent()`
and return `0` as a fact of not inserted to let other logic to work
as expected.

There is no test coverage for this since it is almost impossible to
reproduce such a race condition on DB

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
@@ -19,6 +19,7 @@ package org.springframework.integration.jdbc.metadata;
 import javax.sql.DataSource;
 
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.dao.DuplicateKeyException;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.integration.metadata.ConcurrentMetadataStore;
 import org.springframework.jdbc.core.JdbcOperations;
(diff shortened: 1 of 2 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -142,7 +147,7 @@ public String putIfAbsent(String key, String value) {
 			//try to insert if does not exists
 			int affectedRows = tryToPutIfAbsent(key, value);
 			if (affectedRows > 0) {
-				//it was not in the table, so we have just inserted it
+				//it was not in the table, so we have just inserted it or DuplicateKeyException has been thrown
```
garyrussell, line 150: ? This comment seems wrong; we return 0 on the DKE and go to the else block.

garyrussell (changes requested):

garyrussell:
And cherry-picked to 5.3.x, 5.2.x

Issue #3286, opened 2020-05-20
Consider making FileTailingEvent#getMessage() method public


Toparvion:
**Expected Behavior**

[`FileTailingMessageProducerSupport.FileTailingEvent#getMessage`](https://github.com/spring-projects/spring-integration/blob/75021beae4e1806bee03563628a94f964893eb13/spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java#L221) method might be `public` instead of `protected` to allow application code be aware of what is happening with a file being `tail`ed.

**Current Behavior**

Currently, when handling `FileTailingEvent`s, application listeners have no access to the message produced by the underlying tail implementation. Therefore they cannot distinguish one file event type from another (e.g. absence from rotating) thus making the handling inefficient.

**Context**

I realize that it is generally not a good idea to rely on the text of the event as it comes from the underlying tailing provider (such as Apache Commons or OS Delegate) and thus can vary greatly depending on the platform. Nonetheless, there are some use cases where the message is still needed, for example:
* to handle the event appropriately if the tail implementation is fixed and is known beforehand
* to build some 'heuristic' around the message to cover different implementations where needed
* just to inform the end user (`toString()` method is too verbose for that)

The only alternative I see at the moment is to parse the `toString()` method output but it seems irrational and brittle.

If there is no security/compatibility/etc reason for keeping the method `protected`, I would submit a pull request making it `public` with the corresponding javadoc warning about its changeable nature.

garyrussell:
I am sure this was just a simple mistake (probably some IDE auto-created getter); I can't believe it has taken 7 years for this to be reported.

artembilan:
@Toparvion ,
please, consider to raise a simple PR!

Thanks

Toparvion:
@artembilan , @garyrussell , thank you for the welcoming reply!
I've submitted the [pull request](https://github.com/spring-projects/spring-integration/pull/3287).

PR #3287, opened 2020-05-21
GH-3286: Make FileTailingEvent#getMessage method public

Toparvion:
Fixes `spring-projects/spring-integration#3286`.

This simple change makes it possible for application code to retrieve the original text messages emitted from tailing producers.


artembilan (changes requested):
Please, run `gradlew clean :spring-integration-file:check` go see all the code checkstyle violations.

Toparvion:
@artembilan , yes, there were some trailing spaces (in javadoc) added by my IDE that I didn't noticed when run just `gradle test` instead of `gradle check` command. Now should be OK.

PR #3291, opened 2020-05-27
GH-3288: Add Kotlin DSL `transform(Transformer)`

### LINKED ISSUE #3288: Kotlin dsl- add transform methd for GenericTransformer
Add ability to kotlin dsl to add a generic transformer to a flow without specifiing method to be called.

When i wanted to use UnZipTransformer() i added it to a flow:
```
...
transform(
    UnZipTransformer()
)
...
```
It ended up with error:
```
java.lang.IllegalArgumentException: Found ambiguous parameter type [class java.lang.Boolean] for method match: [public void org.springframework.integration.zip.transformer.AbstractZipTransformer.setZipResultType(org.springframework.integration.zip.transformer.ZipResultType)
```
I had to specify exacltly what method of UnZipTransformer to call one way or another. 

It would be nice to use Transformer without specifying explicitly to use its transform method as is possible using java dsl.

I am creating this ticket following gitter  discussion with @artembilan

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3288

* For better end-user experience with Kotlin DSL and get a gain from
existing `Transformer` implementations add a `transform(Transformer)`
EI-method into the `KotlinIntegrationFlowDefinition`
* Also add `filter(MessageSelector)` for any out-of-the-box `MessageSelector`

**Cherry-pick to 5.3.x**


garyrussell:
And cherry-picked.

PR #3292, opened 2020-05-27
GH-3280: NullChannel as reply for void gateways

### LINKED ISSUE #3280: Gateway: evaluate setting a NullChannel replyChannel in case of void return type method
**Expected Behavior**

The full discussion is at: https://stackoverflow.com/questions/61756786/spring-integration-gateway-with-void-method-return-type-why-doesnt-it-set-a-nu

Substantially, when I set up a gateway method with void return type, I expect that any reply message that is produced by the downstream chain is simply ignored, and that I can rather opt-in for an exception if an unexpected (?) reply message is produced.

**Current Behavior**

If I understand it well, right now a gateway method with void return type only works (with no exception) if the downstream chain ends with an outbound adapter or if I explicitly set a `NullChannel` `replyChannel` somewhere in the chain (perhaps with an annotation on the gateway method). Otherwise an exception is produced:

    org.springframework.messaging.core.DestinationResolutionException: no output-channel or replyChannel header available

In my case I have a gateway interface with many methods and a downstream HTTP outbound gateway that invokes actions on a remote service through REST. The REST API provides some actions with a response, some others with no meaningful response (apart from a status code which is used to correctly wire errors).
On my gateway interface I would expect that a method with a void return type is enough to tell Spring Integration that I don't care about any (unmeaningful) response produced by the HTTP outbound gateway, but this is not the case and an exception is produced. I have to annotate my interface method by specifying a `MessageHeaders.REPLY_CHANNEL` header with `nullChannel` name or `@nullChannel` expression, or rather set this header somewhere else in my downstream chain. But this sounds redundant to me.

By debugging I see that `org.springframework.integration.gateway.GatewayProxyFactoryBean.invokeGatewayMethod(MethodInvocation, boolean)` does a "sendAndReceive" when the gateway method is expected to return something, otherwise a simple "send". The former call causes an appropriate `replyChannel` header to be set on the produced message (sorry, I'm not sure on how this is done, probably by `org.springframework.integration.gateway.MessagingGatewaySupport.registerReplyMessageCorrelator()`?),  while the latter doesn't cause any `replyChannel` header to be set and hence the exception occurs if a reply message arrives from the downstream chain.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3280

To properly support a one-way gateway (`void` return type), it is
better to ignore any possible replies from downstream instead of
unexpected `no output-channel or replyChannel header available`
error

* Populate a `nullChannel` into a `replyChannel` header
for `void` gateway methods when `replyChannel` is not set explicitly
* Remove redundant `GatewayProxyFactoryBean.PARSER` in favor of
similar `EXPRESSION_PARSER` in the super class

### DIFF
--- a/src/reference/asciidoc/gateway.adoc
+++ b/src/reference/asciidoc/gateway.adoc
@@ -83,7 +83,7 @@ You might also want to explicitly provide a reply channel for monitoring or audi
 To configure a channel interceptor, you need a named channel.
 
 NOTE: Starting with version 5.4, when gateway method return type is `void`, the framework populates a `replyChannel` header as a `nullChannel` bean reference if such a header is not provided explicitly.
-Such a behavior allows to ignore possible reply from downstream flow meeting a one-way gateway contract.
+This allows any possible reply from the downstream flow to be discarded, meeting the one-way gateway contract.
 
 [[gateway-configuration-annotations]]
 ==== Gateway Configuration with Annotations and XML
--- a/src/reference/asciidoc/gateway.adoc
+++ b/src/reference/asciidoc/gateway.adoc
@@ -82,6 +82,9 @@ The gateway creates a bridge from it to the temporary, anonymous reply channel t
 You might also want to explicitly provide a reply channel for monitoring or auditing through an interceptor (for example, <<./channel.adoc#channel-wiretap,wiretap>>).
 To configure a channel interceptor, you need a named channel.
 
+NOTE: Starting with version 5.4, when gateway method return type is `void`, the framework populates a `replyChannel` header as a `nullChannel` bean reference if such a header is not provided explicitly.
+Such a behavior allows to ignore possible reply from downstream flow meeting a one-way gateway contract.
+
 [[gateway-configuration-annotations]]
 ==== Gateway Configuration with Annotations and XML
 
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/whats-new.adoc
```diff
@@ -12,3 +12,11 @@ If you are interested in the changes and features that were introduced in earlie
...
 If you are interested in more details, see the Issue Tracker tickets that were resolved as part of the 5.4 development process.
 
+[[x5.4-new-components]]
+=== New Components
+
+[[x5.4-general]]
+=== General Changes
+
+The one-way messaging gateway (the `void` method return type) now sets a `nullChannel` explicitly into the `replyChannel` header to ignore any possible downstream replies.
+See for more information in the <<./gateway.adoc#gateway-default-reply-channel,Setting the Default Reply Channel>>.
```
garyrussell, line 22:
```suggestion
See <<./gateway.adoc#gateway-default-reply-channel,Setting the Default Reply Channel>> for more information.
```

src/reference/asciidoc/gateway.adoc
```diff
@@ -82,6 +82,9 @@ The gateway creates a bridge from it to the temporary, anonymous reply channel t
 You might also want to explicitly provide a reply channel for monitoring or auditing through an interceptor (for example, <<./channel.adoc#channel-wiretap,wiretap>>).
 To configure a channel interceptor, you need a named channel.
 
+NOTE: Starting with version 5.4, when gateway method return type is `void`, the framework populates a `replyChannel` header as a `nullChannel` bean reference if such a header is not provided explicitly.
+Such a behavior allows to ignore possible reply from downstream flow meeting a one-way gateway contract.
```
garyrussell, line 86:
```suggestion
This allows any possible reply from the downstream flow to be discarded, meeting the one-way gateway contract.
```

garyrussell (changes requested):

PR #3293, opened 2020-06-01
GH-2736: TCP Async GW - Fix async check

garyrussell:
Checking for NIO cannot be done in `onInit()` because the factory is
not yet started.

**cherry-pick to 5.3.x**


artembilan:
... and cherry-picked to `5.3.x`

PR #3296, opened 2020-06-09
Fix NPE for component name after FactoryBeans

artembilan:
SO: https://stackoverflow.com/questions/62190244/spring-integration-upgrade-from-5-2-x-to-5-3-problem

When the component is populated to the application context via `FactoryBean`,
all its `BeanFactory` callbacks should be propagated from that `FactoryBean` -
only lifecycle control for this component we have through its `FactoryBean`

**Cherry-pick to 5.3.x**


garyrussell:
Cherry-picked to 5.3.x

Issue #3307, opened 2020-06-12
TransactionTimedOutException during DefaultLockRepository.acquire()


astrubel:
Hi,
I'm facing to `TransactionTimedOutException` that occurs at `DefaultLockRepository.acquire()`. I think is the parameter `timeout = 1` on the `@Transactional` annotation on it.
Why the lock is not retried in this case ? It's a naive question, really.
Thank you in advance!

artembilan:
I think you are right: we need to add a `TransactionTimedOutException` into this catch:
```
catch (TransientDataAccessException e) {
					// try again
}
```

astrubel:
I will create a pull request for it.
Could you backport to 4.3.x too ?

artembilan:
> Could you backport to 4.3.x too ?

Marked respectively.
Will backport after the Pull Request.

Thanks

astrubel:
What does represent the parameter `timeout` on the `@Transaction` annotation? Is it contains the access time to the database or only the transaction opening and processing time?

artembilan:
Not sure what is your concern on the matter, but see its JavaDocs:
```
/**
	 * The timeout for this transaction (in seconds).
	 * <p>Defaults to the default timeout of the underlying transaction system.
	 * <p>Exclusively designed for use with {@link Propagation#REQUIRED} or
	 * {@link Propagation#REQUIRES_NEW} since it only applies to newly started
	 * transactions.
	 * @return the timeout in seconds
	 * @see org.springframework.transaction.interceptor.TransactionAttribute#getTimeout()
	 */
	int timeout() default TransactionDefinition.TIMEOUT_DEFAULT;
```

So, therefore it a timeout for the whole call including Java and DB access.

astrubel:
So I think that 1s is not enough for db access through particular networks.

How can we expose this parameter to setup into project side?

Is it possible to replace `@Transactional` annotation by `TransactionTemplate` into `DefaultLockRepository.acquire()` and setup `TransactionTemplate.timeout`?
```
@Override
public boolean acquire(String lock) {
   // save previous transactionTemplate values
   int prevIso = transactionTemplate.getIsolationLevel();
   int prevTimeout = transactionTemplate.getTimeout();

   // setup acquire transaction
   transactionTemplate.setIsolationLevel(Isolation.SERIALIZABLE);
   transactionTemplate.setTimeout(this.getTimeout());
   transactionTemplate.execute(st -> {
      deleteExpired(lock);
[...]

PR #3308, opened 2020-06-12
GH-3307: Retry DefaultLockRepository.acquire for TransTimedOutEx

astrubel:
Retry lock for TransactionTimedOutException
issue #3307

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ import org.springframework.dao.DataAccessResourceFailureException;
 import org.springframework.dao.TransientDataAccessException;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.integration.util.UUIDConverter;
+import org.springframework.transaction.TransactionTimedOutException;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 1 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -131,7 +132,7 @@ public void lock() {
 					}
 					break;
 				}
-				catch (TransientDataAccessException e) {
+				catch (TransientDataAccessException | TransactionTimedOutException e) {
```
artembilan, line 135: You know that we can't use this structure for `4.3.x`. That one is still based on Java 6.
Saying that it raises a question to you: why do you ask for `4.3.x` if you don't use Java 6 any more?

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
astrubel: Ho crap sorry..

We keep 4.3.x because of spring version. We not upgrade the project to a 5+ spring version yet :/

astrubel: I have to create a new pull request for `4.3.x`? I'm not familiar with your usage. I'm so sorry

artembilan: No, you don't: I'll back port it properly. Just curious why do you need fix in the version for Java 6, when you definitely could move already to `5.x` generation if you don't use Java 6 any more...

artembilan (changes requested):
Please, consider to incorporate the change we've agreed in the issue: https://github.com/spring-projects/spring-integration/issues/3307#issuecomment-644167738

artembilan (changes requested):
Please, add your name to the `@author` list of the changed class and update its Copyright into the current year.

Then we'll merge and back port it immediately.

Thank you!

PR #3310, opened 2020-06-16
Support Java 14

artembilan:
* Provide changes to avoid deprecated Java API
and have a compatibility back to Java 8
* Change affected test classes to JUnit 5 whenever it is possible
* Ignore/Disable some TCP/IP tests which don't pass on Java 14

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/SocketSupportTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/SocketSupportTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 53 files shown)


artembilan:
@garyrussell ,

I need your help with TCP/IP.

It looks like there are some changes in the UDP multicast with their `NetworkInterface` and `joinGroup()`.
Plus there is no `open` field in the `SocketChannel` any more.
See `@EnabledOnJre(JRE.JAVA_8)` tests.
And `SocketSupportTests.testNioClientAndServerSSLDifferentContexts()` fails for no reason for me.

Thanks

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/SocketSupportTests.java
```diff
@@ -518,7 +518,7 @@ public void testNioClientAndServerSSL() throws Exception {
 	}
 
 	@Test
-	@EnabledOnJre(JRE.JAVA_8)
+	@EnabledOnJre(JRE.JAVA_8) // Need to wait for write complete on Java 14
```
artembilan: @garyrussell , what does it mean?
Is there a way to do something similar what you have with that `if (System.getProperty("java.version").startsWith("1.8")) {` ?

garyrussell: I fixed it in the next commit.

garyrussell (approved):

artembilan:
Gary, I'm OK to merge this, but may be you have some idea about that deprecated `MulticastSocket.joinGroup()`?
Or you just can't investigate it altogether because your Mac doesn't support multicast?
Thanks

garyrussell:
Right; I need to test in a Linux VM. 

Please open an issue and assign to me so I don't forget about it.

garyrussell:
Should we add a build plan?

artembilan:
> Should we add a build plan?

Sure! On it...

PR #3313, opened 2020-06-24
Merge Spring Integration Kafka into Main Repo

garyrussell:
Only the last commit needs review; the others were merged from the SIK repo, to retain history.


artembilan (changes requested):
Gary, check, please, Travis report. 
We probably don't run `checkstyleNohttp` locally.
Thanks

src/reference/asciidoc/kafka.adoc
```diff
@@ -0,0 +1,846 @@
+[[kafka]]
+== Spring for Apache Kafka Support
+
+=== Overview
+
+This documentation pertains to versions 2.0.0 and above.
+For documentation for earlier releases, see the https://github.com/spring-projects/spring-integration-kafka/blob/1.3.x/README.md[1.3.x README].
+
+Spring Integration Kafka is an extension module to the https://spring.io/projects/spring-integration[Spring Integration Project].
```
artembilan, line 9: We probably need to revise this doc.
May be later

artembilan (changes requested):
Also would be great to have a `What's New` entry.

Thanks

artembilan (changes requested):
LGTM after NoHttp fix.

Thanks

artembilan:
Gary, any chances to fix NoHttp violations?
Btw, are you going to release new SK today?

garyrussell:
No; tomorrow.

garyrussell:
Why does the nohttp check not run with `testAll`? (which I ran).

artembilan:
Merged after some docs clean up and changing `@since` to `5.4`.

See commit history to be sure that everything from SIK is present: https://github.com/spring-projects/spring-integration/commits/master

artembilan:
> No; tomorrow.

OK. So, I move SI release for tomorrow as well.

> Why does the nohttp check not run with testAll?

We have a condition like this:
```
if (System.getenv('TRAVIS') || System.getenv('bamboo_buildKey')) {
	apply plugin: 'io.spring.nohttp'
```

I think it is not related to `testAll`. 
It is fully OK with me to have it failed on Travis or CI afterwards.
The report is clean enough to fix it in subsequent commit.
The `NoHTTP` is not that common task when we should slow down our regular development routine.

PR #3316, opened 2020-07-01
[WIP][INT-4566 ]-Implement Inbound channel adapter for spring data r2dbc 

rohanmukesh:
I have added Inbound channel adapter for spring data r2db.
The inbound adapter (R2dbcMessageSource) allows users to send query in two forms:
 a) String -> This uses databaseclient.execute(..) to execute sql query and return result.
 b) Query Object -> This uses r2dbcEntityOperations.select(..) to execute Query and return result.

I have added Test Class: R2dbcMessageSourceTests with 8 test cases. The test cases tests all the scenarions which includes:
String query Test with Single result or All Results, Query Object Test with Single Result or All Results and check for bad sql grammer, IllegalStateException.

### DIFF
--- a/spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
+++ b/spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
@@ -157,29 +157,29 @@ public class R2dbcMessageSource extends AbstractMessageSource<Publisher<?>> {
 		return queryMono.flatMapMany(this::select);
 	}
 
-	private Mono<?> selectOne(Object queryMono) {
-		String query = evaluateQueryObject(queryMono);
+	private Mono<?> selectOne(Object queryObject) {
+		String queryString = evaluateQueryObject(queryObject);
 		return this.databaseClient
-				.execute(query)
+				.execute(queryString)
 				.as(this.payloadType)
 				.fetch()
 				.one();
 	}
 
-	private Flux<?> select(Object queryMono) {
-		String query = evaluateQueryObject(queryMono);
+	private Flux<?> select(Object queryObject) {
+		String queryString = evaluateQueryObject(queryObject);
 		return this.databaseClient
-				.execute(query)
+				.execute(queryString)
 				.as(this.payloadType)
 				.fetch()
 				.all();
 	}
 
-	private String evaluateQueryObject(Object query) {
-		if (query instanceof String) {
-			return (String) query;
+	private String evaluateQueryObject(Object queryObject) {
+		if (queryObject instanceof String) {
+			return (String) queryObject;
 		}
 		throw new IllegalStateException("'queryExpression' must evaluate to String " +
-				"or org.springframework.data.relational.core.query.Query, but not: " + query);
+				"or org.springframework.data.relational.core.query.Query, but not: " + queryObject);
 	}
 }
--- a/spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
+++ b/spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.r2dbc.inbound;
+
+import org.reactivestreams.Publisher;
+
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.data.r2dbc.core.R2dbcEntityOperations;
+import org.springframework.data.relational.core.query.Query;
+import org.springframework.expression.Expression;
+import org.springframework.expression.TypeLocator;
+import org.springframework.expression.spel.support.StandardEvaluationContext;
+import org.springframework.expression.spel.support.StandardTypeLocator;
+import org.springframework.integration.endpoint.AbstractMessageSource;
+import org.springframework.integration.expression.ExpressionUtils;
+import org.springframework.integration.expression.ValueExpression;
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+
+/**
+ * An instance of {@link org.springframework.integration.core.MessageSource} which returns
+ * a {@link org.springframework.messaging.Message} with a payload which is the result of
+ * execution of a {@link Query}. When {@code expectSingleResult} is false (default), the R2dbc
+ * {@link Query} is executed using {@link R2dbcEntityOperations#select(Query, Class)} method which
+ * returns a {@link reactor.core.publisher.Flux}.
+ * The returned {@link reactor.core.publisher.Flux} will be used as the payload of the
+ * {@link org.springframework.messaging.Message} returned by the {@link #receive()}
+ * method.
+ * <p>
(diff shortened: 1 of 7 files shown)


artembilan (changes requested):
Would you mind to rebase your branch to the latest `upstream master`?
Plus squash all these commits: we need separate commits only after review.
Right now I see some unrelated commits in this PR. Probably because your local `master` is not fresh according to upstream...

spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
```diff
@@ -0,0 +1,212 @@
...
+					.fetch()
+					.all();
+		}
+		return result;
+	}
+
+	private Object executeQuery(Query query) {
+		Object result;
+		if (this.expectSingleResult) {
+			result = this.r2dbcEntityOperations.selectOne(query, this.entityClass);
```
artembilan, line 203: I think we should not rely on the `@Table` on the provided class.
I really believe that in most cases people are going to deal with just a `HashMap`.
Just because we interact here with messaging system and it is really hard to carry a particular domain class from one endpoint to another. 

So, I think we should get rid of `r2dbcEntityOperations` contract over here and just deal with the plain `DatabaseClient`. This way end-user will be able to not tie his/her logic with persistence model and those extra annotations.

We probably need to revise the `Query`  logic then in favor of `SelectSpec`...

spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
```diff
@@ -0,0 +1,280 @@
...
+import reactor.test.StepVerifier;
+
+
+/**
+ *  @author Rohan Mukesh
+ *
+ *  @since 5.4
+ */
+@SpringJUnitConfig (R2dbcIntegrationTestConfiguration.class)
+public class R2dbcMessageSourceTests {
```
artembilan, line 63: `@DirtiesContext`, please, to stop the application context and clean up the memory from that H2 DB.

artembilan (changes requested):
Some review so far.
Thanks for the effort, but I know this is not an easy task.
So, let's engineer the final solution together!

rohanmukesh:
Hello Artem, I have updated pull request with changes you requested and also with only "String Query". Although I have kept the constructor for queryExpression.  Removed the dependency on Entity @Table and just using databaseclient execute method for now. Let me know your feedback.

spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
```diff
@@ -0,0 +1,232 @@
...
+	@Test
+	public void testAnyOtherObjectQueryExpression() {
+		R2dbcMessageHandler handler = new R2dbcMessageHandler(this.entityTemplate);
+		handler.setBeanFactory(mock(BeanFactory.class));
+		handler.setApplicationContext(mock(ApplicationContext.class, Answers.RETURNS_MOCKS));
+		handler.afterPropertiesSet();
+		Message<Person> message = MessageBuilder.withPayload(this.createPerson("Bob", 35)).build();
+		waitFor(handler.handleMessage(message));
+
+		message = MessageBuilder.withPayload(this.createPerson("Tom", 40)).build();
```
artembilan, line 212: Don't use `this.` for method calls

artembilan (commented):
Please, be careful here in GH comments when you try to use something with `@` symbol. That "table" is a user here on GH and now he/she gets all the notifiations from our communication over here 😄 .
See GH comments formatting recommendations: https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks

spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
```diff
@@ -0,0 +1,212 @@
...
+	private ApplicationContext applicationContext;
+
+	private volatile boolean initialized = false;
+
+	/**
+	 * Create an instance with the provided {@link R2dbcEntityOperations} and SpEL expression
+	 * which should resolve to a Relational 'query' string.
+	 * It assumes that the {@link R2dbcEntityOperations} is fully initialized and ready to be used.
+	 * The 'queryExpression' will be evaluated on every call to the {@link #receive()} method.
+	 *
```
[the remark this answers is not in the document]
artembilan: Not addressed.

spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
```diff
@@ -0,0 +1,188 @@
...
+					Object result;
+					if (value instanceof String) {
+						result = executeQuery((String) value);
+					}
+					else {
+						throw new IllegalStateException("'queryExpression' must evaluate to String " +
+								"or org.springframework.data.relational.core.query.Query, but not: " + value);
+					}
+					return result;
+				}).block();
```
artembilan, line 167: You know I have changed my mind a little bit and figure out something like this for better optimization:
```
Mono<?> queryMono = Mono.fromSupplier(() -> this.queryExpression.getValue(this.evaluationContext));
if (this.this.expectSingleResult) {
     return queryMono.flatMap(this::selectOne)
}
else {
     return queryMono.flatMapMany(this::select)
}
```

The `selectOne` should return a `Mono` and `select` a `Flux`

artembilan (changes requested):

spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
```diff
@@ -0,0 +1,185 @@
...
+	protected Object doReceive() {
+		Assert.isTrue(this.initialized, "This class is not yet initialized. Invoke its afterPropertiesSet() method");
+		Mono<?> queryMono = Mono.fromSupplier(() -> this.queryExpression.getValue(this.evaluationContext));
+		if (this.expectSingleResult) {
+			return queryMono.flatMap(this::selectOne);
+		}
+		return queryMono.flatMapMany(this::select);
+	}
+
+	private Mono<?> selectOne(Object queryMono) {
```
artembilan, line 160: Well, its not `Mono` here any more. Isn't it? 
I mean the name for the param is misleading 😄

spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	@Configuration
+	@Import(R2dbcDatabaseConfiguration.class)
+	static class R2dbcMessageSourceConfiguration {
+
+		@Autowired
+		DatabaseClient databaseClient;
+
+		@Bean(name = "r2dbcMessageSourceSelectOne")
```
artembilan, line 147: FYI: the method name becomes a bean name by default.
So, probably for this use-case you definitely don't need a `name` attribute on this annotation...

spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/outbound/R2dbcMessageHandlerTests.java
```diff
@@ -29,79 +27,49 @@
...
 import reactor.test.StepVerifier;
 
 /**
- *  @author Rohan Mukesh
+ * @author Rohan Mukesh
  *
- *  @since 5.4
+ * @since 5.4
  */
 @SpringJUnitConfig
```
artembilan, line 56: The `@DirtiesContext` here as well to close an application context and shoultdown embedded DB altogether.

artembilan (changes requested):
Very close to the merge!

artembilan (changes requested):
Please, consider to rebase to the latest `master` so Travis for your PR won't fail.

artembilan (changes requested):
Good. Let me pull it locally for final review and if there are only minor concerns I'll fix them and merge

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/eb3c6a017f2df2d954de1a842a22b2a15071c060.

Pay attention, please, how I have optimized a fetch preparation code.

PR #3318, opened 2020-07-01
GH-3315: Fix (S)FTP Stream with Fair Rotation

### LINKED ISSUE #3315: Streaming remote files with a single file filter breaks `fair` rotation of sources
**Describe the bug**
 https://github.com/spring-projects/spring-integration/blame/master/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java#L262.  There may have been some changes here that affect the behavior of the fair rotation when streaming in the sftp source.  Since we have SftpPersistentAcceptOnceFileListFilter configured, this supports single file filtering so maxFetch is ignored.
**To Reproduce**

Steps to reproduce the behavior.

Comment out this line https://github.com/dturanski/stream-applications-1/blob/sftp-source/functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierConfiguration.java#L160

Then run this test
https://github.com/dturanski/stream-applications-1/blob/sftp-source/functions/supplier/sftp-supplier/src/test/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierApplicationTests.java#L199

**Expected behavior**

The test should pass, verifying fair rotations

**Sample**

See above

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3315

`maxFetchSize` ignored with filters supporting single file filtering (default).
This breaks "fair" rotation with the `RotatingServerAdvice`.

Honor `maxFetchSize`, even with filters that support single file filtering.

**cherry-pick to 5.3.x, 5.2.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


artembilan:
... and cherry-picked to `5.3.x`.
Back-ported to `5.2.x` as https://github.com/spring-projects/spring-integration/commit/6d60b9065b399a23b171b893821f514eca624027 after fixing the `RotatingServersTests` for JUnit 4.

PR #3319, opened 2020-07-01
Fix MessagingGatewaySupport for reactive error

artembilan:
The `onErrorResume()` was in a wrong place for the
`doSendAndReceiveMessageReactive()`: we have to catch all the exceptions
from the top level `Mono`, not only a reply one as it was before.

Ensure in HTTP and WebFlux test that behavior is fixed


artembilan:
See https://stackoverflow.com/questions/62645695/how-to-use-custom-error-channel-and-webflux-inboundgateway-to-reply-with-an-erro for more info.

garyrussell:
Cherry-picked to 5.3.x, 5.2.x

PR #3322, opened 2020-07-02
GH-3320: Refine lifecycle control in StdIntFlow

### LINKED ISSUE #3320: Consider Honoring Lifecycle.stop() in StandardIntegrationFlow
**Expected Behavior**

`stop()` should stop all the components in an integration flow; which might be autowired as a `Lifecycle`.

**Current Behavior**

`stop()` does nothing unless the flow was started via `start()`; e.g. via dynamic flow registration.

**Context**

It is not intuitive that `stop()` does nothing in this case.

**Pros**

- Users don't need to figure out what the first component is (e.g. SPCA).
- Active channels/pollers within the flow will be stopped as well.

**Cons**

?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3320

Turns out that lifecycle control for the whole bunch of components
in one `IntegrationFlow` is useful in fields.

* Change the logic in the `StandardIntegrationFlow` to let to call
`start()` and `stop()` independently how the flow was registered in
the application context.
This way it can be autowired as a `Lifecycle` to let end-user to
avoid the search for proper component in the flow to stop or start
manually - all the components registered with the flow are going
to be stopped or started respectively

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -63,12 +63,11 @@ import org.springframework.messaging.MessageChannel;
  * @see IntegrationFlows
  * @see org.springframework.integration.dsl.context.IntegrationFlowBeanPostProcessor
  * @see org.springframework.integration.dsl.context.IntegrationFlowContext
+ * @see SmartLifecycle
  */
 public class StandardIntegrationFlow implements IntegrationFlow, SmartLifecycle {
 
 	private final Map<Object, String> integrationComponents;
-
-	private final List<SmartLifecycle> lifecycles = new LinkedList<>();
 
 	private MessageChannel inputChannel;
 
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
```diff
@@ -127,30 +124,29 @@ public void start() {
 
 	@Override
 	public void stop(Runnable callback) {
-		if (this.lifecycles.size() > 0) {
-			AggregatingCallback aggregatingCallback = new AggregatingCallback(this.lifecycles.size(), callback);
-			ListIterator<SmartLifecycle> iterator = this.lifecycles.listIterator(this.lifecycles.size());
-			while (iterator.hasPrevious()) {
```
garyrussell, line 133: I didn't notice this, but stopping right to left was wrong; corrected in the new code.

garyrussell (approved):

garyrussell:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java:128:41: Reference to instance variable 'integrationComponents' needs "this.". [RequireThis]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java:143:41: Reference to instance variable 'integrationComponents' needs "this.". [RequireThis]
```

spring-integration-core/src/main/java/org/springframework/integration/dsl/StandardIntegrationFlow.java
artembilan: No, Gary, it was right. Because we started before from right to left and filled out the local `lifecycles` property.
Then we used this `lifecycles` in the `stop()` and therefore the first one in the flow was in the end of that list, so we again should iterate it backwards to stop in proper order.
Right now indeed we don't have an intermediate list therefore the logic is changed around a normal `integrationComponents` state.

garyrussell: Ah... right.

PR #3327, opened 2020-07-06
GH-3326: TCP: Support Unsolicited Server Messages

### LINKED ISSUE #3326: Enhance the TcpOutboundGateway to send late replies and other unsolicited messages to a message channel
**Expected Behavior**

I need capability for TcpOutboundGateway to support also single way communication that comes from tcp server without any request from client.

**Current Behavior**

If i am correct according to my experience on TcpOutboundGateway, it is working two direction that is request/response but also in some use cases server should send to client periodical arbitrary data but without any request if server sends data to client, client generates error like "ERROR 54816 --- [pool-2-thread-1] o.s.i.ip.tcp.TcpOutboundGateway : Cannot correlate response - no pending reply for ::58628:62fd67b6-af2d-42f1-9c4d-d232fbe9c8ca". 

If this capability is supported by TcpOutboundGateway than it will not generate any Error then server can send data to connected client so both "request/response" type of communication and "only push message" from server type of communication can be supported.

**Context**

I could not receive push (arbitrary) messages from server with using TcpOutboundGateway and this issue forced me to use adapters but message correlation in adapters is complicated if tcp server is not allowed to support such an extension. In my case tcp server was already written and is a a black box product. When i try adapters, I lost req/resp capability because as i understand adapters working asynchronous. 

What are you trying to accomplish?

I tried to accomplished that both i have to use req/resp type of communication between client and server but also incase of any push message comes from server, that client should receive it.

To be honest, i did not have other alternative but i am totally open to hear other alternatives.

@garyrussell supported me to extend TcpOutboundGateway to inject message channel for receive push messages from server and also he supported me to use adapters but also he mentioned that message correlation is complicated if tcp server is intelligent.

Here some links about this topic

[https://stackoverflow.com/questions/62702192/spring-integration-gateway-vs-adapters](url)
[https://stackoverflow.com/questions/62715732/is-it-possible-to-use-tcpsendingmessagehandler-and-tcpouboundgateway-according-t](url)
[https://stackoverflow.com/questions/62731457/spring-integration-tcpoutboundgateway-serviceactivator-message-channel-and-err](url)
[https://stackoverflow.com/questions/46021319/how-to-extend-tcp-outbound-gateway-to-just-receive-messages-in-spring-integratio?rq=1](url)
[https://stackoverflow.com/questions/33902543/spring-integration-tcp-server-push-to-client](url)

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3326

- OB Gateway - send unsolicited messages and late replies to a channel
- Support multiple `TcpSender`s

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/TcpOutboundGatewaySpec.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/TcpOutboundGatewaySpec.java
@@ -114,6 +114,11 @@ public class TcpOutboundGatewaySpec extends MessageHandlerSpec<TcpOutboundGatewa
 		return this;
 	}
 
+	public TcpOutboundGatewaySpec unsolictedMessageChannelName(String channelName) {
+		this.target.setUnsolicitedMessageChannelName(channelName);
+		return this;
+	}
+
 	@Override
 	public Map<Object, String> getComponentsToRegister() {
 		return this.connectionFactory != null
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/TcpOutboundGatewaySpec.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/TcpOutboundGatewaySpec.java
@@ -26,6 +26,7 @@ import org.springframework.integration.expression.FunctionExpression;
 import org.springframework.integration.ip.tcp.TcpOutboundGateway;
 import org.springframework.integration.ip.tcp.connection.AbstractClientConnectionFactory;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
 
 /**
  * A {@link MessageHandlerSpec} for {@link TcpOutboundGateway}s.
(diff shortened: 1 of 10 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/dsl/TcpOutboundGatewaySpec.java
```diff
@@ -114,6 +114,11 @@ public TcpOutboundGatewaySpec async(boolean async) {
 		return this;
 	}
 
+	public TcpOutboundGatewaySpec unsolictedMessageChannelName(String channelName) {
```
artembilan, line 117: Looks like you have there this setter as well `setUnsolicitedMessageChannel(MessageChannel unsolicitedMessageChannel)`. Don't we need to have similar option here on DSL level?
Thanks

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
```diff
@@ -118,14 +129,53 @@ public void setIntegrationEvaluationContext(EvaluationContext evaluationContext)
...
-		}
-		Assert.state(!this.closeStreamAfterSend || this.isSingleUse,
-				"Single use connection needed with closeStreamAfterSend");
+	/**
+	 * Specify the Spring Integration reply channel. If this property is not
+	 * set the gateway will check for a 'replyChannel' header on the request.
+	 * @param replyChannel The reply channel.
+	 */
+	public void setReplyChannel(MessageChannel replyChannel) {
+		this.setOutputChannel(replyChannel);
```
artembilan, line 138: `this.` on method calls?
I guess you have just moved those existing setter over here for better class structure.
Nevertheless can we fix `this.` as well since we are already here in a change?

artembilan (changes requested):
I thought first of all that this has to be supported via an `ApplicationEvent` support, but technically it doesn't matter that we are going to send them into some configured channel. This is similar to `discardChannel` in other places, but just with a different name and of course different purpose.

So, give me an answer to those my two minor concerns and we are good to go with merge!

Thank you!

PR #3330, opened 2020-07-07
Diagnostics for travis failures

garyrussell:


artembilan:
> The job exceeded the maximum log length, and has been terminated.

Yeah... Not going to help...

garyrussell:
Surprising since we're only running one test.

PR #3331, opened 2020-07-07
Add diagnostics for travis failure

garyrussell:


artembilan:
Gary, it is green now. Does it say anything to you?
Is that  connection config for FTP server enough to treat as a fix for intermittent problem?

Thanks

garyrussell:
I think so, but I need to do some more investigation to understand why we are not closing sessions somewhere.

PR #3332, opened 2020-07-08
Fix CachedSessionFactory Race

garyrussell:
Close the pool so that any sessions returned after the factory is
`destroy()`ed are closed.

**cherry-pick to all supported branches**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/Pool.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/Pool.java
@@ -72,4 +72,11 @@ public interface Pool<T> {
 	 */
 	int getAllocatedCount();
 
+	/**
+	 * Close the pool; returned items will be destroyed.
+	 * @since 4.3.23
+	 */
+	default void close() {
+	}
+
 }
--- a/spring-integration-core/src/test/java/org/springframework/integration/util/SimplePoolTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/util/SimplePoolTests.java
@@ -256,10 +256,12 @@ public class SimplePoolTests {
 	@Test
 	void testClose() {
 		SimplePool<String> pool = stringPool(10, new HashSet<>(), new AtomicBoolean());
-		String item = pool.getItem();
+		String item1 = pool.getItem();
+		String item2 = pool.getItem();
+		pool.releaseItem(item2);
+		assertThat(pool.getAllocatedCount()).isEqualTo(2);
 		pool.close();
-		assertThat(pool.getAllocatedCount()).isEqualTo(1);
-		pool.releaseItem(item);
+		pool.releaseItem(item1);
 		assertThat(pool.getAllocatedCount()).isEqualTo(0);
 		assertThatIllegalStateException().isThrownBy(pool::getItem);
 	}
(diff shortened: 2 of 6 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/Pool.java
```diff
@@ -72,4 +72,11 @@
 	 */
 	int getAllocatedCount();
 
+	/**
+	 * Close the pool; returned items will be destroyed.
+	 * @since 4.3.23
+	 */
+	default void close() {
```
artembilan, line 79: The `default` is not going to work for Java 6, but we may not just introduce this method into the `Pool` interface for backward bytecode compatibility.

spring-integration-core/src/test/java/org/springframework/integration/util/SimplePoolTests.java
```diff
@@ -27,7 +29,7 @@
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
```
artembilan, line 32: I'll take a look what we need to do during back-port to `4.3.x` ...

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/CachingClientConnectionFactory.java
```diff
@@ -389,6 +390,12 @@ public synchronized void stop() {
 		this.pool.removeAllIdleItems();
 	}
 
+	@Override
+	public void destroy() throws Exception {
+		this.pool.close();
+		this.pool.removeAllIdleItems();
```
artembilan, line 396: Why that new `close()` can't do this for us, too?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/Pool.java
garyrussell: Probably easiest to cast to `SimplePool` on 4.3.x?

artembilan:
Any thoughts about another failure:
```
SftpStreamingMessageSourceTests > testMaxFetchLambdaFilter() FAILED
    org.springframework.messaging.MessagingException at SftpStreamingMessageSourceTests.java:159
        Caused by: org.springframework.integration.util.PoolItemNotAvailableException at SftpStreamingMessageSourceTests.java:159
            Caused by: java.lang.IllegalStateException at SftpStreamingMessageSourceTests.java:159
                Caused by: java.lang.IllegalStateException at SftpStreamingMessageSourceTests.java:159
                    Caused by: com.jcraft.jsch.JSchException at SftpStreamingMessageSourceTests.java:159
```
which is probably related to the pool as well...
I've seen it failing before though.

Thanks

artembilan (approved):
Cool! 👍

artembilan:
Cherry-picked to `5.3.x` & `5.2.x`.
Back-ported to `5.1.x` as https://github.com/spring-projects/spring-integration/commit/8df8028537f06acca3b833f8442f8560e1ab5c8f after fixing `SimplePoolTests` conflict.
Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/c95e6dcc3b4c03570cb7d3065c8cecadd879a48a after removing `default` from the `Pool` interface.

PR #3335, opened 2020-07-10
Implement Inbound Channel Adapter - Update functionality for Spring Data R2dbc

rohanmukesh:
I have added Update functionality for Inbound Channel Adapter. 
End user can set Update query using setUpdateQuery method.
End user can also use setBindFunction method to bind parameters used in Update Query.
Example of Update query for bindFunction -> update person set name=Foo where id=:id


artembilan (changes requested):
Please, consider to make a rebase onto the latest `upstream master`.
There likely some conflicts present.

Thanks

spring-integration-r2dbc/src/main/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSource.java
```diff
@@ -151,9 +175,24 @@ protected Object doReceive() {
...
+			return queryMono.flatMap(FetchSpec::one)
+					.flatMap(this::executeUpdate);
+		}
+
+		return queryMono.flatMapMany(FetchSpec::all)
+				.flatMap(this::executeUpdate);
+	}
+
+	private Mono<Object> executeUpdate(Object result) {
+		if (this.updateSql != null) {
```
artembilan, line 187: Since this option is static and doesn't depend on a value we should apply `if...else` logic only once.
See `Flux.transform()` operator:

```
	 * Transform this {@link Flux} in order to generate a target {@link Flux}. Unlike {@link #transformDeferred(Function)}, the
	 * provided function is executed as part of assembly.
```

spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
```diff
@@ -129,6 +135,135 @@ public void validateSuccessfulQueryWithMultipleElementOfFluxDBObject() {
...
+				"id",	o.getId()));
+		StepVerifier.create(r2dbcMessageSourceSelectMany.receive().getPayload())
+				.expectNextCount(2)
+				.verifyComplete();
+
+		this.entityTemplate.select(Person.class)
+				.all()
+				.as(StepVerifier::create)
+				.assertNext(person -> assertThat(person.getName()).isEqualTo("Foo"))
+				.assertNext(person -> assertThat(person.getName()).isEqualTo("Foo"))
```
artembilan, line 185: May we be sure here that these persons are really different?
Maybe checking their age...
Right now their names are the same, so it isn't clear if both asserts are not against the same record.

artembilan (changes requested):

artembilan:
Please, consider to rebase to the latest master and fix conflicts: https://github.com/spring-projects/spring-integration/commit/cd63763bb47b210decff565a83ecd3f8f556c9d2#diff-a17e3e0e28b8487fec47c7f9f4c99c21.

There were a lot of deprecations in Spring Data R2DBC...

artembilan (approved):
OK. Since you don't answer and we really need this feature for the upcoming release, I'm pulling this locally for the final review and clean up.

Hope to see more contribution from you in the future!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b54085c99dff0f878fd59674c0bb52406f4f76f9.

Thank you for contribution; Really looking forward for more from you!

PR #3337, opened 2020-07-13
JDBC lock acquire optimization

astrubel:
I think a `DELETE` query at each call of `acquire()` is unoptimized.
So I merged the `deleteExpiredQuery` into `updateQuery` like this:
```
UPDATE %sLOCK SET CREATED_DATE=? WHERE REGION=? AND LOCK_KEY=? AND (CLIENT_ID=? OR CREATED_DATE<?)
```

I kept `deleteExpired()` to clean table for a retro-compatibility.

Does that make sense?

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
@@ -73,16 +73,18 @@ public class DefaultLockRepository implements LockRepository, InitializingBean {
 
 	private String deleteQuery = "DELETE FROM %sLOCK WHERE REGION=? AND LOCK_KEY=? AND CLIENT_ID=?";
 
-	private String deleteExpiredQuery = "DELETE FROM %sLOCK WHERE REGION=? AND LOCK_KEY=? AND CREATED_DATE<?";
+	private String deleteExpiredQuery = "DELETE FROM %sLOCK WHERE REGION=? AND CREATED_DATE<?";
 
 	private String deleteAllQuery = "DELETE FROM %sLOCK WHERE REGION=? AND CLIENT_ID=?";
 
-	private String updateQuery = "UPDATE %sLOCK SET CREATED_DATE=? WHERE REGION=? AND LOCK_KEY=? AND CLIENT_ID=?";
+	private String updateQuery = "UPDATE %sLOCK SET CLIENT_ID=?, CREATED_DATE=? WHERE REGION=? AND LOCK_KEY=? AND (CLIENT_ID=? OR CREATED_DATE<?)";
 
 	private String insertQuery = "INSERT INTO %sLOCK (REGION, LOCK_KEY, CLIENT_ID, CREATED_DATE) VALUES (?, ?, ?, ?)";
 
 	private String countQuery =
 			"SELECT COUNT(REGION) FROM %sLOCK WHERE REGION=? AND LOCK_KEY=? AND CLIENT_ID=? AND CREATED_DATE>=?";
+
+	private String renewQuery = "UPDATE %sLOCK SET CREATED_DATE=? WHERE REGION=? AND LOCK_KEY=? AND CLIENT_ID=?";
 
 	/**
 	 * Constructor that initializes the client id that will be associated for
@@ -142,6 +144,7 @@ public class DefaultLockRepository implements LockRepository, InitializingBean {
 		this.updateQuery = String.format(this.updateQuery, this.prefix);
 		this.insertQuery = String.format(this.insertQuery, this.prefix);
 		this.countQuery = String.format(this.countQuery, this.prefix);
+		this.renewQuery = String.format(this.renewQuery, this.prefix);
 	}
 
 	@Override
@@ -157,8 +160,8 @@ public class DefaultLockRepository implements LockRepository, InitializingBean {
 	@Transactional(isolation = Isolation.SERIALIZABLE, timeout = 1)
 	@Override
 	public boolean acquire(String lock) {
-		deleteExpired(lock);
-		if (this.template.update(this.updateQuery, new Date(), this.region, lock, this.id) > 0) {
+		if (this.template.update(this.updateQuery, this.id, new Date(), this.region, lock, this.id,
+				new Date(System.currentTimeMillis() - this.ttl)) > 0) {
 			return true;
 		}
 		try {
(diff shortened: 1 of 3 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
```diff
@@ -157,8 +160,8 @@ public void delete(String lock) {
 	@Transactional(isolation = Isolation.SERIALIZABLE, timeout = 1)
 	@Override
 	public boolean acquire(String lock) {
-		deleteExpired(lock);
-		if (this.template.update(this.updateQuery, new Date(), this.region, lock, this.id) > 0) {
+		if (this.template.update(this.updateQuery, new Date(), this.region, lock, this.id,
```
artembilan, line 163: Wll, I think without `deleteExpired()` over here we are missing the fact that another node in the cluster won't be able to acquire the lock since it was removed. We are not able to reacquire it with the current node because it is expired, but the lock still is not available for another clients to insert because the row for an old is still there.

Any ideas how to deal with that, please, with your new approach?

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
astrubel: > another node in the cluster won't be able to acquire the lock since it was removed
This point is covered by `JdbcLockRegistryDifferentClientTests.testOutOfDateLockTaken()`

> reacquire it with the current node because it is expired
I wrote `JdbcLockRegistryTests.testReentrantLockAfterExpiration()` to cover this case

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
```diff
@@ -73,17 +73,19 @@
 
 	private String deleteQuery = "DELETE FROM %sLOCK WHERE REGION=? AND LOCK_KEY=? AND CLIENT_ID=?";
 
-	private String deleteExpiredQuery = "DELETE FROM %sLOCK WHERE REGION=? AND LOCK_KEY=? AND CREATED_DATE<?";
+	private String deleteExpiredQuery = "DELETE FROM %sLOCK WHERE REGION=? AND CREATED_DATE<?";
 
 	private String deleteAllQuery = "DELETE FROM %sLOCK WHERE REGION=? AND CLIENT_ID=?";
 
-	private String updateQuery = "UPDATE %sLOCK SET CREATED_DATE=? WHERE REGION=? AND LOCK_KEY=? AND CLIENT_ID=?";
+	private String updateQuery = "UPDATE %sLOCK SET CLIENT_ID=?, CREATED_DATE=? WHERE REGION=? AND LOCK_KEY=? AND (CLIENT_ID=? OR CREATED_DATE<?)";
```
artembilan, line 80: Would you mind to provide a brief explanation for a new logic which really could be as a commit message.
So, we know the reason for a behavior change.

Thanks

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
astrubel: Thanks to the pair of `CLIENT_ID`/`CREATED_DATE`, it is possible to know if the lock is expired, already held or can be reacquire.
I think that the pre delete row is unnecessary and redundant with these two columns. The rows doesn't need to be purged because the number of row in the table is finite and equal to the number of locks. At worst, the table will be able to be purged by `deleteExpired()`.

I propose to overwrite these two columns at acquire lock time instead of pre delete the row.
So `delete` then `update` or `insert` become `update` or `insert`.

**Sample:**
One client held a lock.
- Another client will be able to hold the same lock only if:
1. the row doesn't exist. Done when the first client will call `unlock()`.
2. the lock expire in case of hardware shutdown. Done by "`OR CREATED_DATE<?`" in `updateQuery`.
- The same client will be able to reacquire it only if:
1. the 2 points above.
2. the lock is already held by him. Done by "`CLIENT_ID=? OR`" in `updateQuery`.

**Cost with PostgreSQL:**
```
EXPLAIN DELETE FROM INT_LOCK WHERE REGION='DEFAULT' AND LOCK_KEY='FOO' AND CREATED_DATE<'2020-07-14';
Delete on int_lock  (cost=0.14..8.17 rows=1 width=6)
  ->  Index Scan using int_lock_pk on int_lock  (cost=0.14..8.17 rows=1 width=6)
        Index Cond: (((region)::text = 'DEFAULT'::text) AND (lock_key = 'FOO'::bpchar))
        Filter: (created_date < '2020-07-14 00:00:00'::timestamp without time zone)

EXPLAIN UPDATE INT_LOCK SET CREATED_DATE='2020-07-15' WHERE REGION='DEFAULT' AND LOCK_KEY='FOO' AND CLIENT_ID=NULL;
Update on int_lock  (cost=0.00..11.40 rows=1 width=520)
  ->  Result  (cost=0.00..11.40 rows=1 width=520)
        One-Time Filter: false
        ->  Seq Scan on int_lock  (cost=0.00..11.40 rows=1 width=520)
```
```
EXPLAIN UPDATE INT_LOCK SET CLIENT_ID=NULL, CREATED_DATE='2020-07-15' WHERE REGION='DEFAULT' AND LOCK_KEY='FOO' AND (CLIENT_ID=NULL OR CREATED_DATE<'2020-07-14')
Update on int_lock  (cost=0.14..8.17 rows=1 width=372)
  ->  Index Scan using int_lock_pk on int_lock  (cost=0.14..8.17 rows=1 width=372)
        Index Cond: (((region)::text = 'DEFAULT'::text) AND (lock_key = 'FOO'::bpchar))
        Filter: (created_date < '2020-07-14 00:00:00'::timestamp without time zone)
```

artembilan: Nice! I apply this comment as a commit message and merge.
However I think there is a bit mismatch: the current holder can update its row even if it expired. 
`CLIENT_ID=? OR CREATED_DATE<?` - we just don't care about the date since the `CLIENT_ID` is equal.

Confirm, please, while I'm cleaning up the code a little bit on merge locally.

Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/e4611d97a1412bc78e34fe0355433a1ef40b5953.

Thank you for contribution!

PR #3339, opened 2020-07-14
Fix Kafka Send Timeout

### LINKED ISSUE #928: Synchronous Producer is not fully synchronous since it returns after 10s by default
Hello,

I have a few questions and recommendations for binder-kafka.

Here is some background information:
- version: Hoxton.RELEASE
- Kafka producer retries is set to Int.MAX
- delivery.timeout.ms is default to 120s.

1. Using async producer, the result of `MessageChannel.send(Message<?>)` always returns true.
If there is a problem with Kafka brokers (for example, NOT_ENOUGH_REPLICAS exception due to disk full), the spring boot app would get `true` as the result of the `send()` but it won't actually produce due to an error. This would end up in a case where the result of the application differs to what actually happened with kafka producer. This would lead me to use `sync` option.

2. Using sync producer (to get the actual response of `send()`), I noticed the timeout given using `MessageChannel.send()` is set to -1 internally. However, going through the code I see that the default value of sendTimeout gathered from `sendTimeoutExpression` is 10s in [KafkaProducerMessageHandler](https://github.com/spring-projects/spring-integration-kafka/blob/a66955513938d8dadbfc462a1ef0848e5a1bf5d4/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java#L597). (I am aware of the sendTimeoutExpression config btw) In Kafka producers, the default `delivery.timeout.ms` is 120s, but why is it designed to be returned after 10s? I do believe and agree that waiting for 120s might not be the most efficient way of handling errors but there could be a chance that the result of `send()` might be different after 10s and `delivery.timeout.ms` due to the retries? What if the Nth retry went through? If so, even with `sync` option, the result is not fully synchronous. Instead of 10s, maybe it should be dependent on `delivery.timeout.ms`. 

Then the question comes to "what is the most golden source of the producer send result?" I think that one case that is always valid is if an exception is thrown from the `ProducerListener.onError()`, then the spring boot app would know that `send()` has failed. Implementation of `ProducerListener` would be optional to the user, but I am wondering what is the most accurate way of the spring boot app knowing that `send()` has actually failed even after all the retries.

Please let me know.

Regards,

Jay

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/928

Kafka has a longer default send timeout; this means a send could be successful long
after Spring has timed out the send.

**I will backport to the 3.3.x extension after merge**


spring-integration-kafka/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
```diff
@@ -175,6 +181,25 @@ public KafkaProducerMessageHandler(final KafkaTemplate<K, V> kafkaTemplate) {
...
+			dt = ProducerConfig.configDef().defaultValues().get(ProducerConfig.DELIVERY_TIMEOUT_MS_CONFIG);
+		}
+		if (dt instanceof Long) {
+			this.sendTimeoutExpression = new ValueExpression<>(((Long) dt) + TIMEOUT_BUFFER);
+		}
+		else if (dt instanceof Integer) {
+			this.sendTimeoutExpression = new ValueExpression<>(Long.valueOf((Integer) dt) + TIMEOUT_BUFFER);
+		}
+		else if (dt instanceof String) {
+			this.sendTimeoutExpression = new ValueExpression<>(Long.parseLong((String) dt) + TIMEOUT_BUFFER);
```
artembilan, line 201: Why can't we call `setSendTimeout(long sendTimeout)` instead?

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
```diff
@@ -257,13 +282,14 @@ public void setSendTimeout(long sendTimeout) {
...
+	 * {@link KafkaProducerMessageHandler} should wait wait for send operation results.
+	 * Defaults to the kafka {@code delivery.timeout.ms} property + 5 seconds. The timeout
+	 * is applied only in {@link #sync} mode.
 	 * @param sendTimeoutExpression the {@link Expression} for timeout to wait for result
-	 * fo send operation.
+	 * for a send operation.
 	 * @since 2.1.1
 	 */
-	public void setSendTimeoutExpression(Expression sendTimeoutExpression) {
+	public final void setSendTimeoutExpression(Expression sendTimeoutExpression) {
```
artembilan, line 292: Looks like you have changed your mind and decided to not use this setter in the ctor.
I mean doesn't look like there is still a justification for a `final`...

artembilan (changes requested):

PR #3342, opened 2020-07-16
GH-3334: Add "embedded reaper" into CorrelationMH

### LINKED ISSUE #3334: Aggregator - Orphaned Groups Don't Time Out After Restart

All

**Describe the bug**

When using an external message group store, orphaned groups are not timed out.
https://stackoverflow.com/questions/62748832

**Expected behavior**

When there is a group timeout, and a persistent group store, the aggregator should schedule a timeout for all existing groups that were created by this aggregator. We don't currently have a mechanism to determine this; it could be a problem when the same store is used for different aggregators - we shouldn't schedule a timeout for groups we don't "own".

```java
@SpringBootApplication
public class So62748832Application {

	public static void main(String[] args) throws InterruptedException {
		ConfigurableApplicationContext ctx = SpringApplication.run(So62748832Application.class, args);
		Thread.sleep(10000);
		ctx.close();
	}

	@Bean
	public IntegrationFlow flow(RedisConnectionFactory cf) {
		return f -> f.aggregate(agg -> agg.correlationExpression("'foo'")
				.messageStore(store(cf))
				.sendPartialResultOnExpiry(true)
				.groupTimeout(15000))
				.handle(System.out::println);
	}

	@Bean
	MessageGroupStore store(RedisConnectionFactory connectionFactory) {
		return new RedisMessageStore(connectionFactory);
	}

//	@Bean
	@DependsOn("flow")
	public ApplicationRunner runner(@Qualifier("flow.input") MessageChannel channel) {
		return args -> channel.send(new GenericMessage<>("foo"));
	}

}
```

Run it once with the runner bean uncommented. Run again with it commented - no timeout.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3334

* Add `expireTimeout` property into `AbstractCorrelatingMessageHandler`
to call newly introduced `purgeOrphanedGroups()` API for removing old
groups from the store
* Add `expireDuration` to perform `purgeOrphanedGroups()` task periodically

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.aggregator;
 
+import java.time.Duration;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -53,6 +54,7 @@ import org.springframework.integration.support.AbstractIntegrationMessageBuilder
 import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.LockRegistry;
 import org.springframework.integration.util.UUIDConverter;
+import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageDeliveryException;
@@ -82,6 +84,11 @@ import org.springframework.util.ObjectUtils;
  * Use proper {@link CorrelationStrategy} for cases when same
  * {@link org.springframework.integration.store.MessageStore} is used
  * for multiple handlers to ensure uniqueness of message groups across handlers.
+ * <p>
+ * When the {@link #expireTimeout} is more than 0, the groups which older then this timeout
+ * is purged from the store on start up (or when {@link #purgeOrphanedGroups()} is called).
+ * If {@link #expireDuration} is provide, the task is scheduled to perform
+ * {@link #purgeOrphanedGroups()} periodically.
  *
  * @author Iwein Fuld
  * @author Dave Syer
@@ -131,6 +138,11 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 	private Expression groupTimeoutExpression;
 
 	private List<Advice> forceReleaseAdviceChain;
+
+	private long expireTimeout;
+
+	@Nullable
+	private Duration expireDuration;
 
 	private MessageGroupProcessor forceReleaseProcessor = new ForceReleaseMessageGroupProcessor();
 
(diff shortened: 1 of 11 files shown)


artembilan:
If this is OK, I'll go ahead with Java DSL and XML options.
Plus docs, of course.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -310,6 +322,32 @@ public void setReleaseLockBeforeSend(boolean releaseLockBeforeSend) {
 		this.releaseLockBeforeSend = releaseLockBeforeSend;
 	}
 
+	/**
+	 * Configure a timeout in milliseconds for purging old orphaned groups from the store.
+	 * Used on startup and if {@link #expireDuration} is provided the task for {@link #purgeOrphanedGroups()}
+	 * is scheduled with that period.
+	 * The {@link #forceReleaseProcessor} is used to process those expired groups according
+	 * the "force complete" options.
```
garyrussell, line 330:
```suggestion
	 * Used on startup and when an {@link #expireDuration} is provided, the task for running
	 * {@link #purgeOrphanedGroups()} is scheduled with that period.
	 * The {@link #forceReleaseProcessor} is used to process those expired groups according
	 * the "force complete" options. A group can be orphaned if a persistent message group
	 * store is used and no new messages arrive for that group after a restart.
```

garyrussell (changes requested):
LGTM to proceed with other changes (and fix travis checkstyle).

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -337,6 +338,18 @@ public void setExpireTimeout(long expireTimeout) {
 		this.expireTimeout = expireTimeout;
 	}
 
+	/**
+	 * Configure a {@link Duration} (ini millis) how often to clean up old orphaned groups from the store.
```
garyrussell, line 342:
```suggestion
	 * Configure a {@link Duration} (in millis) how often to clean up old orphaned groups from the store.
```

src/reference/asciidoc/aggregator.adoc
```diff
@@ -535,6 +536,17 @@ Timed-out groups are either discarded or a partial release occurs (based on `sen
...
+Starting version 5.4, the aggregator (and resequencer) can be configured for so-called "embedded reaper".
+The `expireTimeout` (if greater than `0`) indicates how old groups in the store should be purged.
+The `purgeOrphanedGroups()` method is called on start up and together with the provided `expireDuration` periodically within a scheduled task.
+This method is also can be called externally at any time.
+The expiration logic is fully delegated into a `forceComplete(MessageGroup)` functionality according the provided expiration options mentioned above.
+Such a periodic purge functionality is useful when a message store is needed to be cleaned up from those old groups which are not going to be released any more with regular message arrival logic.
+In most cases this happens after an application start up.
+The functionality is similar to the `MessageGroupStoreReaper` with a scheduled task, but provides a convenient way to deal with old groups withing specific components.
+The `MessageGroupStore` is still have to be provided exclusively for the current correlation endpoint.
+Otherwise one aggregator may purge groups from another.
```
garyrussell, line 549:
```suggestion
Starting with version 5.4, the aggregator (and resequencer) can be configured to expire orphaned groups (groups in a persistent message store that might not otherwise be released).
The `expireTimeout` (if greater than `0`) indicates that groups older than this value in the store should be purged.
The `purgeOrphanedGroups()` method is called on start up and, together with the provided `expireDuration`, periodically within a scheduled task.
This method is also can be called externally at any time.
The expiration logic is fully delegated to the `forceComplete(MessageGroup)` functionality according to the provided expiration options mentioned above.
Such a periodic purge functionality is useful when a message store is needed to be cleaned up from those old groups which are not going to be released any more with regular message arrival logic.
In most cases this happens after an application restart, when using a persistent message group store.
The functionality is similar to the `MessageGroupStoreReaper` with a scheduled task, but provides a convenient way to deal with old groups within specific components, when using group timeout instead of a reaper.
The `MessageGroupStore` must be provided exclusively for the current correlation endpoint.
Otherwise one aggregator may purge groups from another.
With the aggregator, groups expired using this technique will either be discarded or released as a partial group, depending on the `expireGroupsUponCompletion` property.
```

src/reference/asciidoc/whats-new.adoc
```diff
@@ -50,3 +50,6 @@ See <<./ip.adoc#ip-collaborating-adapters,Collaborating Channel Adapters>> and <
 
 The one-way messaging gateway (the `void` method return type) now sets a `nullChannel` explicitly into the `replyChannel` header to ignore any possible downstream replies.
 See <<./gateway.adoc#gateway-default-reply-channel,Setting the Default Reply Channel>> for more information.
+
+The aggregator (and resequencer) has now an "embedded reaper" support functionality provided via an `expireTimeout` and `expireDuration` options.
```
garyrussell, line 54:
```suggestion
The aggregator (and resequencer) can now expire orphaned groups (groups in a persistent store where no new messages arrive after an application restart).
```

garyrussell (changes requested):

garyrussell:
Some problem with Flux aggregator test in travis - maybe not related but suspicious given this code is in the aggregator.

artembilan:
No, not related. I so it failing here on Travis before.
Probably really some timing issue when CPU is too busy with other tasks...

PR #3346, opened 2020-07-17
GH-3344: Treat kotlin.Unit return as null in MMIH

### LINKED ISSUE #3344: Kotlin DSL: handle(handler: (P, MessageHeaders) -> Any) endpoint should not require output channel if lambda returns void
**Expected Behavior**

`handle<T> { payload, headers -> .. }` should work without output channel configured if lambda returns `void`. 
Like `handle { message -> .. }` currently does.

**Current Behavior**

```
integrationFlow(inputChannel) {
    wireTap {
        transform(Transformers.toJson(NODE))
        handle<ObjectNode> { payload, headers ->
            logger.debug(...)
        }
    }
    ...
}
```
throws `DestinationResolutionException: no output-channel or replyChannel header available`. While

```
integrationFlow(inputChannel) {
    wireTap {
        transform(Transformers.toJson(NODE))
        handle { message ->
            logger.debug(...)
        }
    }
    ...
}
```
works as intended.

**Context**
Now I am using this workaround, not sure if it is the best way to go:
```
handle<ObjectNode> { payload, headers ->
    logger.debug(...)
}
channel(nullChannel)
```

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3344

When function lambda doesn't return anything (e.g. a `void` method call is the last one),
Kotlin produces a `kotlin.Unit` instance as a return value which is not null and produced
as a reply message payload.

* Fix `MessagingMethodInvokerHelper` to treat a `kotlin.Unit` as `null` for reply
making Kotlin lambdas working the same way as Java lambdas when we don't return anything
from from there

**Cherry-pick to `5.3.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -1102,7 +1102,9 @@ public class MessagingMethodInvokerHelper extends AbstractExpressionEvaluator im
 			}
 			try {
 				Object result = this.invocableHandlerMethod.invoke(message);
-				if (result != null && result.getClass().getName().equals("kotlin.Unit")) {
+				if (result != null
+						&& org.springframework.integration.util.ClassUtils.isKotlinUnit(result.getClass())) {
+
 					result = null;
 				}
 				return result;
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -576,6 +576,7 @@ public class MessagingMethodInvokerHelper extends AbstractExpressionEvaluator im
 		localHandlerMethodFactory.afterPropertiesSet();
 	}
 
+	@Nullable
 	private Object invokeHandlerMethod(HandlerMethod handlerMethod, ParametersWrapper parameters) {
 		try {
 			return handlerMethod.invoke(parameters);
@@ -1093,13 +1094,18 @@ public class MessagingMethodInvokerHelper extends AbstractExpressionEvaluator im
 			this.invocableHandlerMethod = newInvocableHandlerMethod;
 		}
 
+		@Nullable
 		public Object invoke(ParametersWrapper parameters) {
 			Message<?> message = parameters.getMessage();
 			if (this.canProcessMessageList) {
 				message = new MutableMessage<>(parameters.getMessages(), parameters.getHeaders());
 			}
 			try {
-				return this.invocableHandlerMethod.invoke(message);
+				Object result = this.invocableHandlerMethod.invoke(message);
+				if (result != null && result.getClass().getName().equals("kotlin.Unit")) {
+					result = null;
+				}
+				return result;
 			}
 			catch (RuntimeException ex) { // NOSONAR no way to handle conditional catch according Sonar rules
 				throw ex;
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -1093,13 +1094,18 @@ void setInvocableHandlerMethod(InvocableHandlerMethod newInvocableHandlerMethod)
...
+		@Nullable
 		public Object invoke(ParametersWrapper parameters) {
 			Message<?> message = parameters.getMessage();
 			if (this.canProcessMessageList) {
 				message = new MutableMessage<>(parameters.getMessages(), parameters.getHeaders());
 			}
 			try {
-				return this.invocableHandlerMethod.invoke(message);
+				Object result = this.invocableHandlerMethod.invoke(message);
+				if (result != null && result.getClass().getName().equals("kotlin.Unit")) {
```
garyrussell, line 1105: Can we not capture this condition as a final `boolean` by checking the return type in a CTOR (to avoid running this code on every non-null return)?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
artembilan: Well, we can avoid something indeed, when a return type is `void`  or `Unit`. However in this case the return type is `Object` so end-user can return `null` ()or just don't have `return` in the function lambda and we don't produce a reply message.
In this case Kotlin for non-return in function lambda produces for us that `Unit` which is not a `null` apparently.
So, we produce some weird reply message.
Therefore even if we have some final checks for return type, we still would have some logic to check the returned value to be sure that we don't produce a reply message with `Unit` payload.

To summarize: when have `void` return this `result` is always going to be `null`, therefore we don't go to check a class name for `Unit`. If the value not null, it already doesn't matter what `void` check we have in advance: we still have to check this `result` for `Unit` type.

Does it make sense?

garyrussell: Could we at least do something like...
```java
if (result != null && this.isKotlinLambda && result.getClass().getName().equals("kotlin.Unit")) {
```

?

This `result.getClass().getName().equals("kotlin.Unit")` just seems like a lot of overhead for this one corner case and not good for the 99.9% of calls.

artembilan: Ah! OK. Makes sense.
I'll think what we can do...
Thanks for the pointer!

artembilan:
@garyrussell ,

how about this solution?

garyrussell:
Yes, that's better but you've tagged it `@since 5.4` so no more backport?

spring-integration-core/src/main/java/org/springframework/integration/util/ClassUtils.java
```diff
@@ -247,4 +264,14 @@ public static boolean isKotlinFaction1(Class<?> aClass) {
 		return KOTLIN_FUNCTION_1_CLASS != null && KOTLIN_FUNCTION_1_CLASS.isAssignableFrom(aClass);
 	}
 
+	/**
+	 * Check if class is {@code kotlin.Unit}.
+	 * @param aClass the {@link Class} to check.
+	 * @return true if class is a {@code kotlin.Unit} implementation.
+	 * @since 5.4
```
artembilan, line 271:
```suggestion
	 * @since 5.3.2
```

artembilan:
Fixed here: https://github.com/spring-projects/spring-integration/pull/3346/commits/db6786c462b7425f7b866461511f418ad7c23347.

Thanks for pointing that out!

garyrussell:
And cherry-picked.

PR #3347, opened 2020-07-17
GH-3340: IntegrationEvents  -add getSourceAsType()

### LINKED ISSUE #3340: Add server url property to MqttIntegrationEvent
**Expected Behavior**
`MqttIntegrationEvent` should have `server` property or other means to find what MQTT server caused it.

**Current Behavior**
`MqttIntegrationEvent` does not have `server` property.

**Context**
I have multiple inbound MQTT adapters connected to different servers. Right now it is impossible to find what server caused  application event. For example `MqttConnectionFailedEvent` may not have any mentions of failed server depending on type of the underlying exception.

`AbstractMqttMessageDrivenChannelAdapter.getUrl()` is protected. Also I am not sure if it is set correctly if urls are configured through `MqttConnectOptions`. At the very least `AbstractMqttMessageDrivenChannelAdapter` should log server url on errors.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3340

- add common super-interface for MQTT components
- add `getConnectionInfo()` so users can examine server URIs etc

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/DefaultMqttPahoClientFactory.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/DefaultMqttPahoClientFactory.java
@@ -16,6 +16,12 @@
 
 package org.springframework.integration.mqtt.core;
 
+import java.util.List;
+import java.util.Properties;
+
+import javax.net.SocketFactory;
+import javax.net.ssl.HostnameVerifier;
+
 import org.eclipse.paho.client.mqttv3.IMqttAsyncClient;
 import org.eclipse.paho.client.mqttv3.IMqttClient;
 import org.eclipse.paho.client.mqttv3.MqttAsyncClient;
@@ -23,6 +29,8 @@ import org.eclipse.paho.client.mqttv3.MqttClient;
 import org.eclipse.paho.client.mqttv3.MqttClientPersistence;
 import org.eclipse.paho.client.mqttv3.MqttConnectOptions;
 import org.eclipse.paho.client.mqttv3.MqttException;
+import org.eclipse.paho.client.mqttv3.MqttMessage;
+import org.eclipse.paho.client.mqttv3.MqttTopic;
 
 import org.springframework.util.Assert;
 
@@ -42,6 +50,8 @@ public class DefaultMqttPahoClientFactory implements MqttPahoClientFactory {
 	private MqttClientPersistence persistence;
 
 	private ConsumerStopAction consumerStopAction = ConsumerStopAction.UNSUBSCRIBE_CLEAN;
+
+	private List<String> serverUris;
 
 	/**
 	 * Set the persistence to pass into the client constructor.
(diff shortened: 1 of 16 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -54,14 +55,35 @@
...
+	/**
+	 * Deprecated - do not use; implementations must maintain the url or other connection
+	 * information.
+	 * @deprecated in favor of
+	 * {@link #AbstractMqttMessageDrivenChannelAdapter(String, String...)}.
+	 * @param url the url.
+	 * @param clientId the client id.
+	 * @param topics the topics.
+	 */
+	@Deprecated
```
artembilan, line 78: Yeah... I didn't mean to deprecate per-channel-adapter URL configuration.
My idea was just to deprecate a `getUrl()` since it not always reflect a connection reality.

Since we have now that `MqttConnectOptions getConnectionInfo()` may be it is better to clone its instance and modify its `serverURIs` with the provided URL. So, we still can keep existing apps without respective modification for per-channel-adapter-client-factory.

WDYT?

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
garyrussell: Good point - I hadn't thought of the shared factory.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/DefaultMqttPahoClientFactory.java
```diff
@@ -95,9 +105,270 @@ public void setConnectionOptions(MqttConnectOptions options) {
...
+	private class ExtendedOptions extends MqttConnectOptions {
+
+		@Override
+		public char[] getPassword() {
+			return DefaultMqttPahoClientFactory.this.options.getPassword();
+		}
+
+		@Override
+		public void setPassword(char[] password) {
+			DefaultMqttPahoClientFactory.this.options.setPassword(password);
```
artembilan, line 130: M-m-m. May be this delegate should be just "read-only"?
And all the setters should throw `UnsupportedOperationException` instead of delegation?

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -89,42 +94,91 @@
...
+	 * @param clientId The client id.
+	 * @param clientFactory The client factory.
+	 * @param topic The topic(s).
+	 * @since 5.4
+	 */
+	public MqttPahoMessageDrivenChannelAdapter(List<String> serverUris, String clientId,
+			MqttPahoClientFactory clientFactory, String... topic) {
+		super(clientId, topic);
+		this.clientFactory = clientFactory;
+		this.clientFactory.setServerUris(serverUris);
```
artembilan, line 129: I wouldn't do this. It is mutation of an external shared factory.
What drove you to drop that `url`-based configuration?
I thought we just need to have a consistency on the read side...

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
garyrussell: Oh, you're right - need to rethink.

The reason is if you want to connect to a cluster, the only way you can do it is via the connect options - here I am restoring the functionality, but enhanced to allow cluster per adapter with a single factory.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -174,6 +183,16 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
 		this.applicationEventPublisher = applicationEventPublisher; // NOSONAR (inconsistent synchronization)
 	}
 
+	@Override
+	public MqttConnectOptions getConnectionInfo() {
+		MqttConnectOptions options = this.clientFactory.getConnectionOptions();
+		if (options.getServerURIs() == null && getUrl() != null) {
+			options = MqttUtils.cloneConnectOptions(options);
+			options.setServerURIs(new String[] { getUrl() });
```
artembilan, line 191: Better to extract local variable for `URL` to avoid noise from Sonar that we defer NPE calling the method several times.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -258,6 +277,7 @@ public void removeTopic(String... topic) {
 		}
 	}
 
+	@SuppressWarnings("deprecation")
```
artembilan, line 280: I think there is no deprecations any more.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -68,6 +72,8 @@
 
 	private final MqttPahoClientFactory clientFactory;
 
+	private Supplier<List<String>> uriSupplier;
```
artembilan, line 75: Looks like this one is out of use

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/MqttPahoMessageHandler.java
```diff
@@ -92,13 +94,14 @@ public MqttPahoMessageHandler(String url, String clientId, MqttPahoClientFactory
 	 * @param clientFactory The client factory.
 	 * @since 4.1
 	 */
+	@SuppressWarnings("deprecation")
```
artembilan, line 97: I don't see deprecations any more

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/MqttPahoMessageHandler.java
```diff
@@ -154,6 +157,19 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
 		this.applicationEventPublisher = applicationEventPublisher;
 	}
 
+	@Override
+	public MqttConnectOptions getConnectionInfo() {
+		MqttConnectOptions options = this.clientFactory.getConnectionOptions();
+		if (options.getServerURIs() == null && getUrl() != null) {
+			if (options.getServerURIs() == null && getUrl() != null) {
```
artembilan, line 164: Looks like duplication for this `if`s

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/support/MqttUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * Clone the {@link MqttConnectOptions}, except the serverUris.
+	 * @param options the options to clone.
+	 * @return the clone.
+	 */
+	public static MqttConnectOptions cloneConnectOptions(MqttConnectOptions options) {
+		MqttConnectOptions options2 = new MqttConnectOptions();
+		BeanUtils.copyProperties(options, options2, "password", "serverURIs");
+		if (options.getPassword() != null) {
```
artembilan, line 43: Why does `password` have a special treatment?

src/reference/asciidoc/mqtt.adoc
```diff
@@ -403,3 +403,14 @@ Certain application events are published by the adapters.
 * `MqttSubscribedEvent` - published by the inbound adapter after subscribing to the topics.
 
 These events can be received by an `ApplicationListener<MqttIntegrationEvent>` or with an `@EventListener` method.
+
+To determine the source of an event, use the following; you can check the bean name and/or the connect options (to access the server URIs etc).
+
+====
+[source, java]
+----
+MqttPahoComponent source = event.getSourceAsType();
```
artembilan, line 412: Would be great to mention such a type requirement on the `MqttIntegrationEvent` class JavaDocs.

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/support/MqttUtils.java
garyrussell: `setPassword` clones the `Char[]`. Same thing with `serverURIs` - but we know they are null.

src/reference/asciidoc/mqtt.adoc
garyrussell: I don't see how we can -  the event is not aware of Paho - we have an open issue to migrate to another client which will have a different type.

artembilan: OK. Can we then document it on Paho components about such a specific for events?

PR #3350, opened 2020-07-28
Addition of the inbound part of Redis Stream support.

akuma8:
https://github.com/spring-projects/spring-integration/issues/3226
This commit is for a first review since tests are not yet added.

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
@@ -65,7 +65,7 @@ public class ReactiveRedisStreamMessageProducer extends MessageProducerSupport {
 
 	private StreamReceiver<String, ?> streamReceiver;
 
-	private ReadOffset readOffset = ReadOffset.lastConsumed();
+	private ReadOffset readOffset = ReadOffset.latest();
 
 	private boolean extractPayload = true;
 
@@ -88,7 +88,10 @@ public class ReactiveRedisStreamMessageProducer extends MessageProducerSupport {
 	}
 
 	/**
-	 * Define the offset from which we want to read message. By defaut the {@link ReadOffset#lastConsumed()} is used.
+	 * Define the offset from which we want to read message. By default the {@link ReadOffset#latest()} is used.
+	 * {@link ReadOffset#latest()} is equal to '$', which is the Id used with {@code XREAD} to get new data added to
+	 * the stream. Note that when switching to the Consumer Group feature, we set it to
+	 * {@link ReadOffset#lastConsumed()} if it is still equal to {@link ReadOffset#latest()}.
 	 * @param readOffset the desired offset
 	 */
 	public void setReadOffset(ReadOffset readOffset) {
@@ -174,9 +177,8 @@ public class ReactiveRedisStreamMessageProducer extends MessageProducerSupport {
 		Flux<Message<?>> messageFlux;
 
 		if (StringUtils.isEmpty(this.consumerName)) {
-			messageFlux = this.streamReceiver
-					.receive(offset)
-					.map(event -> getMessageBuilderFactory().withPayload(this.extractPayload ? event.getValue() : event)
+			messageFlux = this.streamReceiver.receive(offset).map(event ->
+					getMessageBuilderFactory().withPayload(this.extractPayload ? event.getValue() : event)
 							.setHeader(RedisHeaders.STREAM_KEY, event.getStream())
 							.setHeader(RedisHeaders.STREAM_MESSAGE_ID, event.getId())
 							.build());
@@ -206,21 +208,36 @@ public class ReactiveRedisStreamMessageProducer extends MessageProducerSupport {
 
 			Consumer consumer = Consumer.from(this.consumerGroup, this.consumerName);
 
-			messageFlux = this.streamReceiver
-					.receive(consumer, offset)
-					.map(event -> {
-						Message<?> message = getMessageBuilderFactory().withPayload(this.extractPayload ? event.getValue() : event)
+			if (offset.getOffset().equals(ReadOffset.latest())) {
+				// for consumer group offset id should be equal '>'
+				offset = StreamOffset.create(this.streamKey, ReadOffset.lastConsumed());
+			}
+
+			if (this.autoAck) {
+				messageFlux = this.streamReceiver.receiveAutoAck(consumer, offset).map(event ->
+						getMessageBuilderFactory()
+								.withPayload(this.extractPayload ? event.getValue() : event)
 								.setHeader(RedisHeaders.STREAM_KEY, event.getStream())
 								.setHeader(RedisHeaders.STREAM_MESSAGE_ID, event.getId())
 								.setHeader(RedisHeaders.CONSUMER_GROUP, this.consumerGroup)
 								.setHeader(RedisHeaders.CONSUMER, this.consumerName)
-								.build();
-						if (this.autoAck) {
-							message.getHeaders().put(IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK,
-									this.reactiveStreamOperations.acknowledge(this.consumerGroup, event).subscribe());
-						}
-						return message;
-					});
+								.build()
+				);
+			}
+			else {
+				messageFlux = this.streamReceiver.receive(consumer, offset).map(event ->
+						getMessageBuilderFactory()
+								.withPayload(this.extractPayload ? event.getValue() : event)
+								.setHeader(RedisHeaders.STREAM_KEY, event.getStream())
+								.setHeader(RedisHeaders.STREAM_MESSAGE_ID, event.getId())
+								.setHeader(RedisHeaders.CONSUMER_GROUP, this.consumerGroup)
+								.setHeader(RedisHeaders.CONSUMER, this.consumerName)
+								.setHeader(IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK,
+										this.reactiveStreamOperations.acknowledge(this.consumerGroup, event)
+												.subscribe())
+								.build()
+				);
+			}
 		}
 
 		subscribeToPublisher(messageFlux);
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
@@ -16,24 +16,22 @@
 
 package org.springframework.integration.redis.inbound;
 
-import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 
 import org.springframework.data.redis.connection.ReactiveRedisConnectionFactory;
 import org.springframework.data.redis.connection.stream.Consumer;
 import org.springframework.data.redis.connection.stream.ReadOffset;
+import org.springframework.data.redis.connection.stream.StreamInfo;
 import org.springframework.data.redis.connection.stream.StreamOffset;
+import org.springframework.data.redis.core.ReactiveRedisTemplate;
+import org.springframework.data.redis.core.ReactiveStreamOperations;
 import org.springframework.data.redis.core.StreamOperations;
+import org.springframework.data.redis.serializer.RedisSerializationContext;
 import org.springframework.data.redis.serializer.StringRedisSerializer;
 import org.springframework.data.redis.stream.StreamReceiver;
-import org.springframework.expression.EvaluationContext;
-import org.springframework.expression.Expression;
-import org.springframework.expression.common.LiteralExpression;
+import org.springframework.integration.IntegrationMessageHeaderAccessor;
 import org.springframework.integration.endpoint.MessageProducerSupport;
-import org.springframework.integration.expression.ExpressionUtils;
 import org.springframework.integration.redis.support.RedisHeaders;
-import org.springframework.integration.support.MessageBuilder;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.util.Assert;
@@ -44,8 +42,9 @@ import reactor.core.publisher.Flux;
 /**
  * A {@link MessageProducerSupport} for Redis that reads message from a Redis Stream and publishes it to the provided
  * output channel.
- * By default this adapter reads message as a standalone client {@code XREAD} Redis command but we can switch to a
- * Consumer Group feature {@code XREADGROUP} by setting {@link #consumerGroupName} and {@link #consumerName} fields.
+ * By default this adapter reads message as a standalone client {@code XREAD} (Redis command) but can be switched to a
+ * Consumer Group feature {@code XREADGROUP} by setting {@link #consumerName} field. By default the Consumer Group name
+ * is the id the bean {@link #getBeanName()}
  *
  * @author Attoumane Ahamadi
  *
@@ -55,9 +54,11 @@ public class ReactiveRedisStreamMessageProducer extends MessageProducerSupport {
 
 	private final ReactiveRedisConnectionFactory reactiveConnectionFactory;
 
-	private final Expression streamKeyExpression;
+	private final String streamKey;
 
-	private EvaluationContext evaluationContext;
+	private ReactiveRedisTemplate<String, ?> reactiveRedisTemplate;
+
+	private ReactiveStreamOperations<String, ?, ?> reactiveStreamOperations;
 
 	private StreamReceiver.StreamReceiverOptions<String, ?> streamReceiverOptions = StreamReceiver
 			.StreamReceiverOptions.builder().pollTimeout(Duration.ZERO).build();
@@ -68,25 +69,22 @@ public class ReactiveRedisStreamMessageProducer extends MessageProducerSupport {
 
 	private boolean extractPayload = true;
 
-	private String streamKey;
+	private boolean autoAck = true;
 
 	@Nullable
-	private String consumerGroupName;
+	private String consumerGroup;
 
 	@Nullable
 	private String consumerName;
 
-	private boolean createConsumerGroupIfNotExist;
+	private boolean createConsumerGroup;
 
 	public ReactiveRedisStreamMessageProducer(ReactiveRedisConnectionFactory reactiveConnectionFactory,
 			String streamKey) {
-		this(reactiveConnectionFactory, new LiteralExpression(streamKey));
-	}
-
-	public ReactiveRedisStreamMessageProducer(ReactiveRedisConnectionFactory connectionFactory,
-			Expression streamKeyExpression) {
-		this.reactiveConnectionFactory = connectionFactory;
-		this.streamKeyExpression = streamKeyExpression;
+		Assert.notNull(reactiveConnectionFactory, "'connectionFactory' must not be null");
+		Assert.hasText(streamKey, "'streamKey' must be set");
+		this.reactiveConnectionFactory = reactiveConnectionFactory;
+		this.streamKey = streamKey;
 	}
 
 	/**
@@ -106,17 +104,26 @@ public class ReactiveRedisStreamMessageProducer extends MessageProducerSupport {
 	}
 
 	/**
-	 * Set the name of the Consumer Group. It is possible to create that Consumer Group if desired, see:
-	 * {@link #createConsumerGroupIfNotExist}. Note that if this value is set, {@link #consumerName} should also be set
-	 * @param consumerGroupName the Consumer Group which this adapter should register to listen messages.
+	 * Set whether or not acknowledge message read in the Consumer Group. {@code true} by default.
+	 * @param autoAck the acknowledge option.
 	 */
-	public void setConsumerGroupName(@Nullable String consumerGroupName) {
-		this.consumerGroupName = consumerGroupName;
+	public void setAutoAck(boolean autoAck) {
+		this.autoAck = autoAck;
 	}
 
 	/**
-	 * If the Consumer Group is used, this value should be provided. Note that this value should be unique in the group
-	 * @param consumerName the conusmer name in the Consumer Group
+	 * Set the name of the Consumer Group. It is possible to create that Consumer Group if desired, see:
+	 * {@link #createConsumerGroup}. If not set, the defined bean name {@link #getBeanName()} is used.
+	 * @param consumerGroup the Consumer Group on which this adapter should register to listen messages.
+	 */
+	public void setConsumerGroup(@Nullable String consumerGroup) {
+		this.consumerGroup = consumerGroup;
+	}
+
+	/**
+	 * Set the name of the consumer. When a consumer name is provided, this adapter is switched to the Consumer Group
+	 * feature. Note that this value should be unique in the group
+	 * @param consumerName the consumer name in the Consumer Group
 	 */
 	public void setConsumerName(@Nullable String consumerName) {
 		this.consumerName = consumerName;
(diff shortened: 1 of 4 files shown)


akuma8:
@artembilan,
This is the inbound PR. I am working on adding tests while you are reviewing. 
Thanks

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+
+	private boolean createConsumerGroupIfNotExist;
+
+	public ReactiveRedisStreamMessageProducer(ReactiveRedisConnectionFactory reactiveConnectionFactory,
+			String streamKey) {
+		this(reactiveConnectionFactory, new LiteralExpression(streamKey));
+	}
+
+	public ReactiveRedisStreamMessageProducer(ReactiveRedisConnectionFactory connectionFactory,
+			Expression streamKeyExpression) {
```
artembilan, line 87: There is no reason in such an expression complexity. This component is the beginning of the flow, so there is no special context to really evaluate a stream key on demand.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+		this.extractPayload = extractPayload;
+	}
+
+	/**
+	 * Set the name of the Consumer Group. Is it possible to create that Consumer Group if desired, see:
+	 * {@link #createConsumerGroupIfNotExist}. Note that if this value is set, {@link #consumerName} should also be set
+	 * @param consumerGroupName the Consumer Group which this adapter should register to listen messages.
+	 */
+	public void setConsumerGroupName(@Nullable String consumerGroupName) {
+		this.consumerGroupName = consumerGroupName;
```
artembilan, line 114: We had a chat with Gary and decided to make the group name based on a `beanName`.
Therefore we only going to request a `consumerName` and only if that is provided we switch to a consumer group receiving. This convention therefore has to be documented on the class level JavaDocs and that `setConsumerName()`. And eventually in the docs, of course.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+		this.consumerName = consumerName;
+	}
+
+	/**
+	 * Create the Consumer Group if and only if it does not exist. During the
+	 * creation we also create the stream {@see MKSTREAM}. If the stream already exists {@code MKSTREAM}
+	 * has no effect.
+	 * @param createConsumerGroupIfNotExist specify if we should create the Consumer Group, {@code false} by default
+	 */
+	public void setCreateConsumerGroupIfNotExist(boolean createConsumerGroupIfNotExist) {
```
artembilan, line 131: Let's call it just `setCreateConsumerGroup()` !

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+	@Override
+	protected void onInit() {
+		super.onInit();
+		if (this.streamReceiverOptions != null) {
+			this.streamReceiver = StreamReceiver.create(this.reactiveConnectionFactory, this.streamReceiverOptions);
+		}
+		else {
+			this.streamReceiver = StreamReceiver.create(this.reactiveConnectionFactory);
+		}
+		this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
```
artembilan, line 154: Since we drop expression for the stream key, we don't need this context, too.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+		StreamOffset<String> offset = StreamOffset.create(this.streamKey, this.readOffset);
+
+		Flux<Message<?>> messageFlux;
+
+		if (StringUtils.isEmpty(this.consumerGroupName)) {
+			messageFlux = this.streamReceiver
+					.receive(offset)
+					.map(event -> MessageBuilder.withPayload(this.extractPayload ? event.getValue() : event)
+							.setHeader(RedisHeaders.STREAM_KEY, event.getStream())
+							.setHeader(RedisHeaders.STREAM_MESSAGE_ID, event.getId())
```
artembilan, line 174: Please, add also a consumer group and consumer name headers.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+		Assert.notNull(this.streamKey, "'streamKey' must not be null");
+
+		StreamOffset<String> offset = StreamOffset.create(this.streamKey, this.readOffset);
+
+		Flux<Message<?>> messageFlux;
+
+		if (StringUtils.isEmpty(this.consumerGroupName)) {
+			messageFlux = this.streamReceiver
+					.receive(offset)
+					.map(event -> MessageBuilder.withPayload(this.extractPayload ? event.getValue() : event)
```
artembilan, line 172: Not `MessageBuilder`, but `getMessageBuilderFactory()`

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+
+		Assert.notNull(this.streamKey, "'streamKey' must not be null");
+
+		StreamOffset<String> offset = StreamOffset.create(this.streamKey, this.readOffset);
+
+		Flux<Message<?>> messageFlux;
+
+		if (StringUtils.isEmpty(this.consumerGroupName)) {
+			messageFlux = this.streamReceiver
+					.receive(offset)
```
artembilan, line 171: See this API:
```
* Every record must be acknowledged using
* {@link org.springframework.data.redis.connection.ReactiveStreamCommands#xAck(ByteBuffer, String, String...)}
```
Therefore we need to have a ` boolean autoAck` option on this channel adapter and consider to build an `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` as a `SimpleAcknowledgment` implementation.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	private void createConsumerGroup() {
+		if (this.createConsumerGroupIfNotExist) {
+			Assert.hasText(this.consumerGroupName, "'consumerGroupName' must be set");
+			Assert.hasText(this.consumerName, "'consumerName' must be set");
+
+			this.reactiveConnectionFactory.getReactiveConnection()
+					.streamCommands()
+					.xGroupCreate(ByteBuffer.wrap(this.streamKey.getBytes(StandardCharsets.UTF_8)),
```
artembilan, line 203: Let's make this as a part of the `ReactiveStreamOperations.createGroup()` reactive stream!

I mean it is better to do such a `Mono` call in that `doStart()` and then do `thenMany()`  for that `this.streamReceiver.receive()`.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+					.map(event -> MessageBuilder.withPayload(this.extractPayload ? event.getValue() : event)
+							.setHeader(RedisHeaders.STREAM_KEY, event.getStream())
+							.setHeader(RedisHeaders.STREAM_MESSAGE_ID, event.getId())
+							.build());
+		}
+		else {
+			createConsumerGroup();
+			Consumer consumer = Consumer.from(this.consumerGroupName, this.consumerName);
+			messageFlux = this.streamReceiver
+					.receiveAutoAck(consumer, offset)
```
artembilan, line 181: We should consider an `autoAck == false` here as well.

artembilan (changes requested):
Some review while you're writing tests...

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,207 @@
...
+	 * If the Consumer Group is used, this value should be provided. Note that this value should be unique in the group
+	 * @param consumerName the conusmer name in the Consumer Group
+	 */
+	public void setConsumerName(@Nullable String consumerName) {
+		this.consumerName = consumerName;
+	}
+
+	/**
+	 * Create the Consumer Group if and only if it does not exist. During the
+	 * creation we also create the stream {@see MKSTREAM}. If the stream already exists {@code MKSTREAM}
```
artembilan, line 127: Looks like JavaDocs is with errors:
```
/home/travis/build/spring-projects/spring-integration/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java:127: error: no tag name after @

	 * creation we also create the stream {@see MKSTREAM}. If the stream already exists {@code MKSTREAM}
```

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
artembilan: So, after more thoughts Gary shared with me an idea that users may use the same application as different instances for different groups.

So, let's leave this setter, but still let's try to make a default one based on bean id!

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
akuma8: I think I need a little briefing on whether or not to use `Expression`. Why and when to use `Expression` instead of `String`? In the outbound side I had to replace `String streamKey` by `Expression streamKeyExpression`. Why not here? 
Thanks

artembilan: We deal with expressions in the place where we have a dynamic context of some object, so the target value is resolved against that root object. On the outbound side it is a request message, therefore `streamKey` can be resolved at runtime every request we send to that outbound channel adapter. It also makes sense sometimes in polling inbound channel adapters when `receive()` is called by poller, so, when some environment in the application is changed in between polls, we get an actual value evaluating expression.

But if such an expression is going to evaluated only once during application startup, it does not give us any advantages to calculate the target value from the expression. Well, you still can use a SpEL, but as a matter of configuration phase - the target value will be populated to the inbound channel adapter only once anyway.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
akuma8: I'm involved in the development  of this feature because I really need it in my project. I planned to use it like this:
- `consumerGroupName = ${spring.application.name}`
- `consumerName = ${spring.application.name}-${instanceNumber}` (I am still wondering how to get the `instanceNumber` for an application deployed in Kubernetes? ^^)

@artembilan What do you mean by making a default setter based on bean id? Should I initialize `consumerGroupName` using the bean id? (I mean in the `onInit()` method, setting `consumerGroupName` if not provided by the client).

artembilan: That's true. In the `onInit()` you do `this.consumerGroupName = getBeanName()` if `consumerName` is provided and `this.consumerGroupName == null`.

I'm not familiar with Kubernetes and it fully not related to this project. 
I think what you do with Spring Boot properties is OK, only need to figure out what is there for that `instanceNumber` which I assume you borrowed from PCF.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
akuma8: @artembilan I think I will need your help for this. Also, I though that acknowledgement was only for the case where Consumer Group was involved.

artembilan: Yeah... Looks like you are right. Only consumer group needs an acknowledge: https://redis.io/topics/streams-intro.
Also see `ReactiveStreamOperations.acknowledge()` API: all of them require a `group` to ack the record.

Probably Spring Data team made a copy/paste mistake to mention ack on the regular `receive()`:
```
/**
 * Starts a Redis Stream consumer that consumes {@link Record records} from the {@link StreamOffset stream}. Records
 * are consumed from Redis and delivered on the returned {@link Flux} when requests are made on the Flux. The receiver
 * is closed when the returned {@link Flux} terminates.
 * <p/>
 * Every record must be acknowledged using
 * {@link org.springframework.data.redis.connection.ReactiveStreamCommands#xAck(ByteBuffer, String, String...)}
 *
 * @param streamOffset the stream along its offset.
 * @return Flux of inbound {@link Record}s.
 * @see StreamOffset#create(Object, ReadOffset)
 */
Flux<V> receive(StreamOffset<K> streamOffset);
```

artembilan: The `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` could be as simple as:
```
.setHeader(IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK, 
                     () -> this.redisStreamOperations.acknowledge(this.consumerGroup, event))
```

akuma8:
Hi @artembilan,
I am currently developping tests and would like to have your opinion on this issue:
I defined this bean: 
```
               @Bean
		public ReactiveMessageHandlerAdapter reactiveMessageHandlerAdapter(
				ReactiveRedisConnectionFactory connectionFactory) {

			ReactiveRedisStreamMessageHandler streamMessageHandler = new ReactiveRedisStreamMessageHandler(connectionFactory, STREAM_KEY); 

			return new ReactiveMessageHandlerAdapter(streamMessageHandler);
		}
```
But when I call: 
```
reactiveMessageHandlerAdapter.handleMessage(....); // NPE is thrown here because streamMessageHandler#reactiveMessageHandlerAdapter is null
```
With: 
```
ReactiveRedisStreamMessageHandler streamMessageHandler = new ReactiveRedisStreamMessageHandler(connectionFactory, STREAM_KEY); 
```
the `onInit()` method of  `streamMessageHandler` is not called so its `reactiveStreamOperations` is not initialized. I would like to know if it's a normal usage or should I review the  `ReactiveRedisStreamMessageHandler#reactiveStreamOperations` initialization? 
Thanks

artembilan:
You should have a `ReactiveRedisStreamMessageHandler` as a bean. This not a `ReactiveMessageHandlerAdapter` responsibility to initialize its delegate.

akuma8:
Hi @artembilan,
I completed outbound side tests by adding the test with the `List` class. 
I have an issue with inbound tests where one of them turn into an infinite loop. 
I am not enough familiar with the Reactive world, which explains my problems. 
Could you take a look and let me know? 
Thanks a lot

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,228 @@
...
+			messageFlux = this.streamReceiver
+					.receive(consumer, offset)
+					.map(event -> {
+						Message<?> message = getMessageBuilderFactory().withPayload(this.extractPayload ? event.getValue() : event)
+								.setHeader(RedisHeaders.STREAM_KEY, event.getStream())
+								.setHeader(RedisHeaders.STREAM_MESSAGE_ID, event.getId())
+								.setHeader(RedisHeaders.CONSUMER_GROUP, this.consumerGroup)
+								.setHeader(RedisHeaders.CONSUMER, this.consumerName)
+								.build();
+						if (this.autoAck) {
```
artembilan, line 218: That's something wrong: why do we need to add a manual ack callback header, if we do that when `autoAck`?
Maybe you mean exactly opposite?
Why don't you call `receiveAutoAck()` anymore?

spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java
```diff
@@ -0,0 +1,166 @@
...
+
+	@Autowired
+	ReactiveMessageHandlerAdapter messageHandler;
+
+	@Before
+	public void delKey() {
+		this.template.delete(STREAM_KEY).subscribe();
+	}
+
+	@AfterEach
```
artembilan, line 83: This is JUnit 5 feature. We can't use it with our Redis test suite yet because we stack there with JUnit 4 rules.
I mean something might be wrong with your tests since you combine JUnit 4 and 5.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,228 @@
...
+					.map(event -> getMessageBuilderFactory().withPayload(this.extractPayload ? event.getValue() : event)
+							.setHeader(RedisHeaders.STREAM_KEY, event.getStream())
+							.setHeader(RedisHeaders.STREAM_MESSAGE_ID, event.getId())
+							.build());
+		}
+		else {
+			Assert.notNull(this.consumerGroup, "'consumerGroup' must not be null");
+
+			if (this.createConsumerGroup) {
+				Assert.hasText(this.consumerName, "'consumerName' must be set");
```
artembilan, line 188: This is redundant: guarded by the previous `if (StringUtils.isEmpty(this.consumerName)) {`

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
```diff
@@ -0,0 +1,228 @@
...
+										.subscribe();
+							}
+							else {
+								this.reactiveStreamOperations.groups(this.streamKey)
+										.map(StreamInfo.XInfoGroup::groupName)
+										.filter(groupName -> groupName.equals(this.consumerGroup))
+										.switchIfEmpty(this.reactiveStreamOperations.createGroup(this.streamKey,
+												this.consumerGroup))
+										.subscribe();
+							}
```
artembilan, line 203: Why so complex logic?
Can't we just call `this.reactiveStreamOperations.createGroup()`?
And also I asked to not call `subscribe()`.
There is `Mono.thenMany()` for our reactive stream.

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
akuma8: How to handle the case where `autoAck` is `false` since `receiveAutoAck()` will always acknowledge?

artembilan: Right, we do `receiveAutoAck()` when `autoAck` is `true` and we add `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` when it is `false`.
And yes: when it is `false`, we perform `receive()`.

When `autoAck == true`, we call `receiveAutoAck()` and don't add header.
We do exactly opposite, when `autoAck == false`.

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java
akuma8: `this.reactiveStreamOperations.createGroup()` throws an exception if the group already exists:
```
Caused by: io.lettuce.core.RedisBusyException: BUSYGROUP Consumer Group name already exists
	at io.lettuce.core.internal.ExceptionFactory.createExecutionException(ExceptionFactory.java:126) ~[lettuce-core-6.0.0.M1.jar:6.0.0.M1]
```
That's why I try to verify it following this logic: 
 ```
  IF streamKey NOT exist
       CREATE the group and the stream
  ELSE 
       CHECK the groups subscribed to the stream // XinfoGroup
            IF the group is not absent 
                CREATE it
```

File: spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java (hunk 11 above)
Comment (akuma8) [reply spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java:218]: Right, a question. Why adding `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` header when `autoAck==false` since we do not want to acknowledge.

Comment (artembilan) [reply spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java:218]: Huh? We don't want to ack automatically. With that header we just defer an ack to end-user decision. 
This is typical messaging pattern: when we auto ack, we leave the durability for a message up to end-user, since message is going to be removed from the broker.
If we don't ack and defer to end-user, we still keep a message on the broker and it is going to be available for re-fetch after crash. This way durability is on broker and it is up to end-user to decide when to ack that processed message.

See here more explanation: https://stackoverflow.com/questions/11673361/jms-auto-acknowledge-when-is-it-acknowledged

File: spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java (hunk 14 above)
Comment (artembilan) [reply spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java:203]: If group already exists, and an exception is thrown, we always can fallback to the rest of the stream using `onErrorResume()`.
I understand that this is not what language does for us, but I feel like with just Reactor operators the logic should be simpler and we just do less calls over the network when group really doesn't exist.

Comment (akuma8) [reply spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java:203]: Sure, but I lack experience on Reactor.
Learning Reactor is in my TODO list.

Comment (artembilan) [reply spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java:203]: No problem! 
I can clean that up for you during merge.
Let's nail all the other concerns and we are good to go!

**Comment by akuma8** [comment]
Hi @artembilan,
I did requested changes. 
I still have the same issue with my tests. 
I think there is an issue with the configuration of `streamReceiver`. 
I breakpointed to check the content of `messageFlux` (result of the `this.streamReceiver.receive()`), I can see a `StackOverflow` exception, see attached image. 
StackOverflow
I don't kknow what I am doing wrong since I use a default `streamReceiver`:
```
StreamReceiver.StreamReceiverOptions<String, ?> streamReceiverOptions = StreamReceiver.StreamReceiverOptions.builder().pollTimeout(Duration.ZERO).build();
this.streamReceiver = StreamReceiver.create(this.reactiveConnectionFactory, this.streamReceiverOptions);
```

artembilan (commented):
Pulling locally for review and possible merge after cleaning up...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4ae6b52c001e3e06c334eec8370c8dd96d3b0fe4.

Please, pull the latest `master` locally to investigate what I did for the code.

I'm really looking forward for more tests since we have fixed the problem with connection factory.
Plus it would be great to have some docs in the `redis.adoc` on the matter.

Thank you for contribution!

akuma8:
Thanks for your investigation. 
Did the problem come from the connection factory? 
I am working on adding further tests. 
About documentation, as I don't speak english it will be hard for me to write a good documentation. 
I will try and let you know. 
Thank you

artembilan:
Hi @akuma8 !

Yes, I think the problem for hanging test suite was really about the second call for the `ConnectionFactory.afterPropertiesSet()` where we created uncontrolled `RedisClient`.

Sure! More tests are welcome!
Re. Docs: I'm not good in English, too: @garyrussell is always helping me to make my docs PRs better.
So, feel free to raise whatever you can do best and we together will bring it into the public!

Thank you!

PR #3353, opened 2020-07-30
AMQP: Support CorrelationData Message Headers

garyrussell:
In preparation for https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/303

Now that `CorrelationData` has a `Future<?>`, users might simply add
correlation data in a header and not receive confirm/return messages.

- No longer require channels for returns and confirms
- don't build the confirm message if there are no channels
- reduce the log level for no channels to DEBUG
- complete the user's future when a message is returned (async GW)

### DIFF
--- a/src/reference/asciidoc/amqp.adoc
+++ b/src/reference/asciidoc/amqp.adoc
@@ -1034,8 +1034,8 @@ When a publisher confirmation is received and correlation data is supplied, the 
 The payload of the confirmation is the correlation data as defined by this expression, and the message has its 'amqp_publishConfirm' header set to `true` (`ack`) or `false` (`nack`).
 For `nack` instances, an additional header (`amqp_publishConfirmNackCause`) is provided.
 Examples: `headers['myCorrelationData']`, `payload`.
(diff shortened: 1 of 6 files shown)


src/reference/asciidoc/amqp.adoc
```diff
@@ -1025,24 +1034,29 @@ When a publisher confirmation is received and correlation data is supplied, the
 The payload of the confirmation is the correlation data as defined by this expression, and the message has its 'amqp_publishConfirm' header set to `true` (`ack`) or `false` (`nack`).
 For `nack` instances, an additional header (`amqp_publishConfirmNackCause`) is provided.
 Examples: `headers['myCorrelationData']`, `payload`.
-If the expression resolves to a `Message<?>` instance (such as "`#this`"), the message emitted on the `ack`/`nack` channel is based on that message, with the additional headers added.
+If the expression resolves to a `Message<?>` instance (such as "`#this`"), the message emitted on the `ack`/`nack` channel is based on that message, with the Also see <<alternative-confirms-returns>>.
+additional headers added.
```
artembilan, line 1038: I think this is a copy/paste artefact.
Probably this new `Also...` sentence should go already after `additional headers added.`

artembilan (changes requested):
Docs typo.

PR #3360, opened 2020-08-07
GH-3358: Kafka IB Gateway - implement Pausable

### LINKED ISSUE #3358: Kafka Inbound Gateway should implement Pausable

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3358

Also add `isPaused()` to `Pausable`.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/core/Pausable.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/core/Pausable.java
@@ -46,7 +46,7 @@ public interface Pausable extends ManageableLifecycle {
 	/**
 	 * Check if the endpoint is paused.
 	 * @return true if paused.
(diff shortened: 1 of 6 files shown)


spring-integration-core/src/main/java/org/springframework/integration/core/Pausable.java
```diff
@@ -42,4 +43,14 @@
 	@ManagedOperation(description = "Resume the component")
 	void resume();
 
+	/**
+	 * Check if the endpoint is paused.
+	 * @return true if paused.
+	 * @since 5.3
```
artembilan, line 49:
```suggestion
	 * @since 5.4
```

artembilan (changes requested):

PR #3361, opened 2020-08-07
GH-3359: Add Timer to MessagingGatewaySupport

### LINKED ISSUE #3359: Add Message Timer metric to gateways

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3359

- replace removed simple counter with a `Timer`
- register DSL GPFB as beans so they can be discovered for wiring metrics

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayProxySpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayProxySpec.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.dsl;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.Executor;
@@ -43,7 +44,7 @@ import org.springframework.messaging.MessageChannel;
  *
  * @since 5.2
  */
-public class GatewayProxySpec {
+public class GatewayProxySpec implements ComponentsRegistration {
 
 	protected  static final SpelExpressionParser PARSER = new SpelExpressionParser(); // NOSONAR - final
 
@@ -285,6 +286,12 @@ public class GatewayProxySpec {
 			this.gatewayProxyFactoryBean.setGlobalMethodMetadata(this.gatewayMethodMetadata);
 		}
 		return this.gatewayProxyFactoryBean;
+	}
+
+	@Override
+	public Map<Object, String> getComponentsToRegister() {
+		return Collections.singletonMap(this.gatewayProxyFactoryBean, this.gatewayProxyFactoryBean.getBeanName()
+				+ ".gateway");
 	}
 
 }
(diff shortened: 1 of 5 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayProxySpec.java
```diff
@@ -43,7 +44,7 @@
  *
```
artembilan, line 44: Your name here, please
```suggestion
 * @author Gary Russell
```

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -1020,6 +1032,11 @@ protected void doStop() {
 		}
 	}
 
+	@Override
+	public void destroy() {
+		this.gatewayMap.values().forEach(gw -> gw.destroy());
```
artembilan, line 1037:
```suggestion
		this.gatewayMap.values().forEach(MethodInvocationGateway::destroy);
```

How do we live without this before?
Should we consider to back-port it as a bug?
What is a consequence not having this `destroy()`?

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -802,6 +857,13 @@ protected void doStop() {
 		}
 	}
 
+	@Override
+	public void destroy() {
+		this.timers.forEach(MeterFacade::remove);
```
artembilan, line 862: Ah! I got it.

Gary, take a look, please, what does that `@SuppressWarnings("deprecation")` do on this class?

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -690,6 +726,25 @@ private void sendMessageForReactiveFlow(MessageChannel requestChannel, Message<?
...
+	protected TimerFacade sendTimer() {
+		if (this.successTimer == null) {
+			this.successTimer = buildSendTimer(true, "none");
+		}
+		return this.successTimer;
+	}
+
+	protected TimerFacade buildSendTimer(boolean success, String exception) {
+		TimerFacade timer = this.metricsCaptor.timerBuilder(SEND_TIMER_NAME)
+				.tag("type", "source")
```
artembilan, line 738: Do we really talk here about `source`?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
garyrussell: It really is a message source - I thought it was better than introducing a new type `gateway` because that can be in or out.

PR #3363, opened 2020-08-11
Addition of test that consumes message with Consumer Group.

akuma8:
https://github.com/spring-projects/spring-integration/issues/3226

This is the final test for the inbound side.


artembilan:
Cool!

Let's see if we can make some docs for this feature!

PR #3377, opened 2020-09-09
GH-3376: Remove gauges on application ctx close

### LINKED ISSUE #3376: Gauges for the number of MessageChannels, MessageHandlers, and MessageSources may cause a BeanCreationNotAllowedException when the context is closed

This was reported against Spring Integration 5.2.6, but I believe older and newer versions are also affected.

**Describe the bug**

Spring Integration registers 3 gauges where the value is calculated by calling `getBeansOfType()` on the application context for `MessageChannel`, `MessageHandler`, and `MessageSource` beans. It does so in such a way that there are no bean dependency relationships between the meter registry and the `MessageChannel`, `MessageHandler`, and `MessageSource` beans. During context close processing, this allows those beans to be destroyed before the meter registry is destroyed. As a result query the gauges in the meter registry can result in bean creation being attempted and a `BeanCreationNotAllowedException` is then thrown.

**To Reproduce**

Courtesy of @tkvangorder, a sample that reproduces the problem is available [here](https://github.com/tkvangorder/spring-task-issue). There's also some background information in [this Spring Boot pull request](https://github.com/spring-projects/spring-boot/pull/22994).

**Expected behavior**

Gauges should be implemented or registered in such a way that the `MeterRegistry` can be used during application context close processing.

**Sample**

https://github.com/tkvangorder/spring-task-issue

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3376

The `MeterRegistry` may request meters on application shutdown.
The gauges for channels, handlers and message sources don't make sense
at the moment since all those beans are going to be destroyed.

* Remove gauges for channel, handler and message source numbers from the
`IntegrationManagementConfigurer.destroy()`


wilkinsona:
Is there an existing relationship between beans that will ensure that `IntegrationManagementConfigurer` is disposed before the Micrometer `MeterRegistry`? I couldn't spot anything but I'm not too familiar with the code so may well have missed something.

artembilan:
@wilkinsona ,

thank you for feedback!

So, let's see together what we have so far!

Since Micrometer is an `optional` dependency we can't make a hard class reference into that `MeterRegistry` everywhere we need metrics to expose. So, we have introduced a `facade` in front of it as a `MetricsCaptor`.
The further logic is like this then in the `IntegrationManagementConfigurer`:
```
public void afterSingletonsInstantiated() {
	Assert.state(this.applicationContext != null, "'applicationContext' must not be null");
	Assert.state(MANAGEMENT_CONFIGURER_NAME.equals(this.beanName), getClass().getSimpleName()
			+ " bean name must be " + MANAGEMENT_CONFIGURER_NAME);
	if (ClassUtils.isPresent("io.micrometer.core.instrument.MeterRegistry",
			this.applicationContext.getClassLoader())) {
		this.metricsCaptor = MicrometerMetricsCaptor.loadCaptor(this.applicationContext);
	}
```
And already that `MicrometerMetricsCaptor.loadCaptor()` does this:
```
MeterRegistry registry = applicationContext.getBean(MeterRegistry.class);
```

So, I think according your requirement there is no such a dependency injection relationship between `IntegrationManagementConfigurer` and `MeterRegistry`.

Maybe we should think about applying `@Conditional` on the inner `Micrometer`-specific `@Configuration` instead and then have an optional injection into the `IntegrationManagementConfigurer` bean definitiont. See an `IntegrationManagementConfiguration`.

Thanks again

artembilan:
@wilkinsona ,

any further feedback, please?

We probably need to see how this can make it into the upcoming release this Wednesday.

Thanks

wilkinsona:
I wonder if you could use a BeanFactoryPostProcessor or ImportBeanDefinitionRegistrar to register the `MetricsCaptor` implementation when Micrometer is on the classpath. It would be configured to depend on the `MeterRegistry` and could then be injected into `IntegrationManagementConfigurer` using an `ObjectProvider` so that it can be optional. This injection should then result in `IntegrationManagementConfigurer` being destroyed before the `MeterRegistry`.

artembilan:
OK. I have added a `MicrometerMetricsCaptorRegistrar implements ImportBeanDefinitionRegistrar` to satisfy dependency tree.

Please, take a look how is this now, @wilkinsona .

Thank you!

wilkinsona:
With the caveat that I haven't tried to run the code, those changes look good to me now.

artembilan:
Back-ported to `5.3.x` as https://github.com/spring-projects/spring-integration/commit/d0cab670ebc52d28abb9d8def8626e966af4ba55 after fixing some conflicts.

Back-ported to `5.2.x` as https://github.com/spring-projects/spring-integration/commit/a2a8764481c5433ddbc35a839bd35a947edadc0f with more conflict resolutions.

PR #3380, opened 2020-09-10
GH-3370: Remove synchronized from RemoteFileUtils

### LINKED ISSUE #3370: Don't make RemoteFileUtils.makeDirectories syncronized
5.3.1.RELEASE

**Describe the bug**
It caused by org.springframework.integration.file.remote.RemoteFileUtils#makeDirectories, which is synchronized static method, when there are lots of (S)ftp move operation concurrently and slow speed of network, all requests of AbstractRemoteFileOutboundGateway#mv are queued and observed as very slow speed.

https://stackoverflow.com/questions/62772002/outbound-gateway-executes-the-mv-command-very-slowly

**Expected behavior**
in api doc, RemoteFileUtils#makeDirectories is static not synchronized, 
https://docs.spring.io/spring-integration/docs/5.3.1.RELEASE/api/
However, in source code is static and syncronized.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3370

The `synchronized` on the `RemoteFileUtils.makeDirectories()` makes an application too
 slow, especially when we deal with different paths in different sessions

* Remove the `synchronized` from that method and rework `SftpSession.mkdir()`
to return `false` when "A file cannot be created if it already exists" exception
is thrown from the server.
Essentially make an `exists()` call to be sure that an exception is really related
to "file-already-exists" answer from the server

**Cherry-pick to 5.3.x, 5.2.x & 4.3.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileUtils.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -46,7 +46,7 @@ public final class RemoteFileUtils {
 	 * @param logger The logger.
 	 * @throws IOException Any IOException.
 	 */
-	public static synchronized <F> void makeDirectories(String path, Session<F> session, String remoteFileSeparator,
+	public static <F> void makeDirectories(String path, Session<F> session, String remoteFileSeparator,
 			Log logger) throws IOException {
 
 		if (!session.exists(path)) {
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileUtils.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileUtils.java
@@ -28,6 +28,7 @@ import org.springframework.integration.file.remote.session.Session;
  * Utility methods for supporting remote file operations.
  *
  * @author Gary Russell
+ * @author Artem Bilan
  *
  * @since 3.0
  *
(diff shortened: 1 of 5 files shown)


garyrussell (changes requested):

garyrussell:
I still think we need to check the result in RFU (for the FTP case).

artembilan:
> I still think we need to check the result in RFU 

Hm. Probably that `true/false` contract is not good. I may guess that `Session` API was designed after that `FTPClient`, so we blindly got a `boolean` result which confuses us now: what really could be the case when we got `false` reply, but not an exception?..

I think for now I'll just go with your suggestion to make a minimal breaking change and I'll throw some exception from the `RFU` to align with the `sftpSession` when it throws an exception instead of `false` return.

garyrussell:
>a minimal breaking change 

It's not really a breaking change, we will just detect the problem earlier - some downstream operation would have failed if the mkdir was not successful.

artembilan:
OK. I just mean that the `FtpSession` contract is not the same as an `SftpSession` one. So, it is not clear when should return `false` for SFTP case, or throw an exception for FTP to align them.
Right now we match both contracts in that RFU, but still have a difference in `mkdir()` behavior. That's what I don't want to break at the moment.

garyrussell:
>a minimal breaking change 

It's not really a breaking change, we will just detect the problem earlier - some downstream operation would have failed if the mkdir was not successful.

garyrussell:
>FtpSession contract is not the same as an SftpSession

I don't think that matters - FTP can throw an exception too.

garyrussell:
```
RemoteFileOutboundGatewayTests > testMoveWithMkDirs FAILED
    org.springframework.messaging.MessagingException at RemoteFileOutboundGatewayTests.java:304
        Caused by: java.io.IOException at RemoteFileOutboundGatewayTests.java:304
```

garyrussell:
Cherry-picked to 5.3.x, 5.2.x - 4.3.x needs a back-port.

artembilan:
Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/90196767dbe7a141fbd6d61fdc1f78528859b274 after fixing conflicts.

PR #3382, opened 2020-09-11
GH-3366: Document HTTP request mapping limitation

### LINKED ISSUE #3366: inbound-gateway: calling a overloaded resource mapped with same path but different method throws exception
5.3.2.RELEASE

**Describe the bug**
on one side we have a controller which is listening  to a generic GET endpoint /api1/whatevertypes

```java

@RestController
@RequestMapping(
    value = "/api1",
    produces = {"application/json"})
@AllArgsConstructor
@CrossOrigin(origins = "*")
@Slf4j
public class GenericApi {

  @GetMapping(value = {"/{whatevertype}s")
  ResponseEntity getList(
      @PathVariable() final String whatevertype,
      @RequestHeader HttpHeaders httpHeaders) {
  

      return ResponseEntity.status(HttpStatus.BAD_REQUEST)
          .contentType(MediaType.TEXT_PLAIN)
          .body("1234");
    
  }
}

```

on the other side we have Spring Integration http inbound-gateway which should listen to POST endpoint with the same path as above /api1/whatevertypes

```java
  
 <int-http:inbound-gateway request-channel="request-channel"
                              reply-channel="inbound-reply-channel"
                              supported-methods="POST"
                              path="/api1/{whatevertype}s"
                              request-payload-type="java.lang.String"

    >
        <int-http:request-mapping consumes="application/json" produces="application/json"/>
        <int-http:header name="whatevertype" expression="#pathVariables.whatevertype"/>
    </int-http:inbound-gateway>

```

**when we call the POST endpoint with the path like "%/api1/mytypes" it throws the error: Request method 'POST' not supported
,an error <405 METHOD_NOT_ALLOWED> is thrown.**

On debug we can see that both mappings exists but when spring is looping through it just tries to match the first map and throws an error if it doesn't find a match.

If we configure the order/prio (RequestMappingHandlerMapping)  where spring i trying to resolve the path it is failing the other way round (POST is working GET doesn't).  

**To Reproduce**
Set up a GET restcontroller with a generic path and in parallel set up an Spring Integration http inbound gateway with a POST method and same path as the GET restcontroller.

Call the POST endpoint leads to <405 METHOD_NOT_ALLOWED>

**Expected behavior**
spring must loop over all possible mappings for matching before throwing an error.

Calling the POST and the GET resource with the configuration above works.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3366

The same path cannot be mapped both Spring Integration and MVC ways

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/inbound/IntegrationRequestMappingHandlerMapping.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/inbound/IntegrationRequestMappingHandlerMapping.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


garyrussell:
Pushed some polishing for your review.

artembilan:
> Pushed some polishing for your review.

👍 . Thank you Gary!

PR #3385, opened 2020-09-14
GH-3336: Change MongoDb Store sequence to long

### LINKED ISSUE #3336: AbstractConfigurableMongoDbMessageStore's getNextId might overflow to long leading to "java.lang.Long cannot be cast to java.lang.Integer"

It happens from version 4.3.12.RELEASE up to current. But it is possible that it happens with prior/future versions as well.

**Describe the bug**

Because getNextId performs an "increment" (by 1) on the "sequence" field of the special "messagesSequence" document, after Integer.MAX_VALUE calls to getNextId, next value will overflow from Integer.MAX_VALUE to a Long and when trying to retrieve it, will get fail with:

Caused by: java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer

```
	/**
	 * Perform MongoDB {@code INC} operation for the document, which contains the {@link MessageDocument}
	 * {@code sequence}, and return the new incremented value for the new {@link MessageDocument}.
	 * The {@link #SEQUENCE_NAME} document is created on demand.
	 * @return the next sequence value.
	 */
	protected int getNextId() {
		Query query = Query.query(Criteria.where("_id").is(SEQUENCE_NAME));
		query.fields().include(MessageDocumentFields.SEQUENCE);
		return (Integer) this.mongoTemplate.findAndModify(query,
				new Update().inc(MessageDocumentFields.SEQUENCE, 1),
				FindAndModifyOptions.options().returnNew(true).upsert(true),
				Map.class, this.collectionName)
				.get(MessageDocumentFields.SEQUENCE);
	}
```

**To Reproduce**

Add Integer.MAX_VALUE + 1 of messages in aggregation store.

```
Caused by: java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer
	at org.springframework.integration.mongodb.store.AbstractConfigurableMongoDbMessageStore.getNextId(AbstractConfigurableMongoDbMessageStore.java:180)
	at org.springframework.integration.mongodb.store.ConfigurableMongoDbMessageStore.addMessagesToGroup(ConfigurableMongoDbMessageStore.java:196)
	at org.springframework.integration.mongodb.store.ConfigurableMongoDbMessageStore.addMessageToGroup(ConfigurableMongoDbMessageStore.java:167)
	at org.springframework.integration.aggregator.AbstractCorrelatingMessageHandler.store(AbstractCorrelatingMessageHandler.java:621)
	at org.springframework.integration.aggregator.AbstractCorrelatingMessageHandler.handleMessageInternal(AbstractCorrelatingMessageHandler.java:413)
	at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:127)
```

**Expected behavior**

Well, getNextId returns an int which populates the MessageDocument's sequence which is an int, so most likely, the MessageDocument's sequence's type needs to be changed from int to long and then getNextId, needs to return a long.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3336

Turns out there are some scenarios where too many messages
are transferred through the message store, so `int` for
sequence is not enough as a type

* Change sequence to `long` to widen a sequence lifespan

### DIFF
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
@@ -150,13 +150,15 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 
 		indexOperations.ensureIndex(new Index(MessageDocumentFields.MESSAGE_ID, Sort.Direction.ASC));
 
-		indexOperations.ensureIndex(new Index(MessageDocumentFields.GROUP_ID, Sort.Direction.ASC)
-				.on(MessageDocumentFields.MESSAGE_ID, Sort.Direction.ASC)
-				.unique());
+		indexOperations.ensureIndex(
+				new Index(MessageDocumentFields.GROUP_ID, Sort.Direction.ASC)
+						.on(MessageDocumentFields.MESSAGE_ID, Sort.Direction.ASC)
+						.unique());
 
-		indexOperations.ensureIndex(new Index(MessageDocumentFields.GROUP_ID, Sort.Direction.ASC)
-				.on(MessageDocumentFields.LAST_MODIFIED_TIME, Sort.Direction.DESC)
-				.on(MessageDocumentFields.SEQUENCE, Sort.Direction.DESC));
+		indexOperations.ensureIndex(
+				new Index(MessageDocumentFields.GROUP_ID, Sort.Direction.ASC)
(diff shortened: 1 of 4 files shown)


garyrussell:
What's new, migration? Breaking change? (i.e. can documents created with int be read as long?). If not, we probably need to provide a conversion utility.

artembilan:
Sure! I'll prepare Migration Guide note.
Probably the best way to go is to remove an existing `messagesSequence` document and let the framework to recreate it with the `long` type already.

I don't think there is the problem with casting existing `int` values for allocated sequences in message documents to `long`...

artembilan:
Pushed some fix to keep compatibility with the previous MongoDb collection state.

The Migration Guide note still has to be done.

PR #3389, opened 2020-09-23
Fix `IntegrationReactiveUtils`

artembilan:
The `Mono.doOnSuccess()` is always called for completed `MonoSink`
even if the value is `null`

* Fix `IntegrationReactiveUtils.messageSourceToFlux()` to check a
message for `null` before calling `AckUtils.autoAck()`
* Add an `logger.error()` message when `doOnError()` is invoked for that `Mono`

**Cherry-pick to 5.3.x**


garyrussell:
And cherry-picked as e649480350

PR #3391, opened 2020-09-25
GH-3194: Add generics to Amqp and Jms channel impls

### LINKED ISSUE #3194: Refactor JmsPollableMessageChannelSpec for target channel impl generic type
See TODO: https://github.com/spring-projects/spring-integration/blob/master/spring-integration-jms/src/test/java/org/springframework/integration/jms/dsl/JmsTests.java#L374

Do the same with `AmqpPollableMessageChannelSpec`.

oxcafedead:
For now we keep the hierarchy, but try to generalize channel specs a bit, so that ones who use `publishSubscribeChannel` can count on `BroadcastCapableChannel` as interface to utilize. 
Should close #3194

### DIFF
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/dsl/JmsPollableMessageChannelSpec.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/dsl/JmsPollableMessageChannelSpec.java
@@ -35,11 +35,12 @@ import org.springframework.lang.Nullable;
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Artem Vozhdayenko
  *
  * @since 5.0
  */
-public class JmsPollableMessageChannelSpec<S extends JmsPollableMessageChannelSpec<S>>
-		extends MessageChannelSpec<S, AbstractJmsChannel> {
+public class JmsPollableMessageChannelSpec<S extends JmsPollableMessageChannelSpec<S, T>, T extends AbstractJmsChannel>
+		extends MessageChannelSpec<S, T> {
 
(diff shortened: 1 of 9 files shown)


oxcafedead:
Strage, I've set up IDEA according to this manual https://github.com/spring-projects/spring-framework/wiki/IntelliJ-IDEA-Editor-Settings and tests format causes checkstyle to fail.

artembilan:
Yeah... Because we prefer to have `import static all other methods` in the beginning of the import section.

We probably would need to revise our Checkstyle preferences, too. But still: in `6.0` 😄 

Thank you for your effort, @oxcafedead !
Will review your PRs today

spring-integration-jms/src/main/java/org/springframework/integration/jms/dsl/JmsPollableMessageChannelSpec.java
```diff
@@ -35,11 +35,12 @@
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Artem Vozhdayenko
  *
  * @since 5.0
  */
-public class JmsPollableMessageChannelSpec<S extends JmsPollableMessageChannelSpec<S>>
-		extends MessageChannelSpec<S, AbstractJmsChannel> {
+public class JmsPollableMessageChannelSpec<S extends JmsPollableMessageChannelSpec<S, T>, T extends AbstractJmsChannel>
```
artembilan, line 42: Yeah... I see now your pain: the `SubscribableJmsChannel` is not a `PollableJmsChannel`, so their specs really must not extend each other.

Let's really revise this flaw in the next `6.0`!

Merging your fix so far...

PR #3397, opened 2020-10-02
GH-3395: Fix XML expression default for ARFOGateway

### LINKED ISSUE #3395: SFTP Outbound Gateway: UnsatisfiedDependencyException when no expression is specified

5.3.2.RELEASE

**Describe the bug**

When you define a SFTP outbound gateway using the XML namespace, if you don't specify the `expression` attribute, instantiation of the gateway fails with an obscure message.

**To Reproduce**

Consider this:

```xml
<int-sftp:outbound-gateway
  id="myGateway"
  request-channel="inputChannel"
  reply-channel="outputChannel"
  session-factory="mySessionFactory"
  remote-directory="${sftp.base-path}"
  command="put" />

<bean id="mySessionFactory"
  class="org.springframework.integration.file.remote.session.CachingSessionFactory">
    <constructor-arg>
      <bean class="org.springframework.integration.sftp.session.DefaultSftpSessionFactory">
        <property name="host" value="${sftp.host}" />
        <property name="port" value="${sftp.port}" />
        <property name="user" value="${sftp.username}" />
        <property name="password" value="${sftp.password}" />
        <property name="knownHostsResource" value="classpath:/config/integration/sftp_host" />
        <property name="timeout" value="${sftp.timeout}" />
      </bean>
    </constructor-arg>
    <constructor-arg value="3" />
    <property name="sessionWaitTimeout" value="120000" />
    <property name="testSession" value="true" />
</bean>
```

**Expected behavior**

This is expected to work. As documentation for `put` command says, "the payload of the message can be a `java.io.File`, a `byte[]`, or a `String`". So, it's natural for me that the "default expression" is the message payload, unless I want to submit something else. This is quite standard in Spring Integration.

(quick hint: what about `java.nio.file.Path` support?)

**Actual behavior**

This fails on application context refresh with:

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.integration.sftp.gateway.SftpOutboundGateway#0' defined in class path resource [config/integration/applicationContext-integration-sftp.xml]: Unsatisfied dependency expressed through constructor parameter 0: Could not convert argument value of type [org.springframework.integration.sftp.session.SftpRemoteFileTemplate] to required type [org.springframework.integration.file.remote.session.SessionFactory]: Failed to convert value of type 'org.springframework.integration.sftp.session.SftpRemoteFileTemplate' to required type 'org.springframework.integration.file.remote.session.SessionFactory'; nested exception is java.lang.IllegalStateException: Cannot convert value of type 'org.springframework.integration.sftp.session.SftpRemoteFileTemplate' to required type 'org.springframework.integration.file.remote.session.SessionFactory': no matching editors or conversion strategy found
	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:765) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:227) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1356) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1203) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:556) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:516) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:324) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:226) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:322) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:897) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:879) ~[spring-context-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:551) ~[spring-context-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext.refresh(ServletWebServerApplicationContext.java:143) ~[spring-boot-2.3.3.RELEASE.jar:2.3.3.RELEASE]
```

**Workaround**

Add `expression="payload"` to the outbound gateway definition.

This was also submitted to Stack Overflow at: https://stackoverflow.com/questions/64160194/unsatisfieddependencyexception-when-setting-up-a-spring-integration-sftp-outboun

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3395

Outbound remote file gateway parser requires `expression` even though some
commands don't need or use it.

To avoid adding a factory bean, set a default value in the schema and interpret
that default as needed, depending on the command.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileOutboundGatewayParser.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileOutboundGatewayParser.java
@@ -58,13 +58,7 @@ public abstract class AbstractRemoteFileOutboundGatewayParser extends AbstractCo
 		}
 		else {
 			builder.addConstructorArgValue(element.getAttribute("command"));
-			if (element.hasAttribute(EXPRESSION_ATTRIBUTE)) {
-				String expression = element.getAttribute(EXPRESSION_ATTRIBUTE);
-				if ("++xsd.expression.default++".equals(expression)) {
-					expression = null;
-				}
-				builder.addConstructorArgValue(expression);
-			}
+			builder.addConstructorArgValue(element.getAttribute(EXPRESSION_ATTRIBUTE));
 		}
 		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "command-options", "options");
 		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "reply-timeout", "sendTimeout");
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileOutboundGatewayParser.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileOutboundGatewayParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,7 +59,11 @@ public abstract class AbstractRemoteFileOutboundGatewayParser extends AbstractCo
 		else {
 			builder.addConstructorArgValue(element.getAttribute("command"));
 			if (element.hasAttribute(EXPRESSION_ATTRIBUTE)) {
-				builder.addConstructorArgValue(element.getAttribute(EXPRESSION_ATTRIBUTE));
+				String expression = element.getAttribute(EXPRESSION_ATTRIBUTE);
+				if ("++xsd.expression.default++".equals(expression)) {
+					expression = null;
+				}
+				builder.addConstructorArgValue(expression);
 			}
 		}
 		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "command-options", "options");
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
@@ -196,18 +196,14 @@ public abstract class AbstractRemoteFileOutboundGateway<F> extends AbstractReply
 		this.remoteFileTemplate = remoteFileTemplate;
 		this.command = command;
 		String expression = expressionArg;
-		boolean noExpressionNeeded = Command.LS.equals(this.command)
+		boolean expressionNeeded = !(Command.LS.equals(this.command)
 				|| Command.NLST.equals(this.command)
 				|| Command.PUT.equals(this.command)
-				|| Command.MPUT.equals(this.command);
-		if ("++xsd.expression.default++".equals(expression)) {
-			expression = noExpressionNeeded ? null : "payload";
+				|| Command.MPUT.equals(this.command));
+		if (expression == null && expressionNeeded) {
+			expression = "payload";
 		}
 		if (expression == null) {
-			Assert.state(noExpressionNeeded,
-					"Only LS, and NLST commands can rely on the working directory.\n" +
-							"PUT and MPUT commands always use the payload.\n" +
-							"All other commands must be supplied with a filename expression");
 			this.fileNameProcessor = null;
 		}
 		else {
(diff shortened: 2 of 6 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
```diff
@@ -187,21 +187,27 @@ public AbstractRemoteFileOutboundGateway(RemoteFileTemplate<F> remoteFileTemplat
...
 		Assert.notNull(remoteFileTemplate, "'remoteFileTemplate' cannot be null");
 		this.remoteFileTemplate = remoteFileTemplate;
 		this.command = command;
+		String expression = expressionArg;
+		boolean noExpressionNeeded = Command.LS.equals(this.command)
+				|| Command.NLST.equals(this.command)
+				|| Command.PUT.equals(this.command)
+				|| Command.MPUT.equals(this.command);
+		if ("++xsd.expression.default++".equals(expression)) {
+			expression = noExpressionNeeded ? null : "payload";
```
artembilan, line 204: Why can't we set it to `null` from the parser and override it to `"payload"` if needed according command logic?

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/config/AbstractRemoteFileOutboundGatewayParser.java
```diff
@@ -59,7 +59,11 @@ protected BeanDefinitionBuilder parseHandler(Element element, ParserContext pars
 		else {
 			builder.addConstructorArgValue(element.getAttribute("command"));
 			if (element.hasAttribute(EXPRESSION_ATTRIBUTE)) {
-				builder.addConstructorArgValue(element.getAttribute(EXPRESSION_ATTRIBUTE));
+				String expression = element.getAttribute(EXPRESSION_ATTRIBUTE);
+				if ("++xsd.expression.default++".equals(expression)) {
+					expression = null;
+				}
+				builder.addConstructorArgValue(expression);
```
artembilan, line 66: Why just this not enough?
I mean we just call `builder.addConstructorArgValue(element.getAttribute(EXPRESSION_ATTRIBUTE));`.
Nothing more. 
What am I missing with `null` propagation and defering logic into the `AbstractRemoteFileOutboundGateway`.
Probably that point is that `expression` was not optional in the past, so we left that `if (element.hasAttribute(EXPRESSION_ATTRIBUTE)) {` logic in the parser.
But it turns out that we just need to always call `builder.addConstructorArgValue(expression);`, so we got the proper ctor resolution even if `expression` is `null`.

artembilan (changes requested):

artembilan:
... and cherry-picked to `5.3.x`.

PR #3398, opened 2020-10-02
GH-3396: Add Docs for Micrometer MeterFilters

### LINKED ISSUE #3396: Document how to disable meters with Micrometer configuration
**Expected Behavior**

Have a Micrometer enabled in the application, but be able to exclude some specific components from metrics gathering.
Looks like it is possible with the `registry.config().meterFilter()`, but should be explained in the [docs](https://docs.spring.io/spring-integration/docs/5.4.0-M3/reference/html/system-management.html#micrometer-integration) how to do that for Spring Integration components.

**Current Behavior**

The legacy metrics relied on the `@EnableIntegrationManagement(metersEnabled)` option. Now it is deprecated, but users still would like to disable some components from gathering metrics. It is not clear from docs at the moment how to do that.

**Context**

The user wants to disable metrics gathering for some specific `MessageChannel` bean. Not having enough knowledge for Micrometer and Spring Integration components naming, it becomes as a blocker how to disable metrics.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3396

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsTests.java
@@ -59,6 +59,7 @@ import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 import org.springframework.test.context.support.DependencyInjectionTestExecutionListener;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.config.MeterFilter;
 import io.micrometer.core.instrument.search.MeterNotFoundException;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 
@@ -91,6 +92,9 @@ public class MicrometerMetricsTests {
 
 	@Autowired
 	private QueueChannel queue;
+
+	@Autowired
+	private QueueChannel noMeters;
 
 	@Autowired
 	private PollableChannel badPoll;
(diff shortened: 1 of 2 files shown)


src/reference/asciidoc/metrics.adoc
```diff
@@ -118,3 +120,11 @@ and
...
 * `tag`: `name:<componentName>`
 * `description`: `The remaining capacity of the queue channel`
+
+===== Disabling Meters
+
+With the <<legacy-metrics>> (which have now been removed), you could specify which integration components would collect metrics.
+By default, all meters are registered when first used.
+Now, with Micrometer, you can add `MeterFilter` s to the `MeterRegistry` to prevent some or all from being registered.
+You can filter out (deny) meters by any of the properties provided, `name`, `tag`, etc.
+See https://micrometer.io/docs/concepts#_meter_filters[Meter Filters] in the Micrometer documentation for more information.
```
artembilan, line 130: Any chances to have some sample how to disable metrics for example for `myChannel` bean?

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsTests.java
```diff
@@ -264,7 +275,11 @@ public void testMicrometerMetrics() {
 
 		@Bean
 		public static MeterRegistry meterRegistry() {
-			return new SimpleMeterRegistry();
+			SimpleMeterRegistry registry = new SimpleMeterRegistry();
+			registry.config().meterFilter(MeterFilter.deny(id ->
+					"channel".equals(id.getTag("type")) &&
+					"noMeters".equals(id.getTag("name"))));
```
artembilan, line 281: Probably writing this test you were going to add something to docs.
Aren't you?

artembilan (changes requested):

PR #3400, opened 2020-10-07
ReactiveRedisMessageProducer autoacknowledgment when set to false does not work. 

rohanmukesh:
Reactive Message Producer is auto acknowledging messages even though we set messageproducet.setAutoAck(false). 
R2dbc outbound adapter component type had reactive-outbound-channel-adapter which has been renamed to outbound-channel-adapter and tests have been added for both inbound and outbound component types.

### DIFF
--- a/spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
+++ b/spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
@@ -17,14 +17,12 @@
 package org.springframework.integration.r2dbc.inbound;
 
 
-import static org.assertj.core.api.Assertions.*;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 
-import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
--- a/spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
+++ b/spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
@@ -17,7 +17,6 @@
 package org.springframework.integration.r2dbc.inbound;
 
 
-import static org.assertj.core.api.Assertions.*;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.time.Duration;
--- a/spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
+++ b/spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
@@ -23,6 +23,7 @@ import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 
+import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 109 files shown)


spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/inbound/R2dbcMessageSourceTests.java
```diff
@@ -89,6 +89,12 @@ public void setup() {
 				.verifyComplete());
 	}
 
+	@Test
+	public void validateComponentType() {
+		assertThat(this.defaultR2dbcMessageSource.getComponentType()).isEqualTo("r2dbc:inbound-channel" +
+				"-adapter");
```
artembilan, line 95: Isn't that would be better to wrap the line on the `.isEqualTo()` instead of splitting string?

spring-integration-r2dbc/src/test/java/org/springframework/integration/r2dbc/outbound/R2dbcMessageHandlerTests.java
```diff
@@ -115,6 +116,11 @@ public void validateMessageHandlingWithInsertQueryCollection() {
 
 	}
 
+			@Test
```
artembilan, line 119: Wrong indents for this new piece of code.

spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java
```diff
@@ -28,10 +31,11 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.data.redis.connection.stream.ReadOffset;
+import org.springframework.data.redis.connection.stream.*;
```
artembilan, line 34: No asterisk imports at all, please.

spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java
```diff
@@ -162,6 +170,34 @@ public void testReadingMessageAsConsumerInConsumerGroup() {
...
+					 .as(StepVerifier::create)
+					 .assertNext(message -> assertThat(message).isEqualTo("OK"))
+					 .thenCancel()
+					 .verify(Duration.ofSeconds(10));
+
+		this.redisStreamMessageProducer.setCreateConsumerGroup(false);
+		this.redisStreamMessageProducer.setAutoAck(false);
+		this.redisStreamMessageProducer.setConsumerName(CONSUMER);
+		this.redisStreamMessageProducer.afterPropertiesSet();
+		this.redisStreamMessageProducer.start();
```
artembilan, line 190: Any chances to see consumed messages and verify their headers for that `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK`?


And probably after `pending` verification call the ack and see if there is no pending afterward left.

artembilan (changes requested):
Please, fix Checkstyle issue:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-integration/spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducer.java:226: Line has leading space characters; indentation should be performed with tabs only. [RegexpSinglelineJava]
```

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/6d6f53fa8792b398e9b528b14591790f30d1201f after fixing Redis Streams tests and resolving conflicts.

The problem with the test was not clear: I've just combined an existing test with your code to be sure that ack callback works as expected. Plus that we really have pending in between. 
Also I decided to remove all the consumer groups between tests. Seems helped.

Thank you for contribution!
Was great to have such a feedback before our GA 😄

PR #3403, opened 2020-10-13
Add Error Handling to ReactiveRedisStreamMessageProducer

rohanmukesh:
* Use 'onErrorContinue' to continue receiving messages from Stream

### DIFF
--- a/spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java
+++ b/spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java
@@ -22,6 +22,8 @@ import static org.awaitility.Awaitility.await;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java
```diff
@@ -277,6 +366,20 @@ ReactiveRedisStreamMessageProducer reactiveRedisStreamProducer() {
...
 		}
 
+		@Bean
+		public IntegrationFlow errorHandlingFlow(ReactiveRedisTemplate<String, ?> reactiveStreamOperations) {
+			return IntegrationFlows.from("errorChannel")
+									.handle(message -> {
+										reactiveStreamOperations.opsForStream()
+																.acknowledge("testGroup",
+																((MapRecord) ((ErrorMessage) message)
+																.getOriginalMessage().getPayload()))
```
artembilan, line 376: We probably need to think about adding all of those headers for the record into a failed message we send with an error, so end-user would still be able to call simple `acknowledge()` from the `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` header and also get all the useful information from other headers.

I also wouldn't advertise a `MapRecord` over here as an entity to deal with, but more its super, plain `Record<S, V>` interface. Also, see those raw types warnings: better to have it as `Record<?, ?>`.

OK. You ack here in the error handling, but I don't see a verification of this fact in the test-case.
Plus it is not clear how there could be a pending in the stream since you ack it here.
Isn't there going to be some race condition? Or our ack assumption is wrong?

artembilan (changes requested):
See Travis report:
```
 Task :spring-integration-redis:compileTestJava

/home/travis/build/spring-projects/spring-integration/spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java:332: warning: [rawtypes] found raw type: Jackson2JsonRedisSerializer

			Jackson2JsonRedisSerializer jacksonSerializer = new Jackson2JsonRedisSerializer(Person.class);

			^

  missing type arguments for generic class Jackson2JsonRedisSerializer<T>

  where T is a type-variable:

    T extends Object declared in class Jackson2JsonRedisSerializer

/home/travis/build/spring-projects/spring-integration/spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java:332: warning: [rawtypes] found raw type: Jackson2JsonRedisSerializer

			Jackson2JsonRedisSerializer jacksonSerializer = new Jackson2JsonRedisSerializer(Person.class);

			                                                    ^

  missing type arguments for generic class Jackson2JsonRedisSerializer<T>

  where T is a type-variable:

    T extends Object declared in class Jackson2JsonRedisSerializer

/home/travis/build/spring-projects/spring-integration/spring-integration-redis/src/test/java/org/springframework/integration/redis/inbound/ReactiveRedisStreamMessageProducerTests.java:332: warning: [unchecked] unchecked call to Jackson2JsonRedisSerializer(Class<T>) as a member of the raw type Jackson2JsonRedisSerializer

			Jackson2JsonRedisSerializer jacksonSerializer = new Jackson2JsonRedisSerializer(Person.class);
```

Please, be sure to fix all the warnings.

artembilan:
Pulling locally for fixes and clean up we agreed in the private discussion. 
It is really good to have this fix in the release!

I'll share here the final commit when it is ready.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/b1a383060d0940890816f9a50155e6bce972e53a.

Pay attention, please, how I have extracted a `buildMessageFromRecord()` method. 
Also see the test modification: there is no reason to build the whole flow for error handling since we just need to get an `ErrorMessage` and deal with its content, which is very useful for the test-case flow.
I also added another message to Redis stream to be sure that we continue to process after conversion exception.

Thank you for your contribution, @rohanmukesh !
Looking forward for more (as usual)!

PR #3404, opened 2020-10-15
Fix MeterRegistry eager load

artembilan:
If there is a `MeterRegistry` bean (any) in the application context,
we add a `MicrometerMetricsCaptor` bean which populates meters further
from the components.
In some case we may load a `MicrometerMetricsCaptor` bean too early
so, not all the stuff around `MeterRegistry` maybe ready.
See Spring Boot and its `MetricsAutoConfiguration`

* Fix the `MicrometerMetricsCaptorRegistrar` the way to rely on the
`ObjectProvider<MeterRegistry>` instead
* Add package protected ctor to the `MicrometerMetricsCaptor` to
provide a target `MeterRegistry` on demand

All of that will ensure that we use an already post-processed `MeterRegistry`
including Spring Boot auto-configuration

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsCaptor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsCaptor.java
@@ -20,6 +20,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.ToDoubleFunction;
 
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.support.GenericApplicationContext;
 import org.springframework.integration.support.management.metrics.CounterFacade;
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsCaptor.java
```diff
@@ -49,31 +50,44 @@
...
 		this.meterRegistry = meterRegistry;
 	}
 
+	MicrometerMetricsCaptor(ObjectProvider<MeterRegistry> meterRegistryProvider) {
+		this.meterRegistryProvider = meterRegistryProvider;
+	}
+
+	public MeterRegistry getMeterRegistry() {
+		if (this.meterRegistry == null) {
+			this.meterRegistry = this.meterRegistryProvider.getIfUnique();
```
garyrussell, line 68: Shouldn't we get the first if there are more than one? (`provider.stream().findFirst()`) This would then be consistent with the existing code.

garyrussell (commented):
Just one question/comment.

spring-integration-core/src/main/java/org/springframework/integration/support/management/micrometer/MicrometerMetricsCaptor.java
artembilan: Well, it is better to use a `getIfUnique()` since this one returns for a `@Primary` one - exactly what we want to have in case of Spring Boot.
Apparently the current solution is still bug, since we inject really only one `MeterRegistry` (Prometheus), but not the composite built by Boot including all the rest available registries, e.g. WaveFront one.

garyrussell: OK; just waiting for travis, then.

garyrussell:
Cherry-picked to 5.3.x, 5.2.x after resolving test conflicts.

PR #3405, opened 2020-10-15
Kafka Outbound Endpoints - Add futuresChannel

garyrussell:
Currently, the only way to block is to set `sync=true` which waits for the future.
The problem with this is it only supports one-at-a-time publication.

Add an option to send the send futures to a channel, allowing the application to
send multiple records and then wait on the futures later.

Also fix long-running test.

**Needs docs before merging**

### DIFF
--- a/spring-integration-kafka/src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
+++ b/spring-integration-kafka/src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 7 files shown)


artembilan:
> The problem with this is it only supports one-at-a-time publication.

I didn't look into the code, but I'm confused. 
Isn't that typical behavior when we call blocked code from a single thread?
Or do you mean that we don't support concurrency over there?

Thanks

garyrussell:
This is to support `send,send,send,wait,wait,wait`, which will run much faster than `send,wait,send,wait,send,wait`.

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
```diff
@@ -354,6 +371,7 @@ public DefaultKafkaHeaderMapper mapper() {
 
 			return Kafka
 					.outboundChannelAdapter(producerFactory)
+					.futuresChannel("futuresChannel")
 					.sync(true)
```
artembilan, line 375: We probably don't need `sync` to prove that "send, send, wait, wait" concept...

artembilan (changes requested):
I don't feel the feature yet, but we can give it a try in the field.
Probably I'll understand more when see docs.

I wonder how do you cover this new functionality when the `KafkaProducerMessageHandler` is used in a gateway, request/reply mode... 
Don't we need to ignore it somehow? Or what is the purpose of such a channel for futures when we really care only about replies from remote service?..

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
garyrussell: It's not really important here; I just wanted to test that the futures were indeed sent to the channel; you are correct in that, in practice, you would not do both.

garyrussell:
I thought about only supporting this with the adapter and not the gateway but, even with a gateway it might useful. For example, after a reply timeout, you can determine if the timeout is because the send failed Vs. the server didn't respond in time.

artembilan:
OK. Thank you! 
You can go ahead with docs

PR #3408, opened 2020-10-21
Gateway: Propagate Error to the errorChannel

artembilan:
See SO for more info:
https://stackoverflow.com/questions/64456946/handle-exceptions-errors-other-than-messagingexception-ie-other-error-excepti

In the versions before `5.2.x` the `Error` was wrapped to the `MessagingException`
in the `AbstractRequestHandlerAdvice` and this one was handled properly
on the gateway level with its `errorChannel` configured

* Fix `MessagingGatewaySupport` to catch all the `Throwable` and try to send them as `ErrorMessage`
to the `errorChannel`.
Otherwise unwrap returned `MessagingException` and re-throw an `Error` as is to keep
the current behavior for non-handled exceptions

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
@@ -528,7 +528,7 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 				sample.stop(sendTimer());
 			}
 		}
-		catch (Throwable ex) {
+		catch (Throwable ex) { // NOSONAR (catch throwable)
 			if (logger.isDebugEnabled()) {
 				logger.debug("failure occurred in gateway sendAndReceive: " + ex.getMessage());
 			}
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -528,7 +528,7 @@ private Object doSendAndReceive(Object object, boolean shouldConvert) {
 				sample.stop(sendTimer());
 			}
 		}
-		catch (Exception ex) {
+		catch (Throwable ex) {
```
garyrussell, line 531: Needs `// NOSONAR` (catch throwable).

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -543,7 +543,14 @@ private Object doSendAndReceive(Object object, boolean shouldConvert) {
...
 							? ((ErrorMessage) reply).getPayload()
 							: (Throwable) reply;
-			return handleSendAndReceiveError(object, requestMessage, error, shouldConvert);
+			try {
+				return handleSendAndReceiveError(object, requestMessage, error, shouldConvert);
+			}
+			catch (MessagingException me) {
+				if (me.getCause() instanceof Error) { // NOSONAR
+					throw (Error) me.getCause();
+				}
```
garyrussell, line 552: Needs `throw me;` to retain existing behavior of throwing error flow exceptions to caller.

It's not clear why this try/catch is here at all; this is an `Error` on the error flow, not the original flow.

garyrussell (changes requested):
Wouldn't it be safer to just fix the retry advice?

garyrussell (approved):

artembilan:
>  garyrussell approved these changes

@garyrussell , do you want me to merge this PR for you?
BTW, do you agree about back-porting it?

Thanks

garyrussell:
Sure. I was waiting for travis and forgot about it.

Yes about back port.

artembilan:
... and cherry-picked to `5.3.x` & `5.2.x`

PR #3412, opened 2020-10-22
Fix `@InChAdapter` for several supplier beans

artembilan:
The `InboundChannelAdapterAnnotationPostProcessor` doesn't use a bean method name
when it parses a `Supplier` bean and only uses a configuration class name + `get`
for method part

* Fix `InboundChannelAdapterAnnotationPostProcessor` to also include a bean method
name into the final bean name for the `MethodInvokingMessageSource` based on
the `Supplier` bean
* Modify `ReactiveInboundChannelAdapterTests` to add one more `Supplier` with the
`@InboundChannelAdapter` to ensure that configuration is still valid after the fix


garyrussell:
Cherry-picked to 5.2.x 0ab395834f and 5.1.x b2d7867ce2

PR #3413, opened 2020-10-22
GH-3410: Add UDP SocketCustomizer

### LINKED ISSUE #3410: Support setting the trafficClass on sockets for udp via DSL.
Setting the trafficClass is now not supported via DSL on UDP sockets.

I like that it is possbile to set the TrafficClass (similar as setting the timeout)

For example: 
`Upd.outboundAdapter(host, port).soTrafficClass(0x10).`

Extra behavior to setting the trafficClass 

I have created a workaround to set the traffic class on a udp socket.
This is not a neat way and may break on future updates for spring integration.

I need it because in our local network this is used to give priority to some udp messages.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3410

**cherry-pick to 5.3.x**

### DIFF
--- a/src/reference/asciidoc/ip.adoc
+++ b/src/reference/asciidoc/ip.adoc
@@ -72,6 +72,7 @@ The following example configures a UDP outbound channel adapter:
     host="somehost"
     port="11111"
     multicast="false"
+    socket-customizer="udpCustomizer"
     channel="exampleChannel"/>
 ----
 ====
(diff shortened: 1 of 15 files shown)


src/reference/asciidoc/ip.adoc
```diff
@@ -201,6 +205,8 @@ By default, reverse DNS lookups are done on inbound packets to convert IP addres
 In environments where DNS is not configured, this can cause delays.
 You can override this default behavior by setting the `lookup-host` attribute to `false`.
 
+Starting with version 5.3.3, you can add a `SocketCustomizer` bean to modify the `DatagramSocket` after it is created. It is called for the receiving socket and any sockets created for sending acks.
```
artembilan, line 208: Every sentence in the docs from a new line, please: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

Probably we need to add a `What's New` entry as well.

I'll take care about it on cherry-pick.
Thanks

artembilan (changes requested):

artembilan:
... and cherry-picked to `5.3.x` as https://github.com/spring-projects/spring-integration/commit/8b70c201a6426e36db50157794d6e30ba19b4249 after fixing conflicts.

PR #3415, opened 2020-10-23
Add Java DSL for R2DBC module

artembilan:
* Add JavaDocs for R2DBC component setters
* Introduce a `R2dbcMessageSource.SelectCreator` to narrow
a `StatementMapper` API access from SpEL
* Add missed `package-info.java` into R2DBC packages
* Add `hamcrest-core` as test dependency into SI-R2DBC
since Awaitility doesn't work without it

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -647,6 +647,7 @@ project('spring-integration-r2dbc') {
 		}
 		api 'org.springframework:spring-r2dbc'
 		testImplementation "io.r2dbc:r2dbc-h2:$r2dbch2Version"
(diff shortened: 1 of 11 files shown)


artembilan:
@garyrussell ,

any chances to merge this?
Looks like that test failure is transient and that `sink.sink.done` `volatile` state is somehow not actual after calling the `this.sink.emitComplete(Sinks.EmitFailureHandler.FAIL_FAST);`...

garyrussell:
One question only - shouldn't we have `@NonNullApi` in the package infos?

PR #3419, opened 2020-10-30
GH-3418: Fix Poller Undeclared Checked Exceptions

### LINKED ISSUE #3418: ClassCastException in AbstractPollingEndpoint
Spring Integration 5.3.2.RELEASE

I am seeing a `ClassCastException` during error handling in `AbstractPollingEndpoint`.  Using jdk15 so there is more detail on the `ClassCastException`.

The thing is, looking at this line
https://github.com/spring-projects/spring-integration/blob/fb35a736d6fc159f9bd34794ff9fdb97b05462d9/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java#L417

For the life of me, I can't understand how a non `RuntimeException` can make it here.  In my case I am getting `java.nio.file.NoSuchFileException`.  Understand this is not an ideal bug report, but as a pure Java dev, I still cant figure out how a non-runtime exception can make it to line 417.  The `receiveMessage()` method above has no throws declared so the exception thrown must be a `RuntimeException`, but `java.nio.file.NoSuchFileException` is not.  The only thing I can think of would be some sort of lombok `@SneakyThrows` somewhere but that would still create a `UndeclaredThrowableException` around the inner exception.

I am not trying to stop the exception (I assume a user had a file locked during polling or something similar), I just want to get a normal throw without a `ClassCastException`.

This error only happens about once per day so it is difficult to reproduce.  The app has about 10 pollers but they are all similar.  Unfortunately I don't know which one is failing since there is no other information in the log or the stack trace regarding bean id. 

The spring configuraiton xml is using this adapter `<int-file:inbound-channel-adapter>`

```
2020-10-29 22:19:09,833 [scheduling-1] ERROR    o.s.i.handler.LoggingHandler - org.springframework.messaging.MessagingException: nested exception is java.lang.ClassCastException: class java.nio.file.NoSuchFileException cannot be cast to class java.lang.RuntimeException (java.nio.file.NoSuchFileException and java.lang.RuntimeException are in module java.base of loader 'bootstrap')
	at org.springframework.integration.endpoint.AbstractPollingEndpoint.pollForMessage(AbstractPollingEndpoint.java:390)
	at org.springframework.integration.endpoint.AbstractPollingEndpoint.lambda$null$3(AbstractPollingEndpoint.java:323)
	at org.springframework.integration.util.ErrorHandlingTaskExecutor.lambda$execute$0(ErrorHandlingTaskExecutor.java:57)
	at org.springframework.core.task.SyncTaskExecutor.execute(SyncTaskExecutor.java:50)
	at org.springframework.integration.util.ErrorHandlingTaskExecutor.execute(ErrorHandlingTaskExecutor.java:55)
	at org.springframework.integration.endpoint.AbstractPollingEndpoint.lambda$createPoller$4(AbstractPollingEndpoint.java:320)
	at org.springframework.scheduling.support.DelegatingErrorHandlingRunnable.run(DelegatingErrorHandlingRunnable.java:54)
	at org.springframework.scheduling.concurrent.ReschedulingRunnable.run(ReschedulingRunnable.java:93)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:832)
Caused by: java.lang.ClassCastException: class java.nio.file.NoSuchFileException cannot be cast to class java.lang.RuntimeException (java.nio.file.NoSuchFileException and java.lang.RuntimeException are in module java.base of loader 'bootstrap')
	at org.springframework.integration.endpoint.AbstractPollingEndpoint.doPoll(AbstractPollingEndpoint.java:417)
	at org.springframework.integration.endpoint.AbstractPollingEndpoint.pollForMessage(AbstractPollingEndpoint.java:376)
	... 13 more
```

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3418

`Class.newInstance()` can propagate checked exceptions that are not declared.

**cherry-pick/back-port to 5.3.x, 5.2.x, 4.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -58,6 +58,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ErrorHandler;
+import org.springframework.util.ReflectionUtils;
 
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -402,7 +403,7 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 
 	private Message<?> doPoll() {
 		IntegrationResourceHolder holder = bindResourceHolderIfNecessary(getResourceKey(), getResourceToBind());
-		Message<?> message;
+		Message<?> message = null;
 		try {
 			message = receiveMessage();
 		}
@@ -414,7 +415,7 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 				return null;
 			}
 			else {
-				throw (RuntimeException) e;
+				ReflectionUtils.rethrowRuntimeException(e);
 			}
 		}
 
--- a/spring-integration-core/src/test/java/org/springframework/integration/endpoint/SourcePollingChannelAdapterTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/endpoint/SourcePollingChannelAdapterTests.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


artembilan:
... and cherry-picked to `5.3.x` & `5.2.x`.

Back-ported to `4.3.x` as https://github.com/spring-projects/spring-integration/commit/b1932942cfc4b88389600d102c03d8b1c78a8c81 after removing new test which cannot be back-ported because of JUnit 5 and Java DSL usage in the code.

PR #3422, opened 2020-11-05
Fix subscribers race condition in FluxChannel

artembilan:
It is possible that
`this.sink.currentSubscriberCount() > 0)` is `true` in the `FluxMessageChannel`
before actually a `subscriber` is added to the source `Publisher`,
so we may end up with a condition when we can already emit, but
there is nothing to accept

* Defer `this.subscribedSignal.emitNext(true)` in the `FluxMessageChannel`
into the `doOnSubscribe()` of the `Flux` from `sink`
* Fix `Flux.destroy()` to emit `false` for the `subscribedSignal` before
calling `emitComplete()`.
Turns out that `delaySubscription()` can react for the completion signal
where the last value was `true` in the `Sinks.many().replay().limit(1)`
* Fix `MessageProducerSupport` to extract an `active` flag and set it before
`isRunning` - the `Flux` subscription relies on the `takeWhile()`
where in case of `autoStartup = false` we will never start consume because
it is set to `true` already after `doStart()`
* Refactor all the `MessageProducerSupport` implementation with similar
`active` state to use already one from the super class

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
@@ -94,11 +94,12 @@ public class FluxMessageChannel extends AbstractMessageChannel
 	@Override
 	public void subscribe(Subscriber<? super Message<?>> subscriber) {
 		this.sink.asFlux()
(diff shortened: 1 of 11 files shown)


artembilan:
After some discussion with @dturanski we decided to divide this PR into two: for `FluxMessageChannel` problem and `active` flag for `MessageProducer`.

Closing this one.

PR #3423, opened 2020-11-06
Rely on `MProducerSupport.active` for `Flux`

artembilan:
* Fix `MessageProducerSupport` to extract an `active` flag and set it before
`isRunning` - the `Flux` subscription relies on the `takeWhile()`
where in case of `autoStartup = false` we will never start consume because
it is set to `true` already after `doStart()`
* Refactor all the `MessageProducerSupport` implementation with similar
`active` state to use already one from the super class

**Cherry-pick to 5.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -161,10 +161,6 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 		return this.messagingTemplate;
 	}
 
-	protected void setActive(boolean active) {
-		this.active = active;
-	}
-
 	public boolean isActive() {
 		return this.active;
 	}
@@ -197,7 +193,7 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 	 */
 	@Override
 	protected void doStart() {
-		setActive(true);
+		this.active = true;
 	}
 
 	/**
@@ -207,7 +203,7 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 	 */
 	@Override
 	protected void doStop() {
-		setActive(false);
+		this.active = false;
 	}
 
 	protected void sendMessage(Message<?> messageArg) {
@@ -306,6 +302,12 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 		else {
 			return message;
 		}
+	}
+
+	@Override
+	public void destroy() {
+		this.active = false;
+		super.destroy();
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -66,8 +66,6 @@ public abstract class MessageProducerSupport extends AbstractEndpoint implements
 
 	private boolean shouldTrack = false;
 
-	private volatile boolean active;
-
 	protected MessageProducerSupport() {
 		this.setPhase(Integer.MAX_VALUE / 2);
 	}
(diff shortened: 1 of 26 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -196,6 +207,7 @@ protected void doStart() {
 	 */
 	@Override
 	protected void doStop() {
+		setActive(false);
 	}
```
garyrussell, line 207: Javadocs for these methods needs polishing. Subclasses MUST call `super` (which is a bit brittle I think).

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -222,7 +234,7 @@ protected void subscribeToPublisher(Publisher<? extends Message<?>> publisher) {
 						.map(this::trackMessageIfAny)
 						.doOnComplete(this::stop)
 						.doOnCancel(this::stop)
-						.takeWhile((message) -> isRunning());
+						.takeWhile((message) -> isActive());
```
garyrussell, line 233: It's not clear why this is different; both flags are set to true in `start()`.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
artembilan: Right, but `running` set after `doStart()` call. My idea (and I see in existing components) to have some flag which is `true` before we call `doStart()`. See that `MessageProducerSupport.start()` impl.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
artembilan: Well, I guess I can move this `active` directly into `start()` and `stop()` impls.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
garyrussell: Right, but why does it matter that it is set after exiting `doStart()`. The requirement is not clear at all.

artembilan: OK. See this `subscribeToPublisher()` details. There is that `subscribe()` word a couple times.
So, essentially we subscribe to the `Flux` in the `doStart()`, but since `isRunning()` returns `false` for us, we stop consuming the `Flux` immediately according that `takeWhile()` contract.

I won't mind any other ideas.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
artembilan: Well, I probably may hack and set `running` to true in that `subscribeToPublisher()`, so we will have less confusion with extra flag 😄

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
garyrussell: Ah; OK; got it now - but why not just set `active=true` in `start()` before calling `doStart()`? I think that would be cleaner.

artembilan: Right. That's an idea, too. Although we would need to move `active` flag from the `MessageProducerSupport` down to the `AbstractEndpoint`...
On the other hand I saw similar `active` in the gateway impls, so probably we are on the right track.

Thanks for confirmation! 
Stay tuned...

garyrussell (approved):

garyrussell:
Cherry-picked as 002382e647 to 5.3.x, after resolving conflicts.

PR #3426, opened 2020-11-10
GH-3425: Remove mngmt gauges from CtxClosedEvent

### LINKED ISSUE #3425: Spring Integration Metrics launch WARN message when the metrics are being collected in Graceful shutdown process.
I can reproduce the "issue" for Spring Integration version 5.2.8.RELEASE but I can see in version 5.4.0.RELEASE (https://github.com/spring-projects/spring-integration/blob/v5.4.0/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java) that the gauge definitions are equals. The conflictive code is at "registerComponentGauges" method.

	private void registerComponentGauges() {
		this.gauges.add(
				this.metricsCaptor.gaugeBuilder("spring.integration.channels", this,
						(c) -> this.applicationContext.getBeansOfType(MessageChannel.class).size())
						.description("The number of message channels")
						.build());

		this.gauges.add(
				this.metricsCaptor.gaugeBuilder("spring.integration.handlers", this,
						(c) -> this.applicationContext.getBeansOfType(MessageHandler.class).size())
						.description("The number of message handlers")
						.build());

		this.gauges.add(
				this.metricsCaptor.gaugeBuilder("spring.integration.sources", this,
						(c) -> this.applicationContext.getBeansOfType(MessageSource.class).size())
						.description("The number of message sources")
						.build());
	}

**Describe the bug**

The bug will be reproducible when any application based on SpringBoot with Micrometer starts to close Spring Application context when the app receive the graceful shutdown signal. In that moment, the micrometer registry try to send the gauge metrics related to "spring.integration.channels", "spring.integration.handlers" and "spring.integration.sources".

The lambda that is executed to collect the values for that metrics invoke to Spring context (`this.applicationContext.getBeansOfType(MessageHandler.class).size()`) producing that the context try to create the beans of the types `MessageHandler`, `MessageChannel `or `MessageSource`. This try to create the beans fails and throw a message 

> Singleton bean creation not allowed while singletons of this factory are in destruction

 because, obviously, the context is in destruction phase.

**To Reproduce**

To reproduce this bug is enough to create an application with micrometer (and a meter registry as a @Bean) and wait that the context start to destruction phase.

**Expected behavior**

In my opinion, the lambda functions to retrieve the metrics should be designed in other way, avoiding to check the spring context each time that we want to retrieve the metric value.

Another suggestion that maybe should be in another different issue is the way to disable the metrics. As I know, we don't have any way/property to disable the registration (and future shipping) of the metrics related to spring integration.
I found a workaround that is using an SpringBoot Actuator property, but that property is not efficient because affect as a meter filter, but In my opinion, we should act at the origin, without interacting with micrometer.

**Sample**

I don't have any standard project now to publish. I have my own code that belongs to my company and I can't publish right now. I think that is easily reproducible.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3425

It turns out that `IntegrationManagementConfigurer.destroy()` is still too
late to remove `gauges` and Micrometer tries to gather them on application
context close, when many beans are already destroyed

* Catch `ContextClosedEvent` in the `IntegrationManagementConfigurer`
and removed `gauges` from there
* Remove `destroy()` impl since it is out of use already:
the `IntegrationManagementConfigurer` is not supposed to be in the
target application directly, so it looks safe to remove the `DisposableBean`
altogether

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
@@ -23,11 +23,12 @@ import java.util.Set;
 
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanNameAware;
-import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextClosedEvent;
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.support.management.IntegrationManagement;
 import org.springframework.integration.support.management.IntegrationManagement.ManagementOverrides;
@@ -53,7 +54,7 @@ import org.springframework.util.Assert;
  */
 public class IntegrationManagementConfigurer
 		implements SmartInitializingSingleton, ApplicationContextAware, BeanNameAware, BeanPostProcessor,
-		DisposableBean {
+		ApplicationListener<ContextClosedEvent> {
 
 	/**
 	 * Bean name of the configurer.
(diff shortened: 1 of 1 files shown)


artembilan:
Note: tested with an application provided in the issue.

artembilan:
Back-ported to `5.3.x` as https://github.com/spring-projects/spring-integration/commit/72232ea02663f47e982a6c062ab82ed886f2cc85 after fixing conflicts.
Into `5.2.x` as https://github.com/spring-projects/spring-integration/commit/fc5fdbf3668fe4d25e2d24bcf8d8d541cf747cf9 after more conflict resolutions.

PR #3427, opened 2020-11-10
Fix mngmt dependency for MetricsCaptor

artembilan:
The `IntegrationManagementConfigurer` is a `BeanPostProcessor`
so it must not have direct dependency injection for other beans.
In our case it is a `MetricsCaptor` injected from the
`IntegrationManagementConfiguration`

* Fix `IntegrationManagementConfiguration` and `IntegrationManagementConfigurer`
to rely on the `ObjectProvider<MetricsCaptor>` instead

Tested against latest Spring Boot


artembilan:
Back-ported to `5.3.x` & `5.2.x` as https://github.com/spring-projects/spring-integration/commit/af725a841f878da4739bcb08740ab9407938ecf7 after fixing conflicts.

PR #3436, opened 2020-12-01
GH-3434: Optimize `unlinkAvailable` in Redis

### LINKED ISSUE #3434: RedisLockRegistry UNLINK command (optimization)

5.4.1

**Describe the bug**

The RedisLockRegistry is creating RedisLock instances that test the Redis backend if it supports the UNLINK command.
If it does not, a local variable `unlinkAvailable` is set to false and the error is logged as a warning.
This causes the RedisLock instance to use the old DELETE command.

There are two issues with this.

First one is that each new RedisLock instance will do the above logic over and over again (for every new instance once). Since Lock instances are mostly short lived that is a lot of UNLINK command testing towards the same Redis backend. We can assume the Redis version does not change during the runtime of the application.

The other, smaller, issue is that the log warning message in its current form can't be silenced/minimized and always outputs an Exception stack trace which adds a great deal of noise to the log.

**To Reproduce**

Point the RedisLockRegistry to an older version of Redis that does not support the UNLINK command.

**Expected behavior**

This detection of UNLINK support should be moved to the RedisLockRegistry and the `unlinkAvailable` command should be passed as a constructor param to RedisLock (base idea by @artembilan https://github.com/spring-projects/spring-integration/issues/3041#issuecomment-735955025)

The log warning message should be logged without the stack trace and if the stack trace is important it should be guarded under a `log.isDebugEnabled` or similar.

**Sample**

Sorry... might be too basic of an example to warrant a sample. It would have to be some kind of docker compose example to include an older version of Redis... or TestContainers...

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3434

Each `RedisLock` has their own `unlinkAvailable` property
therefore we try to `unlink` on every unlock call and also log a WARN
about unavailability of this command in Redis, plus exception

* Move `unlinkAvailable` property to the `RedisLockRegistry` level
to have a check only once on a first `unlock()` call
* Move the whole exception logging onto the DEBUG level, leaving the WARN
only with the `ex.getMessage()`
* Optimize logging the same way in the `RedisMessageStore`

Cherry-pick to `5.3.x & 5.2.x`

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2007-2019 the original author or authors.
+ * Copyright 2007-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


garyrussell:
And cherry-picked to 5.3.x, 5.2.x.

PR #3440, opened 2020-12-07
GH-3438: Add Build from Source into CONTRIBUTING

### LINKED ISSUE #3438: Create a 'Build from Source' section in the README.md documentation
Hello Spring Integration Team:

**Expected Behavior**

When I arrive to the [spring-integration](https://github.com/spring-projects/spring-integration) project there is no information about what JDK version should be used to import the project in the IDE (Java 8, 11?). It of course according for a contributor perspective

For example the [spring-framework](https://github.com/spring-framework/spring-framework) project has in its readme file the `Build from Source` section and indicating: 

* See the [Build from Source](https://github.com/spring-projects/spring-framework/wiki/Build-from-Source) Wiki page and the [CONTRIBUTING.md](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md) file. 

Where in the first link has the respective indications

**Current Behavior**

The readme file does not have a reference about neither `Build from Source` nor `See the Build from Source Wiki page` contents (something similar).

Exists the following in [Wiki](https://github.com/spring-projects/spring-integration/wiki), where appears in the right the [Requirements](https://github.com/spring-projects/spring-integration/wiki/Requirements) link. But really out of date

**Context**

Is not clear with what JDK I should use my IDE to work around (to contribute) Spring Integration.

Thanks for your understanding

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3438

It is not clear at the moment what Java version and what build system is used
in the project from the `CONTRIBUTING.adoc`

* Add `Build from Source` section into the `CONTRIBUTING.adoc`
* Point to the `CONTRIBUTING.adoc` from the top level `README.md`
* Fix the current Java version in the `README.md`
* Mention that `Squash commits` must be avoided for subsequent commits into the PR

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -51,6 +51,69 @@ _you should now see 'upstream' in addition to 'origin' where 'upstream' is the S
 7. `git branch -a`
 _you should see branches on origin as well as upstream, including 'master'_
 
+== Build from Source
+
+The build system for the project is https://gradle.org/[Gradle].
+It is recommended to rely on the `wrapper` provided in the project code based and use a `gradlew` script from command line for the target operation system.
+The current Gradle version in use you can obtain from the `/gradle/gradle-wrapper.properties` file in the source tree.
+It is also recommended to use a Gradle import feature of your IDE for the best contribution experience.
+
+The minimum JDK version at the moment is `11` (mostly for tests - production code is still `1.8`).
+You always can find the currently required Java version in the `build.gradle`.
+For example, for the current project version:
+
+----
+compileJava {
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+}
+
+compileTestJava {
+    sourceCompatibility = 11
+}
+
+compileKotlin {
+    kotlinOptions {
+        jvmTarget = '1.8'
+    }
+}
+compileTestKotlin {
+    kotlinOptions {
+        jvmTarget = '11'
+    }
+}
+----
+
+To build and install jars into your local Maven cache:
+
+----
+./gradlew publishToMavenLocal
+----
+
+To build api Javadoc (results will be in `build/api`):
+
+----
+./gradlew api
+----
+
+To build reference documentation (results will be in `build/reference`):
+
+----
+./gradlew reference
+----
+
+To build complete distribution including `-dist`, `-docs`, and `-schema` zip files (results will be in `build/distributions`)
+
+----
+./gradlew dist
+----
+
+You can build and test only specific module if your contribution is only over there
+
+----
+./gradlew :spring-integration-webflux:test
+----
+
 == A Day in the Life of a Contributor
 
 * _Always_ work on topic branches (Typically use the GitHub issue ID as the branch name).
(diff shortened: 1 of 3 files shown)


artembilan:
@manueljordan , FYI.

CONTRIBUTING.adoc
```diff
@@ -51,6 +51,69 @@ _you should now see 'upstream' in addition to 'origin' where 'upstream' is the S
...
+./gradlew publishToMavenLocal
+----
+
+To build api Javadoc (results will be in `build/api`):
+
+----
+./gradlew api
+----
+
+To build reference documentation (results will be in `build/reference`):
```
garyrussell, line 99:
```suggestion
To build the reference documentation (results will be in `build/docs/asciidoc` and `build/docs/asciidocPdf`):
```

garyrussell (changes requested):

manueljordan:
Huge thanks!. 

Something interesting is about that `spring-integration` about Gradle has the `compileJava`  and `compileTestJava` blocks. It is very valuable and straightforward to indicate all about the JDK to use, but the same blocks do not exists in spring-framework for its [build.gradle](https://github.com/spring-projects/spring-framework/blob/master/build.gradle) file.

As usual, huge thanks for your efforts.

artembilan:
@manueljordan ,

Perhaps Spring Framework really relies on the default Java version. 
See Gradle docs: https://docs.gradle.org/current/userguide/java_plugin.html#sec:java-extension

> Java version compatibility to use when compiling Java source. Default value: version of the current JVM in use.

Although I see this: https://github.com/spring-projects/spring-framework/blob/master/gradle/ide.gradle.
Not sure how this works for IntelliJ IDEA though...

manueljordan:
Interesting, I assumed all the spring projects have the same 'rules' or 'conventions'

Thanks for the links. About the latter, yes it is for Eclipse. but I thought at a first glance: Why that JDK settings were not declared in an agnostic way through the `build.gradle` file from the beginning?. I hope you see my point.

Best regards ...

artembilan:
Well, I pointed you to Gradle preferences according Java version selection by default.

All Spring projects are handled by different teams and to be honest the rules what and how to use to build the specific project is fully up to that team. There is really no strict requirements: someone prefers Gradle, someone Maven. Some people use Eclipse, other - IDEA. And so on. We indeed try to generalize the code style and versions compatibility, but how project is built, tested and released is fully up to project team and really doesn't effect the other projects.

I admit that I don't face the concern you raising because I just don't contribute to all the project. The projects which are maintained by my team indeed have the same contribution process, but others where I contribute have different style because it is still managed by other team where they have Maven and Jenkins preferences. 

I think we just need to get used to the project specifics we would like to contribute rather than try to make everyone to follow some standard where there is really no such a perfect standard. So, IMHO it is better to concentrate on the code than keep forever discussion how all the project should follow the same convention for development.

manueljordan:
Totally agree about the latest paragraph, I wrote that with the best intentions and how an observation, not how a command or order.

I remember many spring projects moved from Maven to Gradle (not sure if some one is based on Maven yet)

As usual, huge thanks for all.

PR #3441, opened 2020-12-08
More GH Actions goodies

artembilan:
* Add RabbitMQ, MongoDB & Redis services into GH actions
* Remove Travis


.github/workflows/pr-build-workflow.yml
```diff
@@ -10,12 +10,30 @@ jobs:
     runs-on: ubuntu-latest
 
     services:
+
       rabbitmq:
         image: rabbitmq
         ports:
           - 5672:5672
```
garyrussell, line 17: `- 15762:15762` for management plugin?

garyrussell: Forget that; not used in this project; only spring-rabbit.

artembilan:
It doesn't show the time for each module to pass its tests, but I think we should believe it that it works well.
And therefore it is ready for merging.

Thanks

PR #3449, opened 2020-12-22
Special handle for void in the RSocketOutGateway

artembilan:
To avoid extra `Publisher` subscription work in the target
application "flatten" a returned empty `Flux` for the `void`
expected type into a `Mono.empty()` for automatic subscription
on the output channel


artembilan:
The Gitter discussion on the matter: https://gitter.im/spring-projects/spring-integration?at=5fdb3542d75db604c1959651

PR #3456, opened 2021-01-07
GH-3454: From MQTT conversion error - to error ch

### LINKED ISSUE #3454: Paho loses connection if the message converter returns null

5.2.8.RELEASE

**Describe the bug**

The paho mqtt client loses the connection if a custom converter returns null. In my case the application pairs each incoming message with some mandatory information. If the information are absent for any reason we return null in order to stop the spi message delivery.

#### MqttPahoMessageDrivenChannelAdapter
```java
@Override
public void messageArrived(String topic, MqttMessage mqttMessage) {
	Message<?> message = this.getConverter().toMessage(topic, mqttMessage);
	try {
		sendMessage(message);
	}
	catch (RuntimeException e) {
		logger.error("Unhandled exception for " + message.toString(), e);
		throw e;
	}
}
```

#### MessageProducerSupport
```java
protected void sendMessage(Message<?> messageArg) {
		Message<?> message = messageArg;
		if (message == null) {
			throw new MessagingException("cannot send a null message");
		}
              
                ...
}
```

**To Reproduce**

1. Implement a PahoMessageConverter
2. Make the `toMessage` method return null
3. Send a message and observe the client getting disconnected

**Expected behavior**

The client should drop the message entirely if the converter returns null.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3454

The message converter may return null when we try to covert from the
MQTT message.
The thrown exception may also reset the client connect.

* Fix `MqttPahoMessageDrivenChannelAdapter` to catch any conversion errors
(including `null` result) and try to send an `ErrorMessage` with that info
into the provided `errorChannel`.
Otherwise re-throw it as as

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -42,6 +42,8 @@ import org.springframework.integration.mqtt.support.MqttUtils;
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
+import org.springframework.messaging.converter.MessageConversionException;
+import org.springframework.messaging.support.GenericMessage;
 import org.springframework.util.Assert;
 
 /**
@@ -205,7 +207,7 @@ public class MqttPahoMessageDrivenChannelAdapter extends AbstractMqttMessageDriv
 		}
 		catch (Exception ex) {
 			logger.error(ex, "Exception while connecting and subscribing, retrying");
-			this.scheduleReconnect();
+			scheduleReconnect();
 		}
 	}
 
@@ -272,7 +274,7 @@ public class MqttPahoMessageDrivenChannelAdapter extends AbstractMqttMessageDriv
 			super.removeTopic(topic);
 		}
 		catch (MqttException e) {
-			throw new MessagingException("Failed to unsubscribe from topic " + Arrays.asList(topic), e);
+			throw new MessagingException("Failed to unsubscribe from topic(s) " + Arrays.toString(topic), e);
 		}
 		finally {
 			this.topicLock.unlock();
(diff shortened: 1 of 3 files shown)


artembilan:
I will do back-port to `5.2.x` after review and merging: it does not support `MessageBuilder` contract yet.

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -399,19 +401,52 @@ public synchronized void connectionLost(Throwable cause) {
...
 		}
-		catch (RuntimeException ex) {
-			logger.error(ex, () -> "Unhandled exception for " + message.toString());
-			throw ex;
+		catch (Exception ex) {
+			conversionError = ex;
+		}
+
+		if (builder == null && conversionError == null) {
+			conversionError = new IllegalStateException("'MqttMessageConverter' produced a null for message");
```
garyrussell, line 432:
```suggestion
			conversionError = new IllegalStateException("'MqttMessageConverter' returned 'null'");
```

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -399,19 +401,52 @@ public synchronized void connectionLost(Throwable cause) {
...
+			if (!sendErrorMessageIfNecessary(message, conversionError)) {
+				MessageConversionException conversionException;
+				if (conversionError instanceof MessageConversionException) {
+					conversionException = (MessageConversionException) conversionError;
+				}
+				else {
+					conversionException = new MessageConversionException(message, "Failed to convert from MQTT Message",
+							conversionError);
+				}
+				throw conversionException;
```
garyrussell, line 446: Should we still throw here? Or just log as ERROR?

src/reference/asciidoc/mqtt.adoc
```diff
@@ -164,6 +164,9 @@ StaticMessageHeaderAccessor.acknowledgment(someMessage).acknowledge();
 ----
 ====
 
+Starting with version `5.2.11`, when message converter throws an exception or returns `null` from the `MqttMessage` conversion, the `MqttPahoMessageDrivenChannelAdapter` sends an `ErrorMessage` into the provided `errorChannel`.
+Re-throws this conversion error otherwise into an MQTT client callback.
```
garyrussell, line 168:
```suggestion
Starting with version `5.2.11`, when the message converter throws an exception or returns `null` from the `MqttMessage` conversion, the `MqttPahoMessageDrivenChannelAdapter` sends an `ErrorMessage` into the `errorChannel`, if provided.
Re-throws this conversion error otherwise into an MQTT client callback.
```

Also see my comment in the code about re-throwing.

garyrussell (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
artembilan: Well, I tried to make it consistent with what we have so far in the `MessageProducerSupport.sendMessage()`:
```
try {
	this.messagingTemplate.send(getRequiredOutputChannel(), message);
}
catch (RuntimeException ex) {
	if (!sendErrorMessageIfNecessary(message, ex)) {
		throw ex;
	}
}
```

We probably may consider some general `ErrorHandler` strategy. However it looks like we have it already for this kind of channel adapters via an `ErrorHandlingTaskExecutor` impl. Nevertheless I don't see that we have some way to inject it into the `MqttAsyncClient`...

Probably the matter for a separate issue and discussion over there...

garyrussell: Right, but the problem here is that throwing an exception causes the client to close. Conversion exceptions are generally fatal anyway (no point in retrying) since a redelivery will fail again.

However, it would be a behavior change so we'd probably have to add a flag.

Or, we can just leave it as you have it; up to you.

artembilan: Well, we have a flag - `errorChannel` and no exceptions are thrown back to the client!
I would stay with this behavior for a while: probably everybody is going to be happy to swallow and process those wrong messages in the error flow instead of just silent (log) drop to satisfy the client connection stability.

artembilan:
@garyrussell ,

let me know if anything else you'd like to see fixed with this PR!
Thanks

garyrussell (approved):

garyrussell:
Cherry-picked to 5.4.x and to 5.3.x after resolving conflicts.

Ready for 5.2.x backport.

artembilan:
Back-ported to `5.2.x` as https://github.com/spring-projects/spring-integration/commit/71d41757cd9655b41e334926968d247cf9e267e1

Issue #3455, opened 2021-01-07
Breaking Change in Version v5.4.2


deen13:
**In what version(s) of Spring Integration are you seeing this issue?**

Spring Boot 2.4.1
Spring Integration 5.4.2

**Describe the bug**

The commit [fbf06e8](https://github.com/spring-projects/spring-integration/commit/fbf06e8bb592e6ad885e4a33384f53e58dc293cf) adds the new `toMessageBuilder(String topic, MqttMessage mqttMessage)` function which is wrapped by the `toMessage(String topic, MqttMessage mqttMessage)`. After we've migrated spring boot from 2.2.9.RELEASE to 2.4.1 our message conversion stopped working. 

#### Client Factory

```java
val adapter = MqttPahoMessageDrivenChannelAdapter("ProdasoConsumer-${UUID.randomUUID()}", mqttClientFactory)

adapter.addTopic("mytopic", 1)

adapter.setCompletionTimeout(5000)
adapter.setOutputChannelName("routingChannel")
adapter.setConverter(mqttMessageConverter) // this is not considered anymore
```

The converter is no longer considered and the handler receives the raw message which results in a ClassCastException.

#### Handler

```java
@ServiceActivator(inputChannel = "lastwillChannel")
fun lastWillEvent(message: Message<MessageWrapper<String>>) 
```

**Question**

How can we make the adapter consider our converter?

garyrussell:
>How can we make the adapter consider our converter?

Unfortunately, this is a breaking change that we failed to document in the [migration guide](https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.2-to-5.3-Migration-Guide). You can change your converter to return a `AbstractIntegrationMessageBuilder<?>` instead of a `Message<?>`.

artembilan:
I don't think this is a breaking change. This is a new API in that new `5.3` version which you just miss to implement in your solution.

We will add migration guide note anyway to avoid the further discussion if it breaks or not.

> How can we make the adapter consider our converter?

Just implement that `toMessageBuilder()` like this:
```
public AbstractIntegrationMessageBuilder<?> toMessageBuilder(String topic, MqttMessage mqttMessage) {
	return MutableMessageBuilder.fromMessage(toMessage(topic, mqttMessage));
}
```

We probably can apply this trick as a `default` one in our `MqttMessageConverter` contract.
Plus it looks like we need to mention this `MqttMessageConverter` in the docs and what end-user have to implement.

[...]

PR #3458, opened 2021-01-08
GH-3455: default MqttMessConv.toMessageBuilder()

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3455

After introduction `MqttMessageConverter.toMessageBuilder()`
(https://github.com/spring-projects/spring-integration/issues/3181)
the existing `MqttMessageConverter` must also implement this new method
where in most cases `toMessage()` must call `toMessageBuilder()` instead though.

* Make `MqttMessageConverter.toMessageBuilder()` as a `default` with a delegation
to the `toMessage()` allowing target implementors to keep their code during
migration without any breaking changes

**Cherry-pick to 5.3.x**


garyrussell:
and cherry-picked to 5.3.x as 040fa703ca78ddadea8cf8cfc67213962df4696b

PR #3459, opened 2021-01-11
GH-3450: Expose PubSub requireSubscribers option

### LINKED ISSUE #3450: MessageProducerSupport loses exception on application shutdown
5.3.0.RELEASE, although this would manifest in any version from the last couple of years.

**Describe the bug**
Upon application context shutdown, `BroadcastingDispatcher`'s list of `MessageHandler`s is cleared. This in itself is not a problem; however it causes [this block](https://github.com/spring-projects/spring-integration/blob/0331933c168f9096195999592a7f169d39f238d2/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java#L210-L214) in `MessageProducerSupport` to lose the exception.

**To Reproduce**

The downstream manifestation on Cloud Pub/Sub implementation of `MessageProducerSupport` was reported by a user in spring-cloud/spring-cloud-gcp#2615 as unexpected acknowledgment of an known bad message (message handling code explicitly throws a `RuntimeException`) when application was shutdown midway through processing. The example can be found in https://github.com/thayhoang/pubsub-demo; however it requires a working GCP Cloud Pub/Sub setup. In this particular case, the application is a Spring Cloud Stream sink, and the difference between the correct scenario (when exception bubbles up to the binder correctly) and the incorrect one (when, after application context shutdown, the binder completes successfully) is the presence or absence of `org.springframework.cloud.stream.binder.FinalRethrowingErrorMessageHandler`.

I've also put together a [unit test demonstrating the issue](https://github.com/elefeint/spring-integration/commit/3bb1d49f025919b5fe9ab5bc88c1208170b29496). the setup is not ideal because it manually unsubscribes the error channel and stops the `MessageProducerSupport`, but it shows both the problem and the potential fix.

**Expected behavior**

Exception from message handlers should always be propagated upstream, unless properly handled by the error handler. Missing error handler and/or known context shutdown should not be considered handled and should instead result in exception propagation.

**Sample**

[Unit test](https://github.com/elefeint/spring-integration/blob/3bb1d49f025919b5fe9ab5bc88c1208170b29496/spring-integration-core/src/test/java/org/springframework/integration/endpoint/MessageProducerSupportTests.java#L123-L165)

[Proposed fix](https://github.com/elefeint/spring-integration/blob/3bb1d49f025919b5fe9ab5bc88c1208170b29496/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java#L211-L215)

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3450

* Add `PublishSubscribeChannel` ctors with the `requireSubscribers` option
with a direct delegation to the same option of underlying `BroadcastingDispatcher`
* Add factory methods to the `Channels` & `MessageChannels` to expose
this new `requireSubscribers` option
* Expose an XML `require-subscribers` attribute for the `<publish-subscribe-channel`
* Document this new `requireSubscribers` option
* Introduce a global `spring.integration.channels.error.requireSubscribers` property
for a default `errorChannel`
* Document this property and explain its default `true` in the `error-handling.adoc`
* Remove docs for `spring.integration.postProcessDynamicBeans` since it purpose was
removed since version `5.1` in favor of "always post-process behavior" as it was
always with all the `BeanPostProcessor`s

### DIFF
--- a/spring-integration-core/src/main/resources/org/springframework/integration/config/spring-integration.xsd
+++ b/spring-integration-core/src/main/resources/org/springframework/integration/config/spring-integration.xsd
@@ -537,12 +537,22 @@
 					</xsd:attribute>
 					<xsd:attribute name="min-subscribers" type="xsd:string">
 						<xsd:annotation>
-							<xsd:documentation><![CDATA[
+							<xsd:documentation>
 								Specifies the minimum subscribers required to be subscribed to this channel; if the minimum number
 								of subscribers receive the message, the send is deemed to be successful (returns true).
 								Defaults to 0.
-							]]></xsd:documentation>
+							</xsd:documentation>
 						</xsd:annotation>
+					</xsd:attribute>
+					<xsd:attribute name="require-subscribers" default="false">
+						<xsd:annotation>
+							<xsd:documentation>
+								Indicates if this channel may ignore or not a message when there is no subscribers at all.
+							</xsd:documentation>
+						</xsd:annotation>
+						<xsd:simpleType>
+							<xsd:union memberTypes="xsd:boolean xsd:string"/>
+						</xsd:simpleType>
 					</xsd:attribute>
 					<xsd:attributeGroup ref="subscribersAttributeGroup"/>
 				</xsd:extension>
--- a/spring-integration-core/src/main/resources/org/springframework/integration/config/spring-integration.xsd
+++ b/spring-integration-core/src/main/resources/org/springframework/integration/config/spring-integration.xsd
@@ -547,7 +547,7 @@
 					<xsd:attribute name="require-subscribers" default="false">
 						<xsd:annotation>
 							<xsd:documentation>
-								Indicates if this channel may ignore or not a message when there is no subscribers at all.
+								Indicates if this channel may ignore or not a message when there are no subscribers.
 							</xsd:documentation>
 						</xsd:annotation>
 						<xsd:simpleType>
--- a/src/reference/asciidoc/channel.adoc
+++ b/src/reference/asciidoc/channel.adoc
@@ -794,8 +794,8 @@ public MessageChannel pubsubChannel() {
 NOTE: The `apply-sequence` value is `false` by default so that a publish-subscribe channel can send the exact same message instances to multiple outbound channels.
 Since Spring Integration enforces immutability of the payload and header references, when the flag is set to `true`, the channel creates new `Message` instances with the same payload reference but different header values.
 
-Starting with version 5.4.3, the `PublishSubscribeChannel` can also be configured with the `requireSubscribers` option of its `BroadcastingDispatcher` to indicate that this channel is not going to ignore a message silently when there is no any subscribers to it.
-A `MessageDispatchingException` with a `Dispatcher has no subscribers` message is thrown when no subscribers and this option is set to `true`.
+Starting with version 5.4.3, the `PublishSubscribeChannel` can also be configured with the `requireSubscribers` option of its `BroadcastingDispatcher` to indicate that this channel will not ignore a message silently when it has no subscribers.
+A `MessageDispatchingException` with a `Dispatcher has no subscribers` message is thrown when there are no subscribers and this option is set to `true`.
 
 [[channel-configuration-executorchannel]]
 ===== `ExecutorChannel`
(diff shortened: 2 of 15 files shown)


artembilan:
I will back-port it to `5.4.x` with the `requireSubscribers = false` for global `errorChannel` to avoid breaking change after review and agreement here.

Thanks

garyrussell:
It's not clear to me why this would need to be false in 5.4.x; the default error channel always has a subscriber so it would only fail if someone explicitly stops the logging hander's endpoint. I just don't see it breaking otherwise. Am I missing something?

spring-integration-core/src/main/resources/org/springframework/integration/config/spring-integration.xsd
```diff
@@ -537,12 +537,22 @@
...
 								of subscribers receive the message, the send is deemed to be successful (returns true).
 								Defaults to 0.
-							]]></xsd:documentation>
+							</xsd:documentation>
+						</xsd:annotation>
+					</xsd:attribute>
+					<xsd:attribute name="require-subscribers" default="false">
+						<xsd:annotation>
+							<xsd:documentation>
+								Indicates if this channel may ignore or not a message when there is no subscribers at all.
```
garyrussell, line 550:
```suggestion
								Indicates if this channel may ignore or not a message when there are no subscribers.
```

src/reference/asciidoc/channel.adoc
```diff
@@ -794,6 +794,9 @@ public MessageChannel pubsubChannel() {
 NOTE: The `apply-sequence` value is `false` by default so that a publish-subscribe channel can send the exact same message instances to multiple outbound channels.
 Since Spring Integration enforces immutability of the payload and header references, when the flag is set to `true`, the channel creates new `Message` instances with the same payload reference but different header values.
 
+Starting with version 5.4.3, the `PublishSubscribeChannel` can also be configured with the `requireSubscribers` option of its `BroadcastingDispatcher` to indicate that this channel is not going to ignore a message silently when there is no any subscribers to it.
+A `MessageDispatchingException` with a `Dispatcher has no subscribers` message is thrown when no subscribers and this option is set to `true`.
```
garyrussell, line 798:
```suggestion
Starting with version 5.4.3, the `PublishSubscribeChannel` can also be configured with the `requireSubscribers` option of its `BroadcastingDispatcher` to indicate that this channel will not ignore a message silently when it has no subscribers.
A `MessageDispatchingException` with a `Dispatcher has no subscribers` message is thrown when there are no subscribers and this option is set to `true`.
```

src/reference/asciidoc/error-handling.adoc
```diff
@@ -70,3 +70,6 @@ With Java & Annotation configuration, a resource is a `@Configuration` class and
 In most case the target integration flow solution is based on the out-of-the-box components and their configuration options.
 When an exception happens at runtime, there is no any end-user code involved in stack trace because an execution is against beans, not their configuration.
 Including a resource and source of the bean definition helps to determine possible configuration mistakes and provides better developer experience.
+
+Starting with version 5.5, the default error channel is configured with the `requireSubscribers = true` to not ignore messages silently when there is no subscribers on this channel (e.g. when application context is stopped).
+In this case a `MessageDispatchingException` is thrown which may lend on the client callback of the inbound channel adapter to negatively acknowledge (or roll back) an original message in the source system for redeliveries or other future consideration.
```
garyrussell, line 75:
```suggestion
Starting with version 5.5, the default error channel is configured with the property `requireSubscribers = true` to not silently ignore messages when there are no subscribers on this channel (e.g. when application context is stopped).
In this case a `MessageDispatchingException` is thrown which may lend on the client callback of the inbound channel adapter to negatively acknowledge (or roll back) an original message in the source system for redeliveriy or other future consideration.
```

garyrussell (changes requested):

artembilan:
>  if someone explicitly stops the logging handler's endpoint

It happens automatically on application context close.
I'm really OK to back-port it as `true`, too - I'm always nervous when we lose messages. 
So, even if this breaks the current behavior when we lose messages silently, it really might be better to keep it throwing an exception by default. Fortunately with that option anyone can come back to the previous dangerous behavior.

Just my conclusion to justify your back-port concern.

So, id we are on the same page in the end, feel free to cherry-pick this commit to `5.4.x`.
Otherwise let's continue a discussion what and how to back-port.

Thanks

garyrussell:
Agreed; I just think we need a bit of doc polishing to tell them how to revert to the old behavior.

garyrussell (approved):

garyrussell:
Cherry-picked to 5.4.x as c880a345f7

PR #3466, opened 2021-01-19
GH-3446: Stream support in the MessageGroupStore

### LINKED ISSUE #3446: DelayHandler, ConfigurableMongoDbMessageStore and PersistentMessageGroup issue at startup/load

5.3.1.RELEASE

**Describe the bug**

If you have set up a DelayHandler using a ConfigurableDbMessageStore and the MongoDB collection behind it is very big (e.g. 1M records), upon startup the DelayHandler will try to re-schedule all messages in the delay store. Because we are using ConfigurableDbMessageStore which extends AbstractMessageGroupStore and also lazyLoadMessageGroups, this class, sets the persistentMessageGroupFactory as a SimpleMessageGroupFactory(SimpleMessageGroupFactory.GroupType.PERSISTENT) which returns PersistentMessageGroup. When the PersistentMessageGroup is called to retrieve the messages, it loads ALL messages in memory before returning the iterator 

https://github.com/spring-projects/spring-integration/blob/08316577cea0be542559b6510d7ef5e0d9c87aa7/spring-integration-core/src/main/java/org/springframework/integration/store/PersistentMessageGroup.java#L171

https://github.com/spring-projects/spring-integration/blob/08316577cea0be542559b6510d7ef5e0d9c87aa7/spring-integration-core/src/main/java/org/springframework/integration/store/PersistentMessageGroup.java#L186

https://github.com/spring-projects/spring-integration/blob/08316577cea0be542559b6510d7ef5e0d9c87aa7/spring-integration-core/src/main/java/org/springframework/integration/store/PersistentMessageGroup.java#L217

https://github.com/spring-projects/spring-integration/blob/08316577cea0be542559b6510d7ef5e0d9c87aa7/spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java#L598

https://github.com/spring-projects/spring-integration/blob/08316577cea0be542559b6510d7ef5e0d9c87aa7/spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java#L570
Depending on the available memory, OOM happens. 

**To Reproduce**

Create a collection behind the DelayHandler/ConfigurableDbMessageStore and load it with millions of records. Start an application that makes use of the DelayHandler

**Expected behavior**

We believe it is OK to reschedule messages upon startup but not loading them all at once in memory. Maybe doing it via MongoDB streaming API?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3446

* For better resources utilization provide a `Stream<Message<?>>` API
on the `MessageGroupStore`, `MessageGroup` and `MessageGroupQueue`
* Use this API in the `DelayHandler` when it reschedules persisted messages

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/DelayHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
 
 import org.aopalliance.aop.Advice;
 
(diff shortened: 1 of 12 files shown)


garyrussell (changes requested):
I don't see any new tests for these `stream` methods.

artembilan:
> I don't see any new tests for these stream methods.

All the delayer tests against persistent message stores do their job I believe.
This is only a place where we actually use messages from the stored group.

garyrussell (approved):

PR #3472, opened 2021-01-22
GH-3461: Readme Polishing, eclipse and build

### LINKED ISSUE #3461: Is not possible build Spring Integration, it fails in many modules
**Describe the bug [Problem]**

Is not possible Build Spring Integration

**To Reproduce**

In MacOS and Windows I have the following situation

I have STS 4.9.0.RELEASE for a workspace dedicated for Spring Integration.
And configured for Java 11 according with the [Checking out and Building](https://github.com/spring-projects/spring-integration#checking-out-and-building) section.

Prior to import `Spring Integration` I did

* `git clone`
* `added remote repository`
* `git fetch`
* `git merge upstream/master`

So I am sure I have the latest and stable code of the project

Then `Spring Integration` is imported and then according with the [Using Eclipse](https://github.com/spring-projects/spring-integration#using-eclipse) section it indicates about `./gradlew eclipse` command. Of course I did do the inverse, I mean, I imported and then executed the command. Anyway it worked 

Then according with the [Checking out and Building](https://github.com/spring-projects/spring-integration#checking-out-and-building) section it indicates about `./gradlew build` command, I executed twice because the first failed and I wanted to confirm the same error message for a second execution. Seems there are more errors about test fails. Below two different executions (and fails in two different modules).

Is stable the project? Or there are some missing step instructions?

BTW Just in case, I get the same even with:

This was tried on Mac, I am going to do the same steps on other machine, Windows 10.

**Expected behavior**

Build in peace the `Spring Integration` project

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3461

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,17 @@ To check out the project and build from the source, do the following:
 
     git clone git://github.com/spring-projects/spring-integration.git
     cd spring-integration
-    ./gradlew build
+    ./gradlew clean test
+
+    or
+
+    ./gradlew clean testAll
+
+The latter runs additional tests (those annotated with `@LongRunningIntegrationTest`); it is a more thorough test but takes quite a lot longer to run.
+
+The test results are captured in `build/reports/tests/test` (or `.../testAll`) under each module (in HTML format).
+
+Add `--continue` to the command to perform a complete build, even if there are failing tests in some modules; otherwise the build will stop after the current module(s) being built are completed.
 
 **NOTE:** While Spring Integration runs with Java SE 8 or higher, a Java 11 compiler is required to build the project.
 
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Not related to your change, but possible fixes in the same file:

1. s/https://projects.spring.io/spring-integration/https://spring.io/projects/spring-integration - no need to wait for redirect
2. `Using IntelliJ IDEA` section. I would change just to this:
> To import the project into IntelliJ IDEA just follow this menu:
`File -> Open... -> and select build.gradle from spring-integration project root directory`

manueljordan:
Hello

Being polite, about: 

```
Add `--continue` to the command to perform a complete build, 
even if there are failing tests in some modules; 
otherwise the build will stop after the current module(s) being built are completed.
```

The reader would be confused, he would think:

- Why the Spring Integration project would fail in some Tests? 
- Is not supposed to be totally ready for production?

Therefore consider to explain or add, about why a SI's test would fail.

garyrussell:
The stable master should not have any failures; if it does, they should be reported.

This is for new changes.

Let's say you are working on a change to `spring-integration-core`, and let's say the change breaks 2 different ways in 2 different modules, say -amqp breaks one way and -jms breaks another way.

If you don't use `--continue`, you will only find the -amqp problem and will have to fix that and run another build to fix the -jms problem.

It could save time (and a build) if you fix both problems at the same time.

I don't think this is necessary to explain in the read me.

garyrussell:
But if you feel strongly about it; pleas issue a PR.

manueljordan:
Gary you are totally correct, but my point was totally based on my original post #3461. I just downloaded the SI and the master fails. I hope you see my point. Tomorrow that I run the tests I must share the outputs.

garyrussell:
We run many, many, builds each week with no failures; locally, and on our Bamboo servers; yes, there are a couple of flaky tests that fail once-in-a-while (usually race conditions in the tests themselves). We fix them when it gets annoying.

If you are seeing lots of failures, there is something unusual about your environment.

We would be happy to look at the failures, but it might be hard to figure them out.

manueljordan:
I wrote the post with the best intentions. We must assume the developer thinks there are none test with a fail scenario.

>>there are a couple of flaky tests that fail once-in-a-while (usually race conditions in the tests themselves).

That's is the reason that would be added about why use the `--continue` for the scenario when the tests fail when the project just was cloned and build

PR #3481, opened 2021-01-28
Expose `error.ignoreFailures` global property

artembilan:
The global default `errorChannel` can have more then only one logging subscriber.
There is no guarantee in which order subscribers are called and one of them fails
the rest are not called.

* Expose `spring.integration.channels.error.ignoreFailures` global property for
default `errorChannel` and make it `true` by default for better end-user experience
when it is expected to have error being handled by all the subscribers

### DIFF
--- a/src/reference/asciidoc/whats-new.adoc
+++ b/src/reference/asciidoc/whats-new.adoc
@@ -18,9 +18,13 @@ If you are interested in more details, see the Issue Tracker tickets that were r
 [[x5.5-general]]
 === General Changes
 
-
 All the persistent `MessageGroupStore` implementation provide a `streamMessagesForGroup(Object groupId)` contract based on the target database streaming API.
 See <<./message-store.adoc#message-store,Message Store>> for more information.
+
+A `spring.integration.channels.error.requireSubscribers=true` global property is added to indicate that global default `errorChannel` must be configured with the `requireSubscribers` option (or not).
+A `spring.integration.channels.error.ignoreFailures=true` global property is added to indicate that global default `errorChannel` must ignore (or not) dispatching errors and pass the message to the next handler.
+See <<./configuration.adoc#global-properties,Global Properties>> for more information.
+
 [[x5.5-amqp]]
 ==== AMQP Changes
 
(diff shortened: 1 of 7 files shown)


src/reference/asciidoc/whats-new.adoc
```diff
@@ -18,9 +18,13 @@ If you are interested in more details, see the Issue Tracker tickets that were r
 [[x5.5-general]]
 === General Changes
 
-
 All the persistent `MessageGroupStore` implementation provide a `streamMessagesForGroup(Object groupId)` contract based on the target database streaming API.
 See <<./message-store.adoc#message-store,Message Store>> for more information.
+
+A `spring.integration.channels.error.requireSubscribers=true` global property is added to indicate that global default `errorChannel` must be configured with the `requireSubscribers` option (or not).
+A `spring.integration.channels.error.ignoreFailures=true` global property is added to indicate that global default `errorChannel` must ignore (or not) dispatching errors and pass the message to the next handler.
```
garyrussell, line 25: that **the** global (2 places).

garyrussell (changes requested):
Just 2 minor doc changes needed.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -18,9 +18,13 @@ If you are interested in more details, see the Issue Tracker tickets that were r
 [[x5.5-general]]
 === General Changes
 
-
 All the persistent `MessageGroupStore` implementation provide a `streamMessagesForGroup(Object groupId)` contract based on the target database streaming API.
 See <<./message-store.adoc#message-store,Message Store>> for more information.
+
+The `spring.integration.channels.error.requireSubscribers=true` global property is added to indicate that global default `errorChannel` must be configured with the `requireSubscribers` option (or not).
+The `spring.integration.channels.error.ignoreFailures=true` global property is added to indicate that global default `errorChannel` must ignore (or not) dispatching errors and pass the message to the next handler.
```
garyrussell, line 25: Sorry, perhaps I wasn't clear; but your change is good too.

```suggestion
The `spring.integration.channels.error.requireSubscribers=true` global property is added to indicate that the global default `errorChannel` must be configured with the `requireSubscribers` option (or not).
The `spring.integration.channels.error.ignoreFailures=true` global property is added to indicate that the global default `errorChannel` must ignore (or not) dispatching errors and pass the message to the next handler.
```

garyrussell (changes requested):

PR #3485, opened 2021-02-01
GH-3483: Fallback to RestTemplate.converters

### LINKED ISSUE #3483: Consider improving the content type handling in int-http:outbound-gateway
**Expected Behavior**

I place a Java bean on the message payload and invoke a remote service with `<int-http:outbound-gateway>`, leaving `extract-request-payload` to `true`. I know Spring Integration uses `RestTemplate` and I know I have Jackson on my classpath, so I expect it to use the Jackson message converter to convert my payload to a JSON request body, as a programmatic use of the `RestTemplate` would normally do.

**Current Behavior**

```
org.springframework.web.client.RestClientException: No HttpMessageConverter for my.java.Bean and content type "application/x-java-serialized-object"
```

**Context**

The problem lies in `org.springframework.integration.http.outbound.AbstractHttpRequestExecutingMessageHandler.createHttpEntityFromPayload(Message<?>, HttpMethod)`, which does this:

- if the HTTP header `Content-Type` is set, it uses it
- otherwise tries to guess it by virtue of `org.springframework.integration.http.outbound.AbstractHttpRequestExecutingMessageHandler.resolveContentType(Object)`, which, after testing for some relevant cases, falls back to setting `x-java-serialized-object` content type

This breaks the ability of `RestTemplate` to detect which converter to use more smartly, based on the class of the body: `org.springframework.http.converter.AbstractHttpMessageConverter.canWrite(MediaType)` by default returns `true` if the media type is `null` and `org.springframework.http.converter.AbstractHttpMessageConverter.canWrite(Class<?>, MediaType)` then returns `true` or `false` solely depending on the body class. Instead, with the content type set to `x-java-serialized-object`, `org.springframework.http.converter.json.AbstractJackson2HttpMessageConverter.canWrite(Class<?>, MediaType)` for instance will return `false` just because `x-java-serialized-object` is not among the list of supported content types.

The solution here is to add an upstream header enricher which sets the `Content-Type` header to `application/json`, but this is quite annoying and potentially verbose (since I have to add an endpoint and possibly an intermediate channel). If the above resolution of the content type (when unspecified) is indeed needed for some reason I am missing, an acceptable alternative could be to have a `contentType` attribute on the `<int-http:outbound-gateway>` tag, so that I can force the content type to use directly on the gateway, with no need for an extra endpoint.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3483

When there is no reasonable way to determine a `Content-Type`
from the request message, do not set an `application/x-java-serialized-object`
as a fallback and let the `RestTemplate` to determine the target type and
conversion through its `HttpMessageConverter` set

* Remove `application/x-java-serialized-object` fallback from the
`AbstractHttpRequestExecutingMessageHandler`
* Adjust its log messages according `LogAccessor`
* Un`@Disable` `WebFluxDslTests` since fix was done in Spring Security
* Add `HttpRequestExecutingMessageHandlerTests.testNoContentTypeAndSmartConverter()`
* Mention change in `What's New`

### DIFF
--- a/src/reference/asciidoc/whats-new.adoc
+++ b/src/reference/asciidoc/whats-new.adoc
@@ -40,6 +40,6 @@ See <<./redis.adoc#redis,Redis Support>> for more information.
 [[x5.5-http]]
 ==== HTTP Changes
 
-The `HttpRequestExecutingMessageHandler` doesn't fallback to the `application/x-java-serialized-object` content type any more and lets the `RestTemplate` to make a final decision for the request body conversion based on the `HttpMessageConverter` provided.
+The `HttpRequestExecutingMessageHandler` doesn't fallback to the `application/x-java-serialized-object` content type any more and lets the `RestTemplate` make the final decision for the request body conversion based on the `HttpMessageConverter` provided.
 
 See <<./http.adoc#http,HTTP Support>> for more information.
--- a/src/reference/asciidoc/whats-new.adoc
+++ b/src/reference/asciidoc/whats-new.adoc
@@ -36,3 +36,10 @@ See <<./amqp.adoc#amqp,AMQP Support>> for more information.
 
 The `ReactiveRedisStreamMessageProducer` has now setters for all the `StreamReceiver.StreamReceiverOptionsBuilder` options, including an `onErrorResume` function.
 See <<./redis.adoc#redis,Redis Support>> for more information.
+
+[[x5.5-http]]
+==== HTTP Changes
+
+The `HttpRequestExecutingMessageHandler` doesn't fallback to the `application/x-java-serialized-object` content type any more and lets the `RestTemplate` to make a final decision for the request body conversion based on the `HttpMessageConverter` provided.
+
+See <<./http.adoc#http,HTTP Support>> for more information.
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/whats-new.adoc
```diff
@@ -36,3 +36,10 @@ See <<./amqp.adoc#amqp,AMQP Support>> for more information.
 
 The `ReactiveRedisStreamMessageProducer` has now setters for all the `StreamReceiver.StreamReceiverOptionsBuilder` options, including an `onErrorResume` function.
 See <<./redis.adoc#redis,Redis Support>> for more information.
+
+[[x5.5-http]]
+==== HTTP Changes
+
+The `HttpRequestExecutingMessageHandler` doesn't fallback to the `application/x-java-serialized-object` content type any more and lets the `RestTemplate` to make a final decision for the request body conversion based on the `HttpMessageConverter` provided.
```
garyrussell, line 43:
```suggestion
The `HttpRequestExecutingMessageHandler` doesn't fallback to the `application/x-java-serialized-object` content type any more and lets the `RestTemplate` make the final decision for the request body conversion based on the `HttpMessageConverter` provided.
```

garyrussell (changes requested):
Just minor doc polishing.

PR #3487, opened 2021-02-02
GH-3464: Treat 0 as special for polling endpoint

### LINKED ISSUE #3464: Polling 0 messages per run
Current code that is doing polling has this code:

```
private Runnable createPoller() {
		return () ->
				this.taskExecutor.execute(() -> {
					int count = 0;
					while (this.initialized && (this.maxMessagesPerPoll <= 0 || count < this.maxMessagesPerPoll)) {
						if (pollForMessage() == null) {
							break;
						}
						count++;
					}
				});
	}
```

Should this `this.maxMessagesPerPoll <= 0` be actually `this.maxMessagesPerPoll < 0`, since I wanted to control this parameter in the runtime, but cannot set 0 to pause polling for a while. I know that I can change polling interval which would have more/less same effect, but this setting I also think does not work as it should. Was this on purpose or?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3464

The `maxMessagePerPoll <= 0` is considered as an unbound `receive()` call.
End-users would like to have a special treatment for `0` value -
skip the `receive()` call altogether for the current polling cycle.

* Change the logic for a scheduled poller to not call `pollForMessage()`
and just log an INFO when `maxMessagePerPoll == 0`
* Fix reactive poller to deal with `maxMessagePerPoll == 0` properly
* Expose `maxMessagesPerPoll` as a `@ManagedAttribute` to let it to be
modified via Control Bus and JMX
* Test and document a new behavior

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -148,9 +148,9 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 
 	/**
 	 * Configure a cap for messages to poll from the source per scheduling cycle.
-	 * The negative number means unlimited amount of messages.
-	 * Zero means no polling at the moment at all -
-	 * can be treated as pausing if 'maxMessagesPerPoll' is changed eventually.
+* A negative number means retrieve unlimited messages until the {@code MessageSource} returns {@code null}.
+	 * Zero means do not poll for any records - it
+	 * can be considered as pausing if 'maxMessagesPerPoll' is later changed to a non-zero value.
 	 * The polling cycle may exit earlier if the source returns null for the current receive call.
 	 * @param maxMessagesPerPoll the number of message to poll per schedule.
 	 */
@@ -342,7 +342,7 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 					int count = 0;
 					while (this.initialized && (this.maxMessagesPerPoll <= 0 || count < this.maxMessagesPerPoll)) {
 						if (this.maxMessagesPerPoll == 0) {
-							logger.info("No poll for 'maxMessagesPerPoll == 0'");
+							logger.info("Polling disabled while 'maxMessagesPerPoll == 0'");
 							break;
 						}
 						if (pollForMessage() == null) {
@@ -377,7 +377,7 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 										Flux
 												.defer(() -> {
 													if (this.maxMessagesPerPoll == 0) {
-														logger.info("No poll for 'maxMessagesPerPoll == 0'");
+														logger.info("Polling disabled while 'maxMessagesPerPoll == 0'");
 														return Mono.empty();
 													}
 													else {
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
@@ -45,6 +45,7 @@ import org.springframework.integration.transaction.IntegrationResourceHolderSync
 import org.springframework.integration.transaction.PassThroughTransactionSynchronizationFactory;
 import org.springframework.integration.transaction.TransactionSynchronizationFactory;
 import org.springframework.integration.util.ErrorHandlingTaskExecutor;
+import org.springframework.jmx.export.annotation.ManagedAttribute;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessagingException;
@@ -145,8 +146,21 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 		this.adviceChain = adviceChain;
 	}
 
+	/**
+	 * Configure a cap for messages to poll from the source per scheduling cycle.
+	 * The negative number means unlimited amount of messages.
+	 * Zero means no polling at the moment at all -
+	 * can be treated as pausing if 'maxMessagesPerPoll' is changed eventually.
+	 * The polling cycle may exit earlier if the source returns null for the current receive call.
+	 * @param maxMessagesPerPoll the number of message to poll per schedule.
+	 */
+	@ManagedAttribute
 	public void setMaxMessagesPerPoll(long maxMessagesPerPoll) {
 		this.maxMessagesPerPoll = maxMessagesPerPoll;
+	}
+
+	public long getMaxMessagesPerPoll() {
+		return this.maxMessagesPerPoll;
 	}
 
 	public void setErrorHandler(ErrorHandler errorHandler) {
@@ -327,6 +341,10 @@ public abstract class AbstractPollingEndpoint extends AbstractEndpoint implement
 				this.taskExecutor.execute(() -> {
 					int count = 0;
 					while (this.initialized && (this.maxMessagesPerPoll <= 0 || count < this.maxMessagesPerPoll)) {
+						if (this.maxMessagesPerPoll == 0) {
+							logger.info("No poll for 'maxMessagesPerPoll == 0'");
+							break;
+						}
 						if (pollForMessage() == null) {
 							break;
 						}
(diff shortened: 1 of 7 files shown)


artembilan:
Note: I haven't implemented `Pausable` on the `AbstractPollingEndpoint` to switch the polling logic to the `maxMessagePerPoll = 0`.
But we can consider that, too.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -145,10 +146,23 @@ public void setAdviceChain(List<Advice> adviceChain) {
 		this.adviceChain = adviceChain;
 	}
 
+	/**
+	 * Configure a cap for messages to poll from the source per scheduling cycle.
+	 * The negative number means unlimited amount of messages.
```
garyrussell, line 151:
```suggestion
* A negative number means retrieve unlimited messages until the {@code MessageSource} returns {@code null}.
```

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -145,10 +146,23 @@ public void setAdviceChain(List<Advice> adviceChain) {
...
+	 * Zero means no polling at the moment at all -
+	 * can be treated as pausing if 'maxMessagesPerPoll' is changed eventually.
```
garyrussell, line 153:
```suggestion
	 * Zero means do not poll for any records - it
	 * can be considered as pausing if 'maxMessagesPerPoll' is later changed to a non-zero value.
```

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -327,6 +341,10 @@ private Runnable createPoller() {
 				this.taskExecutor.execute(() -> {
 					int count = 0;
 					while (this.initialized && (this.maxMessagesPerPoll <= 0 || count < this.maxMessagesPerPoll)) {
+						if (this.maxMessagesPerPoll == 0) {
+							logger.info("No poll for 'maxMessagesPerPoll == 0'");
```
garyrussell, line 345:
```suggestion
							logger.info("Polling disabled while 'maxMessagesPerPoll == 0'");
```

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -357,19 +375,28 @@ private Runnable createPoller() {
 												new Date(), null))
 								.flatMapMany(l ->
 										Flux
-												.<Message<?>>generate(fluxSink -> {
-													Message<?> message = pollForMessage();
-													if (message != null) {
-														fluxSink.next(message);
+												.defer(() -> {
+													if (this.maxMessagesPerPoll == 0) {
+														logger.info("No poll for 'maxMessagesPerPoll == 0'");
```
garyrussell, line 380:
```suggestion
														logger.info("Polling disabled while 'maxMessagesPerPoll == 0'");
```

spring-integration-core/src/test/java/org/springframework/integration/config/SourcePollingChannelAdapterFactoryBeanTests.java
```diff
@@ -222,6 +224,47 @@ public void testStartSourceBeforeRunPollingTask() {
...
+			return invocation.callRealMethod();
+		})
+				.given(logger)
+				.info(anyString());
+
+		pollingChannelAdapter.afterPropertiesSet();
+		pollingChannelAdapter.start();
+
+		assertThat(logCalledLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		verify(logger, atLeastOnce()).info("No poll for 'maxMessagesPerPoll == 0'");
```
garyrussell, line 257:
```suggestion
		verify(logger, atLeastOnce()).info("Polling disabled while 'maxMessagesPerPoll == 0'");
```

src/reference/asciidoc/channel-adapter.adoc
```diff
@@ -76,6 +76,9 @@ However, if you are sure that your method can return null and you need to poll f
 <int:poller max-messages-per-poll="-1" fixed-rate="1000"/>
 ----
 ====
+
+Starting with version 5.5, a `0` value for `max-messages-per-poll` has a special meaning - skip `receive()` call altogether, which may be treated as a pausing for this inbound channel adapter.
+The `maxMessagesPerPoll` can be changed to a different value later on, e.g. via Control Bus.
```
garyrussell, line 81:
```suggestion
Starting with version 5.5, a `0` value for `max-messages-per-poll` has a special meaning - skip the `MessageSource.receive()` call altogether, which may be considered as pausing for this inbound channel adapter until the `maxMessagesPerPoll` is changed to a non-zero value at a later time, e.g. via a Control Bus.
```

src/reference/asciidoc/endpoint.adoc
```diff
@@ -141,9 +141,12 @@ consumer.setReceiveTimeout(5000);
 ====
 
 The `maxMessagesPerPoll` property specifies the maximum number of messages to receive within a given poll operation.
-This means that the poller continues calling receive() without waiting, until either `null` is returned or the maximum value is reached.
+This means that the poller continues calling `receive()` without waiting, until either `null` is returned or the maximum value is reached.
 For example, if a poller has a ten-second interval trigger and a `maxMessagesPerPoll` setting of `25`, and it is polling a channel that has 100 messages in its queue, all 100 messages can be retrieved within 40 seconds.
 It grabs 25, waits ten seconds, grabs the next 25, and so on.
+If `maxMessagesPerPoll` is configured for the negative value, then `receive()` is called withing a single polling cycle until it returns `null`.
+Starting with version 5.5, a `0` value has a special meaning - skip `receive()` call altogether, which may be treated as a pausing for this polling endpoint.
+The `maxMessagesPerPoll` can be changed to a different value later on, e.g. via Control Bus.
```
garyrussell, line 149:
```suggestion
If `maxMessagesPerPoll` is configured with a negative value, then `MessageSource.receive()` is called within a single polling cycle until it returns `null`.
Starting with version 5.5, a `0` value has a special meaning - skip the `MessageSource.receive()` call altogether, which may be considered as pausing for this polling endpoint until the `maxMessagesPerPoll` is changed to a n non-zero value at a later time, e.g. via a Control Bus.
```

src/reference/asciidoc/reactive-streams.adoc
```diff
@@ -73,6 +73,8 @@ A polling trigger is built from the provided options and used for periodic sched
 When an `outputChannel` is a `ReactiveStreamsSubscribableChannel`, the same `Trigger` is used to determine the next time for execution, but instead of scheduling tasks, the `SourcePollingChannelAdapter` creates a `Flux<Message<?>>` based on the `Flux.generate()` for the `nextExecutionTime` values and `Mono.delay()` for a duration from the previous step.
 A `Flux.flatMapMany()` is used then to poll `maxMessagesPerPoll` and sink them into an output `Flux`.
 This generator `Flux` is subscribed by the provided `ReactiveStreamsSubscribableChannel` honoring a back-pressure downstream.
+Starting with version 5.5, when `maxMessagesPerPoll == 0`, the source is not called at all, and `flatMapMany()` is completed immediately via `Mono.empty()` result.
+The `maxMessagesPerPoll` can be changed to different value later on, e.g. via Control Bus.
```
garyrussell, line 77:
```suggestion
Starting with version 5.5, when `maxMessagesPerPoll == 0`, the source is not called at all, and `flatMapMany()` is completed immediately via a `Mono.empty()` result until the `maxMessagesPerPoll` is changed to non-zero value at a later time, e.g. via a Control Bus.
```

src/reference/asciidoc/whats-new.adoc
```diff
@@ -25,6 +25,10 @@ The `spring.integration.channels.error.requireSubscribers=true` global property
 The `spring.integration.channels.error.ignoreFailures=true` global property is added to indicate that the global default `errorChannel` must ignore (or not) dispatching errors and pass the message to the next handler.
 See <<./configuration.adoc#global-properties,Global Properties>> for more information.
 
+An `AbstractPollingEndpoint` (source polling channel adapter and polling consumer) treats a `maxMessagesPerPoll == 0` as to skip calling the source.
+It can be changed to different value later on, e.g. via Control Bus.
```
garyrussell, line 29:
```suggestion
An `AbstractPollingEndpoint` (source polling channel adapter and polling consumer) treats `maxMessagesPerPoll == 0` as to skip calling the source.
It can be changed to different value later on, e.g. via a Control Bus.
```

garyrussell (changes requested):
Just doc polishing.

artembilan:
@garyrussell ,

thank you for review and suggestions.
Applied all of them.

What do you think about `Pausable` idea alongside with this feature?

The Migration Guide note is on me after merging this one...

garyrussell (approved):

garyrussell:
I assume you mean `pause()` would save off the mmpp and set it to zero and `resume()` restores the mmpp.

That would make sense; yes.

artembilan:
> I assume you mean pause() would save off the mmpp and set it to zero and resume() restores the mmpp.

Yes, that's an idea. However it might not be so easy with propagation to the `MessageSource`. See `KafkaMessageSource`. Its pausing/resuming logic is really done in the `doReceive()`. Although we might distinguish when do pause directly in the polling endpoint or delegate to its `MessageSource`...

Probably different story. Let's revise it later!

PR #3489, opened 2021-02-05
GH-3488: Fix Persisent Filters with Recursion

### LINKED ISSUE #3488: When Filtering in a recursive operation, include the directory tree in the metadata
Currently we only store the file name.

https://stackoverflow.com/questions/66053141/spring-integration-sftppersistentacceptoncefilelistfilter-filters-out-new-file

In addition, don't filter directories at all when using recursion - other wise a new file added deep in the tree will not be found.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3488

Resolves two problems:

- When changes are made deep in the directory tree, they were not detected because
  the directory is in the metadata store and only passes the filter if a file
  immediately under it is changed, changing the directory's timestamp.

This is solved by subclassing `AbstractDirectoryAwareFileListFilter`, allowing its
`alwaysAcceptDirectories` property to be set.

- Only the filename was used as a metadata key; causing problems if a file with the
  same name appears multiple times in the tree.

This is solved with a new property on `AbstractDirectoryAwareFileListFilter` used by
the gateways to determine whether to filter the raw file names returned by the session
(previous behavior) or the full path relative to the root directory.

**cherry-pick to 5.4.x, 5.3.x**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractDirectoryAwareFileListFilter.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractDirectoryAwareFileListFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,8 @@ package org.springframework.integration.file.filters;
  * This permits, for example, pattern matching on just files when using recursion
  * to examine a directory tree.
  *
+ * @param <F> the file type.
+ *
  * @author Gary Russell
  * @since 5.0
  *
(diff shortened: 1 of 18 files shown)


artembilan (changes requested):
> Conflicting files
spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java


Looks like your local `master` is not actual...

Thanks

garyrussell:
Yes; rebased and forced.

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/338a26d34dc57d0c827882067219f27a9f572247 after some code clean up.

Cherry-picked to `5.4.x` & `5.3.x` as https://github.com/spring-projects/spring-integration/commit/aa7a47f13d0a527c6103583bb8a93151c2ddfa7d after fixing conflicts.

PR #3495, opened 2021-02-16
FileInbound DSL: Add `recursive` for convenience

artembilan:
Related to https://stackoverflow.com/questions/66171881/how-to-read-nested-txt-file-from-spring-integration-file

The `FileInboundChannelAdapterSpec` can be configured with an external `DirectoryScanner`,
but it sometimes becomes burden for end-users to extract a scanner bean and configure it
with all the required file filters

* Expose a `recursive(boolean)` option for better end-user experience
* Rework `FileTests` for JUnit 5
* Mention a new option in the docs

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 the original author or authors.
+ * Copyright 2016-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,13 +22,14 @@ import java.util.Comparator;
 import java.util.Map;
 import java.util.function.Function;
 
-import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.beans.DirectFieldAccessor;
 import org.springframework.integration.dsl.ComponentsRegistration;
 import org.springframework.integration.dsl.MessageSourceSpec;
 import org.springframework.integration.expression.FunctionExpression;
 import org.springframework.integration.file.DirectoryScanner;
 import org.springframework.integration.file.FileLocker;
 import org.springframework.integration.file.FileReadingMessageSource;
+import org.springframework.integration.file.RecursiveDirectoryScanner;
 import org.springframework.integration.file.config.FileListFilterFactoryBean;
 import org.springframework.integration.file.filters.ExpressionFileListFilter;
 import org.springframework.integration.file.filters.FileListFilter;
@@ -76,6 +77,22 @@ public class FileInboundChannelAdapterSpec
 	}
 
 	/**
+	 * A convenient flag to determine if target message source should use a
+	 * {@link RecursiveDirectoryScanner} or stay with a default one.
+	 * @param recursive to set or not a {@link RecursiveDirectoryScanner}.
+	 * @return the spec.
+	 * @see org.springframework.integration.file.RecursiveDirectoryScanner
+	 * @since 5.5
+	 */
+	public FileInboundChannelAdapterSpec recursive(boolean recursive) {
+		if (recursive) {
+			new DirectFieldAccessor(this.target).setPropertyValue("scanner", new RecursiveDirectoryScanner());
+		}
+		return _this();
+	}
+
+
+	/**
 	 * Specify a custom scanner.
 	 * @param scanner the scanner.
 	 * @return the spec.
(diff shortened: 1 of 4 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
```diff
@@ -75,6 +76,22 @@ FileInboundChannelAdapterSpec directory(File directory) {
...
+	 * A convenient flag to determine if target message source should use a
+	 * {@link RecursiveDirectoryScanner} or stay with a default one.
+	 * @param recursive to set or not a {@link RecursiveDirectoryScanner}.
+	 * @return the spec.
+	 * @see org.springframework.integration.file.RecursiveDirectoryScanner
+	 * @since 5.5
+	 */
+	public FileInboundChannelAdapterSpec recursive(boolean recursive) {
+		if (recursive) {
+			new DirectFieldAccessor(this.target).setPropertyValue("scanner", new RecursiveDirectoryScanner());
```
garyrussell, line 89: Why use recursion when target has `setScanner`?

src/reference/asciidoc/whats-new.adoc
```diff
@@ -58,6 +58,8 @@ In addition, `forRecursion=true` causes the full path to files to be used as the
 IMPORTANT: This means that existing keys in a persistent metadata store will not be found for files beneath the top level directory.
 For this reason, the property is `false` by default; this may change in a future release.
 
+The `FileInboundChannelAdapterSpec` has now a convenient `recursive(boolean)` option instead of requesting an explicit reference to the `RecursiveDirectoryScanner`.
```
garyrussell, line 61:
```suggestion
The `FileInboundChannelAdapterSpec` has now a convenient `recursive(boolean)` option instead of requiring an explicit reference to the `RecursiveDirectoryScanner`.
```

src/reference/asciidoc/file.adoc
```diff
@@ -216,6 +215,8 @@ All other sub-directories inclusions and exclusions are based on the target `Fil
 For example, the `SimplePatternFileListFilter` filters out directories by default.
 See https://docs.spring.io/spring-integration/api/org/springframework/integration/file/filters/AbstractDirectoryAwareFileListFilter.html[`AbstractDirectoryAwareFileListFilter`] and its implementations for more information.
 
+NOTE: Starting with version 5.5, the `FileInboundChannelAdapterSpec` of Java DSL has a convenient `recursive(boolean)` option to use the mentioned `RecursiveDirectoryScanner` in the target `FileReadingMessageSource` instead of default one.
```
garyrussell, line 218:
```suggestion
NOTE: Starting with version 5.5, the `FileInboundChannelAdapterSpec` of the Java DSL has a convenient `recursive(boolean)` option to use a `RecursiveDirectoryScanner` in the target `FileReadingMessageSource` instead of the default one.
```

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
```diff
@@ -83,14 +100,14 @@ FileInboundChannelAdapterSpec directory(File directory) {
 	 */
 	public FileInboundChannelAdapterSpec scanner(DirectoryScanner scanner) {
 		this.scanner = scanner;
-		this.target.setScanner(scanner);
+
```
garyrussell, line 103: Why is this removed?

garyrussell (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
artembilan: I believe you mean "reflection".
It is because the `setFilter` and `setLocker` check for the external scanner. See `FileReadingMessageSource.scannerExplicitlySet` property and its usage.
According end-user experience it is a bit cumbersome to move all the fluent filter configuration to the external scanner, when they just want to change the default scanner to recursive one.

I think we can find some other way to do that with some new setter, but I don't think it hurts at the moment to do this reflection once on configuration phase. If you insist, I'll revise it anyway.

garyrussell: No that's fine; yes, I meant reflection 😄

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
garyrussell: How is this resolved? How does an external scanner get applied now?

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
artembilan: OK. Let me push some fix and we will continue discussion if that!

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/FileInboundChannelAdapterSpec.java
artembilan: It is restored in my local copy. 
Will push soon enough!
Sorry for confusion 😄

garyrussell (approved):

PR #3496, opened 2021-02-17
Deprecate Properties as bean for int global props

artembilan:
Having a bean as a `java.util.Properties` is not confusing
and may lead to some conflicts in the real application.
Plus it is not so easy to configure: need to know all
the possible integration properties - bad end-user experience

* Make an `IntegrationProperties` as a public POJO
for easy configuration of global properties
* Deprecate a presence of the `java.util.Properties`
* Leave framework-created `integrationGlobalProperties` as
a `Properties` instance for backward compatibility
* Fix tests to expose an `IntegrationProperties` bean instead
of deprecated `Properties`
* Fix docs according the change and current recommendations

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
@@ -19,10 +19,7 @@ package org.springframework.integration.config;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.util.Arrays;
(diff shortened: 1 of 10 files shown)


artembilan:
Note: this is going to be a foundation for `@ConfigurationProperties` in Spring Boot: the better end-user experience when code auto-completion in the IDE. And, of course, a single source of truth with all the convention-on-configuration in Spring Boot!

The change is going to be mentioned in the Migration Guide.
In the next major version the `Properties`-based registration is going to be removed.

Thanks

artembilan:
The Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.4-to-5.5-Migration-Guide#integrationproperties

PR #3498, opened 2021-02-22
GH-3497: `ReplyProducerCleaner`: check only MPs

### LINKED ISSUE #3497: Proxy object depending backed by a map will fail when passing through ReplyProducerCleaner#requiresDestruction(Object clean)
Hi.

We use proxy beans backed by map of objects for our multi-tenancy solution.

These proxies depend on environment variables being set.

However, since ReplyProducerCleaner#requiresDestruction calls #hashCode on any object, it will force a resolution of the object on the proxy without said environment variable first being set.

For example:

5.3.4.RELEASE

**Describe the bug**

ReplyProducerCleaner#requiresDestruction(Object clean) triggers hashCode of possibly not ready proxy bean.

**To Reproduce**

Have a proxy bean backed by a map of objects that depend on some environment variable.

Since the mentioned method calls the `#hashCode` of the proxy bean and the environment variable is not available at the moment the method is called, an exception is thrown.

**Expected behavior**

One possible solution is so that `contains` should only be called for the expected bean type of `MessageProducer`

**Sample**

https://gist.github.com/crybat/00371f5f7e5e5b521df0a12ad9293706#file-supplierobjectpooldelegate-java

**Mitigation**

Create a `BeanDefinitionRegistryPostProcessor` that replaces `ReplyProducerCleaner#requiresDestruction ` with a version that first checks if the provided bean is of MessageProducer type before calling the set's contains method.

https://gist.github.com/crybat/00371f5f7e5e5b521df0a12ad9293706#file-apostprocessor-java

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3497

The `BaseIntegrationFlowDefinition.REFERENCED_REPLY_PRODUCERS` is filled up
only with the `MessageProducer` instances.
Therefore no reason to calculate a hash code for every single bean passed
to the `ReplyProducerCleaner.requiresDestruction()`

* Check for the `MessageProducer` instance before passing the bean to the
`BaseIntegrationFlowDefinition.REFERENCED_REPLY_PRODUCERS.contains()`
* Check for the `MessageProducer` in the `ReplyProducerCleaner.postProcessBeforeDestruction()`,
too.
It can be called independently of the `requiresDestruction()` and will calculate a hash code
from the bean again for nothing

**Cherry-pick to 5.4.x & 5.3.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2020 the original author or authors.
+ * Copyright 2019-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


garyrussell:
And cherry-picked to 5.4.x, 5.3.x.

PR #3500, opened 2021-03-03
Miscellaneous improvement for end-user experience

artembilan:
Related to https://stackoverflow.com/questions/66428239/ineligiblemethodexception-found-more-than-one-parameter-type-candidate-spring
Related to https://stackoverflow.com/questions/66397350/spring-integration-how-do-i-call-the-spring-integration-from-spring-controller

The exception thrown from gateway method call for `null` arg is not clear that it cannot be `null`
* Improve `GatewayMethodInboundMessageMapper.mapArgumentsToMessage()` exception message to show args
and explain that it cannot be resolved to `null`

When more than one params are candidates for method invocation, the `MessagingMethodInvokerHelper`
throws an `IneligibleMethodException` which outcome is not clear what to do
* Add `@Payload` annotation recommendation into an exception message to make the reason more cleaner
* Some other code style clean up in the `MessagingMethodInvokerHelper`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -620,10 +620,10 @@ public class MessagingMethodInvokerHelper extends AbstractExpressionEvaluator im
 
 		if (++handlerMethod.failedAttempts >= FAILED_ATTEMPTS_THRESHOLD) {
 			handlerMethod.spelOnly = true;
-				LOGGER.info(() -> "Failed to invoke [ " + handlerMethod.invocableHandlerMethod +
-						"] with provided arguments [ " + parameters + " ]. \n" +
-						"Falling back to SpEL invocation for expression [ " +
-						expression.getExpressionString() + " ]");
+			LOGGER.info(() -> "Failed to invoke [ " + handlerMethod.invocableHandlerMethod +
+					"] with provided arguments [ " + parameters + " ]. \n" +
+					"Falling back to SpEL invocation for expression [ " +
+					expression.getExpressionString() + " ]");
 		}
 
 		return invokeExpression(expression, parameters);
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -1332,7 +1328,8 @@ private void setExclusiveTargetParameterType(TypeDescriptor targetParameterType,
 
 			if (this.targetParameterTypeDescriptor != null) {
 				throw new IneligibleMethodException("Found more than one parameter type candidate: [" +
-						this.targetParameterTypeDescriptor + "] and [" + targetParameterType + "]");
+						this.targetParameterTypeDescriptor + "] and [" + targetParameterType + "].\n" +
+						"Consider to mark one of the parameter with '@Payload' annotation.");
```
garyrussell, line 1332:
```suggestion
						"Consider annotating one of the parameters with '@Payload'.");
```

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -620,12 +620,10 @@ private Object fallbackToInvokeExpression(HandlerMethod handlerMethod, Parameter
 
 		if (++handlerMethod.failedAttempts >= FAILED_ATTEMPTS_THRESHOLD) {
 			handlerMethod.spelOnly = true;
-			if (LOGGER.isInfoEnabled()) {
-				LOGGER.info("Failed to invoke [ " + handlerMethod.invocableHandlerMethod +
+				LOGGER.info(() -> "Failed to invoke [ " + handlerMethod.invocableHandlerMethod +
 						"] with provided arguments [ " + parameters + " ]. \n" +
 						"Falling back to SpEL invocation for expression [ " +
 						expression.getExpressionString() + " ]");
```
garyrussell, line 626:
```suggestion
			LOGGER.info(() -> "Failed to invoke [ " + handlerMethod.invocableHandlerMethod +
					"] with provided arguments [ " + parameters + " ]. \n" +
					"Falling back to SpEL invocation for expression [ " +
					expression.getExpressionString() + " ]");
```

garyrussell (changes requested):

garyrussell (approved):

PR #3508, opened 2021-03-11
GH-3507: Fix OSDelegatingFileTailingMessageProducer filename handling issue

### LINKED ISSUE #3507: OSDelegatingFileTailingMessageProducer cannot handle filename having special characters
I noticed that the way `OSDelegatingFileTailingMessageProducer` passing command string to `Runtime.getRuntime().exec` may cause problems if spaces (and other special characters) are used in the filename.
```
this.command = "tail " + this.options + " " + getFile().getAbsolutePath();
```
e.g `foo bar -c 1` is a valid unix filename. The command would be interpreted as `tail <options> /path/foo bar -c 1`. This command `tail`s two different files `/path/foo` and `bar` with option `-c 1`.
Just a caveat, command options could be injected because `tail` command on many linux distros accepts command options to be placed after file path.
I will make a PR for this issue soon.

trgpa:
Fixes #3507 
I tried to make minimal change and kept the `command` variable to avoid breaking current tests.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/tail/OSDelegatingFileTailingMessageProducer.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/tail/OSDelegatingFileTailingMessageProducer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


artembilan (approved):
LGTM so far!

Pulling locally for final review and possible clean up.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f30ede47e2dd662960509a1b398ed0c6e42e9de8 (and cherry-picked to `5.4.x`) after some clean up. 

Back-ported to `5.3.x` & `5.2.x` as https://github.com/spring-projects/spring-integration/commit/d016dd5b6fb80d99df9be720e92b4b585d2f62f4 after some conflict resolutions.

PR #3511, opened 2021-03-12
INT-4116: Introduce FileAggregator

artembilan:
JIRA: https://jira.spring.io/browse/INT-4116

* Implement a `FileSplitter.FileMarker`-based aggregation strategies
and utilize them in a general `FileAggregator` component
* Make `HeaderAttributeCorrelationStrategy.attributeName` as `final`; add `Assert.notEmpty()`
* Fix `AggregatorFactoryBean` and `AggregatorSpec` to parse the provided processor
for possible `CorrelationStrategy` and/or `ReleaseStrategy`
* Introduce short-cut methods into Java & Kotlin DSL for an `aggregate()` configuration
* Introduce a `FileHeaders.LINE_COUNT` for header to be populated in the `FileSplitter`.
We need this info in the `FileAggregator` to avoid possible overhead with JSON deserialization
of the `FileSplitter.FileMarker` messages
* Improve `FileSplitter` doc for code block switch (tabs)

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileAggregatingMessageGroupProcessor.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileAggregatingMessageGroupProcessor.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.file.aggregator;
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.springframework.integration.aggregator.AbstractAggregatingMessageGroupProcessor;
+import org.springframework.integration.file.FileHeaders;
+import org.springframework.integration.store.MessageGroup;
+import org.springframework.messaging.Message;
+
+/**
+ * An {@link AbstractAggregatingMessageGroupProcessor} implementation for file content collecting
+ * previously splitted by the {@link org.springframework.integration.file.splitter.FileSplitter}
+ * with the {@code markers} option turned on.
+ * <p>
+ * If no file markers present in the {@link MessageGroup}, then behavior of this processor is
+ * similar to the {@link org.springframework.integration.aggregator.DefaultAggregatingMessageGroupProcessor}.
+ * <p>
+ * When no file content (only file markers are grouped), this processor emits an empty {@link ArrayList}.
+ * Note: with no file content and markers turned off,
+ * the {@link org.springframework.integration.file.splitter.FileSplitter} doesn't emit any messages
+ * for possible aggregation downstream.
+ *
+ * @author Artem Bilan
+ *
+ * @since 5.5
+ */
+public class FileAggregatingMessageGroupProcessor extends AbstractAggregatingMessageGroupProcessor {
+
+	@Override
+	protected Object aggregatePayloads(MessageGroup group, Map<String, Object> defaultHeaders) {
+		return group.getMessages()
+				.stream()
+				.filter((message) -> !message.getHeaders().containsKey(FileHeaders.MARKER))
+				.map(Message::getPayload)
+				.collect(Collectors.toList());
+	}
+
+}
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileAggregatingMessageGroupProcessor.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileAggregatingMessageGroupProcessor.java
@@ -17,8 +17,9 @@
 package org.springframework.integration.file.aggregator;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 import org.springframework.integration.aggregator.AbstractAggregatingMessageGroupProcessor;
 import org.springframework.integration.file.FileHeaders;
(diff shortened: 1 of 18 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileMarkerReleaseStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+
+		return messages
+				.stream()
+				.filter((message) ->
+						FileSplitter.FileMarker.Mark.END.name()
+								.equals(message.getHeaders().get(FileHeaders.MARKER)))
+				.findAny()
+				.map((message) -> message.getHeaders().get(FileHeaders.LINE_COUNT, Long.class))
+				.map((lineCount) -> lineCount == messages.size() - 2)
+				.orElse(false);
```
garyrussell, line 49: This won't scale well, especially with large files; it would be better to add `getLast()` to message group.

spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileAggregatingMessageGroupProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+public class FileAggregatingMessageGroupProcessor extends AbstractAggregatingMessageGroupProcessor {
+
+	@Override
+	protected Object aggregatePayloads(MessageGroup group, Map<String, Object> defaultHeaders) {
+		return group.getMessages()
+				.stream()
+				.filter((message) -> !message.getHeaders().containsKey(FileHeaders.MARKER))
+				.map(Message::getPayload)
+				.collect(Collectors.toList());
+	}
```
garyrussell, line 54: I thought we decided not to use Stream APIs in main code flows.

garyrussell (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileMarkerReleaseStrategy.java
artembilan: Oh! I see your point. It looks like the `ReleaseStrategy` contract would be better as `canRelease(MessageGroup group, Message<?> currentMessage)` since we always trigger it whenever we got a new message. 
We can deprecate the current contract in favor of new and remove in the next `6.0` 😄 

But `getLast()` API in the `MessageGroup` won't hurt!

spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileAggregatingMessageGroupProcessor.java
artembilan: Yeah... Will rework.
Thanks for the reminder!

spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileMarkerReleaseStrategy.java
artembilan: Well, after looking into this closely, the `getLast()` doesn't feel OK. My test with the `filter()` and `ExecutorChannel` confirms that the last one is not always the `END` marker. So, I would say that `canRelease(MessageGroup group, Message<?> currentMessage)` makes sense for the main code flow, but `forceComplete(MessageGroup)` is still need to iterate the whole group, since there is no way to be sure that the last added to the group is exactly a `FileMarker` message.

At least some optimization is on its way! 😄

garyrussell: Well, if they go async, all bets are off. If the last marker is processed before the size matches, we'll never release the group.

If we want to support that (async), we'd need another api on the group (e.g. `condition1Passed`, set to true if we ever received the last marker).

Also, the file lines would be re-assembled out of order.

artembilan: We may not care about an order of those items, especially when we convert them to some domain objects for further batch inserts into DB. The `applySequence` on the `FileSplitter` and `Resequencer` were always there for us to reassemble an original order.

I need to give the problem you have described more thoughts...

Have a good weekend!

artembilan:
Pushed a fix without Java Streams.

The feature you are asking would probably like this:
1. The `MessageGroup` gets a new property `String condition` (or `groupCondition` if some DB vendors has the `condition` as a key word)
2. The `MessageGroupStore` get a new option - `coditionSupplier(Function<Message, String>)`
3. When we add a mesasge to the group, we calculate a condition against that message and store it into a group entity
4. The `ReleaseStrategy` can then consult this new group property without walking through the whole group.

I suggest to make this condition as a `String` since it is the only way to have any variety of data structure: JSON, SpEL or just number representation. It is then a `ReleaseStrategy` responsibility to parse such a condition properly for its purposes. Even if it is SpEL to parse, it is still faster, then load the whole group from DB.

I would suggest to revise this feature in a separate issue/PR: really the proposed solution in the current PR fully reflects what we have so far and what an original JIRA is asking.

Any feedback welcome!

Will take a look into that a bit later: I need to fix failing JMS tests and check `TaskScheduler` for `errorChannel`.
Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileMarkerReleaseStrategy.java
```diff
@@ -0,0 +1,63 @@
...
+ * @since 5.5
+ */
+public class FileMarkerReleaseStrategy implements ReleaseStrategy {
+
+	@Override
+	public boolean canRelease(MessageGroup group) {
+		int size = group.size();
+		if (size > 1) { // Need more than only a START marker
+			Collection<Message<?>> messages = group.getMessages();
+			for (Message<?> message : messages) {
```
garyrussell, line 44: This is still no good; we can't iterate over the whole collection each time, it just won't scale with large groups.

garyrussell (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileMarkerReleaseStrategy.java
artembilan: Correct, but see what we have so far in other place - `SequenceAwareMessageGroup`.
Not related to this one, but similar iteration on each added message.

Nevertheless I agree with your concern and really will address i, but I'd like to do that in the separate PR.
See my yesterday's comment.

garyrussell: OK; but let's open an issue so it's not forgotten.

garyrussell (approved):

spring-integration-file/src/main/java/org/springframework/integration/file/aggregator/FileMarkerReleaseStrategy.java
artembilan: If you agree with my proposal, I'll start working on that immediately, so this `FileAggregator` won't be a bottle neck any more.

garyrussell: Yes; proposal looks good.

PR #3513, opened 2021-03-15
GH-3512: Update PostgreSQL priority index to match code implementation

### LINKED ISSUE #3512: PostgreSQL (and possibly other DB) schema definitions have a slightly wrong Priority unique index creation

5.3.6.RELEASE (can be observed in current version source code, as well)

**Describe the bug**

This may potentially affect other database types, as well, but this has only been tested and verified as a defect on PostgreSQL.

The SQL script in spring-integration-jdbc for creating the PostgreSQL schema contains a creation of a unique index if utilizing a `priorityEnabled` queue via a JDBC-backed message store. This index is shown below:

```
-- This is only needed if the message group store property 'priorityEnabled' is true
-- CREATE UNIQUE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (REGION, GROUP_KEY, MESSAGE_PRIORITY DESC, CREATED_DATE, MESSAGE_SEQUENCE);
```

And here is an example of a query that gets generated to poll for new messages from the queue:

```
SELECT int_channel_message.MESSAGE_ID, int_channel_message.MESSAGE_BYTES
FROM int_channel_message
WHERE int_channel_message.GROUP_KEY = :groupKey
  AND int_channel_message.REGION = :region
ORDER BY MESSAGE_PRIORITY DESC NULLS LAST, CREATED_DATE, MESSAGE_SEQUENCE
LIMIT 1 FOR UPDATE;
```

The problem here is that the poller query does not hit the unique index. The reason why is that the index is missing the `NULLS LAST` portion on the `MESSAGE_PRIORITY DESC` definition. Testing this locally on a queue with roughly 750k records of random data, the poller select statement takes roughly 2.5 seconds with the current index definition, and roughly 50 milliseconds after adding the `NULLS LAST` definition.

**Sample**

I unfortunately cannot submit a Pull Request right this minute, so here is a simple patch of the `schema-postgresql.sql` file below. If no one else is able to get to this within the next couple of days, I should have the time to get the project pulled down and get a proper contribution PR up here.

```
avester:~$ diff -p schema-postgresql.sql schema-postgresql-fixed-index.sql
*** schema-postgresql.sql	2021-03-12 17:46:58.815964348 -0600
--- schema-postgresql-fixed-index.sql	2021-03-12 17:47:22.435634860 -0600
*************** CREATE TABLE INT_CHANNEL_MESSAGE (
*** 49,55 ****
  
  CREATE INDEX INT_CHANNEL_MSG_DELETE_IDX ON INT_CHANNEL_MESSAGE (REGION, GROUP_KEY, MESSAGE_ID);
  -- This is only needed if the message group store property 'priorityEnabled' is true
! -- CREATE UNIQUE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (REGION, GROUP_KEY, MESSAGE_PRIORITY DESC, CREATED_DATE, MESSAGE_SEQUENCE);
  
  
  CREATE TABLE INT_METADATA_STORE  (
--- 49,55 ----
  
  CREATE INDEX INT_CHANNEL_MSG_DELETE_IDX ON INT_CHANNEL_MESSAGE (REGION, GROUP_KEY, MESSAGE_ID);
  -- This is only needed if the message group store property 'priorityEnabled' is true
! -- CREATE UNIQUE INDEX INT_CHANNEL_MSG_PRIORITY_IDX ON INT_CHANNEL_MESSAGE (REGION, GROUP_KEY, MESSAGE_PRIORITY DESC NULLS LAST, CREATED_DATE, MESSAGE_SEQUENCE);
  
  
  CREATE TABLE INT_METADATA_STORE  (
```

Again it's entirely possible, and likely, that this affects other databases as well, but I would not submit a bug report with assumptions, rather only the factual data that I have, and that's only for PostgreSQL at the moment.

If any more information is needed, I'm happy to do what I can to assist. Thank you!

runt9:
Fixes https://github.com/spring-projects/spring-integration/issues/3512

Updates SQL schema creation for PostgreSQL to add `NULLS LAST` to the priority index to match what the source code is doing as it currently does not use the index.


artembilan (changes requested):
This is what I see in the source code:

image

Doesn't it sound like we need to fix the rest of indexes for all those matched vendors?

Thanks

artembilan (approved):
Found the answer: https://use-the-index-luke.com/sql/sorting-grouping/order-by-asc-desc-nulls-last

> Only the PostgreSQL database (since release 8.3) supports the NULLS modifier in both the order by clause and the index definition.

So, all good, merging...

artembilan:
Merged with commit message improvement.

PR #3510, opened 2021-03-12
GH-3509: Register TcpSenders on wrapped connection

gigermocas:
Fixes https://github.com/spring-projects/spring-integration/issues/3509

Dead connections are not being removed on `TcpSender` when using a `TcpConnectionInterceptor`
May cause a memory leak

Issue #3509, opened 2021-03-11
Memory leak through TcpSendingMessageHandler's connections map


gigermocas:
`TcpSendingMessageHandler` connections map is not being cleared properly when using a `TcpConnectionInterceptor` and is causing a memory leak

From what I could ascertain, `TcpConnectionSupport.close()` doesn't see the tcp senders as they were registered on `TcpConnectionInterceptor`
A workaround involves overriding `registerSenders()` in the custom interceptor implementation:
```
@Override
public void registerSenders(List<TcpSender> sendersToRegister) {
	TcpConnectionSupport theConnection = getTheConnection();
	while (theConnection instanceof TcpConnectionInterceptorSupport) {
		theConnection = ((TcpConnectionInterceptorSupport) theConnection).getTheConnection();
	}
	theConnection.registerSenders(sendersToRegister);
}
```

I've created a sample project to replicate the issue:
[spring-integration-ip-leak-test](https://github.com/gigermocas/spring-integration-ip-leak-test)

Analysing the heap dump in Eclipse Memory Analyzer showed the following report:
```
One instance of "org.springframework.integration.ip.tcp.TcpSendingMessageHandler" loaded by "jdk.internal.loader.ClassLoaders$AppClassLoader @ 0x68c79c400" occupies 1,506,280 (17.24%) bytes. The memory is accumulated in one instance of "java.util.concurrent.ConcurrentHashMap$Node[]", loaded by "<system class loader>", which occupies 1,505,920 (17.24%) bytes.

Keywords
org.springframework.integration.ip.tcp.TcpSendingMessageHandler
jdk.internal.loader.ClassLoaders$AppClassLoader @ 0x68c79c400
java.util.concurrent.ConcurrentHashMap$Node[]
```
spring-integration version used is 5.4.4

artembilan:
So, sounds like we have to fix it on that `TcpConnectionInterceptorSupport`.
We do have over there this one:
```
@Override
	public void registerSender(TcpSender sender) {
		this.tcpSender = sender;
		this.theConnection.registerSender(this);
	}
```

But looks like have missed to override newly introduced `registerSenders()`.

Thank you for the report and we would appreciate if you issue a Pull Request with the fix and some tests to cover the problem!

artembilan:
Reopen since we don't have a proper solution yet.

PR #3514, opened 2021-03-16
Register TcpSenders on delegated TcpConnection with interceptor instance

gigermocas:
Follow up to [#3510](https://github.com/spring-projects/spring-integration/pull/3510)

Regrettably my previous pull request didn't address the whole underlying issue, although the memory leak was fixed, it introduced a regression as messages to the client were not being sent through the configured interceptor.

- Introduce overloaded `TcpConnectionSupport.registerSenders()` with a `TcpConnection` instance
- Fix `TcpConnectionInterceptorSupport.registerSenders()` to use introduced `registerSenders()`  on delegated connection


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
```diff
@@ -101,7 +101,7 @@ public void registerSender(TcpSender sender) {
 
 	@Override
 	public void registerSenders(List<TcpSender> sendersToRegister) {
-		this.theConnection.registerSenders(sendersToRegister);
+		this.theConnection.registerSenders(sendersToRegister, this);
```
artembilan, line 104: It is indeed confusing that interceptor abstraction implements a connection one.
So, we have such a strange code flow...

I wish @garyrussell will come back to us one day to revise this architecture 😄 

Nevertheless would you mind, @gigermocas , to double check what we have so far with the `TcpConnectionSupport.close()`:
```
public void close() {
		for (TcpSender sender : this.senders) {
			sender.removeDeadConnection(this);
		}
...
```
After this fix it doesn't look like we are going to clean up properly...

I'll debug it locally meanwhile, too

artembilan (changes requested):

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
garyrussell: >I wish @garyrussell will come back to us one day to revise this architecture 😄

Yes, this is a bit of a mess; dates back to very early code and has caused headaches since then.

Maybe in 6.0 @artembilan if you want to raise an issue.

artembilan: Issue created: https://github.com/spring-projects/spring-integration/issues/3515, but it doesn't mean we have to make it in `6.0`. Probably one day in the future when it causes more headaches 😉

artembilan:
Since you are already here, @garyrussell , WDYT about the fix?
Thanks

garyrussell:
I am not really "here" 😄  I was just responding to your mention. But I will take a look when a get a few minutes.

garyrussell (changes requested):
Unfortunately, we have to use the same object for both `addNewConnection()` and `removeDeadConnection()`.

The key to the connections maps is the connection id. While this is the same for the `HelloWorldInterceptor`, that is not the case for other framework interceptors.  

e.g. https://github.com/spring-projects/spring-integration/blob/be2a69861880be48564bd4eaf8f257fc06d24947/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactory.java#L388-L391

If I change the `HelloWorldInterceptor` thus:

```java
	@Override
	public String getConnectionId() {
		return "somethingElse";
	}
```

and add this to the new test...

```java
assertThat(TestUtils.getPropertyValue(handler, "connections", Map.class)).isEmpty();
```

...it fails.

gigermocas:
I see your point @garyrussell, but the only framework class that implements `TcpConnectionInterceptorSupport` is `CachingClientConnectionFactory.CachedConnection` (!) and it's just wrapping an existing connection, so that should be fine.

I guess I've only managed to reduce the problem scope to those users that override their interceptor's `getConnectionId()` which is a narrower use case!? The behavior for those would be the same as before the patches for #3509.

I'm not seeing a full fix for this without requiring some major rework, sorry for the mess guys :disappointed:

artembilan:
I can't find any solution for this at the moment.
So, or we just merge this as is relying on the fact that end-users are not going to override `getConnectionId()`.
Or just let it be for time being. At least until the next release in April...

garyrussell:
We could add this hack to the 3 implementations of TcpSender, at least for today's release, to give us some time for a better solution.

```java
	private static final Pattern FAILOVER_ID = Pattern.compile("-[a-f0-9]{12}:[0-9]+$");

	private final Map<String, String> failoverConnections = new ConcurrentHashMap<>();

        ...

	@Override
	public void addNewConnection(TcpConnection connection) {
		String id = connection.getConnectionId();
		if (FAILOVER_ID.matcher(id).find()) {
			this.failoverConnections.put(id.substring(0, id.lastIndexOf(':')), id);
		}
		this.connections.put(id, connection);
	}

	@Override
	public void removeDeadConnection(TcpConnection connection) {
		String id = connection.getConnectionId();
		this.connections.remove(id);
		String mapped = this.failoverConnections.remove(id);
		if (mapped != null) {
			this.connections.remove(mapped);
		}
	}
```

I tested it by adding this to the Hello World interceptor

```java
	@Override
	public String getConnectionId() {
		return super.getConnectionId() + ":1";
	}
```

garyrussell:
Since the original PR was back-ported, we can't leave the regression in place.

artembilan:
I thought about something like that, but based on the `TcpConnectionInterceptorSupport` conditon. Looks like a `FailoverTcpConnection` does not implement that one, so my idea does not cover everything.

Gary, if you see some intermediate solution, please, go ahead for today's release.

Thanks

artembilan:
I mean it is OK to take this PR and commit some fix with back-port.

garyrussell:
Either that or we have to back out the memory leak commit (at least in 5.4.x), given that we are pressed for time.

garyrussell:
Or, defer 5.4.5 - we can live with the regression on master.

artembilan:
OK. Reverting the change in the `5.4.x` and leaving this PR opened for better times.
It's better do not deffer since we have other fixes over there.

garyrussell:
I think I might have a better fix for the original memory leak - testing now.

garyrussell:
Replaced by https://github.com/spring-projects/spring-integration/pull/3518

PR #3516, opened 2021-03-16
Add Publish Unit Test Results to GH Actions

artembilan:
* Upgrade dependencies
* Fix repos to not use superset to fail for
non-authorized artifacts

### DIFF
--- a/.github/workflows/pr-build-workflow.yml
+++ b/.github/workflows/pr-build-workflow.yml
@@ -44,3 +44,8 @@ jobs:
           NO_REFERENCE_TASK: true
         with:
           arguments: checkAsciidocLinks check
+
+      - name: Publish Unit Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v1
+        with:
+          files: '**/build/test-results/**/*.xml'
--- a/.github/workflows/pr-build-workflow.yml
+++ b/.github/workflows/pr-build-workflow.yml
@@ -46,7 +46,7 @@ jobs:
           arguments: checkAsciidocLinks check
 
       - name: Capture Test Results
-        if: failure()
+        if: always()
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-build-workflow.yml
```diff
@@ -44,3 +44,8 @@ jobs:
           NO_REFERENCE_TASK: true
         with:
           arguments: checkAsciidocLinks check
+
+      - name: Publish Unit Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v1
+        with:
+          files: '**/build/test-results/**/*.xml'
```
garyrussell, line 51: I think capturing the gradle HTML test result rendering is much better than trying to dig through all the XML files in test-results.

garyrussell (changes requested):

.github/workflows/pr-build-workflow.yml
artembilan: I'm trying a different action: https://github.com/marketplace/actions/publish-unit-test-results

garyrussell: But I am not averse to capturing both.

.github/workflows/pr-build-workflow.yml
```diff
@@ -44,3 +44,11 @@ jobs:
...
         with:
           arguments: checkAsciidocLinks check
+
+      - name: Capture Test Results
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-results
+          path: '*/build/reports/tests/**/*.*'
+          retention-days: 1
```
garyrussell, line 54: I chose 3 days for the other projects so that the artifacts would still be there Monday for PR submitted late Friday.

garyrussell (changes requested):

PR #3517, opened 2021-03-16
Introduce a `MessageGroup.condition`

artembilan:
* Add a `MessageGroup.condition` option
* Add a `MessageGroupStore.conditionSupplier` option
* Use it from the `SimpleMessageStore.addMessagesToGroup()` API
to populate `condition` (if any) into a `MessageGroup`
* Introduce a `GroupConditionProvider` contract to be implemented
on those `ReleaseStrategy` contracts which could be aware of group condition
* Populated a `GroupConditionProvider.getGroupConditionSupplier()`
into a `MessageGroupStore` from the `AbstractCorrelatingMessageHandler`
for end-user convenience
* Rework a `FileMarkerReleaseStrategy` to implement a `GroupConditionProvider`
to provide a function which produces a condition from `file_lineCount` header
of the `END` marker message
* Make the `FileMarkerReleaseStrategy` logic already based on the condition from a group
* Delegate `GroupConditionProvider` from the `FileAggregator`
* Add test for empty file aggregation


artembilan:
Some PoC for `condition` in the `MessageGroup` and its usage from the `ReleaseStrategy`.

TODO: condition property implementation for every single `MessageGroupStore` impl...

artembilan:
I think all the work is done with source code.
Only docs are left: I can add them to this PR or we can deffer them into the next milestone.
Or we even may not include this feature to the current milestone as well.

@garyrussell , let me know what works best for you to review this and have it merged one or another way.

garyrussell:
I suggest at this late hour that we defer to the next milestone.

artembilan:
> I suggest at this late hour that we defer to the next milestone.

Too late, I have just added docs, so it is ready for the final review 😉 

But yeah, we definitely can push it back if it doesn't work for you at the moment.

Thanks

garyrussell:
It's too big to review; don't you want to get the builds under way ASAP?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -407,6 +407,11 @@ protected void onInit() {
 		 */
 		this.lockRegistrySet = true;
 		this.forceReleaseProcessor = createGroupTimeoutProcessor();
+
+		if (this.releaseStrategy instanceof GroupConditionProvider) {
+			this.messageStore.setConditionSupplier(
+					((GroupConditionProvider) this.releaseStrategy).getGroupConditionSupplier());
+		}
```
garyrussell, line 414: I don't think this will work - while we recommend a different store for each aggregator, we don't enforce it.

artembilan:
Sure! No problem! Reviewing your PR for TCP/IP problem and that's it.

garyrussell (commented):
Need an answer to my comment before reviewing further.

artembilan:
Moved a `conditionSupplier` from the `MGS` to the `AbstractCorrelatingMessageHandler`, so we don't suffer on the `MGS` any more mutating other groups accidentally.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
```diff
@@ -294,6 +295,15 @@ public void removeMessageGroup(Object groupId) {
 		}
 	}
 
+	@Override
+	public void setGroupCondition(Object groupId, String condition) {
+		MessageGroupMetadata metadata = getGroupMetadata(groupId);
+		if (metadata != null) {
```
garyrussell, line 301: Perhaps we need an assertion here, to ensure `setGroupCondition` is only called after the current record has been added to the group (and metadata will be present).

spring-integration-core/src/main/resources/org/springframework/integration/config/spring-integration.xsd
```diff
@@ -4071,6 +4071,19 @@
...
 				</xsd:attribute>
+				<xsd:attribute name="group-condition-supplier" type="xsd:string">
+					<xsd:annotation>
+						<xsd:appinfo>
+							<tool:annotation kind="ref">
+								<tool:expected-type type="java.util.function.BiFunction"/>
+							</tool:annotation>
+						</xsd:appinfo>
+						<xsd:documentation>
+							A reference to an 'java.util.function.BiFunction' bean that may provide a group
```
garyrussell, line 4082:
```suggestion
							A reference to a 'java.util.function.BiFunction' bean that may provide a group
```

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-db2.sql
```diff
@@ -18,7 +18,7 @@ CREATE TABLE INT_GROUP_TO_MESSAGE  (
 CREATE TABLE INT_MESSAGE_GROUP  (
 	GROUP_KEY CHAR(36) NOT NULL,
 	REGION VARCHAR(100) NOT NULL,
-	MARKED BIGINT,
+    CONDITION VARCHAR(255),
```
garyrussell, line 21: We'll need a migration guide entry for the schema change.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MessageDocument.java
```diff
@@ -151,6 +154,15 @@ public Object getGroupId() {
...
 
+	@Nullable
+	public String getCondition() {
+		return this.condition;
+	}
+
+	public void setCondition(String condition) {
+		this.condition = condition;
+	}
+
```
garyrussell, line 165: Are there any backwards compatibility issues (existing documents)?

src/reference/asciidoc/message-store.adoc
```diff
@@ -147,4 +147,21 @@ However starting with version 5.5, all the persistent `MessageGroupStore` implem
...
 A returned `Stream<Message<?>>` must be closed in the end of processing, e.g. via auto-close by the `try-with-resources`.
-Whenever a `PersistentMessageGroup` is used, its `streamMessages()` delegates to the `MessageGroupStore.streamMessagesForGroup()`.
\ No newline at end of file
+Whenever a `PersistentMessageGroup` is used, its `streamMessages()` delegates to the `MessageGroupStore.streamMessagesForGroup()`.
+
+[[message-group-condition]]
+==== Message Group Condition
+
+Starting with version 5.5, the `MessageGroup` abstraction provides a `condition` string option.
+The value of this option can be anything what could be parsed later on for any reason to make a decision for the group.
```
garyrussell, line 156:
```suggestion
The value of this option can be anything that could be parsed later on for any reason to make a decision for the group.
```

src/reference/asciidoc/message-store.adoc
```diff
@@ -147,4 +147,21 @@ However starting with version 5.5, all the persistent `MessageGroupStore` implem
...
+For example a `ReleaseStrategy` from a <<./aggregator.adoc#aggregator-api, correlation message handler>> may consult this property from the group instead of iterating all the messages in the group.
+The `MessageGroupStore` expose a `setGroupCondition(Object groupId, String condition)` API.
```
garyrussell, line 158:
```suggestion
The `MessageGroupStore` exposes a `setGroupCondition(Object groupId, String condition)` API.
```

src/reference/asciidoc/message-store.adoc
```diff
@@ -147,4 +147,21 @@ However starting with version 5.5, all the persistent `MessageGroupStore` implem
...
+For this purpose a `setGroupConditionSupplier(BiFunction<Message<?>, String, String>)` option has been added to the `AbstractCorrelatingMessageHandler`.
+This function is evaluated against each message to be added to the group and existing condition on the group.
```
garyrussell, line 160:
```suggestion
This function is evaluated against each message after it has been added to the group as well as the existing condition of the group.
```

src/reference/asciidoc/message-store.adoc
```diff
@@ -147,4 +147,21 @@ However starting with version 5.5, all the persistent `MessageGroupStore` implem
...
+The implementation may decide to return new value, existing or reset the target condition to `null`.
```
garyrussell, line 161:
```suggestion
The implementation may decide to return a new value, the existing value, or reset the target condition to `null`.
```

src/reference/asciidoc/message-store.adoc
```diff
@@ -147,4 +147,21 @@ However starting with version 5.5, all the persistent `MessageGroupStore` implem
...
+The value for a `condition` can be a JSON, SpEL expression, number or anything what can be serialized as a string and parsed afterwards.
+For example, the `FileMarkerReleaseStrategy` from the <<./file.adoc#file-aggregator, File Aggregator>> component, populates a condition into a group from the `FileHeaders.LINE_COUNT` header of the `FileSplitter.FileMarker.Mark.END` message and consults with it from its `canRelease()` comparing a group size with the value in this condition.
+This way it doesn't iterate all the messages in group to find a `FileSplitter.FileMarker.Mark.END` message with the `FileHeaders.LINE_COUNT` header.
```
garyrussell, line 164: It also allows the end marker to arrive at the aggregator before all the other records; for example when processing a file in a multi-threaded environment.

src/reference/asciidoc/message-store.adoc
```diff
@@ -147,4 +147,21 @@ However starting with version 5.5, all the persistent `MessageGroupStore` implem
...
+
+In addition, for target configuration convenience, a `GroupConditionProvider` contract has been introduced.
+The `AbstractCorrelatingMessageHandler` checks provided `ReleaseStrategy` for this contract implementation and extracts a `conditionSupplier` for further group condition evaluation logic.
```
garyrussell, line 167:
```suggestion
In addition, for configuration convenience, a `GroupConditionProvider` contract has been introduced.
The `AbstractCorrelatingMessageHandler` checks if the provided `ReleaseStrategy` implements this interface and extracts a `conditionSupplier` for group condition evaluation logic.
```

src/reference/asciidoc/whats-new.adoc
```diff
@@ -41,8 +41,12 @@ This is covered as a `ConsumerEndpointSpec.reactive()` option in Java DSL and as
 See <<./reactive-streams.adoc#reactive-streams,Reactive Streams Support>> for more information.
 
 The `groupTimeoutExpression` for a correlation message handler (an `Aggregator` and `Resequencer`) can now be evaluated to a `java.util.Date` for some fine-grained scheduling use-cases.
+Also  `BiFunction groupConditionSupplier` option is added to the `AbstractCorrelatingMessageHandler` to supply a `MessageGroup` condition against a message to be added to the group.
```
garyrussell, line 44: Also the

garyrussell (changes requested):
Looks good.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
artembilan: I don't think it is relevant. It is only the fact of an aggregator that we add a record then calculate a condition.
In other cases it might be set independently and unconditionally.

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-db2.sql
artembilan: Sure! Will address it after merge

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MessageDocument.java
artembilan: Existing documents comes without a `condition` field: nothing to read - nothing to map.
It is really a trick of all of these NO-SQL DBs: no schema - can change data structure any time!

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
garyrussell: ok

garyrussell (approved):

artembilan:
See also a Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.4-to-5.5-Migration-Guide#messagegroupconditon-option

PR #3518, opened 2021-03-17
GH-3509: Fix Memory Leak with Intercepted TCP Conn

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3509

When a connection was intercepted, the `TcpSender.add...` was called with the
interceptor, but the `removeDead...` was called with the actual connection,
causing a memory leak.

Always call the `TcpSender` with the outer-most interceptor.

**Cherry-pick to master after reverting previous fix**

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/TcpSendingMessageHandlerTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/TcpSendingMessageHandlerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,7 @@ import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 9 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/TcpSendingMessageHandlerTests.java
```diff
@@ -1191,4 +1200,67 @@ public void testConnectionException() throws Exception {
...
+		final AtomicReference<TcpConnection> connection = new AtomicReference<>();
+		scf.setApplicationEventPublisher(event -> {
+			if (event instanceof TcpConnectionOpenEvent) {
+				connection.set(handler.getConnections()
+						.get(((TcpConnectionOpenEvent) event).getConnectionId()));
+				latch.countDown();
+			}
+		});
+		TcpConnectionInterceptorFactoryChain fc = new TcpConnectionInterceptorFactoryChain();
+		fc.setInterceptor(newInterceptorFactory(scf.getApplicationEventPublisher()));
```
artembilan, line 1224: Does it work now if we override a `getConnectionId()` in the `HelloWorldInterceptor`?

artembilan (changes requested):

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/TcpSendingMessageHandlerTests.java
garyrussell: Yes, because the same object (the outermost interceptor) is always used for `addNewConnection` an `removeDeadConnection`.

PR #3524, opened 2021-03-23
GH-3523: TcpConnectionEvent Fixes

### LINKED ISSUE #3523: TcpNetServerConnectionFactory race condition between open notification and reads

This is present in `org.springframework.boot` `2.4.1` and `2.4.4` - which I think maps to  spring-integration `5.4.5`. 

**Describe the bug**

`TcpNetServerConnectionFactory` line number 185 to 187: A notification of a new connection is sent asynchronously to reads from the connection. This means that there are no guarantees that the user will be notified of the connection before receiving messages from the connection.

Specifically, this means that I can not create the necessary objects to handle connection reads before receiving reads from the connection. 

I don't know if the same is the case for close events (though this is not an issue for my application).

Interestingly, `TcpNioServerConnectionFactory` doesn't seem to have the same issue (though I haven't been able to follow its logic closely enough to be sure). 

**To Reproduce**

Write something like this:

```
connectionFactory = new TcpNetServerConnectionFactory(HARDCODED_TCP_PORT);

connectionFactory.registerListener((bar) -> System.out.println("READ"));
connectionFactory.setApplicationEventPublisher((foo) -> System.out.println(foo.getClass().getSimpleName()));
```

And then connect to the above server. Sometimes an open event will be printed first and sometimes "READ".

**Expected behavior**

I would expect the open event to be sent and handled first before any reads are done.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3523

- `TcpNetConnection` - publish open event on reader thread to avoid race with first read.
- Intercepted connections - Ensure that the event source is always the outermost interceptor.
- Also reduce delays during NIO client connect.

**cherry-pick to 5.4.x**

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractClientConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractClientConnectionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -158,7 +158,6 @@ public abstract class AbstractClientConnectionFactory extends AbstractConnection
 					return connection;
 				}
 			}
-
 			return doObtain(singleUse);
 		}
 		catch (RuntimeException e) {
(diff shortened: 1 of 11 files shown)


artembilan:
Cherry-picked to `5.4.x` as https://github.com/spring-projects/spring-integration/commit/2d443f9143cc4b129719a09a520c06ed63c76da7 after fixing conflicts.

artembilan:
@garyrussell , do we really need to back-port this to `5.3.x`? Looks like it is not so easy to back-port commits for GH-3509: we have introduced `TcpConnectionSupport.senders` starting with `5.4`...

Otherwise, place, give it a shot and try to back-port yourself.

Thanks for understanding!

garyrussell:
Good point - removed label.

PR #3525, opened 2021-03-23
GH-3521: Delayer: schedule release task with TX

### LINKED ISSUE #3521: Racing condition in Delayer Channel
5.2.1.RELEASE

**Describe the bug**

We have implemented a delayer channel backed by a persistence message store in our application .

Sometimes the spring integration flow cannot send a message to upstream , this happens due to a throttling limit in the application. In this scenario we send that message to the delayer channel with a delay of `30s`.

So that this can be again retried after `30s` to be sent to the upstream application.

Sometimes we observed that the messages are stuck in int_messages table and they are never getting retried. We don't find any exception or error in the logs.

**As per analysis we found that the issue is happening due to a issue in `DelayHandler` class**

We found that message is inserted in the table `int_message` through the line
`this.messageStore.addMessageToGroup(this.messageGroupId, delayedMessage)`;

After next few lines of code `getTaskScheduler()` is used to schedule the message . When the schedulers `releaseTask` logic is called after delay of `30` seconds , the data in `int_message` is still not committed
so `Message<?> messageToRelease = getMessageById(messageId);` returns `null` and the messages gets stuck in db.

Basically the data in `int_message` will get committed only when this request will get completed. But some times the schedule `releaseTask` method gets invoked but the commit has still not completed and we loose that message. So a racing condition is happening in this code.

I have also discussed issue with Artem Billan on this [issue](https://gitter.im/spring-projects/spring-integration?at=604cd849e562cf54acbb2bd0) and raising this issue as discussed with him

The issue discussed is happening because the code in `DelayHandler` is not able to commit the data immediately in Spring Integration tables and the new `Scheduler` thread is launched before it. So this scenario should be handled.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3521

There is a race condition when transactional `MessageStore` is used
for `DelayHandler`, so the message is not visible for reads until after TX is
committed, but a scheduled release task may be already ready after delay

* Register a `TransactionSynchronization` with scheduling a releasing task
when TX is committed

### DIFF
--- a/src/reference/asciidoc/delayer.adoc
+++ b/src/reference/asciidoc/delayer.adoc
@@ -195,8 +195,9 @@ controlBusChannel.send(delayerReschedulingMessage);
 
 NOTE: For more information regarding the message store, JMX, and the control bus, see <<./system-management.adoc#system-management-chapter,System Management>>.
 
-Starting with version 5.3.7, if transaction is active when message is stored into a `MessageStore`, the release task is scheduled in a `TransactionSynchronization.afterCommit()` callback.
-This is necessary to ensure that there is no a race condition when message is not stored because no transaction commit yet, but scheduled task is ready to be performed after delay.
+Starting with version 5.3.7, if a transaction is active when a message is stored into a `MessageStore`, the release task is scheduled in a `TransactionSynchronization.afterCommit()` callback.
+This is necessary to prevent a race condition, where the scheduled release could run before the transaction has committed, and the message is not found.
+In this case, the message will be released after the delay, or after the transaction commits, whichever is later.
 
 [[delayer-release-failures]]
 ==== Release Failures
--- a/src/reference/asciidoc/delayer.adoc
+++ b/src/reference/asciidoc/delayer.adoc
@@ -25,39 +25,10 @@ The following example delays all messages by three seconds:
 
 If you need to determine the delay for each message, you can also provide the SpEL expression by using the 'expression' attribute, as the following expression shows:
 
-====
-[source,xml]
-----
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/delayer.adoc
```diff
@@ -182,12 +189,15 @@ These operations can be invoked through a `Control Bus` command, as the followin
...
     MessageBuilder.withPayload("@'delayer.handler'.reschedulePersistedMessages()").build();
-    controlBusChannel.send(delayerReschedulingMessage);
+controlBusChannel.send(delayerReschedulingMessage);
 ----
 ====
 
 NOTE: For more information regarding the message store, JMX, and the control bus, see <<./system-management.adoc#system-management-chapter,System Management>>.
 
+Starting with version 5.3.7, if transaction is active when message is stored into a `MessageStore`, the release task is scheduled in a `TransactionSynchronization.afterCommit()` callback.
+This is necessary to ensure that there is no a race condition when message is not stored because no transaction commit yet, but scheduled task is ready to be performed after delay.
```
garyrussell, line 199:
```suggestion
Starting with version 5.3.7, if a transaction is active when a message is stored into a `MessageStore`, the release task is scheduled in a `TransactionSynchronization.afterCommit()` callback.
This is necessary to prevent a race condition, where the scheduled release could run before the transaction has committed, and the message is not found.
In this case, the message will be released after the delay, or after the transaction commits, whichever is later.
```

garyrussell (changes requested):

garyrussell:
Cherry-picked as c5e24fb301 and a7843af045

PR #3528, opened 2021-03-25
GH-3526: Fix Infinite Loop in FailoverCConnFactory

### LINKED ISSUE #3526: FailoverClientConnectionFactory.FailoverTcpConnection.findAConnection() can get caught in infinite loop
say we have 2 connection factories s1 and s2, initially we connect ok to s1

however a subsequent send results in an error on s1 (eg disconnection), findAConnection is called to establish a new connection

1/ we have lastFactoryToTry = s1, we enter the loop and try nextfactory = s2 but connection to this fails and throws exception.  in the exception handling lastFactoryToTry does not equal nextfactory so we reset iterator and loop round.

2/ nextfactory now =s1 connection fails and throws exception.  in the exception handling !this.factoryIterator.hasNext() is not true so we loop round. back to 1/

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3526

`FailoverClientConnectionFactory`

The logic to detect we had iterated over all factories and including
the one from which the previous connection was established was incorrect,
causing an infite loop until one of the factory connections was successful.

Change the logic to detect we have reset the iterator and the current failure
is from the same factory as the one from which the previous connection was
established.

**cherry-pick to 5.4.x, 5.3.x, 5.2.x**

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
@@ -39,7 +39,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -47,8 +46,6 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import javax.net.ServerSocketFactory;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.junit.jupiter.api.Test;
 import org.mockito.InOrder;
 import org.mockito.Mockito;
@@ -64,7 +61,6 @@ import org.springframework.integration.ip.IpHeaders;
 import org.springframework.integration.ip.tcp.TcpInboundGateway;
 import org.springframework.integration.ip.tcp.TcpOutboundGateway;
 import org.springframework.integration.ip.util.TestingUtilities;
-import org.springframework.integration.test.condition.LogLevels;
 import org.springframework.integration.test.util.TestUtils;
 import org.springframework.integration.util.SimplePool;
 import org.springframework.messaging.Message;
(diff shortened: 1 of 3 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
```diff
@@ -193,10 +200,44 @@ public void testFailoverAllDead() throws Exception {
...
-		failoverFactory.getConnection().send(message);
+		assertThatExceptionOfType(UncheckedIOException.class).isThrownBy(() ->
+			failoverFactory.getConnection().send(message));
 		Mockito.verify(conn2).send(message);
 	}
 
+	@Test
+	void failoverAllDeadAfterSuccess() throws Exception {
+		ServerSocket ss1 = ServerSocketFactory.getDefault().createServerSocket(0);
+		ServerSocket ss2 = ServerSocketFactory.getDefault().createServerSocket(0);
```
artembilan, line 211: Do we really need to do the **real** connection for this kind of test?
Won't just mocking target CCFs be enough?

This new test has failed on GH Actions workflow. But there was an OOM, so I'm not sure if that is a real failure...

artembilan (changes requested):

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
```diff
@@ -193,10 +201,52 @@ public void testFailoverAllDead() throws Exception {
...
+		});
+		cf2.setApplicationEventPublisher(event -> { });
+		FailoverClientConnectionFactory fccf = new FailoverClientConnectionFactory(List.of(cf1, cf2));
+		fccf.registerListener(msf -> {
+			latch.countDown();
+			return false;
+		});
+		fccf.start();
+		fccf.getConnection().send(new GenericMessage<>("test"));
+		assertThat(done.get(10, TimeUnit.SECONDS)).isTrue();
```
artembilan, line 244: Do we need this extra barrier to wait?
I think that `latch` on the next line is enough for this test.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
```diff
@@ -193,10 +201,52 @@ public void testFailoverAllDead() throws Exception {
...
-		failoverFactory.getConnection().send(message);
+		assertThatExceptionOfType(UncheckedIOException.class).isThrownBy(() ->
+			failoverFactory.getConnection().send(message));
 		Mockito.verify(conn2).send(message);
 	}
 
+	@LogLevels(level = "DEBUG", classes =
+			{ FailoverClientConnectionFactory.class, TcpNetClientConnectionFactory.class, TcpNetConnection.class },
+			categories =
+			"org.springframework.integration.ip.tcp.connection.FailoverClientConnectionFactory$FailoverTcpConnection")
```
artembilan, line 212: Do we still need a diagnostics then?

artembilan (changes requested):

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
garyrussell: Still failing - will remove when resolved.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
garyrussell: True - can mock the other factory - will do after I resolve the test race.

artembilan:

garyrussell:
Fixed backported tests and cherry-picked to 5.2.x

PR #3530, opened 2021-03-29
GH-3529: Add HTTP & WebFlux extractResponseBody

### LINKED ISSUE #3529: Option on AbstractHttpRequestExecutingMessageHandler to always return the whole ResponseEntity as a payload
**Expected Behavior**
Return ResponseEntity all the time based on some configuration flag

**Current Behavior**
HTTP outbound gateway either returns object or it returns ResponseEntity with given status

**Context**
https://stackoverflow.com/questions/66808030/return-responseentity-from-spring-integrations-gateway-method/66818396#66818396

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3529

* Expose a convenient `extractResponseBody` option on the HTTP client components
to let end-user to decide if the body of `ResponseEntity` must be extracted (default)
or the whole `ResponseEntity` should be produced as a reply message payload
* Remove a deprecated since `5.3` `encode-uri` option
* Document the new feature
* Rework `webflux.adoc` chapter for the code snippet switcher

### DIFF
--- a/src/reference/asciidoc/http.adoc
+++ b/src/reference/asciidoc/http.adoc
@@ -222,8 +222,8 @@ The `expected-response-type` must be compatible with the (configured or default)
 This can be an abstract class or even an interface (such as `java.io.Serializable` when you use Java serialization and `Content-Type: application/x-java-serialized-object`).
 =====
 
-Starting with version 5.5, the `HttpRequestExecutingMessageHandler` exposes an `extractResponseBody` flag (which is `true` by default) to return the whole `ResponseEntity` as a reply message payload or just its body, independently of the provided `expectedResponseType`.
-If body is not present in the `ResponseEntity`, this flag doesn't count and the whole `ResponseEntity` is returned.
+Starting with version 5.5, the `HttpRequestExecutingMessageHandler` exposes an `extractResponseBody` flag (which is `true` by default) to return just the response body, or to return the whole `ResponseEntity` as the reply message payload, independently of the provided `expectedResponseType`.
+If a body is not present in the `ResponseEntity`, this flag is ignored and the whole `ResponseEntity` is returned.
 
 [[http-namespace]]
 === HTTP Namespace Support
--- a/src/reference/asciidoc/http.adoc
+++ b/src/reference/asciidoc/http.adoc
@@ -8,8 +8,8 @@ See also <<./webflux.adoc#webflux,WebFlux Support>>.
 You need to include this dependency into your project:
 
 ====
+[source, xml, subs="normal", role="primary"]
 .Maven
-[source, xml, subs="normal"]
 ----
 <dependency>
     <groupId>org.springframework.integration</groupId>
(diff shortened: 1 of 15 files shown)


src/reference/asciidoc/http.adoc
```diff
@@ -223,6 +222,9 @@ The `expected-response-type` must be compatible with the (configured or default)
 This can be an abstract class or even an interface (such as `java.io.Serializable` when you use Java serialization and `Content-Type: application/x-java-serialized-object`).
 =====
 
+Starting with version 5.5, the `HttpRequestExecutingMessageHandler` exposes an `extractResponseBody` flag (which is `true` by default) to return the whole `ResponseEntity` as a reply message payload or just its body, independently of the provided `expectedResponseType`.
+If body is not present in the `ResponseEntity`, this flag doesn't count and the whole `ResponseEntity` is returned.
```
garyrussell, line 226: This would be better the other way around.

```suggestion
Starting with version 5.5, the `HttpRequestExecutingMessageHandler` exposes an `extractResponseBody` flag (which is `true` by default) to return just the response body, or to return the whole `ResponseEntity` as the reply message payload, independently of the provided `expectedResponseType`.
If a body is not present in the `ResponseEntity`, this flag is ignored and the whole `ResponseEntity` is returned.
```

src/reference/asciidoc/webflux.adoc
```diff
@@ -257,93 +289,30 @@ The following example shows how to configure both an outbound gateway and an out
...
-}
-----
-====
+Starting with version 5.2, the `WebFluxRequestExecutingMessageHandler` supports reactive `Publisher`, `Resource`, and `MultiValueMap` types as the request message payload.
+A respective `BodyInserter` is used internally to be populated into the `WebClient.RequestBodySpec`.
+When the payload is a reactive `Publisher`, a configured `publisherElementType` or `publisherElementTypeExpression` can be used to determine a type for the publisher's element type.
+The expression must be resolved to a `Class<?>`, `String` which is resolved to the target `Class<?>` or `ParameterizedTypeReference`.
 
-The following example shows how to configure a WebFlux outbound gateway with the Java DSL:
+Starting with version 5.5, the `WebFluxRequestExecutingMessageHandler` exposes an `extractResponseBody` flag (which is `true` by default) to return the whole `ResponseEntity` as a reply message payload or just its body, independently of the provided `expectedResponseType` or `replyPayloadToFlux``.
```
garyrussell, line 312: Reword order - see suggestion above.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -63,7 +63,8 @@ See <<./redis.adoc#redis,Redis Support>> for more information.
 ==== HTTP Changes
 
 The `HttpRequestExecutingMessageHandler` doesn't fallback to the `application/x-java-serialized-object` content type any more and lets the `RestTemplate` make the final decision for the request body conversion based on the `HttpMessageConverter` provided.
-
+It also has now an `extractResponseBody` option (`true` by default) to return the whole `ResponseEntity` as a reply message payload or just its body.
+Same option is present for the `WebFluxRequestExecutingMessageHandler`.
```
garyrussell, line 67: The same...

Also reorder previous sentence as in suggestion above.

garyrussell (changes requested):

garyrussell (approved):

PR #3532, opened 2021-04-01
GH-3502: More refactoring to avoid reflection

### LINKED ISSUE #3502: Revise configuration code for better Spring Native support
See more info in [Spring Native](https://docs.spring.io/spring-native/docs/0.9.0-SNAPSHOT/reference/#_designing_native_friendly_spring_libraries).

At the moment it is important to revise `ImportSelector` impls and class presence checks.

Related to: https://github.com/spring-projects/spring-framework/issues/18353

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3502

* Move `ChannelInitializer` bean registration into an `AbstractIntegrationNamespaceHandler` -
it was never used for annotations and Java DSL...
* Rework `IntegrationFlows.fromSupplier()` to call a provided `Supplier` directly -
not via reflection in the `MethodInvokingMessageSource`
* Resolve new Sonar smells
* Rework `EndpointSpec` to accept an expected factory bean instance via ctor arg
instead of reflection
* Rework `Jackson2JsonObjectMapper` to use well-known module instances directly -
not via reflection from their class names

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/DefaultMethodInvokingMethodInterceptor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/DefaultMethodInvokingMethodInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,7 +34,7 @@ import org.springframework.util.ConcurrentReferenceHashMap.ReferenceType;
 import org.springframework.util.ReflectionUtils;
 
 /**
- * Method interceptor to invoke default methods on the repository proxy.
+ * Method interceptor to invoke default methods on the gateway proxy.
  *
  * The copy of {@code DefaultMethodInvokingMethodInterceptor} from Spring Data Commons.
  *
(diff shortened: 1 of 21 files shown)


spring-integration-core/src/main/java/org/springframework/integration/gateway/DefaultMethodInvokingMethodInterceptor.java
```diff
@@ -49,8 +49,7 @@
 
 	private final MethodHandleLookup methodHandleLookup = MethodHandleLookup.getMethodHandleLookup();
 
-	private final Map<Method, MethodHandle> methodHandleCache =
-			new ConcurrentReferenceHashMap<>(10, ReferenceType.WEAK);
+	private final Map<Method, MethodHandle> methodHandleCache = new ConcurrentReferenceHashMap<>(10, ReferenceType.WEAK);
```
garyrussell, line 52: Why? It's now 121 chars.

garyrussell (changes requested):
Trivial, but...

spring-integration-core/src/main/java/org/springframework/integration/gateway/DefaultMethodInvokingMethodInterceptor.java
artembilan: Sometimes I miss an assignment second line during review and I thought `121` is not too far from `120` 😄.

But rule is rule: reverted.

garyrussell (approved):

PR #3534, opened 2021-04-04
Update Smack to 4.4.2

### LINKED ISSUE #3462: Upgrade to Smack 4.4
See more info here: https://download.igniterealtime.org/smack/docs/latest/changelog.html

/CC @Flowdalic

Flowdalic:
The second commit is causing an API change. Not sure what spring integration's policy regarding those is.

Fixes #3462.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -95,7 +95,7 @@ ext {
 	rsocketVersion = '1.1.0'
 	saajVersion = '1.5.2'
 	servletApiVersion = '4.0.1'
-	smackVersion = '4.3.5'
+	smackVersion = '4.4.2'
 	soapVersion = '1.4.0'
 	springAmqpVersion = project.hasProperty('springAmqpVersion') ? project.springAmqpVersion : '2.3.6'
 	springDataVersion = project.hasProperty('springDataVersion') ? project.springDataVersion : '2021.0.0-SNAPSHOT'
@@ -868,7 +868,7 @@ project('spring-integration-xmpp') {
 	dependencies {
 		api project(':spring-integration-core')
 		api "org.igniterealtime.smack:smack-tcp:$smackVersion"
-		api "org.igniterealtime.smack:smack-java7:$smackVersion"
+		api "org.igniterealtime.smack:smack-java8:$smackVersion"
 		api "org.igniterealtime.smack:smack-extensions:$smackVersion"
 
 		testImplementation project(':spring-integration-stream')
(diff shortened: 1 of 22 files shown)


artembilan (changes requested):
Thank you, @Flowdalic , for contribution!

I wonder if we can introduce some intermediary adapter to avoid a third generic argument for `HeaderMapper` and don't confuse the rest impls. 

That intermediary could be created and extracted already in the XMPP channel adapters.

Since it is still going to be a breaking change, I would defer the fix until `6.0` anyway.

Thanks for understanding!

Flowdalic:
> I wonder if we can introduce some intermediary adapter to avoid a third generic argument for `HeaderMapper` and don't confuse the rest impls.

A third? You mean a second, right?

Note that this PR contains two PRs: One that updates Smack and one that adds a second generic parameter to `RequestReplyHeaderMapper`. You could apply the first one, the one that updates Smack now, as it does not change the spring-integration API, and the second one later.

> Thanks for understanding!

No worries. I am not in a rush. :)

artembilan:
> it does not change the spring-integration API

It does because of `public interface XmppHeaderMapper extends RequestReplyHeaderMapper<MessageBuilder>`.
So, even without a second generic arg it is still an new contract for the `XmppHeaderMapper`.

Therefore bear with us how we care about backward compatibility.
Saying that I definitely OK to live with the `XmppHeaderMapper extends RequestReplyHeaderMapper<MessageBuilder>`. This is the one intermediary I've talked about.

Now I remember that I came up with similar solution..
My problem was that I couldn't figure out how to fix extensions, like that GCM.
If that works with your changes, then it is great and I'll review and accept your fix first thing when we switch to `6.0`!

artembilan (changes requested):
Hi @Flowdalic !

sorry for a big delay 😄 

Would you mind to revise your PR? 
With rebasing to the current `main`.
Probably more fresh version of Smack library.
And I'd prefer to not have that second generic arg on the header mapper.
Perhaps that `org.jivesoftware.smack.packet.MessageBuilder` should be enough for us as input type and output...

The next week in the next our milestone release. This one can make it there.
If not, that's OK we can incorporate it later.

Thank you!

artembilan:
Hi @Flowdalic !

I applied only your first commit and merged it like this: https://github.com/spring-projects/spring-integration/commit/330dfd03f2c448215ba082fd9855b2114054c243.

Issue #3451, opened 2021-01-04
Consider to use GreenMail for testing mail module


artembilan:
More info is here: https://greenmail-mail-test.github.io/greenmail/.

It looks like we can replace our `TestMailServer` with more mail integration coverage from such an embedded mail server and its features.

PR #3535, opened 2021-04-04
fixes a race condition preventing idle() to return as expected

apinske:
IMAPFolder.idle() by default keeps idle-ing after each response. We
don't need that, because we want to fetch the new mails immediately
(over the same connection). To make idle() not keep on going, we called
folder.isOpen() which in most cases makes idle() stop by calling noop()
internally (to keep the connection alive) which leads to the current
idle()-call being canceled.

The problem this commits addresses is that the call to noop() only
happens if there were no calls in the last second. There is a check for
that in com.sun.mail.imap.IMAPFolder.keepConnectionAlive. So if a new
message appears less then a second after idle started, we will miss it.

This new way interrupts idle() more often than before, e.g. when there
is an expunge-message (i.e. a message was deleted), which we don't care
about. But the subsequent receive() in IdleTask will simply get no
messages and then turn around and start idle-ing again.

I have a small repo using spring-boot to reproduce the issue here https://github.com/apinske/playground-mail/blob/master/src/test/java/eu/pinske/playground/playground/mail/PlaygroundMailApplicationTests.java

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapMailReceiver.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapMailReceiver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,9 +27,6 @@ import javax.mail.Flags.Flag;
 import javax.mail.Folder;
 import javax.mail.Message;
 import javax.mail.MessagingException;
-import javax.mail.event.MessageCountAdapter;
-import javax.mail.event.MessageCountEvent;
-import javax.mail.event.MessageCountListener;
 import javax.mail.search.AndTerm;
 import javax.mail.search.FlagTerm;
 import javax.mail.search.NotTerm;
@@ -60,8 +57,6 @@ public class ImapMailReceiver extends AbstractMailReceiver {
 	private static final int DEFAULT_CANCEL_IDLE_INTERVAL = 120000;
 
 	private static final String PROTOCOL = "imap";
-
-	private final MessageCountListener messageCountListener = new SimpleMessageCountListener();
 
 	private final IdleCanceler idleCanceler = new IdleCanceler();
 
@@ -188,16 +183,14 @@ public class ImapMailReceiver extends AbstractMailReceiver {
 		else if (!folder.getPermanentFlags().contains(Flags.Flag.RECENT) && searchForNewMessages().length > 0) {
 			return;
 		}
-		imapFolder.addMessageCountListener(this.messageCountListener);
 		try {
 			this.pingTask = this.scheduler.schedule(this.idleCanceler,
 					new Date(System.currentTimeMillis() + this.cancelIdleInterval));
 			if (imapFolder.isOpen()) {
-				imapFolder.idle();
+				imapFolder.idle(true);
 			}
 		}
 		finally {
-			imapFolder.removeMessageCountListener(this.messageCountListener);
 			if (this.pingTask != null) {
 				this.pingTask.cancel(true);
 			}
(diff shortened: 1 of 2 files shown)


apinske:
I made an alternative version of the fix that keeps more of the current flow: https://github.com/apinske/spring-integration/commit/4ea047fd450a3c3fcb2a6d2708f40aa94514d4a1

Hoping for feedback which one you prefer.

artembilan (changes requested):
I like the solution in this PR more.
Anyway: is there a chance to have some test-case to cover this problem?

Thanks

artembilan:
I see you use GreenMail in your sample project. Probably it is a chance for us to add its support, too: https://github.com/spring-projects/spring-integration/issues/3451

apinske:
I thought about tests, but with the current use of mocks it's quite hard/useless. I just added IDLE to Greenmail (not released yet: https://github.com/greenmail-mail-test/greenmail/pull/382). I could try adding some tests with the new version, can't make any promises yet. The test in my repo failed sometimes (depending on CPU load, i.e. if it took more than one second from start to mail-receive).

artembilan:
I see. 

Thanks for letting us know!

OK. So, I'm going to merge it as is (and back-port) and you are free to come back to us with Greenmail move in the future!

Probably the issue you are fixing is the reason why our mock tests fail from time to time...

apinske:
"our mock tests fail from time to time" That sounds interesting. You happen to have a log of such a time? I imagined the mock tests couldn't run in that problem.

artembilan:
> You happen to have a log of such a time?

Not at the moment. But I can keep you posted if you are interested and that happens again in the future.

apinske:
Absolutely, please do. Now you got me hooked. :) Because that means I haven't understood the current tests properly.

artembilan:
Actually, you know: fortunately, it really happened a couple days ago: https://build.spring.io/artifact/INT-MASTERSPRING40/JOB1/build-1406/Test-Results/spring-integration-mail/build/reports/tests/testAll/index.html.

Feel free to look into that!

I won't mind if the problem is not related to what you are fixing here.

apinske:
Wow. 100MB of Logs. That should be very helpful.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/c392d389c6c6044de069a2d8c81a6934b1acc284 and cherry-picked to `5.4.x` & `5.3.x`.

@apinske ,

Here is some remark for you: https://chris.beams.io/posts/git-commit/

apinske:
I have reproduced the test failure locally, with tweaked timeouts - my laptop is too slow otherwise. :) Resulting in the same connection refused exceptions.
The good news: The fix in this PR should address most of the cases. The bad news: There is another race in IdleCanceler, which in the tests gets set to 1 second (should never happen in production system, we e.g. use 29 minutes) and uses the same `folder.isOpen` "trick".
See #3544.

apinske:
Unfortunately I discovered that there is still another possible race: When a new message is added between the last SEARCH and the IDLE. Such a message can be reported in an immediate unsolicited response to IDLE (before the continuation). 
```
A20 IDLE
* 2 EXISTS
* 1 RECENT
+ OK
```
We miss those, as they are only reported to the Listeners and not used to cancel the IDLE immediately. The previous `SimpleMessageCountListener` would have missed those too, because of the 1-second-race. We still might need the solution in apinske@4ea047f, although it feels hack-ish.
I am still trying to come up with a good test case. Currently I'm hacking various sleeps in various places to provoke the behaviour.
I will also have a look into the new `com.sun.mail.imap.IdleManager`. Judging by the [release notes](https://eclipse-ee4j.github.io/mail/docs/CHANGES.txt), it's had its fair share of concurrency problems, though.

artembilan:
Thanks, @apinske , for sharing your investigation!

We probably are going to live with that flaw for some time.

The `IdleManager` sounds promising. 
We have plans to upgrade to the latest Jakarta EE in the next Spring `6.0` generation. That's probably going to be a good oportunity for us to revise an `ImapIdleChannelAdapter` (or just `ImapMailReceiver`) logic in favor of `IdleManager` API.

Feel free to raise a respective GH issue!

PR #3536, opened 2021-04-06
GH-3520 Make TcpConnectionInterceptorSupport.getConnectionId() final

### LINKED ISSUE #3520: Make TcpConnectionInterceptorSupport.getConnectionId() final
Overriding the connection id will cause routing problems because the connection id in inbound messages is from the core connection, not the interceptor.

deepesh-verma:
Fixes [GH-3520](https://github.com/spring-projects/spring-integration/issues/3520)

Overriding the connection id will cause routing problems because the connection id in inbound messages is from the core connection, not the interceptor.

Signed-off-by: Deepesh <deepeshvrm1@gmail.com>

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Thank you for considering to fix this issue!
But see error for the `CachingClientConnectionFactory`.

Please, be sure to run `gradlew clean check` locally before pushing changes.

Regarding the problem: see that `CachedConnection`.
We have two possible solution I guess:

1. Remove an overridden `getConnectionId()` and fully rely on super.
2. Not extend from the `TcpConnectionInterceptorSupport`.

I see @garyrussell has assigned this issue to himself, plus moved it to the current `5.5-RC1` version. So, let's ask him about an opinion! 

In my filing with this `final` we introduce some breaking change. Minor, but still.
Although I believe it worth that since we have some discrepancy in other places allowing end-user to override the `getConnectionId()`...

garyrussell:
The issue is on inbound connections only; since the CCCF is client-side, it's not a problem there.

Since the same base class is used for both inbound and outbound connections, we can't make it final.

I don't think it's worth adding more complexity to solve this issue, so I think we should close the issue as invalid.

deepesh-verma:
Closing this PR, because as @garyrussell suggested, this issue should be marked invalid. 

Will make sure to run the build locally for any future changes, or mark the PR as a draft one.

PR #3538, opened 2021-04-07
Get rid of `master` word

artembilan:


artembilan:
Superseded with #3539

PR #3543, opened 2021-04-09
Revise retry for `IntegrationReactiveUtils`

artembilan:
Related to: https://github.com/spring-cloud/stream-applications/issues/156

The `IntegrationReactiveUtils` uses a general `Flux.retry()`
operator to always retry for all the errors.
On the other hand it has only a `.doOnError(MessagingException.class)`
which leads swallowing all the other exceptions from logs and handling.

* Replace with `retryWhen()` for the `MessagingException` predicate
failing for all other exceptions.
The end-user may add their own `retry` or error handling mechanism to
the returned `Flux` from the `IntegrationReactiveUtils`

**Cherry-pick to 5.4.x**


garyrussell:
...and cherry picked

Issue #3470, opened 2021-01-22
Building With JDK11 Breaks TCP (NIO)


garyrussell:
See https://github.com/spring-projects/spring-integration-samples/issues/291

`ByteBuffer.flip()` now returns `ByteBuffer`, previously it returned `Buffer`.

Cannot run with JDK8.

```
Exception in thread "pool-2-thread-2" java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
	at org.springframework.integration.ip.tcp.connection.TcpNioConnection.doRead(TcpNioConnection.java:437)
	at org.springframework.integration.ip.tcp.connection.TcpNioConnection.readPacket(TcpNioConnection.java:499)
	at org.springframework.integration.ip.tcp.connection.AbstractConnectionFactory.lambda$processNioSelections$7(AbstractConnectionFactory.java:708)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

WorkAround:

https://github.com/plasma-umass/doppio/issues/497#issuecomment-334740243

garyrussell:
Reopening. Other methods suffer from the same problem.

PR #3545, opened 2021-04-09
Fix compatibility with Java 8 for ByteBuffer

artembilan:
When we build with Java 9+ and target for Java 8,
we get an incompatible bytecode around these method for `ByteBuffer`:

```
position(int)
limit(int)
mark()
reset()
clear()
flip()
rewind()
```

The recommendation is to cast to `Buffer` when we call these methods

* Fix all the production code using `ByteBuffer` for recommended cast to `Buffer`

Related to https://build.spring.io/browse/INTEXT-AWS-306

See more info in this Jetty issue: https://github.com/eclipse/jetty.project/issues/3244


garyrussell:
Also https://github.com/spring-projects/spring-integration/issues/3470

We must not have any samples that hit any of these other classes 😦

artembilan:
Yeah... Doesn't matter. It's great to catch it sooner than later. Plus it is just in time for for upcoming RC 😄

garyrussell (approved):

PR #3546, opened 2021-04-09
Use GreenMail for mail tests

apinske:
Wanted to confirm that this is going in the direction you were imagining in #3451?


artembilan:
Yes, @apinske , this is exactly what I was looking for when I found a Greenmail testing library!

My plans were to do this for the next `6.0` generation, but having your contribution we can revise a decision for the current `5.5`.
Although I know that your IMAP support just has been merged over there recently: https://github.com/greenmail-mail-test/greenmail/pull/382, it doesn't mean that it stops us somehow to rework our SMTP and POP3 tests to rely already on this Greenmail  solution.

So, no rush, but feel free to continue your contribution with this whenever you have a time.

Thank you again!

apinske:
I have no idea what's wrong with the build job. Mail-Tests run locally. Here sometimes some other tests fail or the build hangs. @artembilan Could you please have a look?

artembilan:
Yeah... Saw that. Tried to restart. Then went to releases in other projects and forgot.

Would you mind to pull the latest `main`, rebase your branch and re-push?

As long as you cover all the mail protocols in your tests refreshing, the change may make it for tomorrow's release.

I'll review what you you have so far in an hour or so.

apinske:
rebased. 
The test are as before, I just swapped the backing server.

build.gradle
```diff
@@ -641,9 +642,9 @@ project('spring-integration-mail') {
 		api project(':spring-integration-core')
 		api 'org.springframework:spring-context-support'
 
-		providedImplementation "jakarta.mail:jakarta.mail-api:$mailVersion"
-		providedImplementation "com.sun.mail:imap:$mailVersion"
 		providedImplementation "com.sun.mail:jakarta.mail:$mailVersion"
+		testImplementation "com.icegreen:greenmail:$greenmailVersion"
+		testImplementation "org.slf4j:jul-to-slf4j:1.7.30"
```
artembilan, line 647: Why do we need this extra dependency?
Nevertheless it is better to wrap literal into a single quotes in Groovy: otherwise it is treated as GString and evaluated when Gradle build is executed. Kinda some performance optimization.

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
```diff
@@ -100,31 +110,69 @@
...
 
+	private static class ImapSearchLoggingHandler extends Handler {
+		private final List<String> searches = new ArrayList<>();
+		private final List<String> stores = new ArrayList<>();
+		private static final String SEARCH = " SEARCH ";
+		private static final String STORE = " STORE ";
+
+		@Override
+		public void publish(LogRecord record) {
+			if ("com.sun.mail.imap.protocol".equals(record.getLoggerName())) {
```
artembilan, line 127: I see you are making some "hard core" logic to catch and analyze mail logs. If there is no other way to implement our tests, can we consider to do that without extra SLF in the middle?
Looks like we just can bridge JUL to Log4J directly: https://stackoverflow.com/questions/21781026/how-to-send-java-util-logging-to-log4j2

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
```diff
@@ -100,31 +110,69 @@
 
 	private AtomicInteger failed;
 
-	private ImapServer imapIdleServer;
+	private GreenMail imapIdleServer;
+	private GreenMailUser user;
```
artembilan, line 114: Our preference is to wrap every single class member with blank lines.
Much easier to review 😄

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
```diff
@@ -193,6 +237,11 @@ public void testIdleWithServerGuts(ImapMailReceiver receiver, boolean mapped, bo
 		adapter.setTaskScheduler(taskScheduler);
 		adapter.setReconnectDelay(1);
 		adapter.start();
+		MimeMessage message = GreenMailUtil.createTextEmail("Foo <foo@bar>", "Bar <bar@baz>", "Test Email", "foo\r\n", imapIdleServer.getImap().getServerSetup());
```
artembilan, line 240: For better readability we recommend to have code lines not longer than `120` symbols.

artembilan (changes requested):
You can find some of my review on lines.

In addition:

Please, consider to not remove `TestMailServer` class, but rather deprecate.
It is a public API and some one may use it in their own tests.
Having it deprecated right now in the `5.5` version, gives us a freedom to remove in the next `6.0` one.

It is probably would be great to mention GreenMail library in our testing support chapter of the docs: https://docs.spring.io/spring-integration/docs/current/reference/html/testing.html#testing

Thank you for contribution!

If you don't have anything to add to the PR, or no time until tomorrow to fix my concerns, I'll take a look in the morning one more time and will try to include such a great improvement in upcoming `5.5` GA.

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
apinske: Yeah, it's a bit weird. The tests used state of the old TestMailServer to assert parts of a query. I wanted to replicate that. It's not the nicest code. :)
I'll look into the log4j-redirect. Never used it, and didn't word=k on first try, because JUnit initialises JUL "too early". Is there anywhere in the codebase a test that asserts in some log output I could steal?

apinske:
Thanks for the review! I don't think I will be able to incorporate it fully in the next days. I'll take the next release. Its' just tests after all. :)
Will there be a 5.6? Otherwise I will provide a separat PR to just deprecate the TestMailServer.

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
artembilan: OK. No problem. Perhaps you are right and we need to revise the logic of this test instead of trying to find some struts.

Well, don't assert log outputs, but rather "spy" on the logs if we indeed need to assert against some state which is available only via logging.

See something like this pattern:
```
LogAccessor logger = spy(TestUtils.getPropertyValue(channel, "logger", LogAccessor.class));
when(logger.isDebugEnabled()).thenReturn(true);
new DirectFieldAccessor(channel).setPropertyValue("logger", logger);
...
ArgumentCaptor<String> captor = ArgumentCaptor.forClass(String.class);
verify(logger, times(2)).debug(captor.capture());
List<String> logs = captor.getAllValues();
```
The class I took this snippet is `DirectChannelTests`

artembilan:
> Will there be a 5.6? Otherwise I will provide a separate PR to just deprecate the `TestMailServer`.

Apparently no: I'd like to stay away from `5.6`. Too many version to support in parallel.
No problem with `TestMailServer`! I can take care about its deprecation myself.

Plus, as you said, this is just tests, so not a big deal what the code is.
Only my concern from your PR that there are two new dependencies to maintain in the future.
When one of them is indeed a reason of this fix when other, for SLF, we probably could avoid.

Thanks for understanding and no problem if you can't make it for tomorrow!

artembilan:
Pulling locally for final review and some clean up...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/996daca228e764f528bc76edb9e33a432ed552e9.

* restored `TestMailServer` class and deprecate it.
* Remove SLF dependency and rely on the `log4j-jul` instead.
* Clean up the code style.

@apinske ,

thank you again for great contribution; looking forward for more!

PR #3550, opened 2021-04-16
Revise MessageHistory configuration

artembilan:
Currently the `MessageHistoryRegistrar` can parse several sources for message history -
`@EnableMessageHistory` and/or `<message-history>`.
Since its logic relies on the reflection it is not compatible with Spring Native.
Plus it causes confusion when several sources are declared so when time comes to
change something in that configuration, we may miss some place to re-align with
our new requirements.

Better to reject extra configurations and enforce end-users to use only one
`@EnableMessageHistory` or `<message-history>`.
This is actually a preferences for many other `@Enable...` in Spring portfolio.
Plus we got a benefit with a Spring Native compatibility

* Fix `MessageHistoryRegistrar` to parse only one `@EnableMessageHistory`.
Register `MessageHistoryConfigurer` function way for Spring Native compatibility
* Clean up `PublisherRegistrar` for better readability
* Fix message history tests which exposed several configurations
* Add JavaDoc into `EnableMessageHistory`
* Refactor `MessageHistoryConfigurer` to let to override patterns configuration
at runtime
* Clean up `message-history.adoc`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/EnableMessageHistory.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/EnableMessageHistory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 the original author or authors.
+ * Copyright 2014-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,9 +25,11 @@ import java.lang.annotation.Target;
 import org.springframework.context.annotation.Import;
 
 /**
- * Enables {@link org.springframework.integration.history.MessageHistory} for Integration components.
+ * Enables {@link org.springframework.integration.history.MessageHistory}
+ * for Spring Integration components.
  *
  * @author Artem Bilan
+ *
  * @since 4.0
  */
 @Target(ElementType.TYPE)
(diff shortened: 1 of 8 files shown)


artembilan:
I'll add Migration Guide note after merging.

Debated with myself if this should be treated as a serious breaking change...

I have convinced myself like this:

1. All other `@Enable...` in Spring Portfolio don't care about multi-declaration and probably will fail in Spring Boot with its `bean-definition-override=false`. We are just more proactive here and fail fast when several times declared.
2. We still had a protection from mismatching from different config sources and since no one complained so far, it looks like people don't use several `@EnableMessageHistory` anyway.
3. We have already enough small breaking changes in this version for better end-user experience, so probably this one extra simplicity won't hurt too much.

Any feedback welcome!

garyrussell (approved):

garyrussell:
I am ok with the breaking change, but it's a bit unusual to do that between RC and GA, maybe an interim RC2?

Looks like the build failure is unrelated, let me know if you want to merge as-is.

artembilan:
Yes, `RC2` makes sense. Although I treated this change as an internal one - no end-user API changes: just a little behavior 😉 
On the other hand it is a bit unusual to have GA after RC1 only in a month...
Let's see next week if I can come up with similar changes while I'm trying to align with Spring Native compatibility! 
Then we can roll out RC2 in two weeks or so.

artembilan:
OK. Green finally.

PR #3556, opened 2021-04-27
GH-3554: Eval remote dir on each synchToLocal

### LINKED ISSUE #3554: SftpInboundAdapter - RemoteDirectoryExpression is not evaluated on each poll

5.3.2.RELEASE

**Describe the bug**

According to the documentation, it's possible to specify a `remote-directory-expression` to make the sftp inbound adapter dynamically poll from the result of the expression. 

> Starting with version 4.2, you can specify remote-directory-expression instead of remote-directory, letting you dynamically determine the directory on each poll — for example, remote-directory-expression="@myBean.determineRemoteDir()".

This is not working as discussed in this StackOverflow question: https://stackoverflow.com/questions/67218120/poll-different-remote-directories-using-sftpinboundadapter-and-remotedirectoryex

**To Reproduce**

Create an IntegrationFlow that uses a SftpInboundAdapter with a `remoteDirectoryExpression` like:

```
@Bean
    return IntegrationFlows.from(
            Sftp.inboundAdapter(sftpSessionFactory())
                    .preserveTimestamp(true)
                    .localDirectory(new File("sftp-inbound"))
                    .deleteRemoteFiles(true)
                    .autoCreateLocalDirectory(true)
                    .remoteDirectoryExpression(
                            EXPRESSION_PARSER.parseExpression("@directoryProvider.getDirectory()"))
                    .filter(getFilter())
            , e -> e.id("sftpInboundAdapter")
                    .autoStartup(true)
                    .poller(Pollers.fixedDelay(5, TimeUnit.SECONDS)))
            .log()
            .channel("processing")
            .get();
}
``` 

The bean `DirectoryProvider` returns a different folder (in round-robin) on each call

**Expected behavior**

On each poll, the remoteDirectoryExpression should be evaluated

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3554

The `remoteDirectoryExpression` was introduced into an
`AbstractInboundFileSynchronizer` to let end-user to evaluate a remote directory
on each poll (essentially on each `synchronizeToLocalDirectory()` call).
The fix for the https://jira.spring.io/browse/INT-4491 introduced a regression
when such an expression was evaluated only once when we call a `setRemoteDirectory()`.
So, an original purpose of this option was lost and we don't get an actual
remote dir on each poll

* Remove `evaluateRemoteDirectory()` method and its usage since it doesn't
reflect expectation of the remote dir expression property
* Reinstate the `remoteDirectoryExpression` evaluation in the
`synchronizeToLocalDirectory()`
* Propagate the result of that expression into further methods for copying
files
* Remove setting of the `remoteDirectory` variable into a global `EvaluationContext`
of the `AbstractInboundFileSynchronizer` instance since it is not thread-safe
* Instead create an `EvaluationContext` locally for each `synchronizeToLocalDirectory()`
call and set the `remoteDirectory` variable into this scoped instances
* Generate a local file name from the `localFilenameGeneratorExpression`
against locally created `EvaluationContext` with the mentioned `remoteDirectory` variable
* Cover the expected functionality with a unit-test

### DIFF
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/synchronizer/AbstractRemoteFileSynchronizerTests.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/synchronizer/AbstractRemoteFileSynchronizerTests.java
@@ -38,13 +38,12 @@ import org.junit.jupiter.api.io.TempDir;
 
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.expression.EvaluationContext;
-import org.springframework.integration.expression.FunctionExpression;
+import org.springframework.integration.expression.SupplierExpression;
 import org.springframework.integration.file.HeadDirectoryScanner;
 import org.springframework.integration.file.filters.AcceptOnceFileListFilter;
 import org.springframework.integration.file.filters.ChainFileListFilter;
 import org.springframework.integration.file.remote.session.Session;
 import org.springframework.integration.file.remote.session.SessionFactory;
-import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
 
 /**
@@ -258,7 +257,7 @@ public class AbstractRemoteFileSynchronizerTests {
 		Queue<String> remoteDirs = new LinkedList<>();
 		remoteDirs.add("dir1");
 		remoteDirs.add("dir2");
-		sync.setRemoteDirectoryExpression(new FunctionExpression<Message<?>>(s -> remoteDirs.poll()));
+		sync.setRemoteDirectoryExpression(new SupplierExpression<>(remoteDirs::poll));
 		sync.setLocalFilenameGeneratorExpressionString("#remoteDirectory+'/'+#root");
 		sync.setBeanFactory(mock(BeanFactory.class));
 		sync.afterPropertiesSet();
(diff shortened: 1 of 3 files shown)


spring-integration-file/src/test/java/org/springframework/integration/file/remote/synchronizer/AbstractRemoteFileSynchronizerTests.java
```diff
@@ -209,13 +218,60 @@ public void testExclusiveWatchService() {
...
+					protected String protocol() {
+						return "mock";
+					}
+
+				};
+
+		Queue<String> remoteDirs = new LinkedList<>();
+		remoteDirs.add("dir1");
+		remoteDirs.add("dir2");
+		sync.setRemoteDirectoryExpression(new FunctionExpression<Message<?>>(s -> remoteDirs.poll()));
```
garyrussell, line 261: The function input is not a `Message<?>` (it actually doesn't have a root object, but I guess you have to specify something).

My concern is that reading this test might lead the reader to assume we have a `Message<?>` here. `FunctionExpression<Void>` ?

spring-integration-file/src/test/java/org/springframework/integration/file/remote/synchronizer/AbstractRemoteFileSynchronizerTests.java
```diff
@@ -209,13 +218,60 @@ public void testExclusiveWatchService() {
...
+
+		sync.synchronizeToLocalDirectory(localDir);
+		sync.synchronizeToLocalDirectory(localDir);
+
+		/*Files.find(localDir.toPath(),
+				Integer.MAX_VALUE,
+				(filePath, fileAttr) -> fileAttr.isRegularFile())
+				.forEach(System.out::println);*/
+
+		assertThat(localDir.list()).containsAll(remoteDirs);
```
garyrussell, line 274: `remoteDirs` is empty by now, so this will always pass.

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
... and cherry-picked to 5.4.x, 5.3.x.

PR #3562, opened 2021-05-05
GH-3560 Parse mail FROM as comma-delimited header

### LINKED ISSUE #3560: spring-integration-mail: RFC5322 conforming messages containing multiple values in `from` or `reply-to` headers cannot be transformed

spring-integration-mail:5.4.5

**Describe the bug**

https://github.com/spring-projects/spring-integration/blob/main/spring-integration-mail/src/main/java/org/springframework/integration/mail/support/MailUtils.java#L52 and https://github.com/spring-projects/spring-integration/blob/main/spring-integration-mail/src/main/java/org/springframework/integration/mail/support/MailUtils.java#L56 expect a single address, and throw an exception `org.springframework.messaging.MessagingException: conversion of MailMessage headers failed; nested exception is java.lang.IllegalStateException: expected a single value but received an Array` in https://github.com/spring-projects/spring-integration/blob/main/spring-integration-mail/src/main/java/org/springframework/integration/mail/support/MailUtils.java#L69 while RFC 5322 https://tools.ietf.org/html/rfc5322#section-3.6.2 allows for a comma separated list of "mailbox specifications" (for `from`) or "addresses" (for `reply-to`). Messages that conform to RFC 5322 cannot be transformed successfully.

**To Reproduce**

This was just a find after an analysis of production log error messages when receiving mails. It should be reproducible with mail messages containing multiple values in the `from` or `reply-to` fields according to RFC 5322.

**Expected behavior**

`from` and `reply-to` header fields should support multiple comma-separated addresses, be extracted successfully and copied to the `AbstractIntegrationMessageBuilder` at https://github.com/spring-projects/spring-integration/blob/main/spring-integration-mail/src/main/java/org/springframework/integration/mail/transformer/AbstractMailMessageTransformer.java#L83 

Inbound mail messages conforming to RFC 5322 should be transformable without throwing an exception. Currently, this leads to cases where we are not able to successfully receive RFC5322 conforming inbound mail.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3560

According RFC 5322 `FROM` and `REPLY-TO` received mail message can be
as an array of addresses.

* Fix `MailUtils` to present those arrays as comma-delimited strings for
Spring message headers
* Fix tests to deal already with several addresses for `FROM` mime header.

In the future we may change the logic to map those mime headers to arrays
as it states according the mentioned RFC

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/support/MailUtils.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/support/MailUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,12 +25,14 @@ import javax.mail.Message.RecipientType;
 
 import org.springframework.integration.mail.MailHeaders;
 import org.springframework.messaging.MessagingException;
-import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Utilities for handling mail messages.
  *
  * @author Gary Russell
+ * @author Artem Bilan
+ *
  * @since 4.3
  *
  */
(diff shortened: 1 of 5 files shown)


artembilan:
@garyrussell ,

If you still find this as a breaking change, e.g. someone may rely on such a bug and really does some logic according an exception thrown from the current `MailUtils.extractStandardHeaders()`, then I'm OK to leave this change only for the current `5.5` version with respective Migration Guide note.

Thanks

garyrussell (approved):

garyrussell:
...and cherry-picked.

PR #3564, opened 2021-05-11
INT-1828: Docs about filter & release strategy

artembilan:
JIRA: https://jira.spring.io/browse/INT-1828

Document a corner case when filter endpoint is involved in
the flow logic before an aggregator with sequence size release strategy

### DIFF
--- a/src/reference/asciidoc/aggregator.adoc
+++ b/src/reference/asciidoc/aggregator.adoc
@@ -260,8 +260,8 @@ The following example shows how to configure an aggregator with the previous imp
 ----
 ====
 
-NOTE: If the filter endpoint is involved in the flow upstream of an aggregator, the sequence size release strategy (fixed or based on the `sequenceSize` header) is not going to server its purpose because some messages from a sequence may be discarded by the filter.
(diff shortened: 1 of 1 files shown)


src/reference/asciidoc/aggregator.adoc
```diff
@@ -260,6 +260,10 @@ The following example shows how to configure an aggregator with the previous imp
 ----
 ====
 
+NOTE: If the filter endpoint is involved in the flow upstream of an aggregator, the sequence size release strategy (fixed or based on the `sequenceSize` header) is not going to server its purpose because some messages from a sequence may be discarded by the filter.
+In this case it is recommended to choose other `ReleaseStrategy` or use compensation messages sent from a discard sub-flow carrying some information in their content to be skipped in a custom complete group function.
```
garyrussell, line 264:
```suggestion
NOTE: If the filter endpoint is involved in the flow upstream of an aggregator, the sequence size release strategy (fixed or based on the `sequenceSize` header) is not going to serve its purpose because some messages from a sequence may be discarded by the filter.
In this case it is recommended to choose another `ReleaseStrategy`, or use compensation messages sent from a discard sub-flow carrying some information in their content to be skipped in a custom complete group function.
```

garyrussell (changes requested):

garyrussell (approved):

PR #3566, opened 2021-05-18
Set FTPSClient property once

Sintendo:
The sessionCreation property was being set twice by accident.

@pivotal-cla This is an Obvious Fix


artembilan:
Just in time for the current `5.5` GA release!

Thank you very much, @Sintendo , for contribution; looking forward for more!

PR #3569, opened 2021-05-25
GH-3509: Fix regression in TcpNetServerConnectionFactory

gigermocas:
This pull request fixes a regression introduced in https://github.com/spring-projects/spring-integration/commit/a2081137f6bab38c892dba379928ce87e68e59df on `TcpNetServerConnectionFactory`, re-triggering #3509

Fix intercepted tcp connector cleanup tests.
Add missing logic to call `setSenders` on wrapped connection.

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNetServerConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNetServerConnectionFactory.java
@@ -35,6 +35,7 @@ import org.springframework.util.Assert;
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Mário Dias
  *
  * @since 2.0
  *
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Thank you for contribution, but please, confirm how it has some memory leak regression.

The commit you are mentioning has nothing to do with the wrapped connection: it is just a code refactoring to satisfy SonarQube. Perhaps you mean some other changed before that...

I'm fully agree with the change for those tests, but it might be great if you also provide some tests change to confirm the problem and cover it with your fix.

gigermocas:
Sorry for the confusion, it's this one: https://github.com/spring-projects/spring-integration/commit/2d443f9143cc4b129719a09a520c06ed63c76da7

Regarding the tests, those tests that were fixed in this commit were meant to test that exact memory leak issue, introduced in the fix for #3509. They would still fail if not for a posterior code change on that final assertion in https://github.com/spring-projects/spring-integration/commit/90a2f4bff145c2728f5142bcb6c33d032cdc5681

artembilan (approved):
Good point!

It's sad that those tests weren't full enough to catch the problem.
I see that `TcpNioServerConnectionFactory` has the fix for the same problem through the same #3509 issue. 

So, merging your fix and back-porting to `5.4.x!`

artembilan:
... and cherry-picked to `5.4.x`

PR #3571, opened 2021-05-27
GH-3570: Disable the generation of the Gradle metadata

### LINKED ISSUE #3570: Gradle Module Metadata for spring-integration-bom 5.4.7 missing from Maven Central

5.4.7 (also present in 5.5.0)

**Describe the bug**

[`spring-integration-bom-5.4.7.pom`](https://search.maven.org/artifact/org.springframework.integration/spring-integration-bom/5.4.7/pom) was generated with the following comment

```xml
<!-- This module was also published with a richer model, Gradle metadata,  -->
<!-- which should be used instead. Do not delete the following line which  -->
<!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
<!-- that they should prefer consuming it instead. -->
<!-- do_not_remove: published-with-gradle-metadata -->
```

which triggers the download by Gradle of the `spring-integration-bom-5.4.7.module` artifact, however that artifact was not published to Maven Central. This is probably a consequence of an upgrade to Gradle 6.

This is a problem when using Artifactory as a Maven proxy, because by default Artifactory returns a 403 HTTP status code for missing artifacts, and Gradle only ignores an absent `.module` when the repository returns a 404 HTTP status code. In my case (and others, probably), that means my build fails when it cannot download `spring-integration-bom-5.4.7.module`. 😭

See
- rsocket/rsocket-java#746 resolved by rsocket/rsocket-java#751
- remkop/picocli#1152 resolved by [`0874760`](https://github.com/remkop/picocli/commit/0874760e7caa5616dec3318515583b78105e92ab)

for similar issues in other projects that have bit me in the past (through spring-boot-dependencies).

**Expected behavior**

It would be good if either the module file was published on Maven Central, or the Gradle metadata generation was disabled.

fpavageau:
Fixes spring-projects/spring-integration#3570

The module file with the Gradle metadata cannot be published on Maven
Central, which causes issues for some people using a repository manager such
as Artifactory. If it doesn't return a 404 HTTP status, the build fails
instead of ignoring the module.

This should probably be cherry-picked for `5.4.x` and `5.3.x` at least. Should I create other pull requests for those branches?

### DIFF
--- a/gradle/publish-maven.gradle
+++ b/gradle/publish-maven.gradle
@@ -1,6 +1,10 @@
 apply plugin: 'maven-publish'
 apply plugin: 'com.jfrog.artifactory'
 
+tasks.withType(GenerateModuleMetadata) {
+	enabled = false
+}
+
 publishing {
 	publications {
 		mavenJava(MavenPublication) {


artembilan:
... and cherry-picked to `5.4.x` & `5.3.x`.

@fpavageau ,

PR #3577, opened 2021-06-08
Link documentation formats

Buzzardo:
Add a link from the HTML to the PDF and vice-versa.


artembilan (changes requested):
Hi @Buzzardo !

Thank you for contribution!

But we already have such a sentence in the `index.adoc`.
It might not be fully correct and I also don't see it in the PDF 😄 .

Would you mind to take a look there since you are in this task already?
Probably just remove that sentence `This documentation is also available as single..` ?

Buzzardo:
I missed that we have two HTML formats. I have change the index-header.adoc file accordingly. index-header.adoc is included only in the single-page HTML and PDF formats, so it provides a way to link to the home page (which already has the links you mentioned).

PR #3586, opened 2021-07-02
Functional gateway bean definitions

artembilan:
* Rework `MessagingGatewayRegistrar` to parse messaging gateway annotation
an `<gateway>` XML using a supplier variant for bean definition.
Such a feature is required by Spring Native - otherwise we would need to
register reflection info for to many internal Spring Integration classes
* Such a change should benefit from regular JDK perspective, too -
we don't do reflection for this kind of bean registrations

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingGatewayRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingGatewayRegistrar.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 the original author or authors.
+ * Copyright 2014-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


artembilan:
NOTE: I'll let to know when it is ready for merging.
Didn't have a chance to test it from Spring Native perspective, yet.

artembilan:
OK. As long as I see this one in native logs:
```
Caused by: com.oracle.svm.core.jdk.UnsupportedFeatureError: Proxy class defined by interfaces [interface com.example.integration.ControlBusGateway, interface org.springframework.aop.SpringProxy, interface org.springframework.aop.framework.Advised, interface org.springframework.core.DecoratingProxy] not found. 
```
I can tell that `GatewayProxyFactoryBean` is created properly. 
The Proxy configuration concern is out of scope of the work done in this PR and has to be addressed on Spring Native side.

artembilan:
See Spring Native PR for more info: https://github.com/spring-projects-experimental/spring-native/pull/894

garyrussell (approved):

PR #3588, opened 2021-07-13
GH-3424: Refactor to use logging methods from LogAccessor

### LINKED ISSUE #3424: More refactoring to use a LogAccessor from the IntegrationObjectSupport
For example an `AbstractPollingEndpoint` and its code like this:
```
	logger.debug("Poll interrupted - during stop()? : " + e.getMessage());
}
```

trgpa:
Fixes spring-projects/spring-integration#3424

- Logging statements that test cases depend on are kept unchanged (or should I have revised the test cases too?)
- Some include non-final variables so we cannot use LogAcessor.debug(Supplier<? extends CharSequence>). It might be a good idea to have methods similar to LogMessage.format().

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -93,9 +93,7 @@ public class CorrelatingMessageBarrier extends AbstractMessageHandler implements
 		synchronized (lock) {
 			this.store.addMessagesToGroup(correlationKey, message);
 		}
-		if (logger.isDebugEnabled()) {
-			logger.debug(String.format("Handled message for key [%s]: %s.", correlationKey, message));
-		}
+		logger.debug(() -> String.format("Handled message for key [%s]: %s.", correlationKey, message));
 	}
 
 	private Object getLock(Object correlationKey) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
@@ -20,6 +20,7 @@ import java.util.Iterator;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import org.springframework.core.log.LogMessage;
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.handler.AbstractMessageHandler;
 import org.springframework.integration.store.MessageGroup;
@@ -47,6 +48,7 @@ import org.springframework.messaging.Message;
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Trung Pham
  *
  * @see AbstractCorrelatingMessageHandler
  */
@@ -93,7 +95,7 @@ public class CorrelatingMessageBarrier extends AbstractMessageHandler implements
 		synchronized (lock) {
 			this.store.addMessagesToGroup(correlationKey, message);
 		}
-		logger.debug(() -> String.format("Handled message for key [%s]: %s.", correlationKey, message));
+		logger.debug(LogMessage.format("Handled message for key [%s]: %s.", correlationKey, message));
 	}
 
 	private Object getLock(Object correlationKey) {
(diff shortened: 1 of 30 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
```diff
@@ -93,9 +93,7 @@ protected void handleMessageInternal(Message<?> message) {
 		synchronized (lock) {
 			this.store.addMessagesToGroup(correlationKey, message);
 		}
-		if (logger.isDebugEnabled()) {
-			logger.debug(String.format("Handled message for key [%s]: %s.", correlationKey, message));
-		}
+		logger.debug(() -> String.format("Handled message for key [%s]: %s.", correlationKey, message));
```
artembilan, line 96: There is one more in this class using `String.format()`.
I see your general comment about formats support for logging API.
I think we had discussion before that it is better to keep that `LogAccessor` contract as close to other logging frameworks as possible and don't overhead it with something what could cause confuse for target users. I mean that one `debugFormat()` and similar.
The point of this optimization on our side to avoid code flow extra complexity when we remove those `if (logger.isDebugEnabled()) {`.
On the other hand the point of `Supplier<? extends CharSequence>` contract to avoid an overhead for the final strings creation when we don't need to log that message.

See `LogMessage.format()` for that purpose otherwise: ` * Build a lazily formatted message from the given format string and argument.`

I think even this sentence could be reworked to that "lazy formatted" one instead of `Supplier` wrapper around standard `String.format()`...

spring-integration-core/src/main/java/org/springframework/integration/endpoint/ReactiveStreamsConsumer.java
```diff
@@ -89,8 +89,8 @@ public ReactiveStreamsConsumer(MessageChannel inputChannel, Subscriber<Message<?
 		Assert.notNull(subscriber, "'subscriber' must not be null");
 		this.inputChannel = inputChannel;
 
-		if (inputChannel instanceof NullChannel && logger.isWarnEnabled()) {
-			logger.warn("The consuming from the NullChannel does not have any effects: " +
+		if (inputChannel instanceof NullChannel) {
+			logger.warn(() -> "The consuming from the NullChannel does not have any effects: " +
```
artembilan, line 93: We don't need `Supplier` wrapper here: the message is just plain string constant.
The target logging system still checks its `enabled` before printing the message.
And yes: that `isWarnEnabled()` was really an overhead 😄

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -384,8 +384,8 @@ public void afterPropertiesSet() {
 			this.messageRowMapper = new MessageRowMapper(this.deserializer, this.lobHandler);
 		}
 
-		if (this.jdbcTemplate.getFetchSize() != 1 && LOGGER.isWarnEnabled()) {
-			LOGGER.warn("The jdbcTemplate's fetch size is not 1. This may cause FIFO issues with Oracle databases.");
+		if (this.jdbcTemplate.getFetchSize() != 1) {
+			LOGGER.warn(() -> "The jdbcTemplate's fetch size is not 1. This may cause FIFO issues with Oracle databases.");
```
artembilan, line 388: Don't need `Supplier` wrapper over here - just plain string constant.

Would you mind to revise all other log messages in this class in favor of discovered that `LogMessage.format()`?

artembilan (changes requested):
Thank you for contribution!
This is really great change.

We really can merge this as is, but still would like to share my thoughts on the matter in the review.

Also: let us know if that doesn't bother you to add your official name to all the affected classes.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
trungPa: I get the idea but I'm not really sure I'm doing it right. Please take a look.
The `LogMessage` implements CharSequence so I'm passing `LogMessage`s to `LogAccessor.debug(CharSequence)` for lazily formatting instead of wrapping them in `Supplier`s.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/ReactiveStreamsConsumer.java
trungPa: Ok

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
trungPa: Ok

artembilan: There is no reason to comment on the review if you agree with the proposal.
Otherwise I get email for every single your `Ok` 😄

artembilan (changes requested):
Hm. I reviewed this a couple days ago, but looks like I didn't press "Submit review". 🤦

artembilan:
@trungPa ,

PR #3593, opened 2021-07-19
GH-3591: Add `gateway()` section into `dsl.adoc`

### LINKED ISSUE #3591: Please add a section about .gateway() to the online DSL Documentation
**Expected Behavior**

https://docs.spring.io/spring-integration/reference/html/dsl.html#java-dsl-basics mentions `.gateway()` and possibly https://docs.spring.io/spring-integration/reference/html/dsl.html also has its own section about Gateways, e.g. https://docs.spring.io/spring-integration/reference/html/dsl.html#java-dsl-gateways

**Current Behavior**

There is no mention of `.gateway()` and no use of it in the `spring-integration-saples` code either. It's easy to completely miss that this endpoint exists in the DSL

**Context**

On Stackoverflow, @artembilan suggested I post this GH request. 
https://stackoverflow.com/questions/68400511/spring-integration-dsl-how-to-refactor-into-subflows#comment120898078_68402954

It took me way to long to realize that Spring Integration DSL hat a .gateway() method available when .gateway() was exactly what I needed.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3591

### DIFF
--- a/src/reference/asciidoc/dsl.adoc
+++ b/src/reference/asciidoc/dsl.adoc
@@ -602,17 +602,17 @@ Also see <<java-dsl-class-cast>>.
 [[java-dsl-gateway]]
 === Operator gateway()
 
-The `gateway()` operator of `IntegrationFlow` definition is a special service activator implementation to call some other endpoint or integration flow via their input channel and wait for reply from there.
-Technically it plays the same role as a nested `<gateway>` component for `<chain>` definition (see <<./chain.adoc#chain-gateway,Calling a Chain from within a Chain>>) and allows to make a code flow much cleaner and more straightforward.
-And logically and from business perspective it is a messaging gateway to let to distribute and reuse functionality between different parts of the target integration solution (see <<./gateway.adoc#gateway,Messaging Gateways>>).
+The `gateway()` operator in an `IntegrationFlow` definition is a special service activator implementation, to call some other endpoint or integration flow via its input channel and wait for reply.
+Technically it plays the same role as a nested `<gateway>` component in a `<chain>` definition (see <<./chain.adoc#chain-gateway,Calling a Chain from within a Chain>>) and allows a flow  to be cleaner and more straightforward.
+Logically, and from business perspective, it is a messaging gateway to allow the distribution and reuse of functionality between different parts of the target integration solution (see <<./gateway.adoc#gateway,Messaging Gateways>>).
 This operator has several overloads for different goals:
 
-- `gateway(String requestChannel)` to call some endpoint's input channel by its name;
-- `gateway(MessageChannel requestChannel)` to call some endpoint's input channel by its direct injection;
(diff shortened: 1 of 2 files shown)


src/reference/asciidoc/dsl.adoc
```diff
@@ -599,6 +599,41 @@ public IntegrationFlow integerFlow() {
...
+And logically and from business perspective it is a messaging gateway to let to distribute and reuse functionality between different parts of the target integration solution (see <<./gateway.adoc#gateway,Messaging Gateways>>).
+This operator has several overloads for different goals:
+
+- `gateway(String requestChannel)` to call some endpoint's input channel by its name;
+- `gateway(MessageChannel requestChannel)` to call some endpoint's input channel by its direct injection;
+- `gateway(IntegrationFlow flow)` to call the input channel of the provided `IntegrationFlow`.
+
+All of them have a variant with the second `Consumer<GatewayEndpointSpec>` argument to configure the target `GatewayMessageHandler` and respective `AbstractEndpoint`.
+Also the `IntegrationFlow`-based has a power that it allows to call existing `IntegrationFlow` bean or declare the flow as a sub-flow via an in-place lambda for an `IntegrationFlow` functional interface or have it extracted in a `private` method cleaner code style:
```
garyrussell, line 615:
```suggestion
The `gateway()` operator in an `IntegrationFlow` definition is a special service activator implementation, to call some other endpoint or integration flow via its input channel and wait for reply.
Technically it plays the same role as a nested `<gateway>` component in a `<chain>` definition (see <<./chain.adoc#chain-gateway,Calling a Chain from within a Chain>>) and allows a flow  to be cleaner and more straightforward.
Logically, and from business perspective, it is a messaging gateway to allow the distribution and reuse of functionality between different parts of the target integration solution (see <<./gateway.adoc#gateway,Messaging Gateways>>).
This operator has several overloads for different goals:

- `gateway(String requestChannel)` to send a message to some endpoint's input channel by its name;
- `gateway(MessageChannel requestChannel)` to send a message to some endpoint's input channel by its direct injection;
- `gateway(IntegrationFlow flow)` to send a message to the input channel of the provided `IntegrationFlow`.

All of these have a variant with the second `Consumer<GatewayEndpointSpec>` argument to configure the target `GatewayMessageHandler` and respective `AbstractEndpoint`.
Also the `IntegrationFlow`-based methods allows calling existing `IntegrationFlow` bean or declare the flow as a sub-flow via an in-place lambda for an `IntegrationFlow` functional interface or have it extracted in a `private` method cleaner code style:

IMPORTANT: If the downstream flow does not always return a reply, you should set the `requestTimeout` to 0 to prevent hanging the calling thread indefinitely.
In that case, the flow will end at that point and the thread released for further work.
```

garyrussell (changes requested):

PR #3594, opened 2021-07-19
Fix pattern how `Lock.unlock()` is used

artembilan:
Related to https://build.spring.io/browse/INT-MAIN-84/

The `lock.unlock()` must be called in the `finally` block of the
nested `try..catch`, not in the outer which may just fail on `lock.lockInterruptibly()`
in which case there is just not going to be anything we can `unlock()` in the end


garyrussell (approved):

garyrussell:
...and cherry-picked.

PR #3599, opened 2021-07-27
GH-3549: Clean up SonarQube smells

### LINKED ISSUE #3549: Fix minor SonarQube smells
See Sonar report for the current `main`: https://sonar.spring.io/project/issues?id=org.springframework.integration%3Aspring-integration%3Amain&resolved=false&severities=MINOR&types=CODE_SMELL

Mostly those smells are about a `Magic Number` rule violation.

In some places we indeed need to extract a constant for such a default value, but in other places it really would be better just to place a `// NOSONAR` comment after the code line to ignore it from Sonar analysis.

trgpa:
Fixes https://github.com/spring-projects/spring-integration/issues/3549


artembilan:
Thank you, @trungPa !
As usual: looking for more contributions!

PR #3602, opened 2021-07-28
GH-3600: Fix Kafka Tx Synchronization

### LINKED ISSUE #3600: Kafka MessageHandler Does Not Synchronize with a non-Kafka TM for Producer-Only Transactions
`TransactionSynchronizationManager.getResource()` returns `null` when there is an existing synchronizing transaction, so we run in a local transaction.

https://stackoverflow.com/questions/68460690/wrapping-streambridge-send-and-jpa-save-inside-a-transaction/68461298#68461298

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3600

Previously, the `KafkaProducerMessageHandler` did not synchronize a transaction
with some other transaction (e.g. JDBC); it published in a local transaction instead.

Also configure the Gradle Kotlin Daemon to work with JDK 16.

**cherry-pick to 5.4.x**


artembilan:
... and cherry-picked to `5.4.x` after fixing conflict in the `gradle.properties`

Issue #3584, opened 2021-06-23
Add Support for Spring AMQP 2.4


garyrussell:
Make spring-integration-amqp compatible with both spring-amqp 2.3.x and 2.4.x.

artembilan:
Fixed via: https://github.com/spring-projects/spring-integration/commit/76f77ccd2c5f94e000495357e66c4f99fffc6010

ae-govau:
Since updating to Spring Boot 2.5.3 (which bumps 5.5.1 to 5.5.2) one of our apps fails to start and report this error message:

```
Description:
An attempt was made to call a method that does not exist. The attempt was made from the following location:
     org.springframework.cloud.stream.binder.rabbit.RabbitMessageChannelBinder.createConsumerEndpoint(RabbitMessageChannelBinder.java:517)
The following method did not exist:
     'void org.springframework.integration.amqp.inbound.AmqpInboundChannelAdapter.<init>(org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer)'
```

I suspect that's related to this change?

I note the following related JARs in our app, each of which appear to be latest.

```
BOOT-INF/lib/spring-integration-amqp-5.5.2.jar - contains AmqpInboundChannelAdapter.class
BOOT-INF/lib/spring-cloud-stream-binder-rabbit-3.1.3.jar - contains RabbitMessageChannelBinder.class
```

I find the layout of the Spring projects quite complex so I'm not sure where best to report this issue, but figured I'd start here since I think this likely introduced the problem.

artembilan:
That’s correct : we did it here. We just missed the fact that Rabbit Binder is not going to release until September as a part of Spring Cloud release. See this issue for workaround for time being : https://github.com/spring-projects/spring-integration/issues/3601. Sorry for inconvenience

ae-govau:
Thanks @artembilan :+1:

PR #3607, opened 2021-07-30
GH-3601: Bring AMQP byte code compatibility back

### LINKED ISSUE #3601: AmqpInboundChannelAdapter.<init>(Lorg/springframework/amqp/rabbit/listener/AbstractMessageListenerContainer;)V
When the spring-boot-starter-parent version is 2.5.3, the introduction of spring-cloud-starter-bus-amqp will lead to this problem. The constructor of AmqpInboundChannelAdapter exists in the version 5.5.1 spring-integration-amqp , but not exist in the version 5.5.2. Who can correct it? Thank you

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3601

The issue https://github.com/spring-projects/spring-integration/issues/3584
has introduced a regression when old constructor with an
`AbstractMessageListenerContainer` was removed in favor of just
`MessageListenerContainer`.
But with that change all the dependant projects must be recompiled,
which is not a case when Spring Cloud was not released against the
latest Spring Boot.

**Cherry-pick to `5.4.x`**

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
@@ -123,6 +123,11 @@ public class AmqpInboundChannelAdapter extends MessageProducerSupport implements
 
 	private BatchMode batchMode = BatchMode.MESSAGES;
 
+	// TODO Remove in 6.0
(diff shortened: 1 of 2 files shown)


artembilan:
Tested with existing RabbitMQ Binder after generating project via start.spring.io and changing to this `<spring-integration.version>5.5.3-SNAPSHOT</spring-integration.version>` against locally installed artifacts.

garyrussell:
...and cherry-picked.

PR #3611, opened 2021-08-06
Fix to address expectation of proper exception

### LINKED ISSUE #3610: WebFluxRequestExecutingMessageHandler directly creates WebClientResponseException rather than using provided create(...) factory method

5.3.8.RELEASE

**Describe the bug**

The `WebFluxRequestExecutingMessageHandler` is directly creating a `WebClientResponseException` rather than using the provided `create` factory methods. This can cause some confusion when attempting to use the `customizeMonoReply` method to trap and provide a default response payload.

**To Reproduce**

Assume a RESTful endpoint that will return an HTTP/404 error when an entity is not found.

Using an integration flow similar to the following:

```
@Bean
public IntegrationFlow getMixEntryFlowReactive() {
    final WebClient webClient =
                        webClientBuilder.defaultHeaders(
                                        headers -> headers.setBasicAuth("api", "s3cr3t")
                                    )
                                    .build();
    final String defaultPayload = "{ \"id\": -1, \"name\": null }";
    return f->
            f.enrich(
                e -> e.requestSubFlow(
                    sf -> sf.handle(
                        WebFlux.outboundGateway("http://localhost:8080/mix-entry/name/{mixEntryName}", webClient)
                        .httpMethod(HttpMethod.GET)
                        .uriVariable("mixEntryName", "payload")
                        .expectedResponseType(String.class)
                        ,
                        ec -> ec.customizeMonoReply(
                            (Message<?> message, Mono<?> mono) -> 
                                mono.onErrorResume(
                                    WebClientResponseException.NotFound.class,
                                    ex1 -> Mono.just(ex1)
                                                .map(ex -> defaultPayload)
                                                .switchIfEmpty((Mono)mono)
                                )
                        )
                    )
                    .logAndReply()
                )
                .headerExpression("mix-entry-id", "#jsonPath(payload, '$.id')")
        )
        .logAndReply();
}
```

**Expected behavior**

Expectation is that resulting message, when an HTTP/404 error occurs, contains the header `mix-entry-id` with the value of `-1`. 

**Sample**

https://github.com/djgraff209/so68637283

djgraff209:
This fix changes the WebFluxRequestExecutingMessageHandler
The change specifically changes the contruction of the
WebClientResponseException to use the `create` factory method.

Fixes #3610.

### DIFF
--- a/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
+++ b/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
@@ -58,6 +58,7 @@ import reactor.core.publisher.Mono;
  * @author Shiliang Li
  * @author Artem Bilan
  * @author Gary Russell
+ * @author David Graff
  *
  * @since 5.0
  *
--- a/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
+++ b/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/outbound/WebFluxRequestExecutingMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 the original author or authors.
+ * Copyright 2017-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Please, fix it as we discussed on Gitter: https://gitter.im/spring-projects/spring-integration?at=611069649484630efa2a4c46.

Also consider to cover it somehow with the test. 
See https://github.com/spring-projects/spring-integration/pull/3612 for the place where you could inject the logic for your change.

Thank you!

djgraff209:
Will attempt to provide unit test to cover later today.

artembilan (approved):
LGTM.
Merging after adding your name to `@author` list.

artembilan:
Hm.
I meant that you are OK. That's already my responsibility to clean the fix for our standards, but since you have beaten me with that new commit I merge it without my hands 😄 .

The general rule if I request changes, you have to do them. Otherwise that means everything is up to me and you are good.

So, thank you for your patience with that nasty byte buffer issue and contribution.
Looking forward for more!

djgraff209:
Absolutely not a problem. I have been in the corporate world for many years and rigor like I'm trying to follow is meant mainly to reduce the burden on the owners.

Thanks again!

artembilan:

PR #3638, opened 2021-09-30
GH-3637: Fix KafkaMessageSource First Poll

### LINKED ISSUE #3637: KafkaMessageSource Blocks on Initial Poll When No Records
5.5.4.RELEASE

KMS increases the poll timeout for the first poll (assignment).

```java
this.assignTimeout =
		Duration.ofMillis(Math.max(this.pollTimeout.toMillis() * 20, MIN_ASSIGN_TIMEOUT)); // NOSONAR - magic
```

With the default `pollTimeout` (5s), when there are no records, the poll blocks for 100s.

This causes a delay, for example, with a `PollableConsumer` in SCSt the poll is not responsive.

Wake the consumer when the partitions are assigned.

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3637

If there are no records to receive, the poll blocked for `20*pollTimeout`.

Wake the consumer during partition assignment; however, this changes the behavior
when there are records present - the first poll always returns no records.

Detect that the wakeup was due to assignment and perform another poll.

Also add some debug logging, and don't create a new consumer if the source was
stopped.

**cherry-pick to 5.4.x**


artembilan:
... and cherry-picked to `5.4.x`

PR #3639, opened 2021-10-04
GH-3432: Add MQTT v5 channel adapters

### LINKED ISSUE #3432: spring-integration-mqtt: Any plan to integrate with org.eclipse.paho.mqttv5.client
**Expected Behavior**

Spring-integration-mqtt supports mqtt protocol version 5. (Currently, it only support mqtt protocol version 3)

**Current Behavior**

Latest `spring-integration-mqtt` still support mqtt version 3 by integrating with 

```
<dependency>
<groupId>org.eclipse.paho</groupId>
<artifactId>org.eclipse.paho.client.mqttv3</artifactId>
<version>1.2.4</version>
<scope>compile</scope>
</dependency>
```
We have a project which requires a Spring Boot app integrate with mqttv5 client. 
```
        <dependency>
            <groupId>org.eclipse.paho</groupId>
            <artifactId>org.eclipse.paho.mqttv5.client</artifactId>
            <version>1.2.5</version>
        </dependency>
```

I would like to check if we have the plan to have new release of `spring-integration-mqtt` which is able to support mqtt version 5.

How has this issue affected you? our team will develop a springboot app which need to support mqtt version 5. However, the current spring-integration-mqtt dependency only support mqtt version 3.
What are you trying to accomplish?  We would like to have a new release of spring-integration-mqtt which is able to support mqtt version 5.
Are you aware of any workarounds? I have two thoughts: 1. Write a full functional mqtt version 5 client, but it need to handle all connect, reconnect, disconnet and other features. 2. Write a simplified `spring-integration-mqtt` dependency which integrate with `org.eclipse.paho.mqttv5.client` 

I am looking forward to hearing from spring team soon.

Thanks

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3432

* Add `optional` dependency for `org.eclipse.paho:org.eclipse.paho.mqttv5.client`
* Add `MqttProtocolErrorEvent` and emit it from the `mqttErrorOccurred()` callback of the MQTT v5 client
* Add `MqttHeaderMapper` since MQTT v5 has introduced user properties pair to transfer over the protocol
* Add `Mqttv5PahoMessageHandler` as one more extension of the `AbstractMqttMessageHandler`
* Add more convenient `MqttHeaders` constants for easier headers mapping configuration
* Ensure via `Mqttv5BackToBackTests` that MQTT v5 is supported by the provided components
* Change `pr-build-workflow.yml` to use `eclipse-mosquitto` container for testing all the MQTT interactions
* Change `cyrilix/rabbitmq-mqtt` service to the `rabbitmq:management` since RabbitMQ does not support MQTT v5

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
@@ -21,9 +21,10 @@ import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.core.log.LogMessage;
 import org.springframework.integration.endpoint.MessageProducerSupport;
-import org.springframework.integration.mqtt.support.DefaultPahoMessageConverter;
 import org.springframework.integration.mqtt.support.MqttMessageConverter;
 import org.springframework.integration.support.management.IntegrationManagedResource;
 import org.springframework.jmx.export.annotation.ManagedAttribute;
@@ -45,7 +46,13 @@ import org.springframework.util.Assert;
  */
 @ManagedResource
 @IntegrationManagedResource
-public abstract class AbstractMqttMessageDrivenChannelAdapter extends MessageProducerSupport {
+public abstract class AbstractMqttMessageDrivenChannelAdapter extends MessageProducerSupport
+		implements ApplicationEventPublisherAware {
+
+	/**
+	 * The default completion timeout in milliseconds.
+	 */
+	public static final long DEFAULT_COMPLETION_TIMEOUT = 30_000L;
 
 	private final String url;
 
@@ -53,7 +60,13 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter extends MessagePro
 
 	private final Set<Topic> topics;
 
-	private volatile MqttMessageConverter converter;
+	private long completionTimeout = DEFAULT_COMPLETION_TIMEOUT;
+
+	private boolean manualAcks;
+
+	private ApplicationEventPublisher applicationEventPublisher;
+
+	private MqttMessageConverter converter;
 
 	protected final Lock topicLock = new ReentrantLock(); // NOSONAR
 
@@ -147,6 +160,42 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter extends MessagePro
 		return "mqtt:inbound-channel-adapter";
 	}
 
+	@Override
+	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
+		this.applicationEventPublisher = applicationEventPublisher; // NOSONAR (inconsistent synchronization)
+	}
+
+	protected ApplicationEventPublisher getApplicationEventPublisher() {
+		return this.applicationEventPublisher;
+	}
+
+	/**
+	 * Set the acknowledgment mode to manual.
+	 * @param manualAcks true for manual acks.
+	 * @since 5.3
+	 */
+	public void setManualAcks(boolean manualAcks) {
+		this.manualAcks = manualAcks;
+	}
+
+	protected boolean isManualAcks() {
+		return this.manualAcks;
+	}
+
+	/**
+	 * Set the completion timeout for operations. Not settable using the namespace.
+	 * Default {@value #DEFAULT_COMPLETION_TIMEOUT} milliseconds.
+	 * @param completionTimeout The timeout.
+	 * @since 4.1
+	 */
+	public synchronized void setCompletionTimeout(long completionTimeout) {
+		this.completionTimeout = completionTimeout;
+	}
+
+	protected long getCompletionTimeout() {
+		return this.completionTimeout;
+	}
+
 	/**
 	 * Add a topic to the subscribed list.
 	 * @param topic The topic.
@@ -238,18 +287,6 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter extends MessagePro
 			this.topicLock.unlock();
 		}
 	}
-
-	@Override
-	protected void onInit() {
-		super.onInit();
-		if (this.converter == null) {
-			DefaultPahoMessageConverter pahoMessageConverter = new DefaultPahoMessageConverter();
-			pahoMessageConverter.setBeanFactory(getBeanFactory());
-			this.converter = pahoMessageConverter;
-
-		}
-	}
-
 
 	/**
 	 * @since 4.1
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
@@ -188,7 +188,7 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter extends MessagePro
 	 * @param completionTimeout The timeout.
 	 * @since 4.1
 	 */
-	public synchronized void setCompletionTimeout(long completionTimeout) {
+	public void setCompletionTimeout(long completionTimeout) {
 		this.completionTimeout = completionTimeout;
 	}
 
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -1,0 +1,328 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.mqtt.inbound;
+
+import java.util.Arrays;
+import java.util.Map;
+
+import org.eclipse.paho.mqttv5.client.IMqttAsyncClient;
+import org.eclipse.paho.mqttv5.client.IMqttToken;
+import org.eclipse.paho.mqttv5.client.MqttAsyncClient;
+import org.eclipse.paho.mqttv5.client.MqttCallback;
+import org.eclipse.paho.mqttv5.client.MqttConnectionOptions;
+import org.eclipse.paho.mqttv5.client.MqttDisconnectResponse;
+import org.eclipse.paho.mqttv5.common.MqttException;
+import org.eclipse.paho.mqttv5.common.MqttMessage;
+import org.eclipse.paho.mqttv5.common.packet.MqttProperties;
+
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.integration.acks.SimpleAcknowledgment;
+import org.springframework.integration.context.IntegrationContextUtils;
+import org.springframework.integration.mapping.HeaderMapper;
+import org.springframework.integration.mqtt.core.MqttComponent;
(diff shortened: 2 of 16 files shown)


artembilan:
If this is OK, I'll go ahead with Doc.

NOTE: RabbitMQ does not support MQTT v5, so I had to go with Mosquitto for both protocols testing.
We probably need to do something on the CI server to avoid problems with not supported protocol.

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -147,6 +160,42 @@ public String getComponentType() {
...
+	 * Set the completion timeout for operations. Not settable using the namespace.
+	 * Default {@value #DEFAULT_COMPLETION_TIMEOUT} milliseconds.
+	 * @param completionTimeout The timeout.
+	 * @since 4.1
+	 */
+	public synchronized void setCompletionTimeout(long completionTimeout) {
+		this.completionTimeout = completionTimeout;
+	}
+
+	protected long getCompletionTimeout() {
```
garyrussell, line 195: Sync to avoid sonar smell. However, setter may not need sync now that it is obtained via the getter - the sync was added to squash a Sonar issue.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -0,0 +1,343 @@
...
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.MessagingException;
+import org.springframework.messaging.converter.SmartMessageConverter;
+import org.springframework.messaging.support.GenericMessage;
+import org.springframework.util.Assert;
+
+/**
+ * The {@link AbstractMqttMessageDrivenChannelAdapter} implementation for MQTT v5.
+ *
+ * The {@link MqttProperties} are mapped via provided {@link HeaderMapper};
```
garyrussell, line 57:
```suggestion
 * The {@link MqttProperties} are mapped via the provided {@link HeaderMapper};
```

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackTests.java
```diff
@@ -0,0 +1,180 @@
...
+import org.springframework.integration.mqtt.support.MqttHeaders;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.PollableChannel;
+import org.springframework.messaging.converter.AbstractMessageConverter;
+import org.springframework.messaging.converter.SmartMessageConverter;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit4.SpringRunner;
```
garyrussell, line 55: JUnit4 for a new test?

garyrussell (changes requested):
LGTM; you could use TestContainers with mosquitto for the tests 

https://stackoverflow.com/questions/51383126/mqtt-client-testing-with-junit/51384651

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackTests.java
artembilan: Right., because our `BrokerRunning` is JUnit 4 TestWatcher.

I don't want to introduce TestContainers into the project yet...

garyrussell: Right, but this won't work on CI because it will fail (RabbitMQ plugin detected).

artembilan:
If this is OK, I can go ahead and change all the MQTT tests to use that `MosquittoContainerTest` aspect instead of JUnit 4 rule.
The Mosquitto easily supports all the MQTT protocol.
The we can disable respective RabbitMQ plugin everywhere.

NOTE: to reuse container you should add `testcontainers.reuse.enable=true` into your `~/.testcontainers.properties`.
Otherwise it is closed after test class and emits plain error that `reuse` is not enabled.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/MosquittoContainerTest.java
```diff
@@ -0,0 +1,39 @@
...
+ * @author Artem Bilan
+ *
+ * @since 5.5.5
+ */
+@Testcontainers(disabledWithoutDocker = true)
+public interface MosquittoContainerTest {
+
+	@Container
+	GenericContainer<?> MOSQUITTO_CONTAINER =
+			new GenericContainer<>("eclipse-mosquitto:2.0.12")
```
garyrussell, line 34: Why not `latest`?

src/reference/asciidoc/mqtt.adoc
```diff
@@ -421,4 +424,74 @@ MqttConnectOptions options = source.getConnectionInfo();
 [[mqtt-v5]]
 === MQTT v5 Support
 
-TODO
\ No newline at end of file
+Starting with version 5.5.5, `spring-integration-mqtt` module provides channel adapter implementations for MQTT v5 protocol.
+The `org.eclipse.paho:org.eclipse.paho.mqttv5.client` is `optional` dependency, so has to be included explicitly in the target project.
```
garyrussell, line 428:
```suggestion
Starting with version 5.5.5, the `spring-integration-mqtt` module provides channel adapter implementations for the MQTT v5 protocol.
The `org.eclipse.paho:org.eclipse.paho.mqttv5.client` is an `optional` dependency, so has to be included explicitly in the target project.
```

src/reference/asciidoc/mqtt.adoc
```diff
@@ -421,4 +424,74 @@ MqttConnectOptions options = source.getConnectionInfo();
...
+
+Since MQTT v5 protocol supports extra arbitrary properties in MQTT message, the `MqttHeaderMapper` implementation has been introduced to map to/from headers on publish and receive operations.
+By default (`*` pattern) it maps all the received `PUBLISH` frame properties (including user properties).
+On the outbound side it maps these subset of headers for `PUBLISH` frame: `contentType`, `mqtt_messageExpiryInterval`, `mqtt_responseTopic`, `mqtt_correlationData`.
```
garyrussell, line 432:
```suggestion
Since the MQTT v5 protocol supports extra arbitrary properties in an MQTT message, the `MqttHeaderMapper` implementation has been introduced to map to/from headers on publish and receive operations.
By default (via the `*` pattern) it maps all the received `PUBLISH` frame properties (including user properties).
On the outbound side it maps this subset of headers for `PUBLISH` frame: `contentType`, `mqtt_messageExpiryInterval`, `mqtt_responseTopic`, `mqtt_correlationData`.
```

src/reference/asciidoc/mqtt.adoc
```diff
@@ -421,4 +424,74 @@ MqttConnectOptions options = source.getConnectionInfo();
...
+
+The outbound channel adapter for MQTT v5 protocol is present as an `Mqttv5PahoMessageHandler`.
+It requires a `clientId` and MQTT broker URL or `MqttConnectionOptions` reference.
+It does support a `MqttClientPersistence` option, can be `async` and can emit `MqttIntegrationEvent` objects in that case (see `asyncEvents` option).
+If request message payload is an `org.eclipse.paho.mqttv5.common.MqttMessage`, it is published as is via internal `IMqttAsyncClient`.
+If the payload is `byte[]` it is used as is for target `MqttMessage` payload to publish.
+If the payload is a `String` it is converted to `byte[]` to publish.
+The rest use-cases are delegated to the provided `MessageConverter` which is a `IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME` `ConfigurableCompositeMessageConverter` bean from the application context.
+Note: the provided `HeaderMapper<MqttProperties>` is not used when the requested message payload is already an `MqttMessage`.
+The following Java DSL configuration sample demonstrates how to use this channel adapter in the integration flow:
```
garyrussell, line 442:
```suggestion
The outbound channel adapter for the MQTT v5 protocol is present as an `Mqttv5PahoMessageHandler`.
It requires a `clientId` and MQTT broker URL or `MqttConnectionOptions` reference.
It supports a `MqttClientPersistence` option, can be `async` and can emit `MqttIntegrationEvent` objects in that case (see `asyncEvents` option).
If a request message payload is an `org.eclipse.paho.mqttv5.common.MqttMessage`, it is published as is via the internal `IMqttAsyncClient`.
If the payload is `byte[]` it is used as is for the target `MqttMessage` payload to publish.
If the payload is a `String` it is converted to `byte[]` to publish.
The remaining use-cases are delegated to the provided `MessageConverter` which is a `IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME` `ConfigurableCompositeMessageConverter` bean from the application context.
Note: the provided `HeaderMapper<MqttProperties>` is not used when the requested message payload is already an `MqttMessage`.
The following Java DSL configuration sample demonstrates how to use this channel adapter in the integration flow:
```

src/reference/asciidoc/mqtt.adoc
```diff
@@ -421,4 +424,74 @@ MqttConnectOptions options = source.getConnectionInfo();
...
+The inbound channel adapter for MQTT v5 protocol is present as an `Mqttv5PahoMessageDrivenChannelAdapter`.
+It requires a `clientId` and MQTT broker URL or `MqttConnectionOptions` reference, plus topics to subscribe and consume.
+It does support a `MqttClientPersistence` option, which is in-memory by default.
+The expected `payloadType` (`byte[]` by default) can be configured and it is propagated to the provided `SmartMessageConverter` for conversion from `byte[]` of the received `MqttMessage`.
+If the `manualAck` option is set, then an `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` header is added to the message to produce as an instance of `SimpleAcknowledgment`.
+The `HeaderMapper<MqttProperties>` is used to map `PUBLISH` frame properties (including user properties) into the target message headers.
+Standard `MqttMessage` properties, such as `qos`, `id`, `dup`, `retained`, plus received topic are always mapped to headers.
+See `MqttHeaders` for more information.
+
+The following Java DSL configuration sample demonstrates how to use this channel adapter in the integration flow:
```
garyrussell, line 475:
```suggestion
IMPORTANT: The `org.springframework.integration.mqtt.support.MqttMessageConverter` cannot be used with the `Mqttv5PahoMessageHandler` since its contract is aimed only for the MQTT v3 protocol.

See more information in the `Mqttv5PahoMessageHandler` javadocs and its superclass.

The inbound channel adapter for the MQTT v5 protocol is present as an `Mqttv5PahoMessageDrivenChannelAdapter`.
It requires a `clientId` and MQTT broker URL or `MqttConnectionOptions` reference, plus topics to which to subscribe and consume from.
It supports a `MqttClientPersistence` option, which is in-memory by default.
The expected `payloadType` (`byte[]` by default) can be configured and it is propagated to the provided `SmartMessageConverter` for conversion from `byte[]` of the received `MqttMessage`.
If the `manualAck` option is set, then an `IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK` header is added to the message to produce as an instance of `SimpleAcknowledgment`.
The `HeaderMapper<MqttProperties>` is used to map `PUBLISH` frame properties (including user properties) into the target message headers.
Standard `MqttMessage` properties, such as `qos`, `id`, `dup`, `retained`, plus received topic are always mapped to headers.
See `MqttHeaders` for more information.

The following Java DSL configuration sample demonstrates how to use this channel adapter in the integration flow:
```

src/reference/asciidoc/mqtt.adoc
```diff
@@ -421,4 +424,74 @@ MqttConnectOptions options = source.getConnectionInfo();
...
+    messageProducer.setManualAcks(true);
+
+    return IntegrationFlows.from(messageProducer)
+            .channel(c -> c.queue("fromMqttChannel"))
+            .get();
+}
+----
+====
+
+IMPORTANT: The `org.springframework.integration.mqtt.support.MqttMessageConverter` cannot be used with the `Mqttv5PahoMessageDrivenChannelAdapter` since its contract is aimed only for MQTT v3 protocol.
```
garyrussell, line 495:
```suggestion
IMPORTANT: The `org.springframework.integration.mqtt.support.MqttMessageConverter` cannot be used with the `Mqttv5PahoMessageDrivenChannelAdapter` since its contract is aimed only for the MQTT v3 protocol.
```

garyrussell (changes requested):
I would say, yes, go ahead with the other tests, and deprecate `BrokerRunning`.

artembilan:
> I would say, yes, go ahead with the other tests

Let's merge this first, then I'll revise the rest of test in the separate commit not bothering you with review!

Anyway let me know if all of this works well for you locally, before we merge and "break" other tests!

PR #3650, opened 2021-10-20
GH-3648: Fix @Gateway.payloadExpression

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3648

When configuring a gateway proxy with XML, but specifying the payload expression
on the method `@Gateway` annotation, the expression was ignored, even though it
had been parsed.

`@Payload` worked.

With this change, if both `@Payload` and `@Gateway` are defined on a gateway method,
`@Gateway.payloadExpression` wins.

**cherry-pick to 5.4.x, 5.3.x**


artembilan:
... and cherry-picked to `5.4.x` & `5.3.x`.

PR #3657, opened 2021-11-01
Add `toReactivePublisher(autoStartOnSubscribe)`

artembilan:
The `IntegrationFlowBuilder.toReactivePublisher()` returns a `Publisher<Message<?>>`
which may be subscribed somewhere late in the application logic, e.g.
when WebSocket (or RSocket) subscription is initiated by the external client.
In between application context startup and that subscription moment, the `IntegrationFlow`
must not try to produce messages since there is nothing to consumer them from the
`Publisher<Message<?>>` side.
One of the way is to have a source endpoint not started automatically and control its
lifecycle from the point fo reactive subscription

* Introduce an `IntegrationFlowBuilder.toReactivePublisher(boolean autoStartOnSubscribe)`
to let the framework do a job for an `IntegrationFlow` lifecycle control.
This way end-user doesn't need to know autowire a starting endpoint and use `doOnSubscribe()`
and similar callbacks
* Change `ConsumerEndpointFactoryBean` log message about a `FixedSubscriberChannel`
to `INFO` since an `autoStartup = false` really does not have any effect and there is nothing
for end-user to worry about.
The `IntegrationFlow` knows nothing about each endpoint internals and cannot control which
mark as `autoStartup = false` and which not

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
@@ -2940,8 +2940,8 @@ public abstract class BaseIntegrationFlowDefinition<B extends BaseIntegrationFlo
 
 	/**
 	 * Represent an Integration Flow as a Reactive Streams {@link Publisher} bean.
-	 * @param autoStartOnSubscribe start a production and consumption in the flow,
-	 * when subscription to the publisher is initiated.
+	 * @param autoStartOnSubscribe start message production and consumption in the flow,
+	 * when a subscription to the publisher is initiated.
 	 * If this set to true, the flow is marked to not start automatically by the application context.
 	 * @param <T> the expected {@code payload} type
 	 * @return the Reactive Streams {@link Publisher}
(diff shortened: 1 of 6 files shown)


artembilan:
See related SO: https://stackoverflow.com/questions/69769650/java-spring-mqtt-subscriber-toreactivepublisher

spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
```diff
@@ -2934,8 +2934,21 @@ public IntegrationFlow to(IntegrationFlow other) {
...
-	@SuppressWarnings(UNCHECKED)
 	protected <T> Publisher<Message<T>> toReactivePublisher() {
+		return toReactivePublisher(false);
+	}
+
+	/**
+	 * Represent an Integration Flow as a Reactive Streams {@link Publisher} bean.
+	 * @param autoStartOnSubscribe start a production and consumption in the flow,
+	 * when subscription to the publisher is initiated.
+	 * If this set to true, the flow is marked to not start automatically by the application context.
```
garyrussell, line 2945:
```suggestion
	 * @param autoStartOnSubscribe start message production and consumption in the flow,
	 * when a subscription to the publisher is initiated.
	 * If this set to true, the flow is marked to not start automatically by the application context.
```

garyrussell (changes requested):
Minor javadoc polishing.

garyrussell (approved):

PR #3659, opened 2021-11-02
GH-3652 Do not try to delete non-existing file in FileExistMode.REPLACE

### LINKED ISSUE #3652: "Failed to delete" warning reported for each non-replaced file by SftpOutboundGateway
Using Spring Integration 5.4.11

**Describe the bug**

When you setup a SftpOutboundGateway with a `FileExistsMode` set to `REPLACE` you'll get a series of warnings in the logs even if no replace is performed (because the local file does not exist).

**To Reproduce**

Create a SFTP outbound gateway like this and suppose you have the remote directory path in the message payload in the request channel:

```xml
<int-sftp:outbound-gateway
  session-factory="mySftpSessionFactory"
  request-channel="requestChannel"
  reply-channel="replyChannel"
  command="mget"
  command-options="-P -R"
  expression="payload.concat('/*')"
  local-directory-expression="'file:${java.io.tmpdir}/myappdir/'.concat(#remoteDirectory)"
  filter="myfilter"
  mode="REPLACE" />

<bean
  id="myFilter"
  class="org.springframework.integration.sftp.filters.SftpSimplePatternFileListFilter">
    <constructor-arg value="*.txt" />
    <property name="alwaysAcceptDirectories" value="true" />
</bean>
```
If a remote file does not exist locally, a warning gets printed in the logs:

```
Failed to delete /tmp/myappdir/remote/path/myFile.txt
```

**Expected behavior**

If the remote file does not currently exist locally, it does not need to be replaced, so it should not be deleted at all.

I think the problem lies in `org.springframework.integration.file.remote.gateway.AbstractRemoteFileOutboundGateway.get(Message<?>, Session<F>, String, String, String, F)`:

```java
FileExistsMode existsMode = this.fileExistsMode;
boolean appending = FileExistsMode.APPEND.equals(existsMode);
boolean exists = localFile.exists();
boolean replacing = FileExistsMode.REPLACE.equals(existsMode)
		|| (exists && FileExistsMode.REPLACE_IF_MODIFIED.equals(existsMode)
		&& localFile.lastModified() != getModified(fileInfo));
if (!exists || appending || replacing) {
  // ...
  if (replacing && !localFile.delete()) {
    this.logger.warn(() -> "Failed to delete " + localFile);
  }
// ...
```

I believe that the `replacing` boolean value should be turned to something like this:

```java
boolean replacing = exists && (FileExistsMode.REPLACE.equals(existsMode)
		|| (FileExistsMode.REPLACE_IF_MODIFIED.equals(existsMode)
		&& localFile.lastModified() != getModified(fileInfo)));
```

That is: we don't need to replace if the local file does not exist... The `localFile.delete()` call will always fail if the local file does not exist.

mauromol:
Fixes spring-projects/spring-integration#3652

`testAll` task passes (except for `org.springframework.integration.xml.transformer.jaxbmarshaling.JaxbMarshallingIntegrationTests.testFileUnlockedAfterUnmarshallingFailure()`, which fails in my Italian-speaking system because the exception message is localized ;-)).

I don't know if it's worth to think of a possible (and viable) test case to test this change...

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
@@ -73,6 +73,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Mauro Molinari
  *
  * @since 2.1
  */
@@ -1110,9 +1111,9 @@ public abstract class AbstractRemoteFileOutboundGateway<F> extends AbstractReply
 		FileExistsMode existsMode = this.fileExistsMode;
 		boolean appending = FileExistsMode.APPEND.equals(existsMode);
 		boolean exists = localFile.exists();
-		boolean replacing = FileExistsMode.REPLACE.equals(existsMode)
-				|| (exists && FileExistsMode.REPLACE_IF_MODIFIED.equals(existsMode)
-				&& localFile.lastModified() != getModified(fileInfo));
+		boolean replacing = exists && (FileExistsMode.REPLACE.equals(existsMode)
+				|| (FileExistsMode.REPLACE_IF_MODIFIED.equals(existsMode)
+				&& localFile.lastModified() != getModified(fileInfo)));
 		if (!exists || appending || replacing) {
 			OutputStream outputStream;
 			String tempFileName = localFile.getAbsolutePath() + this.remoteFileTemplate.getTemporaryFileSuffix();


artembilan:

mauromol:
Hi @artembilan thank you! I had looked at that section regarding the commit message, but I thought it was enough in this case to specify the first line + "Fixes ...", this is why I didn't add anything else after the "Fixes..." line.

artembilan:
No problem, @mauromol !

Our main concern though, is to have a first (head line) no more than 50 symbols.
That's why I added a link to Chris Beams' article.

Just for your info if you'd like to expedite a bit our collaboration in the future 😉

mauromol:
Sure it's useful for me to know, thank you! And indeed, I did know about the length limit of git commit messages (Eclipse EGit has a line wrap marker at 72 characters), I just didn't know (and missed Chris Beam's raccomendation) about the lower limit (50) for the first line. I will take a note.

PR #3665, opened 2021-11-08
RedisLockRegistry, add automatically deleted for redisLock

### LINKED ISSUE #3655: Clean up RedisLockRegistry cache automatically
spring-integration-redis:5.3.3.RELEASE and more?

**Describe the bug**
I'm trying to use RedisLockRegistry as a Singleton.
```
val lock = redisLockRegistry.obtain("key");
try{
  if(lock.tryLock()){
      //business logic
  }
}finally{
  lock.unlock()
}

```
There seems to be a memoryLeak in the service, so I checked
`RedisLockRegistry.locks` was cumulatively occurring.

[obtain](https://github.com/spring-projects/spring-integration/blob/main/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L156)
put into `RedisLockRegistry.locks`

when i check the code
I didn't see the part where 
[expireUnusedOlderThan()](https://github.com/spring-projects/spring-integration/blob/main/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L163)
was called.

Do I have to call it manually?
but, expireUnusedOlderThan() will search the entire stored lock.
I thought there was a way to remove it from my LockRegistry in the lock.unlock() step, but I couldn't find it.

Or should I not use it by declaring it as a bean?(or use prototype?)

**To Reproduce**
Declare RedisLockRegistry as bean and call obtain().

**Expected behavior**
no memory leak
or
How to properly use RedisLockRegistry

Meteorkor:
Resolves #3655

Currently, if you don't manually call `expireUnusedOlderThan()`, OutOfMemory can happen.

Therefore, I changed the code so that if the cache exceeds a setting capacity, it is automatically deleted.

default of capacity occupies about 25 MB.
(`locks` size is 1_000_000)

When reading, multiple threads can access it through readLock.
It is controlled through writeLock only when writing.

I added a queue .
Because it is effective in removing the oldest items

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -34,6 +34,7 @@ import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
@@ -94,7 +95,12 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 					"return false";
 
 
-	private final Map<String, RedisLock> locks;
+	private final Map<String, RedisLock> locks = new LinkedHashMap<String, RedisLock>(16, 0.75F, true) {
+		@Override
+		protected boolean removeEldestEntry(Entry eldest) {
+			return size() > RedisLockRegistry.this.capacity;
+		}
+	};
 
 	private final String clientId = UUID.randomUUID().toString();
 
@@ -106,7 +112,7 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 
 	private final long expireAfter;
 
-	private final long capacity;
+	private long capacity = DEFAULT_CAPACITY;
 
 	/**
 	 * An {@link ExecutorService} to call {@link StringRedisTemplate#delete} in
@@ -139,30 +145,12 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 	 * @param expireAfter The expiration in milliseconds.
 	 */
 	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, long expireAfter) {
-		this(connectionFactory, registryKey, expireAfter, DEFAULT_CAPACITY);
-	}
-
-	/**
-	 * Constructs a lock registry with the supplied lock expiration, and with cache limit count
-	 * @param connectionFactory The connection factory.
-	 * @param registryKey The key prefix for locks.
-	 * @param expireAfter The expiration in milliseconds.
-	 * @param capacity The capacity of cached lock
-	 */
-	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, long expireAfter, long capacity) {
 		Assert.notNull(connectionFactory, "'connectionFactory' cannot be null");
 		Assert.notNull(registryKey, "'registryKey' cannot be null");
 		this.redisTemplate = new StringRedisTemplate(connectionFactory);
 		this.obtainLockScript = new DefaultRedisScript<>(OBTAIN_LOCK_SCRIPT, Boolean.class);
 		this.registryKey = registryKey;
 		this.expireAfter = expireAfter;
-		this.capacity = capacity;
-		this.locks = new LinkedHashMap<String, RedisLock>(16, 0.75F, true) {
-			@Override
-			protected boolean removeEldestEntry(Entry eldest) {
-				return size() > capacity;
-			}
-		};
 	}
 
 	/**
@@ -176,11 +164,32 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 		this.executorExplicitlySet = true;
 	}
 
+	/**
+	 * Rebalance the capacity of cached locks.
+	 * @param capacity The capacity of cached lock
+	 * @since 5.5.6
+	 */
+	public void setCapacity(long capacity) {
+		synchronized (this.locks) {
+			final int locksSize = this.locks.size();
+			final int newLockMapSize = (int) capacity;
+			if (locksSize > newLockMapSize) {
+				final int removeCnt = locksSize - newLockMapSize;
+				Iterator<Entry<String, RedisLock>> iterator = this.locks.entrySet().iterator();
+				for (int i = 0; i < removeCnt; i++) {
+					iterator.next();
+					iterator.remove();
+				}
+			}
+			this.capacity = capacity;
+		}
+	}
+
 	@Override
 	public Lock obtain(Object lockKey) {
 		Assert.isInstanceOf(String.class, lockKey);
 		String path = (String) lockKey;
-		synchronized (locks){
+		synchronized (this.locks) {
 			return this.locks.computeIfAbsent(path, RedisLock::new);
 		}
 	}
@@ -188,14 +197,15 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 	@Override
 	public void expireUnusedOlderThan(long age) {
 		long now = System.currentTimeMillis();
-		synchronized (locks) {
-			Iterator<Entry<String, RedisLock>> iterator = locks.entrySet().iterator();
+		synchronized (this.locks) {
+			Iterator<Entry<String, RedisLock>> iterator = this.locks.entrySet().iterator();
 			while (iterator.hasNext()) {
 				Entry<String, RedisLock> entry = iterator.next();
 				RedisLock lock = entry.getValue();
 				if (now - lock.getLockedAt() > age && !lock.isAcquiredInThisProcess()) {
 					iterator.remove();
-				} else {
+				}
+				else {
 					break;
 				}
 			}
@@ -213,12 +223,15 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 
 		private final String lockKey;
 
+		private final String path;
+
 		private final ReentrantLock localLock = new ReentrantLock();
 
 		private volatile long lockedAt;
 
 		private RedisLock(String path) {
 			this.lockKey = constructLockKey(path);
+			this.path = path;
 		}
 
 		private String constructLockKey(String path) {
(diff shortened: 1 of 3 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -92,6 +97,10 @@
 
 	private final Map<String, RedisLock> locks = new ConcurrentHashMap<>();
 
+	private final ConcurrentLinkedDeque<String> queue = new ConcurrentLinkedDeque<>();
```
artembilan, line 100: I don't think we need to worry about `put` and `remove` optimization.
Those `obtain()` and `expireUnusedOlderThan()` are not designed to be used in the heavy concurrent environment.
The first one is just for lock reference, which you can store for reuse.
The second one indeed has to be called on from an exclusive thread.

Probably for a solid logic it is enough to make those to method as `synchronized`, if you don't want to wrap `LinkedHashMap` into a `Collections.synchronizedMap()`.

And we might need to unit-test to cover this new functionality.

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
Meteorkor: Thank you for your review.
Fixed to use `LinkedHashMap`.
and
Added test case.

artembilan (changes requested):
Please, consider to run `gradlew clean :spring-integration-redis:testAll` to be sure that all tests pass and Checkstyle rules are not violated.

Thank you for your effort so far!

artembilan (changes requested):
Please, also consider to add some note into the `redis.adoc` about that new `capacity` option starting with version `5.5.6`.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -156,17 +180,26 @@ public void setExecutor(Executor executor) {
...
-				});
+		synchronized (locks) {
+			Iterator<Entry<String, RedisLock>> iterator = locks.entrySet().iterator();
+			while (iterator.hasNext()) {
+				Entry<String, RedisLock> entry = iterator.next();
+				RedisLock lock = entry.getValue();
+				if (now - lock.getLockedAt() > age && !lock.isAcquiredInThisProcess()) {
+					iterator.remove();
+				} else {
+					break;
```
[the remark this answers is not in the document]
Meteorkor: > I don't think this is correct assumption to stop iterating the cache because we don't match that `getLockedAt()` any more. The cache is only for instances to not overhead with them for the same key, but any obtained `Lock` can be locked later in the logic at arbitrary time. Therefore the oldest entry in the cache might be the most fresh with its `lock()` call. So, early exit from such a loop is misleading and may leave really old not locked entries in the cache.

Thanks for the review.

Fixed to be sorted when lock() is called.
For removeEldestEntry() and expireUnusedOlderThan()

Is it ok to change like this?

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -133,12 +139,30 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
...
+	}
+
+	/**
+	 * Constructs a lock registry with the supplied lock expiration, and with cache limit count
+	 * @param connectionFactory The connection factory.
+	 * @param registryKey The key prefix for locks.
+	 * @param expireAfter The expiration in milliseconds.
+	 * @param capacity The capacity of cached lock
+	 */
+	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, long expireAfter, long capacity) {
```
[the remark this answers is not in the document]
Meteorkor: > I think this new `capacity` option could be just a setter. No need to be precise via ctor.

Add setter for capacity and remove ctor

Meteorkor:
> Please, also consider to add some note into the `redis.adoc` about that new `capacity` option starting with version `5.5.6`.

edit redis.adoc

Please tell me if there is anything strange

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -287,7 +333,11 @@ private boolean obtainLock() {
 			boolean result = Boolean.TRUE.equals(success);
 
 			if (result) {
-				this.lockedAt = System.currentTimeMillis();
+				final long now = System.currentTimeMillis();
+				synchronized (RedisLockRegistry.this.locks) {
```
artembilan, line 337: No, this is not OK.
Better to have a heavy `expireUnusedOlderThan()` iterating over all the entries, than have a block on every single `Lock.lock()`.

src/reference/asciidoc/redis.adoc
```diff
@@ -883,3 +883,6 @@ However, the resources protected by such a lock may have been compromised, so su
 You should set the expiry at a large enough value to prevent this condition, but set it low enough that the lock can be recovered after a server failure in a reasonable amount of time.
 
 Starting with version 5.0, the `RedisLockRegistry` implements `ExpirableLockRegistry`, which removes locks last acquired more than `age` ago and that are not currently locked.
+
+String with version 5.5.6, the `RedisLockRegistry` is support automatically clean up cache for redisLocks in `RedisLockRegistry.locks` via `RedisLockRegistry.capacity`.
```
artembilan, line 887:
```suggestion
String with version 5.5.6, the `RedisLockRegistry` is support automatically clean up cache for redisLocks in `RedisLockRegistry.locks` via `RedisLockRegistry.setCaapacity()`.
```

src/reference/asciidoc/redis.adoc
```diff
@@ -883,3 +883,6 @@ However, the resources protected by such a lock may have been compromised, so su
...
+`capacity` field can be modified via setters, The default is 1,000,000.
```
artembilan, line 888:
```suggestion
See its JavaDocs for more information.
```

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -152,21 +164,52 @@ public void setExecutor(Executor executor) {
 		this.executorExplicitlySet = true;
 	}
 
+	/**
+	 * Rebalance the capacity of cached locks.
+	 * @param capacity The capacity of cached lock
```
artembilan, line 169: Mention, please, here in JavaDoc a default value.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -152,21 +164,52 @@ public void setExecutor(Executor executor) {
...
+	public void setCapacity(long capacity) {
+		synchronized (this.locks) {
+			final int locksSize = this.locks.size();
+			final int newLockMapSize = (int) capacity;
+			if (locksSize > newLockMapSize) {
+				final int removeCnt = locksSize - newLockMapSize;
+				Iterator<Entry<String, RedisLock>> iterator = this.locks.entrySet().iterator();
+				for (int i = 0; i < removeCnt; i++) {
+					iterator.next();
+					iterator.remove();
```
artembilan, line 181: I don't think that someone is going to modify the `capacity` at runtime.
Move over we consult with it in that `removeEldestEntry` at runtime any way.
So, new value is going to be an effect and therefore obsolete entries are going to be removed on the next `put()`.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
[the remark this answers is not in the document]
artembilan: It is not. 
See my comment on the `obtainLock()` below.

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
Meteorkor: If it is not sorted when `RedisLockRegistry.RedisLock.obtainLock()` is called, the Lock object that was recently called `RedisLockRegistry.RedisLock.lock()` in `removeEldestEntry()` may be removed.

Is this ok because the cache has been removed?

artembilan: I think it is not a big deal, since the `Lock` object is still valid.
We only miss it in the cache for the next `obtain()` call.
I think more like this cache feature should go at all, but it won't heart to have the logic like we do right now: it is very rear situation that some project would have more then 1,000,000 locks in its logic.

Meteorkor: thanks for your comment
Unfortunately, I recently discovered that it is used in a similar case in another project.
It is being used to prevent simultaneous execution of a specific request by a user.

rolled back `obtainLock()`

src/reference/asciidoc/redis.adoc
Meteorkor: Thanks for the review.
edit redis.adoc

src/reference/asciidoc/redis.adoc
Meteorkor: Thanks for the review.
edit redis.adoc

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
Meteorkor: Thanks for the review.
edit JavaDoc

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
Meteorkor: Thanks for the review.
Fixed not to rebalance in setter

artembilan (approved):
LGTM.

Pulling locally for final review and merge process...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/e51513230afd0ddad403d8f0ab093c3d7ee9eb59.

@Meteorkor ,

For example our `JdbcLockRegistry` suffers from the same issue 😉

Meteorkor:
@artembilan 

Thanks for the review
Thank you for the opportunity to contribute.

Let's take a further look at the `JdbcLockRegistry`

PR #3669, opened 2021-11-10
Fix WebFluxMH for proper response handling

artembilan:
The `WebFluxRequestExecutingMessageHandler` does direct `ClientResponse.create(entity.getStatusCode())`
which comes with a `ExchangeStrategies.withDefaults()`.
Even if end-user configures a `WebClient` properly, the response is created with default strategies.

* Rework `WebFluxRequestExecutingMessageHandler` internal logic to call `ResponseSpec.toEntityFlux(BodyExtractor)`
instead of manual `ClientResponse.create()`
* Add unit test to the `WebFluxRequestExecutingMessageHandlerTests` to ensure that configured `maxInMemorySize`
on the `WebClient` strategies has an effect when response body is bigger than expected size

**Cherry-pick to `5.4.x`**

### DIFF
--- a/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/dsl/WebFluxMessageHandlerSpec.java
+++ b/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/dsl/WebFluxMessageHandlerSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 the original author or authors.
+ * Copyright 2017-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


garyrussell:
...and cherrry-picked.

djgraff209:
@artembilan I just verified this against the `5.5.6-SNAPSHOT` version and it is working as expected and respects either the property `spring.codec.max-in-memory-size` or a configured `WebClient`.

artembilan:
Thank you for confirming that, @djgraff209 !

So, we are good to proceed with releases next week 😉

djgraff209:
Verified the Spring-Boot 2.5.7 release which includes this. No modifications to dependencies necessary. Works as advertised.

PR #3671, opened 2021-11-11
Update RedisLockRegistry default capacity value

Meteorkor:
@artembilan
 
I found that the default was calculated incorrectly.
I'm sorry, but could you please reflect the default value?
(1,000,000 -> 100,000)

Is it better to proceed after registering as an issue?

CAPACITY = 100,000
image

CAPACITY = 1,000,000
image

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -80,7 +80,7 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 
 	private static final long DEFAULT_EXPIRE_AFTER = 60000L;
 
-	private static final int DEFAULT_CAPACITY = 1_000_000;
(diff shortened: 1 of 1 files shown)


artembilan:
I think the concern is really pointless, when you are going to have pretty complex application with different lock registries, so all of them individually are going to take memory slowly.
The `capacity` option comes to the rescue anyway.

Although I think we need to rename that `setCapacity(int capacity)` to `cacheCapacity`. Otherwise it may mislead end-users since this registry deals with the target store, which also may have its own capacity or so.
WDYT?

Meteorkor:
As you said, there may be very few problems.
But I thought it was very important to set the default, so I requested a change.

Thanks for merging.

> `setCapacity(int capacity)` to `cacheCapacity`

It seems good to change the name.

I have a PR I wrote, is it okay to apply it here?
https://github.com/spring-projects/spring-integration/pull/3673

artembilan:
Yes, please, do that in that your PR for JDBC and Zookeeper.

Will review it shortly.

PR #3678, opened 2021-11-16
GH-3677: Doc for URL conn customization in FeedCA

### LINKED ISSUE #3677: UrlConnection timeout required for FeedEntryMessageSource
**Expected Behavior**

FeedEntryMessageSource should timeout the connection when server is unresponsive for desired amount of time

**Current Behavior**

FeedEntryMessageSource holds the connection infinitely which is blocking other threads.

Here is stackoverflow url: https://stackoverflow.com/questions/69822045/threads-are-being-blocked-while-receiving-data-from-rss-feed-adapter?noredirect=1#comment123697506_69822045

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3677

If there is need to have a `URLConnection` customized, the `UrlResource`
has to be used instead of plain `URL` injection into the `FeedEntryMessageSource`

### DIFF
--- a/src/reference/asciidoc/feed.adoc
+++ b/src/reference/asciidoc/feed.adoc
@@ -87,6 +87,31 @@ This is useful when the feed source is not an HTTP endpoint but is any other res
 In the `FeedEntryMessageSource` logic, such a resource (or provided `URL`) is parsed by the `SyndFeedInput` to the `SyndFeed` object for the processing mentioned earlier.
 You can also inject a customized `SyndFeedInput` (for example, with the `allowDoctypes` option) instance into the `FeedEntryMessageSource`.
 
+[NOTE]
+====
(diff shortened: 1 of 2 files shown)


src/reference/asciidoc/feed.adoc
```diff
@@ -127,3 +127,5 @@ public class FeedJavaApplication {
 }
 ----
 ====
+
+NOTE: If the connection to the feed needs some customization, e.g. connection and read timeouts, the `org.springframework.core.io.UrlResource` extension with its `customizeConnection(HttpURLConnection)` override has to be used instead of plain `URL` injection into the `FeedEntryMessageSource`.
```
garyrussell, line 131: Perhaps we could include an example?

garyrussell (approved):

PR #3684, opened 2021-11-25
Make sure DefaultLockRepository always use new transaction

### LINKED ISSUE #3683: DefaultLockRepository can misbehave when DB transaction is already active
Currently `acquire` method of `DefaultLockRepository` has `@Transactional(isolation = Isolation.SERIALIZABLE)` annotation.

https://github.com/spring-projects/spring-integration/blob/a843dae1512bdfb1308ee8b3f7c59754c4ae5a21/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java#L162

But I wander what will be if I try to acquire a LOCK from the method that is already annotated with `@Transactional`?

If the transaction is already started, the default propagation level is `REQUIRED` and that does mean that already opened transaction will be supported instead of creating a new one. And the documentation of `@Transactional.isolation` attribute explicitly says:

> Exclusively designed for use with `Propagation.REQUIRED` or `Propagation.REQUIRES_NEW` since it only applies to newly started transactions.

I'm not sure what does it mean in the context of `REQUIRED` propagation when it's an inner `REQUIRED` transaction, but it looks like it will just do nothing with the isolation level if the transaction is already started. So the isolation level will be the same that is set by most outer transaction (real DB transaction that is already started).

To me it looks like `@Transactional(isolation = Isolation.SERIALIZABLE)` annotation must include `propagation = Propagation.REQUIRES_NEW` attribute to make sure that LOCK acquiring will always be done in a separate transaction with SERIALIZABLE isolation level.

I also see other potential problems with the shared transaction between user's code and LOCK repository. For example, any UPDATE of a record in `INT_LOCK` table will be blocked if other thread already acquired the lock and now executes the user's code under this LOCK (in the same transaction that acquired the lock). In this case an EXCLUSIVE DB LOCK on an INT_LOCK's record will be hold by a first transaction for the whole time while user's (outer) transaction is not committed or rolled back. So, e.g. `JdbcLockRegistry.JdbcLock.tryLock` method will be blocked for a potentially long time.

Another potential problem could be for the retry logic in the case of transient DB exception (e.g. deadlock).
Currently `JdbcLockRegistry.JdbcLock.lock()` method does retry is the case of one of:

* TransientDataAccessException
* TransactionTimedOutException
* TransactionSystemException

But this retry logic could not always work correctly if it will not start a new DB transaction before each retry. With current implementation of `acquire()` method it looks like it's not a problem, but if the method will e.g. `SELECT` before `UPDATE` in a REPEATABLE_READ transaction, the logic could be flawed because SELECT will always return the same result that was returned when it was executed the 1st time in this transaction. Anyway, it looks like the retry logic in `JdbcLockRegistry.JdbcLock` was implemented with the assumption that a new transaction will be started by each `acquire` method execution.

Another problem with a shared transaction I could imagine: a user's transaction that acquires and releases the LOCK and then executes a method with REQUIRES_NEW transaction that tries to acquire the same (already released) LOCK. It will not be able to do this because of a transaction opened by the caller method that is still active and just suspended. The java LOCK is already released by the caller's method, but the DB EXCLUSIVE LOCK on the deleted DB record of `INT_LOCK` table is still held by this suspended transaction, so new transaction in the called method could not UPDATE or INSERT the same record into the table:

```java
public class Service1 {
  @Transactional
  public void method1() {
    Lock lock = lockRegistry.obtain("lock1");
    lock.lock();
    try {
      // some logic
    } finally {
      lock.unlock();
    }
    
    service2.method2(); // current transaction is suspended but is still active
  }
}

public class Service2 {
  @Transactional(propagation = Propagation.REQUIRES_NEW)
  public void method2() {
    Lock lock = lockRegistry.obtain("lock1");
    lock.lock(); // it will never succeed because DB X LOCK is still held by another transaction!
    // JdbcLockRegistry.JdbcLock.lock() will try again indefinitely because MySQL exception is:
    // SQL Error [1205] [40001]: Lock wait timeout exceeded; try restarting transaction
    // So I suppose it will be TransientDataAccessException or TransactionTimedOutException
    // which is trigger for "try again" to execute DefaultLockRepository.acquire().
  }
}
```

Also reusing the same transaction opened by user's code could increase the probability of deadlocks, as user's transaction could already hold some DB locks at the point of acquiring the LOCK, and if it does this in multiple places in reverse order, a deadlock will happen.

**All in all, to me it looks like DefaultLockRepository and JdbcLockRegistry were designed with the assumption that a new transaction will be opened on each call to any repository method instead of supporting an already opened transaction. Am I right?**

A also noticed the `@Transactional` annotation on the whole `DefaultLockRepository` class that looks strange to me as some methods like `setRegion` doesn't require a transaction. The annotation on the class is also applies to all methods of subclasses, so this makes extending of `DefaultLockRepository` class harder. This is not a major problem, though. But `propagation = Propagation.REQUIRES_NEW` attribute also should be added here (or `@Transactional` annotation should be moved from the class level to appropriate methods level).

xak2000:
If a transaction is already active while JdbcLockRegistry uses
DefaultLockRepository to acquire/release a lock, the repository
must execute SQL queries in a separate transaction to prevent
problems with blocking, deadlocking etc.

It also allows to properly follow transaction isolation level that
is set on some methods of DefaultLockRepository.

Previously all methods of DefaultLockRepository supported the current
transaction if there are any. Now all methods will always create new
transaction on each call.

Fixes #3683

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 the original author or authors.
+ * Copyright 2016-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ import org.springframework.dao.DuplicateKeyException;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Repository;
 import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.util.Assert;
 
@@ -44,11 +45,11 @@ import org.springframework.util.Assert;
  * @author Glenn Renfro
  * @author Gary Russell
  * @author Alexandre Strubel
+ * @author Ruslan Stelmachenko
  *
  * @since 4.3
  */
 @Repository
-@Transactional
 public class DefaultLockRepository implements LockRepository, InitializingBean {
 
 	/**
@@ -149,17 +150,19 @@ public class DefaultLockRepository implements LockRepository, InitializingBean {
 		this.renewQuery = String.format(this.renewQuery, this.prefix);
 	}
 
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	@Override
 	public void close() {
 		this.template.update(this.deleteAllQuery, this.region, this.id);
 	}
 
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	@Override
 	public void delete(String lock) {
 		this.template.update(this.deleteQuery, this.region, lock, this.id);
 	}
 
-	@Transactional(isolation = Isolation.SERIALIZABLE)
+	@Transactional(propagation = Propagation.REQUIRES_NEW, isolation = Isolation.SERIALIZABLE)
 	@Override
 	public boolean acquire(String lock) {
 		if (this.template.update(this.updateQuery, this.id, new Date(), this.region, lock, this.id,
@@ -174,17 +177,20 @@ public class DefaultLockRepository implements LockRepository, InitializingBean {
 		}
 	}
 
+	@Transactional(propagation = Propagation.REQUIRES_NEW, readOnly = true)
 	@Override
 	public boolean isAcquired(String lock) {
 		return this.template.queryForObject(this.countQuery, Integer.class, // NOSONAR query never returns null
 				this.region, lock, this.id, new Date(System.currentTimeMillis() - this.ttl)) == 1;
 	}
 
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	@Override
 	public void deleteExpired() {
 		this.template.update(this.deleteExpiredQuery, this.region, new Date(System.currentTimeMillis() - this.ttl));
 	}
 
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	@Override
 	public boolean renew(String lock) {
 		return this.template.update(this.renewQuery, new Date(), this.region, lock, this.id) > 0;
--- a/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/lock/DefaultLockRepositoryTests.java
+++ b/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/lock/DefaultLockRepositoryTests.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2016-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.jdbc.lock;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import java.sql.Connection;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+/**
+ * @author Ruslan Stelmachenko
+ *
+ * @since 5.5.7
+ */
+@SpringJUnitConfig(locations = "JdbcLockRegistryTests-context.xml")
+@DirtiesContext
+public class DefaultLockRepositoryTests {
+
+	@Autowired
+	private LockRepository client;
+
+	@BeforeEach
+	public void clear() {
+		this.client.close();
+	}
+
+	@Transactional
+	@Test
+	public void testNewTransactionIsStartedWhenTransactionIsAlreadyActive() {
+		// Make sure a transaction is active
+		assertThat(TransactionSynchronizationManager.isActualTransactionActive()).isTrue();
+
+		TransactionSynchronization transactionSynchronization = spy(TransactionSynchronization.class);
+		TransactionSynchronizationManager.registerSynchronization(transactionSynchronization);
+
+		this.client.acquire("foo"); // 1
+		this.client.renew("foo"); // 2
+		this.client.delete("foo"); // 3
+		this.client.isAcquired("foo"); // 4
+		this.client.deleteExpired(); // 5
+		this.client.close(); // 6
+
+		// Make sure a transaction is still active
+		assertThat(TransactionSynchronizationManager.isActualTransactionActive()).isTrue();
+		// And was suspended for each invocation of @Transactional methods of DefaultLockRepository,
+		// that confirms that these methods were called in a separate transaction each.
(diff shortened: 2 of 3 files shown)


artembilan:
Hi @xak2000 !

Any chances to have a simple unit test to cover this kinda new behavior?

xak2000:
I'll try but currently I don't have a time. Maybe a week or two.

I suppose the test should start a transaction, then call `DefaultLockRepository.acquire` and then check **somehow** that new transaction was started and successfully committed after `acquire` is finished it's work and current transaction is still active (not committed).

@artembilan Can you point me to similar tests where checks for transaction reuse is made if you have something in mind? If no, don't worry. I'll try to investigate it myself when have a time.

xak2000:
@artembilan I looked into current tests and found out that there is no tests for `DefaultLockRepository` component. All tests are for  `JdbcLockRegistry`. But what we really need to test here is `DefaultLockRepository`'s methods, isn't it?

They could be tested implicitly by calling to `JdbcLockRegistry` methods of course, but is it a good idea? What do you think? Should I add new `DefaultLockRepositoryTests` class or use the existing `JdbcLockRegistryTests`?

Also I'm not sure how to check that new transaction was started and commited (even if currently there is already active transaction) after `DefaultLockRepository.acquire()` finished it's work.

I know how to check if currently there is active transaction (`TransactionSynchronizationManager.isActualTransactionActive()`). But how to check that new transaction was started and finished in the called code? The only idea is to start new REQUIRES_NEW transaction from the test method and then check the contents of DB using jdbcTemplate or by calling `DefaultLockRepository.isAcquired()`. If `acquire()` was called in a separate transaction then new transaction should return the data, but if if `acquire()` was called from the transaction, started in the test (that is currently suspended), then new transaction will return empty result set, as the suspended transaction is not committed yet.

I think other `@Transactional` methods of `DefaultLockRepository` could be tested the same way.

Any thoughts on an easier way to check if called method really was called in a new transaction (rather than supplying an existing transaction)?

artembilan:
Hi @xak2000 !

Thank you for coming back to us!

I'm out of work today, so I'll give it more thoughts next week.
But for now according your observations:
I don't think we need a dedicated `DefaultLockRepositoryTests` since all of its methods are fully covered by ` JdbcLockRegistry` calls. So, I think there should be enough to be sure that existing tests don't fail. Plus add a simple new one to check that new transaction is started even if there is existing one.
Please, take a look into a `org.springframework.transaction.support.TransactionSynchronization` which you can register to the existing transaction via `TransactionSynchronizationManager.registerSynchronization()`.
That `TransactionSynchronization` has this method:
```
	/**
	 * Suspend this synchronization.
	 * Supposed to unbind resources from TransactionSynchronizationManager if managing any.
	 * @see TransactionSynchronizationManager#unbindResource
	 */
	default void suspend() {
```

So, probably to cover your new behavior there is just enough to check in the test that this method is called when we done dealing with the lock when current test method is transactional.

xak2000:
Hello @artembilan !

I tried your suggestion and it works! But to cover all `@Transactional` methods of `DefaultLockRepository` I had to use some methods of repository instance directly. Please take a look.

Also to me it looks like just calling all 6 methods of repository instance directly will be a better idea as it will cover these methods with 100% probability, while calling them indirectly through registry is somewhat brittle. What if registry implementation will be changed at some point and some methods of repository will not be called anymore (but some others will be called instead, so the test will still be successful)?

This indirection also makes the test less readable, because to be able to know what methods of `DefaultLockRepository` are called from e.g. `JdbcLockRegistry.lock()` you need to follow the source code of `JdbcLockRegistry.lock()` method.

So I don't understand why we need to test `DefaultLockRepository` behavior through `JdbcLockRegistry` methods. We need to make sure that all transactional methods of repository are really open-close new transaction. We don't care what registry does and how it uses them. This is responsibility of other tests.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
```diff
@@ -174,17 +177,20 @@ public boolean acquire(String lock) {
 		}
 	}
 
+	@Transactional(propagation = Propagation.REQUIRES_NEW, readOnly = true)
```
artembilan, line 180: I wonder why do we need a dedicated transaction for `SELECT`?..

artembilan (changes requested):
Hi @xak2000 !

Thank you for such a thorough thinking about the problem.
Yes, now I agree with you that it is better to have a dedicated test class to cover the functionality just of that `DefaultLockRepository`.

So, let's see if you can come up with something simple which we really could call unit-tests nail that your concern about changed behavior in the future.

Also, please find the question from me on some line of the code.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
xak2000: Actually I didn't think about it. Just moved `@Transactional` method from the class to each method that executes some query. But now when you asked I researched it and concluded that new transaction is still required here even for methods that only read from DB.

If current transaction's isolation level is REPEATABLE_READ or SERIALIZABLE then after 1st `SELECT` all further `SELECT`s will return the same data, even if DB records were already modified/deleted by concurrent transactions.

This means that if first call to `this.client.isAcquired(path)`  returned `true` (if was called from user's transaction with REPEATABLE_READ isolation level), then all next calls will also return `true` while user's transaction is still active.

This means that `registry.expireUnusedOlderThan(..)` will never expire anything as long as user's transaction is still active.

I written a small test that confirms this.

```java
	@Transactional(isolation = Isolation.REPEATABLE_READ)
	@Test
	public void testIsAcquiredFromRepeatableReadTransaction() {
		// Make sure a transaction with REPEATABLE_READ isolation level is active
		assertThat(TransactionSynchronizationManager.isActualTransactionActive()).isTrue();
		assertThat(TransactionSynchronizationManager.getCurrentTransactionIsolationLevel())
				.isEqualTo(Connection.TRANSACTION_REPEATABLE_READ);

		Lock lock = this.registry.obtain("foo");
		String path = TestUtils.getPropertyValue(lock, "path", String.class);
		lock.lock();
		try {
			assertThat(this.client.isAcquired(path)).isTrue();
		}
		finally {
			lock.unlock();
		}
		assertThat(this.client.isAcquired(path)).isFalse();
	}
```

This test fails when this `SELECT` doesn't execute in a new transaction.

I also tried to use

```java
this.registry.expireUnusedOlderThan(-1000);
assertThat(TestUtils.getPropertyValue(this.registry, "locks", Map.class).size()).isEqualTo(1); // and then isEqualTo(0) after unlock()
```

instead of

```
assertThat(this.client.isAcquired(path)).isTrue(); // and then isFalse()
```

and it also fails.

Should I add this test too? Maybe I should replace `lock.lock()`/`lock.unlock()` with direct usage of `DefaultLockRepository` methods. This case is actually already covered by test that checks transaction's suspension count. This additional test just makes the intention more clear for the `SELECT` case.

xak2000: Actually I thought about it more and maybe it's worth to set `Propagation.NOT_SUPPORTED` on `boolean isAcquired(String lock)`? This way the test still passes as the problem is not the absence of a new transaction but presence of existing transaction. Executing SELECT without opening and closing a new transaction will make this method faster in theory. It will still be executed in a new DB connection and current transaction will still be suspended, though.

artembilan: Good. Thanks for the report. Didn't think about the cause behind `REPEATABLE_READ`.
Let's have that TX on the SELECT and let's include such a test into our suite!

artembilan: Yeah... Let's leave it transactional for consistency and clear separation of concerns that the state around lock record in DB has to be volatile in terms of Java 😸

xak2000: Yeah, it's pretty subtle. The problem when user's transaction is `SERIALIZABLE` will be even worse, as `SELECT` will acquire shared lock on DB record and all concurrent transactions that trying to do something with this DB record will be blocked waiting for this shared lock to be released (bye-bye `tryLock()`). And it will be released only after user's transaction will be committed. 🙈 

Btw, I removed the `JdbcLockRegistry` test and added dedicated tests for `DefaultLockRepository`.

artembilan (approved):
Pulling locally for final review and possible merge...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/ad343107071bf55743a836ec0a0a55167c6b5062 and cherry-picked to `5.4.x` after minor clean up.
Back-ported to `5.3.x` as https://github.com/spring-projects/spring-integration/commit/ad159df6476f42f27ee1d7c6745248d7abb8eb66 after fixing conflicts around yet non-existing methods.

@xak2000 ,

thank you very much for your awesome contribution!

PR #3687, opened 2021-12-09
Quote CONDITION whenever necessary for JDBC

artembilan:
SO: https://stackoverflow.com/questions/70286480/how-can-spring-integration-5-5-x-use-mysql-as-message-store

Turns out the condition word is reserved in many SQL DB vendors, e.g.:
https://dev.mysql.com/doc/refman/8.0/en/keywords.html

* Fix SQL scripts for quoting `CONDITION` column name for
those vendors which have it as a reserved word
* Fix `JdbcMessageStore` to have a `CONDITION` quoted to `""` by default and
replaced to "`" for MySQL
* Add `MySqlContainerTest` to test against MySQL Docker container
* Looks like quoted identifiers work well even if they are not reserved words in the SQL vendor


artembilan:
Also has fixed Migration Guide: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.4-to-5.5-Migration-Guide#messagegroupconditon-option

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
```diff
@@ -87,7 +87,7 @@
 				"(GROUP_KEY, REGION, COMPLETE, LAST_RELEASED_SEQUENCE, CREATED_DATE, UPDATED_DATE)"
 				+ " values (?, ?, 0, 0, ?, ?)"),
 
-		UPDATE_MESSAGE_GROUP("UPDATE %PREFIX%MESSAGE_GROUP set UPDATED_DATE=?, CONDITION=? " +
+		UPDATE_MESSAGE_GROUP("UPDATE %PREFIX%MESSAGE_GROUP set UPDATED_DATE=?, GROUP_CONDITION=? " +
```
garyrussell, line 90: This is a big breaking change in a point release; I think it's ok to change the scripts but I think this code needs to be made conditional on a property (defaulting to CONDITION in the point release and GROUP_CONDITION in 6.0).

garyrussell (changes requested):
I fixed a few typos in the migration guide.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
artembilan: OK. How about to make it "smart" conditional: based on vendor derived from the provided `DataSource`?
All those which have this `condition` as a reserved word are not going to work anyway.
So, we change the word exactly in their scripts and check for them in the `JdbcMessageStore`.
This way we would leave the rest of supported vendors working as before this change.

artembilan:
I have reproduced the issue with MySQL under Testcontainers:
```
Caused by: java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CONDITION VARCHAR(255), COMPLETE BIGINT, LAST_RELEASED_SEQUENCE BIGINT, CREATED_' at line 1
```
This fixes the problem: 
```
 `CONDITION` VARCHAR(255),
```

Now looking for the possible fix in the rest of scripts and `JdbcMessageStore`...

artembilan:
Pushed the fix which is more flexible according the current state of things:
1. If vendor doesn't fail for `CONDITION` identifier, it won't fail with this fix.
2. If it fails, then script is updated.
3. MySQL has a special quotation rules.

garyrussell (approved):

PR #3689, opened 2021-12-10
GH-3688: Fix WS DSL for proper values propagation

### LINKED ISSUE #3688: MarshallingWsOutboundGatewayNoTemplateSpec overwrites useful WebserviceTemplate defaults with null

5.5.6, 5.4.x, but other versions are affected as well

**Describe the bug**

In MarshallingWsOutboundGatewaySpec.MarshallingWsOutboundGatewayNoTemplateSpec class the assemble() method overwrites _messageSenders and faultMessageResolver's_ otherwise useful and valid default not null values
with null values in the WebServiceTemplate created by the builder itself if no values provided in the builder explicitly:
https://github.com/spring-projects/spring-integration/blob/5.4.x/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsOutboundGatewaySpec.java#L144
https://github.com/spring-projects/spring-integration/blob/5.4.x/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java#L181

Also, one MUST provide an encodingMode in the builder, because BaseWsOutboundGatewaySpec does exactly the same (overwrites the correct default value without null checking) with the UriFactory's encodingMode:
https://github.com/spring-projects/spring-integration/blob/5.4.x/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java#L162
https://github.com/spring-projects/spring-integration/blob/5.4.x/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java#L140

**To Reproduce**

create a Ws.marshallingOutboundGateway()

**Expected behavior**

faultMessageResolver, messageSenders and enCodingmode:
- may be left unspecified in builder, in which case the defaults in WebServiceTemplate and UriFactory should be left intact
_(Note: A simple null check before the setter calls would be sufficient.)_
- if one needs null values, they should provide WebServiceTemplate and UriFactory explicitly (with mentioned fields set to null)

**Sample**
```
IntegrationFlow myFlow = IntegrationFlows
        .from(MessageChannels.direct("myChannel"))
        .handle(
            Ws.marshallingOutboundGateway()
                .marshaller(jaxb2Marshaller())
                .destinationProvider(createDestinationProvider(myBackendUri))
                // IntegrationFlows's Ws.marshallingOutBoundGateway a little bit buggy, it
                // overwrites nice defaults already set, so we have to provide encodingMode,
                // faultMessageResolver and MessageSenders here, otherwise they will be nulled,
                // which results in an error...
                .encodingMode(DefaultUriBuilderFactory.EncodingMode.TEMPLATE_AND_VALUES)
                .messageSenders(getWebserviceTemplateDefaultMessageSenders())
                .faultMessageResolver(getWebserviceTemplateDefaultFaultMessageResolver()))
        .get();
```

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3688

The `WebServiceTemplate` is populated with some defaults from its ctor.
We should rely on the target default template values as much as possible
and don't override them to `null` if end-user doesn't ask about that explicitly

* Fix `BaseWsOutboundGatewaySpec` extensions to populate values to the target gateway
only if they are not null - therefore provided by end-user
* If end-user wants them explicitly `null`, it is better to do that via
externally configured template.
See overloaded variants for DSL: `Ws.marshallingOutboundGateway(WebServiceTemplate)`
and `Ws.simpleOutboundGateway(WebServiceTemplate)`

### DIFF
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
@@ -158,8 +158,8 @@ public abstract class BaseWsOutboundGatewaySpec<
 	protected E assemble(E gateway) {
 		gateway.setUriVariableExpressions(this.uriVariableExpressions);
 		JavaUtils.INSTANCE
-				.acceptIfNotNull(this.headerMapper, gateway::setHeaderMapper);
-		gateway.setEncodingMode(this.encodingMode);
+				.acceptIfNotNull(this.headerMapper, gateway::setHeaderMapper)
+				.acceptIfNotNull(this.encodingMode, gateway::setEncodingMode);
 		gateway.setIgnoreEmptyResponses(this.ignoreEmptyResponses);
 		gateway.setRequestCallback(this.requestCallback);
 		return gateway;
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsOutboundGatewaySpec.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsOutboundGatewaySpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 the original author or authors.
+ * Copyright 2020-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


garyrussell (approved):

garyrussell:
... and cherry-picked.

Issue #3700, opened 2021-12-26
I am using RedisLockRegistry. After uploading the version to 5.5.7.RELEASE, a issue occurred.


junhoi-pk:
**In what version(s) of Spring Integration are you seeing this issue?**

5.5.7.RELEASE (with spring boot version 2.6.2)

**Describe the bug**

```
        Lock lock = redisLockRegistry.obtain(key);
        boolean acquired = false;
        try {
            acquired = lock.tryLock(ttl, TimeUnit.MILLISECONDS);
            if (!acquired) {
                throw new RuntimeException(String.format("Failed to acquire lock [%s]", key));
            }
            // code
        } catch (InterruptedException e) {
            log.error(Throwables.getStackTraceAsString(e));
            Thread.currentThread().interrupt();
            throw new RuntimeException(String.format("Interrupted in locking [%s]", key));
        } finally {
            if (acquired) {
                try {
                    lock.unlock();
                } catch (Exception e) {
                    log.error(Throwables.getStackTraceAsString(e));
                }
            }
        }
```

**To Reproduce**

After updating the version, `acquired` is intermittently returned as `false` in the following code.
When I measured the time, it is presumed that it probably took a timeout and returned `false`.

It seems that unlock doesn't work properly locally, presumably intermittently.

There are several instances, but it seems that local lock release does not work properly as it only occurs on the instance with the upgraded version.

**Expected behavior**

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

artembilan:
@Meteorkor ,

would you mind to take a look here?
Feels like a regression after introduction of pub-sub recently.

Meteorkor:
@artembilan 
If a vm uses `tryLock()` and another vm `unlock` it, it will wake up with an unlock message and try to acquire the lock, but if it fails(the other vm obtain lock), vm will return false.

There seems to be a problem that returns false without waiting for the time passed to `tryLock(ttl)`.
I think we should retry until the time passed from `subscribeLock()`

And if you mix the versions, it will be a problem, but is this part okay?
Version 5.5.7 wakes up through unlock message, but lower version doesn't publish unlock message, so you may not be able to wake up.

junhoi-pk:
@Meteorkor 

Thank you for answer.
I tested it, but it returns `false` after waiting all the expire time.

Also, when I mixed it with the lower version, the problem seemed to occur, so I updated all the servers, but it occurred continuously.

[...]

PR #3702, opened 2021-12-27
GH-3700: Fix to retry until expiration time

Meteorkor:
https://github.com/spring-projects/spring-integration/issues/3700

Fixed a case where lock acquisition attempt was abandoned earlier than expected time

If a vm uses tryLock() and another vm unlock it, it will wake up with an unlock message and try to acquire the lock, but if it fails(the other vm obtain lock), vm will return false.

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -359,35 +359,38 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 		}
 
 		private boolean subscribeLock(long time) throws ExecutionException, InterruptedException {
+			if (obtainLock()) {
+				return true;
+			}
+
+			if (!RedisLockRegistry.this.redisMessageListenerContainer.isRunning()) {
+				RedisLockRegistry.this.redisMessageListenerContainer.afterPropertiesSet();
+				RedisLockRegistry.this.redisMessageListenerContainer.start();
+			}
+
 			final long expiredTime = System.currentTimeMillis() + time;
 			while (time == -1 || expiredTime >= System.currentTimeMillis()) {
-				if (!obtainLock()) {
-					if (!RedisLockRegistry.this.redisMessageListenerContainer.isRunning()) {
-						RedisLockRegistry.this.redisMessageListenerContainer.afterPropertiesSet();
-						RedisLockRegistry.this.redisMessageListenerContainer.start();
+				try {
+					Future<String> future =
+							RedisLockRegistry.this.unlockNotifyMessageListener.subscribeLock(this.lockKey);
+					//DCL
+					if (obtainLock()) {
+						return true;
 					}
 					try {
-						Future<String> future =
-								RedisLockRegistry.this.unlockNotifyMessageListener.subscribeLock(this.lockKey);
-						//DCL
-						if (!obtainLock()) {
-							try {
-								//if short expireAfter key expire for ttl, no receive unlock msg
-								long waitTime = time >= 0 ? time : RedisLockRegistry.this.expireAfter;
-								future.get(waitTime, TimeUnit.MILLISECONDS);
-							}
-							catch (TimeoutException ignore) {
-							}
-							if (!obtainLock()) {
-								continue;
-							}
-						}
+						//if short expireAfter key expire for ttl, no receive unlock msg
+						long waitTime = time >= 0 ? time : RedisLockRegistry.this.expireAfter;
+						future.get(waitTime, TimeUnit.MILLISECONDS);
 					}
-					finally {
-						RedisLockRegistry.this.unlockNotifyMessageListener.unSubscribeLock(this.lockKey);
+					catch (TimeoutException ignore) {
+					}
+					if (obtainLock()) {
+						return true;
 					}
 				}
-				return true;
+				finally {
+					RedisLockRegistry.this.unlockNotifyMessageListener.unSubscribeLock(this.lockKey);
+				}
 			}
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -359,31 +359,37 @@ public boolean tryLock(long time, TimeUnit unit) throws InterruptedException {
...
-					//DCL
-					if (!obtainLock()) {
-						try {
-							//if short expireAfter key expire for ttl, no receive unlock msg
-							long waitTime = time >= 0 ? time : RedisLockRegistry.this.expireAfter;
-							future.get(waitTime, TimeUnit.MILLISECONDS);
-						}
-						catch (TimeoutException ignore) {
+			final long expiredTime = System.currentTimeMillis() + time;
+			while (time == -1 || expiredTime >= System.currentTimeMillis()) {
```
Meteorkor, line 363: Wait a minute, This code is changing.
It seems better to move while to the bottom.

artembilan (approved):
Thank you for looking into this @Meteorkor !

Pulling locally for final review...

I think that was a big mistake to introduce this pub/sub feature in a point release.
It really causes a breaking change if mixed versions of the library is used in the cluster.

I wonder how to be more careful in the future... 😢

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -692,6 +692,63 @@ public void multiRedisLockRegistryTest() throws InterruptedException, ExecutionE
...
+			}
+		});
+
+		Thread.sleep(100);
+		tryLockReady.await();
+		lock1.unlock();
+
+		boolean await = awaitTimeout.await(2, TimeUnit.SECONDS);
+
+		assertThat(await).isFalse();
```
artembilan, line 747: I changed the code here to this to optimize a performance a bit:
```
		assertThat(tryLockReady.await(10, TimeUnit.SECONDS)).isTrue();

		lock1.unlock();

		assertThat(awaitTimeout.await(1, TimeUnit.SECONDS)).isFalse();
```
Not reason to sleep if we are good to just wait for both threads to be activated.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -692,6 +692,63 @@ public void multiRedisLockRegistryTest() throws InterruptedException, ExecutionE
...
+		Lock lock1 = registry1.obtain(testKey);
+		Lock lock2 = registry2.obtain(testKey);
+		Lock lock3 = registry3.obtain(testKey);
+		AtomicInteger expectOne = new AtomicInteger();
+
+		lock1.lock();
+		executorService.submit(() -> {
+			try {
+				tryLockReady.countDown();
+				boolean b = lock2.tryLock(3, TimeUnit.SECONDS);
```
artembilan, line 718: Changed to this `boolean b = lock2.tryLock(10, TimeUnit.SECONDS);`.

It is really possible to fail in the test with a race condition if we don't block thread long enough. 
There is no big enough time window between 2 and 3 seconds...

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -692,6 +692,63 @@ public void multiRedisLockRegistryTest() throws InterruptedException, ExecutionE
 	}
 
 
+	@Test
+	@RedisAvailable
+	public void earlyWakeUpTest() throws InterruptedException {
+		final int THREAD_CNT = 3;
```
artembilan, line 698: We use only 2 threads in the `ExecutorService`.
Would you mind to revise, please?

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -692,6 +692,63 @@ public void multiRedisLockRegistryTest() throws InterruptedException, ExecutionE
...
+		final String testKey = "testKey";
+
+		final CountDownLatch tryLockReady = new CountDownLatch(2);
+		final CountDownLatch awaitTimeout = new CountDownLatch(2);
+		final RedisConnectionFactory connectionFactory = getConnectionFactoryForTest();
+		final RedisLockRegistry registry1 = new RedisLockRegistry(connectionFactory, this.registryKey);
+		final RedisLockRegistry registry2 = new RedisLockRegistry(connectionFactory, this.registryKey);
+		final RedisLockRegistry registry3 = new RedisLockRegistry(connectionFactory, this.registryKey);
+		final ExecutorService executorService = Executors.newFixedThreadPool(THREAD_CNT);
```
artembilan, line 707: I think the `ExecutorService` has to be `shutdown()` in the end of test method to free resources.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -359,31 +359,40 @@ public boolean tryLock(long time, TimeUnit unit) throws InterruptedException {
...
+			if (obtainLock()) {
+				return true;
+			}
+
+			if (!RedisLockRegistry.this.redisMessageListenerContainer.isRunning()) {
+				RedisLockRegistry.this.redisMessageListenerContainer.afterPropertiesSet();
+				RedisLockRegistry.this.redisMessageListenerContainer.start();
+			}
+
+			final long expiredTime = System.currentTimeMillis() + time;
```
artembilan, line 371: I believe this line has to be a first in the method to calculate the remaining time properly.
The `obtainLock()` and `this.redisMessageListenerContainer.start()` is going to take some time anyway.
So, to be more accurate with expectation we need to `while()` already with the time what was left after those.

artembilan (changes requested):
Here is my review.

Thanks

artembilan:
@Meteorkor ,

thank you again for contribution!

PR #3709, opened 2022-01-03
Move IntegrationDsl from .kt to .java

artembilan:
Turns out Eclipse Kotlin plugin is not perfect at the moment,
and it causes some code parsing problems in the editor.

* Move Kotlin annotation to Java to mitigate normal source code flow:
 Java to Kotlin, not opposite


artembilan:
@garyrussell ,

please, try this in your STS.

Thanks

garyrussell:
Yes; this solves the problem.

PR #3714, opened 2022-02-01
GH-3713: TCP: Fix Intercepted Sender List

### LINKED ISSUE #3713: TcpConnectionInterceptorSupport.addNewConnection() is no longer called

5.5.8

**Describe the bug**

In 5.2.8.RELEASE, the following code worked:

```java
@Configuration
@IntegrationComponentScan
@EnableIntegration
public class MyServer {
    @Bean
    public AbstractServerConnectionFactory serverConnectionFactory() {
        TcpNetServerConnectionFactory connectionFactory = new TcpNetServerConnectionFactory(listenPort);

        TcpConnectionInterceptorFactoryChain interceptorFactoryChain = new TcpConnectionInterceptorFactoryChain();
        interceptorFactoryChain.setInterceptor(new MyInterceptorFactory());
        connectionFactory.setInterceptorFactoryChain(interceptorFactoryChain);

        return connectionFactory;
    }

    @ServiceActivator(inputChannel = "myOutput")
    @Bean
    public TcpSendingMessageHandler outbound(AbstractServerConnectionFactory connectionFactory) {
        TcpSendingMessageHandler tcpSendingMessageHandler = new TcpSendingMessageHandler();
        tcpSendingMessageHandler.setConnectionFactory(connectionFactory);
        return tcpSendingMessageHandler;
    }

    // ... etc ...

    private class MyInterceptorFactory implements TcpConnectionInterceptorFactory {
        @Override
        public TcpConnectionInterceptorSupport getInterceptor() {
            return new MyInterceptor();
        }
    }

    private class MyInterceptor extends TcpConnectionInterceptorSupport {
        @Override
        public void addNewConnection(TcpConnection connection) {
            super.addNewConnection(connection); // <=== I was called before but I'm not now!!!! ===
        }
   }
}
```

However, in 5.5.8 -- and I assume ever since [this change](https://docs.spring.io/spring-integration/reference/html/history.html#x5.4-tcp) -- the `addNewConnection()` method above is no longer called.

This is because the old code called `org.springframework.integration.ip.tcp.connection.TcpConnectionSupport.registerSender(TcpSender)` with a `MyInterceptor` instance. This called `this.theConnection.registerSender(this)` which eventually called `addNewConnection()` on the interceptor.

But the new code calls `org.springframework.integration.ip.tcp.connection.TcpConnectionSupport.registerSenders(List)`. It then walks through the list of senders and calls `addNewConnection()` on them directly, and those senders are the `TcpSendingMessageHandler` Bean(s) that were returned from  the `outbound` ServiceActivator(s) above.

Let me know if this is clear, or, if not, if I should make working and non-working samples.

Or, alternately, let me know if I'm misreading the documentation and am setting up interceptors wrong!

Thank you so much!
Dave

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3713

https://github.com/spring-projects/spring-integration/issues/3326 added support
for multiple `TcpSenders`. However, when connections are intercepted, the sender
list was not properly chained through the interceptors.

- override `registerSenders` and properly capture the real senders in the last
  interceptor and intermediate interceptors
- this ensures that `addNewConnection` is called on each interceptor
- when removing dead connections, use the (possibly intercepted) connection sender
  list insted of the factory's raw sender list.

**cherry-pick to 5.5.x**

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractConnectionFactory.java
@@ -563,10 +563,15 @@ public abstract class AbstractConnectionFactory extends IntegrationObjectSupport
 		synchronized (this.connections) {
 			Iterator<Entry<String, TcpConnectionSupport>> iterator = this.connections.entrySet().iterator();
 			while (iterator.hasNext()) {
-				TcpConnection connection = iterator.next().getValue();
+				TcpConnectionSupport connection = iterator.next().getValue();
 				connection.close();
 				iterator.remove();
-				getSenders().forEach(sender -> sender.removeDeadConnection(connection));
+				if (connection instanceof TcpConnectionInterceptorSupport) {
+					((TcpConnectionInterceptorSupport) connection).removeDeadConnection(connection);
+				}
+				else {
+					connection.getSenders().forEach(sender -> sender.removeDeadConnection(connection));
+				}
 			}
 		}
 		synchronized (this.lifecycleMonitor) {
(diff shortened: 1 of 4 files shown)


garyrussell:
Incomplete.

artembilan:

PR #3717, opened 2022-02-03
Fix @Poller Javadoc

garyrussell:
`receiveTimeout` was added as a property in 5.1.

**cherry-pick to 5.5.x**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/annotation/Poller.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/annotation/Poller.java
@@ -25,7 +25,7 @@ import java.lang.annotation.Target;
  * for the Messaging annotations for polled endpoints. It is an analogue of the XML
  * {@code <poller/>} element, but provides only simple attributes. If the
  * {@link org.springframework.integration.scheduling.PollerMetadata} requires more options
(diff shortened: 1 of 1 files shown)


spring-integration-core/src/main/java/org/springframework/integration/annotation/Poller.java
```diff
@@ -25,7 +25,7 @@
  * for the Messaging annotations for polled endpoints. It is an analogue of the XML
  * {@code <poller/>} element, but provides only simple attributes. If the
  * {@link org.springframework.integration.scheduling.PollerMetadata} requires more options
- * (e.g. Transactional and other Advices) or {@code initialDelay, receiveTimeout} etc, the
+ * (e.g. Transactional and other Advices) or {@code initialDelay} etc, the
```
artembilan, line 28: Good catch!

Would you mind to fix other typos in this JavaDoc then?

- `etc` should come with `.`
- there is a double `the` in the `errorChannel`
- `whether or not` must be without `or not`

All the stuff my IDEA suggest to me 😄

artembilan (changes requested):

artembilan:

PR #3718, opened 2022-02-04
Add a basic example to Spring Integration with Reactive Streams

migroskub:
I've seen that there is no clear example of how to quickly get started with Spring Integration and Reactive Streams. This PR tries to help with this.

I'm trying to help mainly for questions like [this](https://stackoverflow.com/questions/70942419/how-to-create-a-reactive-inbound-channel-adapter-in-spring-integration-reactor).

### DIFF
--- a/src/reference/asciidoc/reactive-streams.adoc
+++ b/src/reference/asciidoc/reactive-streams.adoc
@@ -245,14 +245,9 @@ public class CustomReactiveMessageHandler extends AbstractReactiveMessageHandler
 
     private final CustomEntityOperations customEntityOperations;
 
-    private Expression queryTypeExpression = new ValueExpression<>(Type.INSERT);
-
-    public CustomReactiveMessageHandler(CustomEntityOperations customEntityOperations, Expression queryTypeExpression) {
+    public CustomReactiveMessageHandler(CustomEntityOperations customEntityOperations) {
         Assert.notNull(customEntityOperations, "'customEntityOperations' must not be null");
         this.customEntityOperations = customEntityOperations;
-        if (queryTypeExpression != null) {
-            this.queryTypeExpression = queryTypeExpression;
-        }
     }
 
     @Override
--- a/src/reference/asciidoc/reactive-streams.adoc
+++ b/src/reference/asciidoc/reactive-streams.adoc
@@ -191,6 +191,32 @@ public class CustomReactiveMessageSource extends MessageProducerSupport {
 }
 ```
 
+And now we will be able to use it:
+```java
+public class MainFlow {
+  @Autowired
+  private CustomReactiveSourceMessageSource customReactiveSourceMessageSource;
+  
+  @Bean
+  public IntegrationFlow buildFlow() {
+     return IntegrationFlows.from(customReactiveSourceMessageSource)
+        .channel(outputChannel)
+        .get();
+  }
+}
+```
+Or in a declarative way:
+```java
+public class MainFlow {  
+  @Bean
+  public IntegrationFlow buildFlow() {
+     return IntegrationFlows.from(new CustomReactiveSourceMessageSource(new CustomReactiveSource()))
+        .handle(outputChannel)
+        .get();
+  }
+}
+``
+
 A reactive outbound channel adapter implementation is about the initiation (or continuation) of a reactive stream to interaction with an external system according to the provided reactive API for the target protocol.
 An inbound payload could be a reactive type per se or as an event of the whole integration flow which is a part of the reactive stream on top.
 A returned reactive type can be subscribed immediately if we are in a one-way, fire-and-forget scenario, or it is propagated downstream (request-reply scenarios) for further integration flow or an explicit subscription in the target business logic, but still downstream preserving reactive streams semantics.
@@ -243,14 +269,20 @@ public class CustomReactiveMessageHandler extends AbstractReactiveMessageHandler
 }
 ```
 
-And now we will be able to use it:
+And now we will be able to use both of the channel adatpers:
 ```java
 public class MainFlow {
+  @Autowired
+  private CustomReactiveSourceMessageSource customReactiveSourceMessageSource;
+  
+  @Autowired
+  private CustomReactiveMessageHandler customReactiveMessageHandler;
   
   @Bean
   public IntegrationFlow buildFlow() {
-     return IntegrationFlows.from(new ReactiveSourceMessageProducer(new ReactiveSource()))
-        .channel(outputChannel)
+     return IntegrationFlows.from(customReactiveSourceMessageSource)
+        .transform(someOperation)
+        .handle(customReactiveMessageHandler)
         .get();
   }
 }
--- a/src/reference/asciidoc/reactive-streams.adoc
+++ b/src/reference/asciidoc/reactive-streams.adoc
@@ -160,22 +160,19 @@ When the target protocol for integration provides a Reactive Streams solution, i
 An inbound, event-driven channel adapter implementation is about wrapping a request (if necessary) into a deferred `Mono` or `Flux` and perform a send (and produce reply, if any) only when a protocol component initiates a subscription into a `Mono` returned from the listener method.
 This way we have a reactive stream solution encapsulated exactly in this component.
 Of course, downstream integration flow subscribed on the output channel should honor Reactive Streams specification and be performed in the on demand, back-pressure ready manner.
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Thank you for a sample!

Please, consider to move it *before* `A reactive outbound channel adapter` paragraph.

Any samples for the outbound part?

The `R2dbcMessageHandler` should be a good one took into.

artembilan:
@migroskub ,

Gentle ping.
Otherwise we will take it from here ourselves and will merge it as we think it should be.

Thank you!

artembilan:
Thank you for feedback, @migroskub !

1. The Function Support is part of Messaging Endpoints chapter. So, it is indeed cannot be visible from the main page. You need to go to this one to see it in the content on the left: https://docs.spring.io/spring-integration/reference/html/messaging-endpoints.html#messaging-endpoints-chapter. Same applies for may others paragraphs which are part for bigger chapters. Perhaps if you go to a single page, it would become clearer for you: https://docs.spring.io/spring-integration/docs/current/reference/html/index-single.html
2. I will take a look into that tool.
3. Please, raise a separate GH issue for that channel adapters development concern.

src/reference/asciidoc/reactive-streams.adoc
```diff
@@ -160,12 +160,101 @@ When the target protocol for integration provides a Reactive Streams solution, i
...
+    }
+
+    public enum Type {
+        INSERT,
+        UPDATE,
+    }
+}
+```
+
+And now we will be able to use it:
```
artembilan, line 246: I think this flow sample belongs to the `producer` section above.
The `CustomReactiveMessageHandler` deserves its own independent usage sample.
Otherwise it is confusing a little bit.

artembilan (changes requested):
Thanks

src/reference/asciidoc/reactive-streams.adoc
```diff
@@ -160,12 +160,134 @@ When the target protocol for integration provides a Reactive Streams solution, i
...
+        Assert.notNull(customEntityOperations, "'customEntityOperations' must not be null");
+        this.customEntityOperations = customEntityOperations;
+        if (queryTypeExpression != null) {
+            this.queryTypeExpression = queryTypeExpression;
+        }
+    }
+
+    @Override
+    protected Mono<Void> handleMessageInternal(Message<?> message) {
+        return Mono.fromSupplier(() -> this.queryTypeExpression.getValue(this.evaluationContext, message, Type.class))
```
migroskub, line 243: @artembilan I'm struggling with the second example - can you simplify this to a more basic supplier usage that will be practical and won't require `evaluationContext`?

src/reference/asciidoc/reactive-streams.adoc
```diff
@@ -160,12 +160,134 @@ When the target protocol for integration provides a Reactive Streams solution, i
...
+
+This is not always available by the nature (or with the current implementation) of `MessageHandler` processor used in the integration flow.
 This limitation can be handled using thread pools and queues or `FluxMessageChannel` (see above) before and after integration endpoints when there is no reactive implementation.
 
-A reactive outbound channel adapter implementation is about initiation (or continuation) of a reactive stream to interaction with an external system according provided reactive API for the target protocol.
-An inbound payload could be a reactive type per se or as an event of the whole integration flow which is a part of reactive stream on top.
-A returned reactive type can be subscribed immediately if we are in one-way, fire-and-forget scenario, or it is propagated downstream (request-reply scenarios) for further integration flow or an explicit subscription in the target business logic, but still downstream preserving reactive streams semantics.
+An example for a reactive inbound channel adapter:
+```java
+public class CustomReactiveMessageSource extends MessageProducerSupport {
```
artembilan, line 169: It's not correct to call it `MessageSource`. Just because it does not implement a `MessageSource`. And definitely a reactive publisher could not be wrapped to the `MessageSource` anyway.
Better to call it `CustomReactiveMessageProducer` since you really do it correctly extending `MessageProducerSupport` and using its `subscribeToPublisher()` from the `doStart()` impl.

src/reference/asciidoc/reactive-streams.adoc
```diff
@@ -160,12 +160,134 @@ When the target protocol for integration provides a Reactive Streams solution, i
...
+
+    private final CustomReactiveSource customReactiveSource;
+
+    public CustomReactiveSourceMessageProducer(CustomReactiveSource customReactiveSource) {
```
artembilan, line 173: I see that constructor for the class you called properly. So, please, revise the name of the class 😄

src/reference/asciidoc/reactive-streams.adoc
migroskub: Can be resolved

src/reference/asciidoc/reactive-streams.adoc
artembilan: Just do `-> message.getHeaders().get("queryType", Type.class)`.
And no `queryTypeExpression` in the code at all!

artembilan (changes requested):

src/reference/asciidoc/reactive-streams.adoc
```diff
@@ -212,12 +212,29 @@ Or in a declarative way:
...
+```java
+public class MainFlow {  
+  @Bean
+  public IntegrationFlow buildFlow() {
+     return IntegrationFlows.from(this.customReactiveSource
+                .map(event - >
+                    MessageBuilder
+                    .withPayload(event.getBody())
+                    .setHeader(MyReactiveHeaders.SOURCE_NAME, event.getSourceName())
+                    .build());)
```
artembilan, line 231: The `;` is wrong here.
I would prefer to have this `Flux` extracted as a variable. It's OK to have it in this bean method definition.

artembilan (changes requested):

src/reference/asciidoc/reactive-streams.adoc
migroskub: Do you mean like this?
````java
public class MainFlow {  
  @Bean
  public IntegrationFlow buildFlow() {
    Flux<Message<?>> myFlux = this.customReactiveSource
                .map(event - >
                    MessageBuilder
                    .withPayload(event.getBody())
                    .setHeader(MyReactiveHeaders.SOURCE_NAME, event.getSourceName())
                    .build());
     return IntegrationFlows.from(myFlux)
        .handle(outputChannel)
        .get();
  }
}
````

Comment (artembilan) [reply src/reference/asciidoc/reactive-streams.adoc:231]: Yep! This is good. It does give us a clue what is that `customReactiveSource` about and which `from()` factory is going to be for building an `IntegrationFlow`.

**Comment by migroskub** [comment]
I think the PR is ready

File: src/reference/asciidoc/reactive-streams.adoc [hunk 6]
```diff
@@ -160,12 +160,152 @@ When the target protocol for integration provides a Reactive Streams solution, i
...
+    private final CustomReactiveSource customReactiveSource;
+
+    public CustomReactiveMessageProducer(CustomReactiveSource customReactiveSource) {
+        Assert.notNull(customReactiveSource, "'customReactiveSource' must not be null");
+        this.customReactiveSource = customReactiveSource;
+    }
+
+    @Override
+    protected void doStart() {
+        Flux<Message<<?>> messageFlux =
```
artembilan, line 180: Typo for diamond definition. Has lead to this error:
```
 Execution failed for task ':checkAsciidocLinks'.
> [*** Anchor reference errors found:, 
  Anchor '?' not found in 'reactive-streams.adoc', if in another file, it needs to be qualified with './fileName.adoc#']
```

But originally I definitely have missed it 😄

artembilan (changes requested):

src/reference/asciidoc/reactive-streams.adoc
artembilan: This code is wrong: `Flux<Message<<?>>`.
Must be `Flux<Message<?>>`. You have to have exact number of closing bracket as many as you have opening.
The error is misleading, but it really helped me to spot the typo in your code.

artembilan (changes requested):

src/reference/asciidoc/reactive-streams.adoc
migroskub: Fixed

artembilan (approved):

artembilan:
Thank you, @migroskub; looking forward for more!

PR #3721, opened 2022-02-08
Fix a mistake in a sentence about EIP components and BeanPostProcessor

almogtavor:
A minor fix for a wrong sentence in the docs.


artembilan:

@almogtavor ,

Thank you for the catch and fix; looking forward for more!

PR #3728, opened 2022-02-15
Remove SocketUtils usage

artembilan:
* Remove usage of non-stable `org.springframework.util.SocketUtils`
* Replace it with `0` for those tests where it is possible to select OS port
* Remove the mentioning of the `SocketUtils` from the `testing.adoc`
* Use `TransportConstants.DEFAULT_STOMP_PORT` for `StompServerIntegrationTests`.
We may disable this test in the future for CI if it is not going to be stable
* Introduce `Supplier<String> connectUrl` variants for `ZeroMqMessageHandler`
to let it defer connection evaluation until subscription to the socket `Mono`
in the `ZeroMqMessageHandler`.
* Move connection logic in the `ZeroMqMessageHandler` to `Lifecycle.start()`

Related to https://github.com/spring-projects/spring-framework/issues/28054

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-stomp/src/test/java/org/springframework/integration/stomp/client/StompServerIntegrationTests.java
+++ b/spring-integration-stomp/src/test/java/org/springframework/integration/stomp/client/StompServerIntegrationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2021 the original author or authors.
+ * Copyright 2015-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.config.impl.ConfigurationImpl;
+import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
 import org.apache.activemq.artemis.core.server.embedded.EmbeddedActiveMQ;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
 import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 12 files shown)


artembilan:
@garyrussell ,

I'd appreciate if you test it locally on your OS.
Not sure why we have used an explicit port for those UDP tests all this time...

Thanks

spring-integration-stomp/src/test/java/org/springframework/integration/stomp/client/StompServerIntegrationTests.java
```diff
@@ -73,22 +73,21 @@
...
-		int port = SocketUtils.findAvailableTcpPort(61613);
 		ConfigurationImpl configuration =
 				new ConfigurationImpl()
 						.setName("embedded-server")
 						.setPersistenceEnabled(false)
 						.setSecurityEnabled(false)
 						.setJMXManagementEnabled(false)
 						.setJournalDatasync(false)
-						.addAcceptorConfiguration("stomp", "tcp://127.0.0.1:" + port)
+						.addAcceptorConfiguration("stomp", "tcp://127.0.0.1:" + TransportConstants.DEFAULT_STOMP_PORT)
```
garyrussell, line 83: It is better to use a flaky `SocketUtils` rather than hard coding to a single port if the component doesn't support `0`.

I'd prefer disabling (per your comment), if you want to remove it completely.

garyrussell (changes requested):
Unfortunately, UDP doesn't work on my new MacBook and I haven't had time to figure out why, but our friends at GH Actions have verified it.

I don't see how we can cherry-pick this (except just the tests) because it is a documented class - it needs to be deprecated in 5.5.x.

artembilan:
>  it needs to be deprecated in 5.5.x.

There is nothing to deprecate for us: we use already that one from Spring Framework.
It is used only in the tests therefore the cherry-pick it straight forward without any impact on the production code.
Plus that ZeroMQ has gotten a new feature to determine the port on demand 😄 .

spring-integration-stomp/src/test/java/org/springframework/integration/stomp/client/StompServerIntegrationTests.java
artembilan: There is no such a class in SF 6.0 any more. 
So, I wouldn't mind to watch this until it fails with the port conflict 😄

garyrussell (approved):

garyrussell:
Stomp needs back-porting for 5.5.x; Artemis is not present there.

artembilan:
Thanks. On it.

PR #3729, opened 2022-02-16
Issue 3501

dmfrey:


spring-integration-graphql/src/test/java/org/springframework/integration/graphql/outbound/GraphQlMessageHandlerTests.java
```diff
@@ -0,0 +1,424 @@
...
+
+		@Override
+		public String toString() {
+			return "QueryResult{" +
+					"id='" + id + '\'' +
+					'}';
+		}
+	}
+
+	static class Update {
```
artembilan, line 394: These two last classes could be as simple as:
```
	record QueryResult(String id) {	}

	record Update(String id) { 	}
```
No need in that extra boilerplate code 😄 !

artembilan (changes requested):

spring-integration-graphql/src/main/java/org/springframework/integration/graphql/outbound/GraphQlMessageHandler.java
```diff
@@ -54,7 +57,7 @@
 	@Nullable
 	private Locale locale;
 
-	private String executionId = null;
+	private Expression idExpression = new SupplierExpression<>(() -> null);
```
artembilan, line 60: Let's make it more clear, so the name should be an `executionIdExpression`.
And this one can be as default expression: `new FunctionExpression<Message<?>>(message -> message.getHeaders().getId())`

artembilan (changes requested):

spring-integration-graphql/src/main/java/org/springframework/integration/graphql/outbound/GraphQlMessageHandler.java
```diff
@@ -119,12 +119,12 @@ public void setLocale(@Nullable Locale locale) {
...
 	/**
-	 * Set a SpEL expression to evaluate Id for GraphQL Operation Request to execute.
-	 * @param idExpression the idExpression to use.
+	 * Set a SpEL expression to evaluate Execution Id for GraphQL Operation Request to execute.
+	 * @param executionIdExpression the executionIdExpression to use.
 	 */
-	public void setIdExpression(Expression idExpression) {
-		Assert.notNull(idExpression, "'idExpression' must not be null");
-		this.idExpression = idExpression;
+	public void setIdExpression(Expression executionIdExpression) {
```
artembilan, line 125: Rename, please, the setter, too 😄

spring-integration-graphql/src/main/java/org/springframework/integration/graphql/outbound/GraphQlMessageHandler.java
```diff
@@ -143,11 +143,11 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
 		}
 		else  {
 			Assert.notNull(this.operationExpression, "'operationExpression' must not be null");
-			Assert.notNull(this.idExpression, "'idExpression' must not be null");
+			Assert.notNull(this.executionIdExpression, "'executionIdExpression' must not be null");
```
artembilan, line 146: We don't need this because it is not null by default and null cannot pass the setter.

artembilan (changes requested):

artembilan (approved):
Looking into this one more time locally and probably merging...

Thank you so far!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f09c665db4291854765a536f50e3cfda1ed0369c.

@dmfrey ,

thank you very much; looking forward for more!

dmfrey:
> Merged as [f09c665](https://github.com/spring-projects/spring-integration/commit/f09c665db4291854765a536f50e3cfda1ed0369c).
> 
> @dmfrey ,
> 
> thank you very much; looking forward for more!

No problem. Glad I could contribute.

PR #3738, opened 2022-03-06
Add serialVersionUID to MessageHistory

### LINKED ISSUE #3737: SerialVersionUID on class MessageHistory Breaking change : from version 5.5.2 to latest

After the version 5.5.2 until now

**bug**

The Class MessageHistory is a serializable class that doesn't have a SerialVersionUID defined, because of that it was taken the default SerialVersionUID created with the previous version.

Now this class was updated in the commit :
 https://github.com/spring-projects/spring-integration/commit/af9e69c251751ad60c7ccba5ec6ace3140d8c758

Where the equals/hashCode was implemented
Because of that, after the version 5.5.2 the serialVersionUID changed

This change of serialVersionUID makes the previos version incompatible with the new one, because the serialVersionUID matching fails when it tries to deserialize the class.

**To Reproduce**

Use spring integration with the version 5.5.1 to store a new record, and then update the version to the 5.5.2 or latest to read this record, and it will fails

**Expected behavior**

Deserialize the class MessageHistory stored in previous versions to the 5.5.2

Is it possible to fix this problem ?
Is there any workaround that we can implement in the meantime ?

Thanks a lot

onobc:
@artembilan not sure what test to add for this one. We could do this:
- Serialize a 5.1 MessageHistory to disk.
- Put that file in `test/resources`
- Deserialize the file w/ the latest `MessageHIistory`.
I'm not sure if this is worth adding to automation though. It is somewhat a one-off and should not be an issue again in this class. 

Fixes gh-3737

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/history/MessageHistory.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/history/MessageHistory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,11 +49,13 @@ import com.fasterxml.jackson.annotation.JsonCreator;
  * @author Mark Fisher
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Chris Bono
  *
  * @since 2.0
  */
-@SuppressWarnings("serial")
 public final class MessageHistory implements List<Properties>, Serializable {
+
+	private static final long serialVersionUID = -2340400235574314134L;
 
 	private static final Log LOGGER = LogFactory.getLog(MessageHistory.class);
 
@@ -303,11 +305,12 @@ public final class MessageHistory implements List<Properties>, Serializable {
 		return entry;
 	}
 
-
 	/**
 	 * Inner class for each Entry in the history.
 	 */
 	public static class Entry extends Properties {
+
+		private static final long serialVersionUID = -8225834391885601079L;
 
 		public String getName() {
 			return this.getProperty(NAME_PROPERTY);


artembilan (changes requested):
Yes, I agree it is pointless to waste our time figuring out how to test it in JUnit.

It is really would be enough to test it once.
We may just accept this PR and merge. 
And then ask @german22 to test his project against latest `5.5.10-SNAPSHOT`.

Just check the project and other similar entities are OK, even if they just only have `serialVersionUID = 1L`.

Although I feel as we have to add an explicit one to all the `ApplicationEvent` impls, and exceptions, like `ReplyRequiredException`...

WDYT?

artembilan (changes requested):
After removing that suppression we got now this:
```
/home/runner/work/spring-integration/spring-integration/spring-integration-core/src/main/java/org/springframework/integration/history/MessageHistory.java:312: warning: [serial] serializable class Entry has no definition of serialVersionUID
	public static class Entry extends Properties {
	              ^
```

So, it has to be fixed with an explicit (extracted) `serialVersionUID` as well.

artembilan:
We are not ready to merge this yet.
Even if this will fix data stored from the `5.5.1`. 
But what about the data stored already starting with `5.5.2`?

We have to come up somehow with a solution which could deal with old and new class versions...

spring-integration-core/src/main/java/org/springframework/integration/support/serializer/VersionTolerantJavaDeserializer.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+
+/**
+ * A {@link Deserializer} implementation that reads an input stream using Java serialization in a
+ * {@link VersionTolerantObjectInputStream serialVersionUID tolerant} manner.
+ *
+ * <pre>
+ * <code>
+ * 	VersionTolerantJavaDeserializer deser = new VersionTolerantJavaDeserializer();
```
onobc, line 33: @artembilan this is the workaround we discussed. 

Because the `serialVersionUID` is verified very early in the `ObjectInputStream` (namely before readObject on the actual class) the only technique to allow us to inspect and decide on that UID is by overriding [readClassDescriptor](https://github.com/spring-projects/spring-integration/pull/3738/files#diff-4dccaf3dfbd729f0f6c2cdde477a3e9440eab9539c23903498e65d620bf620f6R43). 

This makes this more of an intrusive scenario as we have to figure out how to wedge in the custom OIS. We do not want to do this in the framework. We do, however want to give users (@german22) a workaround to use if they so choose until 6.0.0 comes out where we will make sure all serializables have UID fields (breaking change though). 

@artembilan I am not sure what the delivery vehicle for this would be. We can chat more about this tomorrow. 

The workaround is thoroughly tested as I did not want to provide this w/o knowing it was solid. Usage is described below. The idea being one would have to hook in the custom OIS into whatever store they are using that needs the version tolerant deserialization.

artembilan (changes requested):
That's great, @onobc !

But I believe we should leave such a workaround outside of the framework code even for the current version we are going to fix it for.

Better to have a single workaround class (even with that `VersionTolerantObjectInputStream` as nested) and show it in the issue for end-user.

The fix has to be like an explicit for the current `MessageHistory` class version - `serialVersionUID = -2340400235574314134L`.

So, it is not going to be a break point in the next `6.0`.

(I don't think we will migrate a `MessageHistory` to the `record` as IDE suggests...)

onobc:
> That's great, @onobc !
> 
> But I believe we should leave such a workaround outside of the framework code even for the current version we are going to fix it for.
>
 > Better to have a single workaround class (even with that `VersionTolerantObjectInputStream` as nested) and show it in the issue for end-user.
> 

Sounds good @artembilan . Yep, I was not wanting to actually merge it in - but figure out where to put it. The docs idea in a single class sounds good.  I will put it in a branch locally (just to keep around for later if needed), consolidate it into single class, and produce a HowTo in the issue w/ the code sample. 

I will close this PR once I update the issue w/ the sample code. 

> The fix has to be like an explicit for the current `MessageHistory` class version - `serialVersionUID = -2340400235574314134L`.
> 

Yep, that was in the javadoc at the top of the deserializer on how to use. I will make it clear in the issue when I comment. 

> So, it is not going to be a break point in the next `6.0`.
> 
> (I don't think we will migrate a `MessageHistory` to the `record` as IDE suggests...)

Yeh - no changes to framework should come out of this PR at all.

artembilan:
I mean, Chris, that current `MessageHistory` has to got that `serialVersionUID = -2340400235574314134L`. And that's exactly what we will merge and back-port.
It is not going to be a breaking change at all: just will let us to avoid similar issue in the future when we would change something in the `MessageHistory` breaking its class version.

onobc:
> I mean, Chris, that current `MessageHistory` has to got that `serialVersionUID = -2340400235574314134L`. And that's exactly what we will merge and back-port. It is not going to be a breaking change at all: just will let us to avoid similar issue in the future when we would change something in the `MessageHistory` breaking its class version.

Ahh yeh. Makes sense. Explicitly set the `serialVersionUID` in MH to the value it implicitly is (currently). I will amend this PR to do that.

onobc:
Ignore the latest commits. I intended those to a separate branch for safe-keeping. I will clean this up shortly.

spring-integration-core/src/main/java/org/springframework/integration/history/MessageHistory.java
```diff
@@ -49,12 +49,14 @@
...
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Chris Bono
  *
  * @since 2.0
  */
-@SuppressWarnings("serial")
 public final class MessageHistory implements List<Properties>, Serializable {
 
+	private static final long serialVersionUID = -2340400235574314134L;
```
onobc, line 58:
```
cbono@cbono-a01 server-main % serialver -classpath spring-context-5.3.14.jar:spring-context-support-5.3.14.jar:spring-core-5.3.14.jar:spring-integration-core-5.5.8-SNAPSHOT.jar:spring-jcl-5.3.14.jar:spring-messaging-5.3.14.jar org.springframework.integration.history.MessageHistory
org.springframework.integration.history.MessageHistory:    private static final long serialVersionUID = -2340400235574314134L;
```

spring-integration-core/src/main/java/org/springframework/integration/history/MessageHistory.java
```diff
@@ -303,12 +305,13 @@ private static Properties extractMetadata(NamedComponent component) {
 		return entry;
 	}
 
-
 	/**
 	 * Inner class for each Entry in the history.
 	 */
 	public static class Entry extends Properties {
 
+		private static final long serialVersionUID = -8225834391885601079L;
```
onobc, line 313:
```
cbono@cbono-a01 server-main % serialver -classpath spring-context-5.3.14.jar:spring-context-support-5.3.14.jar:spring-core-5.3.14.jar:spring-integration-core-5.5.8-SNAPSHOT.jar:spring-jcl-5.3.14.jar:spring-messaging-5.3.14.jar org.springframework.integration.history.MessageHistory.Entry
org.springframework.integration.history.MessageHistory.Entry:    private static final long serialVersionUID = -8225834391885601079L;
```

artembilan:
... and cherry-picked to `5.5.x`.

Thanks, @onobc ; looking forward for more!

PR #3741, opened 2022-03-08
Add Gitpod support

hogah97323:
Add Gitpod support so it will be possible to run the project with Gitpod and easily help with contributions


artembilan:
I'm not sure why do we need this...
Looks like the preference is a Codespaces instead: https://github.com/spring-projects/spring-petclinic/pull/514#issuecomment-934391253.

Please, elaborate: I don't see that any Spring project does this. So, I'm also reluctant to accept such a change.

hogah97323:
@artembilan seems like Codespaces is not free. Waiting for the response from @dsyer to understand more.

dsyer:
I think the argument for encouraging better contributions is good. Unfortunately I don't think many people will be happy giving Gitpod write access to their repositories (which it needs apparently). Codespaces might be a better option if Microsoft ever make it generally available.

hogah97323:
Ok. I'd like that to be reconsidered in case Microsoft won't open source Codespaces

PR #3746, opened 2022-03-10
Revise @Poller usage to single entry in Messaging Annotations.

### LINKED ISSUE #3506: Revise @Poller usage to a single entry in the Messaging Annotations
See discussion for more info: https://github.com/spring-projects/spring-integration/pull/3503#discussion_r588582368 and new `@Reactive` for an inspiration. Therefore `AbstractMethodAnnotationPostProcessor` logic has to be changed to deal with a single value and parse a default one properly.

onobc:
Fixes #3506


src/reference/asciidoc/changes-5.5-6.0.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+[[x6.0-new-components]]
+=== New Components
+TODO
+
+[[x6.0-general]]
+=== General Changes
+
+[[x6.0-poller]]
+==== @Poller
```
artembilan, line 12: This info has to be added to the `whats-new.adoc`. Because this is exactly what is new in this `6.0` version.
We will create a `changes-5.5-6.0.adoc` file, when we start `6.1` version.

The point is to let end-users to find easily what is new in the project just from a first page of our doc.

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingAnnotationUtils.java
```diff
@@ -69,9 +70,31 @@
...
+		// Empty array
+		if (ObjectUtils.isArray(annotationValue) && ObjectUtils.isEmpty(annotationValue)) {
+			return false;
+		}
+		// Empty string
+		if ((annotationValue instanceof String) && ObjectUtils.isEmpty(annotationValue)) {
+			return false;
+		}
+		// Poller with special 'none' value
+		if ((annotationValue instanceof Poller) && ValueConstants.DEFAULT_NONE.equals(((Poller) annotationValue).value())) {
```
artembilan, line 94: This is not OK. 
Better really to have an utility method as generics as possible.
I mean I believe there is a way to get a value from the annotation without casting to some specific type.
For example see the same `@Reactive` use-case:
```
Reactive reactive() default @Reactive(ValueConstants.DEFAULT_NONE);
```
It is fully similar to what we have so far with the `@Poller`.


Also: if we do this then we don't need that extra check `if (poller != null && !ValueConstants.DEFAULT_NONE.equals(poller.value())) {` in the `AbstractMethodAnnotationPostProcessor.java`

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingAnnotationUtils.java
```diff
@@ -69,9 +70,31 @@
 		return null;
 	}
 
-	public static boolean hasValue(Object value) {
-		return value != null && (!(value instanceof String) || (StringUtils.hasText((String) value))) // NOSONAR
-				&& (!value.getClass().isArray() || ((Object[]) value).length > 0);
+	/**
+	 * Determines if the value of a named attribute from an annotation instance contains an actual value.
```
artembilan, line 74: The method JavaDoc must be imperative, kinda commanding us what to do.
No blank lines in the JavaDoc: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

artembilan (changes requested):
Thanks, Chris.

Please, find some review on the lines.

spring-integration-core/src/main/java/org/springframework/integration/util/MessagingAnnotationUtils.java
onobc: Yep, I meant to add a comment on this controversial section. The alternative approach I was going to talk about would be to consider any annotation whose "value" field was set to `ValueConstants.DEFAULT_NONE` to be considered "empty". This would take the `Poller` or `Reactive` type out of the equation. This would also handle `@Reactive` annotation extension properly. 

> For example see the same @Reactive use-case: 
> It is fully similar to what we have so far with the @Poller.

AFAICT we do not handle `@Reactive` override in annotation hierarchy properly (you will see it is not in the EnableIntegrationTests). We would need to add the special handling for `@Reactive` as well. 

I will amend the proposal w/ the above approach and we can see how that looks.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -36,4 +36,21 @@ See <<./http.adoc#http,HTTP Support>> for more information.
...
+- `@BridgeFrom`
+- `@BridgeTo`
+- `@Filter`
+- `@InboundChannelAdapter`
+- `@Router`
+- `@ServiceActivator`
+- `@Splitter`
+- `@Transformer`
+
+used to contain a `Poller[]` attribute but now contain a single element. The array was only in place  to allow an empty default (no poller). This is now accomplished with a single element using the special value of `ValueConstants.DEFAULT_NONE`.
```
artembilan, line 56: Hm. I thought I send you a suggestion how to make it clean and simple:

> The messaging annotations don't require a `poller` attribute as an array of `@Poller` any more.
See <<./configuration.adoc#configuration-using-poller-annotation,Using the `@Poller` Annotation>> for more information.

And I believe this one has to be as a plain sentence in the existing `=== General Changes` section: no need for its own since it is really a general one.

artembilan (changes requested):

artembilan:
FYI: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.x-to-6.0-Migration-Guide#no-array-for-poller-annotation-attribute

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -363,24 +364,26 @@ protected AbstractEndpoint createEndpoint(MessageHandler handler, @SuppressWarni
 	protected AbstractEndpoint doCreateEndpoint(MessageHandler handler, MessageChannel inputChannel,
 			List<Annotation> annotations) {
 
-		Poller[] pollers = MessagingAnnotationUtils.resolveAttribute(annotations, "poller", Poller[].class);
+		Poller poller = MessagingAnnotationUtils.resolveAttribute(annotations, "poller", Poller.class);
+		boolean pollerProvided = poller != null && !ValueConstants.DEFAULT_NONE.equals(poller.value());
```
artembilan, line 368: Since the `ValueConstants.DEFAULT_NONE` is covered now in the `MessagingAnnotationUtils.hasValue()`, I guess we don't need it here any more.

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -414,19 +417,15 @@ private ReactiveStreamsConsumer reactiveStreamsConsumer(MessageChannel channel,
...
 
-	protected void configurePollingEndpoint(AbstractPollingEndpoint pollingEndpoint, Poller[] pollers) {
+	protected void configurePollingEndpoint(AbstractPollingEndpoint pollingEndpoint, Poller poller) {
 		PollerMetadata pollerMetadata;
-		if (!ObjectUtils.isEmpty(pollers)) {
-			Assert.state(pollers.length == 1,
-					"The 'poller' for an Annotation-based endpoint can have only one '@Poller'.");
-			Poller poller = pollers[0];
-
+		if (poller != null && !ValueConstants.DEFAULT_NONE.equals(poller.value())) {
```
artembilan, line 428: DITTO

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
onobc: It is used [here](https://github.com/spring-projects/spring-integration/pull/3746/files#diff-4968e3d7ef5068b890a520af47f1fc8da4b9c369f7af56fbb05efbd5d07f070bR373) to ensure XOR on reactive|poller though.

onobc: Yeh it is now simplified. Good code review @artembilan !!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/f3d3694836150312e198c05de06e755105af2039 after some commit message clean up.

Thank you, Chris, again!

Issue #3732, opened 2022-02-25
Mqttv5PahoMessageDrivenChannelAdapter.addTopic(topic, qos) throws NPE


mrpiggi:
**In what version(s) of Spring Integration are you seeing this issue?**

5.5.9 and prior (5.5.5)

**Describe the bug**

Creating a new `Mqttv5PahoMessageDrivenChannelAdapter` and using `addTopic(topic, qos)` before `onInit()` was called throws a NPE as `this.mqttClient.subscribe(topic, qos)` is invoked. Same with `Mqttv5PahoMessageDrivenChannelAdapter.removeTopic(String...)` and `this.mqttClient.unsubscribe(topic)`.

**To Reproduce**
```java
final Mqttv5PahoMessageDrivenChannelAdapter adapter = new Mqttv5PahoMessageDrivenChannelAdapter(options, id);
adapter.addTopic(topic, 1);
```

**Workaround**
```java
final Mqttv5PahoMessageDrivenChannelAdapter adapter = new Mqttv5PahoMessageDrivenChannelAdapter(options, id, topic);
adapter.setQos(1);
```

**Expected behavior**

Topic registration or removal without calling `IMqttAsyncClient.subscribe(topic, qos)` or `IMqttAsyncClient.unsubscribe(String...)` when not initialized yet.

The comparison between `Mqttv5PahoMessageDrivenChannelAdapter` and `MqttPahoMessageDrivenChannelAdapter` is quite obvious.

https://github.com/spring-projects/spring-integration/blob/8c57cb78697bec3af71d508d4301fc0e067e97ec/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java#L214-L242

https://github.com/spring-projects/spring-integration/blob/8c57cb78697bec3af71d508d4301fc0e067e97ec/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java#L222-L255

artembilan:
Confirmed. 

Feel free to contribute the fix: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc

Thank you!

mrpiggi:
[...]

PR #3747, opened 2022-03-14
GH-3732 make mqttClient initialize lazily

mipo256:
Fixing #3732 issue. Also simplified the `Topic#equals()` method.

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.integration.mqtt.inbound;
 
 import java.util.LinkedHashSet;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -331,15 +332,7 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter extends MessagePro
 				return false;
 			}
 			Topic other = (Topic) obj;
-			if (this.topic == null) {
-				if (other.topic != null) {
-					return false;
-				}
-			}
-			else if (!this.topic.equals(other.topic)) {
-				return false;
-			}
-			return true;
+			return Objects.equals(this.topic, other.topic);
 		}
 
 		@Override
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -65,6 +65,7 @@ import org.springframework.util.Assert;
  * See {@link #setPayloadType} for more information about type conversion.
  *
  * @author Artem Bilan
+ * @author Mikhail Polivakha
  *
  * @since 5.5.5
  *
@@ -141,13 +142,15 @@ public class Mqttv5PahoMessageDrivenChannelAdapter extends AbstractMqttMessageDr
 	@Override
 	protected void onInit() {
 		super.onInit();
-		try {
-			this.mqttClient = new MqttAsyncClient(getUrl(), getClientId(), this.persistence);
-			this.mqttClient.setCallback(this);
-			this.mqttClient.setManualAcks(isManualAcks());
-		}
-		catch (MqttException ex) {
-			throw new BeanCreationException("Cannot create 'MqttAsyncClient' for: " + getComponentName(), ex);
+		if (this.mqttClient == null) {
+			try {
+				this.mqttClient = new MqttAsyncClient(getUrl(), getClientId(), this.persistence);
+				this.mqttClient.setCallback(this);
+				this.mqttClient.setManualAcks(isManualAcks());
+			}
+			catch (MqttException ex) {
+				throw new BeanCreationException("Cannot create 'MqttAsyncClient' for: " + getComponentName(), ex);
+			}
 		}
 		if (this.messageConverter == null) {
 			setMessageConverter(getBeanFactory()
@@ -189,8 +192,10 @@ public class Mqttv5PahoMessageDrivenChannelAdapter extends AbstractMqttMessageDr
 		this.topicLock.lock();
 		String[] topics = getTopic();
 		try {
-			this.mqttClient.unsubscribe(topics).waitForCompletion(getCompletionTimeout());
-			this.mqttClient.disconnect().waitForCompletion(getCompletionTimeout());
+			if (this.mqttClient != null && this.mqttClient.isConnected()) {
+				this.mqttClient.unsubscribe(topics).waitForCompletion(getCompletionTimeout());
+				this.mqttClient.disconnect().waitForCompletion(getCompletionTimeout());
+			}
 		}
 		catch (MqttException ex) {
 			logger.error(ex, () -> "Error unsubscribing from " + Arrays.toString(topics));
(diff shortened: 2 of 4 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -187,33 +190,40 @@ protected void doStart() {
...
 		this.topicLock.lock();
-		String[] topics = getTopic();
+		initializeMqttAsyncClientIfRequired();
+		String[] topics = getTopicNames();
 		try {
 			this.mqttClient.unsubscribe(topics).waitForCompletion(getCompletionTimeout());
 			this.mqttClient.disconnect().waitForCompletion(getCompletionTimeout());
-		}
-		catch (MqttException ex) {
+		} catch (MqttException ex) {
```
artembilan, line 198: That's not correct code style.

Please, run `./gradlew clean :spring-integration-mqtt:check` to see all the Checkstyle violations.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -187,33 +190,40 @@ protected void doStart() {
 	@Override
 	protected void doStop() {
 		this.topicLock.lock();
-		String[] topics = getTopic();
+		initializeMqttAsyncClientIfRequired();
```
artembilan, line 193: That's something wrong in the `stop()` logic.
if there is not client initialized why would one wants to initialize it form stopping?

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -187,33 +190,40 @@ protected void doStart() {
...
+			}
+		} catch (MqttException ex) {
 			logger.error(ex, "Failed to close 'MqttAsyncClient'");
 		}
 	}
 
 	@Override
 	public void addTopic(String topic, int qos) {
 		this.topicLock.lock();
+		initializeMqttAsyncClientIfRequired();
```
artembilan, line 226: That's not OK, too. 
The `addTopic()` could be called from the Spring context initialization phase. Therefore it is too aggressive to make a low-level resource connection.

I believe the logic here must be similar to the one in the `MqttPahoMessageDrivenChannelAdapter`:
```
if (this.client != null && this.client.isConnected()) {
		this.client.subscribe(topic, qos);
}
```

artembilan (changes requested):
Please, find some review on the lines.
Thank you so far!

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -140,7 +141,7 @@ protected MqttMessageConverter getConverter() {
 	}
 
 	@ManagedAttribute
-	public String[] getTopic() {
+	public String[] getTopicNames() {
```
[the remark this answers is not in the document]
Mikhail2048: I agree, you are right. I will leave the name of the method unchanged, thanks

mipo256 (commented):
@artembilan I have adjusted the PR after your comments, can you, please, take a look again?

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -215,8 +231,12 @@ public void destroy() {
 	public void addTopic(String topic, int qos) {
 		this.topicLock.lock();
 		try {
-			this.mqttClient.subscribe(topic, qos).waitForCompletion(getCompletionTimeout());
-			super.addTopic(topic, qos);
+			if (mqttClient != null && mqttClient.isConnected()) {
+				this.mqttClient.subscribe(topic, qos).waitForCompletion(getCompletionTimeout());
+				super.addTopic(topic, qos);
+			} else {
+				logger.warn(String.format("An attempt to add topic : '%s' with QoS : '%s' without having mqtt client initialized or connected, the topic will not be added!", topic, qos));
```
artembilan, line 238: I don't think it is so bad to store topics (`super.addTopic(topic, qos);`) for the future use.
Something similar what we have so far in the `MqttPahoMessageDrivenChannelAdapter`.
So, please, change the logic over here to the old one and subscribe only if there is an `mqttClient`.

artembilan (changes requested):
I also would like to see some simple unit test to verify the we don't have a reported in the issue NPE any more.

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
Mikhail2048: Well, ok, perhaps) I just retained a warning in case `mqttClient` is not initialized

Mikhail2048:
@artembilan I have supplied the RP with 2 simple unit tests and squashed commits into one, can you please review?

Thanks

artembilan:
> squashed commits into one

That was not good, since now I have to review all your changes.
While without squashing there would be only need to take a look into fresh commits.

Not problem for now: jut want to give you a hint how it would be easier to collaborate in the future. 😄 
We squash it in the end when we merge it upstream.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -215,8 +223,12 @@ public void destroy() {
 	public void addTopic(String topic, int qos) {
 		this.topicLock.lock();
 		try {
-			this.mqttClient.subscribe(topic, qos).waitForCompletion(getCompletionTimeout());
 			super.addTopic(topic, qos);
+			if (mqttClient != null && mqttClient.isConnected()) {
+				this.mqttClient.subscribe(topic, qos).waitForCompletion(getCompletionTimeout());
+			} else {
+				logger.warn(String.format("An attempt to add topic : '%s' with QoS : '%s' without having mqtt client initialized or connected", topic, qos));
```
artembilan, line 230: This one has to be removed, too.
It is fully OK to add topics which could be subscribed later on, when `start()` is called.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -230,7 +242,11 @@ public void addTopic(String topic, int qos) {
 	public void removeTopic(String... topic) {
 		this.topicLock.lock();
 		try {
-			this.mqttClient.unsubscribe(topic).waitForCompletion(getCompletionTimeout());
+			if (mqttClient != null && mqttClient.isConnected()) {
+				this.mqttClient.unsubscribe(topic).waitForCompletion(getCompletionTimeout());
+			} else {
+				logger.warn(String.format("An attempt to remove topics : '%s' without having mqtt client initialized or connected", Arrays.toString(topic)));
```
artembilan, line 248: DITTO

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapterTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Mqttv5PahoMessageDrivenChannelAdapterTest's unit tests
+ *
+ * @author Mikhail Polivakha
+ */
+class Mqttv5PahoMessageDrivenChannelAdapterTest {
+
+	@Test //GH-3732
+	public void testNoNpeIsNotThrownInCaseDoInitIsNotInvokedBeforeTopicAddition() {
```
artembilan, line 14: How about just add these new tests into an existing `Mqttv5BackToBackTests`?

artembilan (changes requested):

mipo256 (commented):
I have fixed all of your suggestions and verified the code with Checkstyle, sorry for being annoying :)

artembilan (approved):
Pulling locally for the final review and possible merge...

artembilan:
Merged and cherry-picked to `5.5.x` as https://github.com/spring-projects/spring-integration/commit/c1d29ab3855444189045fa2400da3d406b92654a after some minor code clean up.

@Mikhail2048 ,

PR #3749, opened 2022-03-15
Cancel subscription for MPS.subscribeToPublisher

artembilan:
The `Flux.takeWhile()` only works if there is data in the `Publisher`
to consume.
We still need to be able to cancel subscription and stop producing even if
there is no data at the moment.

* Change `takeWhile()` to the `doOnSubscribe()` and store `subscription`
in the `volatile` property of the `MessageProducerSupport`
* Cancel such a subscription in the `doStop()` impl
* Propagate `doStop()` to super in the `ZeroMqMessageProducer`
which is only one reactive channel adapter overriding `doStop()`
* Verify in the `ReactiveMessageProducerTests` that subscription is cancelled
for delayed data in the `Publisher`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -196,6 +199,11 @@ protected void doStart() {
 	 */
 	@Override
 	protected void doStop() {
+		Subscription subscriptionToCancel = this.subscription;
```
garyrussell, line 202: `synchronized` ? Or is cancel idempotent?

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
...and cherry-picked.

PR #3753, opened 2022-03-18
GH-3734: Support MessageHistory on Kafka deser

### LINKED ISSUE #3734: Default addition of "org.springframework.integration.history" to trusted packages of DefaultKafkaHeaderMapper
**Expected Behavior**

By default, "org.springframework.integration.history" should be available to the trusted package of DefaultKafkaHeaderMapper.

**Current Behavior**

Message is not getting deserialized due to  untrusted package(i.e "org.springframework.integration.history ")

Also raised a stack-overflow question i.e [](https://stackoverflow.com/questions/71305189/message-history-not-getting-deserialized)

**Context**

Following are the alternatives that i considered
 a) Defining a bean instance for DefaultKafkaHeaderMapper with the addTrustedPackages("*")
 b) Defining a bean for "MessagingMessageConverter" 
 c)  Setting "MessagingMessageConverter" with header mapper mentioned in a) for "ConcurrentKafkaListenerContainerFactory"

No

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3734

The `MessageHistory` is not in the trusted packages of the `DefaultKafkaHeaderMapper`
therefore it fails when `MessageHistory.read()` is performed.

* Configure default `DefaultKafkaHeaderMapper` in the `SubscribableKafkaChannel`,
`KafkaInboundGateway`, `KafkaMessageDrivenChannelAdapter` and `KafkaMessageSource`
to also trust packages exposed via `JacksonJsonUtils.DEFAULT_TRUSTED_PACKAGES`
which include a `MessageHistory`, too.

* Verify in some integration Kafka tests to be sure that `MessageHistory` is
deserialized properly in the transferred headers
* Rework some tests to use `@EmbeddedKafka` instead of `@BeforeAll/@AfterAll`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
@@ -36,15 +36,19 @@ import org.springframework.integration.kafka.support.RawRecordHeaderErrorMessage
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.integration.support.ErrorMessageUtils;
 import org.springframework.integration.support.MessageBuilder;
+import org.springframework.integration.support.json.JacksonJsonUtils;
 import org.springframework.kafka.core.KafkaTemplate;
 import org.springframework.kafka.listener.AbstractMessageListenerContainer;
 import org.springframework.kafka.listener.ConsumerSeekAware;
 import org.springframework.kafka.listener.ContainerProperties;
 import org.springframework.kafka.listener.adapter.RecordMessagingMessageListenerAdapter;
 import org.springframework.kafka.support.Acknowledgment;
+import org.springframework.kafka.support.DefaultKafkaHeaderMapper;
+import org.springframework.kafka.support.JacksonPresent;
 import org.springframework.kafka.support.KafkaHeaders;
 import org.springframework.kafka.support.converter.ConversionException;
 import org.springframework.kafka.support.converter.KafkaMessageHeaders;
+import org.springframework.kafka.support.converter.MessagingMessageConverter;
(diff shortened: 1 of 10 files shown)


spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -192,9 +204,6 @@ protected void onInit() {
 			}
 		}
 		ContainerProperties containerProperties = this.messageListenerContainer.getContainerProperties();
-		Object existing = containerProperties.getMessageListener();
-		Assert.state(existing == null, () -> "listener container cannot have an existing message listener (" + existing
-					+ ")");
```
garyrussell, line 197: Why remove these assertions? You asked me to add them last week.

garyrussell (commented):
Just one comment/question

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
artembilan: Sorry for confusion: we have had them before in the ctor anyway.
I guess I shouldn't do this since it might cause a conflict on cherry-pick...

garyrussell:
...cherry-picked as eec3e956fe after resolving conflicts.

pmahony893:
Changing `Arrays.asList` to `List.of` breaks this for Java 8, which is listed as a compatible version in the [docs](https://docs.spring.io/spring-integration/docs/current/reference/html/preface.html#supported-java-versions).

artembilan:
It has been fixed in the #3761. Will be released in a couple weeks - April 19th.

PR #3754, opened 2022-03-18
GH-3735: Don't mutate FeedEntryMS metadataKey

### LINKED ISSUE #3735: Setting metadatakey on FeedEntryMessageSource using URL object is appending url to metadatakey but when Resource object is used, only metadatakey is set
**Using spring integration core 5.5.7 and spring integration feed 5.5.7**

**Expected Behavior**

If user is already passing metadatakey then he is already aware about metadatakey, so both constructors of FeedEntryMessageSource should only set metdatakey, not modify it.

**Current Behavior**

Currently, if object of FeedEntryMessageSource created with URL, then url is appended with metadatakey and if created with second constructor which takes Resource then it is only setting metadatakey.

**How has this issue affected you?**
I am creating instance of FeedEntryMessageSource with both constructors. so if instance with URL failed then Resource one can be used but i am getting different metadatakey in both cases.

What other alternatives have you considered? None

Are you aware of any workarounds? No

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3735

The `FeedEntryMessageSource` adds an url to the provided `metadataKey`
making it incompatible when we provide a `Resource`-based configuration.

* Remove adding of the url to the `metadataKey` making it rely only
on the provided value
* Remove internal `Comparator` for entries in favor of `Comparator.comparing()`
feature
* Improve some internal logic of the `PropertiesPersistingMetadataStore`
when it emits a false warning: cannot create dirs, but they are present
* Improve `feed.adoc`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/metadata/PropertiesPersistingMetadataStore.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/metadata/PropertiesPersistingMetadataStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


garyrussell (changes requested):
LGTM, but there are some strange unrelated GraphQL errors in the action results.

artembilan:
Correct. See this first: https://github.com/spring-projects/spring-integration/pull/3756

garyrussell (approved):

PR #3758, opened 2022-03-25
Enable functional setters usage by changing FluxAggregatorMessageHandler's …

markusherbert:
…setters return type to `this` instead of `void`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
@@ -141,9 +141,10 @@ public class FluxAggregatorMessageHandler extends AbstractMessageProducingHandle
 	 * {@link IntegrationMessageHeaderAccessor#CORRELATION_ID} header value.
 	 * @param correlationStrategy the {@link CorrelationStrategy} to use.
 	 */
-	public void setCorrelationStrategy(CorrelationStrategy correlationStrategy) {
+	public FluxAggregatorMessageHandler setCorrelationStrategy(CorrelationStrategy correlationStrategy) {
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
This change does not give too much gain when you edit the code in the IDE.
You just declare this as a `@Bean` and use its injection in other places.

On the other hand we have many-many other components which have plain `void` setters. 
Do you suggest to change them, too?
I mean how about consistency?
What is really a strong purpose for this change?

Thanks for understanding!

ipanimnim:
@artembilan You're totally right about the consistency. In general yes. It would be better to return `this` in every component setter. Why? exactly because of what you said. That force us to move the basic settings into another function. Instead of doing all of this on the fly:
```java
from().aggregate(new FluxAggregatorMessageHandler()
                  .setCorrelationStrategy(myCorrelationStrategy))
          .handle();
```

artembilan:
I would say such a style is not recommended.
It is much better to extract those components definition and configuration to separate bean or, at least, `private` methods.
That's your "demo"-ready sample is so easy to ready, but most real components are not so simple in their config, therefore even if we would have builder style instead of `void` setters the whole flow configuration would look very complicated.

Perhaps as a compromise between us, you'd better to look into a `FluxAggregatorSpec` impl instead?..
See similar `AggregatorSpec`. However this one better to make as an extension of the `ConsumerEndpointSpec`.
Probably the best sample would be  `DelayerEndpointSpec` or `EnricherSpec`...

We have that extra layer between plain setter-based component which could be used for XML configuration and that `Spec`, builder-like configurers for better Java DSL experience: https://docs.spring.io/spring-integration/docs/current/reference/html/dsl.html#java-dsl-basics

ipanimnim:
Understand. Totally agree with the idea.

PR #3759, opened 2022-03-25
Add IntegrationFlowAdatper extentions in usage examples

markusherbert:
I'm not sure if the `@Bean` declaration is needed above the `buildFlow()` function.

### DIFF
--- a/src/reference/asciidoc/reactive-streams.adoc
+++ b/src/reference/asciidoc/reactive-streams.adoc
@@ -199,13 +199,15 @@ Usage would look like:
 ====
 [source, java]
 ----
-public class MainFlow {
+@Component
+public class MainFlow extends IntegrationFlowAdapter {
   @Autowired
   private CustomReactiveMessageProducer customReactiveMessageProducer;
   
   @Bean
-  public IntegrationFlow buildFlow() {
-     return IntegrationFlows.from(customReactiveMessageProducer)
+  @Override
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
What is the point of this change?
It just does not make sense to a logic we want to express with those samples.

The `IntegrationFlowAdapter` is explained in its specific place: https://docs.spring.io/spring-integration/docs/current/reference/html/dsl.html#java-dsl-flow-adapter.

The Reactive Streams chapter is already complex enough to bring another not related stuff.
Please, clarify why would we need this in the docs?
Perhaps you'd like to see it really in the samples project instead?..

ipanimnim:
This is not used to demonstrate how to use `IntegrationFlowAdapter`, but to demonstrate in general the direction of correct implementation way. I think it would be more helpful for all.

ipanimnim:
> Typically integration solutions are so task-specific
Why not focus on the ones who are building a full data pipeline with the project and want to understand the basics of full application usage? I'm not saying to do this everywhere, but in pages like this which are introduction for some specific usage, it can be helpful.

artembilan:
This page is about a general info what and how is supported in Spring Integration around Reactive Streams. 
There is no full data pipeline concentration in this chapter.
There is no some specific usage in this chapter.
That's why I still cannot understand what a gain does that `IntegrationFlowAdapter` bring exactly to this chapter...

PR #3760, opened 2022-03-25
Make the reactive gateway example readable

markusherbert:
Relevant in general for the JavaDSL gateway usage. The `...` is super hard for new users to understand. I think it should be replaced with a real-world example in every usage.

### DIFF
--- a/src/reference/asciidoc/gateway.adoc
+++ b/src/reference/asciidoc/gateway.adoc
@@ -756,27 +756,51 @@ The following example shows how to create a gateway with Project Reactor:
 ----
 @MessagingGateway
 public interface TestGateway {
-
 	@Gateway(requestChannel = "promiseChannel")
 	Mono<Integer> multiply(Integer value);
-
-	}
-
-	    ...
-
+}
+----
+It's implementation looks like:
+====
+[source,java]
+----
+@Component
+public class MyServiceActivators {
 	@ServiceActivator(inputChannel = "promiseChannel")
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Doesn't look like your change is relevant somehow to the task of this chapter.
This now has to explain how to use exactly `@MessagingGateway`.
Your change fully eliminate its usage in the target flow.

What is that `someReactiveMessageProducer` for this chapter?
The `e.reactive(flux -> flux.flatMap()`  is not correct to use for this reactive endpoint configurer.
This point is not about data processing.

I agree with you that ellipsis might not be so useful in the sample code, but your change is so drastic for the original purpose and much more confusing and misleading and raising more questions than before, that I cannot accept it yet.

Sorry and thank you for the attempt.

ipanimnim:
The `someReactiveMessageProducer` appears just for the demonstration. Can you suggest improvements to what I submitted? I think a more detailed example can be great in this case

artembilan:
As I said in my previous comment: your change appears to be not relevant to the context of this particular doc and fully break the purpose of that ` might be a part of Reactor stream` sentence.
We try explain in this paragraph some specific feature of the gateway when it returns a `Mono`. 
Most of your changes to this simple sample just makes things more complex and may mislead from the education goal.

I'm closing this in favor of the fix I've just pushed: https://github.com/spring-projects/spring-integration/commit/6a435fbd777177aa0020976b5dbee122aae299b8

Thanks for understanding.

PR #3768, opened 2022-04-07
Close mail folder if no messages to produce

artembilan:
Related to https://stackoverflow.com/questions/71667731/spring-integration-mimemessage-gmail-folder-is-not-open-exception

If no mail messages pulled from the folder or all of them are filtered out,
there is nothing to produce downstream.
Therefore, always close the folder in the end of `AbstractMailReceiver.receive()`
when no messages and even if `autoCloseFolder == false`

**Cherry-pick to `5.5.x`**


garyrussell (approved):

garyrussell:
...and cherry-picked as c75be107fa

PR #3769, opened 2022-04-07
GH-3615: Make log() as non-terminal in DSL

### LINKED ISSUE #3615: .log() should always behave the same, no matter where it is placed in a flow
**Expected Behavior**

Given some code like this:

```
        .scatterGather(
                scatterer -> scatterer
                        .applySequence(true)
                        .recipientFlow(IntegrationFlows.from(MessageChannels.executor(myExecutorService))
                                .gateway(mySubFlow1())
                                .get())
                        .recipientFlow(IntegrationFlows.from(MessageChannels.executor(myExecutorService))
                                .log(/* some logging here */)
                                .gateway(mySubFlow2())
                                .get()),
                gatherer -> gatherer
                        .releaseStrategy(defaultReleaseStrategy)
                        .outputProcessor(myOutputProcessor))
```

if I move the `.log()` statement down one line, like this:

```
        .scatterGather(
                scatterer -> scatterer
                        .applySequence(true)
                        .recipientFlow(IntegrationFlows.from(MessageChannels.executor(myExecutorService))
                                .gateway(mySubFlow1())
                                .get())
                        .recipientFlow(IntegrationFlows.from(MessageChannels.executor(myExecutorService))
                                .gateway(mySubFlow2())
                                .log(/* some logging here */)
                                .get()),
                gatherer -> gatherer
                        .releaseStrategy(defaultReleaseStrategy)
                        .outputProcessor(myOutputProcessor))
```

I would expect everything to work exactly the same, except that the `.log()` statement is executed after the `.gateway()` call.

**Current Behavior**

Changing the position of `.log()` changes the behavior of the flow. In this case, the releaseStrategy never receives the output of the second recipientFlow. The reason for this is explained in the JavaDoc, where we read - "When this operator is used in the end of flow, it is treated as one-way handler without any replies to continue.". 

**Context**

This is not the behavior one would expect, it's not intuitive, and it has caused an unexpected bug in our team that took a long time to debug. (https://en.m.wikipedia.org/wiki/Principle_of_least_astonishment) We were quite surprised to learn about `.logAndReply()` and consider it one of those things that make the SI DSL hard to work with. Sometimes things just don't work and you have not reason why and no easy and no quick way to find the cause.

IMO the `.log()` statement should behave the same no matter where you place it in a flow. If it is placed at the end of a flow, and if this flow has no output channel, throw the usual exception, and the developer will quickly realize they need to add `.nullChannel()` or something similar. Perhaps you could even detect if `.log()` needs to reply or not (it seems you can detect if it is at the end of a flow, so maybe at that moment you can also detect if an output channel is configured for this flow) ?

`logAndReply()` could be deprecated and ultimately removed.

Am I missing something that would prevent such a solution ?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3615

The `log()` operator for Java DSL in the end of flow causes a confusion
for its different behavior and, therefore, inconsistency with expectations

* Populate a `bridge()` in the end of flow instead of `nullChannel` when
the current component is `WireTapSpec` (relevant to `wireTap()` and `log()`)
* Deprecate a `logAndReply()` operator since its behavior is now included into the `log()` in the end if flow
* Fix tests to use just `log()` in the end from now on
* Fix `dsl.adoc` for the new `log()` behaviour in the end of flow

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
@@ -2182,9 +2182,6 @@ public abstract class BaseIntegrationFlowDefinition<B extends BaseIntegrationFlo
 	 * logging level and {@code org.springframework.integration.handler.LoggingHandler}
 	 * as a default logging category.
 	 * <p> The full request {@link Message} will be logged.
-	 * <p> When this operator is used in the end of flow, it is treated
-	 * as one-way handler without any replies to continue.
-	 * The {@link #logAndReply()} should be used for request-reply configuration.
 	 * @return the current {@link BaseIntegrationFlowDefinition}.
 	 * @see #wireTap(WireTapSpec)
 	 */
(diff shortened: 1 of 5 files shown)


garyrussell (approved):

artembilan:
Also see a Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.x-to-6.0-Migration-Guide#the-log-in-the-end-of-flow

garyrussell:
I polished the migration guide a little.

PR #3786, opened 2022-04-20
GH-3785: Close stream for persistent collection

### LINKED ISSUE #3785: JdbcMessageStore doesn't close database connections after streaming stored messages

5.5.10

**Describe the bug**

Usage of `JdbcMessageStore` eventually leads to the database connection pool depletion as the connections are not getting properly closed after streaming the messages. After a while, the application stops functioning correctly, as the database connection pool never receives those connections back and treats them as active.

This seems to happen on:
 - message group expiration via the configured timeout
 - explicit execution of a `MessageGroupStoreReaper` instance
 - streaming through the messages of a message group via `getMessages().stream()`

The issue appears to be with the stream from `PersistentMessageGroup#streamMessages` not getting closed after the usage in certain cases.

**To Reproduce**

Set up the application with a limited database connection pool, configure an aggregation flow with an instance of `JdbcMessageStore` and a scheduled `MessageGroupStoreReaper` execution for that message store.

**Expected behavior**

Internal usages of the message store should properly release database connections when `JdbcMessageStore` is utilized.

If the current behavior of streaming on `MessageGroup#getMessages` is expected, the need for closing the stream should ideally be mentioned in the documentation. 

**Sample**

[Sample application](https://github.com/Seregy/spring-integration-jdbc-connection-leak)

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3785

* Fix `CollectionArgumentResolver` and `PayloadsArgumentResolver` to
close the `Stream` of message after its usage
* Rework `AbstractKeyValueMessageStore.removeMessagesFromGroup()`
to iterate input collection of messages not its stream to avoid
the mentioned problem

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/CollectionArgumentResolver.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.springframework.core.MethodParameter;
 import org.springframework.core.convert.TypeDescriptor;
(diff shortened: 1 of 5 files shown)


garyrussell:
No tests needed?

artembilan:
Well, it doesn't look obvious what test has to do, but I'll think about that tomorrow.
For self-learning purpose.
I just don't want to add more deps for no extreme reason....

garyrussell:
Maybe a mock AMGS with a message stream passed into each of the resolvers and verify that the mock stream is closed?

Perhaps we also should enhance the javadocs of the JDBC MGS to say that the stream must be closed; the AGMS just streams the `getMessages()` so close is not required there, but any subclass that uses true streaming presumably needs the stream closed.

It looks like we don't even have any unit tests for `PMG.streamMessages()`.

garyrussell (approved):

garyrussell:
... and cherry-picked.

PR #3792, opened 2022-05-05
GH-3679: Better caching for SpringIntegrationTest

### LINKED ISSUE #3679: Reconsider @SpringIntegrationTest
Perhaps create a JUnit5 condition to stop/start the specified endpoints instead of adding an `IntegrationEndpointsInitializer` to the context.

The current mechanism defeats the test context caching mechanism because each test class with `@SpringIntegrationTest` gets a different context customizer.

See https://github.com/spring-projects/spring-integration/discussions/3674#discussioncomment-1654268 for more information.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3679

The `@SpringIntegrationTest` makes a test cache key based on its attributes values
when the same application context can be used in different test classes
with different endpoints to have stopped originally.

* Rework an `IntegrationEndpointsInitializer` to the `SpringIntegrationTestExecutionListener`
which consult a `MockIntegrationContext` for endpoints to be started or not before
the test execution and definitely stopped after the test execution to have a flexibility
with the cached context
* Improve a `MockIntegrationContext` to gather `AbstractEndpoint` beans and have them
marked for stopping in the beginning of the application context.
The `SpringIntegrationTestExecutionListener` takes care about startup in its `beforeTestClass()`
* Verify different state for the `SpringIntegrationTest` with the `CachedSpringIntegrationTestAnnotationTests`


artembilan:
BTW, with the `TestExecutionListener` approach the fix works even with JUnit 4. No reason to be hard-dependent on JUnit 5 yet 😄 !

garyrussell (approved):

PR #3796, opened 2022-05-10
GH-3794: Use less memory with scheduled tasks

### LINKED ISSUE #3794: Memory leak in AbstractCorrelatingMessageHandler
SI core 5.x

**Describe the bug**
Hard references to completed and released `MessageGroup` instances stored memory by `AbstractCorrelatingMessageHandler`

**Additional details**
Please see the following discussion: https://github.com/spring-projects/spring-integration/commit/2f0b377cfb564876544c9c1d54f5a4bd5cb06f5b#r73229543, also note this is effectively a sibling of https://jira.spring.io/browse/INT-3999, the fix for which was never applied to the empty group removal logic.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3794

* Replace hard reference to message group with its `id` in the
`AbstractCorrelatingMessageHandler.removeEmptyGroupAfterTimeout()`
* Replace hard reference to message with its `id` in the
`DelayHandler.rescheduleAt()`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


artembilan:
I'm not sure how to test this therefore I just leave it as is with the conclusion that nothing is broken and that's it.

garyrussell:
Not really a memory leak, right? Actually "Reduce Memory Usage with Scheduled Tasks"?

artembilan:
Well, @vfunshteyn-zw called it a "memory leak" because their groups were scheduled for 72 hours. 
More content in the group - more memory consumption.

Anyway: we can call it whatever, but better to let GC to do its job and leave the memory for other purposes.

garyrussell:
Right; I was just concerned that someone browses the change log.

A leak is specifically when some memory is allocated but never released.

In this case, it's unnecessary memory usage that is eventually removed.

I am ok with merging as-is, though, if you like.

artembilan:
Fixed

garyrussell:
...and cherry-picked.

PR #3798, opened 2022-05-11
Improve README.md file so it will have examples, explanations, and logo

migroskub:
The library's README looks pretty bad right now. The library is super maintained and well used, but from its Github page, it doesn't look like this. This PR aims at helping modernize the homepage of Spring Integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<img align="right" width="350" height="350" src="https://user-images.githubusercontent.com/96301480/153731014-72229c13-cea9-4da1-85cd-bcf445e358bf.svg">
+<img align="right" width="290" height="290" src="https://user-images.githubusercontent.com/96301480/153731014-72229c13-cea9-4da1-85cd-bcf445e358bf.svg">
 
 
 Spring Integration [<img src="https://build.spring.io/plugins/servlet/wittified/build-status/INT-MAIN">](https://build.spring.io/browse/INT-MAIN) [![Join the chat at https://gitter.im/spring-projects/spring-integration](https://badges.gitter.im/spring-projects/spring-integration.svg)](https://gitter.im/spring-projects/spring-integration?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,84 @@
...
+
+Using the Spring Framework encourages developers to code using interfaces and use dependency injection (DI) to provide a Plain Old Java Object (POJO) with the dependencies it needs to perform its tasks. Spring Integration takes this concept one step further, where POJOs are wired together using a messaging paradigm and individual components may not be aware of other components in the application. Such an application is built by assembling fine-grained reusable components to form a higher level of functionality. WIth careful design, these flows can be modularized and also reused at an even higher level.
+
+In addition to wiring together fine-grained components, Spring Integration provides a wide selection of channel adapters and gateways to communicate with external systems. Channel Adapters are used for one-way integration (send or receive); gateways are used for request/reply scenarios (inbound or outbound). For a full list of adapters and gateways, refer to the reference documentation.
+
+# Installation and Getting Started
```
artembilan, line 13: Not sure in this `Installation` word.
Plus you don't show any steps to do that.

README.md
```diff
@@ -1,6 +1,84 @@
...
+
+Here is a quick teaser of a complete Spring Integration application in Java:
```
artembilan, line 15: Doesn't look like you sample is really complete...

artembilan (changes requested):
I won't mind to have some improvements in this file, but perhaps better to keep it short and add more links to existing resources.

Something similar to what we have so far in Spring Framework would be preferred: https://github.com/spring-projects/spring-framework

Thanks

artembilan:
OK. I see you took it from Spring Boot README: https://github.com/spring-projects/spring-boot.

Anyway: I'd prefer to not have Spring Boot mentioning over here. It may lead to a false impression that Spring Integration requires Spring Boot.

migroskub:
@artembilan I think you did great and improved the readme a lot. I’d improve a little the code sample you’ve attached so it’ll contain some more concrete example (so the user will get a sneak peek at the `IntegrationFlow`). And I’d also mention the asynchronous/ reactive support because people may consider their usage based on this.

PR #3802, opened 2022-05-17
Addition of Java DSL Package to SMB Module

GregBragg:
@artembilan As promised here is the addition of the Java DSL package for SMB Support. This new branch contains the cherry picked commits from the previous messed up branch.


spring-integration-smb/src/main/java/org/springframework/integration/smb/inbound/SmbStreamingMessageSource.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	@Override
+	public String getComponentType() {
+		return "smb:inbound-streaming-channel-adapter";
+	}
+
+	@Override
+	protected List<AbstractFileInfo<SmbFile>> asFileInfoList(Collection<SmbFile> files) {
+		List<AbstractFileInfo<SmbFile>> canonicalFiles = new ArrayList<AbstractFileInfo<SmbFile>>();
```
artembilan, line 76: We simply can use diamond operators starting with Java 8 support in the project.

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbFileInfo.java
```diff
@@ -0,0 +1,119 @@
...
+	@Override
+	public String getPermissions() {
+		ACE[] aceArray = null;
+		try {
+			aceArray = this.smbFile.getSecurity(true);
+		}
+		catch (IOException se) {
+			logger.error("Unable to determine security descriptor information for this SmbFile", se);
+		}
+		return (aceArray == null) ? null : aceArray.toString();
```
artembilan, line 111: The `Arrays.toString(aceArray)` is recommended way to do this.

spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * The actual SMB share must be configured in file 'SmbTests-context.xml'
+ * with the 'real' server settings for testing.
+ *
+ * @author Gregory Bragg
+ *
+ * @since 6.0
+ */
+
+@Disabled("Actual SMB share must be configured in file [SmbTests-context.xml].")
```
artembilan, line 43: i wonder if we can look into using a Testcontainers with Samba Docker image for this kind of tests: https://hub.docker.com/r/dperson/samba ...

Not the current PR though

spring-integration-smb/src/test/resources/META-INF/spring/integration/SmbTests-context.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
```
artembilan, line 2: Can we somehow go without an XML configs from now on if there is no strict requirements to test exactly our XML components?

I see that you have only an `SmbSessionFactory` over here, but that doesn't really justify a whole `ApllicationContext` just for this.
The `SmbSessionFactory` can be created manually and we can control its lifecycle in the test ourselves.

src/reference/asciidoc/smb.adoc
```diff
@@ -122,6 +122,55 @@ public MessageSource<File> smbMessageSource() {
 
 For XML configuration the `<int-smb:inbound-channel-adapter>` component is provided.
 
+==== Configuring with the Java DSL
```
artembilan, line 125: Can you also add into docs new components you have introduced in this PR: `SmbStreamingMessageSource` and `SmbOutboundGateway`.
Well, perhaps in a new PR as well.

artembilan (changes requested):
It also would be great to add a short sentence into a `whats-new.adoc` that our SMB module has now a Java DSL. And we also have now `SmbStreamingMessageSource` and `SmbOutboundGateway`.

This is great work! 
I really don't see any significant problems with your code.

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbFileInfo.java
GregBragg: I struggled with this method of returning security permissions for a user... it's hard to relate what it looks like on a Windows OS to  what it looks like on a Unix OS. Not sure what value it brings to anybody to see a list of users and what access they have to for a file or directory.

spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
GregBragg: I definitely looked into some sort of embedded SMB Server for a few days, like what we have for (S)FTP however there is none available. I agree we could have Samba container for the build process to use instead, it doesn't help however for localhost testing against an external SMB share, unless we expose it to the outside world to use for development purposes.

spring-integration-smb/src/test/resources/META-INF/spring/integration/SmbTests-context.xml
GregBragg: I originally just used constants in the class that needed to be edited with real values... I figured this was a cleaner approach.

spring-integration-smb/src/main/java/org/springframework/integration/smb/gateway/SmbOutboundGateway.java
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * @since 6.0
+ */
+public class SmbOutboundGateway extends AbstractRemoteFileOutboundGateway<SmbFile> {
+
+	private static final Log logger = LogFactory.getLog(SmbOutboundGateway.class);
+
+	private Expression workingDirExpression;
+
+	private StandardEvaluationContext evaluationContext;
```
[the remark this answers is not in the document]
GregBragg: Yeah I know it's weird... my IDE says the same thing, but wehn I remove them I get a compile error. They are both used in
`setWorkingDirExpression(Expression workingDirExpression)` and `doInit()`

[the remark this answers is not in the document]
artembilan: Correct. That's settings values to them. But they are out of use afterwards in the logic.
See for example `FtpOutboundGateway.doInWorkingDirectory()`

spring-integration-smb/src/main/java/org/springframework/integration/smb/gateway/SmbOutboundGateway.java
```diff
@@ -0,0 +1,216 @@
...
+import jcifs.smb.SmbFile;
+
+/**
+ * Outbound Gateway for performing remote file operations via SMB.
+ *
+ * @author Gregory Bragg
+ *
+ * @since 6.0
+ */
+public class SmbOutboundGateway extends AbstractRemoteFileOutboundGateway<SmbFile> {
```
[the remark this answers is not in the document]
artembilan: No. Just `smb/outbound`. No `gateway` at all.
It is just component implementation details that it is a gateway. 
Technically it is still an Outbound Channel Adapter. 
The fact that it returns something for a reply is just an extension of a general `outbound` pattern.

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbFileInfo.java
artembilan: I guess this should be left for a future issue: not related to this PR.
Feel free to raise one so we don't forget to investigate!

spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
artembilan: No, my point is to to use Testcontainers and its ability to run any Docker image.

That's must not be a part of a build process: just a Unit test scope.
Not sure why you talk about outside world...

spring-integration-smb/src/test/resources/META-INF/spring/integration/SmbTests-context.xml
artembilan: When we switch to Testcontainers, it is going to be OK to use those "real values" in the constant. 
See my approach for MQTT - `MosquittoContainerTest`. That's exactly what I'd like to see for SMB tests against Samba in Docker container.
I just have no idea how that hard going to be to use that Docker image and initialize it properly for our use-cases...

spring-integration-smb/src/main/java/org/springframework/integration/smb/gateway/SmbOutboundGateway.java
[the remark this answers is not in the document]
GregBragg: Yeah ok, see it... I think those methods are redundant then, because I stripped out the logic for FTP style logic such as ls, nlst, get, etc. I don't think it's really applicable to SMB protocol.

src/reference/asciidoc/smb.adoc
GregBragg: Sure, will create a separate branch and PR for smb.adoc

artembilan (approved):
I think we are good so far.
I'll add a couple sentences to docs on merge.

Unless you have something else to add/fix, @GregBragg .

I'm in a release process, so would be great to have this ASAP 😄

GregBragg:
> I think we are good so far. I'll add a couple sentences to docs on merge.
> 
> Unless you have something else to add/fix, @GregBragg .
> 
> I'm in a release process, so would be great to have this ASAP 😄

I'm good for now... I'm always here for any other stuff you need done on this module, Thanks!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/ac6af716e1cce254db15dbf5f92ec0391ff119e7 after some minor clean up.

@GregBragg ,

thank you very much again; looking forward for more!

PR #3806, opened 2022-05-19
Really reuse Testcontainers

artembilan:
The `withReuse(true)` and `testcontainers.reuse.enable=true` don't work together with `@Container`.
The JUnit extension gathers those containers and stop them in the end of test class unconditionally.

* Remove `@Container` annotation usage
* Use `@BeforeAll` and `GenericContainer.start()` manually
This way the container ensures to reuse existing running container and don't start a fresh one.
Since the container instance is stored in a `static` property, it is really started only once.
The rest tests in a suite just reuse that existing container.
Ryuk container will take care about their stopping and removal eventually afte JVM exit.

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlContainerTest.java
+++ b/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlContainerTest.java
@@ -16,12 +16,16 @@
 
 package org.springframework.integration.jdbc.mysql;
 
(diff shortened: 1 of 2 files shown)


artembilan:
Just sharing my findings.
Can merge and back-port myself.

Thanks

artembilan:
... and cherry-picked to `5.5.x` as https://github.com/spring-projects/spring-integration/commit/3e2e2308c13f9388e136b3b8664830b23a9e8434 after minor conflict resolution.

PR #3808, opened 2022-05-19
Enhanced SmbOutboundGateway with additional remote file operations

GregBragg:
@artembilan I've enhanced the SmbOutboundGateway with additional remote file operations.

The documentation will be updated in a later branch for your review, once I have it ready. Thanks!

### DIFF
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
@@ -16,15 +16,11 @@
 
 package org.springframework.integration.smb.outbound;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.net.MalformedURLException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import java.util.concurrent.Callable;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -33,11 +29,9 @@ import org.springframework.integration.file.remote.AbstractFileInfo;
 import org.springframework.integration.file.remote.MessageSessionCallback;
 import org.springframework.integration.file.remote.RemoteFileTemplate;
 import org.springframework.integration.file.remote.gateway.AbstractRemoteFileOutboundGateway;
-import org.springframework.integration.file.remote.session.Session;
 import org.springframework.integration.file.remote.session.SessionFactory;
 import org.springframework.integration.smb.session.SmbFileInfo;
 import org.springframework.integration.smb.session.SmbRemoteFileTemplate;
-import org.springframework.messaging.Message;
 
 import jcifs.smb.SmbException;
 import jcifs.smb.SmbFile;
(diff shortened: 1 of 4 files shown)


spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
```diff
@@ -134,4 +187,88 @@ protected SmbFile enhanceNameWithSubDirectory(SmbFile file, String directory) {
...
+				() -> super.mv(message, session, remoteFilePath, remoteFileNewPath));
+	}
+
+	@Override
+	protected String put(Message<?> message, Session<SmbFile> session, String subDirectory) {
+		try {
+			return doTask(message, session,
+					() -> super.put(message, session, subDirectory));
+		}
+		catch (IOException e) {
```
artembilan, line 235: Why is this? Doesn't look like `super.put()` throws any exceptions...

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
```diff
@@ -134,4 +187,88 @@ protected SmbFile enhanceNameWithSubDirectory(SmbFile file, String directory) {
...
+		}
+	}
+
+	@Override
+	protected List<String> mPut(Message<?> message, Session<SmbFile> session, File localDir) {
+		try {
+			return doTask(message, session,
+					() -> super.mPut(message, session, localDir));
+		}
+		catch (IOException e) {
```
artembilan, line 246: Same here.

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
```diff
@@ -134,4 +187,88 @@ protected SmbFile enhanceNameWithSubDirectory(SmbFile file, String directory) {
...
+			throw new UncheckedIOException(e);
+		}
+	}
+
+	private <V> V doTask(Message<?> message, Session<SmbFile> session, Callable<V> task)
+			throws IOException {
+
+		try {
+			return task.call();
```
artembilan, line 255: Doesn't look like there is an extra work in this `doTask()`: you just do plain delegation to the provided task and therefore you call that `super.` and that's.
From here it looks like all these new method over here are without value and we simply can rely on the impls in the super class.

What do I miss, please?

artembilan (changes requested):

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
GregBragg: It follows the same logic in FtpOutboundGateway

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
GregBragg: Same as above... It follows the same logic in FtpOutboundGateway

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
GregBragg: This code is based/follows the same logic in FtpOutboundGateway, and is a refactored version of 'doInWorkingDirectory' private method.

There is no concept of a working directory in the SMB client code/dependent library like what we have in 'FTPClient'. This method simply wraps the 'task.call()' in a safe try/catch.

All this logic was in the 'FtpOutboundGateway' class which I originally stripped out during the port to SMB. I have since found out that they are useful methods and will work with the SMB client code and dependent libraries.

GregBragg: I'll refactor it if it doesn't make sense for this new gateway... and remove 'doTask'. it is somewhat redundant. looks like the code needs cleanup anyways...

artembilan:
The `FtpOutboundGateway` was not good sample to follow. It has that `doInWorkingDirectory()` extra logic which throws an `IOException` from its `ftpClient.changeWorkingDirectory()`.
Since we don't have anything similar for SMB, I would say the `SftpOutboundGateway` is better sample to follow.

I guess you did that already in your latest commit.
Will look shortly.

GregBragg:
Yep, not a good example... I'm in the process of adding more unit tests and adding code to SmbSession to support `listNames` method which is missing from this API for use by `Command.NLST` .

Hold off for now until I can get these enhancements and unit tests in. Thanks!

GregBragg:
@artembilan Ok all done, I believe this code change is good now... Thanks!

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
```diff
@@ -480,9 +482,46 @@ public String getHostPort() {
...
+
+		if (logger.isDebugEnabled()) {
+			logger.debug("Successfully listed " + fileNames.size() + " resource(s) in [" + _path + "]"
+					+ ": " + fileNames.toString());
+		}
+		else if (logger.isInfoEnabled()) {
+			logger.info("Successfully listed " + fileNames.size() + " resource(s) in [" + _path + "]" + ".");
+		}
+
+		return fileNames.toArray(new String[fileNames.size()]);
```
artembilan, line 524: This is how my IDEA doesn't such a style:
image

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
```diff
@@ -480,9 +482,46 @@ public String getHostPort() {
...
+				if (logger.isWarnEnabled()) {
+					logger.warn("Remote directory [" + _path + "] does not exist. Cannot list resources.");
+				}
+				return fileNames.toArray(new String[0]);
+			}
+			else if (!smbDir.isDirectory()) {
+				throw new IOException("Resource [" + _path + "] is not a directory. Cannot list resources.");
+			}
+
+			SmbFile[] files = smbDir.listFiles();
```
artembilan, line 507: Hm. Looks like there is just `String[] list ()`  API in that `SmbFile`.
Why don't we use it for optimization?

artembilan (changes requested):
Nothing critical so far.
Any chances to see docs for this in the end, please?

Thanks

GregBragg:
> Nothing critical so far. Any chances to see docs for this in the end, please?
> 
> Thanks

It'll be next week, I promise. Some critical things going on at work these days...

GregBragg:
> > Nothing critical so far. Any chances to see docs for this in the end, please?
> > Thanks
> 
> It'll be next week, I promise. Some critical things going on at work these days...

I was able to get some time in this afternoon for PR https://github.com/spring-projects/spring-integration/pull/3810 
Thanks!

artembilan:
Thank you again!

FYI how to write a good a proper commit message, at least the fist one: https://cbea.ms/git-commit/

The GH has also a great feature to map that first commit message to the PR title and description.
Kinda saves our time a bit 😄

PR #3810, opened 2022-05-20
Updated SMB Module Documentation For New Components

GregBragg:
Addition of documentation for new components `SmbStreamingMessageSource` and `SmbOutboundGateway`.
As promised this is my first crack at adding the documentation in for these new components. We can review next week. Thanks!

### DIFF
--- a/src/reference/asciidoc/endpoint-summary.adoc
+++ b/src/reference/asciidoc/endpoint-summary.adoc
@@ -175,7 +175,7 @@ The following table summarizes the various endpoints with quick links to the app
 | <<./sftp.adoc#sftp-outbound-gateway,SFTP Outbound Gateway>>
 
 | *SMB*
-| <<./smb.adoc#smb-inbound,SMB Inbound Channel Adapter>>
+| <<./smb.adoc#smb-inbound,SMB Inbound Channel Adapter>>, <<./smb.adoc#smb-streaming-inbound,SMB Streaming Inbound Channel Adapter>>
 | <<./smb.adoc#smb-outbound,SMB Outbound Channel Adapter>>
 | N
 | <<./smb.adoc#smb-outbound-gateway,SMB Outbound Gateway>>
--- a/src/reference/asciidoc/endpoint-summary.adoc
+++ b/src/reference/asciidoc/endpoint-summary.adoc
@@ -178,7 +178,7 @@ The following table summarizes the various endpoints with quick links to the app
 | <<./smb.adoc#smb-inbound,SMB Inbound Channel Adapter>>
 | <<./smb.adoc#smb-outbound,SMB Outbound Channel Adapter>>
 | N
-| N
+| <<./smb.adoc#smb-outbound-gateway,SMB Outbound Gateway>>
 
 | *STOMP*
 | <<./stomp.adoc#stomp-inbound-adapter,STOMP Inbound Channel Adapter>>
--- a/src/reference/asciidoc/smb.adoc
+++ b/src/reference/asciidoc/smb.adoc
@@ -186,7 +186,7 @@ public class SmbJavaApplication {
 ----
 ====
 
-[[smb-streaming]]
+[[smb-streaming-inbound]]
 === SMB Streaming Inbound Channel Adapter
 
 This adapter produces message with payloads of type `InputStream`, letting files be fetched without writing to the local file system.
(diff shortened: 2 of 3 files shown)


src/reference/asciidoc/endpoint-summary.adoc
```diff
@@ -178,7 +178,7 @@ The following table summarizes the various endpoints with quick links to the app
 | <<./smb.adoc#smb-inbound,SMB Inbound Channel Adapter>>
```
artembilan, line 178: I think it won't be bad to mention `SMB Streaming Inbound Channel Adapter` over here as well.
See for example Redis line in this table: it does describe all the channel adapter implementations from that module.

If you do the same fix for (S)FTP modules, it would be great 😄

src/reference/asciidoc/smb.adoc
```diff
@@ -171,6 +186,141 @@ public class SmbJavaApplication {
...
+For more information about the `SmbPersistentAcceptOnceFileListFilter`, and how it is used, see <<./file.adoc#remote-persistent-flf,Remote Persistent File List Filters>>.
+
+Use the `max-fetch-size` attribute to limit the number of files fetched on each poll when a fetch is necessary.
+Set it to `1` and use a persistent filter when running in a clustered environment.
+See <<smb-max-fetch>> for more information.
+
+The adapter puts the remote directory and file name in the `FileHeaders.REMOTE_DIRECTORY` and `FileHeaders.REMOTE_FILE` headers, respectively.
+The `FileHeaders.REMOTE_FILE_INFO` header provides additional remote file information (represented in JSON by default).
+If you set the `fileInfoJson` property on the `SmbStreamingMessageSource` to `false`, the header contains an `SmbFileInfo` object.
+The `SmbFile` object provided by the underlying Apache Net library can be accessed by using the `SmbFileInfo.getFileInfo()` method.
```
artembilan, line 235: I think this is a copy/paste artifact. 
I don't think that can treat our `jcifs` library as a `Apache Net`...

src/reference/asciidoc/smb.adoc
```diff
@@ -171,6 +186,141 @@ public class SmbJavaApplication {
...
+The `fileInfoJson` property is not available when you use XML configuration, but you can set it by injecting the `SmbStreamingMessageSource` into one of your configuration classes.
```
artembilan, line 236: You know, I don't see that we have an XML support for the `<int-smb:inbound-streaming-channel-adapter>` at all...
Are you planning to add it or just back off with such a documentation then just in favor of Java DSL configuration?

src/reference/asciidoc/smb.adoc
```diff
@@ -246,3 +396,420 @@ public class SmbJavaApplication {
...
+====
+
+Framework components, such as the <<./file.adoc#file-splitter,File Splitter>> and <<./transformer.adoc#stream-transformer,Stream Transformer>>, automatically close the session after the data is transferred.
+
+The following example shows how to consume a file as a stream:
+
+====
+[source, xml]
+----
+<int-smb:outbound-gateway session-factory="smbSessionFactory"
```
artembilan, line 482: Same here: no XML support for this custom tag (yet).

artembilan (changes requested):
I see that you copy/pasted a lot of docs from (S)FTP.
Are there some confirmation that all those remote commands we claim for SMB Outbound Gateway work well?

Thanks

GregBragg:
> I see that you copy/pasted a lot of docs from (S)FTP. Are there some confirmation that all those remote commands we claim for SMB Outbound Gateway work well?
> 
> Thanks

Yes agreed, I fully admit that I plagiarized a lot of the FTP Module design and documentation as this part of the code for the Java DSL I am not fully familiar with except from actually using the SFTP module in our batch-service.

All of this code for the SMB Module was carefully patterned after the FTP module design and was compared line by line as I refactored it to work with the code that came from the extensions project so it should be fine, however I admit I need to do some more work around adding more JUnit test cases to verify fully.

When is v6.0 supposed to be GA? If I knew your timelines I would be able to plan to contribute more as my work commitments allow... Thanks!

src/reference/asciidoc/smb.adoc
GregBragg: I'm glad you are catching this stuff... oh my lord, my bad.

src/reference/asciidoc/smb.adoc
```diff
@@ -171,6 +186,138 @@ public class SmbJavaApplication {
...
+Since the session remains open, the consuming application is responsible for closing the session when the file has been consumed.
+The session is provided in the `closeableResource` header (`IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE`).
+Standard framework components, such as the `FileSplitter` and `StreamTransformer`, automatically close the session.
+See <<./file.adoc#file-splitter,File Splitter>> and <<./transformer.adoc#stream-transformer,Stream Transformer>> for more information about these components.
+The following example shows how to configure an `inbound-streaming-channel-adapter`:
+
+====
+[source, xml]
+----
+<int-smb:inbound-streaming-channel-adapter id="smbInbound"
```
artembilan, line 202: But this one is also missed in the XSD...

As I said: we can add an XML support for them or just omit it and go ahead with only Java & Annotations or DSL configuration approach.
But doc anyway has to reflect the current reality to be merged.

artembilan (changes requested):

src/reference/asciidoc/smb.adoc
GregBragg: Yes of course, as I don't have experience with XML configuration in general as I always use Java & Annotations or DSL configuration approach, I will omit those XML configuration examples.

PR #3812, opened 2022-05-24
Move hazelcast to spring integration

RobbanHoglund:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1080,6 +1080,17 @@ project('spring-integration-bom') {
     }
 }
 
+
+project('spring-integration-hazelcast') {
+    description = 'Spring Integration Hazelcast Support'
+    dependencies {
+	implementation "com.hazelcast:hazelcast:$hazelcastVersion"
+	api project(':spring-integration-core')
+
+	testImplementation project(':spring-integration-test')
+    }
+}
+
 sonarqube {
     properties {
         property 'sonar.links.homepage', linkHomepage
--- a/build.gradle
+++ b/build.gradle
@@ -1084,7 +1084,7 @@ project('spring-integration-bom') {
 project('spring-integration-hazelcast') {
     description = 'Spring Integration Hazelcast Support'
     dependencies {
-	implementation "com.hazelcast:hazelcast:$hazelcastVersion"
+	api "com.hazelcast:hazelcast:$hazelcastVersion"
 	api project(':spring-integration-core')
 
 	testImplementation project(':spring-integration-test-support')
--- a/build.gradle
+++ b/build.gradle
@@ -1087,7 +1087,7 @@ project('spring-integration-hazelcast') {
 	implementation "com.hazelcast:hazelcast:$hazelcastVersion"
 	api project(':spring-integration-core')
 
-	testImplementation project(':spring-integration-test')
+	testImplementation project(':spring-integration-test-support')
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -1086,9 +1086,7 @@ project('spring-integration-hazelcast') {
     dependencies {
 	api "com.hazelcast:hazelcast:$hazelcastVersion"
 	api project(':spring-integration-core')
-
-	testImplementation project(':spring-integration-test-support')
-    }
+	}
 }
 
 sonarqube {
--- a/spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/store/HazelcastMessageStoreTests.java
+++ b/spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/store/HazelcastMessageStoreTests.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2017-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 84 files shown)


build.gradle
```diff
@@ -69,6 +69,7 @@ ext {
     hsqldbVersion = '2.6.1'
     h2Version = '2.1.210'
     jacksonVersion = '2.13.3'
+    javaxAnnotationApi = '1.3.2'
```
artembilan, line 72: I believe we don't need this: I don't see you use it someway in deps below.

build.gradle
```diff
@@ -106,6 +107,7 @@ ext {
     springSecurityVersion = project.hasProperty('springSecurityVersion') ? project.springSecurityVersion : '6.0.0-M5'
     springVersion = project.hasProperty('springVersion') ? project.springVersion : '6.0.0-M4'
     springWsVersion = '4.0.0-M2'
+    slf4jVersion = '1.7.36'
```
artembilan, line 110: I wonder if existing `testRuntimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'` is enough.
My point is to keep deps as short as possible and if we really can use some other existing logging tool instead of extra SLF4J, that would be great.

spring-integration-core/src/test/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandlerTests.java
```diff
@@ -157,6 +158,7 @@ void testCustomCombineFunction() {
 	}
 
 	@Test
+	@Disabled("Time-sensitive")
```
artembilan, line 161: Please, consider to do something like this:
```
git checkout main
git pull upstream main
git checkout -
git rebase main
```

and then `push --force` to this PR.

I mean it might not be a single not related to your PR change, so we will suffer reviewing some extra code existing already in the `upstream/main`

artembilan (changes requested):

artembilan (changes requested):

artembilan (changes requested):
Please, run `./gradlew clean check` locally to be sure that you don't violate Checkstyle rules.

artembilan (changes requested):
Thanks. 

Any chances that you can replace all those `@since 1.0.0` in this module to `6.0`?

artembilan:
Still Checkstyle failures in the XML config: must point to `https` for XSD.

spring-integration-hazelcast/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,8 @@
+log4j.rootCategory=WARN, stdout
```
artembilan, line 1: This file has to be changed to `log4j2-test.xml`.
We don't support Log4J 1 any more.

build.gradle
```diff
@@ -1080,6 +1080,17 @@ project('spring-integration-bom') {
...
 }
 
+
+project('spring-integration-hazelcast') {
+    description = 'Spring Integration Hazelcast Support'
+    dependencies {
+	implementation "com.hazelcast:hazelcast:$hazelcastVersion"
+	api project(':spring-integration-core')
+
+	testImplementation project(':spring-integration-test')
```
artembilan, line 1090: I'm not sure why we need this.
Perhaps existing, imported to every modules `testImplementation(project(':spring-integration-test-support')) {` is enough...

build.gradle
```diff
@@ -1080,6 +1080,17 @@ project('spring-integration-bom') {
     }
 }
 
+
+project('spring-integration-hazelcast') {
+    description = 'Spring Integration Hazelcast Support'
+    dependencies {
+	implementation "com.hazelcast:hazelcast:$hazelcastVersion"
```
artembilan, line 1087: Why not `api`?

spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/store/HazelcastMessageStoreTests.java
```diff
@@ -0,0 +1,149 @@
...
+ */
+public class HazelcastMessageStoreTests {
+
+	private static HazelcastMessageStore store;
+
+	private static HazelcastInstance instance;
+
+	private static IMap<Object, Object> map;
+
+	@BeforeClass
```
artembilan, line 54: No need to do this right now, but one day we definitely need to revise all of these tests in favor of JUnit 5...

artembilan (changes requested):
Nothing critical so far.

Please, address those my minor concerns this or other way and I'll move on for final local review and merging.

Thanks

spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/store/HazelcastMessageStoreTests.java
RobbanHoglund: Okay I can take this in future iteration

build.gradle
RobbanHoglund: Hmm I thought Spring integration _only_ used Hazelcast _internally_ and hence not exposing it as an Api, but I might be wrong here?
I did a quick analysis of the Spring integration Api and it seems that Hazelcast´s own Api is not leaked through. But maybe we want to expose its Api as well?

artembilan: Yeah... I know what you mean, but our goal is let end-user just add `spring-integration-hazelcast` dependency and `hazelcast` by itself is going to be pulled transitively.
The `api` is an equivalent of `compile` in Maven, where an `implementation` is mapped to `runtime`.
See more info here: https://stackoverflow.com/questions/44413952/gradle-implementation-vs-api-configuration

RobbanHoglund: Ok fair enough I will change that

artembilan (changes requested):

artembilan (approved):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/a322f5c35d091b8a2a0917c84a3335c79927ff9f after minor clean up.

@RobbanHoglund ,

See my comment above for possible tasks around this new module.

PR #3819, opened 2022-06-06
GH-3805: Change to select between spinLock and pub-sub for the lock method

### LINKED ISSUE #3805: Connection failures and application hangs with RedisLockRegistry since Spring Boot upgrade

5.5.11

**Describe the bug**

We've just recently updated Spring Boot from 2.5.5 to 2.5.13 (so basically upgrade from 5.5.4 to 5.5.11). 
Immediately we received constant connection failure logs:

    ERROR 7 --- [lock-registry-1] o.s.d.r.l.RedisMessageListenerContainer  : Connection failure occurred. Restarting subscription task after 5000 ms

The application also randomly hangs for multiple seconds.

We are using RedisLockRegistry with a ElastiCache Redis (6.0.5) in Primary/Replica mode. For RedisLockRegistry we only connect to the primary.

Our configuration is basically this:

```kotlin
@Bean
fun redisConnectionFactory(
    builderCustomizers: ObjectProvider<LettuceClientConfigurationBuilderCustomizer>,
    clientResources: ClientResources
): LettuceConnectionFactory {
    val clientConfig = getLettuceClientConfiguration(builderCustomizers, clientResources, redisProperties.lettuce.pool)
    val config = RedisStaticMasterReplicaConfiguration(
        redisProperties.host,
        redisProperties.port
    )
    config.username = redisProperties.username
    config.password = RedisPassword.of(redisProperties.password)
    config.database = redisProperties.database
    return LettuceConnectionFactory(config, clientConfig)
}

@Bean
fun lockRegistry(
    redisConnectionFactory: RedisConnectionFactory
): ExpirableLockRegistry = RedisLockRegistry(
    redisConnectionFactory,
    "exampleService",
    120_000
)
```

We had to downgrade back to 2.5.5 to fix the issues for now

**To Reproduce**

Not sure

**Expected behavior**

No hangs / error logs

**Sample**

Not sure how to show

Meteorkor:
Resolves https://github.com/spring-projects/spring-integration/issues/3805

It has been changed so that the user can choose between the spinLock method and the pub-sub method.

In the default, spinLock is set as the default so that there is no problem in the past version.

Change RedisLock to abstract class and leave common parts
Separated into RedisSpinLock and RedisPubSubLock for each specialized part.

Any comments are always welcome.
thank you

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -141,7 +141,7 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 	/**
 	 * It is set via lazy initialization when it is a {@link RedisLockType#PUB_SUB_LOCK}.
 	 */
-	private volatile RedisUnLockNotifyMessageListener unlockNotifyMessageListener;
+	private volatile RedisPubSubLock.RedisUnLockNotifyMessageListener unlockNotifyMessageListener;
 	/**
 	 * It is set via lazy initialization when it is a {@link RedisLockType#PUB_SUB_LOCK}.
 	 */
@@ -175,7 +175,7 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 		Assert.isNull(RedisLockRegistry.this.redisMessageListenerContainer, "'redisMessageListenerContainer' must not have been re-initialized.");
 		Assert.isNull(RedisLockRegistry.this.unlockNotifyMessageListener, "'unlockNotifyMessageListener' must not have been re-initialized.");
 		RedisLockRegistry.this.redisMessageListenerContainer = new RedisMessageListenerContainer();
-		RedisLockRegistry.this.unlockNotifyMessageListener = new RedisUnLockNotifyMessageListener();
+		RedisLockRegistry.this.unlockNotifyMessageListener = new RedisPubSubLock.RedisUnLockNotifyMessageListener();
 		final Topic topic = new ChannelTopic(this.unLockChannelKey);
 		this.redisMessageListenerContainer.setConnectionFactory(connectionFactory);
 		this.redisMessageListenerContainer.setTaskExecutor(this.executor);
@@ -208,18 +208,14 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 
 	/**
 	 * <p>
-	 * 	Because pub-sub does not work in some settings(RedisStaticMasterReplicaConfiguration), The default is {@link RedisLockType.SPIN_LOCK}
-	 * <ul>
-	 *    <li>{@link RedisLockType#SPIN_LOCK}: The lock is acquired by periodically(100ms) checking whether the lock can be acquired.</li>
-	 *    <li>{@link RedisLockType#PUB_SUB_LOCK}: The lock is accuired by redis pub-sub.</li>
-	 * 	</ul>
+	 * 	Because pub-sub does not work in some settings(RedisStaticMasterReplicaConfiguration), The default is {@link RedisLockType#SPIN_LOCK}
 	 * <p>
 	 * Set the type of unlockType, Select the lock method.
 	 *
 	 * @param redisLockType obtain RedisLockType
 	 * @since 5.5.13
 	 */
-	public void setRedisLockType(@NonNull RedisLockType redisLockType) {
+	public void setRedisLockType(RedisLockType redisLockType) {
 		Assert.notNull(redisLockType, "'redisLockType' cannot be null");
 		this.redisLockType = redisLockType;
 	}
@@ -258,17 +254,26 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 		}
 	}
 
+	/**
+	 * <ul>
+	 * <li>{@link RedisLockType#SPIN_LOCK}: The lock is acquired by periodically(100ms) checking whether the lock can be acquired.</li>
+	 * <li>{@link RedisLockType#PUB_SUB_LOCK}: The lock is accuired by redis pub-sub.</li>
+	 * </ul>
+	 */
 	public enum RedisLockType {
-		PUB_SUB_LOCK, SPIN_LOCK;
+		SPIN_LOCK, PUB_SUB_LOCK
 	}
 
 	private Function<String, RedisLock> getRedisLockConstructor(@NonNull RedisLockType redisLockType) {
-		return switch (redisLockType) {
-			case PUB_SUB_LOCK -> RedisPubSubLock::new;
-			case SPIN_LOCK -> RedisSpinLock::new;
-		};
+		switch (redisLockType) {
+			case SPIN_LOCK:
+				return RedisSpinLock::new;
+			case PUB_SUB_LOCK:
+				return RedisPubSubLock::new;
+			default:
+				throw new IllegalArgumentException();
+		}
 	}
-
 
 	private abstract class RedisLock implements Lock {
 
@@ -283,7 +288,8 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 				"end\n" +
 				"return false";
 
-		protected static final RedisScript<Boolean> obtainLockScript = new DefaultRedisScript<>(OBTAIN_LOCK_SCRIPT, Boolean.class);
+		protected static final RedisScript<Boolean>
+				OBTAIN_LOCK_REDIS_SCRIPT = new DefaultRedisScript<>(OBTAIN_LOCK_SCRIPT, Boolean.class);
 
 		protected final String lockKey;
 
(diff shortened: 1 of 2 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -523,11 +495,108 @@ public boolean equals(Object obj) {
...
+		}
+
+		private boolean subscribeLock(long time) throws ExecutionException, InterruptedException {
+			final long expiredTime = System.currentTimeMillis() + time;
+			if (obtainLock()) {
+				return true;
+			}
+
+			if (!(RedisLockRegistry.this.isRunningRedisMessageListenerContainer
+				&& RedisLockRegistry.this.redisMessageListenerContainer.isRunning())) {
```
artembilan, line 560: I think this `redisMessageListenerContainer` property has to be optional so we don't create extra objects if we don't deal with pub-sub.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -563,4 +635,58 @@ private void unlockNotify(String lockKey) {
...
+				"local lockClientId = redis.call('GET', KEYS[1])\n" +
+				"if lockClientId == ARGV[1] then\n" +
+				"  redis.call('PEXPIRE', KEYS[1], ARGV[2])\n" +
+				"  return true\n" +
+				"elseif not lockClientId then\n" +
+				"  redis.call('SET', KEYS[1], ARGV[1], 'PX', ARGV[2])\n" +
+				"  return true\n" +
+				"end\n" +
+				"return false";
+		private final RedisScript<Boolean> obtainLockScript = new DefaultRedisScript<>(OBTAIN_LOCK_SCRIPT, Boolean.class);
```
artembilan, line 649: Isn't this property the same what we have in the `RedisPubSubLock`?
Cannot this go to the super class?

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -563,4 +635,58 @@ private void unlockNotify(String lockKey) {
...
+		protected void removeLockKeyInnerUnlink() {
+			RedisLockRegistry.this.redisTemplate.unlink(this.lockKey);
+		}
+
+		@Override
+		protected void removeLockKeyInnerDelete() {
+			RedisLockRegistry.this.redisTemplate.delete(this.lockKey);
+		}
+
+		private boolean obtainLock() {
```
artembilan, line 684: Isn't this method the same as in the `RedisPubSubLock`?
Cannot it go to the super class, too?

artembilan (changes requested):
This is all great, but need some clarification.

Plus it would be great to add some documentation note with reasoning and links to reliable sources.

Thanks

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -150,6 +125,8 @@ protected boolean removeEldestEntry(Entry<String, RedisLock> eldest) {
 
 	private int cacheCapacity = DEFAULT_CAPACITY;
 
+	private RedisLockType redisLockType = RedisLockType.SPIN_LOCK;
```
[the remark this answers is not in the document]
Meteorkor: Agree that no other types will be added.

However, when provided as a boolean, it was not clear what the operation other than SpinLock would be, so I defined it as an enum.

Should it be managed as a boolean?

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -260,9 +258,34 @@ public void destroy() {
...
 
-	private final class RedisLock implements Lock {
+	public enum RedisLockType {
+		PUB_SUB_LOCK, SPIN_LOCK;
+	}
+
+	private Function<String, RedisLock> getRedisLockConstructor(@NonNull RedisLockType redisLockType) {
+		return switch (redisLockType) {
+			case PUB_SUB_LOCK -> RedisPubSubLock::new;
+			case SPIN_LOCK -> RedisSpinLock::new;
```
artembilan, line 268: We cannot use switch expression because we have to back-port to `5.5.x` which is still based on Java 8.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -226,12 +205,31 @@ public void setCacheCapacity(int cacheCapacity) {
...
+	 *    <li>{@link RedisLockType#SPIN_LOCK}: The lock is acquired by periodically(100ms) checking whether the lock can be acquired.</li>
+	 *    <li>{@link RedisLockType#PUB_SUB_LOCK}: The lock is accuired by redis pub-sub.</li>
+	 * 	</ul>
+	 * <p>
+	 * Set the type of unlockType, Select the lock method.
+	 *
+	 * @param redisLockType obtain RedisLockType
+	 * @since 5.5.13
+	 */
+	public void setRedisLockType(@NonNull RedisLockType redisLockType) {
```
artembilan, line 222: We assume that everything is not null, unless it is marked as `@Nullable`.
So, we typically don't use this annotation at all, unless we declare a `@NonNullApi` in the `package-info.java`.
Then we need to revise all the public API to honor nullability contract, so Kotlin projects won't fail.

Since there is only this `RedisLockRegistry` in this `org.springframework.integration.redis.util`, i won't mind if you address this concern in this PR as well.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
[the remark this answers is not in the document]
artembilan: Yeah... After giving it more thoughts I agree with you.
Just have it as `spinLock = true` doesn't say us immediately and what is going to be if we set it to `false` 😄

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -260,9 +258,34 @@ public void destroy() {
 		}
 	}
 
-	private final class RedisLock implements Lock {
+	public enum RedisLockType {
+		PUB_SUB_LOCK, SPIN_LOCK;
```
artembilan, line 262: These constants have to have JavaDocs: they are public API.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -524,18 +526,111 @@ private RedisLockRegistry getOuterType() {
...
+				"return false";
+
+		private static final String DELETE_UNLOCK_SCRIPT =
+				"if (redis.call('del', KEYS[1]) == 1) then " +
+				"redis.call('publish', ARGV[1], KEYS[1]) " +
+				"return true " +
+				"end " +
+				"return false";
+
+		private static final RedisScript<Boolean> unLinkUnLockScript = new DefaultRedisScript<>(UNLINK_UNLOCK_SCRIPT, Boolean.class);
```
artembilan, line 547: If it is `static final`, then it is a constant and therefore has to be declared only in capital letters.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -524,18 +526,111 @@ private RedisLockRegistry getOuterType() {
...
 				}
 			}
 		}
+
 	}
 
+	/**
+	 * Used by RedisPubSubLock
+	 */
 	private static final class RedisUnLockNotifyMessageListener implements MessageListener {
```
artembilan, line 634: This is `private` class - cannot it move into `RedisPubSubLock` class directly to preserve a context?

artembilan (changes requested):
I'm asking about docs one more time 😄 

Thanks

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -260,9 +254,44 @@ public void destroy() {
...
+	 * <li>{@link RedisLockType#SPIN_LOCK}: The lock is acquired by periodically(100ms) checking whether the lock can be acquired.</li>
+	 * <li>{@link RedisLockType#PUB_SUB_LOCK}: The lock is accuired by redis pub-sub.</li>
+	 * </ul>
+	 */
+	public enum RedisLockType {
+		SPIN_LOCK, PUB_SUB_LOCK
+	}
 
-		private final String lockKey;
+	private Function<String, RedisLock> getRedisLockConstructor(@NonNull RedisLockType redisLockType) {
```
artembilan, line 267: Remove, please, this `@NonNull` as we discussed before.

artembilan (changes requested):
I'd like to merge this, but there is need some documentation note to be added into `redis.adoc` in the `[[redis-lock-registry]]` section.

Let me know if you add it or you want me to do that for you!

Thank you.

Meteorkor:
> redis.adoc
@artembilan 
Oh thanks, could you please fix it for me?

Thank you.

artembilan (approved):
Pulling this locally for final review and docs.

I will fix Checkstyle violation on merge:

```
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java:54:8: Unused import - org.springframework.lang.NonNull. [UnusedImports]
```

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/cf6ce961a2bb1be650b9f6fa0b343267b3a63757 and cherry-picked to `5.5.x`

Issue #3822, opened 2022-06-07
Mqtt V5 improve automaticReconnect


kroutal:
**Expected Behavior**
Mqtt V5 automaticReconnect should work when first connection fails.
This is related to : https://github.com/eclipse/paho.mqtt.java/issues/930

**Current Behavior**
When application starts and server is down or something goes wrong during connection then In Mqttv5PahoMessageHandler/Mqttv5PahoMessageDrivenChannelAdapter there will never be a reconnection attempt even though automaticReconnect is set to true

**Context**

The only way to reconnect is to add a MqttConnectionFailedEvent listener. And recode the reconnection mechanism using Timers.
If this is the intended behavior then it should be documented.

artembilan:
If such a behavior is not available in the Paho client, then I'm not sure what would you expect from us to here in our channel adapters.
And indeed the `MqttConnectionFailedEvent` you've mentioned should be a good workaround for time being. 
This is the current logic in the `Mqttv5PahoMessageHandler`:
```
	protected void doStart() {
		try {
			this.mqttClient.connect(this.connectionOptions).waitForCompletion(getCompletionTimeout());
		}
		catch (MqttException ex) {
			ApplicationEventPublisher applicationEventPublisher = getApplicationEventPublisher();
			if (applicationEventPublisher != null) {
				applicationEventPublisher.publishEvent(new MqttConnectionFailedEvent(this, ex));
			}
			logger.error(ex, "MQTT client failed to connect. Will retry if 'ConnectionOptions.isAutomaticReconnect()'.");
		}
	}
```

so, if we fail to start/connect we emit that `MqttConnectionFailedEvent`.
And I believe it is explained well in [docs](https://docs.spring.io/spring-integration/docs/current/reference/html/mqtt.html#mqtt-events):
> `MqttConnectionFailedEvent` - published by both adapters if we fail to connect or a connection is subsequently lost.

Does it make sense?

kroutal:
Thanks for your answer.

I was thinking of something similar to what is done in the MQTT V3 channel adapters, where a scheduler is started when connection fails on startup:
https://github.com/spring-projects/spring-integration/blob/e855f13d8e70eb1cbe280c5bdb41aa98e24ecb79/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java#L183

[...]

PR #3823, opened 2022-06-09
GH-3822 Allow MQTTv5 reconnection, even when first connection fails

kroutal:
GH-3822: Allow MQTTv5 reconnection, even when first connection fails

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -90,7 +90,7 @@ public class Mqttv5PahoMessageDrivenChannelAdapter extends AbstractMqttMessageDr
 	public Mqttv5PahoMessageDrivenChannelAdapter(String url, String clientId, String... topic) {
 		super(url, clientId, topic);
 		this.connectionOptions = new MqttConnectionOptions();
-		this.connectionOptions.setServerURIs(new String[]{ url });
+		this.connectionOptions.setServerURIs(new String[] {url});
 		this.connectionOptions.setAutomaticReconnect(true);
 	}
 
@@ -165,14 +165,17 @@ public class Mqttv5PahoMessageDrivenChannelAdapter extends AbstractMqttMessageDr
 		ApplicationEventPublisher applicationEventPublisher = getApplicationEventPublisher();
 		try {
 			this.mqttClient.connect(this.connectionOptions).waitForCompletion(getCompletionTimeout());
-		} catch (MqttException ex) {
+		}
+		catch (MqttException ex) {
 			if (this.connectionOptions.isAutomaticReconnect()) {
 				try {
 					this.mqttClient.reconnect();
-				} catch (MqttException e) {
+				}
+				catch (MqttException e) {
 					logger.error(ex, "MQTT client failed to connect. Will retry.");
 				}
-			} else {
+			}
+			else {
 				if (applicationEventPublisher != null) {
 					applicationEventPublisher.publishEvent(new MqttConnectionFailedEvent(this, ex));
 				}
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -172,7 +172,7 @@ public class Mqttv5PahoMessageDrivenChannelAdapter extends AbstractMqttMessageDr
 					this.mqttClient.reconnect();
 				}
 				catch (MqttException e) {
-					logger.error(ex, "MQTT client failed to connect. Will retry.");
+					logger.error(ex, "MQTT client failed to connect. Never happens.");
 				}
 			}
 			else {
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -90,7 +90,7 @@ public class Mqttv5PahoMessageDrivenChannelAdapter extends AbstractMqttMessageDr
 	public Mqttv5PahoMessageDrivenChannelAdapter(String url, String clientId, String... topic) {
 		super(url, clientId, topic);
 		this.connectionOptions = new MqttConnectionOptions();
-		this.connectionOptions.setServerURIs(new String[] {url});
+		this.connectionOptions.setServerURIs(new String[]{ url });
 		this.connectionOptions.setAutomaticReconnect(true);
 	}
 
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -66,6 +66,7 @@ import org.springframework.util.Assert;
  *
  * @author Artem Bilan
  * @author Mikhail Polivakha
+ * @author Lucas Bowler
  *
  * @since 5.5.5
  *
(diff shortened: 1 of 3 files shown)


artembilan (changes requested):
There are some Checktyle violations:
```
 Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java:168:17: '}' at column 3 should be alone on a line. [RightCurly]
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java:172:33: '}' at column 5 should be alone on a line. [RightCurly]
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java:175:25: '}' at column 4 should be alone on a line. [RightCurly]
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java:176:25: '}' at column 4 should be alone on a line. [RightCurly]
```

Please, run this command locally `./gradlew clean :spring-integration-mqtt:check`

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -79,7 +80,7 @@ public class Mqttv5PahoMessageHandler extends AbstractMqttMessageHandler
 	public Mqttv5PahoMessageHandler(String url, String clientId) {
 		super(url, clientId);
 		this.connectionOptions = new MqttConnectionOptions();
-		this.connectionOptions.setServerURIs(new String[]{ url });
+		this.connectionOptions.setServerURIs(new String[] {url});
```
artembilan, line 83: Please, revert this one.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -162,28 +163,24 @@ protected void onInit() {
...
-				logger.debug(message);
-				if (applicationEventPublisher != null) {
-					applicationEventPublisher.publishEvent(new MqttSubscribedEvent(this, message));
-				}
-			}
 		}
 		catch (MqttException ex) {
-			if (applicationEventPublisher != null) {
-				applicationEventPublisher.publishEvent(new MqttConnectionFailedEvent(this, ex));
+			if (this.connectionOptions.isAutomaticReconnect()) {
```
artembilan, line 170: This one is a bit different.
The message driven channel adapter is really an active one and does its stuff automatycally - it produces messages whenever they are available on the MQTT topic(s).
So, I guess the logic you did here is OK: 
1. If we connected and automatic reconnect is set, then we are good to go.
2. If we didn't connect, we must reconnect, but if there is no automatic reconnect, we cannot guarantee that it will reconnect on that async `reconnect()`, so we indeed should just throw that exception and let end-user to control it from the outside.
3. If it us auto, then we call `reconnect()` and it has to be enough for eventual success connection.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -162,28 +163,24 @@ protected void onInit() {
...
+				try {
+					this.mqttClient.reconnect();
+				}
+				catch (MqttException e) {
+					logger.error(ex, "MQTT client failed to connect. Will retry.");
```
artembilan, line 175: This will never happen here, but won't hurt to have it. Although `Will retry` is a bit misleading.
Would you mind to remove it and add a comment like - `Never happens`?

artembilan (changes requested):
Here is my further review.

Thank you!

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
kroutal: Done

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
kroutal: Done

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -165,11 +166,21 @@ protected void doStart() {
 			this.mqttClient.connect(this.connectionOptions).waitForCompletion(getCompletionTimeout());
 		}
 		catch (MqttException ex) {
-			ApplicationEventPublisher applicationEventPublisher = getApplicationEventPublisher();
-			if (applicationEventPublisher != null) {
-				applicationEventPublisher.publishEvent(new MqttConnectionFailedEvent(this, ex));
+			if (this.connectionOptions.isAutomaticReconnect()) {
```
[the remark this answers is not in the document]
kroutal: I'm not sure why we need to call `reconnect()`. For me the only purpose of calling `reconnect()` was to fix the automaticReconnect option. But maybe I've missed something. 
I've updated the code to connect in `handleMessageInternal()`. But the code probably needs to be improved in order to avoid spamming the MQTT broker with invalid connection (bad credentials, ... )

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
kroutal: I'm not sure why we need to call `reconnect()` when we are not using auto

artembilan: Yeah... I see your point: let's handle that as a `MqttConnectionFailedEvent` instead!

In other words: I agree with what you have so far.


Sorry for a noise!

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -250,6 +247,9 @@ else if (payload instanceof String) {
 	protected void publish(String topic, Object mqttMessage, Message<?> message) {
 		Assert.isInstanceOf(MqttMessage.class, mqttMessage, "The 'mqttMessage' must be an instance of 'MqttMessage'");
 		try {
+			if (!this.mqttClient.isConnected()) {
```
artembilan, line 250: Would you mind removing that `if (!this.connectionOptions.isAutomaticReconnect()) {` in the ctor since now we do reconnect ourselves.
Or if it is there, then we rely on the auto-reconnect.
Nothing critical for us from now on.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackAutomaticReconnectTests.java
```diff
@@ -0,0 +1,207 @@
...
+import org.springframework.messaging.PollableChannel;
+import org.springframework.messaging.converter.AbstractMessageConverter;
+import org.springframework.messaging.converter.SmartMessageConverter;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+
+/**
+ * @author Lucas Bowler
+ * @since 5.5.x
```
artembilan, line 64: `5.5.13`.
And remove, please, that `2002` in the copyright header.

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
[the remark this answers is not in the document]
artembilan: I think that one is already out of our scope.
For that reason this Paho MQTT client raises this:
```
		if ((reasonCode == MqttClientException.REASON_CODE_FAILED_AUTHENTICATION) || 
			(reasonCode == MqttClientException.REASON_CODE_NOT_AUTHORIZED)) {
			return new MqttSecurityException(reasonCode);
		}
```
So, it's up to end user to catch this exception this or other way and deal with it respectively.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/8cf5f9083bd7834ca4d7f4956338a445b4d76ca3 and cherry-picked to `5.5.x` after some clean up and minor tests & docs improvements.

@kroutal ,

PR #3825, opened 2022-06-22
Use `lookupHost = false` by default for TCP & UDP

artembilan:
The applications these days more and more are deployed and managed in the containers
where DNS is not configured by default.
Having `lookupHost = true` by default leads to a bad experience when some delays happen
for reverse host lookups.

* Use `lookupHost = false` by default for both TCP & UDP to have a reliable behavior
independently of the environment.
The `hostName` is used for `connectionId` and as a header in the message -
semantically it doesn't matter for the application logic what value is present over there.


artembilan:
See more info here: https://stackoverflow.com/questions/72700072/spring-integration-container-application-hangs-for-a-few-seconds-during-connect
I won't mind, though, rejecting this if there is really a big reason to keep `hostName` in headers as a name, not IP.

Thanks.

garyrussell (changes requested):

artembilan:
Also Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.x-to-6.0-Migration-Guide#no-lookup-dns-host-by-default

PR #3829, opened 2022-06-27
GH-3826: Fix SimplePool for resizing from MAX

### LINKED ISSUE #3826: CachingSessionFactory slowness if setPoolSize is called without initializing initial sessionCacheSize
Version :
5.5.10

Migrating from 5.1.13 and 5.5.10

**Describe the bug**

We were doing a spring 2.6 migratoin and hence also upgrading spring integration version to 5.5.10 . Our existing code used to initialize CachingSessionFactory as below

```
      CachingSessionFactory<LsEntry> csf = new CachingSessionFactory<LsEntry>(sf ); 
       scsf.setPoolSize(connectionInfo.getPoolSize());
      // Refer https://github.com/spring-projects/spring-integration/pull/2606
      csf.setTestSession(true);
      csf.setSessionWaitTimeout(props.getSftp()
            .getSessionWaitTimeout());
```
Post migrating we observed that the `setPoolSize` method was degrading our performance and in itself taking aroudn 30 seconds . This is mainly due to the method calling `SimplePool.setPoolSize` which as the poolsize was not initialized keeps a default pool size as Integer.MAX_VALUE and hence when setPoolSize tries to reduce the pool size it goes into a long loop in order to clear it up . The same worked in the old version as the old code used to break in case there was nothing returned on poll . 

**Diff here** https://github.com/spring-projects/spring-integration/compare/v5.1.13.RELEASE...v5.5.10#diff-1da249bba02f27de259b1f47687338cb6204ca4754afd457ef82a1d17c912ca1 

**To Reproduce**

The below code 

```
      CachingSessionFactory<LsEntry> csf = new CachingSessionFactory<LsEntry>(sf ); 
       scsf.setPoolSize(connectionInfo.getPoolSize());
      // Refer https://github.com/spring-projects/spring-integration/pull/2606
      csf.setTestSession(true);
      csf.setSessionWaitTimeout(props.getSftp()
            .getSessionWaitTimeout());
```

**Expected behavior**
 
I think the design makes sense to clear the pool but it could lead to performance issues in case the poolsize is not initialized at the start like ` CachingSessionFactory<LsEntry> csf = new CachingSessionFactory<LsEntry>(sf, connectionInfo.getPoolSize()); ` . There could be a fix or this needs to be documented in code so people take care in case of upgrades.  

**Sample**

```
      CachingSessionFactory<LsEntry> csf = new CachingSessionFactory<LsEntry>(sf ); 
       scsf.setPoolSize(connectionInfo.getPoolSize());
      // Refer https://github.com/spring-projects/spring-integration/pull/2606
      csf.setTestSession(true);
      csf.setSessionWaitTimeout(props.getSftp()
            .getSessionWaitTimeout());
```

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3826

By default, the `SimplePool` is used with an `Integer.MAX_VALUE` pool size.
There is a performance degradation in the `setPoolSize()` when we try to
decrease the pool size: the `while` for `permits.tryAcquire()` is too long
close to the current `Integer.MAX_VALUE` pool size

* Revise the logic in the `setPoolSize()` to use `Semaphore.reducePermits()`
instead of the loop.
* Change the calculation for a new pool size for the current pool state:
or it is a size of a new request, or iti s equal to the `inUse.size()`.
It will be reduced on subsequent `releaseItem()` calls
* Reduce the number of `available` according a new pool size based on the `inUse`.
So, if `inUse > newPoolSize`, the `available` is cleared.
Otherewise, it is reduced to the number which would give `newPoolSize` together
with the `inUse` size

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -48,7 +48,7 @@ public class SimplePool<T> implements Pool<T> {
 
 	protected final Log logger = LogFactory.getLog(getClass()); // NOSONAR final
 
-	private final Semaphore permits = new Semaphore(0);
+	private final PoolSemaphore permits = new PoolSemaphore(0);
 
 	private final AtomicInteger poolSize = new AtomicInteger();
 
@@ -56,11 +56,11 @@ public class SimplePool<T> implements Pool<T> {
 
 	private long waitTimeout = Long.MAX_VALUE;
 
-	private final BlockingQueue<T> available = new LinkedBlockingQueue<T>();
+	private final BlockingQueue<T> available = new LinkedBlockingQueue<>();
 
-	private final Set<T> allocated = Collections.synchronizedSet(new HashSet<T>());
+	private final Set<T> allocated = Collections.synchronizedSet(new HashSet<>());
 
-	private final Set<T> inUse = Collections.synchronizedSet(new HashSet<T>());
+	private final Set<T> inUse = Collections.synchronizedSet(new HashSet<>());
 
 	private final PoolItemCallback<T> callback;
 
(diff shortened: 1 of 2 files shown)


artembilan:
@gauravbrills,

here is some possible fix for the problem.
I'm not sure that is perfect, but still would like to share to start some discussion and gather other options.

Thanks

/CC @garyrussell

spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
```diff
@@ -266,6 +272,19 @@ public synchronized void close() {
 		removeAllIdleItems();
 	}
 
+	private static class PoolSemaphore extends Semaphore {
+
+		public PoolSemaphore(int permits) {
```
garyrussell, line 277: Should be package protected (will solve checkstyle too).

garyrussell (changes requested):
LGTM, with change.

spring-integration-core/src/main/java/org/springframework/integration/util/SimplePool.java
```diff
@@ -105,21 +105,27 @@ public synchronized void setPoolSize(int poolSize) {
...
+			this.poolSize.set(newPoolSize);
+
+			for (int i = this.available.size(); i > newPoolSize - inUseSize; i--) {
 				T item = this.available.poll();
 				if (item != null) {
 					doRemoveItem(item);
 				}
-				this.poolSize.decrementAndGet();
-				delta++;
+				else {
```
gauravbrills, line 119: Guess this should help in the default init case we had

artembilan:
This has been merged as https://github.com/spring-projects/spring-integration/commit/6fe59a76aa7def82a8f38bee742e9fe5f021e511.
Thank you!

PR #3830, opened 2022-06-30
Add assertions and comment on the utility class in Spring Integration IP.

ZahidFKhan:
Added assertion on the utility class, so that it can't be mistakenly ever initialized in the class.

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
@@ -35,7 +35,7 @@ public final class TestingUtilities {
 
 	// Non-instantiable utility class
 	private TestingUtilities() {
-		throw new AssertionError();
+		throw new AssertionError(""Class Instantiation not allowed."");
 	}
 
 	/**
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
@@ -33,7 +33,9 @@ import org.springframework.lang.Nullable;
  */
 public final class TestingUtilities {
 
+	// Non-instantiable utility class
 	private TestingUtilities() {
+		throw new AssertionError();
 	}
 
 	/**
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
@@ -35,7 +35,7 @@ public final class TestingUtilities {
 
 	// Non-instantiable utility class
 	private TestingUtilities() {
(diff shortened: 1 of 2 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
```diff
@@ -33,7 +33,9 @@
  */
 public final class TestingUtilities {
 
+	// Non-instantiable utility class
 	private TestingUtilities() {
+		throw new AssertionError();
```
artembilan, line 38: This is `private`, so it cannot be instantiate from the outside of the class.
Even if you do some hardcore via reflection, it still does not harm since there is no any state in these classes to break something.
I'm not sure that is going to be mistake in the framework code to call this ctor directly in this class: we always review changes, so such a code won't pass it to the mainstream.

Please, elaborate on the purpose of this change.

Thanks

artembilan (changes requested):

spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
ZFK07: I agree @artembilan that this is a `private` utility class and it can't be instantiated from outside of the class.

The `AssertionError` isn’t strictly required at all but it provides insurance in case the constructor is accidentally invoked from within the class not now but at some point in time. It guarantees that the class will never be instantiated under any circumstances. It is therefore wise to include a comment as well.

artembilan: OK. So, add, please, a message to that error and revert that JavaDocs format. We prefer to not have blank line in the method JavaDocs.
Plus, bring that blank line before `@since` on the class level - for better readability 😄

artembilan (changes requested):

spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
artembilan: Tell me, please, if you are going to address other my suggestions.
Thanks

artembilan (changes requested):

spring-integration-ip/src/main/java/org/springframework/integration/ip/util/TestingUtilities.java
ZFK07: Yes, @artembilan I am working on the above suggestions. :smile: 
And I haven't modified any of the java docs. Seems there is some problem with CRLF and LF in my IDE. :see_no_evil:

ZFK07:
Hi @artembilan would you like to verify again, please?

ZFK07:
Thank you so much @artembilan 

This is my first contribution to the Open Source, definitely, I would love to contribute more to the Spring Ecosystem.
Looking forward to working with you in the future. 😄😇

PR #3837, opened 2022-07-07
Fix IntegrationMBeanExporter logic for endpoints

artembilan:
Related to https://stackoverflow.com/questions/72851234/error-in-startup-application-when-using-serviceactivator-in-spring-cloud-stream

The `MessagingAnnotationPostProcessor` register an `endpoint` bean for Messaging Annotation on POJO methods.
The `IntegrationMBeanExporter` post-process such a bean and registers respective MBean.
It does this not in optimal way scanning all the `IntegrationConsumer` beans for requested `MessageHandler`
which may cause a `BeanCurrentlyInCreationException`.

* Rework the logic of the `IntegrationMBeanExporter.postProcessAbstractEndpoint()` to propagate provided endpoint
for the monitor registration to bypass application context scanning for matched name.
* Swap `equals()` for `monitor` since an `extractTarget()` may return `null`
* Some other code clean in the `IntegrationMBeanExporter`
* Change one of the `@ServiceActivator` in the configuration for the `ScatterGatherHandlerIntegrationTests`
to POJO method.
However, this didn't fail for me with original code unlike in the sample application provided in the mentioned SO thread.

**Cherry-pick to `main`**

### DIFF
--- a/spring-integration-jmx/src/main/java/org/springframework/integration/monitor/IntegrationMBeanExporter.java
+++ b/spring-integration-jmx/src/main/java/org/springframework/integration/monitor/IntegrationMBeanExporter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -264,8 +264,7 @@ public class IntegrationMBeanExporter extends MBeanExporter
 	}
 
 	private void populateMessageSources() {
-		this.applicationContext.getBeansOfType(
-				IntegrationInboundManagement.class)
+		this.applicationContext.getBeansOfType(IntegrationInboundManagement.class)
 				.values()
 				.stream()
 				// If the source is proxied, we have to extract the target to expose as an MBean.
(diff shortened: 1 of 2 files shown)


garyrussell:
FYI, I see the same error for both SAs in the app I wrote for this answer https://stackoverflow.com/questions/72900966/producer-callback-in-spring-cloud-stream-using-streambridge/72901322#72901322

I tried to test it by merging the patch into 5.5.x and `./gradlew clean publishToMavenLocal`, but it failed:

```
% ./gradlew clean publishToMavenLocal
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details
> Task :spring-integration-graphql:generatePomFileForMavenJavaPublication FAILED
> Task :spring-integration-hazelcast:generatePomFileForMavenJavaPublication FAILED

> Task :spring-integration-core:compileKotlin
Could not perform incremental compilation: Could not connect to Kotlin compile daemon
Could not connect to kotlin daemon. Using fallback strategy.
warning: API version 1.3 is deprecated and its support will be removed in a future version of Kotlin

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/Users/grussell/Development/spring-integration/build.gradle' line: 391

* What went wrong:
Execution failed for task ':spring-integration-graphql:generatePomFileForMavenJavaPublication'.
> Could not apply withXml() to generated POM
   > Cannot access first() element from an empty List

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
==============================================================================
```

Perhaps the recent gradle upgrade?

artembilan:
Saw that Gary, as well.
The problem with Gradle is in our `settings.gradle`:
```
	if (dir.name.startsWith('spring-integration-')) {
		include ":${dir.name}"
	}
```
But those new modules exist only in `main`.
You need to remove `spring-integration-graphql`, `spring-integration-hazelcast` and `spring-integration-smb` dirs from the project to let it to build artifacts properly.
Not sure why Git doesn't remove those dirs after swithing between branches...

Issue #3804, opened 2022-05-19
Setup/Installation of a SMB Server for JUnit Testing in SMB Module


GregBragg:
I wonder if we can look into using a Testcontainers with Samba Docker image for these kinds of JUnit tests: https://hub.docker.com/r/dperson/samba ...

I definitely looked into some sort of embedded SMB Server for a few days, like what we have for (S)FTP however there is none available. I agree we could have Samba container for the build process to use instead, it doesn't help however for localhost testing against an external SMB share, however we can update the JUnit test classes with a different config while the SMB Module code is under development.

_Originally posted by @GregBragg in https://github.com/spring-projects/spring-integration/pull/3802#discussion_r876067290_

PR #3842, opened 2022-07-11
GH-3804: Migrate SMB unit tests to Testcontainers

oxcafedead:
Instead of manual SMB server setup and preparation, make it all
automated using convenient Testcontainers library and a bit of extra
code executed before tests.

More detailed,
* Change `localhost` to `127.0.0.1` for a proper integration with
docker container
* A few new config values to support container environment, like tmpFs,
share permissions and a dynamic port value
* Introduce Testcontainers API for container management
* Few configuration changes to support docker container usage
* Automatic dirs and files layout creation in `@BeforeAll`
* A few bugfixes where the assertion has not been done actually
* Couple of small changes related to assertions readability as well

### DIFF
--- a/spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
+++ b/spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
@@ -52,12 +52,10 @@ public class SmbTestSupport extends RemoteFileTestSupport {
 
 	private static final String INNER_SHARE_DIR = "/tmp";
 
-	// Configuration details can be found at https://hub.docker.com/r/dperson/samba
-	private static final GenericContainer<?> SMB_CONTAINER = new GenericContainer<>("dperson/samba:latest")
-			.withExposedPorts(445)
-			.withEnv("USER", USERNAME + ";" + PASSWORD)
+	private static final GenericContainer<?> SMB_CONTAINER = new GenericContainer<>("elswork/samba:4.15.5")
 			.withTmpFs(Map.of(INNER_SHARE_DIR, "rw"))
-			.withEnv("SHARE", SHARE_AND_DIR + ";" + INNER_SHARE_DIR + ";yes;no;no;all;none");
+			.withCommand("-u", "1000:1000:" + USERNAME + ":" + USERNAME + ":" + PASSWORD, "-s", SHARE_AND_DIR + ":" + INNER_SHARE_DIR + ":rw:" + USERNAME)
+			.withExposedPorts(445);
 
 	private static SmbSessionFactory smbSessionFactory;
 
--- a/spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
+++ b/spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
@@ -16,8 +16,11 @@
 
 package org.springframework.integration.smb;
 
+import java.util.Map;
+
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.junit.jupiter.Testcontainers;
 
 import org.springframework.integration.file.remote.RemoteFileTestSupport;
 import org.springframework.integration.file.remote.session.CachingSessionFactory;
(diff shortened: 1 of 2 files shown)


oxcafedead:
Seems like there's some issue in core part, not related to this PR:
```
> Task :spring-integration-core:test
*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" with message transform method call failed at ../src/java.instrument/share/native/libinstrument/JPLISAgent.c line: 873
...
```

artembilan:
Hi @oxcafedead !

Please, consider to rebase your branch to the the latest `upstream/main`: I've pushed the fixes for `TaskScheduler` and `Trigger` breaking changes in Spring Framework.

Re. `*** java.lang.instrument ASSERTION FAILED ***`. Yes, I've seen this for all the years I work with this project.
I couldn't find a proper explanation of the problem in the Internet, e.g.: https://stackoverflow.com/questions/60526670/what-is-the-meaning-cause-of-java-lang-instrument-assertion-failed-er
Doesn't look like this affect somehow the project build or its artifacts.

IF you wish to investigate it and find the cause, I would suggest to do that in the separate branch.
Start from the plain `./gradlew :spring-integration-core:check` to see if it still appears.
Then probably comment/remove test dirs to minimize the scope and so on.

spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
```diff
@@ -28,35 +31,46 @@
...
+	public static final String SHARE_AND_DIR = "smb-share";
 
 	public static final String USERNAME = "sambaguest";
 
 	public static final String PASSWORD = "sambaguest";
 
+	private static final String INNER_SHARE_DIR = "/tmp";
+
+	// Configuration details can be found at https://hub.docker.com/r/dperson/samba
+	private static final GenericContainer<?> SMB_CONTAINER = new GenericContainer<>("dperson/samba:latest")
```
artembilan, line 56: The `latest` is not OK. Seems for me most of the images are resolved against snapshots.
So, better to use an explicit version.

artembilan (changes requested):
I'll accept this if it really confirms that all tests for SMB in the project works well.

/CC @GregBragg

spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
oxcafedead: Agree, however this particular image does not have version tags (https://github.com/dperson/samba/issues/394).
In theory we could use sha256 digests, but I think this is not a good solution as well as it will be dependent on a concrete CPU architecture (if my understanding is correct).
Maybe you have any ideas?

oxcafedead:
> Hi @oxcafedead !
> 
> Please, consider to rebase your branch to the the latest `upstream/main`: I've pushed the fixes for `TaskScheduler` and `Trigger` breaking changes in Spring Framework.

Hi @artembilan.
Sure, I can do it, just want to double check: won't force pushing break existing pull request mappings? like comments etc.

spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java
artembilan: I don't know. Maybe we need to use some other vendor image, e.g. https://github.com/crazy-max/docker-samba ?..

oxcafedead: It, makes sense, I will try to change the code to be able to use another one.

GregBragg:
> I'll accept this if it really confirms that all tests for SMB in the project works well.
> 
> /CC @GregBragg

Yes, as long as all the test case passes I'm good with the change. Thanks for the extra effort!

oxcafedead:
My quick research: seems like libs incompatibility:
we have:
> hibernateVersion = '5.6.8.Final'
but the class org/hibernate/query/BindableType is only in hibernate `6.0` version

artembilan:
Thank you @oxcafedead !

I'm looking now into that Hibernate problem.
Looks like Spring Data JPA has moved to Hibernate 6.x, so we need to catch up.
This really has nothing to do with the module you are working in.
There is just enough for you to test your changes with this command: `./gradlew :spring-integration-smb:check`

oxcafedead:
> There is just enough for you to test your changes with this command: `./gradlew :spring-integration-smb:check`


Yeah, sure, I have checked it locally before pushing changes to the branch. Smb tests work.

oxcafedead:
@artembilan my bad, did it occasionally, usually I sync my master branch in the fork by fetching via UI, but this time feature branch was selected. I have rebased it already.

artembilan (approved):
LGTM.

Pulling locally for final review and merge.

GregBragg:
> Merged as [f349e1a](https://github.com/spring-projects/spring-integration/commit/f349e1a114236e0e15fc712574b5d58f774c4769) after some minor code clean up and making tests as cross-platform. On Windows the file separator is `\`, but we use `/` in our tests, so I had to do this for example: `TestUtils.applySystemFileSeparator("smbSource/smbSource1.txt")` to make it compatible with the target `SmbSession` expectations.
> 
> Note.
> 
> Probably not related to this change and nothing fails any more, but I see this failure in logs:
> 
> ```
> ```
> 
> Probably there is need in some extra security configuration for that Samba server to let us to get access to the file permissions...
> 
> @oxcafedead ,
> 
> thank you for a great contribution anyway!

I concur, probably related to file permission access for the userid that is logged into to the test container SMB server. This is a non-critical method call anyways as it just gets a formatted list of the directory and file permissions that the userid has access to, and is informational.

@oxcafedead Great work! I can still modify the test classes as-is to connect to a 'real' Windows file server that I use today for development testing. Thanks!

PR #3845, opened 2022-07-13
Initial support for Micrometer Observation

artembilan:
* Add respective Observation dependencies
* Refactor an `AbstractMessageHandler` logic for potential Observation hooks
* Introduce an `ObservationPropagationChannelInterceptor` to propagate
an `Observation` from one thread to another through message channels

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/interceptor/ObservationPropagationChannelInterceptor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/interceptor/ObservationPropagationChannelInterceptor.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.channel.interceptor;
+
+import org.springframework.aop.support.AopUtils;
+import org.springframework.integration.channel.DirectChannel;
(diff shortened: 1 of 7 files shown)


artembilan:
This is just initial support and some clean up.
I decided to go baby steps to make it easier for review.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/channel/interceptor/ObservationPropagationChannelInterceptor.java
```diff
@@ -0,0 +1,77 @@
...
+ * flow's thread to another through the {@link MessageChannel}s involved in the flow.
+ * Opens a new {@link Observation.Scope} on another thread and cleans up it in the end.
+ *
+ * @author Artem Bilan
+ *
+ * @since 6.0
+ */
+public class ObservationPropagationChannelInterceptor extends ThreadStatePropagationChannelInterceptor<Observation> {
+
+	private final ThreadLocal<Observation.Scope> scopes = new ThreadLocal<>();
```
marcingrzejszczak, line 42: Theoretically instead of storing scopes yourself you can think of using `ObservationThreadLocalAccessor` (from micrometer-core) that pretty much does the same. The only thing to consider is that you would need to add `io.micrometer:context-propagation` to the classpath as a dependency (since in micrometer-core that dependency is optional)

marcingrzejszczak (approved):

marcingrzejszczak:
I've created a PR to your PR to see if the `Sender` and `Receiver` Contexts are working well with messaging and it seems that they are (https://github.com/spring-projects/spring-integration/pull/3856).

spring-integration-core/src/main/java/org/springframework/integration/channel/interceptor/ObservationPropagationChannelInterceptor.java
artembilan: I thought about that and we had with you private discussion on the matter.
My only concern that `ObservationThreadLocalAccessor` is not able to be injected with an external `ObservationRegistry` as it would be in our Spring cases.
But if you insist, I'll change this interceptor to reuse that propagation API.

Thanks

marcingrzejszczak: No, I don't insist at all, it was just an idea. You can leave the scopes, I don't see a problem with that.

PR #3847, opened 2022-07-14
GH-3840: Migrate Redis tests to Testcontainers

### LINKED ISSUE #3840: Migrate Redis integration tests to Testcontainers
Currently our integration tests rely on the Redis server provided by the host of the test execution and use JUnit 4 rule for condition evaluation.
We cannot have externally configured Redis server on CI environment anymore, so SonarQube reports pure coverage, since most of Redis tests are ignored.

Switch integration tests for Redis to Testcontainers since Docker is available in most modern environments.
This way we also will be able to migrate all those tests to JUnit 5.

See `MosquittoContainerTest` for example.

oxcafedead:
Fixes spring-projects/spring-integration#3840

* Create a new abstraction `RedisTest` for Testcontainers-based tests
* Move existing common utility methods to the new interface
* Migrate all existing Redis tests to use this new `RedisTest` interface
* Migrate all existing Redis tests to JUnit5
* Make all existing Redis tests confirming to JUnit5 standards, such as
default methods and classes visibility instead of public
* Add a dependency for parametrized tests in JUnit5
* Improve assertions readability across tests, for instance:
assertThat(thing.size()).isEqualTo(2) -> assertThat(thing).hasSize(2)
* Add real assertions to
`RedisLockRegistryTests.twoRedisLockRegistryTest` (earlier it did not
have assertions at all)
* Reformat, rearrange and cleanup the code


oxcafedead:
Sorry for many changes to review, but I decided that is would be better that 'chunk-based' approach.
Just for the note, locally I've run `:spring-integration-redis:testAll` and `:spring-integration-file:testAll` to verify the migration.

spring-integration-redis/src/test/java/org/springframework/integration/redis/RedisTest.java
```diff
@@ -33,36 +38,64 @@
...
+ * @author Artem Vozhdayenko
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
  *
+ * @since 6.0
  */
-public abstract class RedisAvailableTests {
+@Testcontainers(disabledWithoutDocker = true)
+public interface RedisTest {
```
artembilan, line 59: You know I call them like this: `MySqlContainerTest`, `MosquittoContainerTest`, `LocalstackContainerTest` etc...

What was the reason to avoid a `Container` indication for this interface?

spring-integration-redis/src/test/java/org/springframework/integration/redis/RedisTest.java
```diff
@@ -33,36 +38,64 @@
 import org.springframework.integration.test.util.TestUtils;
 import org.springframework.messaging.Message;
 
+import io.lettuce.core.ClientOptions;
+import io.lettuce.core.SocketOptions;
+
 /**
+ * The base contract for all tests requiring a Redis connection.
+ * The Testcontainers 'reuse' option must be disabled,so, Ryuk container is started
```
artembilan, line 46: Typo in the Javadocs: space before `so` 😄 

I know, copy/paste from other place, but still let's fix at least here since we have a chance during review 😉

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -516,19 +509,19 @@ public void concurrentObtainCapacityTest() throws InterruptedException {
 
 
 		registry.expireUnusedOlderThan(-1000);
-		assertThat(TestUtils.getPropertyValue(registry, "locks", Map.class).size()).isEqualTo(0);
+		assertThat(TestUtils.getPropertyValue(registry, "locks", Map.class).size()).isZero();
```
artembilan, line 512: `assertThat(TestUtils.getPropertyValue(registry, "locks", Map.class)).isEmpty();` ? 😃

artembilan (changes requested):
Just a couple nit-picks.

Otherwise this is great work, @oxcafedead , as usual!

spring-integration-redis/src/test/java/org/springframework/integration/redis/RedisTest.java
oxcafedead: No specific reason except my inattentiveness 😄

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
oxcafedead: Yeah, I initially fixed such places with `isEmpty` but in this test class it gives compilation warning about unchecked assignments, so I decided that leaving one more `@SuppressWarnings` will not be worth it.
Should I add it anyway?

oxcafedead: Woah, just found that the class has already 
`org.springframework.integration.redis.util.RedisLockRegistryTests#getRedisLockRegistryLocks`
defined.
Will reuse it !

artembilan (approved):
LGTM.

Pulling locally for final verification and merging...

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
oxcafedead: Finally fixed!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/1c461a3e2c51604f0dc4f9ad26e88925152af2f3 after some minor compatibility fix.

@oxcafedead ,

thank you again.

PR #3856, opened 2022-07-21
Adds an example of propagation

marcingrzejszczak:


artembilan (commented):
Thanks, @marcingrzejszczak , for education!

I probably  can merge it and we will continue from here.

My idea for the observation logic is like this:
to implement send/receive/handle/produce in the respective places. mostly where we have at the moment metrics - `AbstractMessageChannel`, `AbstractMessageHandler`, `SourcePollingChannelAdapter`, `MessageProducerSupport`.

My bigger problem at the moment that Micrometer is an `optional` dependency, so I cannot use all the API you mention directly in my code. I have to implement some kind of Facade. Or is that not supposed to be optional from now on?

Thanks

PR #3865, opened 2022-07-27
Fix deprecations around ListenableFuture

artembilan:
SF has deprecated a `ListenableFuture` and API around it

* Migrate to `CompletableFuture` everywhere a `ListenableFuture` has been used
* Suppress a deprecation for `ListenableFuture` keeping the functionality until the next version
* Resolve deprecations nad removals from the latest Spring for Apache Kafka
* Fix documentation for the `ListenableFuture` in favor of `CompletableFuture`

NOTE: the AMQP module is left as is until `ListenableFuture` deprecation is resolved in Spring AMQP

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/gateway/GatewayInterfaceTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/gateway/GatewayInterfaceTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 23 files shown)


spring-integration-core/src/test/java/org/springframework/integration/gateway/GatewayInterfaceTests.java
```diff
@@ -683,7 +671,7 @@ public interface ExecGateway {
 		Future<Thread> test1(Thread caller);
 
 		@Gateway(requestChannel = "gatewayThreadChannel")
-		ListenableFuture<Thread> test2(Thread caller);
+		CompletableFuture<Thread> test2(Thread caller);
```
garyrussell, line 674: Shouldn't we leave some of these tests as Listenable, to make sure they're not broken?

garyrussell (commented):
Just one comment.

garyrussell (approved):

PR #3868, opened 2022-08-03
Use spring-asciidoctor-backends

Buzzardo:
rather than spring-doc-resources, which is to be retired.

### DIFF
--- a/gradle/docs.gradle
+++ b/gradle/docs.gradle
@@ -1,31 +1,17 @@
 ext {
-	docResourcesVersion = '0.2.5'
-	blockSwitchVersion = '0.5.0'
+	backendVersion = '0.0.3'
 }
 
 configurations {
-	docs
-	asciidoctorExt
+	asciidoctorExtensions
 }
 
 dependencies {
-	docs "io.spring.docresources:spring-doc-resources:${docResourcesVersion}@zip"
-	asciidoctorExt "io.spring.asciidoctor:spring-asciidoctor-extensions-block-switch:$blockSwitchVersion"
-}
-
-task prepareAsciidocBuild(type: Sync) {
-	dependsOn configurations.docs
-	from {
-		configurations.docs.collect { zipTree(it) }
-	}
-	duplicatesStrategy = DuplicatesStrategy.INCLUDE
-	from 'src/reference/asciidoc/'
-	into "$buildDir/asciidoc"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -5,6 +5,7 @@ buildscript {
         mavenCentral()
         gradlePluginPortal()
         maven { url 'https://repo.spring.io/plugins-release-local' }
+        maven { url "https://repo.spring.io/milestone" }
```
artembilan, line 8: I wonder why do we need this since I don't see you have added any `M` dependency. 
You have just removed two deps and added one with `backendVersion = '0.0.3'`.
So, why do we need `milestone` repo then?

gradle/docs.gradle
```diff
@@ -119,9 +104,12 @@ asciidoctor {
 		jvmArgs '--add-opens', 'java.base/sun.nio.ch=ALL-UNNAMED', '--add-opens', 'java.base/java.io=ALL-UNNAMED'
 	}
 	baseDirFollowsSourceFile()
-	configurations 'asciidoctorExt'
-	sourceDir "$buildDir/asciidoc"
+	configurations 'asciidoctorExtensions'
+	sourceDir "src/reference/asciidoc/"
 	inputs.dir(sourceDir)
+	outputOptions {
+		backends "spring-html"
```
artembilan, line 111: Just curious if you have build the doc for the project and checked if it looks as expected in both HTML and PDF.
Plus HTML-single.

artembilan (changes requested):
Just a couple questions.

Thank you!

build.gradle
Buzzardo: We don't. I got carried away with copying code. I'll remove it.

gradle/docs.gradle
Buzzardo: I did, but, on checking again after you asked, I found two places that need work. I'll look into them. Thanks.

Buzzardo:
Update: I have resolved part of it (a fencing problem around callouts), but I have a question for the Boot team. I created another commit to be able to show it to them. Once that is resolved, I'll put together a final commit.

gradle/docs.gradle
Buzzardo: It's all fixed now. Thanks, Artem.

Buzzardo:
It should work now. Please review again. Thanks.

PR #3874, opened 2022-08-12
Add documentation for a new MQTT shared client feature

oxcafedead:
Add an overview with reason for the feature as well as basic
capabilities listing. Give an example with Java DSL usage for several
adapters.

### DIFF
--- a/src/reference/asciidoc/mqtt.adoc
+++ b/src/reference/asciidoc/mqtt.adoc
@@ -505,10 +505,10 @@ Otherwise, only the manual restart of `Mqttv5PahoMessageDrivenChannelAdapter` ca
 [[mqtt-shared-client]]
 === Shared MQTT Client Support
 
-If you need to use a single MQTT ClientID for several integrations, you cannot use different MQTT client instances because MQTT brokers may have a limitation on a number of connections per ClientID (typically, a single connection is allowed).
-For having a single client reused for different channel adapters, you may use a `org.springframework.integration.mqtt.core.ClientManager` component and pass it to any channel adapter you need.
+If a single MQTT ClientID is required for several integrations, multiple MQTT client instances cannot be used because MQTT brokers may have a limitation on a number of connections per ClientID (typically, a single connection is allowed).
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Thank you, @oxcafedead , for quick turn around! Would you mind also to add a note into `whats-new.adoc` with a link to this new paragraph? Perhaps in a section for new components .

We have a day off today, so I’ll come back with review on Monday

src/reference/asciidoc/mqtt.adoc
```diff
@@ -500,4 +500,65 @@ IMPORTANT: The `org.springframework.integration.mqtt.support.MqttMessageConverte
...
+public IntegrationFlow mqttOutFlow(
+        ClientManager<IMqttAsyncClient, MqttConnectionOptions> clientManager) {
+
+    return f -> f.handle(new Mqttv5PahoMessageHandler(clientManager));
+}
+----
+====
+
+IMPORTANT: You should use a single client manager instance so that the connection is reused.
+Ensure you do not create it for each channel adapter.
```
artembilan, line 564: I don't think we need this `IMPORTANT`. Not sure how it is important at all since client per channel adapter is still a valid use-case.
Please, consider to remove this altogether.

A couple notes about technical docs:

1. It is better to express the matter in the impersonal manner. In other words we must not say `you`, `we`, `they` or so. I mean not `you cannot use different`, but `the same MQTT client id cannot be used`. 
2. `Foo` and `bar` are not professional words and are considered as a historical mistake for software engineering practice. So, better to call them `topic1` and `topic2`.


Both of these are optional at the moment and feel free to pass such a clean up task up to me.
Just let me know your thoughts!

artembilan (changes requested):
Nothing critical.

Thank you!

src/reference/asciidoc/mqtt.adoc
oxcafedead: These all make sense, I have changed it.

PR #3876, opened 2022-08-17
GH-3875: Initialize ClassUtils as early as possible

### LINKED ISSUE #3875: NoClassDefFoundError for org.springframework.integration.util.ClassUtils if first called from JMX
* Spring Integration - 5.5.11
* Spring Framework - 5.3.19

**Describe the bug**

We have a Spring Integration setup in an application, and we have some healthchecks that exercise this and are triggered via JMX. In general this works fine, but in some cases we instead get `java.lang.NoClassDefFoundError: Could not initialize class org.springframework.integration.util.ClassUtils` errors.

After much digging, it turns out that this is because:
* `org.springframework.integration.util.ClassUtils` is entirely static. 
  * This means that it is first initialized when some code referencing it is executed, and it is initialized in the thread of the code that referenced it.
* JMX runs on a thread in the Tomcat server, using the Tomcat common classloader.

This all means that, if the first time `org.springframework.integration.util.ClassUtils` is referenced is by code executing within an MBean, then it is running in the Tomcat common classloader and not in the webapp classloader. This in turn means that it is unable to see other classes that it expects to see - for example, [`org.springframework.integration.core.GenericSelector`](https://github.com/spring-projects/spring-integration/blob/main/spring-integration-core/src/main/java/org/springframework/integration/util/ClassUtils.java#L108)

**To Reproduce**

I'll see if I can put together a minimal reproducing project, but essentially it will be:
* Build an app using Spring Integration
* Write an MBean that triggers Spring Integration
* Call the MBean before anything in the app is called

**Expected behavior**

This would all still work correctly and not fail.

**Sample**

Coming soon.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3875

The class `static` variables and initialization block are performed on first class access.

* Add "fake" `ClassUtils.resolvePrimitiveType(Integer.class)` call to the `IntegrationRegistrar`
to trigger `ClassUtils` initialization with the current Spring `ClassLoader`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2021 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


garyrussell:
...and cherry-picked as b7710fcd4d

PR #3877, opened 2022-08-21
GH-3844: Rework messaging annotation with @Bean

### LINKED ISSUE #3844: A BeanPostProcessor should not create bean definitions
Right now, `MessagingBeanPostProcessor` is creating bean definitions. In order to do that, it has to inject the `BeanFactory` and downcast it to a `BeanRegistry`, see https://github.com/spring-projects/spring-integration/blob/7f631fc947c1b4c5f284b887aa144ba4c100486c/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/MessagingAnnotationPostProcessor.java#L233

As a side effect of this, we might be able to get rid of [that `@Bean` lookup](https://github.com/spring-projects/spring-integration/blob/7f631fc947c1b4c5f284b887aa144ba4c100486c/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java#L151). `@Bean` is a configuration class parsing concern and shouldn't be relied upon outside of it.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3844

* Make `MessagingAnnotationPostProcessor` as a `BeanDefinitionRegistryPostProcessor`
to process bean definitions as early as possible and register respective messaging
components at that early phase
* Make bean definitions parsing logic optional for AOT and native mode since beans
have bean parsed during AOT building phase
* Introduce a `BeanDefinitionPropertiesMapper` for easier mapping
of the annotation attributes to the target `BeanDefinition`
* Remove `@Bean`-related logic from method parsing process
* Change the logic for `@Bean`-based endpoint bean names:
since we don't deal with methods on the bean definition phase, then method name
does not make sense.
It even may mislead if we `@Bean` name is based on a method by default, so we end up
with duplicated word in the target endpoint bean name.
Now we don't
* Fix `configuration.adoc` respectively for a new endpoint bean name logic
* In the end the new logic in the `AbstractMethodAnnotationPostProcessor`
is similar to XML parsers: we feed annotation attributes to the
`AbstractStandardMessageHandlerFactoryBean` impls

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
@@ -241,7 +241,9 @@ public abstract class AbstractMethodAnnotationPostProcessor<T extends Annotation
 
 		Poller poller = MessagingAnnotationUtils.resolveAttribute(annotations, "poller", Poller.class);
 		Reactive reactive = MessagingAnnotationUtils.resolveAttribute(annotations, "reactive", Reactive.class);
-		Assert.state(reactive == null || poller == null, "The 'poller' and 'reactive' are mutually exclusive.");
+		Assert.state(reactive == null || poller == null,
+				() -> "The 'poller' and 'reactive' attributes are mutually exclusive." +
+						"The bean definition with the problem is: " + beanName);
 		if (poller != null) {
 			applyPollerForEndpoint(endpointBeanDefinition, poller);
 		}
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
@@ -20,43 +20,63 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.aopalliance.aop.Advice;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.reactivestreams.Publisher;
 
-import org.springframework.aop.TargetSource;
 import org.springframework.aop.framework.Advised;
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.aop.support.DefaultBeanFactoryPointcutAdvisor;
 import org.springframework.aop.support.NameMatchMethodPointcut;
 import org.springframework.aop.support.NameMatchMethodPointcutAdvisor;
+import org.springframework.beans.BeanMetadataElement;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.BeanFactoryAware;
+import org.springframework.beans.factory.FactoryBean;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.beans.factory.annotation.AnnotatedBeanDefinition;
+import org.springframework.beans.factory.config.BeanDefinition;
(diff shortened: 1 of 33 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -106,67 +125,298 @@
...
+				createEndpointBeanDefinition(new BeanComponentDefinition(handlerBeanDefinition, handlerBeanName),
+						new BeanComponentDefinition(beanDefinition, beanName), annotations);
+
+		new BeanDefinitionPropertiesMapper(endpointBeanDefinition, annotations)
+				.setPropertyValue("autoStartup")
+				.setPropertyValue("phase");
+
+		Poller poller = MessagingAnnotationUtils.resolveAttribute(annotations, "poller", Poller.class);
+		Reactive reactive = MessagingAnnotationUtils.resolveAttribute(annotations, "reactive", Reactive.class);
+		Assert.state(reactive == null || poller == null, "The 'poller' and 'reactive' are mutually exclusive.");
```
garyrussell, line 244: Either `"'poller' and 'reactive' are mutually exclusive."` or `"The 'poller' and 'reactive' attributes are mutually exclusive."` Also should include the bean name.

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -106,67 +125,298 @@
...
+							.setPropertyValue(SEND_TIMEOUT_ATTRIBUTE)
+							.setPropertyReference("outputChannel")
+							.setPropertyReference("defaultOutputChannel");
+				}
+
+				if (isClassIn(handlerBeanClass, AbstractMessageProducingHandler.class,
+						AbstractSimpleMessageHandlerFactoryBean.class)
+						&& !RouterFactoryBean.class.isAssignableFrom(handlerBeanClass)) {
+
+					applyAdviceChainIfAny(handlerBeanDefinition, annotations);
```
garyrussell, line 220: Why are we doing this here - this will add all the advices to the handler - the CEFB (which also gets the full advice chain) selectively applies advices to the handler.

src/reference/asciidoc/configuration.adoc
```diff
@@ -676,12 +676,15 @@ NOTE: The bean names are generated with the following algorithm:
 
 * The `MessageHandler` (`MessageSource`) `@Bean` gets its own standard name from the method name or `name` attribute on the `@Bean`.
 This works as though there were no messaging annotation on the `@Bean` method.
-* The `AbstractEndpoint` bean name is generated with the following pattern: `[configurationComponentName].[methodName].[decapitalizedAnnotationClassShortName]`.
-For example, the `SourcePollingChannelAdapter` endpoint for the `consoleSource()` definition <<annotations_on_beans,shown earlier>> gets a bean name of `myFlowConfiguration.consoleSource.inboundChannelAdapter`.
+* The `AbstractEndpoint` bean name is generated with the following pattern: `[@Bean name].[decapitalizedAnnotationClassShortName]`.
+For example, the `SourcePollingChannelAdapter` endpoint for the `consoleSource()` definition <<annotations_on_beans,shown earlier>> gets a bean name of `consoleSource.inboundChannelAdapter`.
+Unlike with POJO methods the bean method name is not included into an endpoint bean name.
```
garyrussell, line 681:
```suggestion
Unlike with POJO methods, the bean method name is not included in the endpoint bean name.
```

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/annotation/AbstractMethodAnnotationPostProcessor.java
artembilan: See the comment exactly in that `ConsumerEndpointFactoryBean.adviceChain()`:
```
 *  ARPMHs advise the handleRequestMessage method internally and already have the advice chain injected.
*  So we only advise handlers that are not reply-producing.
```

so, we don't set an `adviceChain` to the `AbstractReplyProducingMessageHandler`, neither from XML parser, nor Java DSL - `ConsumerEndpointSpec.doGet()`.

We may revise it though, but I guess there was a reason why you implemented that way in the XML parser day first.
Different issue either way.

garyrussell (approved):

PR #3879, opened 2022-09-02
Add some infrastructure for Observation

artembilan:
* Populate an `ObservationRegistry` bean from the `IntegrationManagementConfigurer`
into all the `IntegrationManagement` components
* Introduce `MessageReceiverContext` and `MessageSenderContext` for easier usage
in the target code
* Implement `Observation` handling in the `AbstractMessageHandler`
* Modify `ObservationPropagationChannelInterceptorTests` for new `MessageSenderContext`
* Use `BridgeHandler` to ensure that `Observation` is propagated and handled properly
* Verify that tags from the `AbstractMessageHandler` are preset on the consumer span


spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageHandler.java
```diff
@@ -40,18 +43,31 @@ public abstract class AbstractMessageHandler extends MessageHandlerSupport
...
+			}
+			else {
+				doHandleMessage(message);
+			}
 		}
 	}
 
+	private void handleWithObservation(Message<?> message, ObservationRegistry observationRegistry) {
+		Observation.createNotStarted(CONSUME_OBSERVATION_NAME, new MessageReceiverContext(message), observationRegistry)
+				.lowCardinalityKeyValue("type", "handler")
```
marcingrzejszczak, line 66: We suggest moving this out to use `DocumentedObservation` together with `ObservationConvention`. That way you give users an option to control the key values, naming etc. + you can get the documentation generated automatically

marcingrzejszczak (approved):
LGTM! I've added some small comments.

Things to consider / test

- Use `SampleTestRunner` to see how the spans will look like in Zipkin / Wavefront
- If we have multiple channels via which we are pushing a message, do we want each passing via a channel to be a separate span? For example, if we get a message from rabbit, then we go through 10 different other processing via different channels, and then we push the message out of process to kafka, then should we 12 spans, or 1 span for input, 1 for processing and 1 for output ?

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfiguration.java
```diff
@@ -64,12 +66,16 @@ public void setImportMetadata(AnnotationMetadata importMetadata) {
 
 	@Bean(name = IntegrationManagementConfigurer.MANAGEMENT_CONFIGURER_NAME)
 	@Role(BeanDefinition.ROLE_INFRASTRUCTURE)
-	public IntegrationManagementConfigurer managementConfigurer(ObjectProvider<MetricsCaptor> metricsCaptorProvider) {
+	public IntegrationManagementConfigurer managementConfigurer(
+			ObjectProvider<MetricsCaptor> metricsCaptorProvider,
+			ObjectProvider<ObservationRegistry> observationRegistryProvider) {
```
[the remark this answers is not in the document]
artembilan: Yes, that's what we have discussed with you a couple weeks ago. 
The conclusion was like: since a `micrometer-observation` is tiny API jar there is no reason in extra facade layer complexity on our side.
The bean by itself is `optional` though, so we just don't have any observation by default either way.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageHandler.java
artembilan: Sounds interesting, but we definitely need to come up with that convention for end-users 😄 .

I think these couple keys I have here so far are mandatory and really specific to component.
I really design an `Observation` usage similar to what we have so far with `Timer` here or there.

The `DocumentedObservation` looks like - will investigate that one.

Thank you!

artembilan:
I'm not familiar with  Zipkin / Wavefront so either view won't make too much sense for me 😄 

Will look into that when I come up with some observation for message channels.

And answering to your last question: indeed there have to be 12 spans (at least). I plan to observe message channels as well - for now just it is place where we handle timers: `AbstractMessageChannel` and `MessagingGatewaySupport` (not talking about `MessageHandlerSupport` which is handled by this PR).

Thank you for review! 
Will go to learn a `DocumentedObservation`...

artembilan:
Pushed some `DocumentedObservation` API impls.

spring-integration-core/src/main/java/org/springframework/integration/support/management/observation/IntegrationObservations.java
```diff
@@ -0,0 +1,53 @@
...
+import io.micrometer.observation.ObservationRegistry;
+
+/**
+ * A convenient factory for {@link IntegrationObservation} instances.
+ *
+ * @author Artem Bilan
+ *
+ * @since 6.0
+ */
+public final class IntegrationObservations {
```
artembilan, line 31: As you see I had to create this factory for better target usage.
I really don't see reason in the `DefaultMessageReceiverObservationConvention` instance per request.
And I don't feel yet that I need some `customConvention`.

It probably would be great if `getDefaultConvention()` would return an instance instead of `Class` and this one would always be used in case of plain `observation(registry, context)` call.

@marcingrzejszczak , you mentioned in [Spring AMQP PR](https://github.com/spring-projects/spring-amqp/pull/1500#discussion_r965296869) that we always have to use a convention for contextual name propagation, but as you see it is a bit involved in the target code since we have to call the `observation()` method with much more arguments, than just rely on a default convention by default...

rishiraj88:
👍

spring-integration-core/src/main/java/org/springframework/integration/support/management/observation/IntegrationObservations.java
marcingrzejszczak: >  I really don't see reason in the DefaultMessageReceiverObservationConvention instance per request.

We typically create singleton fields for those in instrumentations.

>  And I don't feel yet that I need some customConvention.

You don't, but the users might ;) What if they want to follow their internal naming convention and you don't give them such an option?

> It probably would be great if getDefaultConvention() would return an instance instead of Class and this one would always be used in case of plain observation(registry, context) call.

That will be harder to parse for the micrometer docs generator. 

> @marcingrzejszczak , you mentioned in https://github.com/spring-projects/spring-amqp/pull/1500#discussion_r965296869 that we always have to use a convention for contextual name propagation, but as you see it is a bit involved in the target code since we have to call the observation() method with much more arguments, than just rely on a default convention by default...

Yes, I mentioned that it would be preferable to use the `ObservationConvention` to name observations either via contextual or non-contextual methods. 

The way you did it ATM doesn't allow to set any customizations by the users at the level of Spring Integration. Those can be still done by `GlobalObservationConvention` and the `ObservationRegistry` (which may be OK if you're OK with that). That means that if I want to change the tags or naming for Spring Integration then I can ONLY do it by registering a proper `GlobalObservationConvention` and attaching it to the `ObservationRegistry`. The other option would be to give the user a setter to provide a custom implementation of your `MessageRecevierObservationConvention`. It's up to you to decide (e.g. Reactor Netty does not allow any customizations).

spring-integration-core/src/main/java/org/springframework/integration/support/management/observation/IntegrationObservation.java
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	enum HandlerTags implements KeyName {
+
+		/**
+		 * Name of the message handler component.
+		 */
+		COMPONENT_NAME {
+			@Override
+			public String asString() {
+				return "name";
```
marcingrzejszczak, line 63: I wonder if the full `spring.integration.handler.name` wouldn't be better (same for type). Imagine that you're in zipkin and you're searching for a tag `name`. You can have a lot of different names but you're looking specifically for a handler name. WDYT ?

spring-integration-core/src/main/java/org/springframework/integration/support/management/observation/IntegrationObservation.java
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	enum HandlerTags implements KeyName {
+
+		/**
+		 * Name of the message handler component.
+		 */
+		COMPONENT_NAME {
+			@Override
+			public String asString() {
+				return "spring.integration.name";
```
marcingrzejszczak, line 68: Shouldn't this be `spring.integration.handler.name` ? And below, `spring.integration.handler.type` ?

spring-integration-core/src/test/java/org/springframework/integration/channel/interceptor/ObservationPropagationChannelInterceptorTests.java
```diff
@@ -199,49 +207,55 @@ void observationPropagatedOverQueueChannel() throws InterruptedException {
...
+
+		assertThat(this.meterRegistry.get("spring.integration.handler")
+				.tag("spring.integration.name", "testBridge")
+				.tag("spring.integration.type", "handler")
+				.tag("error", "none")
+				.timer().count()).isEqualTo(1);
+	}
+
+	@SuppressWarnings("unchecked")
+	private static SpansAssert assertSpans(Collection<? extends FinishedSpan> actual) {
```
marcingrzejszczak, line 257: Have you filed a PR / issue about this in micrometer-tracing?

spring-integration-core/src/main/java/org/springframework/integration/support/management/observation/IntegrationObservation.java
artembilan: No, I don't think so.
It would sound a bit awkward: `spring.integration.handler.type=handler`.

The idea is to have a common tag on integration meters, so we can easy search them.
There are going to be also: `spring.integration.type=source` and `spring.integration.type=channel`.
The `spring.integration.name` really belongs to the bean name of that integration component.

Not sure why would you request me to do that fine-graining...

spring-integration-core/src/test/java/org/springframework/integration/channel/interceptor/ObservationPropagationChannelInterceptorTests.java
artembilan: No, I did not yet.
Will look into that today if time permits.

artembilan: Here you are: https://github.com/micrometer-metrics/tracing/pull/65 !

marcingrzejszczak: Merged, thanks!

spring-integration-core/src/main/java/org/springframework/integration/support/management/observation/IntegrationObservation.java
marcingrzejszczak: Ah, ok, I misunderstood what the type is. I thought that you would have `spring.integration.handler.type=some handler type`. Disregard my comment.

marcingrzejszczak:
It still would be super beneficial to use the `SampleTestRunner` with Zipkin to check how the whole trace looks like - https://micrometer.io/docs/tracing#_running_integration_tests

artembilan:
@marcingrzejszczak ,

does this look like you'd expect?

image

Sorry, I'm not familiar with Zipkin and what it could show me 😄

spring-integration-core/src/test/java/org/springframework/integration/support/management/observation/IntegrationObservabilityZipkinTests.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * @author Artem Bilan
+ *
+ * @since 6.0
+ */
+public class IntegrationObservabilityZipkinTests extends SampleTestRunner {
+
+	@Override
+	public TracingSetup[] getTracingSetup() {
+		return new TracingSetup[]{ TracingSetup.IN_MEMORY_BRAVE, TracingSetup.ZIPKIN_BRAVE };
```
marcingrzejszczak, line 56: Why are you leaving out OTel ? BTW the Wavefront setup will be automatically ignored cause you didn't provide any tokens.

Do you have a screenshot from Zipkin? I'm super interested in how it looks like :)

marcingrzejszczak:
> does this look like you'd expect?

You tell me :D So by looking at the graph I see that via Spring Integration we've managed to send a message from one channel and receive it on the other one. It took close to no time to process it. 

If you were to debug latency issues or production bugs - would this information be useful? Would you need more / less (would you add more tags) ?

spring-integration-core/src/test/java/org/springframework/integration/support/management/observation/IntegrationObservabilityZipkinTests.java
artembilan: Why should I care about OTel?
It is really not Spring Integration responsibility to check the target tracing protocol.
I have even excluded OTel and Wavefront dependencies: it is not my project responsibility to test Observation low-level libraries.
I think I just have to believe to your asserts and that's it 😄

artembilan:
> we've managed to send a message from one channel and receive it on the other one.

Well, it's a bit complicated than you think and it is not finished from the framework perspective yet.

So, I wrap sending into a start observation - this part will be implemented eventually in the `MessageChannel.send()`. (Was a bit surprised though to see extra parent span: the sample runner does that for us 😄 )
Then a bit propagation magic happens. My channel under test is a `QueueChannel` where consumption from it happens on another, scheduled thread. I transfer an `Observation` via specific message property.
On that consumer thread I restore the `Observation` into current scope. See my `ObservationPropagationChannelInterceptor`.
The this message is handled by the `BridgeHandler` and we see a span for it contributed by the infrastructure introduced in this PR.

We indeed see "no time to process" because there is really no latency: producer just place a message into a queue and close its observation.
The consumer happens on another thread and just replies that message for assertion.

I don't have a production experience with tracing to judge how is this useful.
However I'd like to see some visual connection between `producer` and `consumer` kinds...

Thank you for looking into this anyway!

spring-integration-core/src/test/java/org/springframework/integration/support/management/observation/IntegrationObservabilityZipkinTests.java
marcingrzejszczak: > Why should I care about OTel?

Because we're providing 2 tracers for this `SampleTestRunner` and we want to be sure that using both of them things are looking good.

>  It is really not Spring Integration responsibility to check the target tracing protocol.

You can also just write unit tests with mocks - it's all about being more certain that things are working fine

>  I have even excluded OTel and Wavefront dependencies: it is not my project responsibility to test Observation low-level libraries.

`SampleTestRunner` is about giving you more confidence that things are looking good with tracing visualization tools. If I was a Spring Integration project maintainer then I would like to be pretty sure that things are looking good regardless of the used tracer. Also I would like to be sure that things are looking good in Wavefront (have you checked that?)

artembilan: Well, this sounds like testing a `RestTemplate` against different HTTP client impls, which is really not my project responsibility where I should worry proper `RestTemplate` calls from my code. How `RestTemplate` interacts with a provided HTTP client is really outside of my project scope.
OTel, Wavefront, Zipkin - all of these are new for me and I need to read about them more.
Meanwhile I still don't see what and how I would fix in  my code if something is not OK with those techs. Just because I don't do anything from my code to affect it somehow.
Doesn't look like it is similar to RDBMS where each vendor may have its own specific SQL extensions we should care about.

If you still insist on enabling them, I'll revert my restrictions, but, please, be my guest to help me to determine "that things are looking good" 😄

artembilan:
> SHOW ALL ANNOTATIONS

Yep! Saw them. I worry more about if I'm going to have many producer and consumer spans in my trace. It seems there has to be a parent-child connection between them. 
Now imaging I use publish-subscribe. So, one producer span an many consumer spans. That might be great to see a connection between them. Perhaps not a parent-child, but like a `source span` since with an async handling we may stop the parent far before those other threads handles their own spans...

marcingrzejszczak:
>  It seems there has to be a parent-child connection between them.

There is a parent child relationship because the spans are nested below each other. Look to the left that you can fold the child spans of `test send`. In this case you have only 1 so there's not much to fold :)

marcingrzejszczak:
Yes, looks good to me!

garyrussell (approved):

PR #3880, opened 2022-09-02
Rename CONDITION column to GROUP_CONDITION

artembilan:
The `CONDITION` word is reserved one in some RDBMS vendors

* Rename the `CONDITION` column to `GROUP_CONDITION` to avoid possible failures
and get rid of the conditional code on the `JdbcMessageStore`


artembilan:
I will add a Migration Guide note after merging.

garyrussell (approved):

PR #3881, opened 2022-09-02
Rework infra logic onto IntegrationProperties

artembilan:
The `Properties` bean has been deprecated since `5.5`

* Register a default `integrationGlobalProperties` as a `IntegrationProperties`
* Fix respective usages from the `Properties`


garyrussell (changes requested):
Just one nit.

PR #3883, opened 2022-09-07
GH-3660: Invoke well-known lambdas explicitly

### LINKED ISSUE #3660: Revise LambdaMessageProcessor to call well-known lambdas directly, not via reflection
Right now it does `ReflectionUtils.makeAccessible(this.method);` and then `Object result = this.method.invoke(this.target, args);`, but probably we can call `Function.apply()`, `GenericHandler.handle()`, `GenericSelector.accept()`, etc. directly since we know the places `LambdaMessageProcessor` is used, plus we can infer the type of the lambda and delegate to respective strategy.

NOTE: With Java 17 we have to add `--add-opens java.base/java.util.function=ALL-UNNAMED` when `Function` lambda is in use.
Not having `method.setAccessible(true)` in this case does not help: it still fail on `method.invoke()` because cannot get access to the method even if it is `public` by definition of lambda itself.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3660

* Change the `LambdaMessageProcessor` to invoke `Function`, `GenericHandler`, `GenericSelector` & `GenericTransformer` explicitly without reflection. This allows to avoid `--add-opens` for Java types to be used from reflections. Plus this removes some overhead with native images. And in general it is faster than reflection invocation.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 the original author or authors.
+ * Copyright 2016-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


garyrussell:
Your "Fixes" points to the wrong place; should be #3660

artembilan:
Edited and re-pushed.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
```diff
@@ -170,4 +185,41 @@ else if (Map.class.isAssignableFrom(parameterType)) {
...
+			return transformer.transform(args[0]);
+		}
+		else if (this.target instanceof GenericHandler handler) {
+			return handler.handle(args[0], (MessageHeaders) args[1]);
+		}
+		else {
+			return this.method.invoke(this.target, args);
+		}
+
+		/* TODO when preview features are available in the next Java version
```
garyrussell, line 206: This might need a NOSONAR (TODO).

artembilan: I think that's OK: will give us an extra reminder to not forget to fix this eventually 😄

artembilan: Either way our Sonar instance does not support Java 17 yet: https://stackoverflow.com/questions/71528775/migrate-sonarqube-to-support-java-17

Issue #3872, opened 2022-08-08
Embrace queue APIs in Spring's JDBC message stores for polling messages


raphw:
Modern databases support much better queuing APIs than what they offered years ago, so articles [like the one included in the Spring integration](https://mikehadlow.blogspot.com/2012/04/database-as-queue-anti-pattern.html) documentation are rather dated.

For example, [Postgres offers `LISTEN` and `NOTIFY`](https://jdbc.postgresql.org/documentation/81/listennotify.html) or [Oracle database change norification](https://docs.oracle.com/cd/E11882_01/java.112/e16548/dbchgnf.htm#JJDBC28820). These APIs would allow for a more efficient `Queue` usage of the JDBC queues that are offered, if a database offers it.

artembilan:
Yes, we may consider that as a contribution, for example like `spring-integration-postgresql` and `spring-integration-oracle` to implement those vendor-specific features.
It might be not as a store impl, but rather just `PostgresMessageChannel` and perhaps message-driven channel adapter to react to notifications and emit respective data.

[Contribution is welcome](https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc)!

raphw:
Thanks, I will look into it and try to solve this for my own application first and then try to extract it to a pull request.

raphw:
I played around with this on Oracle and Postgres and here are some observations:

1. The Oracle JDBC driver is rather buggy, on its newest version, the driver fails with an internal assertion error upon trying to register a change notification. With an older version, I get it to work nicely, but only if the Oracle database runs on the same host as the JVM. There is a property `NTF_LOCAL_HOST` to define a host name that the Oracle DB should send notifications to, but it is [falsely submitted by the JDBC driver](https://gist.github.com/raphw/dc7f9bef0786d64a8cb96091abc6c3d7). I got around this as well by registering the change notification in a PL/SQL script that I run via JDBC from which I call back via TCP. It requires some effort to run this setup and I found it rather limiting as the Oracle DB must be able to call the JVM. In most setups with Kubernetes, it is for example not possible to callback a specific replica. Also, often it requires to change firewall setup.

2. The official Postgres JDBC driver [only supports blocking polls](https://jdbc.postgresql.org/documentation/head/listennotify.html) but I assume that would not be an issue as that represents the queue API. However, it consumes a full connection.

[...]

PR #3884, opened 2022-09-08
Adds a listener for Postgres-based channel message stores.

raphw:
Addresses #3872.

This listener allows for push notifications upon the arrival of new messages for a given group id and region.


raphw:
I added this as a draft for creating a better environment for discussing this. I addressed your last comments, however not the point about the `DataSource` as it will most likely not work. If one used

```java
dataSource.getConnection().unwrap(PgConnection.class)
```

for getting a connection, the `DataSource` would likely be a pooled one. These pools normally expect a connection to return to the pool after a given timeout, otherwise, they close the physical connection and report a connection leak. This would abort the listener, and repeat the process as a new connection was checked out. Therefore, a connection must normally be read directly from the driver, often via `Driver.getConnection`, that's why the interface is added.

artembilan:
>  the DataSource would likely be a pooled one

So, you mean that the logic like this `dataSource.getConnection().unwrap(PgConnection.class)` is not not always correct because we cannot be sure what type of `DataSource` is provided.
If that is a case, then we must not tempt the fate and avoid problems as much as possible.
Therefore and approach with a `Supplier<PgConnection>` is the way to go - and we leave a choice up to end-user. And here already we can show a sample as this `dataSource.getConnection().unwrap(PgConnection.class)`.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
```diff
@@ -0,0 +1,48 @@
+package org.springframework.integration.jdbc.channel;
+
+import org.springframework.integration.jdbc.store.JdbcChannelMessageStore;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHandler;
+
+public final class PostgresChannelMessageTableSubscription {
```
artembilan, line 7: I don't think that I can accept this abstraction:
1. It is probably better to have a `JdbcChannelMessageStore` as a ctor for the subscriber.
2. Have getter for it over there as well
3. The Postgres message channel should expect the subscriber as its ctor and take that store from there. 
This way we would avoid the problem when subscriber and store look into different DBs or tables.
4. The subscription should be an inner interface of the subscriber to avoid classes tangle.
5. This way a `region` would be just an internal of the subscriber since both subscriber and channel would rely on the same store, therefore on the same region.
6. I don't believe in external `Runnable onPossibleUpdate`: it is really has to be an `onUpdate()` impl details. That's why I'm asking about a `SubscribableChannel` impl together with the mentioned subscription contract.
7. Somehow "subscriber" and "subscription" words don't work for me here: they are clashing with Reactive Streams. Not sure yet in other names though...

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PgConnectionSupplier.java
```diff
@@ -0,0 +1,11 @@
+package org.springframework.integration.jdbc.channel;
+
+import org.postgresql.jdbc.PgConnection;
+
+import java.sql.SQLException;
+
+@FunctionalInterface
+public interface PgConnectionSupplier {
+
+	PgConnection get() throws SQLException;
```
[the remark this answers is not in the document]
raphw: My thought was to make it lambda-friendly. Most people would likely just wrap the exception and rethrow it, I guess.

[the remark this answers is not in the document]
artembilan: OK. Let it be! We an always change it afterwards if that.

raphw:
> > the DataSource would likely be a pooled one
> 
> So, you mean that the logic like this `dataSource.getConnection().unwrap(PgConnection.class)` is not not always correct because we cannot be sure what type of `DataSource` is provided. If that is a case, then we must not tempt the fate and avoid problems as much as possible. Therefore and approach with a `Supplier<PgConnection>` is the way to go - and we leave a choice up to end-user. And here already we can show a sample as this `dataSource.getConnection().unwrap(PgConnection.class)`.

For example Spring Boot would always auto-configure a data source where you could not use it without constant reconnects. And I have rarely seen unpooled data sources being used. I think that's the right approach.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
raphw: Not sure if I follow, you suggest to make the JdbcChannelMessageStore a constructor argument to the subscriber? Then it would no longer be possible to reuse the subscriber for different regions. This would require multiple connections to block for each region, would it not? Ideally, I wanted to avoid this.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class PostgresChannelMessageTableSubscriber implements SmartLifecycle {
+
+	private final LogAccessor logger = new LogAccessor(this.getClass()); // NOSONAR final
```
artembilan, line 22: We don't need `NOSONAR` here since the property is `private`.
We use it in those case where it is `protected`.
Also this one can simply be `static`: I doubt any one is going to extend this class.

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-postgresql.sql
```diff
@@ -58,3 +58,19 @@ CREATE TABLE INT_METADATA_STORE  (
 	REGION VARCHAR(100) NOT NULL,
 	constraint INT_METADATA_STORE_PK primary key (METADATA_KEY, REGION)
 );
+
+-- This is only needed if using PostgresChannelMessageSubscriber
+-- CREATE FUNCTION INT_CHANNEL_MESSAGE_NOTIFY_FCT()
+-- RETURNS TRIGGER AS
+-- $BODY$
+-- BEGIN
+--     EXECUTE 'NOTIFY int_channel_message_notify, ''' ||  QUOTE_LITERAL(NEW.GROUP_KEY || NEW.REGION) || '''';
```
artembilan, line 67: Why haven't you yet flipped as `region | groupKey`?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
artembilan: Yeah... Sorry, I forgot that `JdbcChannelMessageStore` is restricted to specific region.
We will have to document it properly (and in JavaDocs as well), that `JdbcChannelMessageStore` on the channel we want to subscribe from and `PgConnectionSupplier` must belong to the same table.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -0,0 +1,161 @@
...
+				Thread t = new Thread(job);
+				t.setDaemon(true);
+				t.setName("postgres-channel-message-table-subscriber");
+				return t;
+			});
+			this.executor = executor;
+		}
+		// Avoid that stop/start sequence reactivates previously stopped thread.
+		Object current = new Object();
+		identity = current;
```
artembilan, line 79: You just avoid this via making those `start()` and `stop()` as `synchronized`.

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
raphw: Indeed, I fixed that.

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-postgresql.sql
raphw: Just an oversight, just fixed it.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
raphw: I made the class final to avoid its subclassing as I agree that there's no point in it.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
raphw: That would not work as the method returns before the executor thread might have ended. If the executor's thread dangles around, it could pick up a subsequent start. By switching the identity, it is certain that this additional listener should no longer listen or produce events, even if it takes the thread some time to complete.

artembilan: We probably need to wait in the `stop()` until that thread is done with its loop before returning from the method. Something like `CountDownLatch`

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-postgresql.sql
```diff
@@ -58,3 +58,19 @@ CREATE TABLE INT_METADATA_STORE  (
 	REGION VARCHAR(100) NOT NULL,
 	constraint INT_METADATA_STORE_PK primary key (METADATA_KEY, REGION)
 );
+
+-- This is only needed if using PostgresChannelMessageSubscriber
+-- CREATE FUNCTION INT_CHANNEL_MESSAGE_NOTIFY_FCT()
+-- RETURNS TRIGGER AS
+-- $BODY$
+-- BEGIN
+--     EXECUTE 'NOTIFY int_channel_message_notify, ''' ||  QUOTE_LITERAL(NEW.REGION || NEW.GROUP_KEY) || '''';
```
artembilan, line 67: Why there is no some readable concatenator between region and group?
What if I have them like this: `myregion` and `mygroup` ?
So, with this I'll end up as `myregionmygroup`. something what is not end-user friendly 😢

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
raphw: Yeah, I think this will need some extensive javadoc, but I wanted to get the API right before I started on writing anything there. Due to the sharing requirement, the API will get slightly more complicated than it otherwise would have been, but I think it's worth that performance-wise.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
raphw: I changed that to use a latch now.

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-postgresql.sql
raphw: I thought that, since it's not exposed to the user, it does not matter much, but it's probably nice for debugging purposes. I added a space now.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -0,0 +1,206 @@
...
+	private final PgConnectionSupplier connectionSupplier;
+
+	private final String tablePrefix;
+
+	@Nullable
+	private ExecutorService executor;
+
+	private CountDownLatch latch = new CountDownLatch(0);
+
+	private Future<?> future = CompletableFuture.completedFuture(null);
```
artembilan, line 54: I don't think we need to have this props initialized, but `volatile` indeed

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -273,6 +273,14 @@ public void setRegion(String region) {
 		this.region = region;
 	}
 
+	/**
+	 * Returns the current region that was set or {@link #DEFAULT_REGION}, which is the default.
+	 * @return the set region name
```
artembilan, line 278: `@since 6.0`

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-postgresql.sql
```diff
@@ -58,3 +58,19 @@ CREATE TABLE INT_METADATA_STORE  (
...
+--     EXECUTE 'NOTIFY int_channel_message_notify, ''' ||  QUOTE_LITERAL(NEW.REGION || ' ' || NEW.GROUP_KEY) || '''';
+--     RETURN NEW;
+-- END;
+-- $BODY$
+-- LANGUAGE PLPGSQL;
+--
+-- CREATE TRIGGER INT_CHANNEL_MESSAGE_NOTIFY_TRG
+-- AFTER INSERT ON INT_CHANNEL_MESSAGE
+-- FOR EACH ROW
+-- EXECUTE PROCEDURE INT_CHANNEL_MESSAGE_NOTIFY_FCT()
```
artembilan, line 76: New line in the end of each file.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
```diff
@@ -0,0 +1,65 @@
...
+
+	public String getRegion() {
+		return this.region;
+	}
+
+	public void onPossibleUpdate() {
+		this.onPossibleUpdate.run();
+	}
+
+	public static PostgresChannelMessageTableSubscription ofMessageStore(
```
artembilan, line 53: I'm still against this solution.
I'd like to see a subscription interface and some `SubscribableChannel` implementation.

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
raphw: I redid it to yield a `SubscribableChannel`. Of course, the implementation could be changed to only allow for this abstraction, but this would take away the possibility to being informed that there are new messages without polling them off the queue.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
```diff
@@ -0,0 +1,165 @@
...
+
+	@Override
+	public int hashCode() {
+		int result = this.groupId.hashCode();
+		result = 31 * result + this.region.hashCode();
+		result = 31 * result + this.onPossibleUpdate.hashCode();
+		return result;
+	}
+
+	public static SubscribableChannel asSubscribableChannel(
```
artembilan, line 81: Why do you still insist on this factory?
This is not Spring-friendly and I doubt anyone is going to use a subscription outside of the channel. 
The `PostgresSubscribableChannel` must be a top-level API and no any extra classes for some `onPossibleUpdate` .
and I doubt in this `onPossibleUpdate()` name yet. Why just not `notify()` ?
What is wrong with extracting contract `Subscription` interface and implement it on the channel?
The `PostgresChannelMessageTableSubscriber` must be a ctor argument for that channel.
Also dispatching must be handled by the `MessageDispatcher` abstraction.

artembilan (changes requested):

artembilan (changes requested):
One more not forgot to say:

No reason to squash your commits all the time: we are going to do that on merge.
Meanwhile, during review and feedback process, it is better to keep commits history.
This way we know what was before we started and concluded some discussion.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
raphw: My intuition is to expose the most primitive abstraction on the base level, which is receiving a notification whenever a message was added to the group. If you wanted to use this information to read a message from a worker thread or directly, should be up to the user. I made a few changes now to expose a different API that always goes via a `SubscribableChannel`, if you'd prefer not introducing more API. I think this would satisfy the most common use.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
```diff
@@ -0,0 +1,161 @@
...
+
+package org.springframework.integration.jdbc.channel;
+
+import org.springframework.integration.jdbc.store.JdbcChannelMessageStore;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.SubscribableChannel;
+import org.springframework.util.Assert;
+
+final class PostgresChannelMessageTableSubscription {
```
artembilan, line 25: I still don't see this class as a Spring Integration structure.
Your original concern was exactly about a `MessageChannel` to be ready for notification from the PostgreSQL and there is no a `MessageChannel` as a top-level Spring Integration API in your PR.
It is fully not clean why you resist to follow a simple `Subscription` contract interface and a `MessageChannel`.
Plus I said before that without an inner interface in the subscriber there is going to be a class tangle, which you exactly have just introduced with that `asSubscribableChannel()` factory method.

There is also no tests yet for this PR.
We have in the `jdbc` module a `MySqlContainerTest` to test against MySQL container in Docker.
Probably we can have something similar for Postgres to be sure in the functionality you introduce.

artembilan (changes requested):
If you are not going to implement a `MessageChannel` as a first-class citizen for Spring Integration functionality and not going to add tests, I cannot accept the PR as it is right now and take it over for original expectations to follow up. We have some other priorities at the moment, so it may take some time until we incorporate this feature.

Thank you for understanding!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscription.java
raphw: I still don't understand what API you are looking for. Before that, I would not want to spend time on cleaning up the implementation or adding tests as this would require additional effort for every API suggestion. Once we find out that we're on the same page regarding the API, it's easy to add tests and clean up, that's why I set this PR as a draft.

I made the subscription class package-private to indicate that it's not the API I would go for but rather expose a factory for a SubscribableChannel. We have not discussed ChannelDispatcher before, only subscribable channel, would you rather have it as a factory for such a dispatcher?

What do you mean by "inner interface"? That the PostgresChannelMessageTableSubscriber should implement one? There's no interface in Spring integration that fits the necessary API, so I would need to create one. This goes back to what we discussed that it needs to be shared across regions and group ids as it fully occupies a physical Postgres connection.

artembilan:
@raphw ,

my idea about an implementation and API is like this: https://github.com/artembilan/spring-integration/commit/92de4ca4d7718135a91327524fa3481157d49c3f

raphw:
As an example: a very similar implementation using Testcontainers that works: https://github.com/Skatteetaten/skatteprosessen-task/blob/main/task-processor-jdbc/src/main/java/no/skatteetaten/fastsetting/formueinntekt/felles/task/processor/jdbc/PostgresProcessorNotifier.java

raphw:
For some reason, Postgres does not like the inline concatenation when `NOTIFY` is used, but accepts it via the `pg_notify` function. The integration test is not passing and I would consider this a working solution.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresContainerTest.java
```diff
@@ -0,0 +1,93 @@
...
+ * started only once per JVM, therefore the target Docker container is reused automatically.
+ *
+ * @since 6
+ */
+@Testcontainers(disabledWithoutDocker = true)
+public interface PostgresContainerTest {
+
+	@SuppressWarnings({"unchecked", "rawtypes"})
+	PostgreSQLContainer<?> POSTGRES_CONTAINER = new PostgreSQLContainer("postgres:11") {{
+		this.setPortBindings(Collections.singletonList("5432:5432"));
```
artembilan, line 43: I don't think it is good idea to bind to specific port: better to rely on a randomly assigned and really use just that `getJdbcUrl()` in the target test.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresContainerTest.java
```diff
@@ -0,0 +1,93 @@
...
+	}};
+
+	@BeforeAll
+	static void startContainer() throws SQLException {
+		POSTGRES_CONTAINER.start();
+		try (Connection conn = DriverManager.getConnection(POSTGRES_CONTAINER.getJdbcUrl(),
+				POSTGRES_CONTAINER.getUsername(),
+				POSTGRES_CONTAINER.getPassword()); Statement stmt = conn.createStatement()) {
+			stmt.execute("CREATE SEQUENCE INT_MESSAGE_SEQ START WITH 1 INCREMENT BY 1 NO CYCLE");
```
artembilan, line 52: I think this DDL has to be a part of the target test.
I have an idea to add tests for other Spring Integration JDBC stuff against Postgres afterwards.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriberTest-context.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:jdbc="http://www.springframework.org/schema/jdbc"
+	xsi:schemaLocation="http://www.springframework.org/schema/jdbc https://www.springframework.org/schema/jdbc/spring-jdbc.xsd
+			http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd">
+
+	 <bean id="dataSource" class="org.apache.commons.dbcp2.BasicDataSource" destroy-method="close">
```
artembilan, line 8: Let's consider to rework this to the Java & Annotations configuration!

artembilan (changes requested):
This is good progress.
This probably won't make it into today's release, but definitely next milestone.
Thank you!

raphw:
I addressed all comments but am left with not having the subscriber being registered as a bean to allow for two tests where subscribers are registered in a started and in a stopped condition.

artembilan (changes requested):
Good.
Would you mind to rebase your branch to the latest `main` and I'll it locally for final review.
Perhaps it may make it into today's release after all 😃 .

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresContainerTest.java
```diff
@@ -0,0 +1,59 @@
...
+ * Since the Postgres container instance is shared via static property, it is going to be
+ * started only once per JVM, therefore the target Docker container is reused automatically.
+ *
+ * @author Rafael Winterhalter
+ * @since 6.0
+ */
+@Testcontainers(disabledWithoutDocker = true)
+public interface PostgresContainerTest {
+
+	PostgreSQLContainer<?> POSTGRES_CONTAINER = new PostgreSQLContainer<>("postgres:11");
```
artembilan, line 36: Why not the latest `14.5` tag?

artembilan (changes requested):

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresContainerTest.java
raphw: To make sure no functions or features being used are not only supported by the newest Postgres versions. Plenty of old installations out there but 11 is the oldest one that is still common.

raphw:
I rebased on master.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/14d85977d7005e4b408c67a2d58888b0c9e53a60.

Unfortunately I don't have time today to add docs for this new feature, but we are OK to go with then in the next milestone.

@raphw ,

thank you very much for the contribution; looking forward for more! For example those docs for this new feature 😄

raphw:
Of course, thank you for your feedback along the way.

As for the documentation: https://github.com/spring-projects/spring-integration/pull/3890

PR #3892, opened 2022-09-23
GH-3572: Migrate SFTP from `jsch` to `sshd-sftp`

### LINKED ISSUE #3572: Sftp lacks support for ssh-ed25519 due to unmaintained library

For example:
5.5.0 and all previous version

**Describe the bug**

The DefaultSftpSessionFactory does not support host keys of type `ssh-ed25519`.
If it tries to validate one it will return with 2 messages:

```
INFO: The authenticity of host '****' can't be established.\nRSA key fingerprint is ****.\nAre you sure you want to continue connecting?
ERROR: MessagingException Failed to execute on session; nested exception is java.lang.IllegalStateException: failed to create SFTP Session
``` 

However: If all unknown keys are accepted, the connection can be established. This might cause people to rather not validate keys because of convenience and would open the door for vulnerabilities.

**To Reproduce**

Start an SFTP server using `ssh-ed25519` based host keys. In my case I'm using the docker image `atmoz/sftp`.
Connect to the server via sftp cli and accept the key.
Use the known hosts to instantiate a SftRemoteFileTemplate.

**Expected behavior**

The host key is properly validated and used.

**Background information**
The reason for this issue is the lack of support for this key type in the JSch library. Furthermore the library seems to be no longer maintained as nobody answers on maintenance requests on the mailing list.
The lack of key type support will be a security issue in the future. More details can also be found here: http://www.matez.de/index.php/2020/06/22/the-future-of-jsch-without-ssh-rsa/

Several forks such as https://github.com/mwiede/jsch have appeared that resolve these issues. Maybe it's possible to swap the dependency while keeping compatibility.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3572

* Rework SFTP module from the JSch API to more modern `sshd-sftp`
* Migrate generics of most API from `ChannelSftp.LsEntry` to the `SftpClient.DirEntry`
* Rework `DefaultSftpSessionFactory` to deal with an `SshClient` and create `SftpClient` wrapped to the `SftpSession`
* Implement a `ResourceKnownHostsServerKeyVerifier` to load `known-hosts` from any possible resource
* Implement an expected `SftpSession.list()` with just file name to take or pattern matching
* Remove some unused tests and their config
* Remove tests for custom `UserInfo` since we don't provide any custom out-of-the-box
* Test a new `ResourceKnownHostsServerKeyVerifier` against default `known-hosts` file


artembilan:
This is huge breaking change for SFTP module, but it is still justified with today's trends.
Plus JSch is really obsoleted and abandoned library.

src/reference/asciidoc/sftp.adoc
```diff
@@ -10,6 +10,12 @@ The SFTP protocol requires a secure channel, such as SSH, and visibility to a cl
 Spring Integration supports sending and receiving files over SFTP by providing three client side endpoints: inbound channel adapter, outbound channel adapter, and outbound gateway.
 It also provides convenient namespace configuration to define these client components.
 
+NOTE: Starting with version 6.0, an outdated JCraft JSch client has been replaced with modern https://mina.apache.org/sshd-project/index.html[Apache MINA SSHD] framework.
```
garyrussell, line 13:
```suggestion
NOTE: Starting with version 6.0, the outdated JCraft JSch client has been replaced with the https://mina.apache.org/sshd-project/index.html[Apache MINA SSHD] framework.
```

garyrussell (changes requested):
Just one minor doc change and a question/comment.

garyrussell (approved):

garyrussell:
Please rebase to address what's new conflct.

artembilan:
A Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.x-to-6.0-Migration-Guide#migrate-sftp-module-from-jsch-to-apache-mina

PR #3895, opened 2022-09-26
GH-3626: RabbitMQ Stream Support

### LINKED ISSUE #3626: Add RabbitStreamMessageHandler
Using a StreamRabbitTemplate

https://github.com/spring-projects/spring-amqp/pull/1366

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/3626

- spec for stream listener container
- move message handler from scst to here; add spec

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/Amqp.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/Amqp.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2021 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,11 @@ import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
 import org.springframework.integration.amqp.channel.PollableAmqpChannel;
 import org.springframework.integration.amqp.inbound.AmqpMessageSource.AmqpAckCallbackFactory;
 import org.springframework.lang.Nullable;
+import org.springframework.rabbit.stream.listener.StreamListenerContainer;
+import org.springframework.rabbit.stream.producer.RabbitStreamTemplate;
+
+import com.rabbitmq.stream.Codec;
+import com.rabbitmq.stream.Environment;
 
 /**
  * Factory class for AMQP components.
@@ -240,12 +245,64 @@ public final class Amqp {
 	}
 
 	/**
+	 * Create an initial {@link RabbitInboundChannelAdapterSLCSpec}
+	 * with the provided {@link StreamListenerContainer}.
+	 * Note: only endpoint options are available from spec.
+	 * The {@code listenerContainer} options should be specified
+	 * on the provided {@link StreamListenerContainer} using
+	 * {@link RabbitInboundChannelAdapterSLCSpec#configureContainer(java.util.function.Consumer)}.
+	 * @param listenerContainer the listenerContainer.
+	 * @return the RabbitInboundChannelAdapterSLCSpec.
+	 */
+	public static RabbitInboundChannelAdapterSLCSpec inboundAdapter(StreamListenerContainer listenerContainer) {
+		return new RabbitInboundChannelAdapterSLCSpec(listenerContainer);
+	}
+
+	/**
+	 * Create an initial {@link RabbitInboundChannelAdapterSLCSpec}
+	 * with the provided {@link Environment}.
+	 * Note: only endpoint options are available from spec.
+	 * The {@code listenerContainer} options should be specified
+	 * on the provided {@link StreamListenerContainer} using
+	 * {@link RabbitInboundChannelAdapterSLCSpec#configureContainer(java.util.function.Consumer)}.
+	 * @param environment the environment.
+	 * @return the RabbitInboundChannelAdapterSLCSpec.
+	 */
+	public static RabbitInboundChannelAdapterSLCSpec inboundAdapter(Environment environment) {
+		return new RabbitInboundChannelAdapterSLCSpec(environment, null);
+	}
+
+	/**
+	 * Create an initial {@link RabbitInboundChannelAdapterSLCSpec}
+	 * with the provided {@link Environment}.
+	 * Note: only endpoint options are available from spec.
+	 * The {@code listenerContainer} options should be specified
+	 * on the provided {@link StreamListenerContainer} using
+	 * {@link RabbitInboundChannelAdapterSLCSpec#configureContainer(java.util.function.Consumer)}.
+	 * @param environment the environment.
+	 * @param codec the codec.
+	 * @return the RabbitInboundChannelAdapterSLCSpec.
+	 */
+	public static RabbitInboundChannelAdapterSLCSpec inboundAdapter(Environment environment, Codec codec) {
+		return new RabbitInboundChannelAdapterSLCSpec(environment, codec);
+	}
+
+	/**
 	 * Create an initial AmqpOutboundEndpointSpec (adapter).
 	 * @param amqpTemplate the amqpTemplate.
 	 * @return the AmqpOutboundEndpointSpec.
 	 */
 	public static AmqpOutboundChannelAdapterSpec outboundAdapter(AmqpTemplate amqpTemplate) {
 		return new AmqpOutboundChannelAdapterSpec(amqpTemplate);
+	}
+
+	/**
+	 * Create an initial {@link RabbitStreamMessageHandlerSpec} (adapter).
+	 * @param template the amqpTemplate.
+	 * @return the RabbitStreamMessageHandlerSpec.
+	 */
+	public static RabbitStreamMessageHandlerSpec outboundStreamAdapter(RabbitStreamTemplate template) {
+		return new RabbitStreamMessageHandlerSpec(template);
 	}
 
 	/**
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/Amqp.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/Amqp.java
@@ -25,11 +25,6 @@ import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
 import org.springframework.integration.amqp.channel.PollableAmqpChannel;
 import org.springframework.integration.amqp.inbound.AmqpMessageSource.AmqpAckCallbackFactory;
 import org.springframework.lang.Nullable;
-import org.springframework.rabbit.stream.listener.StreamListenerContainer;
-import org.springframework.rabbit.stream.producer.RabbitStreamTemplate;
-
-import com.rabbitmq.stream.Codec;
-import com.rabbitmq.stream.Environment;
 
 /**
  * Factory class for AMQP components.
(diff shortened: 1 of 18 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/Amqp.java
```diff
@@ -25,6 +25,11 @@
 import org.springframework.integration.amqp.channel.PollableAmqpChannel;
 import org.springframework.integration.amqp.inbound.AmqpMessageSource.AmqpAckCallbackFactory;
 import org.springframework.lang.Nullable;
+import org.springframework.rabbit.stream.listener.StreamListenerContainer;
+import org.springframework.rabbit.stream.producer.RabbitStreamTemplate;
+
+import com.rabbitmq.stream.Codec;
+import com.rabbitmq.stream.Environment;
```
artembilan, line 32: I saw these all as an `optional` API in the dependencies, so this `Amqp` factory class is going to fail when `srteam` is not available. No? What do I miss?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/Amqp.java
```diff
@@ -248,6 +296,15 @@ public static AmqpOutboundChannelAdapterSpec outboundAdapter(AmqpTemplate amqpTe
 		return new AmqpOutboundChannelAdapterSpec(amqpTemplate);
 	}
 
+	/**
+	 * Create an initial {@link RabbitStreamMessageHandlerSpec} (adapter).
+	 * @param template the amqpTemplate.
+	 * @return the RabbitStreamMessageHandlerSpec.
```
artembilan, line 302: `@since 6.0` for all the new factory methods?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/RabbitInboundChannelAdapterSLCSpec.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.springframework.amqp.rabbit.listener.DirectMessageListenerContainer;
+import org.springframework.lang.Nullable;
+import org.springframework.rabbit.stream.listener.StreamListenerContainer;
+
+import com.rabbitmq.stream.Codec;
+import com.rabbitmq.stream.Environment;
+
+/**
+ * Spec for an inbound channel adapter with a {@link DirectMessageListenerContainer}.
```
artembilan, line 29: Doesn't look like this JavaDoc reflects the reality of the class...

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/RabbitInboundChannelAdapterSLCSpec.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author Gary Russell
+ * @author Artem Bilan
+ *
+ * @since 6.0
+ *
+ */
+public class RabbitInboundChannelAdapterSLCSpec
+		extends AmqpInboundChannelAdapterSpec<RabbitInboundChannelAdapterSLCSpec, StreamListenerContainer> {
```
artembilan, line 38: Looks like this spec is just for the `StreamListenerContainer`. Why then the name of the class is so vague and doesn't say anything about Streams?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/RabbitStreamMessageHandlerSpec.java
```diff
@@ -0,0 +1,114 @@
...
+
+import org.springframework.integration.amqp.outbound.AbstractAmqpOutboundEndpoint;
+import org.springframework.integration.amqp.outbound.RabbitStreamMessageHandler;
+import org.springframework.integration.amqp.support.AmqpHeaderMapper;
+import org.springframework.integration.amqp.support.DefaultAmqpHeaderMapper;
+import org.springframework.integration.dsl.MessageHandlerSpec;
+import org.springframework.rabbit.stream.producer.RabbitStreamOperations;
+
+/**
+ * The base {@link MessageHandlerSpec} for {@link AbstractAmqpOutboundEndpoint}s.
```
artembilan, line 27: Please, fix the JavaDoc for real class purpose around `RabbitStreamMessageHandler`

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/RabbitStreamMessageHandlerSpec.java
```diff
@@ -0,0 +1,114 @@
+/*
+ * Copyright 2016-2020 the original author or authors.
```
artembilan, line 2: Copy/paste artifact?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -0,0 +1,231 @@
+/*
+ * Copyright 2021-2022 the original author or authors.
```
artembilan, line 2: Just `2022`

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -0,0 +1,231 @@
...
+				amqpMessageProperties);
+	}
+
+	@Override
+	public void start() {
+	}
+
+	@Override
+	public void stop() {
+		this.streamOperations.close();
```
artembilan, line 199: This is not OK. The `streamOperations` is an external object: we cannot mutate it in this handler.
What if the same `streamOperations` is used in other places where it must not be stopped yet?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -0,0 +1,231 @@
...
+		 * Note that Exceptions raised by this method are ignored.
+		 * @param result the result of the future
+		 */
+		void onSuccess(Message<?> result);
+	}
+
+	/**
+	 * Callback for when publishing fails.
+	 */
+	public interface FailureCallback {
```
artembilan, line 222: Why is that?
Why not channels like we do in the `KafkaProducerMessageHandler`?

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/support/RabbitTestContainer.java
```diff
@@ -0,0 +1,56 @@
...
+import java.time.Duration;
+
+import org.testcontainers.containers.RabbitMQContainer;
+
+/**
+ * Provides a static {@link RabbitMQContainer} that can be shared across test classes.
+ *
+ * @author Chris Bono
+ */
+public final class RabbitTestContainer {
```
artembilan, line 28: See `MySqlContainerTest`, `MongoDbContainerTest`, `MosquittoContainerTest` etc.
I really prefer to have them as interfaces.
And no any `IMAGE_CACHE`, please.

artembilan (changes requested):
And docs, please.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/RabbitInboundChannelAdapterSLCSpec.java
garyrussell: It says SLC (like the others) I removed the Amqp because it is not amqp; but I can rename to include Stream.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
garyrussell: Well, strictly, 2021 is correct because this class was written in SCSt.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
garyrussell: I just copied what's in SCSt; but I agree.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/support/RabbitTestContainer.java
artembilan: No, we don't. The Harbor is used by default from its Docker environment on Bamboo.
It might be different on Jenkins and so, but I indeed sure that we have a proper mirroring on Bamboo.
I fact I removed those extra ENV vars from Spring AMQP CI plan to realign it with whatever works for me in Spring Integration plans.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
artembilan: Well, I didn't know the license works that way: it is really just a new class in this project, so its data is just actual, nothing more.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/RabbitInboundChannelAdapterSLCSpec.java
artembilan: OK. Why it cannot be then just `RabbitStreamInboundChannelAdapterSpec`?
I don't think it make sense to say something about "listener container" in its name...

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -44,7 +45,7 @@
  * @since 6.0
  *
  */
-public class RabbitStreamMessageHandler extends AbstractMessageHandler implements Lifecycle {
+public class RabbitStreamMessageHandler extends AbstractMessageProducingHandler {
```
artembilan, line 48: Why is that? What do we produce from here?

src/reference/asciidoc/whats-new.adoc
```diff
@@ -104,3 +104,8 @@ See <<./jdbc.adoc#jdbc-lock-registry,JDBC Lock Registry>> for more information.
 The `lookupHost` property of the `AbstractConnectionFactory` and `DatagramPacketMessageMapper` is now set to `false` by default to avoid delays in the environments where DNS is not configured.
 
 See <<./ip.adoc#ip,TCP and UDP Support>> for more information.
+
+=== RabbitMQ Changes
+
+The AMQP module has been enhanced to provide support for inbound and outbound channel adapters using RabbitMQ Stream Queues.
+See <<./amqp.adocrmq-streams>> for more information.
```
artembilan: Cannot we move this up to the `New Components` section?
These are really new features even if in the existing module

artembilan (changes requested):

garyrussell:
Do not merge - needs another change...

garyrussell:
Ignore that last message - the change is needed in SCSt.

PR #3898, opened 2022-09-30
Fix new classes and packages tangles

artembilan:
* Move `MessagingAnnotationPostProcessor` and `MethodAnnotationPostProcessor` impls out of the `config.annotation` package due to usage of the `FactoryBean` configs
* Move `GenericHandler` and `GenericTransformer` to the `core` package to break a tangle with a `LambdaMessageProcessor`
* Clean up affected tests and docs


artembilan:
This is some kind of a breaking change, but I guess it is still legit to do this in the current major version milestone.
Will add a Migration Guide note after merging.

artembilan:
The Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.x-to-6.0-Migration-Guide#class-and-package-tangles

PR #3899, opened 2022-09-30
GH-3635: Add Future<Void> & Mono<Void> to gateway

### LINKED ISSUE #3635: Future<Void> method with @Gateway causes thread leak

org.springframework.boot:spring-boot-starter-integration: '2.2.6.RELEASE'

**Describe the bug**

If response type is `Future<Void>`,  the method with `@Gateway`  causes thread leak. New threads are constantly spawned without reusing or disposing. It was confirmed on thread dump I collected in my application.  Even though each memory usage is quite small, threads being racked up will eventually lead OOM.
I assume that this is happening because of no acknowledgement with Void. 

However, there's a solution. It simply can be avoided by replacing it with `void send` along with `ExecutorService` in order to keep it asynchronous. (Non-blocking messaging is what I wanna do)
e.g.
```
@Bean
public MessageChannel sampleChannel() {
    return new PublishSubscribeChannel(Executors.newCachedThreadPool());
}

```

This is maybe a bug or should be described clearly in the docs if not a bug imhp.

**To Reproduce**

* Register bean at a config class
```
@Bean
public MessageChannel sampleChannel() {
    return MessageChannels.publishSubscribe().get();
}
```

* Make MessagingGateway have a method returns Future<Void> as following
```
@MessagingGateway
public interface SampleMessagingGateway { 
  @Gateway(requestChannel = "sampleChannel")  
  Future<Void> send(Payload payload);
}
```

* `send(payload)` at a producer class.

*  Have a consumer class which contains method like
```
@ServiceActivator(inputChannel = "sampleChannel")
public void onSample(Payload payload){
  // do something 
}
```
As a result, threads spawned by that consumers stays in the lifetime of application apparently unless the application is restarted or so. Thus this will eventually lead OOM. 

**Expected behavior**

* It won't cause thread leak with `Future<Void>`

Or 

* Described in doc well. This behavior seems not described explicitly at [asyn-gateway doc](https://docs.spring.io/spring-integration/reference/html/gateway.html#async-gateway).  If you wanna deal with it void type, the solution using `ExecutorService` should be recommended. 

**Sample**

See above. 

Thx in advance.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3635

When `Future<Void>` & `Mono<Void>` is used as a messaging gateway return type, the application hangs out on this barrier which may lead to the out of memory eventually

* Add support for the `Future<Void>` & `Mono<Void>` messaging gateway return type and ensure an asynchronous call for the `gateway.send(Message)` operation and its exception handling.
In case of successful call, the `Future` is fulfilled with `null` and `Mono` is completed as empty

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/gateway/AsyncGatewayTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/gateway/AsyncGatewayTests.java
@@ -30,10 +30,13 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.jupiter.api.Test;
 
+import org.springframework.beans.DirectFieldAccessor;
 import org.springframework.beans.factory.BeanFactory;
+import org.springframework.integration.MessageDispatchingException;
 import org.springframework.integration.annotation.Gateway;
 import org.springframework.integration.annotation.GatewayHeader;
 import org.springframework.integration.channel.DirectChannel;
+import org.springframework.integration.channel.NullChannel;
 import org.springframework.integration.channel.QueueChannel;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
@@ -42,6 +45,7 @@ import org.springframework.messaging.support.ChannelInterceptor;
 import org.springframework.messaging.support.MessageBuilder;
 
 import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
 
 /**
  * @author Mark Fisher
@@ -70,7 +74,7 @@ public class AsyncGatewayTests {
 	}
 
 	@Test
-	public void futureWithError() throws Exception {
+	public void futureWithError() {
 		final Error error = new Error("error");
 		DirectChannel channel = new DirectChannel() {
 
@@ -140,7 +144,7 @@ public class AsyncGatewayTests {
 	}
 
 	@Test
-	public void nonAsyncFutureReturned() throws Exception {
+	public void nonAsyncFutureReturned() {
 		QueueChannel requestChannel = new QueueChannel();
 		addThreadEnricher(requestChannel);
 		startResponder(requestChannel);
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/gateway.adoc
```diff
@@ -801,7 +800,9 @@ public interface MyGateway {
 ----
 ====
 
-Unlike the `Future<?>` return types, there is no way to inform the caller if some exception is thrown by the flow, unless some custom `TaskExecutor` (such as an `ErrorHandlingTaskExecutor`) is associated with the `@Async` annotation.
+But downstream exceptions are not going to be propagated back to the caller.
+To ensure asynchronous behaviour for downstream flow invocation and exception propagation to the caller, starting with version 6.0, the framework provides support for the `Future<Void>` and `Mono<Void>` return types.
```
garyrussell, line 804: American English (`behavior`).

garyrussell (changes requested):
Also refer to our out-of-band discussion about a two-way flow that returns one of these types.

spring-integration-core/src/test/java/org/springframework/integration/gateway/AsyncGatewayTests.java
```diff
@@ -204,6 +210,59 @@ public void futureWithWildcardReturned() throws Exception {
...
+		new Thread(() -> {
+			try {
+				Message<?> input = requestChannel.receive();
+				CompletableFuture<Void> reply = new CompletableFuture<>();
+				((MessageChannel) input.getHeaders().getReplyChannel()).send(new GenericMessage<>(reply));
+				readyForReplyLatch.await(10, TimeUnit.SECONDS);
+				reply.complete(null);
+			}
+			catch (InterruptedException ex) {
+				ReflectionUtils.rethrowRuntimeException(ex);
```
garyrussell, line 251: `Thread.currentThread.interrupt()` before rethrow.

garyrussell (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/gateway/AsyncGatewayTests.java
artembilan: This is a unit test. Why do we need to care about that?

garyrussell: It's just always good practice to do that - e.g. if JUnit does a wait anywhere later on this thread, it will block because the interrupted flag has been reset.

Probably not an issue, but might be.

garyrussell: I didn't notice it was a `new Thread()` so less important, I agree, but my eye is always drawn to these; years ago, I got into the practice of automatically adding it; I think it's a good practice so you don't miss one in an important place.

PR #3906, opened 2022-10-06
Improve messaging gateway mapping

artembilan:
The `#args` and `#gatewayMethod` SpEL variables have been deprecated for a while

* Remove their population and usage in favor of `MethodArgsHolder` `root` of the evaluation context This change optimize a gateway mapping logic the way that there is no need in evaluation context for every call: we can just reuse a global one
* Some other `GatewayMethodInboundMessageMapper` code style refactoring
* Fix effected test classes and their configs


garyrussell (approved):

artembilan:
Here it is: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-5.x-to-6.0-Migration-Guide#no-args-and-gatewaymethod-for-gateway-expressions.

Thank you for the pointer, @Shiftuia !

PR #3907, opened 2022-10-07
INT-3333: Return empty list as is from DB gateway

artembilan:
Fixes https://jira.spring.io/browse/INT-3333

In `JdbcOutboundGateway` and `JpaOutboundGateway` the empty result list is treated as "no reply" and therefore `null` is returned cause the flow to stop at this point.
It is better to return such a result as is and the target application to decided what to do with it, e.g. a `discardChannel` on a downstream splitter configuration.

NOTE: the `MongoDbOutboundGateway` doesn't treat an empty result as a `null`

### DIFF
--- a/src/reference/asciidoc/jdbc.adoc
+++ b/src/reference/asciidoc/jdbc.adoc
@@ -365,6 +365,10 @@ It can also have a `SqlParameterSourceFactory` injected to control the binding o
 Starting with the version 4.2, the `request-prepared-statement-setter` attribute is available on the `<int-jdbc:outbound-gateway>` as an alternative to `request-sql-parameter-source-factory`.
 It lets you specify a `MessagePreparedStatementSetter` bean reference, which implements more sophisticated `PreparedStatement` preparation before its execution.
 
+Starting with the version 6.0, the `JdbcOutboundGateway` returns an empty list result as is instead of converting it to `null` as it was before with the meaning "no reply".
+This caused an extra configuration in applications where handling of empty lists is a part of downstream logic.
+See <<./splitter.adoc#split-stream-and-flux,Splitter Discard Channel>> for possible empty list handling option.
+
 See <<jdbc-outbound-channel-adapter>> for more information about `MessagePreparedStatementSetter`.
 
 [[jdbc-message-store]]
--- a/src/reference/asciidoc/jpa.adoc
+++ b/src/reference/asciidoc/jpa.adoc
@@ -1011,10 +1011,89 @@ public class JpaJavaApplication {
 [[jpa-retrieving-outbound-gateway]]
 ==== Retrieving Outbound Gateway
 
-The following example shows all the attributes that you can set on a retrieving outbound gateway and describes the key attributes:
+The following example demonstrates how to configure a retrieving outbound gateway:
 
 ====
-[source,xml]
+[source, java, role="primary"]
+.Java DSL
+----
+@SpringBootApplication
+@EntityScan(basePackageClasses = StudentDomain.class)
+public class JpaJavaApplication {
+
+    public static void main(String[] args) {
+        new SpringApplicationBuilder(JpaJavaApplication.class)
+            .web(false)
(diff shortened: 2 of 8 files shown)


src/reference/asciidoc/jdbc.adoc
```diff
@@ -365,6 +365,10 @@ It can also have a `SqlParameterSourceFactory` injected to control the binding o
 Starting with the version 4.2, the `request-prepared-statement-setter` attribute is available on the `<int-jdbc:outbound-gateway>` as an alternative to `request-sql-parameter-source-factory`.
 It lets you specify a `MessagePreparedStatementSetter` bean reference, which implements more sophisticated `PreparedStatement` preparation before its execution.
 
+Starting with the version 6.0, the `JdbcOutboundGateway` returns an empty list result as is instead of converting it to `null` as it was before with the meaning "no reply".
+This caused an extra configuration in applications where handling of empty lists is a part of downstream logic.
+See <<./splitter.adoc#split-stream-and-flux,Splitter Discard Channel>> for possible empty list handling option.
```
garyrussell, line 370:
```suggestion
Starting with the version 6.0, the `JdbcOutboundGateway` returns an empty list result when there are no records returned by the query. 
Previously, `null` was returned, ending the flow, or throwing an exception, depending on `requiresReply`.
This required extra configuration in applications where handling of empty lists is part of the downstream logic.
See <<./splitter.adoc#split-stream-and-flux,Splitter Discard Channel>> for possible empty list handling options.
Or, to revert to the previous behavior, add a `filter` after the gateway to filter out empty lists.
```

src/reference/asciidoc/jpa.adoc
```diff
@@ -1171,6 +1170,11 @@ It does not cascade to related entities.`"
 For more information, see https://jcp.org/en/jsr/detail?id=317[JSR 317: Java™ Persistence 2.0]
 ====
 
+NOTE: Starting with version 6.0, the `Jpa.retrievingGateway()` returns an empty list result as is.
+Previously it was converted to `null` with the meaning "no reply".
+It causes an extra configuration in the applications where empty list handling is a part of business logic.
+See <<./splitter.adoc#split-stream-and-flux,Splitter Discard Channel>> for possible empty list handling option.
```
garyrussell, line 1176:
```suggestion
NOTE: Starting with version 6.0, the `Jpa.retrievingGateway()` returns an empty list result when there are no entities returned by the query.
Previously `null` was returned ending the flow, or throwing an exception, depending on `requiresReply`.
Or, to revert to the previous behavior, add a `filter` after the gateway to filter out empty lists.
It requires extra configuration in applications where empty list handling is a part of the downstream logic.
See <<./splitter.adoc#split-stream-and-flux,Splitter Discard Channel>> for possible empty list handling options.
```

src/reference/asciidoc/whats-new.adoc
```diff
@@ -93,6 +93,9 @@ See <<./gateway.adoc#async-gateway, Asynchronous Gateway>> for more information.
 
 The `integrationGlobalProperties` bean is now declared by the framework as an instance of `org.springframework.integration.context.IntegrationProperties` instead of the previously deprecated `java.util.Properties`.
 
+The message handlers which produce a collection as a reply (e.g. `JpaOutboundGateway`, `JdbcOutboundGateway` and other DB-based gateways) now return the empty result list as is.
+Previously it was converted to `null` with the meaning no reply as it states about `null` payloads in the framework at the moment.
```
garyrussell, line 97:
```suggestion
Message handlers which produce a collection as a reply (e.g. `JpaOutboundGateway`, `JdbcOutboundGateway` and other DB-based gateways) now return an empty result list if no records are returned by the query.
Previously, `null` was returned ending the flow, or throwing an exception, depending on `requiresReply`.
```

garyrussell (changes requested):
Just doc polishing.

garyrussell (approved):

PR #3911, opened 2022-10-11
GH-3664: Re-enable JavaScript support via GraalVM

### LINKED ISSUE #3664: Rework JavaScript support to GraalVM
See more info in GraalVM: https://www.graalvm.org/reference-manual/js/JavaInteroperability/.

It may turn out just to a simple `GraalVmJavaScriptExecutor` implementation since making it working we need to expose these variables:
```
<script:variable name="polyglot.js.allowHostAccess" value="#{true}"/>
<script:variable name="polyglot.js.allowHostClassLookup" value="#{true}"/>
```

For now JavaScript tests are disable if no GraalVM in classpath.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3664

The JavaScript JSR223 engine has been removed from Java.

Migrate JavaScript support into GraalVM Polyglot API:
* Implement `PolyglotScriptExecutor` which can also support other GraalVM languages
* Change the `ScriptExecutorFactory` to use a `PolyglotScriptExecutor` when it encounters JavaScript
* Re-enable tests for JavaScript
* Add GraalVM Polyglot support into docs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,6 +65,7 @@ ext {
     derbyVersion = '10.14.2.0'
     findbugsVersion = '3.0.1'
     ftpServerVersion = '1.2.0'
+    graalvmVersion = '22.2.0'
     greenmailVersion = '2.0.0-alpha-2'
     groovyVersion = '3.0.13'
     hamcrestVersion = '2.2'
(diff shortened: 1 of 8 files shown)


artembilan:
I have decided to not touch Ruby and Python yet.
We may revise them in favor of GraalVM as well eventually.
Plus it will open doors for other languages, like R.

PR #3916, opened 2022-10-14
GH-3421: Resolve `throws Exception;` in the API

### LINKED ISSUE #3421: Revise all the throws Exception; // NOSONAR
For example we have a code in the `WebSocketListener` like this:
```
void onMessage(WebSocketSession session, WebSocketMessage<?> message)
			throws Exception; // NOSONAR Remove in 5.2
```

so, all those `throws Exception` have to be removed.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3421

Remove `throws Exception;` from production code to honor the rule `Generic exceptions should never be thrown` which is enabled on SonarQube

* Rework affected usages to `try..catch` with throwing respective runtime exception or just logging
* Some other refactoring for the affected classes

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/transformer/MailToStringTransformer.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/transformer/MailToStringTransformer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ import jakarta.mail.Multipart;
 import jakarta.mail.Part;
 
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
+import org.springframework.integration.transformer.MessageTransformationException;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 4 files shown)


spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
```diff
@@ -224,39 +224,54 @@ public List<String> getSubProtocols() {
...
+	public void afterSessionEnded(WebSocketSession session, CloseStatus closeStatus) {
 
 		if (isActive()) {
-			this.subProtocolHandlerRegistry.findProtocolHandler(session)
-					.afterSessionEnded(session, closeStatus, this.subProtocolHandlerChannel);
+			try {
+				this.subProtocolHandlerRegistry.findProtocolHandler(session)
+						.afterSessionEnded(session, closeStatus, this.subProtocolHandlerChannel);
+			}
+			catch (Exception ex) {
```
garyrussell, line 257: I think you'll need `// NOSONAR' for these to avoid Sonar complaining.

spring-integration-mail/src/main/java/org/springframework/integration/mail/transformer/MailToStringTransformer.java
```diff
@@ -51,27 +52,34 @@ public void setCharset(String charset) {
...
-		if (content instanceof String) {
-			return this.getMessageBuilderFactory().withPayload((String) content);
+			return getMessageBuilderFactory().withPayload(payload);
 		}
-		if (content instanceof Multipart) {
-			ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-			((Multipart) content).writeTo(outputStream);
-			return this.getMessageBuilderFactory().withPayload(
-					new String(outputStream.toByteArray(), this.charset));
+		catch (Exception ex) {
```
garyrussell, line 79:
```suggestion
		catch (Exception ex) { // NOSONAR
```

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
```diff
@@ -224,39 +224,54 @@ public List<String> getSubProtocols() {
...
+					accessor.setSessionId(session.getId());
+					accessor.setLeaveMutable(true);
+					accessor.setAcceptVersion("1.1,1.2");
+
+					Message<?> connectMessage =
+							MessageBuilder.createMessage(EMPTY_PAYLOAD, accessor.getMessageHeaders());
+					protocolHandler.handleMessageToClient(session, connectMessage);
+				}
+			}
+			catch (Exception ex) {
```
garyrussell, line 243:
```suggestion
			catch (Exception ex) { // NOSONAR
```

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
garyrussell, line 257:
```suggestion
			catch (Exception ex) { // NOSONAR
```

garyrussell (changes requested):

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
artembilan: Why? What is that rule we would violate otherwise?

garyrussell: I might be confusing it with `catch (Throwable t)`; I guess leave as is.

artembilan: I see. Probably you mean then this: https://rules.sonarsource.com/java/tag/error-handling/RSPEC-1181, which indeed doesn't talk about `catch Exception`

garyrussell (approved):

PR #3919, opened 2022-10-18
GH-3912: Handler: ignore Groovy generated methods

### LINKED ISSUE #3912: MessagingMethodInvokerHelper should disregard synthetic methods

5.5.14

**Describe the bug**

I have a service object written in Groovy that I'm providing to the `handle` instruction on a DSL integration flow. This service method has a single public method plus some private fields and closures; the presence of the closures causes the generation of some synthetic methods to access the private fields in the top-level class. The `MessagingMethodInvokerHelper` throws an exception claiming ambiguous methods because it's processing the synthetics as fallbacks.

**To Reproduce**

Pass a class that carries synthetic methods to `handle`.

**Expected behavior**

I expect synthetic methods to be excluded. Specifically, at `MessagingMethodInvokerHelper:776`, the existing condition
```
!Modifier.isPublic(method1.getModifiers())
```
should be logically equivalent to
```
Modifier.isSynthetic(method1.getModifiers()) || !Modifier.isPublic(method1.getModifiers())
```

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3912

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
@@ -53,6 +53,7 @@ import org.springframework.core.KotlinDetector;
 import org.springframework.core.LocalVariableTableParameterNameDiscoverer;
 import org.springframework.core.MethodParameter;
 import org.springframework.core.ParameterNameDiscoverer;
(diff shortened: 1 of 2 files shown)


artembilan:
Back-port to `5.5.x` only a `MessagingMethodInvokerHelper`  change since there is no those Groovy tests.
Although Git won't let to cherry-pick it anyway and will just suggest to delete...

garyrussell:
... and back ported as described.

PR #3921, opened 2022-10-19
Fix Messaging annotations for MH bean factories

artembilan:
The `AbstractMethodAnnotationPostProcessor` makes a decision to use a `MessageHandler` bean definition as is without wrapping by the type of `MessageHandler`. There are some of them can be configured via `AbstractSimpleMessageHandlerFactoryBean`, e.g. an `AggregatorFactoryBean`

* Check for the `AbstractSimpleMessageHandlerFactoryBean` to return as is
* Widen the replying producer by the `AbstractMessageProducingHandler` type
* Add an `AggregatorFactoryBean` with a `@ServiceActivator` configuration to test coverage


artembilan:
The problem has popped up in the https://github.com/spring-cloud/stream-applications/issues/373

Issue #3923, opened 2022-10-21
Support `@Import`ing gateway interfaces


chrylis:
**Expected Behavior**
```
@MessagingGateway
interface MyGateway {
  void enqueue(MyObject o)
}

@Configuration
@Import(MyGateway)
class IntegrationConfig {}
```

A gateway proxy for `MyGateway` is added to the context.

**Current Behavior**

Currently, this import produces `BeanInstantiationException: Specified class is an interface`.

**Context**
[We usually use `@Import` rather than `@ComponentScan`](https://stackoverflow.com/questions/74158211/is-there-a-directive-to-individually-import-a-spring-integration-gateway-in-java) for more fine-grained control (and slightly faster startup). I'd like to be able to `@Import` a gateway interface in the same way we do for service classes. I think this would be possible with a `BeanFactoryPostProcessor`.

artembilan:
The `@Import` annotation is out of Spring Integration scope and its behavior cannot be controlled.

I don't know if Spring Data has something like manual Repository registration, but this `@MessagingGateway` has really been designed after Spring Data Repository interfaces scanning.

More over whatever you are asking is already available like this:
```
@Bean 
AnnotationGatewayProxyFactoryBean myGateway() {
    return new AnnotationGatewayProxyFactoryBean(MyGateway.class);
}
```

Can you tell us, please, what is wrong with this approach?

I gave you similar answer in your other issue: https://github.com/spring-projects/spring-integration/issues/3920.

Why did you ignore that and raised a new one, plus SO thread?

chrylis:
I asked a question on SO primarily so that whatever the best answer is found will be visible to anyone else who is trying to write similar code. The feature request here is very much not ignoring #3920; I'm suggesting something new that I know isn't supported right now but that would be convenient to have.

Please note that your solution _does not_ provide what I'm asking for, because the `AnnotationGatewayProxyFactoryBean` will not match `@ConditionalOnBean(MyGateway)`.

artembilan:
OK. So, you suggest then to have some thing like `@IntegrationImport` like a brother for existing `@IntegrationComponentScan`.
Kinda realign what we have so far with a `@ComponentScan` and its brother `@Import`.

I think what you are talking about with a conditional based on type can be fixed with a generic argument on the `AnnotationGatewayProxyFactoryBean`.
This is that kind of limitation Spring Framework has at the moment around target type guessing from the `FactoryBean`.
[...]

PR #3904, opened 2022-10-04
GH-3903: Improve AOT for gateway proxy beans

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3903

* Fix `MessagingGatewayRegistrar` to set a `targetType` on the bean definition instead of `FactoryBean.OBJECT_TYPE_ATTRIBUTE`
* Rework `GatewayProxyBeanRegistrationAotProcessor` to the `GatewayProxyInitializationAotProcessor implements BeanFactoryInitializationAotProcessor` to avoid custom code generation
* Remove tests which rely on the `FactoryBean.OBJECT_TYPE_ATTRIBUTE`

PR #3925, opened 2022-10-25
Add generic arg to the `GatewayProxyFactoryBean`

artembilan:
Related to https://github.com/spring-projects/spring-integration/issues/3923

When gateway proxy is declared manually, a `GatewayProxyFactoryBean` is used as a `@Bean`. In this case the info about target interface is not available on a `BeanDefinition`, unlike with a programmatic registration via `MessagingGatewayRegistrar`.

* Expose `<T>` on a `GatewayProxyFactoryBean` to make end-user to specify the type this gateway is going to be based on.
This allows Spring container to determine the type of the `FactoryBean` bean definition properly
* Migrate a programmatic bean definition registration from the `FactoryBean.OBJECT_TYPE_ATTRIBUTE` to the `targetType` property of the `BeanDefinition` based on a `ResolvableType.forClassWithGenerics()` to simulate generic arg for the application context
* Remove `ComponentsRegistration` from the `GatewayProxySpec` since it us out of use
* Fix affected tests for newly added generic arg on the `GatewayProxyFactoryBean`

### DIFF
--- a/src/reference/asciidoc/gateway.adoc
+++ b/src/reference/asciidoc/gateway.adoc
@@ -312,7 +312,7 @@ public interface TestGateway {
 IMPORTANT: Similarly to the XML version, when Spring Integration discovers these annotations during a component scan, it creates the `proxy` implementation with its messaging infrastructure.
 To perform this scan and register the `BeanDefinition` in the application context, add the `@IntegrationComponentScan` annotation to a `@Configuration` class.
 The standard `@ComponentScan` infrastructure does not deal with interfaces.
-Consequently, we introduced the custom `@IntegrationComponentScan` logic  to fine the `@MessagingGateway` annotation on the interfaces and register `GatewayProxyFactoryBean` instances for them.
+Consequently, we introduced the custom `@IntegrationComponentScan` logic to fine the `@MessagingGateway` annotation on the interfaces and register `GatewayProxyFactoryBean` instances for them.
 See also <<./configuration.adoc#annotations,Annotation Support>>.
 
 Along with the `@MessagingGateway` annotation you can mark a service interface with the `@Profile` annotation to avoid the bean creation, if such a profile is not active.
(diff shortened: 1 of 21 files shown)


artembilan:
Even if this looks like a breaking change, it is still legit for the current major release.
Plus it is really long overdue for this generic feature required by the `FactoryBean` contract.
With this fix the change in the #3904 would make sense and probably will generate the proper AOT code - already with a generic argument on the `GatewayProxyFactoryBean` bean definition 😄

src/reference/asciidoc/gateway.adoc
```diff
@@ -312,7 +312,7 @@ public interface TestGateway {
 IMPORTANT: Similarly to the XML version, when Spring Integration discovers these annotations during a component scan, it creates the `proxy` implementation with its messaging infrastructure.
 To perform this scan and register the `BeanDefinition` in the application context, add the `@IntegrationComponentScan` annotation to a `@Configuration` class.
 The standard `@ComponentScan` infrastructure does not deal with interfaces.
-Consequently, we introduced the custom `@IntegrationComponentScan` logic  to fine the `@MessagingGateway` annotation on the interfaces and register `GatewayProxyFactoryBean` instances for them.
+Consequently, we introduced the custom `@IntegrationComponentScan` logic to fine the `@MessagingGateway` annotation on the interfaces and register `GatewayProxyFactoryBean` instances for them.
```
garyrussell, line 315:
```suggestion
Consequently, we introduced the custom `@IntegrationComponentScan` logic to find the `@MessagingGateway` annotation on the interfaces and register `GatewayProxyFactoryBean` instances for them.
```

garyrussell (changes requested):
typo

chrylis:
This definitely looks helpful for a number of cases!

garyrussell (approved):

PR #3927, opened 2022-10-25
GH-3926: BeanNameGenerator for @MessagingGateway

### LINKED ISSUE #3926: Change default MessagingGateway bean name in 6.0
Currently, if `@MessagingGateway` is applied without a `name` parameter, Spring Integration applies the convention of camel-casing the interface's simple name. This convention is prone to collisions; for example, I have an interface `DeliverSink` that I've been retrofitting, and in the process I ended up trying to also name its channel `deliverSink` and the `@Bean IntegrationFlow deliverSink`, leading to the predictable failures.

Spring Boot moved to using the FQCN for `@ConfigurationProperties` and similar automatically-declared beans. As autowiring by type is now the usual style, I suggest changing the default name for declarative gateways to be the FQCN in Spring Integration 6.0. This would be a breaking change but adaptable for applications that need the old behavior by adding the `name` parameter.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3926

Current approach for generated bean name in the `MessagingGatewayRegistrar` is to decapitalize simple class name, which is similar to standard `AnnotationBeanNameGenerator`

* Make logic in the `MessagingGatewayRegistrar` based on the provided `BeanNameGenerator`
* Expose an `@IntegrationComponentScan.nameGenerator()` attribute to allow to customize default bean name generation strategy
* Introduce `IntegrationConfigUtils.annotationBeanNameGenerator()` to take a provided `AnnotationConfigUtils.CONFIGURATION_BEAN_NAME_GENERATOR` singleton or fallback to the `AnnotationBeanNameGenerator.INSTANCE`
* Use this utility in the `IntegrationComponentScanRegistrar` if no custom strategy is provided in the `@IntegrationComponentScan`
* Use same util from the `GatewayParser` since there is no custom naming strategy configuration
* Some other current Java level refactoring in the `IntegrationComponentScanRegistrar` and `MessagingGatewayRegistrar`


src/reference/asciidoc/gateway.adoc
```diff
@@ -319,6 +319,12 @@ Along with the `@MessagingGateway` annotation you can mark a service interface w
 
 Starting with version 6.0, an interface with the `@MessagingGateway` can also be marked with a `@Primary` annotation for respective configuration logic as its possible with any Spring `@Component` definition.
 
+The `@MessagingGateway.name()` attribute specifies a name for the target bean registered.
+If it is not provided, the  `@IntegrationComponentScan.nameGenerator()` (since version `6.0`) is consulted.
+And if that is not provided as well, an `AnnotationConfigUtils.CONFIGURATION_BEAN_NAME_GENERATOR` bean, provided for the application context, is used.
+Otherwise, the bean name generation falls back to the `AnnotationBeanNameGenerator` which results with an uncapitalized simple class name.
+In other words, the current naming strategy for gateway proxy beans is aligned with standard Spring Framework `@ComponentScan` approach.
+
```
garyrussell, line 327:
```suggestion
The `@MessagingGateway.name()` attribute specifies a name for the target bean that will be registered.
If it is not provided, the  `@IntegrationComponentScan.nameGenerator()` (since version `6.0`) is consulted.
And, if that is not provided, an `AnnotationConfigUtils.CONFIGURATION_BEAN_NAME_GENERATOR` bean, if present in the application context, is used.
Otherwise, the bean name generation falls back to the `AnnotationBeanNameGenerator` which results with an uncapitalized simple class name.
In other words, the current naming strategy for gateway proxy beans is aligned with standard Spring Framework `@ComponentScan` approach.

```

garyrussell (changes requested):

artembilan:
This probably can be revised when #3929 is merged, since it looks like we would need from now on only a `ClassPathBeanDefinitionScanner` customized to allow interfaces.
The rest can be delegated to the regular `ScannedGenericBeanDefinition` and in the end to a new `InstantiationAwareBeanPostProcessor`.
Thanks

artembilan:
Sorry, it took a bit more time than I expected: the annotation stuff has eaten my brain...

garyrussell (approved):

PR #3928, opened 2022-10-26
Fix double start for `AbstractEndpoint`

artembilan:
When we use POJO methods in the `IntegrationFlowAdapter`, the `IntegrationFlowAdapter` is set as a `target` for the `MessagingMethodInvokerHelper`. When endpoint is started by the application context, such a `start()` is propagated down to the `MessagingMethodInvokerHelper`.
And in our case back into an `IntegrationFlowAdapter` instance. This one, in turn, starts its `IntegrationFlow` internally which leads to the start of the mentioned endpoint in the beginning.
Therefore, we cause a recursive `start()` call on this endpoint from itself. The `running` flag is set when we are already done with the `doStart()` logic. Therefore a recursive `start()` call leads to two concurrent polling tasks in the `AbstractPollingEndpoint`.

* Check also for the `active` flag in the `AbstractEndpoint.start()` and reset it in case of exception in the `doStart()`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
@@ -16,7 +16,6 @@
 
 package org.springframework.integration.endpoint;
 
-import java.lang.reflect.UndeclaredThrowableException;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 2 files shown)


artembilan:
@garyrussell ,

this was the easiest way to fix the problem.
I'm not sure if there is an other option to check if we are causing a recursive `start()` call from itself.

Thank you!

garyrussell:
...and cherry-picked as 8f809a7489

PR #3929, opened 2022-10-28
GH-3923: Add @MessagingGateway @Import support

### LINKED ISSUE #3923: Support `@Import`ing gateway interfaces

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3923

Currently, the `@MessagingGateway` interfaces can be scanned or created explicitly as `@Bean` definition for `AnnotationGatewayProxyFactoryBean`

* Implement a `GatewayProxyBeanDefinitionPostProcessor` which is invoked before instantiation attempt on the `BeanDefinition`
* Verify `@Import` for `@MessagingGateway` interface in the `GatewayInterfaceTests`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -121,7 +121,6 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
 		if (!registry.containsBeanDefinition("gatewayProxyBeanDefinitionPostProcessor")) {
 			BeanDefinitionBuilder builder =
 					BeanDefinitionBuilder.genericBeanDefinition(GatewayProxyInstantiationPostProcessor.class,
-							() -> new GatewayProxyInstantiationPostProcessor(registry))
 							.addConstructorArgValue(registry)
 							.setRole(BeanDefinition.ROLE_INFRASTRUCTURE);
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -63,6 +63,7 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
 		if (importingClassMetadata != null) {
 			registerMessagingAnnotationPostProcessors(registry);
 		}
+		registerGatewayProxyInstantiationPostProcessor(registry);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


artembilan:
With this `InstantiationAwareBeanPostProcessor` feature it looks like we don't need a `MessagingGatewayRegistrar` to be used for a `@IntegrationComponentScan`: we just simply can rely on a standard `AnnotatedBeanDefinition` registration against an interface and and delegate to this post-processor to let the `AnnotationGatewayProxyFactoryBean` to do the trick with annotation attributes processing.
But this is going to be a different story already after merging this one and naming PR.

chrylis:
Very interesting! Will this functionality make it back to 5.x, or will it be 6.x only? (The re-baseline from Java 11 to Java 17 will be a hassle if required.)

artembilan:
Well, I'm a bit reluctant to introduce any breaking changes into maintenance versions.
More over this is not a bug, but a new feature: we never advertised an `@Import` for `@MessagingGateway`.
As a workaround it is simply enough to add that `GatewayProxyInstantiationPostProcessor` as a `static` bean into your own configuration.

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -116,4 +117,16 @@ private void registerMessagingAnnotationPostProcessors(BeanDefinitionRegistry re
 		}
 	}
 
+	private void registerGatewayProxyInstantiationPostProcessor(BeanDefinitionRegistry registry) {
+		if (!registry.containsBeanDefinition("gatewayProxyBeanDefinitionPostProcessor")) {
+			BeanDefinitionBuilder builder =
+					BeanDefinitionBuilder.genericBeanDefinition(GatewayProxyInstantiationPostProcessor.class,
+							() -> new GatewayProxyInstantiationPostProcessor(registry))
+							.addConstructorArgValue(registry)
```
garyrussell, line 125: You don't need both of these (instance supplier + constructor arg).
```suggestion
							.addConstructorArgValue(registry)
```

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
artembilan: I think we do need to AOT generation.
The supplier works in for regular JVM, but AOT will generate for us an new bean definition based on the args provided: there is no way in a generated code to get access to those suppliers.
Do I miss anything?
If you want I can verify that a bit later.
Thanks

garyrussell: I understand that, but why bother using a supplier for the regular JVM when just the class and ctor arg is needed?

chrylis:
Thanks for this! We'll be stuck on 5.x for a while, but I can continue using the proof-of-concept I have until we overhaul to Java 17.

PR #3930, opened 2022-10-31
Add observation to inbound endpoints

artembilan:

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegrationManagement.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegrationManagement.java
@@ -62,4 +62,12 @@ public @interface EnableIntegrationManagement {
 	 */
 	String defaultLoggingEnabled() default "true";
 
+	/**
+	 * Set simple pattern component names matching for observation registry injection.
+	 * @return simple pattern component names matching for observation registry injection.
+	 * @since 6.0
+	 * @see org.springframework.integration.support.utils.PatternMatchUtils#smartMatch(String, String...)
+	 */
+	String[] observationPatterns() default "*";
+
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegrationManagement.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegrationManagement.java
@@ -54,7 +54,7 @@ public @interface EnableIntegrationManagement {
 	 * <p>
 	 * Set this to false for disabling logging by default in all framework components that implement
 	 * {@link org.springframework.integration.support.management.IntegrationManagement}
-	 * (channels, message handlers etc). It turns off logging such as "PreSend on channel", "Received message" etc.
+	 * (channels, message handlers etc.). It turns off logging such as "PreSend on channel", "Received message" etc.
 	 * <p>
 	 * After the context is initialized, individual components can have their setting changed by invoking
 	 * {@link org.springframework.integration.support.management.IntegrationManagement#setLoggingEnabled(boolean)}.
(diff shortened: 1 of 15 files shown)


garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegrationManagement.java
```diff
@@ -62,4 +62,12 @@
 	 */
 	String defaultLoggingEnabled() default "true";
 
+	/**
+	 * Set simple pattern component names matching for observation registry injection.
+	 * @return simple pattern component names matching for observation registry injection.
+	 * @since 6.0
+	 * @see org.springframework.integration.support.utils.PatternMatchUtils#smartMatch(String, String...)
+	 */
+	String[] observationPatterns() default "*";
```
garyrussell, line 71: I think I would make the default `""` - observation is quite expensive so I would suggest that users should opt-in.

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
```diff
@@ -80,6 +83,8 @@
 
 	private ObjectProvider<ObservationRegistry> observationRegistryProvider;
 
+	private String[] observationPatterns = { "*" };
```
garyrussell, line 86: ditto

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegrationManagement.java
artembilan: Well, you opted-in, when you have added this `@EnableIntegrationManagement` into your configuration.
If not -there is no any instrumentation at all.
So, in my feeling if you have added such a marker, then you probably would expect everything to be instrumented.
Otherwise it sounds double: add annotation and then add an `*` pattern to enable all 🤷

garyrussell: I doubt whether anyone would ever want to instrument every component in an integration flow with observation. I would expect that, in most cases, they would only enable it on inbound and outbound endpoints.

Also, people might want simple timers (metrics) without the extra overhead of observation. Using observation for just timers is less performant.

src/reference/asciidoc/metrics.adoc
```diff
@@ -150,7 +150,13 @@ registry.config().meterFilter(MeterFilter.deny(id ->
 
 Starting with version 6.0, Spring Integration utilizes a Micrometer Observation abstraction which can handle metrics as well as https://micrometer.io/docs/tracing[tracing] via appropriate `ObservationHandler` configuration.
 
-The observation handling is enabled on the `IntegrationManagement` components whenever an `ObservationRegistry` bean is present in the application context.
+The observation handling is enabled on the `IntegrationManagement` components whenever an `ObservationRegistry` bean is present in the application context and an `@EnableIntegrationManagement` is configured.
+To customize what set of components should be instrumented, an `observationPatterns()` attribute is exposed on the `@EnableIntegrationManagement` annotation.
+See its javadocs for a pattern matching algorithm.
+
+IMPORTANT: By default, none of the `IntegrationManagement` components are instrumented with an `ObservationRegistry` bean.
+Can be configured as `*` to match all components.
```
garyrussell, line 158:
```suggestion
Can be configured as `*` to match all components but, typically, patterns to match only inbound and outbound endpoints would be used, to instrument just those components.
```

garyrussell (changes requested):

src/reference/asciidoc/metrics.adoc
artembilan: Why would we say that? 
You know I'm still going to add an observation to `MessageChannel` as well. I guess that one would be as a `PRODUCER` kind.
What objections do you have to not observe channels?
FYI in Spring Cloud Sleuth there was spans processing only on channels though...
Thanks

garyrussell: It will kill performance if all the timers are switched from simple timers to observations.

artembilan: Right. I know what you mean about all those extra objects in messages processing, but I'm still going to expose an observation on channels like we do with timers at the moment.
Just to fulfill the picture and realign with whatever we have so far with SC Sleuth.
As we state now in docs: it is up end-user to decide what components to instrument.
Perhaps they will rely on an observation from Apache Kafka or Rabbit MQ on the input side and on the JDBC or HTTP on the outbound.
The Spring Integration will be configured only for channel.

Also: it doesn't matter where we instrument an observation, the garbage collection noise will be the same.

Therefore I don't see a reason in incorporate your suggestion to this sentence in the docs.

PR #3939, opened 2022-11-07
GH-3938: Fix HTTP XML configuration for ambiguity

### LINKED ISSUE #3938: encoding-mode attribute of http:outbound-gateway is silently ignore if rest-template is present

Version 6.0.0-SNAPSHOT and previous 

**Describe the bug**

In http:outbound-gateway, when "encoding-mode" and "rest-template" attributes are both present, "encoding-mode" is silently ignored

**To Reproduce**

Create an http:output-gateway with both attributes "rest-template" and "encoding-mode" and set, for example, "encoding-mode" to None to discover that the URL is nonetheless encoded.

**Expected behavior**

Get an error from method "HttpAdapterParsingUtils.verifyNoRestTemplateAttributes". This is as simple as adding "encoding-mode" to the list of attributes in the class member SYNC_REST_TEMPLATE_REFERENCE_ATTRIBUTES

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3938

The `encoding-mode` is a property of the `RestTemplate`. Therefore, it cannot be set on the component configuration together with an externally injected `rest-template`

Even if `HttpRequestExecutingMessageHandler` has an assertion for such an ambiguity, the XML parser just ignores this `encoding-mode` when it encounters the `rest-template`

* Fix `HttpAdapterParsingUtils.verifyNoRestTemplateAttributes()` to check for not allowed attributes with a `encoding-mode` as well
* Remove a `default` from the `encoding-mode` to not cause an ambiguity in the parser
* Fix some typos in the `spring-integration-http.xsd`
* Rework `OutboundResponseTypeTests` to JUnit 5

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpAdapterParsingUtils.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpAdapterParsingUtils.java
@@ -42,7 +42,7 @@ import org.springframework.util.xml.DomUtils;
 final class HttpAdapterParsingUtils {
 
 	static final String[] SYNC_REST_TEMPLATE_REFERENCE_ATTRIBUTES = {
-			"request-factory", "error-handler", "message-converters", "encoding-mode"
+			"request-factory", "error-handler", "message-converters"
 	};
 
 	static void verifyNoRestTemplateAttributes(Element element, ParserContext parserContext) {
@@ -52,6 +52,12 @@ final class HttpAdapterParsingUtils {
 								+ attributeName + "' attribute is not allowed.",
 						parserContext.extractSource(element));
 			}
+		}
+
+		if (element.hasAttribute("encoding-mode")) {
+			parserContext.getReaderContext().error("When providing a 'rest-template' reference, " +
+							"the 'encoding-mode' must be set on the 'RestTemplate.uriTemplateHandler' property.",
+					parserContext.extractSource(element));
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpAdapterParsingUtils.java
```diff
@@ -36,12 +36,13 @@
...
  * @author Shiliang Li
+ *
  * @since 2.0.2
  */
-abstract class HttpAdapterParsingUtils {
+final class HttpAdapterParsingUtils {
 
 	static final String[] SYNC_REST_TEMPLATE_REFERENCE_ATTRIBUTES = {
-			"request-factory", "error-handler", "message-converters"
+			"request-factory", "error-handler", "message-converters", "encoding-mode"
```
garyrussell, line 45: Looks like this will break the `for` loops that use this constant (gateway and adapter parsers). (Set reference instead of value). We don't appear to have test coverage for this condition.

I also think we need to improve the error message to say that this property must be set on the template's `UriBuilderFactory`, whereas the other 3 are set on the template itself.

garyrussell (changes requested):

garyrussell:
Could we also improve

```java
parserContext.getReaderContext().error("When providing a 'rest-template' reference, the '"
				+ attributeName + "' attribute is not allowed.",
		parserContext.extractSource(element));
```
to

```java
parserContext.getReaderContext().error("When providing a 'rest-template' reference, the '"
				+ attributeName + "' attribute is not allowed, it must be set on the provided template instead",
		parserContext.extractSource(element));
```

??

garyrussell:
... and cherry-picked as e4e80419c5

PR #3948, opened 2022-11-15
GH-3946: Revise Router channelKeyFallback option

### LINKED ISSUE #3946: Spring DSL defaultOutputToParentFlow does not behave as described
5.1.4.RELEASE

**Describe the bug**

When creating an IntegrationFlow bean using `IntegrationFlows.from(CHANNEL)`, in the `route(Function<S, T> router, Consumer<RouterSpec<T, MethodInvokingRouter>> routerConfigurer)` method, when configuring the router mappings, the method `defaultOutputToParentFlow()` does not what the documentation says it does.

Example:

Given this Integration Flow definition:
```
 @Bean
    public IntegrationFlow orderIntegrationFlow() {
        return IntegrationFlows.from(CHANNEL)
                .<Order, OrderType>route(Order::getType,
                        r -> r.channelMapping(OrderType.DINE_IN, DineInOrderFlow.CHANNEL)
                                .channelMapping(OrderType.ONLINE, OnlineOrderFlow.CHANNEL)
                                .defaultOutputToParentFlow()
                )
                .log("It only reaches here if Order::getType is null.")
                .get();
    }

```

The log will only get printed if `Order::getType` is null. This is contrary to the javadoc which says that:

```
Make a default output mapping of the router to the parent flow. Use the next, after router, parent flow MessageChannel as a AbstractMessageRouter.setDefaultOutputChannel(MessageChannel) of this router.
```

**To Reproduce**

Create a Spring project in Java 11, using Spring 5.3.20 and Spring Integration 5.1.4.RELEASE .

Create an IntegrationFlow bean using the default `IntegrationFlows.from()` method:

```
    public static final String CHANNEL = "orderInputChannel";

    @Bean(name = CHANNEL)
    public MessageChannel getChannel() {
        DirectChannel dc = new DirectChannel();
        dc.setComponentName(CHANNEL);
        return dc;
    }


    @Bean
    public IntegrationFlow orderIntegrationFlow() {
        return IntegrationFlows.from(CHANNEL)
                .<Order, OrderType>route(Order::getType,
                        r -> r.channelMapping(OrderType.DINE_IN, DineInOrderFlow.CHANNEL)
                                .channelMapping(OrderType.ONLINE, OnlineOrderFlow.CHANNEL)
                                .defaultOutputToParentFlow()
                )
                .log("It only reaches here if Order::getType is null.")
                .get();
    }
```

Create a class called Order:

```
public class Order {

    private OrderType orderType;
    public OrderType getType() {
        return orderType;
    }

    public void setType(OrderType orderType) {
        this.orderType = orderType;
    }

}
```

And an enum called OrderType:

```
public enum OrderType {
    DINE_IN,
    ONLINE,
    TAKEAWAY;
}
```

And with everything setup correctly, send a message to CHANNEL of an `Order` with `OrderType.TAKEAWAY`.

**Expected behavior**

The `log()` at the end of the `orderIntegrationFlow` will be called because the `defaultOutputToParent` reached a default state.

** Actual behavior **

An Exception is thrown where TAKEAWAY is not found to be a proper CHANNEL name defined in the Spring Context. Only if Order::getOrderType returns null, then the `log()` function is called.

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3946

The `AbstractMappingMessageRouter` has both `resolutionRequired` and `channelKeyFallback` as `true` by default.
End-users expects them to back off when they set a `defaultOutputChannel`. They really want something similar to Java `switch` statement

* Change the logic in the `AbstractMappingMessageRouter` to reset `channelKeyFallback` to `false` when `defaultOutputChannel` to avoid attempts to resolve channel from name, but rather fallback to `defaultOutputChannel` as it states from th mentioned Java `switch` statement experience
* Deprecate `RouterSpec.noChannelKeyFallback()` in favor of newly introduced `channelKeyFallback(boolean)`
* Call `channelKeyFallback(false)` from an overloaded `defaultOutputToParentFlow()` to reflect the mentioned expected behavior in Java DSL as well.
* Respectively, deprecate `KotlinRouterSpec.noChannelKeyFallback()` wrapper in favor of newly introduced `channelKeyFallback(channelKeyFallback: Boolean)`
* Remove redundant already `noChannelKeyFallback()` option in the `NoFallbackAllowedTests`
* Document the change and new behavior

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
@@ -226,7 +226,7 @@ public final class RouterSpec<K, R extends AbstractMappingMessageRouter>
 	 * Use the next, after router, parent flow {@link MessageChannel} as a
 	 * {@link AbstractMessageRouter#setDefaultOutputChannel(MessageChannel)} of this router.
 	 * This option also disables {@link AbstractMappingMessageRouter#setChannelKeyFallback(boolean)},
-	 * if not called explicitly, to skip an attempt to resolve channel name.
+	 * if not called explicitly afterwards, to skip an attempt to resolve the channel name.
 	 * @return the router spec.
 	 * @since 6.0
 	 */
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 the original author or authors.
+ * Copyright 2016-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.integration.context.IntegrationObjectSupport;
 import org.springframework.integration.router.AbstractMappingMessageRouter;
+import org.springframework.integration.router.AbstractMessageRouter;
 import org.springframework.integration.support.context.NamedComponent;
 import org.springframework.integration.support.management.MappingMessageRouterManagement;
 import org.springframework.messaging.MessageChannel;
@@ -118,9 +119,24 @@ public final class RouterSpec<K, R extends AbstractMappingMessageRouter>
 	 * cause a stack overflow.
 	 * @return the router spec.
 	 * @since 5.2
+	 * @deprecated since 6.0 in favor of {@link  #channelKeyFallback(boolean)}
 	 */
+	@Deprecated(since = "6.0", forRemoval = true)
 	public RouterSpec<K, R> noChannelKeyFallback() {
-		this.handler.setChannelKeyFallback(false);
+		return channelKeyFallback(false);
+	}
+
+	/**
+	 * When true (default), if a resolved channel key does not exist in the channel map,
+	 * the key itself is used as the channel name, which we will attempt to resolve to a
+	 * channel. Set to {@code false} to disable this feature.
+	 * @param channelKeyFallback false to disable the fallback.
+	 * @return the router spec.
+	 * @since 6.0
+	 * @see AbstractMappingMessageRouter#setChannelKeyFallback(boolean)
+	 */
+	public RouterSpec<K, R> channelKeyFallback(boolean channelKeyFallback) {
+		this.handler.setChannelKeyFallback(channelKeyFallback);
 		return _this();
 	}
 
@@ -203,6 +219,20 @@ public final class RouterSpec<K, R extends AbstractMappingMessageRouter>
 
 		this.mappingProvider.addMapping(key, (NamedComponent) channel);
 		return _this();
+	}
+
+	/**
+	 * Make a default output mapping of the router to the parent flow.
+	 * Use the next, after router, parent flow {@link MessageChannel} as a
+	 * {@link AbstractMessageRouter#setDefaultOutputChannel(MessageChannel)} of this router.
+	 * This option also disables {@link AbstractMappingMessageRouter#setChannelKeyFallback(boolean)},
+	 * if not called explicitly, to skip an attempt to resolve channel name.
+	 * @return the router spec.
+	 * @since 6.0
+	 */
+	public RouterSpec<K, R> defaultOutputToParentFlow() {
+		return super.defaultOutputToParentFlow()
+				.channelKeyFallback(false);
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
@@ -251,7 +251,7 @@ public abstract class AbstractMappingMessageRouter extends AbstractMessageRouter
 	@Override
 	protected void onInit() {
 		super.onInit();
-		Assert.state(this.channelKeyFallback && this.resolutionRequired && this.defaultOutputChannelSet,
+		Assert.state(!this.channelKeyFallback || !this.resolutionRequired || !this.defaultOutputChannelSet,
 				"The 'defaultOutputChannel' cannot be reached " +
 						"when both 'channelKeyFallback' & 'resolutionRequired' are set to true. " +
 						"See their javadocs for more information.");
--- a/spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
@@ -79,7 +79,9 @@ public abstract class AbstractMappingMessageRouter extends AbstractMessageRouter
 
 	private boolean channelKeyFallback = true;
 
-	private boolean isChannelKeyFallbackSetExplicitly;
+	private boolean defaultOutputChannelSet;
+
+	private boolean channelKeyFallbackSetExplicitly;
 
 	private volatile Map<String, String> channelMappings = new LinkedHashMap<>();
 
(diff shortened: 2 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
```diff
@@ -205,6 +221,20 @@ public RouterSpec<K, R> subFlowMapping(K key, IntegrationFlow subFlow) {
 		return _this();
 	}
 
+	/**
+	 * Make a default output mapping of the router to the parent flow.
+	 * Use the next, after router, parent flow {@link MessageChannel} as a
+	 * {@link AbstractMessageRouter#setDefaultOutputChannel(MessageChannel)} of this router.
+	 * This option also disables {@link AbstractMappingMessageRouter#setChannelKeyFallback(boolean)},
+	 * if not called explicitly, to skip an attempt to resolve channel name.
```
garyrussell, line 229:
```suggestion
	 * if not called explicitly afterwards, to skip an attempt to resolve the channel name.
```

spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
```diff
@@ -124,15 +128,70 @@ public void setResolutionRequired(boolean resolutionRequired) {
...
 	public void setChannelKeyFallback(boolean channelKeyFallback) {
 		this.channelKeyFallback = channelKeyFallback;
+		this.channelKeyFallbackSetExplicitly = true;
+	}
+
+	/**
+	 * Set the default channel where Messages should be sent if channel resolution
+	 * fails to return any channels.
+	 * It also sets {@link #channelKeyFallback} to {@code false} to avoid
+	 * an attempt to resolve a channel from its key, but instead send message
```
garyrussell, line 147:
```suggestion
	 * an attempt to resolve a channel from its key, but instead send the message
```

spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
```diff
@@ -124,15 +128,70 @@ public void setResolutionRequired(boolean resolutionRequired) {
...
+	 * directly to this channel.
+	 * If {@link #channelKeyFallback} is set explicitly to {@code true},
+	 * the further logic depend on a {@link #resolutionRequired} options
```
garyrussell, line 150:
```suggestion
	 * the logic depends on the {@link #resolutionRequired} option
```

spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
```diff
@@ -124,15 +128,70 @@ public void setResolutionRequired(boolean resolutionRequired) {
...
+	 * ({@code true} by default), and therefore a fallback to
+	 * this default output channel may never happen.
+	 * The configuration where default output channel is present and
+	 * {@link #resolutionRequired} and {@link #resolutionRequired} are set to {@code true}
+	 * is rejected since it leads to ambiguity.
```
garyrussell, line 155:
```suggestion
	 * The configuration where a default output channel is present and
	 * {@link #setResolutionRequired resolutionRequired} and 
	 * {@link #setChannelKeyFallback channelKeyFallback} are set to {@code true}
	 * is rejected since it leads to ambiguity.
```

spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
```diff
@@ -124,15 +128,70 @@ public void setResolutionRequired(boolean resolutionRequired) {
...
```
garyrussell, line 182: Same changes here.

spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
```diff
@@ -189,15 +248,14 @@ public Collection<String> getDynamicChannelNames() {
...
-	protected abstract List<Object> getChannelKeys(Message<?> message);
-
+	@Override
+	protected void onInit() {
+		super.onInit();
+		Assert.state(!this.channelKeyFallback || !this.resolutionRequired || !this.defaultOutputChannelSet,
+				"The 'defaultOutputChannel' cannot be reached " +
+						"when both 'channelKeyFallback' & 'resolutionRequired' are set to true. " +
+						"See their javadocs for more information.");
+	}
```
garyrussell, line 258: I don't think we can enforce this; what if both are true and channel resolution is successful.

src/reference/asciidoc/router.adoc
```diff
@@ -349,7 +349,10 @@ If set, this attribute provides a reference to the channel where messages should
 If no default output channel is provided, the router throws an exception.
 If you would like to silently drop those messages instead, set the default output channel attribute value to `nullChannel`.
 +
-NOTE: A message is sent only to the `default-output-channel` if `resolution-required` is `false` and the channel is not resolved.
+NOTE: Starting with version 6.0, setting a default output channel also resets `channelKeyFallback` option to `false`.
+So, no attempts to resolve a channel from name, but rather fallback to this default output channel - similar to Java `switch` statement.
```
garyrussell, line 353:
```suggestion
NOTE: Starting with version 6.0, setting a default output channel also resets the `channelKeyFallback` option to `false`.
So, no attempts will be made to resolve a channel from its name, but rather fallback to this default output channel - similar to a Java `switch` statement.
```

src/reference/asciidoc/router.adoc
```diff
@@ -349,7 +349,10 @@ If set, this attribute provides a reference to the channel where messages should
...
+If `channelKeyFallback` is set to `true` explicitly, the further logic depends on the `resolutionRequired` option: the message can reach a `defaultOutputChannel` only if `resolutionRequired` is `false`.
+Therefore, a configuration where `defaultOutputChannel` is provided and both `channelKeyFallback` & `resolutionRequired` are set to `true` is rejected by the `AbstractMappingMessageRouter` initialization phase.
```
garyrussell, line 355: We can't enforce this. See comment above.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -110,6 +110,11 @@ For convenience, the XML and Java DSL for Scatter-Gather, based on the `Recipien
 
 See <<./scatter-gather.adoc#scatter-gather,Scatter-Gather>> for more information.
 
+Another convenient behaviour change has been done for an `AbstractMappingMessageRouter`.
+Now setting a `defaultOutputChannel` also resets a `channelKeyFallback` to `false`, so no attempts to resolve channel from its key, but immediate fallback to `defaultOutputChannel`.
```
garyrussell, line 114:
```suggestion
Another convenient behavior change has been made to the `AbstractMappingMessageRouter`.
Now, setting a `defaultOutputChannel` also resets the `channelKeyFallback` property to `false`, so no attempts will be made to resolve a channel from its key, but the logic immediately falls back to sending the mesage to the `defaultOutputChannel`.
```

garyrussell (changes requested):
Mostly docs, but one important.

spring-integration-core/src/main/java/org/springframework/integration/router/AbstractMappingMessageRouter.java
artembilan: Correct. And what is the reason to have a `defaultOutputChannel` then?
You simply can set `channelKeyFallback = true`, but `resolutionRequired = false`.
So, if resolution is OK, we go there, if it is not we fallback to a default channel.

We probably can improve docs for this as well.

artembilan:
I don't think we need to say anything else.
This sentence in docs covers everything:

> If `channelKeyFallback` is set to `true` explicitly, the further logic depends on the `resolutionRequired` option: the message to non-resolved channel from key can reach a `defaultOutputChannel` only if `resolutionRequired` is `false`.

artembilan:
My point  with this change is that end-users expects a behavior a-la Java `swtich` expression.
So, they don't consider a resolution from key.
Kinda let's make a default behavior as obvious as possible 😄

garyrussell (approved):

Issue #3869, opened 2022-08-04
Implement ContextHolderRequestHandlerAdvice


artembilan:
Sometimes we may have a context during message handling. See a `DelegatingSessionFactory` as an example. So, we store some `key` into the thread local, then `SftpMessageHandler` obtain the session and and performs respective logic.
All this is done in the `SftpMessageHandler` transparently, but in case of `DelegatingSessionFactory` we just chose the target (S)FTP connection according previously mentioned `key` stored in the thread local.

See more info in [SO question](https://stackoverflow.com/questions/73229125/spring-integration-sftpoutbound-channel-adapter-dynamic-sftpsessionfactory) and in [docs](https://docs.spring.io/spring-integration/docs/current/reference/html/messaging-endpoints.html#message-handler-advice-chain).

So, current solution for the context holder is a bit involved, especially when we use one-way channel adapter for (S)FTP.

The proposed `ContextHolderRequestHandlerAdvice` could be a single way to inject into such an outbound endpoint.
Its logic is pretty simple: 

1. Take a key from the request message
2. Store the value into the context holder
3. Invoke the `MessageHandler`
4. Clean the value from the context holder in the `finally`

I feel like this advice can have these options:

1. `keyProvider` - the `Function<Message<?>, Object>` to obtain the key from the message, or whatever source for the key is available. Could be as simple as `setKeyProvider((m) -> m.getHeaders().get("myKey"));`
2. `contextSetHook` - a `Consumer<Object>` - just delegates to the method on the target context holder. For example in case of the mentioned `DelegatingSessionFactory` it could be like this `setContextSetHook(myDelegatingFactory::setThreadKey);`
3. `contextClearHook` - a `Consumer<Object>` - just delegates to the method on the target context holder.For example in case of the mentioned `DelegatingSessionFactory` it could be like this `setContextClearHook((key) -> myDelegatingFactory.clearThreadKey());`

Probably internal logic could be implemented as a `try-with-resource`.

PR #3960, opened 2022-12-04
GH-3869: Create The Context Holder Request Handler Advice.

adel-haidar:
GH-3869: Add unit tests.

GH-3869: Create Advice.

GH-3869: Correct typo.

Issue:  Implement ContextHolderRequestHandlerAdvice #3869

This is my first open source PR ever. I therefore would like to apologise in advance for any stupid mistake I might have made during the process.
Your Feedback is much appreciated!


adel-haidar:
When running `gradle clean testAll` on my mac the following test always fails: `testOs`.

- OS: MacOS Ventura
- IDE: IntelliJ IDEA Community Edition
-  Gradle version: 7.5.1

artembilan:
Thank you for contribution!

This is a new feature and we will look adding it into the next `6.1` version.

Please, ignore that failing test: it has nothing to do with your changes.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.springframework.messaging.Message;
+
+/**
+ * Provides a key for the context holder.This key could for example be stored in a message header.
+ *
+ * @author Adel Haidar
+ * @since 6.0
+ */
+public class ContextHolderRequestHandlerAdvice {
```
artembilan, line 30: You are missing an implementation of `AbstractRequestHandlerAdvice`.
Otherwise this class have no value.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.integration.support.MessageBuilder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+public class ContextHolderRequestHandlerAdviceTests {
```
artembilan, line 30: This test class does not bring any value.
Or probably better to say: it is just redundant and can be covered by other more realistic tests where we really apply this new advice on endpoints.
See docs for more info: https://docs.spring.io/spring-integration/docs/current/reference/html/messaging-endpoints.html#message-handler-advice-chain

artembilan (changes requested):

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
```diff
@@ -0,0 +1,131 @@
...
+		given(foo.mockSession.list(anyString()))
+				.willReturn(new String[0]);
+		final Map<String, Object> headers = Map.of("foo", "foo");
+		this.in.send(new GenericMessage<>("foo", headers));
+		verify(foo.mockSession).list("foo/");
+	}
+
+	@Configuration
+	@ImportResource(
+			"classpath:/org/springframework/integration/file/remote/session/context-holder-request-handler-advice-context.xml")
```
artembilan, line 78: May we just don't do any XML configs, please?
See a `@ServiceActivator` and its `advice` property instead: https://docs.spring.io/spring-integration/docs/current/reference/html/configuration.html#annotations

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
```diff
@@ -0,0 +1,131 @@
...
+	@Autowired
+	DefaultSessionFactoryLocator<String> sessionFactoryLocator;
+
+	@Test
+	public void testFlow() throws IOException {
+		given(foo.mockSession.list(anyString()))
+				.willReturn(new String[0]);
+		final Map<String, Object> headers = Map.of("foo", "foo");
+		this.in.send(new GenericMessage<>("foo", headers));
+		verify(foo.mockSession).list("foo/");
```
artembilan, line 73: I don't believe into this test since there is nowhere the code is interact with the `Session`.

artembilan (changes requested):

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
adel-haidar: I replaced the tests with a new one based on what I understood from the documentation. But I cannot get it to work. Although I added the advice to a service activator, created a delegatingSessionFactory and set the functions in the new advice as shown in the example. But, I am still getting the following error: 

```
Wanted but not invoked:
session.list("foo/");
-> at org.springframework.integration.file.remote.session.ContextHolderRequestHandlerAdviceTests.testDelegates(ContextHolderRequestHandlerAdviceTests.java:73)
Actually, there were zero interactions with this mock.
```

I know the test still does not look good, many parts are unused (and maybe unneeded).

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
adel-haidar: Happy to. I prefer java over xml 😄

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,117 @@
...
+		applyKeyProvider(message);
+		return result;
+	}
+
+	private void applyKeyProvider(Message<?> message) {
+		if (this.keyProvider != null) {
+			final Object key = this.keyProvider.apply(message);
+			if (key != null) {
+				setContext(key);
+				clearContext(key);
```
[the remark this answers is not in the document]
adel-haidar: How did you imagine the clearContext should be called. You wrote in the ticket description that it could be called in the finally section of a try-with-resources block. However, I don't see how this could be implemented.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
adel-haidar: I will try to rework the test 😄

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,127 @@
...
+import org.springframework.messaging.Message;
+
+/**
+ * Provides a key for the context holder.This key could for example be stored in a message header.
+ *
+ * @author Adel Haidar
+ * @since 6.1
+ */
+public class ContextHolderRequestHandlerAdvice extends AbstractRequestHandlerAdvice {
+	public Function<Message<?>, Object> keyProvider;
```
artembilan, line 32: Every member in a class must be surrounded with blank line - better code readability.
Also: all of these properties must be marked with a `final`.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,127 @@
...
+	 */
+	public Consumer<Object> getContextClearHook() {
+		return this.contextClearHook;
+	}
+
+	/**
+	 * Sets the context clear hook function.
+	 * @param contextClearHook the context clear hook function.
+	 */
+	public void setContextClearHook(Consumer<Object> contextClearHook) {
```
artembilan, line 97: Since we configure this class via constructor I don't see a reason in setters.
Also it doesn't look like there is a reason in getters, too...

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,127 @@
...
+	protected Object doInvoke(ExecutionCallback callback, Object target, Message<?> message) {
+		Object result;
+		final Object key = this.keyProvider.apply(message);
+		try {
+			setContext(key);
+			result = callback.execute();
+		}
+		catch (Exception e) {
+			result = e;
+			logger.error("Failure setting context for " + message + ": " + e.getMessage());
```
artembilan, line 111: That's not correct: the exception might be as a result of the `callback.execute()`.
I believe that we must not catch any exception over here at all and just let them to be propagated upstream as is.
Kinda not this component responsibility.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,127 @@
...
+		}
+		return result;
+	}
+
+	private void setContext(Object key) {
+		this.contextSetHook.accept(key);
+	}
+
+	private void contextClearHook(Object key) {
+		this.contextClearHook.accept(key);
```
artembilan, line 124: I don't see reasons in these single-line methods.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright 2015-2022 the original author or authors.
```
artembilan, line 2: Just current year: this is a new class and apparently it will make it into the `main` already in the current `2023`

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
```diff
@@ -0,0 +1,152 @@
...
+	Config config;
+
+	@Test
+	public void testFlow() {
+		final Map<String, Object> headers = Map.of("test-key", "foo");
+		config.in().send(new GenericMessage<>("any-payload", headers));
+		Message<?> received = config.out().receive(0);
+		assertThat(received).isNotNull();
+		assertThat(Config.testSessionDuringHandling).isNotNull().isEqualTo(foo.getSession());
+		assertThat(config.dsf().getSession()).isEqualTo(bar.getSession());
```
artembilan, line 67: I think this test is still meaningless and configuration is too complicated.

I believe we simple could test against some `ThreadLocal` used in the `@ServiceActivator` method with such an advice.
And then in the test we ensure that the value from that `ThreadLocal` has been accessed from that method.
Plus in the end verify that `ThreadLocal` is cleared according to the `ContextHolderRequestHandlerAdvice` expectations.

Then `DelegatingSessionFactoryTests` might be modified for this `ContextHolderRequestHandlerAdvice` instead of those `<int:service-activator>`.
And perhaps the whole XML config could be removed altogether since the logic for context holder is supported via `ContextHolderRequestHandlerAdvice` now.

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,70 @@
...
+		this.contextClearHook = contextClearHook;
+	}
+
+	@Override
+	protected Object doInvoke(ExecutionCallback callback, Object target, Message<?> message) {
+		Object result;
+		final Object key = this.keyProvider.apply(message);
+		try {
+			this.contextSetHook.accept(key);
+			result = callback.execute();
```
artembilan, line 62: I think we can just return from here without any extra local variable.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,70 @@
...
+
+	public final Function<Message<?>, Object> keyProvider;
+
+	public final Consumer<Object> contextSetHook;
+
+	public final Consumer<Object> contextClearHook;
+
+	/**
+	 * Constructor for the ContextHolderRequestHandlerAdvice.
+	 *
```
artembilan, line 42: We prefer do not have blank line in the method Javadocs.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.springframework.integration.handler.advice.AbstractRequestHandlerAdvice;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * Provides a key for the context holder.This key could for example be stored in a message header.
```
artembilan, line 27: New sentence on a new line.
Either way no whitespace between period and a new sentence is wrong.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,70 @@
...
+		Assert.notNull(contextClearHook, "'contextClearHook' must not be null");
+		this.keyProvider = keyProvider;
+		this.contextSetHook = contextSetHook;
+		this.contextClearHook = contextClearHook;
+	}
+
+	@Override
+	protected Object doInvoke(ExecutionCallback callback, Object target, Message<?> message) {
+		Object result;
+		final Object key = this.keyProvider.apply(message);
```
artembilan, line 59: How about to `logger.trace()` this key for this message?

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
artembilan: So, you have agreed with me about this test class problem, but you did nothing to fix it.

Are you still going?

artembilan (changes requested):
Any chances also to document this feature in the `handler-advice.adoc` ?
Also, please, consider to issue your PR against `6.1-WIP` branch.
I hope we are going to release `M1` in the end of this month.
Your contribution is a good candidate to make it there.

Thank you!

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
adel-haidar: I am sorry, I haven't got the chance to properly work on this yet. That's why I just pushed the easy stuff and left this for later. I honestly didn't quite understand what you mean. But I will try anyway.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
adel-haidar: Done

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
adel-haidar: Done

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
adel-haidar: Removed the blank line.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
adel-haidar: Done

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
adel-haidar: I don't understand what you meant by the following:

> I believe we simple could test against some ThreadLocal used in the @ServiceActivator method with such an advice.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdvice.java
```diff
@@ -0,0 +1,68 @@
...
+		Assert.notNull(contextClearHook, "'contextClearHook' must not be null");
+		this.keyProvider = keyProvider;
+		this.contextSetHook = contextSetHook;
+		this.contextClearHook = contextClearHook;
+	}
+
+	@Override
+	protected Object doInvoke(ExecutionCallback callback, Object target, Message<?> message) {
+		final Object key = this.keyProvider.apply(message);
+		logger.trace("Setting context key to: " + key + " message: " + message);
```
artembilan, line 58: It has to start with `()-> ` to make the string build lazy.
Also missed `for message`

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -57,6 +57,9 @@ In addition to providing the general mechanism to apply AOP advice classes, Spri
 * `RateLimiterRequestHandlerAdvice` (described in <<rate-limiter-advice>>)
 * `CacheRequestHandlerAdvice` (described in <<cache-advice>>)
 * `ReactiveRequestHandlerAdvice` (described in <<reactive-advice>>)
+* `ContextHolderRequestHandlerAdvice` (described in <<context-holder-advice>>)
+
+[[expression-advice]]
```
artembilan, line 62: I think this is some copy-paste artifact.

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -571,6 +574,29 @@ The `message` argument is the request message for the message handler and can be
 The `mono` argument is the result of this message handler's `handleRequestMessage()` method implementation.
 A nested `Mono.transform()` can also be called from this function to apply, for example, a https://spring.io/projects/spring-cloud-circuitbreaker[Reactive Circuit Breaker].
 
+[[context-holder-advice]]
+==== Context Holder Advice
+Starting with version 6.1, the `ContextHolderRequestHandlerAdvice` has been introduced.
```
artembilan, line 579: New line after the head line.

src/reference/asciidoc/handler-advice.adoc
```diff
@@ -571,6 +574,29 @@ The `message` argument is the request message for the message handler and can be
...
+    final DefaultSessionFactoryLocator<String> sff = new DefaultSessionFactoryLocator<>(factories, foo);
+    final DelegatingSessionFactory<String> dsf = new DelegatingSessionFactory<>(sff);
+    final Function<Message<?>, Object> keyProviderFn = m -> m.getHeaders().get("<KEY_NAME>");
+    final Consumer<Object> contextSetHookFn = dsf::setThreadKey;
+    final Consumer<Object> contextClearHookFn = key -> dsf.clearThreadKey();
+    return new ContextHolderRequestHandlerAdvice(keyProviderFn, contextSetHookFn, contextClearHookFn);
+}
+----
+====
 [[custom-advice]]
```
artembilan, line 600: New line after code snippet.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/session/ContextHolderRequestHandlerAdviceTests.java
```diff
@@ -0,0 +1,148 @@
...
+
+	@Autowired
+	Config config;
+
+	@Test
+	public void testFlow() {
+		final Map<String, Object> headers = Map.of("test-key", "foo");
+		config.in().send(new GenericMessage<>("any-payload", headers));
+
+		Message<?> received = config.out().receive(0);
```
artembilan, line 66: This test is going to pass even without `ContextHolderRequestHandlerAdvice`.
You just really don't demonstrate a usage of this feature in that `TestService`.
Even if that `foo` is selected properly from the `DelegatingSessionFactory`, there is no proof anywhere in the code that we indeed have taken a `foo` `ConnectionFactory`.

Pay attention that `DelegatingSessionFactoryTests` you took ideas from has that `verify(foo.mockSession).list("foo/");` with the meaning that target handler really interacted with mocked `foo` `SessionFactory`.

Your current test does not have that logic.

artembilan (changes requested):

artembilan:
image


Oh! Sorry, I didn't think it is going to close a PR against that branch.
If you still going to contribute, please, consider to raise a new PR against latest `main` already.
Otherwise we can take your work as is and merge it some other way.

Thank you for understanding!

artembilan:
So, this one has been merged as https://github.com/spring-projects/spring-integration/commit/bfc6931329fee651cc3b398156642057da2693ed after significant rework and clean up.
Especially tests and docs.

Anyway, thank you for the contribution, @adel-haidar !

PR #3961, opened 2022-12-06
GH-3959: MqttConFailedEvent for normal disconnect

### LINKED ISSUE #3959: MQTTv5: Disconnect message does not trigger

sprint-boot 2.7.5
spring-integration-mqtt 5.5.15
org.eclipse.paho.mqttv5.client 1.2.5

**Describe the bug**

If an MQTTv5 broker disconnects a client via a disconnect message then MqttConnectionFailedEvent is never triggered.

**To Reproduce**

```
    @EventListener
    public void handleFailedEvent(MqttConnectionFailedEvent mfe) {
        System.out.println("MqttConnectionFailedEvent "+mfe.getCause().getMessage());
    }

```
and trigger a disconnect event from the broker.

**Expected behavior**

MqttConnectionFailedEvent is called.

**Details**
The problem occurs because

```
org.eclipse.paho.mqttv5.client.internal.CommsCallback:

	public void connectionLost(MqttException cause, MqttDisconnect message) {
		final String methodName = "connectionLost";
		// If there was a problem and a client callback has been set inform
		// the connection lost listener of the problem.
		try {
			if (mqttCallback != null && message != null) {
				MqttDisconnectResponse disconnectResponse = new MqttDisconnectResponse(message.getReturnCode(),
						message.getProperties().getReasonString(),
						(ArrayList<UserProperty>) message.getProperties().getUserProperties(),
						message.getProperties().getServerReference());
				mqttCallback.disconnected(disconnectResponse);
```
this constructur does not set the cause as exception in MqttDisconnectResponse. 

```
	public MqttDisconnectResponse(int returnCode, String reasonString, ArrayList<UserProperty> userProperties,
			String serverReference) {
		this.returnCode = returnCode;
		this.reasonString = reasonString;
		this.userProperties = userProperties;
		this.serverReference = serverReference;
	}

```
this disconnectResponse eventually triggers disconnected in Mqttv5PahoMessageDrivenChannelAdapter. The check cause != null is false (because cause == null) and thus the event is never publisher.

```
    public void disconnected(MqttDisconnectResponse disconnectResponse) {
        MqttException cause = disconnectResponse.getException();
        ApplicationEventPublisher applicationEventPublisher = getApplicationEventPublisher();
        if (cause != null && applicationEventPublisher != null) {
            applicationEventPublisher.publishEvent(new MqttConnectionFailedEvent(this, cause));
        }
    }
```

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3959

The `MqttCallback.disconnected(MqttDisconnectResponse)` in Paho v5 client is also called when server initiates a disconnection.
In this case that `MqttDisconnectResponse` does not have a `cause` value

* Modify `MqttConnectionFailedEvent` to make a `cause` property optional
* Fix `Mqttv5PahoMessageDrivenChannelAdapter` & `Mqttv5PahoMessageHandler` to not check for `cause`, but emit an `MqttConnectionFailedEvent` for any `disconnected()` calls

Unfortunately current Paho v3 client does not call `connectionLost()` for normal disconnections and we cannot react for this callback with an `MqttConnectionFailedEvent`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/src/reference/asciidoc/mqtt.adoc
+++ b/src/reference/asciidoc/mqtt.adoc
@@ -403,6 +403,7 @@ public class MqttJavaApplication {
 Certain application events are published by the adapters.
 
 * `MqttConnectionFailedEvent` - published by both adapters if we fail to connect or a connection is subsequently lost.
+For Paho client for MQTT v5 this event is also emitted when server performs a normal disconnection, in which case the `cause` of lost connection is `null`.
 * `MqttMessageSentEvent` - published by the outbound adapter when a message has been sent, if running in asynchronous mode.
 * `MqttMessageDeliveredEvent` - published by the outbound adapter when the client indicates that a message has been delivered, if running in asynchronous mode.
 * `MqttSubscribedEvent` - published by the inbound adapter after subscribing to the topics.
--- a/src/reference/asciidoc/mqtt.adoc
+++ b/src/reference/asciidoc/mqtt.adoc
@@ -403,7 +403,7 @@ public class MqttJavaApplication {
 Certain application events are published by the adapters.
 
 * `MqttConnectionFailedEvent` - published by both adapters if we fail to connect or a connection is subsequently lost.
-For Paho client for MQTT v5 this event is also emitted when server performs a normal disconnection, in which case the `cause` of lost connection is `null`.
+For the MQTT v5 Paho client, this event is also emitted when the server performs a normal disconnection, in which case the `cause` of the lost connection is `null`.
 * `MqttMessageSentEvent` - published by the outbound adapter when a message has been sent, if running in asynchronous mode.
 * `MqttMessageDeliveredEvent` - published by the outbound adapter when the client indicates that a message has been delivered, if running in asynchronous mode.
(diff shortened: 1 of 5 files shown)


src/reference/asciidoc/mqtt.adoc
```diff
@@ -403,6 +403,7 @@ public class MqttJavaApplication {
 Certain application events are published by the adapters.
 
 * `MqttConnectionFailedEvent` - published by both adapters if we fail to connect or a connection is subsequently lost.
+For Paho client for MQTT v5 this event is also emitted when server performs a normal disconnection, in which case the `cause` of lost connection is `null`.
```
garyrussell, line 406:
```suggestion
For the for MQTT v5 Paho client, this event is also emitted when the server performs a normal disconnection, in which case the `cause` of the lost connection is `null`.
```

garyrussell (changes requested):

garyrussell:
...and cherry-picked as 280eb29f62

PR #3967, opened 2022-12-09
Obtain observed timers in the Integration Graph

artembilan:
The `Observation` populates slightly different Micrometer timer names and their tags

* Fix `MicrometerNodeEnhancer` to search timers in the registry according to the `Observation` convention if the provided component is observed
* For that purpose expose an `IntegrationManagement.isObserved()` option to check of component is instrumented with an `Observation`
* Improve the logic in the `IntegrationGraphServer` to delegate into a `Function` to call `micrometerEnhancer.enhance()` instead of `static` property which might not be OK in the environment where several integration applications are ran in the same JVM

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/graph/MicrometerNodeEnhancer.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/graph/MicrometerNodeEnhancer.java
@@ -179,7 +179,7 @@ public class MicrometerNodeEnhancer {
 				(long) (failures == null ? 0 : failures.count()));
 	}
 
-	private static TimerStats buildTimerStats(Timer timer) {
+	private static TimerStats buildTimerStats(@Nullable Timer timer) {
 		return timer == null
 				? ZERO_TIMER_STATS
 				: new TimerStats(timer.count(), timer.mean(TimeUnit.MILLISECONDS), timer.max(TimeUnit.MILLISECONDS));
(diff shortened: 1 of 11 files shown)


spring-integration-core/src/main/java/org/springframework/integration/graph/MicrometerNodeEnhancer.java
```diff
@@ -150,5 +179,11 @@ private <T extends IntegrationNode> ReceiveCounters retrieveCounters(T node, Str
 				(long) (failures == null ? 0 : failures.count()));
 	}
 
+	private static TimerStats buildTimerStats(Timer timer) {
```
garyrussell, line 182:
```suggestion
	private static TimerStats buildTimerStats(@Nullable Timer timer) {
```

spring-integration-core/src/main/java/org/springframework/integration/support/management/IntegrationManagement.java
```diff
@@ -106,6 +106,15 @@ default void registerObservationRegistry(ObservationRegistry observationRegistry
 		// no op
 	}
 
+	/**
+	 * True if this implementation is going to dela with not {@link  ObservationRegistry#NOOP} instance.
+	 * @return true if this implementation is going to dela with not {@link  ObservationRegistry#NOOP} instance.
```
garyrussell, line 111:
```suggestion
	 * True if this implementation is going to deal with a registry other than the {@link  ObservationRegistry#NOOP} instance.
	 * @return true if this implementation is going to deal with a registry other than the {@link  ObservationRegistry#NOOP} instance.
```

garyrussell (changes requested):

garyrussell (approved):

PR #3971, opened 2022-12-12
Make IntegrationObjectSupport#setConversionService public

### LINKED ISSUE #3968: IntegrationObjectSupport.conversionService property has flipped access modifiers

5.5.14

**Describe the bug**

In `IntegrationObjectSupport`, most of the properties have public setters (for external configuration) and protected getters. However, the `conversionService` property has this reversed: The setter is protected but the getter is public. I am wanting to externally set the `conversionService` on a channel adapter (specifically SNS, because I want to customize the JSON formatting but after the adapter has evaluated expressions against the payload POJO), but the property isn't visible unless I subclass it. Based on the other dependency properties, this appears to be a typo in the declarations.

chrylis:
fixes #3968

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
@@ -268,7 +268,7 @@ public abstract class IntegrationObjectSupport implements BeanNameAware, NamedCo
 		return this.conversionService;
 	}
 
-	protected void setConversionService(ConversionService conversionService) {
+	public void setConversionService(ConversionService conversionService) {
 		this.conversionService = conversionService;
 	}
 
--- a/spring-integration-core/src/test/java/org/springframework/integration/router/RouterConcurrencyTest.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/router/RouterConcurrencyTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -57,7 +57,7 @@ public class RouterConcurrencyTest {
 			}
 
 			@Override
-			protected void setConversionService(ConversionService conversionService) {
+			public void setConversionService(ConversionService conversionService) {
 				try {
 					if (count.incrementAndGet() > 1) {
 						Thread.sleep(20);


spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
```diff
@@ -268,7 +268,7 @@ public ConversionService getConversionService() {
 		return this.conversionService;
 	}
 
-	protected void setConversionService(ConversionService conversionService) {
+	public void setConversionService(ConversionService conversionService) {
```
artembilan, line 271: Looks like you didn't run tests in the project after this change.
The `RouterConcurrencyTest` is going to fail.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
chrylis: This is strange. Gradle output a message on the console but didn't record it as an error:
```
> Task :spring-integration-core:compileTestJava
/home/christopher/git/spring/spring-integration/spring-integration-core/src/test/java/org/springframework/integration/router/RouterConcurrencyTest.java:60: error: setConversionService(ConversionService) in <anonymous org.springframework.integration.router.RouterConcurrencyTest$1> cannot override setConversionService(ConversionService) in IntegrationObjectSupport
                        protected void setConversionService(ConversionService conversionService) {
                                       ^
  attempting to assign weaker access privileges; was public

> Task :spring-integration-kafka:compileKotlin
```

chrylis: This does bring up a concern, though, that this is a breaking change for implementers (SemVer minor). Not sure how this changes whether it's worth accepting.

artembilan: Oh, right! I wouldn't worry too much for version `6.0.1`, since it is unlikely this method is overridden without increasing access, but still bad.
I worry more about the version `5.5.x` where indeed a bytecode must not be made incompatible.
Therefore we can accept your change, but for consistency only in the next `6.1.0` version.
For now we are stuck with a workaround to extend the class to be able to increasing its access.
Although I believe with an XML configuration it should work even with a `protected` modifier 😄 

I suggest to close this PR since it is not appropriate for the code base it is requested.

Thank you for spotting this!

chrylis: This would be appreciated for the 6.x branch, but for now I can reflectively access it for our specific use case (we're using Groovy anyway).

PR #3973, opened 2022-12-15
GH-3954: Fix WebFlux XML config for ambiguity

### LINKED ISSUE #3954: WebFluxOutboundChannelAdapterParser silently ignores encoding-mode attribute if web-client attribute is present

2.7.1 but I believe that 6.0.0 also has this problem

**Describe the bug**

When the ```web-client attribute``` is included in the declararation of a ```webflux:outbound-gateway``` using XML namespace support, the ```encoding-mode``` attribute will be ignored silently because the encoding mode from the ```UriBuilderFactory``` of the ```web-client``` will be used instead.

**To Reproduce**

Include an ```encoding-mode``` attribute and a ```web-client``` with a different ```encoding-mode``` within the ```UriBuilderFactory```.

**Expected behavior**

An error or warning to signal that ```encoding-mode``` is ignored.

**Sample**

Similar to https://github.com/spring-projects/spring-integration/pull/3939

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3954

When `web-client` attribute is provided for the WebFlux outbound components configuration via XML, the `encoding-mode` is silently ignored

* Check for `encoding-mode` attribute presence in the `WebFluxOutboundChannelAdapterParser` when `web-client` is provided and throw respective exception to reject such a config

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/config/WebFluxOutboundChannelAdapterParser.java
+++ b/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/config/WebFluxOutboundChannelAdapterParser.java
@@ -53,7 +53,7 @@ public class WebFluxOutboundChannelAdapterParser extends HttpOutboundChannelAdap
 		if (StringUtils.hasText(webClientRef)) {
 			if (element.hasAttribute("encoding-mode")) {
 				parserContext.getReaderContext()
-						.error("The 'web-client' and 'encoding-mode' are mutually exclusive.", element);
+						.error("The 'web-client' and 'encoding-mode' attributes are mutually exclusive.", element);
 			}
 
 			builder.getBeanDefinition()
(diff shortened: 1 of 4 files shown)


spring-integration-webflux/src/main/java/org/springframework/integration/webflux/config/WebFluxOutboundChannelAdapterParser.java
```diff
@@ -51,6 +51,11 @@ static BeanDefinitionBuilder buildWebFluxRequestExecutingMessageHandler(Element
 
 		String webClientRef = element.getAttribute("web-client");
 		if (StringUtils.hasText(webClientRef)) {
+			if (element.hasAttribute("encoding-mode")) {
+				parserContext.getReaderContext()
+						.error("The 'web-client' and 'encoding-mode' are mutually exclusive.", element);
```
garyrussell, line 56:
```suggestion
						.error("The 'web-client' and 'encoding-mode' attributes are mutually exclusive.", element);
```

garyrussell (changes requested):
Minor change.

garyrussell (approved):

PR #3975, opened 2022-12-16
Replace javadoc mentions of JSch with MINA SSHD

otlaitil:
JSch was replaced with MINA SSHD in v6 upgrade.


artembilan:
@otlaitil ,

thank you for spotting this and fixing; looking forward for more!

otlaitil:
Thanks @artembilan !

PR #3981, opened 2023-01-04
GH-3980: Fix DefaultSftpSFactory for concurrency

### LINKED ISSUE #3980: Concurrent threads might use uninitialised SFTP client

spring-integration-sftp 6.0.1

**Describe the bug**

If SFTP client is not yet initialised and two threads simultaneously try to use SFTP client, one of them may fail.

`DefaultSftpSessionFactory::initClientSession` calls `DefaultSftpSessionFactory::initClient` which initialises the client.
Client initialisation is done only by the first thread.
The second thread skips initialisation and may use uninitialised/partially initialised client.
It causes exception:
```java
Caused by: java.lang.IllegalStateException: failed to create SFTP Session
	at org.springframework.integration.sftp.session.DefaultSftpSessionFactory.getSession(DefaultSftpSessionFactory.java:291)
	at org.springframework.integration.sftp.session.DefaultSftpSessionFactory.getSession(DefaultSftpSessionFactory.java:67)
	at org.springframework.integration.file.remote.session.CachingSessionFactory$1.createForPool(CachingSessionFactory.java:85)
	at org.springframework.integration.file.remote.session.CachingSessionFactory$1.createForPool(CachingSessionFactory.java:82)
	at org.springframework.integration.util.SimplePool.doGetItem(SimplePool.java:206)
	at org.springframework.integration.util.SimplePool.getItem(SimplePool.java:187)
	... 20 common frames omitted
Caused by: java.lang.IllegalStateException: SshClient not started. Please call start() method before connecting to a server
	at org.apache.sshd.client.SshClient.doConnect(SshClient.java:626)
	at org.apache.sshd.client.SshClient.doConnect(SshClient.java:615)
	at org.apache.sshd.client.SshClient.connect(SshClient.java:546)
	at org.apache.sshd.client.session.ClientSessionCreator.connect(ClientSessionCreator.java:199)
	at org.apache.sshd.client.session.ClientSessionCreator.connect(ClientSessionCreator.java:188)
	at org.springframework.integration.sftp.session.DefaultSftpSessionFactory.initClientSession(DefaultSftpSessionFactory.java:314)
	at org.springframework.integration.sftp.session.DefaultSftpSessionFactory.getSession(DefaultSftpSessionFactory.java:281)
	... 25 common frames omitted
```

**To Reproduce**

After application startup when SFTP client is not yet initialised make two SFTP requests simultaneously

**Expected behavior**

All threads must wait for client initialisation. 
Uninitialised/partially initialised client should never be used.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3980

When not `isSharedSession`, the `initClient()` is called for every session we request from the factory and in concurrent calls we end up with not initialized SSH client in some threads.

* Add `synchronized` to the `initClient()` to block other threads while the first one initialize the client

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -320,7 +320,7 @@ public class DefaultSftpSessionFactory implements SessionFactory<SftpClient.DirE
 		return clientSession;
 	}
 
-	private void initClient() throws IOException {
+	private synchronized void initClient() throws IOException {
 		if (this.initialized.compareAndSet(false, true)) {
 			if (this.port <= 0) {
 				this.port = SshConstants.DEFAULT_PORT;
--- a/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
+++ b/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2022 the original author or authors.
+ * Copyright 2014-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


kdebski85:
I prepared slightly different approach with Double-checked locking: https://github.com/spring-projects/spring-integration/pull/3982

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
```diff
@@ -320,7 +320,7 @@ private ClientSession initClientSession() throws IOException {
 		return clientSession;
 	}
 
-	private void initClient() throws IOException {
+	private synchronized void initClient() throws IOException {
```
garyrussell, line 323: Synchronization has to be higher up - otherwise if `isSharedSession` is true, we could orphan a session; see code in `getSession`.

https://github.com/spring-projects/spring-integration/blob/f8fff8188622e4a9602cc46e430ab38776c58f31/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java#L286-L288

garyrussell (changes requested):

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
artembilan: If `isSharedSession` we do `this.sharedSessionLock.lock();`

garyrussell: Oops - right; ignore.

garyrussell (approved):

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
```diff
@@ -320,7 +320,7 @@ private ClientSession initClientSession() throws IOException {
...
 		if (this.initialized.compareAndSet(false, true)) {
```
kdebski85, line 324: `this.initialized` should not be set at the beginning of the method.
If initialisation fails, the variable will be set to true for un this.initialized client

artembilan: Yeah... Makes sense.
OK closing my PR.
Feel free to copy my unit test into your PR then we will review it and merge properly.

artembilan:
Closed in favor: https://github.com/spring-projects/spring-integration/pull/3982

PR #3982, opened 2023-01-04
GH-3980: Fix concurrent initialization in DefaultSftpSessionFactory

### LINKED ISSUE #3980: Concurrent threads might use uninitialised SFTP client

kdebski85:
fixes: https://github.com/spring-projects/spring-integration/issues/3980

Fix concurrent initialization in DefaultSftpSessionFactory

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,7 +22,6 @@ import java.security.GeneralSecurityException;
 import java.security.KeyPair;
 import java.time.Duration;
 import java.util.Collection;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -61,6 +60,7 @@ import org.springframework.util.Assert;
  * @author David Liu
  * @author Pat Turner
  * @author Artem Bilan
+ * @author Krzysztof Debski
  *
  * @since 2.0
  */
@@ -68,7 +68,7 @@ public class DefaultSftpSessionFactory implements SessionFactory<SftpClient.DirE
 
 	private final SshClient sshClient;
 
-	private final AtomicBoolean initialized = new AtomicBoolean();
+	private volatile boolean initialized;
 
 	private final boolean isSharedSession;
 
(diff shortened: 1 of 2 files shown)


artembilan:
Thank you, but I have already made the fix here: https://github.com/spring-projects/spring-integration/pull/3981.
WDYT?

kdebski85:
My approach might be slightly faster because:

1. I do not use synchronized
2. I do not use AtomicBoolean

Can I copy test from your PR?

garyrussell:
See my comment on https://github.com/spring-projects/spring-integration/pull/3981

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
```diff
@@ -321,41 +321,50 @@ private ClientSession initClientSession() throws IOException {
...
 
 	private void initClient() throws IOException {
-		if (this.initialized.compareAndSet(false, true)) {
-			if (this.port <= 0) {
-				this.port = SshConstants.DEFAULT_PORT;
+		if (!this.initialized) {
+			synchronized (this) {
+				if (!this.initialized) {
+					doInitClient();
+					this.initialized = true;
```
artembilan, line 328: @garyrussell , WDYT about this double-check solution instead?

garyrussell: It's marginally better, I suppose, but performance is rarely an issue here because most people cache sessions.

That said, the patch looks more complicated than it really is, review with `?w=1`.

artembilan (changes requested):
@kdebski85 ,

any chances to see a unit test to confirm that we don't have concurrency issue any more?
As we discussed before you can simply copy-paste a test I provide in my rejected PR: https://github.com/spring-projects/spring-integration/pull/3981/files#diff-ea9ddb7293f89610654c162b3b862b3b590008c3feb48e12ec2d57c69027d16eR96

Thanks

Krzysztof-Debski (approved):

artembilan:
@kdebski85 ,

thank you for contribution and reasonable discussion; looking forward for more!

PR #3983, opened 2023-01-04
Do not build exception instances in advance

artembilan:
When no `cause` and no clean context of the exception thrown, the code place where an exception has been created is confusing.

* Fix `SimpleMessageStore` to create an `out of capacity` exception in the `addMessagesToGroup()` exactly at the point where it is thrown
* Fix `FunctionExpression` and `SupplierExpression` do not pre-create `readOnlyException`: it is unlikely these kind of expressions are going to be used in the SpEL `write` context, so we save some time and memory not creating extra object and the plain call `throw new` gives us a clean context what method call has ended up with such an exception

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
@@ -281,9 +281,7 @@ public class SimpleMessageStore extends AbstractMessageGroupStore
 				MessageGroup group = this.groupIdToMessageGroup.get(groupId);
 				if (group == null) {
 					if (this.groupCapacity > 0 && messages.length > this.groupCapacity) {
-						throw new MessagingException(getClass().getSimpleName() +
-								" was out of capacity (" + this.groupCapacity + ") for group '" + groupId +
-								"', try constructing it with a larger number.");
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
```diff
@@ -278,13 +279,11 @@ public void addMessagesToGroup(Object groupId, Message<?>... messages) {
...
-				MessagingException outOfCapacityException =
-						new MessagingException(getClass().getSimpleName() +
-								" was out of capacity (" + this.groupCapacity + ") for group '" + groupId +
-								"', try constructing it with a larger capacity.");
 				if (group == null) {
 					if (this.groupCapacity > 0 && messages.length > this.groupCapacity) {
-						throw outOfCapacityException;
+						throw new MessagingException(getClass().getSimpleName() +
+								" was out of capacity (" + this.groupCapacity + ") for group '" + groupId +
+								"', try constructing it with a larger number.");
```
garyrussell, line 286: This should be extracted to a method to avoid the duplicate code.

garyrussell (changes requested):

garyrussell (approved):

PR #3989, opened 2023-01-11
Revise JS support for regular JVM

artembilan:
Turns out that we just need to have `org.graalvm.sdk:graal-sdk` and `org.graalvm.js:js` dependencies for regular JVM to enable JavaScript support for Spring Integration Scripting module

* Add respective `providedImplementation` dependencies
* Remove `@EnabledIfSystemProperty` from unit tests
* Mention those dependencies in the doc

### DIFF
--- a/src/reference/asciidoc/scripting.adoc
+++ b/src/reference/asciidoc/scripting.adoc
@@ -295,4 +295,4 @@ By default, the framework sets `allowAllAccess` to `true` on the shared Polyglot
 
 This can be customized via overloaded `PolyglotScriptExecutor` constructor which accepts a `org.graalvm.polyglot.Context.Builder`.
 
-The usage of JavaScript in the framework components remains the same, on ly the difference that now it is going to work only on GraalVM with installed `js` component.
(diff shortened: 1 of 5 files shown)


src/reference/asciidoc/scripting.adoc
```diff
@@ -295,4 +295,4 @@ By default, the framework sets `allowAllAccess` to `true` on the shared Polyglot
 
 This can be customized via overloaded `PolyglotScriptExecutor` constructor which accepts a `org.graalvm.polyglot.Context.Builder`.
 
-The usage of JavaScript in the framework components remains the same, on ly the difference that now it is going to work only on GraalVM with installed `js` component.
+To enable this JavaScript support the GraalVM with installed `js` component has to be used or `org.graalvm.sdk:graal-sdk` and `org.graalvm.js:js` dependencies must be included for regular host JVM.
```
garyrussell, line 298: I am having trouble parsing this sentence so can't offer any suggestions. 

Are you saying that you need these graalvm dependencies, even with a regular JVM?

If that's the case, my suggestion would be

```suggestion
To enable this JavaScript support, GraalVM with the `js` component installed has to be used or, when using a regular JVM, the `org.graalvm.sdk:graal-sdk` and `org.graalvm.js:js` dependencies must be included.
```

garyrussell (changes requested):

PR #3995, opened 2023-01-20
GH-3993: Fix async race condition in TcpOutGateway

### LINKED ISSUE #3993: TcpOutboundGateway. Semaphore not released after exception
Hello, I have found next behaviour which could be an issue.

I'm working around with the HelloWorldInterceptor example, a TcpOutboundGateway in async mode and a ConnectionFactory with singleUse=false field. 

In this test the Server is who should start the handshake conversation. 
Once the client tries to send a message it waits for hello" message from the server, if the await time expires it throws a MessagingException (negotiation error); then the exception is catched by TcpOutboundGateway, and cleanup method in finally block is not invoked so semaphore is not released and next invocations to handleMessage are locked (until request timeout passed) when acquiring the semaphore.

Attached proofOfConcept and test execution log.

[Git proofOfConcept](https://github.com/opardor/spring-tcp-demo)
[test.log](https://github.com/spring-projects/spring-integration/files/10453822/test.log)

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3993

When `TcpOutboundGateway` is in an `async` mode and `CCF` is configured not for `singleUse` an `Semaphore` around an obtained `TcpConnection` is involved. If we fail on `TcpConnection.send()`, resources are not clean up, including the mentioned `Semaphore`: in async mode this happens only when we receive a reply.

* Catch an exception on the `TcpConnection.send()` and perform `cleanUp()` in async mode.
* Add `cleanUp()` into a scheduled task from the `TcpOutboundGateway.AsyncReply` when no reply arrives in time.
* Optimize `TcpOutboundGateway.AsyncReply` behavior to cancel no-reply scheduled task when reply arrives into a `CompletableFuture`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
@@ -457,10 +457,12 @@ public class TcpOutboundGateway extends AbstractReplyProducingMessageHandler
 				this.noResponseFuture =
 						getTaskScheduler()
 								.schedule(() -> {
-									cleanUp(this.haveSemaphore, this.connection, this.connection.getConnectionId());
-									this.future.completeExceptionally(
+									if (this.future.completeExceptionally(
 											new MessageTimeoutException(requestMessage,
-													"Timed out waiting for response"));
+													"Timed out waiting for response"))) {
+
+										cleanUp(this.haveSemaphore, this.connection, this.connection.getConnectionId());
+									}
 								}, Instant.now().plusMillis(remoteTimeout));
 			}
 			else {
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2001-2022 the original author or authors.
+ * Copyright 2001-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
```diff
@@ -440,13 +454,25 @@ private final class AsyncReply {
...
-				getTaskScheduler()
-						.schedule(() -> {
-							TcpOutboundGateway.this.pendingReplies.remove(connection.getConnectionId());
-							this.future.completeExceptionally(
-									new MessageTimeoutException(requestMessage, "Timed out waiting for response"));
-						}, Instant.now().plusMillis(remoteTimeout));
+				this.noResponseFuture =
+						getTaskScheduler()
+								.schedule(() -> {
+									cleanUp(this.haveSemaphore, this.connection, this.connection.getConnectionId());
```
garyrussell, line 460: I think there's a new race here; we could end up releasing the semaphore twice because `onMessage` calls cleanup before completing the future.

Perhaps check if `pendingReplies` was actually removed in cleanup before releasing the semaphore?

garyrussell (changes requested):

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
artembilan: See some fixed I have pushed.

garyrussell (approved):

garyrussell:
Forget that; since it's internal, we can use `CompletableFuture`.

garyrussell:
...and cherry-picked as 1aef04171f after resolving conflicts.

artembilan:
That's OK, Gary.
We indeed can use it internally.
However the `AbstractMessageProducingHandler` cannot handle `CompletableFuture` in that `5.5.x` version:
```
if (this.async && (reply instanceof ListenableFuture<?> || reply instanceof Publisher<?>)) {
```
So, your catch is good.

I'll provide the fix in subsequent commit...

garyrussell:
Oops - compile error too.

artembilan:
An adaptation to `ListenableFuture`: https://github.com/spring-projects/spring-integration/commit/070d1868e59cbad73e798ffe33ec5bb48f4c07a1

PR #3999, opened 2023-02-01
Propagate and Observation from Reactive context

artembilan:
To propagate an `Observation` from reactive stream (e.g. WebFlux) we have to capture its context and set it into the current thread scope.

* Add a `io.micrometer:context-propagation` dependency to support reactive context propagation
* Populate a `parentObservation` in the `IntegrationObservation.PRODUCER.observation()` since it is not available for tracing on `Observation.onStart()`. Might be tentative until upcoming fix in Micrometer Observation
* Populate from reactive context in the `WebFluxInboundEndpoint` where we use just `send()` operation downstream
* Populate from reactive context in the `MessagingGatewaySupport` where we use `send()` operation downstream or `FluxMessageChannel.subscribeTo()`
* Use `contextCapture()` in the `FluxMessageChannel` to gather a `ThreadLocal` info into a Reactor context and then set back to `ThreadLocal` in the `transformDeferredContextual()` which really happens on a different thread
* Verify a trace propagation from WebFlux to an integration flow via Brave instrumentation in the `WebFluxObservationPropagationTests`

### DIFF
--- a/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/inbound/WebFluxInboundEndpoint.java
+++ b/spring-integration-webflux/src/main/java/org/springframework/integration/webflux/inbound/WebFluxInboundEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2022 the original author or authors.
+ * Copyright 2017-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import io.micrometer.context.ContextSnapshot;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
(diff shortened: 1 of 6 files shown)


spring-integration-webflux/src/main/java/org/springframework/integration/webflux/inbound/WebFluxInboundEndpoint.java
```diff
@@ -151,16 +152,20 @@ private Mono<Void> doHandle(ServerWebExchange exchange) {
...
-					}
-				})
+				.flatMap(requestTuple ->
+						Mono.deferContextual(contextView -> {
+							if (isExpectReply()) {
+								return sendAndReceiveMessageReactive(requestTuple.getT1())
+										.flatMap(replyMessage -> populateResponse(exchange, replyMessage));
+							}
+							else {
+								var scope = ContextSnapshot.setAllThreadLocalsFrom(contextView);
```
garyrussell, line 162: Why is this not required in the send/receive case?

artembilan: It is covered by the `MessagingGatewaySupport` change in this PR. 
We need such an action only when we lose a reactive scope.
This `send()` returns nothing, but `sendAndReceiveMessageReactive()` returns a `Mono` where respective context is still present. Therefore we have to repeat a set to `ThreadLocal` over there where we lose the reactive stream context again.
That's why I have those separate tests in this PR.

NOTE: The RSocket support will work as well just because we have a fix in the `MessagingGatewaySupport`  and it does not use a plain `send()`.

PR #4003, opened 2023-02-02
GH-4001: Doc for cooperation with some HZ objects

### LINKED ISSUE #4001: When configuring a HazelcastEventDrivenMessageProducer I'm not seeing the message actually dequeue from my hazelcast queue.
I'm trying to configure a mechanism using spring integration to consume from a hazelcast queue into my spring boot project running on Java 11

I'm using spring integration with the following dependencies in my pom.xml

```
<dependency>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-integration</artifactId>
</dependency>
<dependency>
  <artifactId>spring-integration-hazelcast</artifactId>
  <version>3.0.0</version>
</dependency>

```        
No matter how I configure my  HazelcastEventDrivenMessageProducer, MessageChannel, and QueueChannel I can't get my consumer code to run once per message and dequeue from my hazelcast queue. I am trying to read from an externalized queue in my external hazelcast cluster and dequeue the message once done.

Here is my configuration setup..

```
 @Bean
    HazelcastEventDrivenMessageProducer hazelcastEventDrivenMessageProducer(@Qualifier("myQueue") IQueue<String> myQueue) {
        HazelcastEventDrivenMessageProducer hazelcastEventDrivenMessageProducer = new HazelcastEventDrivenMessageProducer(myQueue);
        hazelcastEventDrivenMessageProducer.setOutputChannel(hazelcastQueueChannel());
        hazelcastEventDrivenMessageProducer.setCacheEventTypes("ADDED,REMOVED");
        hazelcastEventDrivenMessageProducer.setCacheListeningPolicy(CacheListeningPolicyType.SINGLE);
        return hazelcastEventDrivenMessageProducer;
    }
    
    @Bean
    IQueue<String> myQueue(@Qualifier("externalHazelcastInstance") HazelcastInstance hazelcastInstance) {
        return hazelcastInstance.getQueue("myQueue");
    }

    @Bean
    public PollableChannel hazelcastQueueChannel() {
        return new QueueChannel();
    }

```    
    I also have a consumer class that looks like this.
    
```
@Slf4j
@Component
public class QueueTaskConsumer {
    @ServiceActivator(inputChannel = "hazelcastQueueChannel")
    public void readFromTaskQueue(String message) {
        log.info("received the message inside QueueTaskConsumer " + message);
    }
}
```

**To Reproduce**

Configure your setup as the same above and try to add 10 messages(String) to a hazelcast queue 

**Expected behavior**

I expect to see the output of "received the message inside QueueTaskConsumer" in my test 10 times and the queue to be empty.

Currently I have 17 messages in my queue and I have a test that writes 10 messages to my queue with my consumer set up this way. I can see the logging of these 10 messages with "received the message inside QueueTaskConsumer" but the other 17 remain and now I have 27 messages in my queue. It never seems to dequeue at all from my queue.

The consumer should have logged those 10 + the 17 that already existed and dequeued all 27.

I have also tried adding a consumer inside my class representing a Spring Boot app with @SpringBootTest. There I try to inject my channel itself and call receive and the item variable is always null when trying to get from the channel.

`@SpringBootApplication(scanBasePackages = "com.mypackage")
public class ExternalHazelcastApp {
        SpringApplication app = new SpringApplication(ExternalHazelcastApp.class);
        app.setDefaultProperties(Collections.singletonMap("server.port", "8085"));
        app.run(args);
    }

    @Component
    @Slf4j
    public static class QueueTaskConsumerInternal {
        @Autowired @Qualifier("hazelcastQueueChannel") public PollableChannel hazelcastQueueChannel;

        @EventListener(value = ApplicationReadyEvent.class)
        public void start() {
            log.info("CONSUMER HAS STARTED");
            while (true)
            {
                try {
                    log.info("INSIDE WHILE LOOP");
                    Message<?> item = hazelcastQueueChannel.receive(1_000); // I always get a null here and a npe on line below.
                    log.info("Consumed: " + item.getPayload());
                }
                catch (Exception e) {
                    log.info("exception " + e.getCause());
                    e.printStackTrace();
                }
            }
        }
    }
}`

I have seen my queue actually dequeue messages when my queue setup is like the following below. With this though, I see consumption code run 50 times for one message and that cannot happen. The following is how i configured it.

```
@Bean
    IQueue<Message<?>> myQueue(@Qualifier("externalHazelcastInstance") HazelcastInstance hazelcastInstance) {
        return hazelcastInstance.getQueue(MY_TEST_QUEUE);
    }

    @Bean
    public MessageChannel hazelcastQueueChannel(@Qualifier("myQueue") IQueue<Message<?>> myQueue) {
        return new QueueChannel(myQueue);
    }
    
    @Bean
    HazelcastEventDrivenMessageProducer hazelcastEventDrivenMessageProducer(@Qualifier("myQueue") IQueue<Message<?>> myQueue) {
        HazelcastEventDrivenMessageProducer hazelcastEventDrivenMessageProducer = new HazelcastEventDrivenMessageProducer(myQueue);
        hazelcastEventDrivenMessageProducer.setOutputChannel(hazelcastQueueChannel(myQueue));
        hazelcastEventDrivenMessageProducer.setCacheEventTypes("ADDED");
        hazelcastEventDrivenMessageProducer.setCacheListeningPolicy(CacheListeningPolicyType.SINGLE); //must be single AND ADDED to SEE CONSUMPTION (NO DEQUEUE) regardless of cache type
        return hazelcastEventDrivenMessageProducer;
    }
```
    
 In this case i get the message and then it wrapped again and again over and over in a class called GenericMessage in my output. I abandoned this approach for now.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/4001

The `IQueue`, `ITopic` and `IExecutorService` can be used with Spring Integration channel as is without any extra component implementations.

* Document the cooperation feature with Hazelcast objects via samples

### DIFF
--- a/src/reference/asciidoc/hazelcast.adoc
+++ b/src/reference/asciidoc/hazelcast.adoc
@@ -598,3 +598,78 @@ public LockRegistry lockRegistry() {
 When used with a shared `MessageGroupStore` (e.g. `Aggregator` store management), the `HazelcastLockRegistry` can be used to provide this functionality across multiple application instances, such that only one instance can manipulate the group at a time.
 
 NOTE: For all the distributed operations the CP Subsystem must be enabled on `HazelcastInstance`.
+
+[[hazelcast-message-channels]]
+=== Message Channels with Hazelcast
+
+The Hazelcast `IQueue` and `ITopic` distributed objects are, essentially, messaging primitives and can be use with Spring Integration core components without extra implementations in this Hazelcast module.
+
+The <<./channel.adoc#channel-implementations-queuechannel,`QueueChannel`>> can be supplied by any `java.util.Queue`, including the mentioned Hazelcast distributed `IQueue`:
+
+====
+[source,java]
+----
+@Bean
+PollableChannel hazelcastQueueChannel(HazelcastInstance hazelcastInstance) {
+    return new QueueChannel(hazelcastInstance.Message<?>>getQueue("springIntegrationQueue"));
+}
+----
+====
+
+Placing this config on several nodes in Hazelcast cluster of the application, will make the `QueueChannel` as distributed and only one node will be able to poll a single `Message` from that `IQueue`.
+This works similar to <<./jms.adoc#jms-channel,`PollableJmsChannel`>>, <<./kafka.adoc#kafka-channels,`PollableKafkaChannel`>> or <<./amqp.adoc#amqp-channels,`PollableAmqpChannel`>>.
+
+An `ITopic` abstraction in Hazelcast has similar semantics to a `Topic` in JMS: all subscribers receive published messages.
+With a pair of simple `MessageChannel` beans this mechanism as an out-of-the-box feature:
+
+====
+[source,java]
+----
+@Bean
+public ITopic<Message<?>> springIntegrationTopic(HazelcastInstance hazelcastInstance,
+        MessageChannel fromHazelcastTopicChannel) {
+
+    ITopic<Message<?>> topic = hazelcastInstance.getTopic("springIntegrationTopic");
+	topic.addMessageListener(m -> fromHazelcastTopicChannel.send(m.getMessageObject()));
+	return topic;
+}
+
+@Bean
+public MessageChannel publishToHazelcastTopicChannel(ITopic<Message<?>> springIntegrationTopic) {
+    return new FixedSubscriberChannel(springIntegrationTopic::publish);
+}
+
+@Bean
+public MessageChannel fromHazelcastTopicChannel() {
+    return new DirectChannel();
+}
+----
+====
+
+The `FixedSubscriberChannel` is an optimized variant of `DirectChannel`, which requires a `MessageHandler` on initialization.
+Since the `MessageHandler` is a functional interface a simple lambda for the `handleMessage` method can be provided.
+When a message is sent to the `publishToHazelcastTopicChannel` it is just published onto the Hazelcast `ITopic`.
+The `com.hazelcast.topic.MessageListener` is a functional interface, too, hence a lambda to the `ITopic#addMessageListener` can be provided.
+So, when ever in the Hazelcast the `fromHazelcastTopicChannel` is subscribed, it is going to consume a message sent to the mentioned `ITopic`.
+
+An `ExecutorChannel` can be supplied with an `IExecutorService`.
+For example, with respective configuration a cluster-wide singleton can be achieved:
+
+====
+[source,java]
+----
+@Bean
+public HazelcastInstance hazelcastInstance() {
+    return Hazelcast.newHazelcastInstance(
+                new Config()
+                    .addExecutorConfig(new ExecutorConfig()
+                         .setName("singletonExecutor")
+                         .setPoolSize(1)));
+}
+
+@Bean
+public MessageChannel hazelcastSingletonExecutorChannel(HazelcastInstance hazelcastInstance) {
+    return new ExecutorChannel(hazelcastInstance.getExecutorService("singletonExecutor"));
+}
+----
+====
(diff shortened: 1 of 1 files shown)


artembilan:
@vsalvati,

I broke a GH issue you've raised a bit trying to transfer it to discussions.
Which is really a mistake since we have agreed that this has to be reflected some way in the docs.]
Not one can comment on the issue(or discussion) at the moment.

Please, see my second commit for an Inbound Channel Adapter on the `IQueue`. That's the way to consume data from non-Spring producer.
More over if there is no producer on a `QueueChannel` at all this would be a pattern abuse and better to stay away from channel in this case at all.

src/reference/asciidoc/hazelcast.adoc
```diff
@@ -598,3 +598,97 @@ public LockRegistry lockRegistry() {
...
+    return new QueueChannel(hazelcastInstance.Message<?>>getQueue("springIntegrationQueue"));
+}
+----
+====
+
+Placing this config on several nodes in Hazelcast cluster of the application, will make the `QueueChannel` as distributed and only one node will be able to poll a single `Message` from that `IQueue`.
+This works similar to <<./jms.adoc#jms-channel,`PollableJmsChannel`>>, <<./kafka.adoc#kafka-channels,`PollableKafkaChannel`>> or <<./amqp.adoc#amqp-channels,`PollableAmqpChannel`>>.
+
+If the producer side is not a Spring Integration, there is no way to configure a `QueueChannel`, and rather a plain Hazelcast `IQueue` API is used to produce the data.
+In this case a `QueueChannel` approach is wrong on the consumer side: an <<./channel-adapter.adoc#channel-adapter-namespace-inbound,Inbound Channel Adapter>> solution must be used instead:
```
garyrussell, line 623:
```suggestion
If the producer side is not a Spring Integration application, there is no way to configure a `QueueChannel`, and therefore the plain Hazelcast `IQueue` API is used to produce the data.
In this case, the `QueueChannel` approach is wrong on the consumer side: an <<./channel-adapter.adoc#channel-adapter-namespace-inbound,Inbound Channel Adapter>> solution must be used instead:
```

src/reference/asciidoc/hazelcast.adoc
```diff
@@ -598,3 +598,97 @@ public LockRegistry lockRegistry() {
...
+@Bean
+@InboundChannelAdapter(channel = "stringValuesFromHzQueueChannel")
+Supplier<String> fromHzIQueueSource(IQueue<String> myStringHzQueue) {
+    return myStringHzQueue::poll;
+}
+----
+====
+
+An `ITopic` abstraction in Hazelcast has similar semantics to a `Topic` in JMS: all subscribers receive published messages.
+With a pair of simple `MessageChannel` beans this mechanism as an out-of-the-box feature:
```
garyrussell, line 642:
```suggestion
The `ITopic` abstraction in Hazelcast has similar semantics to a `Topic` in JMS: all subscribers receive published messages.
With a pair of simple `MessageChannel` beans this mechanism is supported as an out-of-the-box feature:
```

src/reference/asciidoc/hazelcast.adoc
```diff
@@ -598,3 +598,97 @@ public LockRegistry lockRegistry() {
...
+    return new DirectChannel();
+}
+----
+====
+
+The `FixedSubscriberChannel` is an optimized variant of `DirectChannel`, which requires a `MessageHandler` on initialization.
+Since the `MessageHandler` is a functional interface a simple lambda for the `handleMessage` method can be provided.
+When a message is sent to the `publishToHazelcastTopicChannel` it is just published onto the Hazelcast `ITopic`.
+The `com.hazelcast.topic.MessageListener` is a functional interface, too, hence a lambda to the `ITopic#addMessageListener` can be provided.
+So, when ever in the Hazelcast the `fromHazelcastTopicChannel` is subscribed, it is going to consume a message sent to the mentioned `ITopic`.
```
garyrussell, line 672:
```suggestion
So, a subscriber to the `fromHazelcastTopicChannel` will consume all messages sent to the mentioned `ITopic`.
```

garyrussell (changes requested):

PR #4006, opened 2023-02-03
GH-4002: Deprecate `ConsumerStopAction` in MQTT

### LINKED ISSUE #4002: MQTT integration: Possiblity to set the consumerStopAction when using a ClientManager
**Expected Behavior**
In the MQTT integration when using a client manager, I need to be able to set the ConsumerStopAction. This could be achieved by adding a constructor with the possibility to provide both a ClientManager and a MqttPahoClientFactory.

**Current Behavior**
The ConsumerStopAction are defined on the MqttPahoClientFactory which is private. It is not possible to define the value when using a ClientManager. As it use a DefaultMqttPahoClientFactory which set it to UNSUBSCRIBE_CLEAN.

**Context**
I need to be able to have the messages queue up in the broker while the spring boot application restart for maintenance purposes. And i also can only use one client but I need several channelAdaptors. Currently restarting the spring application causes the loss of message that are critical for my use case.

A possible alternative would be to make a custom channelAdaptor but I feel it would be appropriate to implement it on the base class. I am willing to submit a pull request to implement those changes if they are approved.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/4002

Turns out the `ConsumerStopAction` was introduced in a point version to mitigate an unsubscription bug and let to preserve a previous behaviour.

* Deprecate `ConsumerStopAction` in favor of just `cleanSession` flag in the `MqttConnectOptions`

**Cherry-pick to `5.5.x`**

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ConsumerStopAction.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ConsumerStopAction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


garyrussell (approved):

garyrussell:
...and cherry-picked as b08809190250b5cb8d211434fc3ae23af194fdd9 after resolving conflicts.

PR #4011, opened 2023-02-06
GH-4008: Add leader and participants support

### LINKED ISSUE #4008: Impossible to retrieve current leader id from LeaderInitiator
Hello,

Currently it is impossible to retrieve current leader id from [LeaderInitiator](https://github.com/spring-projects/spring-integration/blob/main/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/leader/LeaderInitiator.java) because LeaderSelector field is private and has no getter.

It would be really helpful to have this functionality for cases when we need to understand what node is a leader.
I think it is better to add getter for LeaderSelector instead of delegate method for current leader. It could cover cases when we need for example retrieve list of all participants.

ivan-zaitsev:
Fixes spring-projects/spring-integration#4008

* change return type of LeaderInitiator.getContext() from Context to CuratorContext
* add getLeader() method to CuratorContext
* add getParticipants() to CuratorContext

### DIFF
--- a/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/leader/LeaderInitiator.java
+++ b/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/leader/LeaderInitiator.java
@@ -16,12 +16,16 @@
 
 package org.springframework.integration.zookeeper.leader;
 
+import java.util.Collection;
+import java.util.List;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Please, see code style of the project: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-Framework-Code-Style.

You need to reconfigure your IDE for our editor config, so imports are ordered according to our Checkstyle requirements.

The rest looks good to me.
Will double check locally tomorrow and merge.

👍

ivan909020:
@artembilan Fixed code style

PR #4015, opened 2023-02-08
GH-4013: TCP DSL Improvements

### LINKED ISSUE #4013: Spring Integration - TCP SSL Support via DSL
### Discussed in https://github.com/spring-projects/spring-integration/discussions/4012

<sup>Originally posted by **ppussar** February  8, 2023</sup>
Hi,

I am trying to implement a TCP Server, which does client certificate authentication on its connections. I added a TcpSocketSupport Bean as described in the [documentation](https://docs.spring.io/spring-integration/reference/html/ip.html#ssl-client-authentication-example).

However I get a ClassCastException during the serverSocket cast, as it is still a ServerSocket and not a SSLServerSocket.

So I tryed (without success) to overwrite the TcpSocketFactorySupport Bean:

```
@Bean
    public TcpSocketSupport tcpSocketSupport(serverProperties serverProperties) {
        if (serverProperties.getSsl().isEnabled()) {
            return new DefaultTcpSocketSupport() {
                @Override
                public void postProcessServerSocket(ServerSocket serverSocket) {
                    ((SSLServerSocket) serverSocket).setNeedClientAuth(true);
                }
            };
        } else {
            return new DefaultTcpSocketSupport();
        }
    }

    @Bean
    public TcpSocketFactorySupport tcpSocketFactorySupport(
            ServerProperties serverProperties,
            TcpSSLContextSupport tcpSSLContextSupport) {

        if (serverProperties.getSsl().isEnabled()) {
            return new DefaultTcpNetSSLSocketFactorySupport(tcpSSLContextSupport);
        } else {
            return new DefaultTcpNetSocketFactorySupport();
        }
    }
```

But it has no effect. I guess it is not working as I am using the DSL syntax, which does not support SSL?:

```
IntegrationFlow.from(Tcp.inboundGateway(
                                Tcp.netServer(serverPort)
                                        .tcpSocketSupport(tcpSocketSupport)
                                        .deserializer(mySerde)
                                        .serializer(mySerde)
                                        .soTimeout(Integer.MAX_VALUE)
                                        .singleUseConnections(false)
                                        .mapper(tcpMessageMapper))
                        .id(TCP_GATEWAY_ID)
                        .errorChannel(ERROR_CHANNEL_NAME)
                        .replyChannel(REPLY_CHANNEL_NAME))
```

Is there a way to set the tcpSocketFactorySupport to DefaultTcpNetSSLSocketFactorySupport via DSL?</div>

garyrussell:
Resolves https://github.com/spring-projects/spring-integration/issues/4013

Add Net/NIO specific Server/Client connection factory specs.

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/dsl/ConnectionFacforyTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/dsl/ConnectionFacforyTests.java
@@ -101,9 +101,9 @@ public class ConnectionFacforyTests {
 		TcpNetConnectionSupport conSupp = mock(TcpNetConnectionSupport.class);
 		TcpSocketFactorySupport factSupp = mock(TcpSocketFactorySupport.class);
 		TcpNetServerConnectionFactory server = Tcp.netServer(0)
-				.tcpSocketSupport(sockSupp)
-				.tcpNetConnectionSupport(conSupp)
-				.tcpSocketFactorySupport(factSupp)
+				.socketSupport(sockSupp)
+				.connectionSupport(conSupp)
+				.socketFactorySupport(factSupp)
 				.get();
 		assertThat(TestUtils.getPropertyValue(server, "tcpSocketSupport")).isSameAs(sockSupp);
 		assertThat(TestUtils.getPropertyValue(server, "tcpNetConnectionSupport")).isSameAs(conSupp);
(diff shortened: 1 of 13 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/dsl/ConnectionFacforyTests.java
```diff
@@ -89,4 +95,62 @@ public void shouldReturnNetFlavor() throws Exception {
...
+	}
+
+	@Test
+	void nioCustomClient() {
+		TcpSocketSupport sockSupp = mock(TcpSocketSupport.class);
+		TcpNioConnectionSupport conSupp = mock(TcpNioConnectionSupport.class);
+		TcpNioClientConnectionFactory client = Tcp.nioClient("localhost", 0)
+				.tcpSocketSupport(sockSupp)
+				.usingDirectBuffers(true)
+				.tcpNioConnectionSupport(conSupp)
```
artembilan, line 149: Cannot we call methods as simple as `socketSupport`, `connectionSupport` and `socketFactorySupport`?
We are really in the context of ` Tcp.netClient` builder or `Tcp.nioClient`, so it sounds redundant for me to mention `tcp` or `nio` in every method of the builder again.

artembilan (changes requested):

Issue #3772, opened 2022-04-08
Protobuf Support


devlap2:
**Expected Behavior**
I'd like there to be a `Transformer` of `ProtobufToObjectTransformer`. The `Transformers` class should include a `fromProto` function and a `toProto` function.

**Current Behavior**
There is no Protocol Buffers support in Spring Integration at all.

**Context**
Protocol Buffers are a popular and wide used mechanism for serializing structured data. This is a standard in the big data development and support to it will, later on, enable the development of grpc support.

artembilan:
I suggest to align the impl with existing Avro transformers - `SimpleFromAvroTransformer` & `SimpleToAvroTransformer`: `FromProtobufTransformer` and `ToProtobufTransformer`.

We also might borrow some ideas from Spring Cloud Function to support gRPC channel adapters: https://spring.io/blog/2021/12/02/spring-cloud-function-3-2-is-out.

See also: https://github.com/spring-projects/spring-integration/issues/2626

and contribution is welcome: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc !

ipanimnim:
Agree with @artembilan. This can be renamed to Protobuf & grpc support / we should add another grpc focused ticket. Grpc support is super important either.

PR #4016, opened 2023-02-09
Add Protobuf transformation support

tzolov:
- Add basic To/From Protocol Buffer's com.google.protobuf.Message transformers.
 - Allow the proto_type header to specify the type.
 - Add tests.

 Resoles: #3772


build.gradle
```diff
@@ -524,6 +525,7 @@ project('spring-integration-core') {
     apply plugin: 'org.jetbrains.dokka'
 
     dependencies {
+        api "com.google.protobuf:protobuf-java:$protobufVersion"
```
artembilan, line 528: This has to be an `optionalApi` and therefore move it to respective section of the dependencies.
The framework by itself doesn't use this library directly, like we do, for example, with `reactor-core`, so this `protobuf` has to be realigned same way as we do for `avro`.

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+import org.springframework.integration.context.IntegrationContextUtils;
+import org.springframework.integration.expression.FunctionExpression;
+import org.springframework.integration.transformer.support.ProtoHeaders;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.ConcurrentReferenceHashMap;
+
+/**
+ * An Protocol Buffer transformer to instantiate {@link com.google.protobuf.Message} objects from {@code byte[]}.
```
artembilan, line 38: Language `A Protocol ...` 😄

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+		}
+		if (type == null) {
+			type = this.defaultType;
+		}
+
+		try {
+			Builder messageBuilder = this.getMessageBuilder(type);
+			byte[] payload = (byte[]) message.getPayload();
+			com.google.protobuf.Message boza = messageBuilder.mergeFrom(payload).build();
+			return boza;
```
artembilan, line 147: I'm not sure in word `boza` and seems for me this variable can be eliminated altogether as a simple `return messageBuilder.mergeFrom(payload).build()`.
Although this might be for now: we might to do some post-processing eventually.
But still `boza` is not what I'd like to see in the production code 😄

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+			catch (ClassNotFoundException | LinkageError e) {
+				throw new IllegalStateException(e);
+			}
+		}
+		if (type == null) {
+			type = this.defaultType;
+		}
+
+		try {
+			Builder messageBuilder = this.getMessageBuilder(type);
```
artembilan, line 144: We don't use `this.` for methods.

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+	 * Create a new {@code Message.Builder} instance for the given class.
+	 * <p>
+	 * This method uses a ConcurrentReferenceHashMap for caching method lookups.
+	 */
+	private com.google.protobuf.Message.Builder getMessageBuilder(Class<? extends com.google.protobuf.Message> clazz) {
+		try {
+			Method method = methodCache.get(clazz);
+			if (method == null) {
+				method = clazz.getMethod("newBuilder");
+				methodCache.put(clazz, method);
```
artembilan, line 164: `computeIfAbsent()` instead?
And why do use use reflection for this method at all?
Why Protobuf does not have an API for `Message.newBuilder()` ?

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+		return this;
+	}
+
+	/**
+	 * Set the expression to evaluate against the message to determine the type id. Default
+	 * {@code headers['proto_type']}.
+	 * @param expression the expression.
+	 * @return the transformer
+	 */
+	public FromProtobufTransformer typeExpression(String expression) {
```
artembilan, line 87: I think we don't need extra methods for builder style.
Can't it be just a single `public FromProtobufTransformer setTypeExpression(Expression expression) {`?
This way we will satisfy XML configuration support and will have a way to chain methods in a builder style.

spring-integration-core/src/main/java/org/springframework/integration/transformer/ToProtobufTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    protected void onInit() {
+        this.evaluationContext = IntegrationContextUtils.getEvaluationContext(getBeanFactory());
+    }
+
+    @Override
+    protected Object doTransform(Message<?> message) {
+        Assert.state(message.getPayload() instanceof com.google.protobuf.Message,
```
artembilan, line 95: See `Assert.isInstanceOf()` instead.

spring-integration-core/src/main/java/org/springframework/integration/transformer/ToProtobufTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+                "Payload must be an implementation of 'com.google.protobuf.Message'");
+        com.google.protobuf.Message protobufMessage = (com.google.protobuf.Message) message.getPayload();
+
+        return getMessageBuilderFactory().withPayload(protobufMessage.toByteArray())
+                .copyHeaders(message.getHeaders())
+                .setHeader(ProtoHeaders.TYPE, this.typeIdExpression.getValue(this.evaluationContext, message))
```
artembilan, line 101: Do you mean we cannot infer the proto type from that `com.google.protobuf.Message`?
How can we be sure that expression will return a proper type we have just deserialzed from?

spring-integration-core/src/main/java/org/springframework/integration/transformer/support/ProtoHeaders.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2019 the original author or authors.
```
artembilan, line 2: `2023`

spring-integration-core/src/test/java/org/springframework/integration/transformer/ProtoTest.java
```diff
@@ -0,0 +1,306 @@
...
+import static org.mockito.Mockito.verify;
+
+/**
+ * 
+ * @author Christian Tzolov
+ * @since 6.1
+ */
+@SpringJUnitConfig
+@LogLevels(categories = "foo", level = "DEBUG")
+public class ProtoTest {
```
artembilan, line 53: We prefer to name our test classes with a `Tests` suffix.

spring-integration-core/src/test/resources/proto/ProtoTest.proto
```diff
@@ -0,0 +1,17 @@
...
+
+message TestClass1 {
+  optional string bar = 1;
+  optional int32 qux = 2;
+}
+
+message TestClass2 {
+  optional string bar = 1;
+  optional int32 qux = 2;
+}
```
artembilan, line 17: Need to ensure that every file ends with a new line.
My IDEA can do that for me on save.
Perhaps you can configure VSCode some how to do that, too...

artembilan (changes requested):
Please, see if we can add docs for this new feature into the `transformer.adoc`.
Probably just after `[[Avro-transformers]]` paragraph.
It would be also great to have a bullet in the `whats-new.adoc` in a `=== New Components` section.

Thank you! This is great start!

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,157 @@
...
+		else if (value instanceof String) {
+			try {
+				type = (Class<? extends com.google.protobuf.Message>) ClassUtils.forName((String) value,
+						this.beanClassLoader);
+			}
+			catch (ClassNotFoundException | LinkageError e) {
+				throw new IllegalStateException(e);
+			}
+		}
+		if (type == null) {
```
artembilan, line 116: Well, I'm not fully OK with this logic, since it is misleading when end-user expression returns `null`.
but I see that you borrowed it from a `SimpleFromAvroTransformer`.
So, then OK - let it be!

spring-integration-core/src/main/java/org/springframework/integration/transformer/ToProtobufTransformer.java
artembilan: I think you still have a question here.
So, what I see so far: since `FromProtobufTransformer` relies on a `com.google.protobuf.Message.class` in that `ProtoHeaders.TYPE`, then I don't see a reason in such an extra expression and just provide whatever info is our `payload` based on.
It is probably up to target `FromProtobufTransformer` (or any other consumer) side to decide to rely on our `ProtoHeaders.TYPE` header or use some other  type for Proto deserialization.

WDYT if we just remove this `typeIdExpression` altogether? 

I see that you took this logic from the `SimpleToAvroTransformer`, but I still wonder why would one use different type for header value from the one we have just serialized from?...

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,126 @@
...
+			try {
+				targetClass = (Class<? extends com.google.protobuf.Message>) ClassUtils.forName((String) value,
+						this.beanClassLoader);
+			}
+			catch (ClassNotFoundException | LinkageError e) {
+				throw new IllegalStateException(e);
+			}
+		}
+		if (targetClass == null) {
+			targetClass = this.defaultType;
```
artembilan, line 121: I think we need at least `trace()` when we fallback to the default type.
Kinda say that `typeIdExpression` returned `null` for request message and therefore falling back to the default one.

spring-integration-core/src/main/java/org/springframework/integration/transformer/ToProtobufTransformer.java
```diff
@@ -0,0 +1,57 @@
...
+        Assert.isInstanceOf(com.google.protobuf.Message.class, message.getPayload(),
+                "Payload must be an implementation of 'com.google.protobuf.Message'");
+
+        MessageHeaderAccessor accessor = new MessageHeaderAccessor(message);
+        accessor.setHeader(ProtoHeaders.TYPE, message.getPayload().getClass().getName());
+        if (!message.getHeaders().containsKey(MessageHeaders.CONTENT_TYPE)) {
+            accessor.setContentType(ProtobufMessageConverter.PROTOBUF);
+        }
+
+        return this.protobufMessageConverter.toMessage(message.getPayload(), accessor.getMessageHeaders());
```
artembilan, line 55: How about that `textToBytes` option we have chatted before?

spring-integration-core/src/test/java/org/springframework/integration/transformer/ProtoTests.java
```diff
@@ -0,0 +1,264 @@
...
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+/**
+ *
+ * @author Christian Tzolov
+ * @since 6.1
+ */
+@SpringJUnitConfig
+@LogLevels(categories = "foo", level = "DEBUG")
```
artembilan, line 55: Is it really necessary to still have the logging on debug level?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transformer/ToProtobufTransformer.java
tzolov: I thought about it, but, IMO, the right solution is to override the ProtobufMessageConverter#convertToInternal() method and make it return Message<byte[]> instead of Message<String>.
It would be a waste to get back a Message<String> (from the ProtobufMessageConverter) just rewrite it on the fly into  Message<byte[]>.
What do you think?

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,126 @@
...
+	protected void onInit() {
+		this.evaluationContext = IntegrationContextUtils.getEvaluationContext(getBeanFactory());
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	protected Object doTransform(Message<?> message) {
+
+		Class<? extends com.google.protobuf.Message> targetClass = null;
+		Object value = this.typeIdExpression.getValue(this.evaluationContext, message);
```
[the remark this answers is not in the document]
tzolov: It would have been nice but apparently the SpEL is not that smart. I'm getting:
"EL1001E: Type conversion problem, cannot convert from java.lang.String to java.lang.Class<?>"

build.gradle
```diff
@@ -543,7 +544,8 @@ project('spring-integration-core') {
         optionalApi('com.fasterxml.jackson.module:jackson-module-kotlin') {
             exclude group: 'org.jetbrains.kotlin'
         }
-
+        optionalApi "com.google.protobuf:protobuf-java:$protobufVersion"
+        optionalApi "com.google.protobuf:protobuf-java-util:$protobufVersion"
```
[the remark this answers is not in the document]
tzolov: moved the "com.google.protobuf:protobuf-java-util:$protobufVersion" to testImplementation deps, for the test to work. Also added brief desc in the javadoc.

tzolov:
@artembilan last 2 commits (6 and 7) address the last round to comments.

spring-integration-core/src/main/java/org/springframework/integration/transformer/ToProtobufTransformer.java
tzolov: I've tried to override the default [ProtobufMessageConverter](https://github.com/spring-projects/spring-framework/blob/main/spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java) JSON handling. 
In `ToProtobufTransformer` I added:
```java
    public static class ProtobufMessageConverterEx extends ProtobufMessageConverter {
        @Override
        protected Object convertToInternal(Object payload, MessageHeaders headers, Object conversionHint) {
            Object outputPayload = super.convertToInternal(payload, headers, conversionHint);
            return (outputPayload instanceof String) ? ((String) outputPayload).getBytes() : outputPayload;
        }
    }
```
It does the string-to-byte conversion but then if you try to transform it back (using the `FromProtobufTransformer` and in turn the `ProtobufMessageConverter`) I we get stuck with the 
[ProtobufMessageConverter.ProtobufJavaUtilSupport#merge(...)](https://github.com/spring-projects/spring-framework/blob/88e6544d9d3498359b49a2548a1b6ef52b32a5ad/spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java#L268) that assumes `String` payload only and fails if the payload is `byte[]`.
IMO there is no way to override this behaviour. If we want to pursue this path we need to fix the ProtobufMessageConverter impl. to use the `merge(Reader...` instead of `merge(String ...`

So to me it seems that we should stick with the String json representation, but document it well and explain how to perform `textToBytes` & `bytesToText` transformations as additional processing steps.

artembilan: Yes, fully agree: since there is no way for non-String JSON handling in the `ProtobufMessageConverter`, then for consistency our transformers will stick with the same behavior.

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,124 @@
...
+import org.springframework.integration.context.IntegrationContextUtils;
+import org.springframework.integration.expression.FunctionExpression;
+import org.springframework.integration.transformer.support.ProtoHeaders;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.converter.ProtobufMessageConverter;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+
+/**
+ * A Protocol Buffer transformer to instantiate {@link com.google.protobuf.Message} objects from {@code byte[]}.
```
artembilan, line 31: Is that still correct to say "from `byte[]`" since we have just talked about JSON which is accepted only as as String?
Shouldn't we improve this Javadocs respectively with a `ProtobufMessageConverter` logic?
And still mention that we fully rely on that converter internally.

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,124 @@
...
+			try {
+				targetClass = (Class<? extends com.google.protobuf.Message>) ClassUtils.forName((String) value,
+						this.beanClassLoader);
+			}
+			catch (ClassNotFoundException | LinkageError e) {
+				throw new IllegalStateException(e);
+			}
+		}
+		if (targetClass == null) {
+			this.logger.trace("Empty SpEL type expression. Fallback to the defaultType!");
```
artembilan, line 118: How about this:
```suggestion
			this.logger.trace(() -> "The 'typeIdExpression' (" + this.typeIdExpression + ") returned 'null' for: " + message + ". Falling back to the defaultType: " + this.defaultType);
```
?
I mean to give as much info to end-user as possible to let them to understand the reason behind behavior.

spring-integration-core/src/test/java/org/springframework/integration/transformer/ProtoTests.java
```diff
@@ -0,0 +1,263 @@
...
+                                .isNotSameAs(test);
+                assertThat(received.getHeaders().get("flow")).isEqualTo("flow1");
+                ArgumentCaptor<String> captor = ArgumentCaptor.forClass(String.class);
+                verify(spied, atLeastOnce()).debug(captor.capture());
+                assertThat(captor.getAllValues()).anyMatch(s -> s.contains("preSend on channel"));
+                assertThat(captor.getAllValues()).anyMatch(s -> s.contains("postSend (sent=true) on channel"));
+        }
+
+        @Test
+        @LogLevels(classes = DirectChannel.class, categories = "bar", level = "DEBUG")
```
artembilan, line 85: I still see these on method.
What do you have in mind, please, to keep them?

src/reference/asciidoc/transformer.adoc
```diff
@@ -446,6 +446,61 @@ If the expression returns `null`, the `defaultType` is used.
...
 This allows decoupling of the producer and consumer where the sender can set the header to some token representing the type and the consumer then maps that token to a type.
 
+[[Protobuf-transformers]]
+===== Protocol Buffers Transformers
+
+Version 6.1 adds support for transforming from and to https://protobuf.dev/[Protocol Buffers] data content.
+
+The `ToProtobufTransformer` transforms `com.google.protobuf.Message` message payloads into native byte array or json text payloads.
+The `application/x-protobuf` content type (default) produces byte array output payload.
+Set the `application/json` content type and add the `com.google.protobuf:protobuf-java-util` on the classpath to output text message payload. If the content type header is not set the `ToProtobufTransformer` defaults to `application/x-protobuf`.
```
artembilan, line 456: We prefer one line per sentence: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

src/reference/asciidoc/transformer.adoc
```diff
@@ -446,6 +446,61 @@ If the expression returns `null`, the `defaultType` is used.
...
+
+The `FromProtobufTransformer` transforms byte array or text protobuf payload (depending on the content type) back into `com.google.protobuf.Message` instances. You can also specify a `SpEL expression` to determine the type to deserialize using the `setTypeExpression` method.
```
artembilan, line 458: I don't think it make sense to wrap "SpEL expression" as a code...

src/reference/asciidoc/transformer.adoc
```diff
@@ -446,6 +446,61 @@ If the expression returns `null`, the `defaultType` is used.
...
+The default SpEL expression is `headers[proto_type]` (`ProtoHeaders.TYPE`) which is populated by the `ToProtobufTransformer` with the fully qualified class name of the source `com.google.protobuf.Message` class. If the expression returns `null`, the `defaultType` is used.
+
+For example, compiling the following IDL
```
artembilan, line 461: Colon in the end.

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/transformer/ProtoTests.java
tzolov: I've borrowed this form the Simple Avro Transfomer and apparently it is requred for the `LogAccessor spied` logic and verification to work. 
Apparently it uses reflection to substitute the parent IntegrationObjectSupport#logger with a spy instance and then verifies that the loger `debug` is beaing used at lest once. Do you know why this is important? otherwise i will remove all this logger tests.

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
tzolov: More descriptive trace message is better, but I start to think that we should remove it all together. Here is why: 

The `typeIdExpression` and the `proto_type` msg header only make sence when the message was encoded via the `ToProtobufTransofmer`. But the `FromProbufTransformer` can be used to de-ser protobuf content creted with custom `ProtobufMessageConverter` implementations, 3rd party tools and even different programming language. In this case the `defaultType` will be the normal way to configure and use the `FromProtobufTransoformer`. Logging trace messges in this case would be incorrect, anoying and could affect the performance.

spring-integration-core/src/test/java/org/springframework/integration/transformer/ProtoTests.java
artembilan: OK. I see what you are talking about the spying on the `logger`.
And why do we need to do that in this our tests?
Isn't that enough to be sure that we receive from transformers whatever we expect?
Why do we need to verify those `preSend on channel` and `postSend (sent=true) on channel`?
How is that logic relevant to Proto?

src/reference/asciidoc/transformer.adoc
artembilan: Doesn't look like you have resolved this and other concerns in docs.

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
artembilan: After thinking about this more I would suggest this behavior:
1. Change the `defaultType` from ctor option to the `expectedType` setter.
2. Have a ctor free from the type at all.
3. Still have an `expectedTypeExpression` and with that default value.
4. Have a setter for it anyway.
5. There is a `ValueExpression` for just constant `expectedType`.
6. If expression returns `null` we fail with an exception.

This way end-user by default can use this transformer against a header.
If other non-standard behavior is expected, you are free to override that expression or just set a static type.
My point is to avoid some unexpected logic, when we fallback to something what might not be relevant to the request at all.
You are free to provide a complex expression which would do a fallback for you, but that's already out of this transformer responsibility.
Another argument is to not require some default type when we are going to deal only with types resolved through expression.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,160 @@
...
+		}
+		if (targetClass == null) {
+			targetClass = this.expectedType;
+		}
+
+		if (targetClass == null) {
+			this.logger.error(() -> "The 'expectedTypeExpression' (" + this.expectedTypeExpression
+					+ ") returned 'null' for: "
+					+ message + ". No falling back expectedType is configured. Consider setting the expectedType.");
+			throw new RuntimeException("The 'expectedTypeExpression' (" + this.expectedTypeExpression
```
artembilan, line 149: The exception must be `MessageTransformationException` to shorten stack trace a little bit upstream.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/transformer/FromProtobufTransformer.java
```diff
@@ -0,0 +1,143 @@
...
+		if (targetClass == null) {
+			throw new MessageTransformationException("The 'expectedTypeExpression' (" + this.expectedTypeExpression
+					+ ") returned 'null' for: "
+					+ message + ". No falling back expectedType is configured. Consider setting the expectedType.");
+		}
+
+		return this.protobufMessageConverter.fromMessage(message, targetClass);
+	}
+
+	public void setEvaluationContext(EvaluationContext evaluationContext) {
```
artembilan, line 140: Well, no one must do this.
Whatever we have in the `onInit()` is enough.

src/reference/asciidoc/transformer.adoc
```diff
@@ -446,6 +446,62 @@ If the expression returns `null`, the `defaultType` is used.
...
+===== Protocol Buffers Transformers
+
+Version 6.1 adds support for transforming from and to https://protobuf.dev/[Protocol Buffers] data content.
+
+The `ToProtobufTransformer` transforms `com.google.protobuf.Message` message payloads into native byte array or json text payloads.
+The `application/x-protobuf` content type (default) produces byte array output payload.
+Set the `application/json` content type and add the `com.google.protobuf:protobuf-java-util` on the classpath to output text message payload.
+If the content type header is not set the `ToProtobufTransformer` defaults to `application/x-protobuf`.
+
+The `FromProtobufTransformer` transforms byte array or text protobuf payload (depending on the content type) back into `com.google.protobuf.Message` instances. You can also specify a SpEL expression to determine the type to deserialize using the `setTypeExpression` method.
```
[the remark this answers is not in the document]
artembilan: Was not addressed.

artembilan (changes requested):

src/reference/asciidoc/transformer.adoc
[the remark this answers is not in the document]
tzolov: addressed

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/451374dd119551348fbae96ab9955797207a5327 after some simple code clean up and fixing some typos in docs.

Thank you!

PR #8551, opened 2023-02-14
GH-8550: MQTT: Always re-subscribe on re-connect

### LINKED ISSUE #8550: MQTT is not re-subscribing after reconnect

**6.0.2 with Spring Boot 3.0.2**

**Describe the bug**

When the client loses connection (and auto-reconnect is on) and then reconnects, existing subscriptions are not re-subscribbed. Application restart is needed to re-subscribe.

**To Reproduce**

- Start the sample application
- Connect to the MQTT broker
- Send message
- Stop the MQTT broker
- Start the MQTT broker
- Send message

After the reconnection, no message will appear in the sample application.

**Expected behavior**

I expect to re-subscribe to the topic and receive messages

**Sample**

The sample includes MQTT broker in the `infra` folder.

[Sample Here](https://github.com/iamceph/spring-integration-mqtt-bug).

[Original thread on StackOverflow](https://stackoverflow.com/questions/75404056/spring-integration-mqtt-not-re-subscribing-after-reconnect)

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8550

Turns out the Paho MQTT client does not re-subscribe when connection re-established on automatic reconnection

* Fix `AbstractMqttMessageDrivenChannelAdapter` to always subscribe to their topics in the `connectComplete()` independently of the `reconnect` status
* Verfiy behavior with `MOSQUITTO_CONTAINER` image restart in Docker

**Cherry-pick to `5.5.x`**


artembilan:
A fully different solution is needed for `5.5.x` branch (if any)

PR #8552, opened 2023-02-15
GH-4014: MQTT ClientManager: completion timeouts

### LINKED ISSUE #4014: Reusing of MQTT connection (clientId) results in error log "... Never happens."
This issue is slightly realated to GH-3685 (https://github.com/spring-projects/spring-integration/issues/3685).

I am trying to use one MQTT connection with a single clientId for subscriber as well as publisher channels.

Problem is that in <tt>MqttConnectOptions</tt> is the timeout specified in *seconds* while in <tt>org.eclipse.paho.client.mqttv3.MqttToken.waitForCompletion(long)</tt> is expected in *milliseconds*.

<pre>
MqttConnectOptions.setConnectionTimeout(int seconds)
client.connect(this.connectionOptions).waitForCompletion(long milliseconds)
</pre>

So calling:

<pre>
Mqttv3ClientManager.start() {
  ...
  client.connect(this.connectionOptions).waitForCompletion(this.connectionOptions.getConnectionTimeout());
  ...
}
</pre>

times out very quickly (default seems to be 30 seconds but they actually time out in 30 milliseconds) and an attempt to call the <tt>reconnect()</tt> method is made even though the connection is still becoming connected. That results in:

<pre>
17:17:04,334 ERROR [main] o.s.i.m.c.Mqttv3ClientManager:110 MQTT client failed to connect. Never happens.
Connect already in progress (32110)
        at org.eclipse.paho.client.mqttv3.MqttAsyncClient.reconnect(MqttAsyncClient.java:1399)
        at org.springframework.integration.mqtt.core.Mqttv3ClientManager.start(Mqttv3ClientManager.java:107)
        at org.springframework.context.support.DefaultLifecycleProcessor.doStart(DefaultLifecycleProcessor.java:178)
</pre>

This just prints the error message in log, but the connection seems to be established without further problems and the application works.

### LINKED ISSUE #8550: MQTT is not re-subscribing after reconnect

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/4014

The `ClientManager` implementations uses by mistake a `connectionTimeout` for operations with completion wait

* Introduce `completionTimeout` and `disconnectCompletionTimeout` for `ClientManager` abstraction to realign the logic with existing channel adapters and Paho Client by itself.
* Deprecate `DEFAULT_COMPLETION_TIMEOUT` and `DISCONNECT_COMPLETION_TIMEOUT` constants in the `AbstractMqttMessageDrivenChannelAdapter` in favor of respective replacement in the `ClientManager`
* Pull `disconnectCompletionTimeout` property from the `MqttPahoMessageDrivenChannelAdapter` to its superclass
* Use new `disconnectCompletionTimeout` in the `Mqttv5PahoMessageDrivenChannelAdapter` for similar `disconnectForcibly()` call
* Fix Lifecycle race condition when `ClientManager` is started by the outbound channel adapter (`Integer.MIN_VALUE` phase and auto-startup - see `DefaultLifecycleProcessor.doStart()` and logic around `dependencies`) which is much earlier than `MessageProducerSupport` (`Integer.MAX_VALUE / 2` phase) and there a `connectComplete()` callback might be called before the `MessageProducerSupport.start()`. For that purpose check for an `isRunning()` in the `connectComplete()` before subscribing and set `readyToSubscribeOnStart` flag to `subscribe()` in a `doStart()` of this `MqttMessageDrivenChannelAdapter`
* Remove redundant `MqttPahoMessageDrivenChannelAdapter.cleanSession` property in favor of `this.clientFactory.getConnectionOptions().isCleanSession()` call

GH-8550: MQTT: Always re-subscribe on re-connect

Fixes https://github.com/spring-projects/spring-integration/issues/8550

Turns out the Paho MQTT client does not re-subscribe when connection re-established on automatic reconnection

* Fix `AbstractMqttMessageDrivenChannelAdapter` to always subscribe to their topics in the `connectComplete()` independently of the `reconnect` status
* Verify behavior with `MOSQUITTO_CONTAINER` image restart in Docker

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2022 the original author or authors.
+ * Copyright 2022-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -53,6 +53,10 @@ public abstract class AbstractMqttClientManager<T, C> implements ClientManager<T
 
 	private int phase = DEFAULT_MANAGER_PHASE;
 
+	private long completionTimeout = ClientManager.DEFAULT_COMPLETION_TIMEOUT;
+
+	private long disconnectCompletionTimeout = ClientManager.DISCONNECT_COMPLETION_TIMEOUT;
+
 	private boolean manualAcks;
 
 	private ApplicationEventPublisher applicationEventPublisher;
@@ -96,6 +100,34 @@ public abstract class AbstractMqttClientManager<T, C> implements ClientManager<T
 		return this.connectCallbacks;
 	}
 
+	/**
+	 * Set the completion timeout for operations.
+	 * Default {@value #DEFAULT_COMPLETION_TIMEOUT} milliseconds.
+	 * @param completionTimeout The timeout.
+	 * @since 6.0.3
+	 */
+	public void setCompletionTimeout(long completionTimeout) {
+		this.completionTimeout = completionTimeout;
+	}
+
+	protected long getCompletionTimeout() {
+		return this.completionTimeout;
+	}
+
+	/**
+	 * Set the completion timeout when disconnecting.
+	 * Default {@value #DISCONNECT_COMPLETION_TIMEOUT} milliseconds.
+	 * @param completionTimeout The timeout.
+	 * @since 6.0.3
+	 */
+	public synchronized void setDisconnectCompletionTimeout(long completionTimeout) {
+		this.disconnectCompletionTimeout = completionTimeout;
+	}
+
+	protected long getDisconnectCompletionTimeout() {
+		return this.disconnectCompletionTimeout;
+	}
+
 	@Override
 	public boolean isManualAcks() {
 		return this.manualAcks;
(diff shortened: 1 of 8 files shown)


artembilan:
This PR also includes the fix for https://github.com/spring-projects/spring-integration/issues/8550.
Otherwise an existing PR for that one is in conflict.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
```diff
@@ -53,6 +53,10 @@
 
 	private int phase = DEFAULT_MANAGER_PHASE;
 
+	private long completionTimeout = ClientManager.DEFAULT_COMPLETION_TIMEOUT;
```
garyrussell, line 56: Consider using the more modern `Duration`?

garyrussell (commented):
Just one comment/suggestion. Will merge as-is if you prefer.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
artembilan: I will, when the target client lib does, but now for consistency I would stay with the same type the target API expects from us.

garyrussell (approved):

PR #8554, opened 2023-02-16
Gh 4014

### LINKED ISSUE #4014: Reusing of MQTT connection (clientId) results in error log "... Never happens."

aretap:
GH-4014: Fix inconsistent usage of MQTT timeouts

Fixes https://github.com/spring-projects/spring-integration/issues/4014

Also fixes error log with text "... Never happens."

* Introduce `completionTimeout` into a `ClientManager` abstraction
and fix inappropriate usages of the `connectionTimeout` over there.

* Deprecate the `disconnectCompletionTimeout` in favor of
`completionTimeout` for consistency with other similar places.

* Update MqttAdapterTests.testDifferentQos JUnit test.

* TODO: There is still a problematic JUnit test ClientManagerBackToBackTests.testV3ClientManagerReconnect. It passes when executed alone, but fails when executed together with other tests. It seems to me that there is some race condition or improper Spring context initialization. I am not able to sort this out. The test used to pass before this fix, but that was "a pass by an accident" as it logged the problematic exception  "... Never happens" and then passed.

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
@@ -43,6 +43,13 @@ import org.springframework.util.Assert;
  */
 public abstract class AbstractMqttClientManager<T, C> implements ClientManager<T, C>, ApplicationEventPublisherAware {
 
+	/**
+	 * The default completion timeout in milliseconds.
+	 * @since 6.0.3
+	 * @author Jiří Patera
+	 */
+	public static final long DEFAULT_COMPLETION_TIMEOUT = 30_000L;
+
 	protected final Log logger = LogFactory.getLog(this.getClass()); // NOSONAR
 
 	private static final int DEFAULT_MANAGER_PHASE = 0;
@@ -62,6 +69,8 @@ public abstract class AbstractMqttClientManager<T, C> implements ClientManager<T
 	private String beanName;
 
 	private T client;
+
+	private long completionTimeout = DEFAULT_COMPLETION_TIMEOUT;
 
 	protected AbstractMqttClientManager(String clientId) {
 		Assert.notNull(clientId, "'clientId' is required");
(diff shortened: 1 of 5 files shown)


aretap:
I see that I ran into a partial conflict with somebody else who was also solving this issue...

artembilan:
Sorry, this has been fixed via https://github.com/spring-projects/spring-integration/pull/8552.
And yes: I also noticed that test failure and fixed it in that PR as well.

Thank you for your effort any way!
There are still a lot of issues you can take for contribution though

PR #8560, opened 2023-02-22
GH-3957: Add JmsInboundGateway.replyToExpression

### LINKED ISSUE #3957: Jms InboundGateway to support Dynamic reply queue
Jms InboundGateway via DSL should support dynamic destination queue based on Jms or custom message headers

Currently it does not support dynamic destination queue for a Jms InboundGateway. Creating this feature request based on Artem's comment on the below SO question

https://stackoverflow.com/questions/74617227/spring-integration-jms-inboundgateway-dynamic-reply-queue/74617452

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3957

Sometimes we cannot use a standard `JmsReplyTo` property for sending replies from the server. A `DestinationResolver` API does not have access to the request message.

* Introduce a `ChannelPublishingJmsMessageListener.replyToExpression` property to evaluate a reply destination against request JMS `Message`
* Use this expression only of no `JmsReplyTo` property
* Expose this property on Java DSL level
* To simplify end-user experience with lambda configuration for this property, introduce a `CheckedFunction` which essentially re-throws exception "sneaky" way

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/CheckedFunction.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/CheckedFunction.java
@@ -21,6 +21,9 @@ import java.util.function.Function;
 /**
  * A Function-like interface which allows throwing Error.
  *
+* @param <T> the input type.
+* @param <R> the output type.
+ *
  * @author Artem Bilan
  *
  * @since 6.1
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/CheckedFunction.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/CheckedFunction.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.util;
+
+import java.util.function.Function;
+
+/**
+ * A Function-like interface which allows throwing Error.
+ *
+ * @author Artem Bilan
+ *
+ * @since 6.1
+ */
+@FunctionalInterface
+public interface CheckedFunction<T, R> {
+
+	R apply(T t) throws Throwable;
+
+	default Function<T, R> unchecked() {
+		return t1 -> {
+			try {
+				return apply(t1);
+			}
+			catch (Throwable t) {
+				return sneakyThrow(t);
+			}
+		};
+	}
+
+	@SuppressWarnings("unchecked")
+	private static <T extends Throwable, R> R sneakyThrow(Throwable t) throws T {
+		throw (T) t;
+	}
+
+}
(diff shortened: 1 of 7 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/CheckedFunction.java
```diff
@@ -0,0 +1,52 @@
...
+	default Function<T, R> unchecked() {
+		return t1 -> {
+			try {
+				return apply(t1);
+			}
+			catch (Throwable t) {
+				return sneakyThrow(t);
+			}
+		};
+	}
```
garyrussell, line 45: I am not really comfortable with institutionalizing this JVM abuse. Is there no alternative?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/CheckedFunction.java
artembilan: OK. Will it be better if there is no that `unchecked()` then?
Something like plain:
```
@FunctionalInterface
public interface CheckedFunction<T, R> {

	R apply(T t) throws Throwable;

}
```
?

garyrussell: Right; that's the bit I was objecting to. You probebly need a `// NOSONAR` here too.

spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
```diff
@@ -440,6 +458,24 @@ private Destination getReplyDestination(jakarta.jms.Message request, Session ses
 			}
 		}
 		return replyTo;
+
+	}
+
+	@Nullable
+	private Destination resolveReplyTo(jakarta.jms.Message request, Session session) throws JMSException {
```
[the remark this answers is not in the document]
artembilan: I don't think so. Then it is probably better to make the whole `getReplyDestination()` as `protected`.
And then we will dive to a mess what is `protected` and what is not and can be accessed from the inheritors.
If you wish, we can revise this in a separate issue.

spring-integration-jms/src/main/java/org/springframework/integration/jms/dsl/JmsInboundGatewaySpec.java
```diff
@@ -137,6 +141,44 @@ public S destinationResolver(DestinationResolver destinationResolver) {
...
+	 * Set a function to resolve a 'replyTo' destination from a request {@link jakarta.jms.Message}
+	 * as a root evaluation object if {@link jakarta.jms.Message#getJMSReplyTo()} is null.
+	 * @param replyToFunction the function for 'replyTo' destination.
+	 * @return the spec.
+	 * @since 6.1
+	 * @see ChannelPublishingJmsMessageListener#setReplyToExpression(Expression)
+	 */
+	public S replyToFunction(CheckedFunction<Message, ?> replyToFunction) {
+		return replyToExpression(new FunctionExpression<>(replyToFunction.unchecked()));
+	}
```
garyrussell, line 167: Why do we need this; why not just

```suggestion
	public S replyToFunction(Function<Message, ?> replyToFunction) {
		return replyToExpression(new FunctionExpression<>(replyToFunction));
	}
```

??

artembilan: Sure, but then `.replyToFunction(message -> message.getStringProperty("myReplyTo"))` is not going to be clean and end-user has to handle that `JMSException` himself which is not runtime one.

garyrussell: So, you are assuming that the user will use a JMS operation to determine the destination? I guess that's reasonable given that the root object is a `Message`.

OK.

PR #8564, opened 2023-02-28
GH-8562: Fix streaming source for remote calls

### LINKED ISSUE #8562: Allow AbstractRemoteFileStreamingMessageSource#remoteFileToMessage Retry
**Expected Behavior**
When using `SmbStreamingMessageSource`, in the doReceive() function, should be able to retry remoteFileToMessage() call with a `RetryTemplate` that can be set via `SmbStreamingInboundChannelAdapterSpec`.

**Current Behavior**
`AbstractRemoteFileStreamingMessageSource#remoteFileToMessage` is only called once.  Network errors can cause the call to fail when resolving `SmbFileInfo` and drop the file from being processed.

**Context**
I wrote an integration flow that streams/processes zip files from a shared folder in a defined order.  Occasionally, the process hits a network error when resolving the `SmbFileInfo` for the remote file and gets dropped from processing.  Handling the messageException in the errorChannel will cause the zip files to be processed out of order.

Could not simply extend `SmbStreamingMessageSource` and add a `RetryTemplate` because the class properties used by the doReceive() function are private.  I had to copy/paste `AbstractRemoteFileStreamingMessageSource` into my project, add a `RetryTemplate` and modify the call to remoteFileToMessage (line 214).

return remoteFileToMessage(file);

->

var processFile = file;
return retry.execute(ctx -> {
  if (ctx.getRetryCount() > 0)
    logger.info(ctx::toString);
  return remoteFileToMessage(processFile);
});

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8562

The `AbstractRemoteFileStreamingMessageSource.doReceive()` takes files first from a `toBeReceived` queue. When `AbstractRemoteFileStreamingMessageSource.remoteFileToMessage()` fails to fetch the file content because of interim connection issue, we reset this file from a filter and rethrow an exception. The next `receive()` call will just go ahead to the next entry in the `toBeReceived` queue, but the file we have just failed for will be retried only on the next list call to the remove directory. This essentially breaks a possible in-order target application logic.

* Introduce `AbstractRemoteFileStreamingMessageSource.strictOrder` option to clear the `toBeReceived` queue when we fail in the `remoteFileToMessage()`, so the next `receive()` call would re-fetch files from remote dir, because the filter has been reset for those files.
* Fix `AbstractFileInfo.toString()` to not perform remote calls when we just log this file. For example, we reset the file for connection failure and log the message about it, but it fails again because we request `size` of the file which may require a remote connection.

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractFileInfo.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractFileInfo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,8 +16,6 @@
 
 package org.springframework.integration.file.remote;
 
-import java.util.Date;
-
 import org.springframework.integration.json.SimpleJsonSerializer;
 
 /**
@@ -27,6 +25,7 @@ import org.springframework.integration.json.SimpleJsonSerializer;
  * @param <F> The target protocol file type.
  *
  * @author Gary Russell
+ * @author Artem Bilan
  *
  * @since 2.1
  */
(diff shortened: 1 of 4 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractFileInfo.java
```diff
@@ -57,10 +56,9 @@ public String toJson() {
 
 	@Override
 	public String toString() {
-		return "FileInfo [isDirectory=" + isDirectory() + ", isLink=" + isLink()
-				+ ", Size=" + getSize() + ", ModifiedTime="
-				+ new Date(getModified()) + ", Filename=" + getFilename()
-				+ ", RemoteDirectory=" + getRemoteDirectory() + ", Permissions=" + getPermissions() + "]";
+		return "FileInfo [isLink=" + isLink()
+				+ ", Filename=" + getFilename()
+				+ ", RemoteDirectory=" + getRemoteDirectory() + "]";
```
garyrussell, line 61: As discussed out of bounds; it would be better to leave this as-is and overload in any subclass that needs this change.

garyrussell (changes requested):

garyrussell:
..and cherry-picked to 5.5.x as efab65b2fa after resolving conflicts.

garyrussell:
...and to 6.0.x as d9f3821b9b

PR #8568, opened 2023-03-02
Reduce instance supplier use to appease AOT

onobc:
* Remove the use of instance suppliers on bean definitions that are processed during the AOT phase.

* The remaining areas that use instance suppliers do so at runtime and do not use reflection, but instead
are passed the configured bean to register.

See https://github.com/spring-cloud/spring-cloud-stream/issues/2655

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -70,6 +70,7 @@ import org.springframework.util.ClassUtils;
  * @author Gary Russell
  * @author Michael Wiles
  * @author Pierre Lakreb
+ * @author Chris Bono
  *
  * @see IntegrationContextUtils
  */
@@ -153,14 +154,14 @@ public class DefaultConfiguringBeanFactoryPostProcessor
 	private void registerBeanFactoryChannelResolver() {
 		if (!this.beanFactory.containsBeanDefinition(ChannelResolverUtils.CHANNEL_RESOLVER_BEAN_NAME)) {
 			this.registry.registerBeanDefinition(ChannelResolverUtils.CHANNEL_RESOLVER_BEAN_NAME,
-					new RootBeanDefinition(BeanFactoryChannelResolver.class, BeanFactoryChannelResolver::new));
+					new RootBeanDefinition(BeanFactoryChannelResolver.class));
 		}
 	}
 
 	private void registerMessagePublishingErrorHandler() {
 		if (!this.beanFactory.containsBeanDefinition(ChannelUtils.MESSAGE_PUBLISHING_ERROR_HANDLER_BEAN_NAME)) {
 			this.registry.registerBeanDefinition(ChannelUtils.MESSAGE_PUBLISHING_ERROR_HANDLER_BEAN_NAME,
-					new RootBeanDefinition(MessagePublishingErrorHandler.class, MessagePublishingErrorHandler::new));
+					new RootBeanDefinition(MessagePublishingErrorHandler.class));
 		}
 	}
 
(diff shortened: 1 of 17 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
```diff
@@ -173,8 +174,7 @@ private void registerNullChannel() {
 			BeanDefinition nullChannelDefinition = null;
 			BeanFactory beanFactoryToUse = this.beanFactory;
 			do {
-				if (beanFactoryToUse instanceof ConfigurableListableBeanFactory) {
-					ConfigurableListableBeanFactory listable = (ConfigurableListableBeanFactory) beanFactoryToUse;
+				if (beanFactoryToUse instanceof ConfigurableListableBeanFactory listable) {
```
onobc, line 177: **[UNRELATED]** I agreed w/ IDEA suggestion

spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
```diff
@@ -421,8 +420,7 @@ private boolean noBeanPresentForComponent(Object instance, String parentBeanName
 		}
 
 		return !this.beanFactory.getBeansOfType(instance.getClass(), false, false)
-				.values()
-				.contains(instance);
+				.containsValue(instance);
```
onobc, line 423: **[UNRELATED]** I like this new API

spring-integration-core/src/main/java/org/springframework/integration/config/AbstractMethodAnnotationPostProcessor.java
```diff
@@ -577,7 +577,7 @@ protected AbstractEndpoint createEndpoint(MessageHandler handler, @SuppressWarni
 			catch (DestinationResolutionException e) {
 				if (e.getCause() instanceof NoSuchBeanDefinitionException) {
 					this.definitionRegistry.registerBeanDefinition(inputChannelName,
-							new RootBeanDefinition(DirectChannel.class, DirectChannel::new));
+							new RootBeanDefinition(DirectChannel.class));
```
artembilan, line 580: Revert this one, please: it is used at runtime and not in AOT at all.

spring-integration-core/src/main/java/org/springframework/integration/config/InboundChannelAdapterAnnotationPostProcessor.java
```diff
@@ -122,12 +121,12 @@ public Object postProcess(Object bean, String beanName, Method method, List<Anno
...
-		methodInvokingMessageSource.setObject(bean);
-		methodInvokingMessageSource.setMethod(method);
+		BeanDefinition beanDefinition = BeanDefinitionBuilder.rootBeanDefinition(MethodInvokingMessageSource.class)
+				.addPropertyValue("object", bean)
+				.addPropertyValue("method", method)
+				.getBeanDefinition();
 		String messageSourceBeanName = generateHandlerBeanName(beanName, method);
-		getDefinitionRegistry().registerBeanDefinition(messageSourceBeanName,
-				new RootBeanDefinition(MethodInvokingMessageSource.class, () -> methodInvokingMessageSource));
+		getDefinitionRegistry().registerBeanDefinition(messageSourceBeanName, beanDefinition);
```
artembilan, line 129: Same here: not AOT, but rather runtime stuff.

spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
```diff
@@ -408,8 +408,7 @@ private boolean noBeanPresentForComponent(Object instance, String parentBeanName
 						&& !instance.equals(this.beanFactory.getBean(beanName))) {
 
 					AbstractBeanDefinition beanDefinition =
-							BeanDefinitionBuilder.genericBeanDefinition((Class<Object>) instance.getClass(),
-											() -> instance)
+							BeanDefinitionBuilder.genericBeanDefinition(instance.getClass())
```
artembilan, line 411: Revert this one. You essentially making a bug here: a bean definition is going to be based on a new instance of the class, but not provided one as intention.
More over this class does its stuff at runtime, so the change is redundant.

artembilan (changes requested):
Thank you, Chris!
Just a couple requests to restore existing behavior in your change.
The mostly important concern that we would like to deal with existing instances in those bean definitions, but not creating new as it would be with a default class-based configuration.

artembilan:
... and cherry-picked to `6.0.x`

PR #8570, opened 2023-03-14
Some `LockRegistryLeaderInitiator` improvements

artembilan:
It is better to not go to the target lock provider at all if the current thread is already interrupted.

* Check for the `Thread.currentThread().isInterrupted()` in the `while` loop and `restartSelectorBecauseOfError()` immediately without checking for a lock
* Fix some other simple typos in the `LockRegistryLeaderInitiator`

**Cherry-pick to `6.0.x`**


artembilan:
Do not merge yet: looking into `yielding` logic which is not always OK when we just cancel the `Future`...

garyrussell (approved):

garyrussell:
...and cherry-picked as 2adf33467c

PR #8572, opened 2023-03-15
Add check of RedisLock#getLockedAt when lock is still not acquired

AntonGabov:
### Description
Fix race condition, when methods `RedisLockRegistry#expireUnusedOlderThan` and `RedisLockRegistry#obtain` are executed successively. 

### Fix
It's possible to delete the lock from `RedisLockRegistry#expireUnusedOlderThan` method, when lock is created but is not acquired (`RedisLock#getLockedAt = 0`)
It can lead to the situation, when `RedisLockRegistry#obtain` returns multiple locks with the same redis-key, which shouldn't happen at all.

### Actual result
`RedisLockRegistry#obtain` may return different lock instances for the same key after invoked method `RedisLockRegistry#expireUnusedOlderThan`.

### Expected result
`RedisLockRegistry#obtain` must return the same lock instance for the same key after invoked method `RedisLockRegistry#expireUnusedOlderThan`.

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -235,7 +235,11 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 			this.locks.entrySet()
 					.removeIf(entry -> {
 						RedisLock lock = entry.getValue();
-						return now - lock.getLockedAt() > age && !lock.isAcquiredInThisProcess();
+						long lockedAt = lock.getLockedAt();
+						return now - lockedAt > age
+								// 'lockedAt = 0' means that the lock is still not acquired!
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Please, add your name to the `@author` list of affected classes.
And that's it: I'll merge it then.

Thank you!

AntonGabov:
> Please, add your name to the `@author` list of affected classes. And that's it: I'll merge it then.
> 
> Thank you!

Done!

@artembilan also is it possible to backport this change to 5.5.x version?

artembilan:
... and cherry-picked to `6.0.x`.
Back-ported to `5.5.x` as https://github.com/spring-projects/spring-integration/commit/bdce2b9c8f5ef2f0eb240bf194f701f570e303b7 after fixing conflicts.

@AntonGabov ,

`5.5.17` is going to be released next week.

AntonGabov:
Wow, already next week! It's really fast!

@artembilan thank you!

PR #8579, opened 2023-03-20
Warn about dropped message in filter

artembilan:
Buy default the `MessageFilter` just drops a discarded message silently. If a request-reply gateway is used upstream, then it becomes unclear why the flow sometimes doesn't work.

* Add a waring log ot emit for default behavior. This still doesn't fix the request-reply problem, but at least it can give a clue what is going on

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/filter/MessageFilter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/filter/MessageFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.context.Lifecycle;
 import org.springframework.core.convert.ConversionService;
+import org.springframework.core.log.LogMessage;
 import org.springframework.integration.IntegrationPatternType;
 import org.springframework.integration.MessageRejectedException;
 import org.springframework.integration.core.MessageSelector;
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/filter/MessageFilter.java
```diff
@@ -179,13 +180,16 @@ protected Object doHandleRequestMessage(Message<?> message) {
...
-				this.messagingTemplate.send(channel, message);
+			MessageChannel channelToDiscard = getDiscardChannel();
+			if (channelToDiscard != null) {
+				this.messagingTemplate.send(channelToDiscard, message);
 			}
 			if (this.throwExceptionOnRejection) {
 				throw new MessageRejectedException(message, "message has been rejected in filter: " + this);
 			}
+			else if (channelToDiscard == null) {
+				logger.warn(LogMessage.format("The message [%s] has been rejected in filter: %s", message, this));
```
garyrussell, line 191: WARN seems too strong; this might be the desired behavior; I would suggest DEBUG, perhaps with an option to increase the log level.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/filter/MessageFilter.java
garyrussell: I have no objection to requiring `nullChannel`, if it's only a 6.1 change.

artembilan: Yes, it is. Let me add such a sentence into doc and you are feel free to merge thereafter!

src/reference/asciidoc/filter.adoc
```diff
@@ -88,6 +88,10 @@ If you want rejected messages to be routed to a specific channel, provide that r
 ----
 ====
 
+If the `throwExceptionOnRejection == false` and no `discardChannel` provided, the message is silently dropped and an `o.s.i.filter.MessageFilter` instance just emits a warning into logs (starting with version 6.1) about this discarded message.
+To fully drop the message off and even lose the mentioned warning in logs, a `nullChannel` can be configured as a `discardChannel` on the filter.
+This way, the goal of the framework is to never be silent by default until an explicit option is set.
```
garyrussell, line 93:
```suggestion
If the `throwExceptionOnRejection == false` and no `discardChannel` is provided, the message is silently dropped and an `o.s.i.filter.MessageFilter` instance just emits a warning log message (starting with version 6.1) about this discarded message.
To drop the message with no warning in the logs, a `NullChannel` can be configured as the `discardChannel` on the filter.
The goal of the framework is to not be completely silent, by default, requiring an explicit option to be set, if that is the desired behavior.
```

garyrussell (changes requested):

garyrussell (approved):

PR #8580, opened 2023-03-21
Do not block by default

artembilan:
Currently, many timeouts in the project are like `-1` or other negative value with a meaning to wait indefinitely.

According to distributed systems design and bad demo developing experience it is not OK to block forever.

* Rework most of the timeouts in the framework to be `30` seconds. Only one remained as `1` seconds is a `PollingConsumer` where it is better to not block even for those 30 seconds when no messages in the queue, but let the polling task be rescheduled.
* Remove the `MessagingGatewaySupport.replyTimeout` propagation down to the `PollingConsumer` correlator where it was a `-1` before and blocked the polling thread on the `Queue.poll()`.
This fixed the problem with a single thread in a pool for auto-configured `TaskScheduler`. Now with 1 seconds wait time we are able to switch to other scheduled tasks even with only 1 thread in the pool

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -782,14 +782,9 @@ public class GatewayProxyFactoryBean<T> extends AbstractEndpoint
 		Expression requestTimeout = this.defaultRequestTimeout;
 
 		if (gatewayAnnotation != null) {
-			/*
-			 * INT-2636 Unspecified annotation attributes should not
-			 * override the default values supplied by explicit configuration.
-			 * There is a small risk that someone has used Long.MIN_VALUE explicitly
-			 * to indicate an indefinite timeout on a gateway method and that will
-			 * no longer work as expected; they will need to use, say, -1 instead.
-			 */
-			if (requestTimeout == null || gatewayAnnotation.requestTimeout() != Long.MIN_VALUE) {
+			if (requestTimeout == null ||
+					gatewayAnnotation.requestTimeout() != IntegrationContextUtils.DEFAULT_TIMEOUT) {
+
 				requestTimeout = new ValueExpression<>(gatewayAnnotation.requestTimeout());
 			}
 			if (StringUtils.hasText(gatewayAnnotation.requestTimeoutExpression())) {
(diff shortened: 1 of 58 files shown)


spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -968,31 +956,27 @@ private void channels(@Nullable String requestChannelName, @Nullable String repl
...
 	private void timeouts(@Nullable Expression requestTimeout, @Nullable Expression replyTimeout,
 			GatewayMethodInboundMessageMapper messageMapper, MethodInvocationGateway gateway) {
-		if (requestTimeout == null) {
-			gateway.setRequestTimeout(-1);
-		}
-		else if (requestTimeout instanceof ValueExpression) {
-			Long timeout = requestTimeout.getValue(Long.class);
-			if (timeout != null) {
-				gateway.setRequestTimeout(timeout);
+		if (requestTimeout != null) {
```
garyrussell, line 959: We still need a null check here and use the default; `GenericMessagingTemplate` has `-1` as its default.

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -782,14 +782,9 @@ private Expression extractRequestTimeoutFromAnnotationOrMetadata(@Nullable Gatew
...
-			 * There is a small risk that someone has used Long.MIN_VALUE explicitly
-			 * to indicate an indefinite timeout on a gateway method and that will
-			 * no longer work as expected; they will need to use, say, -1 instead.
-			 */
-			if (requestTimeout == null || gatewayAnnotation.requestTimeout() != Long.MIN_VALUE) {
+			if (requestTimeout == null ||
+					gatewayAnnotation.requestTimeout() != IntegrationContextUtils.DEFAULT_TIMEOUT) {
+
 				requestTimeout = new ValueExpression<>(gatewayAnnotation.requestTimeout());
 			}
```
garyrussell, line 789: See comment in `timeouts()`.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
artembilan: Sure! See `MessagingGatewaySupport`:
```
private long replyTimeout = IntegrationContextUtils.DEFAULT_TIMEOUT;
...
template.setReceiveTimeout(this.replyTimeout);
```
So, if we don't call `setReplyTimeout()`, then that default value is used in the `GenericMessagingTemplate`.

garyrussell: Ah - ok.

garyrussell (approved):

artembilan:
I addition some notes in the Migration Guide: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-6.0-to-6.1-Migration-Guide

Issue #8582, opened 2023-03-24
Transactional support in PostgresSubscribableChannel 


igorlovich:
**Expected Behavior**

Opening an issue as discussed on stackoverflow [Spring Integration PostgresSubscribableChannel is not transactional](https://stackoverflow.com/questions/75825803/spring-integration-postgressubscribablechannel-is-not-transactional).  

The expectation is that if there is an exception in the downstream the message returns to the underlying message store.

**Current Behavior**

When replacing current QueueChannel + Transactional Polling  flow with a PostgresSubscribableChannel the behavior of the flow changes, since messages are now lost if there is an exception in the downstream

**Context**

I have made the following modifications as a POC to get the desired behavior.  This does change the semantics somewhat, as before each message  (for a subscriber) was dispatched on a separate thread and now all the messages for (a subscriber) are emitted sequentially on the same thread

**PostgresChannelMessageTableSubscriber**

1.  Add another executor service that will be used for notification dispatch, initialized in the start method with `notificationThreadPoolSize` (defaults to 1)

```java
if (subscriptionExecutorService == null) {
    CustomizableThreadFactory threadFactory =
            new CustomizableThreadFactory("postgres-channel-dispatch-");
    threadFactory.setDaemon(true);
    subscriptionExecutorService = Executors.newFixedThreadPool(notificationThreadPoolSize);
}
```
2. Modify the subscriber notifications to happen asynchronously 
```java
subscriptions.forEach( it-> subscriptionExecutorService.submit(it::notifyUpdate));
```
3. Add a method shutDownAndAwaitTermination as per [ExecutorService javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ExecutorService.html) with a 5 second timeout for graceful termination. Use this method to shutdown the `subscriptionExecutorService` in the `stop` method

**PostgresSubscribableChannel**

1.  Inject a PlatformTransactionManager in the constructor and create a TransactionTemplate <br> `this.transactionTemplate = new TransactionTemplate(platformTransactionManager);`
5.  Remove the Executor from the UnicastDispatcher (and remove the method that allows setting the Executor). 
6.  Add an AtomicBoolean field `errorOnNotify`
7.  Modify the `notifyUpdate` method  
```java 
@Override
public void notifyUpdate() {
    try {
        transactionTemplate.executeWithoutResult(it -> {
            Message<?> message;
[...]

PR #8587, opened 2023-03-28
GH-8582:Transactional support in PostgresSubscribableChannel

igorlovich:
As discussed in [Transactional support in PostgresSubscribableChannel](https://github.com/spring-projects/spring-integration/issues/8582)

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 the original author or authors.
+ * Copyright 2022-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,6 +59,7 @@ import org.springframework.util.Assert;
  *
  * @author Rafael Winterhalter
  * @author Artem Bilan
+ * @author Igor Lovich
  *
  * @since 6.0
  */
@@ -149,6 +150,8 @@ public final class PostgresChannelMessageTableSubscriber implements SmartLifecyc
 			this.executor = executorToUse;
 		}
 		this.latch = new CountDownLatch(1);
+
+		CountDownLatch startingLatch = new CountDownLatch(1);
 		this.future = executorToUse.submit(() -> {
 			try {
 				while (isActive()) {
@@ -171,6 +174,8 @@ public final class PostgresChannelMessageTableSubscriber implements SmartLifecyc
 						try {
 							this.connection = conn;
 							while (isActive()) {
+								startingLatch.countDown();
+
 								PGNotification[] notifications = conn.getNotifications(0);
 								// Unfortunately, there is no good way of interrupting a notification
 								// poll but by closing its connection.
@@ -208,6 +213,18 @@ public final class PostgresChannelMessageTableSubscriber implements SmartLifecyc
 				this.latch.countDown();
 			}
 		});
+
+		try {
+			if (!startingLatch.await(5, TimeUnit.SECONDS)) {
+				throw new IllegalStateException("Failed to start "
+						+ PostgresChannelMessageTableSubscriber.class.getName());
+			}
+		}
+		catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			throw new IllegalStateException("Failed to start "
+					+ PostgresChannelMessageTableSubscriber.class.getName(), e);
+		}
 	}
 
 	private boolean isActive() {
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 the original author or authors.
+ * Copyright 2022-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,8 +16,10 @@
 
 package org.springframework.integration.jdbc.channel;
 
+import java.util.Optional;
 import java.util.concurrent.Executor;
 
+import org.springframework.core.log.LogAccessor;
 import org.springframework.core.task.SimpleAsyncTaskExecutor;
 import org.springframework.integration.channel.AbstractSubscribableChannel;
 import org.springframework.integration.dispatcher.MessageDispatcher;
@@ -25,6 +27,9 @@ import org.springframework.integration.dispatcher.UnicastingDispatcher;
 import org.springframework.integration.jdbc.store.JdbcChannelMessageStore;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHandler;
+import org.springframework.retry.support.RetryTemplate;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.Assert;
 
 /**
@@ -39,11 +44,14 @@ import org.springframework.util.Assert;
  *
  * @author Rafael Winterhalter
  * @author Artem Bilan
+ * @author Igor Lovich
  *
  * @since 6.0
  */
 public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 		implements PostgresChannelMessageTableSubscriber.Subscription {
+
+	private static final LogAccessor LOGGER = new LogAccessor(PostgresSubscribableChannel.class);
 
 	private final JdbcChannelMessageStore jdbcChannelMessageStore;
 
@@ -51,7 +59,13 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 
 	private final PostgresChannelMessageTableSubscriber messageTableSubscriber;
 
-	private UnicastingDispatcher dispatcher = new UnicastingDispatcher(new SimpleAsyncTaskExecutor());
+	private final UnicastingDispatcher dispatcher = new UnicastingDispatcher();
+
+	private TransactionTemplate transactionTemplate;
+
+	private RetryTemplate retryTemplate = RetryTemplate.builder().maxAttempts(1).build();
+
+	private Executor executor = new SimpleAsyncTaskExecutor();
 
 	/**
 	 * Create a subscribable channel for a Postgres database.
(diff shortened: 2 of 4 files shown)


igorlovich:
The failing test has nothing to do with the MR

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -208,6 +217,29 @@ public synchronized void start() {
 				this.latch.countDown();
 			}
 		});
+
+		try {
+			if (!startingLatch.await(5, TimeUnit.SECONDS)) {
```
artembilan, line 222: To mitigate a race condition you can just swap subscription to this publisher in the test with its start call.
But I think I'm OK with this `startingLatch` since it give us a chance to ensure that we call `LISTEN` before any inserts into a table.

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
igorlovich: My opinion is that having a latch gives a more deterministic behavior

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -242,6 +279,35 @@ public synchronized void stop() {
 		}
 	}
 
+
+	/**
+	 * Gracefully shutdown an executor service. Taken from @see ExecutorService javadoc
+	 *
+	 * @param pool The pool to shut down
+	 */
+	private void shutdownAndAwaitTermination(@Nullable ExecutorService pool) {
```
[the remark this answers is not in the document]
artembilan: Probably no. The default `SimpleAsyncTaskExecutor` we use over there has no hook for shutdown.
It is really recommended to provide externally managed `Executor`, which in most cases is a `org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor`.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -208,6 +217,29 @@ public synchronized void start() {
...
+			Thread.currentThread().interrupt();
+			throw new IllegalStateException("Failed to start "
+					+ PostgresChannelMessageTableSubscriber.class.getName(), e);
+		}
+	}
+
+	private void notifyAll(Set<Subscription> subscriptions) {
+		subscriptions.forEach(it -> {
+			try {
+				this.executor.submit(it::notifyUpdate);
```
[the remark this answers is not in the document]
artembilan: I don't think this is already our responsibility.
We can warn, though, in that `Subscription` Javadocs.

From here I wonder if that would be really better to have this feature implemented against this library: https://impossibl.github.io/pgjdbc-ng/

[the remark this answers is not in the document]
igorlovich: I have moved the executor into the channel.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -110,10 +143,37 @@ protected boolean doSend(Message<?> message, long timeout) {
 
 	@Override
 	public void notifyUpdate() {
-		Message<?> message;
-		while ((message = this.jdbcChannelMessageStore.pollMessageFromGroup(this.groupId)) != null) {
-			this.dispatcher.dispatch(message);
-		}
+		this.executor.execute(() -> {
```
artembilan, line 146: Can we look into the way to schedule a task per message?
Kinda perform an execute for single poll, if it was not null, then schedule the next one.

If there is a FIFO requirements, that's where this task executor can be customized externally.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -75,7 +89,26 @@ public PostgresSubscribableChannel(JdbcChannelMessageStore jdbcChannelMessageSto
 	 */
 	public void setDispatcherExecutor(Executor executor) {
 		Assert.notNull(executor, "An executor must be provided.");
-		this.dispatcher = new UnicastingDispatcher(executor);
+		this.executor = executor;
+	}
+
+	/**
+	 * Sets the transaction manager to use for message processing. Each message will be processed in a
```
artembilan, line 96: Imperative language in method Javadocs, please: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -75,7 +89,26 @@ public PostgresSubscribableChannel(JdbcChannelMessageStore jdbcChannelMessageSto
...
+	 * separate transaction
+	 * @param transactionManager The transaction manager to use
```
artembilan, line 98: `@since 6.0.5`

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -39,19 +44,28 @@
...
  * @since 6.0
  */
 public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 		implements PostgresChannelMessageTableSubscriber.Subscription {
 
+	private static final LogAccessor LOGGER = new LogAccessor(PostgresSubscribableChannel.class);
+
 	private final JdbcChannelMessageStore jdbcChannelMessageStore;
 
+	private TransactionTemplate transactionTemplate;
```
artembilan, line 58: Please, move this non-`final` property to respective section of props.

src/reference/asciidoc/jdbc.adoc
```diff
@@ -632,6 +632,14 @@ Such connection pools do normally expect that issued connections are closed with
 For this need of an exclusive connection, it is also recommended that a JVM only runs a single `PostgresChannelMessageTableSubscriber` which can be used to register any number of subscriptions.
 ====
 
+==== Transaction support
```
artembilan, line 635: I don't think we need a dedicated section id for these paragraphs. It is fully OK to have this info withing that top-level `postgresql-push`.

src/reference/asciidoc/jdbc.adoc
```diff
@@ -632,6 +632,14 @@ Such connection pools do normally expect that issued connections are closed with
...
+Specifying a `PlatformTransactionManager` on `PostgresSubscribableChannel` will notify subscribers in a transaction.
```
artembilan, line 636: We also need to mention that this feature is available `starting with version 6.0.5`

src/reference/asciidoc/jdbc.adoc
```diff
@@ -632,6 +632,14 @@ Such connection pools do normally expect that issued connections are closed with
...
+An exception in a subscriber will cause the transaction to be rolled back and the message to be put back in the message store.
+Transactional support is not active by default.
+
+==== Retries
+A retry policy can be controlled by providing a `RetryTemplate` to `PostgresSubscribableChannel`. By default, no retries are performed.
```
artembilan, line 641: One sentence per line, please: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

artembilan (changes requested):
This is very close to what we all can agree in the end.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
igorlovich: By injecting a AsyncTaskExecutor and not an Executor we can get a CompletableFuture with the result. However getting the result (to determine whether to continue the loop) implies blocking, and is therefore equivalent to doing the loop as a single task, no matter what kind of an executor is passed in.

I can see 2 approaches (there may be more )  to achieve what you're proposing

1.  Have an AtomicReference outside the loop that is set inside the task to the result of the poll. The exit condition for the while loop is  the reference != null.  Disadvantage of this approach is that in the case of a multi-threaded executor, depending on the order of tasks finishing, we can get useless iterations through the loop if the null from a later submitted, but earlier finishing task is overridden.

2.  Ignore what is happening inside the task. The condition to exit from the while is `this.jdbcChannelMessageStore.messageGroupSize(groupId) == 0` .  There would need to be a sleep(X ms) to avoid hammering the message store. Disadvantage is that the call to notifyUpdate will take an X ms latency penalty.

igorlovich: Does this not equally imply sequential message processing, irrespective of the `Executor` , since each task only submits the next one after it finishes?

artembilan: Yeah... I see now. This is indeed the same what you have just explained with a `CompletableFuture` approach.
We could swap an new task submitting with the message processing. This way we can try to pull a new message in a new thread while we process the current one.
But is it really a goal for us? We would just eliminate any FIFO from this channel logic.
It might be as an option though, or can be implemented downstream as a `bridge().channel(new ExecutorChannel())` to not care about FIFO.
But that's fully different story.

I think it is time for me to step back and agree with your current solution with a single-threaded process for the whole loop.

I'm going to pull your PR locally for the final review if you don't have anything else to add.

Thank you for your patience!

igorlovich: No worries it is important to get it right (at least with our current understanding of "right"). I will address your feedback on the documentation

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
igorlovich: done

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
igorlovich: done

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
igorlovich: done

src/reference/asciidoc/jdbc.adoc
igorlovich: done

src/reference/asciidoc/jdbc.adoc
igorlovich: done

src/reference/asciidoc/jdbc.adoc
igorlovich: done

artembilan: FYI: I prefer to not have these kind of comments in the PR.
If you agree with me, just push the change.
Otherwise feel free to discuss.
In the end this `done` looks like a spam in my e-mail box 😄

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/e39449b6436339adbce46f75c8e65434297b2003 after some minor clean up.
And cherry-picked to `6.0.x`.

@igorlovich ,

thank you very much for contribution; looking forward for more!

Issue #8577, opened 2023-03-18
FolderClosedException when downloading large file attachments


jayChrono:
**In what version(s) of Spring Integration are you seeing this issue?**
v6.0.3

**Describe the bug**
When downloading a large file attachment downstream, the IMAP folder can be closed in subsequent `AbstractMailSender.receive()` when there are no more `messagesToReturn` throwing a `FolderClosedException`.

From the documentation,
> Starting with version 5.5.11, the folder is closed automatically after AbstractMailReceiver.receive() if no messages received or all of them are filtered out independently of the autoCloseFolder flag. In this case there is nothing to produce downstream for possible logic around IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE header.

And indeed lines 384-386 in `AbstractMailReceiver` describes this behaviour.
>		if (this.autoCloseFolder || ObjectUtils.isEmpty(messagesToReturn)) {
>			closeFolder();
>		}

**To Reproduce**
1. Create `IntegrationFlow` with `Mail.imapIdleAdapter()` and `autoCloseFolder(false)`
2. Attach a `MessageHandler` that runs the `MimeBodyPart.saveFile()` function
3. Put an email in the mailbox with a large (1MB) file attachment

**Expected behavior**
Expect to save attachments regardless of size (save for mail server limits).

**Sample**
Here's the trace of the error.

DEBUG 24076 --- [scheduling-1] o.s.integration.mail.ImapMailReceiver    : opening folder [imaps://xxxxxxxx%40hotmail.com@outlook.office365.com:993/Test]
DEBUG 24076 --- [scheduling-1] com.sun.mail.imap                        : connection available -- size: 1
 INFO 24076 --- [scheduling-1] com.sun.mail.imap.messagecache           : create cache of size 3
DEBUG 24076 --- [scheduling-1] com.sun.mail.imap.messagecache           : create message number 3
DEBUG 24076 --- [scheduling-1] com.sun.mail.imap                        : IMAPProtocol noop
DEBUG 24076 --- [scheduling-1] o.s.integration.mail.ImapMailReceiver    : attempting to receive mail from folder [Test]
DEBUG 24076 --- [scheduling-1] o.s.integration.mail.ImapMailReceiver    : found 1 new messages
DEBUG 24076 --- [scheduling-1] o.s.integration.mail.ImapMailReceiver    : Received 1 messages
[...]

PR #8589, opened 2023-03-28
GH-8577: Deprecate ImapIdleCA.sendingTaskExecutor

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8577

When we process mail messages in async manner, it is possible that we end up in a race condition situation where the next idle cycle closes the folder.

It is possible to reopen the folder, but feels better to block the current idle cycle until we are done with the message and therefore keep folder opened.

* Deprecate `ImapIdleChannelAdapter.sendingTaskExecutor` in favor of an `ExecutorChannel` as an output for this channel adapter or similar async hand-off downstream.
* Make a default one as a `SyncTaskExecutor` to make a sense of this deprecation

### DIFF
--- a/src/reference/asciidoc/mail.adoc
+++ b/src/reference/asciidoc/mail.adoc
@@ -143,6 +143,8 @@ In this case, the only header populated is the mentioned above `IntegrationMessa
 Starting with version 5.5.11, the folder is closed automatically after `AbstractMailReceiver.receive()` if no messages received or all of them are filtered out independently of the `autoCloseFolder` flag.
 In this case there is nothing to produce downstream for possible logic around `IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE` header.
 
+Starting with version 6.0.5, an `ImapIdleChannelAdapter.sendingTaskExecutor` option is deprecated in favor of an asynchronous message handling in downstream flow, e.g. via an `ExecutorChannel` as an output channel of this channel adapter.
+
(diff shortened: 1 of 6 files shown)


src/reference/asciidoc/mail.adoc
```diff
@@ -143,6 +143,8 @@ In this case, the only header populated is the mentioned above `IntegrationMessa
 Starting with version 5.5.11, the folder is closed automatically after `AbstractMailReceiver.receive()` if no messages received or all of them are filtered out independently of the `autoCloseFolder` flag.
 In this case there is nothing to produce downstream for possible logic around `IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE` header.
 
+Starting with version 6.0.5, an `ImapIdleChannelAdapter.sendingTaskExecutor` option is deprecated in favor of an asynchronous message handling in downstream flow, e.g. via an `ExecutorChannel` as an output channel of this channel adapter.
```
garyrussell, line 146:
```suggestion
Starting with version 6.0.5, the `ImapIdleChannelAdapter.sendingTaskExecutor` option is deprecated in favor of using asynchronous message handling in the downstream flow, e.g. via an `ExecutorChannel` as the output channel of this channel adapter.
```

garyrussell (changes requested):
Just doc polishing.

garyrussell (approved):

PR #8594, opened 2023-04-12
GH-8586: Deprecate IntegrationComponentSpec.get()

### LINKED ISSUE #8586: Revise the `IntegrationComponentSpec.get()` API to disallow to be called from end-user code
The `IntegrationComponentSpec` implements a `FactoryBean` and we usually don't call its `getObject()` in the target project.
We rather rely on the Spring Framework dependency injection lifecylce.
The `IntegrationComponentSpec.get()` manual call may lead the component to the state where it is not fully initialized.
The Java DSL API and the framework by itself is able to deal properly with `IntegrationComponentSpec` as a bean as it can deal with `FactoryBean`.

Related issue: https://github.com/spring-projects/spring-integration/issues/2995

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8586

The `IntegrationComponentSpec` is not a plain wrapper around single component. Sometimes it comes with several components where all of them must be registered as beans.
If `IntegrationComponentSpec.get()` is called from end-user code, we may lose other related components, for example filters in the `FileInboundChannelAdapterSpec`.

* Deprecate `IntegrationComponentSpec.get()` with no-op for end-user, rather encourage to leave it as is and let the framework take care about its lifecycle and related components registration
* Fix `IntegrationComponentSpec` logic to deal as a simple `FactoryBean` instead of extra overhead via `AbstractFactoryBean`
* Use `IntegrationComponentSpec.getObject()` in the framework code where `get()` was called
* Fix tests to expose `IntegrationComponentSpec` as beans instead of previously called `get()`
* Some other clean up and typos fixes in the affected classes
* Document the change

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationComponentSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationComponentSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,15 +16,21 @@
 
 package org.springframework.integration.dsl;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.FactoryBean;
 import org.springframework.beans.factory.InitializingBean;
-import org.springframework.beans.factory.config.AbstractFactoryBean;
 import org.springframework.context.Lifecycle;
 import org.springframework.context.SmartLifecycle;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 
 /**
- * The common Builder abstraction. The {@link #get()} method returns the final component.
+ * The common Builder abstraction.
+ * If used as a bean definition, must be treated as an {@link FactoryBean},
+ * therefore its {@link #getObject()} method must not be called in the target configuration.
  *
  * @param <S> the target {@link IntegrationComponentSpec} implementation type.
  * @param <T> the target type.
@@ -35,10 +41,11 @@ import org.springframework.expression.spel.standard.SpelExpressionParser;
  */
 @IntegrationDsl
 public abstract class IntegrationComponentSpec<S extends IntegrationComponentSpec<S, T>, T>
-		extends AbstractFactoryBean<T>
-		implements SmartLifecycle {
+		implements FactoryBean<T>, InitializingBean, DisposableBean, SmartLifecycle {
 
 	protected static final SpelExpressionParser PARSER = new SpelExpressionParser();
+
+	protected final Log logger = LogFactory.getLog(getClass()); // NOSONAR - final
 
 	protected volatile T target; // NOSONAR
 
@@ -61,8 +68,25 @@ public abstract class IntegrationComponentSpec<S extends IntegrationComponentSpe
 
 	/**
 	 * @return the configured component.
+	 * @deprecated since 6.1 with no-op for end-user:
+	 * the {@link #getObject()} is called by the framework at the appropriate phase.
 	 */
+	@Deprecated(since = "6.1", forRemoval = true)
 	public T get() {
+		return getObject();
+	}
+
+	@Override
+	public Class<?> getObjectType() {
+		return getObject().getClass();
+	}
+
+	/**
+	 * !!! This method must not be called from the target configuration !!!
+	 * @return the object backed by this factory bean.
+	 */
+	@Override
+	public T getObject() {
 		if (this.target == null) {
 			this.target = doGet();
 		}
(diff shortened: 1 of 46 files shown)


src/reference/asciidoc/dsl.adoc
```diff
@@ -63,6 +63,10 @@ You need not replace all of your existing XML configuration to use Java configur
 The `org.springframework.integration.dsl` package contains the `IntegrationFlowBuilder` API mentioned earlier and a number of `IntegrationComponentSpec` implementations, which are also builders and provide the fluent API to configure concrete endpoints.
 The `IntegrationFlowBuilder` infrastructure provides common https://www.enterpriseintegrationpatterns.com/[enterprise integration patterns] (EIP) for message-based applications, such as channels, endpoints, pollers, and channel interceptors.
 
+IMPORTANT:: The `IntegrationComponentSpec` is a `FactoryBean` implementation, therefore its `getObject()` method must not be called from the target configuration.
```
garyrussell, line 66:
```suggestion
IMPORTANT:: The `IntegrationComponentSpec` is a `FactoryBean` implementation, therefore its `getObject()` method must not be called from bean definitions.
```

src/reference/asciidoc/dsl.adoc
```diff
@@ -63,6 +63,10 @@ You need not replace all of your existing XML configuration to use Java configur
...
+The `IntegrationComponentSpec` implementation must be left as is for bean definition and the framework will take about its lifecycle.
```
garyrussell, line 67:
```suggestion
The `IntegrationComponentSpec` implementation must be left as is for bean definitions and the framework will manage its lifecycle.
```

src/reference/asciidoc/dsl.adoc
```diff
@@ -63,6 +63,10 @@ You need not replace all of your existing XML configuration to use Java configur
...
+A bean method parameter injection for the target `IntegrationComponentSpec` type (a `FactoryBean` value) must be used for `IntegrationFlow` bean definitions instead of bean method references.
```
garyrussell, line 68:
```suggestion
Bean method parameter injection for the target `IntegrationComponentSpec` type (a `FactoryBean` value) must be used for `IntegrationFlow` bean definitions instead of bean method references.
```

src/reference/asciidoc/whats-new.adoc
```diff
@@ -41,6 +41,11 @@ See <<./filter.adoc#filter, Filter>> for more information.
  - The default timeout for send and receive operations in gateways and replying channel adapters has been changed from infinity to `30` seconds.
 Only one left as a `1` second is a `receiveTimeout` for `PollingConsumer` to not block a scheduler thread too long and let other queued tasks to be performed with the `TaskScheduler`.
 
+ - An `IntegrationComponentSpec.get()` method has been deprecated with removal plans in the next version.
```
garyrussell, line 44:
```suggestion
 - The `IntegrationComponentSpec.get()` method has been deprecated with removal planned for the next version.
```

src/reference/asciidoc/whats-new.adoc
```diff
@@ -41,6 +41,11 @@ See <<./filter.adoc#filter, Filter>> for more information.
...
+Since `IntegrationComponentSpec` is a `FactoryBean`, its bean definition must stay as is without any target object resolutions.
+The Java DSL and the framework by itself will take care about the proper `IntegrationComponentSpec` lifecycle.
```
garyrussell, line 46:
```suggestion
The Java DSL and the framework by itself will manage the `IntegrationComponentSpec` lifecycle.
```

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationComponentSpec.java
```diff
@@ -61,78 +68,83 @@ public final String getId() {
...
 	@Override
-	protected T createInstance() {
-		T instance = get();
-		if (instance instanceof InitializingBean) {
-			try {
-				((InitializingBean) instance).afterPropertiesSet();
-			}
-			catch (Exception e) {
-				throw new IllegalStateException("Cannot initialize bean: " + instance, e);
+	public T getObject() {
```
garyrussell, line 89: Perhaps we should rename this to `internalGetObject()` with javadocs clearly specifying that it must only be called from framework components (or custom DSL specs for advanced users who might implement DSL extensions).

Then implement `getObject()` to throw an `UnsupportedOperationException`.

WDYT?

garyrussell (changes requested):
Just one comment, and some doc polishing.

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationComponentSpec.java
artembilan: Sorry, we cannot. The `IntegrationComponentSpec` can be declared as a bean, like any `FactoryBean`.
See our many examples in the tests like:
```
		@Bean
		public KafkaPublishSubscribeChannelSpec pubSub(KafkaTemplate<Integer, String> template,
				ConcurrentKafkaListenerContainerFactory<Integer, String> containerFactory) {

			return Kafka.publishSubscribeChannel(template, containerFactory, TEST_TOPIC7);
		}
```
In this case the target `IntegrationFlow` deals only with a result of this spec. Same way as it is done with regular `FactoryBean`.
That's why I pointed this in the doc.
At the same time the `IntegrationFlowBeanPostProcessor` is able to process a `ComponentsRegistration` from such a `IntegrationComponentSpec` in parallel.

garyrussell: Oh; right; sorry.

garyrussell (approved):

PR #8597, opened 2023-04-14
Fix NPEs in DSL Specs

artembilan:
The `BaseWsInboundGatewaySpec` and `TailAdapterSpec` don't override super methods and when we call them we fail with NPE since `target` was not populated.

* Fix `BaseWsInboundGatewaySpec` and its inheritors to populate the `target` from their ctors.
* Remove redundant methods from `BaseWsInboundGatewaySpec` hierarchy
* Propagate `AbstractWebServiceInboundGateway` properties directly to the target from the `BaseWsInboundGatewaySpec` inheritors
* Override `MessageProducerSpec` methods in the `TailAdapterSpec` to populate respective option into the `FileTailInboundChannelAdapterFactoryBean`
* Expose more missed options for producer endpoint in the `FileTailInboundChannelAdapterFactoryBean`


garyrussell (approved):

garyrussell:
...and cherry-picked to 6.0.x as a45c3c5396 after resolving conflicts, and to 5.5.x as c44c2e6545 after more conflict resolution.

PR #8603, opened 2023-04-26
Demonstrate `@Publisher` with an `@EventListener`

artembilan:
The `@EventListener` from Spring Framework is a good tool for POJO configuration to subscribe to events from an `ApplicationContext`. The `@Publisher` is an AOP tool to publish a `Message` from POJO method call.

* Add a test and docs to demonstrate how the `@Publisher` can be used together with an `@EventListener`

### DIFF
--- a/spring-integration-event/src/test/java/org/springframework/integration/event/dsl/IntegrationFlowEventsTests.java
+++ b/spring-integration-event/src/test/java/org/springframework/integration/event/dsl/IntegrationFlowEventsTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-integration-event/src/test/java/org/springframework/integration/event/dsl/IntegrationFlowEventsTests.java
```diff
@@ -176,6 +190,17 @@ public IntegrationFlow delayFlow() {
 					.channel(MessageChannels.queue("delayedResults"));
 		}
 
+		@Bean
+		QueueChannel eventFromPublisher() {
+			return new QueueChannel();
+		}
+
+		@EventListener
+		@Publisher("eventFromPublisher")
```
garyrussell, line 199: While this would work, do we really want to promote it? An `ApplicationEventListeningMessageProducer` is a much simpler approach.

garyrussell (commented):
The `@RabbitListener` one looks a bit awkward too.

What drove you to create this PR?

artembilan:
> What drove you to create this PR?

Yeah... When I asked end-user to catch an application event and publish it into a channel it turned out to not be so smooth to digest.
Their first question was: why do we need extra dependency for events?
Another one that `ApplicationEventListeningMessageProducer` is a bit long name. 
Plus `eventTypes` is a setter, so it leads to more code lines, then just couple annotations on a POJO method.
Plus method signature dictates type of the event to consume and body result is naturally a message to produce.

Right, with `@RabbitListener` it is a bit awkward, but at least we it covered in docs if someone comes to us with the question that they know this annotation, but would like to produce to the channel without code, but rather annotation style.

Another thought (but that's for other day contribution) is to have something like `IntegrationFlowBuilder fromQueue()`  in the `Amqp` factory to avoid extra codding for `IntegrationFlow.from(Amqp.)`.
Something similar for events - `fromEvent(Class<?>... eventTypes)`.

Plus it might be great to look into sensible default beans for something like `ConnectionFactory` like we do for container factory in the `@RabbitListener`.
There idea is to come up with some hook which would let end-users to end up with a solution with  Spring Integration faster than it is now like they claim to have it with just two annotations for `@RabbitListener`.

garyrussell (approved):

PR #8607, opened 2023-05-01
docs(kotlin-dsl): correct IntegrationFlow import

exaucae:
I was writing a guide to Spring Integration and Koltin DSL. I noticed a case mistake in the `IntegrationFlow` [import](https://docs.spring.io/spring-integration/reference/html/kotlin-dsl.html#kotlin-dsl), hence this PR.

integration-flow-err

### DIFF
--- a/src/reference/asciidoc/kotlin-dsl.adoc
+++ b/src/reference/asciidoc/kotlin-dsl.adoc
@@ -3,7 +3,7 @@
 
 The Kotlin DSL is a wrapper and extension to <<./dsl.adoc#java-dsl,Java DSL>> and aimed to make Spring Integration development on Kotlin as smooth and straightforward as possible with interoperability with the existing Java API and Kotlin language-specific structures.
 
-All you need to get started is just an import for `org.springframework.integration.dsl.integrationFlow` - an overloaded global function for Kotlin DSL.
+All you need to get started is just an import for `org.springframework.integration.dsl.IntegrationFlow` - an overloaded global function for Kotlin DSL.
 
 For `IntegrationFlow` definitions as lambdas we typically don't need anything else from Kotlin and just declare a bean like this:
 


src/reference/asciidoc/kotlin-dsl.adoc
```diff
@@ -3,7 +3,7 @@
 
 The Kotlin DSL is a wrapper and extension to <<./dsl.adoc#java-dsl,Java DSL>> and aimed to make Spring Integration development on Kotlin as smooth and straightforward as possible with interoperability with the existing Java API and Kotlin language-specific structures.
 
-All you need to get started is just an import for `org.springframework.integration.dsl.integrationFlow` - an overloaded global function for Kotlin DSL.
+All you need to get started is just an import for `org.springframework.integration.dsl.IntegrationFlow` - an overloaded global function for Kotlin DSL.
```
artembilan, line 6: No, that's not correct. There is exactly an overloaded global function called `integrationFlow()`. See `IntegrationFlowDsl.kt` and respective KotlinDocs: https://docs.spring.io/spring-integration/docs/current/kdoc-api/spring-integration-core/org.springframework.integration.dsl/integration-flow.html

artembilan (changes requested):

src/reference/asciidoc/kotlin-dsl.adoc
exaucae: I was on `Spring Integration 5.0.6.RELEASE`. Glad to see it working on 6.0.5

artembilan: That version is long ago at the End of Life. 
So, better to upgrade to the latest supported.
If you cannot move to Java 17 and latest Kotlin, then consider to use `5.5.17`: https://spring.io/projects/spring-integration#learn

PR #8610, opened 2023-05-04
GH-8609: Fix bug that cannot applied correctly TcpConnectorInterceptor

### LINKED ISSUE #8609: Cannot applied correctly TcpConnectorInterceptor on sending with server connection

5.5.17 (Probably 6.0.x line too)

**Describe the bug**

If define multiple `TcpConnectionInterceptor`, cannot applied correctly `TcpConnectorInterceptor` **on sending a message with server connection**. As actual result, it has been applied only the first interceptor(interceptor that wrapped the real `TcpConnection`). Probably c1aa9b7958d92cc1ce13a78c77e0fadce102bb6c 's changed are affect this behavior.

**To Reproduce**

For example, When following beans are defined, `characterAppendingInterceptor1` has been applied but `characterAppendingInterceptor2` has not been applied on sending a message. As a side note, When receive a message, all interceptors are applied.

```java
  @Bean
  @Order(1)
  public TcpConnectionInterceptorFactory characterAppendingInterceptor1() {
    return new CharacterAppendingInterceptor.Factory("1");
  }

  @Bean
  @Order(2) // This interceptor cannot applied on sending
  public TcpConnectionInterceptorFactory characterAppendingInterceptor2() {
    return new CharacterAppendingInterceptor.Factory("2");
  }

  @Bean
  public TcpConnectionInterceptorFactoryChain tcpConnectionInterceptorFactoryChain(
      List<TcpConnectionInterceptorFactory> tcpConnectionInterceptorFactories) {
    TcpConnectionInterceptorFactoryChain chain = new TcpConnectionInterceptorFactoryChain();
    chain.setInterceptor(tcpConnectionInterceptorFactories.toArray(new TcpConnectionInterceptorFactory[0]));
    return chain;
  }

  @Bean
  public AbstractServerConnectionFactory serverConnectionFactory(
      TcpConnectionInterceptorFactoryChain chain) {
    return Tcp.netServer(1234)
        .deserializer(TcpCodecs.crlf())
        .interceptorFactoryChain(chain)
        .get();
  }
```

**Expected behavior**

We expected all interceptors are applied on sending a message. As workaround, I add the following `TcpConnectorInterceptor`, it work fine as expected behavior.

```java
// The patch interceptor for binding a connection that wrapped all interceptor to TcpSender(in this case TcpSendingMessageHandler)
public class PatchInterceptor extends TcpConnectionInterceptorSupport {

  private PatchInterceptor(ApplicationEventPublisher publisher) {
    super(publisher);
  }

  @Override public void addNewConnection(TcpConnection connection) {
    super.addNewConnection(connection);
    // Call addNewConnection of TcpSender with wrapped connection(interceptor)
    Optional.ofNullable(getSender()).ifPresent(x -> x.addNewConnection(getTheConnection()));
  }

  static class Factory extends ApplicationObjectSupport implements TcpConnectionInterceptorFactory {

    @Override public TcpConnectionInterceptorSupport getInterceptor() {
      return new PatchInterceptor(obtainApplicationContext());
    }

  }

}
```

**Sample**

Just wait a few time!! I will add  the repro project in My GitHub repository at after.

* https://github.com/kazuki43zoo/spi-gh-8609

**How to reproduce:**
Please run the `SpiGh8609ApplicationTests#connectAndReceive`.

**FYI:**

* `SpiGh8609ApplicationWithPatchTests` : Tests for applying workaround patch interceptor
* `SpiGh8609ApplicationWithoutInterceptorTests` : Tests for no interceptors

kazuki43zoo:
Fixes spring-projects/spring-integration#8609

* ~~add getLastWrapper() in TcpConnectionSupport~~
* ~~change to passed the instance that geted via getLastWrapper() instead of self instance to TcpSender#addNewConnection~~
* Changed to passed the self instance to addNewConnection instead of argument's connection

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import org.springframework.messaging.support.ErrorMessage;
  * to the underlying {@link TcpConnection}.
  *
  * @author Gary Russell
+ * @author Kazuki Shimizu
  *
  * @since 2.0
  */
@@ -232,7 +233,7 @@ public abstract class TcpConnectionInterceptorSupport extends TcpConnectionSuppo
 	@Override
 	public void addNewConnection(TcpConnection connection) {
 		if (this.interceptedSenders != null) {
-			this.interceptedSenders.forEach(sender -> sender.addNewConnection(connection));
+			this.interceptedSenders.forEach(sender -> sender.addNewConnection(this));
 		}
 	}
 
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpSenderTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpSenderTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 the original author or authors.
+ * Copyright 2022-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,7 +18,9 @@ package org.springframework.integration.ip.tcp.connection;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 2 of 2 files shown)


artembilan (changes requested):
I wonder what is wrong with my suggestion to fix `TcpConnectionInterceptorSupport`: https://github.com/spring-projects/spring-integration/issues/8609#issuecomment-1533480685.

The test looks OK and I believe the outcome will be exactly the same with my fix.

artembilan:
Yeah... It would be better to not have any wrapping at all, but rather just a chain of interceptors injected into a `Connection`.
But that's fully different story.

Let's listener what @garyrussell has in his mind on the matter!
He is an author of this TCP/IP framework, so better to make the final decision counting on his opinion.

Thank you for the contribution anyway!

garyrussell:
>I thought it would be better to pass the last wrapped connection to all interceptors, what do you think?

My intuition is that an interceptor in the chain should only reference the next in the chain, not the outermost one.

So I prefer @artembilan 's approach.

kazuki43zoo:
@garyrussell @artembilan 
Thanks for your comment!! I will change the code by @artembilan 's approach!!

kazuki43zoo:
@garyrussell @artembilan 
I've changed the code by @artembilan 's approach, please review again when you have time! Thanks!

artembilan (approved):

artembilan:
... and cherry-picked to `6.0.x` & `5.5.x`.

@kazuki43zoo ,

PR #8612, opened 2023-05-04
MH: Async by default for reactive output channel

artembilan:
When we configure an output channel for the handler as a `FluxMessageChannel`, in most cases we assume an async processing for the reply. Especially this is critical when reply is a reactive type, so in the async mode it is "flattened" by the mentioned `FluxMessageChannel`. Therefore, it is a bit awkward to require to set async explicitly, when we have already configured output channel as a `FluxMessageChannel`

* Remove redundant config for async from `R2dbcDslTests`
* Mention the change in the docs


artembilan:
Will add a Migration Guide note when this is merged.

garyrussell (approved):

artembilan:
The Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-6.0-to-6.1-Migration-Guide#endpoint-is-async-by-default-for-output-channel-as-fluxmessagechannel

PR #8617, opened 2023-05-09
Introduce `PartitionedChannel`

artembilan:
* Implement a `PartitionedChannel` as an extension of the `AbstractExecutorChannel`
* Supply this channel with a `PartitionedDispatcher` which is an extension of the `AbstractDispatcher`
* The target partition is essentially a `UnicastingDispatcher` with a single thead executor

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.dispatcher;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.function.Function;
+
+import org.springframework.integration.util.ErrorHandlingTaskExecutor;
+import org.springframework.lang.Nullable;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
+import org.springframework.util.Assert;
+import org.springframework.util.ErrorHandler;
+
+/**
+ * An {@link AbstractDispatcher} implementation for distributing messages to
+ * dedicated threads according to the key determined by the provided function against
+ * the message to dispatch.
+ * <p>
+ * Every partition, created by this class, is a {@link UnicastingDispatcher}
+ * delegate based on a single thread {@link Executor}.
+ * <p>
+ * The rest of the logic is similar to {@link UnicastingDispatcher} behavior.
+ *
+ * @author Artem Bilan
+ *
+ * @since 6.1
+ */
+public class PartitionedDispatcher extends AbstractDispatcher {
+
+	private final Map<Integer, UnicastingDispatcher> partitions = new HashMap<>();
+
+	private final List<ExecutorService> executors = new ArrayList<>();
+
+	private final int partitionCount;
+
+	private final Function<Message<?>, Object> partitionKeyFunction;
+
+	private ThreadFactory threadFactory = new CustomizableThreadFactory("partition-thread-");
+
+	private boolean failover = true;
+
+	@Nullable
+	private LoadBalancingStrategy loadBalancingStrategy;
+
+	private ErrorHandler errorHandler;
+
+	private MessageHandlingTaskDecorator messageHandlingTaskDecorator = task -> task;
+
+	public PartitionedDispatcher(int partitionCount, Function<Message<?>, Object> partitionKeyFunction) {
+		Assert.isTrue(partitionCount > 0, "'partitionCount' must be greater than 0");
+		Assert.notNull(partitionKeyFunction, "'partitionKeyFunction' must not be null");
+		this.partitionKeyFunction = partitionKeyFunction;
+		this.partitionCount = partitionCount;
+	}
+
+	public void setThreadFactory(ThreadFactory threadFactory) {
+		Assert.notNull(threadFactory, "'threadFactory' must not be null");
+		this.threadFactory = threadFactory;
+	}
+
+	public void setFailover(boolean failover) {
+		this.failover = failover;
+	}
+
+	public void setLoadBalancingStrategy(@Nullable LoadBalancingStrategy loadBalancingStrategy) {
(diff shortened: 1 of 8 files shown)


artembilan:
If this is OK, I'll add some docs and DSL configuration.
Thanks

spring-integration-core/src/test/java/org/springframework/integration/channel/PartitionedChannelTests.java
```diff
@@ -0,0 +1,108 @@
...
+		assertThat(handleLatch.await(10, TimeUnit.SECONDS)).isTrue();
+
+		assertThat(partitionedMessages).hasSize(2);
+		partitionedMessages.values()
+				.forEach(messagesInPartition -> {
+					assertThat(messagesInPartition).hasSize(2);
+					assertThat(messagesInPartition.get(0).getHeaders().get("partitionKey"))
+							.isEqualTo(messagesInPartition.get(1).getHeaders().get("partitionKey"));
+				});
+
```
garyrussell, line 81: If we match the thread name to the partition, we could assert that the records went to the right partition.

garyrussell (changes requested):
Some suggestions; otherwise LGTM.

spring-integration-core/src/test/java/org/springframework/integration/channel/PartitionedChannelTests.java
artembilan: Well, it is not always true, since we don't use a partition key as is, but rather its `hashCode()`.
So, while all the same keys goes to the same partition, it does not mean that `1` for key will always determine the partition `0`.

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
```diff
@@ -0,0 +1,191 @@
...
+
+	@Override
+	public boolean dispatch(Message<?> message) {
+		prePopulatedPartitionsIfAny();
+		int partition = Math.abs(this.partitionKeyFunction.apply(message).hashCode()) % this.partitionCount;
+		UnicastingDispatcher partitionDispatcher = this.partitions.get(partition);
+		return partitionDispatcher.dispatch(message);
+	}
+
+	private void prePopulatedPartitionsIfAny() {
```
garyrussell, line 154:
```suggestion
	private synchronized void prePopulatedPartitionsIfAny() {
```

src/reference/asciidoc/channel.adoc
```diff
@@ -203,6 +203,30 @@ CAUTION: The sender can sometimes block.
...
+[[partitioned-channel]]
+===== `PartitionedChannel`
+
+Starting with version 6.1, a `PartitionedChannel` implementation is provided.
+This is an extension of `AbstractExecutorChannel` and represents a point-to-point dispatching logic where an actual consumption happens on a specific thread determined by the partition key evaluated from a message sent to this channel.
+This channel is similar to the `ExecutorChannel` mentioned above, but with a difference that messages with the same partition key are always handled in the same thread preserving ordering.
+It does not require an external `TaskExecutor`, but can be configured with a custom `ThreadFactory` (e.g. `Thread.ofVirtual().name("partition-", 0).factory()`).
+This factory then used to populate single-thread executors into `MessageDispatcher` delegates per partition.
+By default, the `IntegrationMessageHeaderAccessor.CORRELATION_ID` message header is used for partition key.
+This channel can be configured as a simple bean:
```
garyrussell, line 215:
```suggestion
Starting with version 6.1, a `PartitionedChannel` implementation is provided.
This is an extension of `AbstractExecutorChannel` and represents point-to-point dispatching logic where the actual consumption is processed on a specific thread, determined by the partition key evaluated from a message sent to this channel.
This channel is similar to the `ExecutorChannel` mentioned above, but with the difference that messages with the same partition key are always handled in the same thread, preserving ordering.
It does not require an external `TaskExecutor`, but can be configured with a custom `ThreadFactory` (e.g. `Thread.ofVirtual().name("partition-", 0).factory()`).
This factory is used to populate single-thread executors into a `MessageDispatcher` delegate, per partition.
By default, the `IntegrationMessageHeaderAccessor.CORRELATION_ID` message header is used as the partition key.
This channel can be configured as a simple bean:
```

src/reference/asciidoc/channel.adoc
```diff
@@ -203,6 +203,30 @@ CAUTION: The sender can sometimes block.
...
+[source,java]
+----
+@Bean
+PartitionedChannel somePartitionedChannel() {
+    return new PartitionedChannel(3, (message) -> message.getHeaders().get("partitionKey"));
+}
+----
+====
+
+The channel will have `3` partitions - dedicated threads; will use a `partitionKey` header to determine in which partition the message must be handled.
```
garyrussell, line 227:
```suggestion
The channel will have `3` partitions - dedicated threads; will use the `partitionKey` header to determine in which partition the message must be handled.
```

garyrussell, line 227:
```suggestion
The channel will have `3` partitions - dedicated threads; will use the `partitionKey` header to determine in which partition the message will be handled.
```

garyrussell (changes requested):
Just doc polishing.

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
```diff
@@ -145,13 +145,15 @@ public void shutdown() {
 
 	@Override
 	public boolean dispatch(Message<?> message) {
-		prePopulatedPartitionsIfAny();
+		if (this.partitions.isEmpty()) {
```
garyrussell, line 148: This won't work, thread 2 might proceed before thread 1 has populated all partition execs, and NPE on the dispatcher returned by the `get()`.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
artembilan: How about now?
See the latest commit.

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * Set a {@link ThreadFactory} for executors per partitions.
+	 * Defaults to the {@link CustomizableThreadFactory} based on a {@code partition-thread-} prefix.
+	 * @param threadFactory the {@link ThreadFactory} to use.
+	 */
+	public void setThreadFactory(ThreadFactory threadFactory) {
+		Assert.notNull(threadFactory, "'threadFactory' must not be null");
+		this.threadFactory = threadFactory;
+		populatedPartitions();
```
garyrussell, line 117: Need to shut down and clear existing executors.

Perhaps call `shutdown()` first?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
artembilan: Yeah... Nothing to shutdown I guess: those executors have nit been used yet.
But we indeed have to `this.executors.clear()` .

garyrussell: We don't know that; they could call this after we started processing messages.

garyrussell: And if not, it's a no-op; so why worry?

artembilan: Doh... We cannon populate them from the ctor: some other options are required for delegate dispatchers.
So, coming back to the `synchronized` solution 😄

garyrussell (approved):

PR #8619, opened 2023-05-12
GH-8014: Improve doc for Service Activator

### LINKED ISSUE #8014: Service Activator referred to as outbound gateway in the documentation [INT-4073]
**[Pradeep Kumar L](https://jira.spring.io/secure/ViewProfile.jspa?name=pradeep.kumar44@gmail.com)** opened **[INT-4073](https://jira.spring.io/browse/INT-4073?redirect=false)** and commented

I was going through the definition of service activator. **_The definition says that it is used to "call a method and wrap the result in the response message"_**. The definition also tells that it is a **_outbound gateway for invoking the bean method_**. I am not clear on the second statement. As I understand outbound gateway is to send the request from the application to external application and get the response back into the application. So, if a bean is invoked, it is invoked within the application and hence it should be inbound gateway right. Please let me know where I am wrong.

Reference to the statement above is done in the "enterprise-spring-4.0.a.Release-student-handout.pdf" which I received when I enrolled for the "Enterprise Spring Integration" certification course.

---

**Reference URL:** http://stackoverflow.com/questions/38428729/service-activator-acting-as-outbound-gateway-for-invoking-bean-method

**Attachments:**
- [screenshot-for-reference.png](https://jira.spring.io/secure/attachment/23446/screenshot-for-reference.png) (_256.75 kB_)

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8014

### DIFF
--- a/src/reference/asciidoc/service-activator.adoc
+++ b/src/reference/asciidoc/service-activator.adoc
@@ -7,9 +7,9 @@ Alternatively, an output-producing service may be located at the end of a proces
 This is the default behavior if no output channel is defined.
 As with most of the configuration options described here, the same behavior actually applies for most of the other components.
 
-The service activator is essentially a generic endpoint for calling method on some object with an input message and its headers.
-Its internal logic is based on a `MessageHandler` which can be any possible implementation for specific use-case, for example `DefaultMessageSplitter`, `AggregatingMessageHandler`, `SftpMessageHandler`, `JpaOutboundGateway` etc.
-Therefore any outbound gateway and outbound channel adapter mentioned in this reference manual should be treated as a specific extension of this service activator endpoint: they all in the end call some object's method.
+The service activator is essentially a generic endpoint for calling a method on some object with an input message (payload and headers).
+Its internal logic is based on a `MessageHandler` which can be any possible implementation for a specific use-case, for example `DefaultMessageSplitter`, `AggregatingMessageHandler`, `SftpMessageHandler`, `JpaOutboundGateway` etc.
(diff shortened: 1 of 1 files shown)


src/reference/asciidoc/service-activator.adoc
```diff
@@ -7,6 +7,10 @@ Alternatively, an output-producing service may be located at the end of a proces
 This is the default behavior if no output channel is defined.
 As with most of the configuration options described here, the same behavior actually applies for most of the other components.
 
+The service activator is essentially a generic endpoint for calling method on some object with an input message and its headers.
+Its internal logic is based on a `MessageHandler` which can be any possible implementation for specific use-case, for example `DefaultMessageSplitter`, `AggregatingMessageHandler`, `SftpMessageHandler`, `JpaOutboundGateway` etc.
+Therefore any outbound gateway and outbound channel adapter mentioned in this reference manual should be treated as a specific extension of this service activator endpoint: they all in the end call some object's method.
+
```
garyrussell, line 13:
```suggestion
The service activator is essentially a generic endpoint for calling a method on some object with an input message (payload and headers).
Its internal logic is based on a `MessageHandler` which can be any possible implementation for a specific use-case, for example `DefaultMessageSplitter`, `AggregatingMessageHandler`, `SftpMessageHandler`, `JpaOutboundGateway` etc.
Therefore, any outbound gateway and outbound channel adapter mentioned in this reference manual should be treated as a specific extension of this service activator endpoint; they all, in the end, call some object's method.

```

garyrussell (changes requested):
Doc polishing.

garyrussell (approved):

PR #8622, opened 2023-05-15
Fix memory leak in the `FluxMessageChannel`

artembilan:
The `FluxMessageChannel` can subscribe to any volatile `Publisher`. For example, we can call Reactor Kafka `Sender.send()` for input data and pass its result to the `FluxMessageChannel` for on demand subscription.
These publishers are subscribed in the `FluxMessageChannel` and their `Disposable` is stored in the internal `Disposable.Composite` which currently only cleared on `destroy()`

* Extract `Disposable` from those internal `subscribe()` calls into an `AtomicReference`.
* Use this `AtomicReference` in the `doOnTerminate()` to remove from the `Disposable.Composite` and `dispose()` when such a volatile `Publisher` is completed

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/FluxMessageChannel.java
@@ -18,6 +18,7 @@ package org.springframework.integration.channel;
 
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.LockSupport;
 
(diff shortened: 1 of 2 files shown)


artembilan:
Forgot to mention : it might not cherry-pick clean . So, let me know and I’ll figure out the change for those versions !

garyrussell (approved):

garyrussell:
...and cherry-picked to 6.0.x as 485aa2a493

Please back-port to 5.5.x - conflicts.

artembilan:
... and back-ported to `5.5.x` as https://github.com/spring-projects/spring-integration/commit/ccd3574dadb902d0ffc63fafb999e0944e502ce0 after fixing conflicts.

PR #8627, opened 2023-05-19
GH-8625: Add Duration support for `<poller>`

### LINKED ISSUE #8625: Poller fixed-delay and fixed-rate support for Duration syntax
**Expected Behavior**

It would be nice and useful (from a configuration point of view) if `<int:poller>` attributes `fixed-delay` and `fixed-rate` supported the `java.time.Duration` syntax, with support for resolvable configuration values (via a `StringValueResolver`) just like `@Scheduled`/`ScheduledAnnotationBeanPostProcessor` do.

**Current Behavior**

`fixed-delay` and `fixed-rate` attributes in `<int:poller>` must be integer values, interpreted as milliseconds or amount of `time-unit` units. Resolvable configuration values are supported, but configuration must then specify these values as milliseconds, or provide two distinct configuration values for delay/rate and time unit.

**Context**

Using Spring Integration with Spring Boot 2.7.10. Duration syntax is nicely supported in a lot of places, but not in `<int:poller>` tag.
The alternative to have maximum flexibility is to leave `fixed-delay`/`fixed-rate` in milliseconds and externalize them in configuration. But milliseconds are not so readable when resolutions of minutes or hours is needed.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8625

The duration can be represented in a ISO 8601 format, e.g. `PT10S`, `P1D` etc. The `<poller>` and `@Poller` don't support such a format.

* Introduce a `PeriodicTriggerFactoryBean` to accept string values for trigger options and parse them manually before creating the target `PeriodicTrigger`
* Use this `PeriodicTriggerFactoryBean` in the `PollerParser` and `AbstractMethodAnnotationPostProcessor` where we parse options for the `PeriodicTrigger`
* Modify tests to ensure that feature works
* Document the duration option
* Add more cross-links into polling docs
* Fix typos in the affected doc files
* Add `-parameters` for compiler options since SF 6.1 does not support `-debug` anymore for method parameter names discovery

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/PeriodicTriggerFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/PeriodicTriggerFactoryBean.java
@@ -28,7 +28,7 @@ import org.springframework.util.StringUtils;
 /**
  * The {@link FactoryBean} to produce a {@link PeriodicTrigger}
  * based on parsing string values for its options.
- * This class is mostly driver by the XML configuration requirements for
+ * This class is mostly driven by the XML configuration requirements for
  * {@link Duration} value representations for the respective attributes.
  *
  * @author Artem Bilan
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/PeriodicTriggerFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/PeriodicTriggerFactoryBean.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.config;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+import org.springframework.beans.factory.FactoryBean;
+import org.springframework.lang.Nullable;
+import org.springframework.scheduling.support.PeriodicTrigger;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * The {@link FactoryBean} to produce a {@link PeriodicTrigger}
+ * based on parsing string values for its options.
+ * This class is mostly driver by the XML configuration requirements for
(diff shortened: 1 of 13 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/PeriodicTriggerFactoryBean.java
```diff
@@ -0,0 +1,115 @@
...
+import org.springframework.beans.factory.FactoryBean;
+import org.springframework.lang.Nullable;
+import org.springframework.scheduling.support.PeriodicTrigger;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * The {@link FactoryBean} to produce a {@link PeriodicTrigger}
+ * based on parsing string values for its options.
+ * This class is mostly driver by the XML configuration requirements for
```
garyrussell, line 31:
```suggestion
 * This class is mostly driven by the XML configuration requirements for
```

src/reference/asciidoc/channel-adapter.adoc
```diff
@@ -125,6 +125,9 @@ However, if you are sure that your method can return null and you need to poll f
 
 Starting with version 5.5, a `0` value for `max-messages-per-poll` has a special meaning - skip the `MessageSource.receive()` call altogether, which may be considered as pausing for this inbound channel adapter until the `maxMessagesPerPoll` is changed to a non-zero value at a later time, e.g. via a Control Bus.
 
+Starting with version 6.2, the `fixed-delay` and `fixed-rate` can be configured in https://en.wikipedia.org/wiki/ISO_8601#Durations[ISO 8601 Duration] format, e.g. `PT10S`, `P1D` etc.
+In addition, an `initial-interval` of the underlying `PeriodicTrigget` is also expose with similar value formats as `fixed-delay` and `fixed-rate`.
```
garyrussell, line 129:
```suggestion
In addition, an `initial-interval` of the underlying `PeriodicTrigger` is also exposed with similar value formats as `fixed-delay` and `fixed-rate`.
```

spring-integration-core/src/main/java/org/springframework/integration/config/PeriodicTriggerFactoryBean.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	@Override
+	public Class<?> getObjectType() {
+		return PeriodicTrigger.class;
+	}
+
+	private static Duration toDuration(String value, TimeUnit timeUnit) {
+		if (isDurationString(value)) {
+			return Duration.parse(value);
```
garyrussell, line 98: Assert that timeUnit is not provided in this case?

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/PeriodicTriggerFactoryBean.java
artembilan: Having `time-unit` does not cause any ambiguity, so ignoring it as it is also done in the `ScheduledAnnotationBeanPostProcessor` should be totally OK.

PR #8633, opened 2023-05-25
Debezium DSL support

### LINKED ISSUE #8632: Add Java DSL for Debezium Channel Adapter
Something similar to `Cassandra` or `Camel` factories and respective `DebeziumMessageProducerSpec` impl.

tzolov:
Resolves #8632


spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public final class Debezium {
+
+	/**
+	 * Create an instance of {@link DebeziumMessageProducerSpec} for the provided {@link DebeziumEngine.Builder}.
+	 * @param debeziumEngineBuilder the {@link DebeziumEngine.Builder} to use.
+	 * @return the spec.
+	 */
+	public static DebeziumMessageProducerSpec inboundChannelAdapter(
+			Builder<ChangeEvent<byte[], byte[]>> debeziumEngineBuilder) {
```
artembilan, line 37: I wonder if we should add an overloaded version for something simple as `Properties, SerializationFormat` and create a `Builder` on the fly. That's exactly what I see in your `DebeziumTestConfiguration`.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/DebeziumMessageProducerSpec.java
```diff
@@ -0,0 +1,134 @@
...
+	 * You should prefix the names passed to the 'headerNames' with the prefix configured by the Debezium
+	 * 'add.headers.prefix' property. Later defaults to '__'. For example for 'add.headers=op,name' and
+	 * 'add.headers.prefix=__' you should use header hames like: '__op', '__name'.
+	 * @param headerNames The values in this list can be a simple patterns to be matched against the header names.
+	 * @return the spec.
+	 */
+	public DebeziumMessageProducerSpec headerNames(String... headerNames) {
+		Assert.notEmpty(headerNames, "'inboundHeaderNames' must not be empty");
+		this.inboundHeaderNames = Arrays.copyOf(headerNames, headerNames.length);
+		return addDebeziumHeaderMapper();
```
artembilan, line 115: I don't see a reason in this method: you simply can create a `DefaultDebeziumHeaderMapper`, initialize it here and then call `headerMapper()`. 
Also must be mentioned that this `headerNames()` and `headerMapper()` are mutually exclusive.

And from here we might not need that `inboundHeaderNames` property at all.

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ *
+ * @author Christian Tzolov
+ *
+ * @since 6.2
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class DebeziumDslTest {
```
artembilan, line 30: We prefer `*Tests` suffix 😄

artembilan (changes requested):
Not a big deal so far.
Please, also consider to add docs for this new API.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
tzolov: In addition to the properties and the serialization formats, the `Builder` and in turn the [DebeziumEngineBuilderAutoConfiguration.java](https://github.com/spring-cloud/stream-applications/blob/main/functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumEngineBuilderAutoConfiguration.java) provide configuration other things like: `offsetCommitPolicy`, `clock`, `completionCallback`, `connectorCallback`.

artembilan: I know. And for that reason we have this factory based on a `Builder`.
I'm just talking about a sensible defaults where end-user is just interested in format and some props for his/her DB.
Well, we even can look into a `inboundChannelAdapter(Properties debeziumProperties)` where JSON would be a default format.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
```diff
@@ -0,0 +1,144 @@
...
+	 * completion state of each component running within the engine.
+	 * @param debeziumClock Set a specific clock when needing to determine the current time. Defaults to
+	 * {@link Clock#systemDefaultZone() system clock}
+	 * @return the spec.
+	 */
+	public static DebeziumMessageProducerSpec inboundChannelAdapter(Properties debeziumConfig,
+			Class<? extends SerializationFormat<byte[]>> messageFormat,
+			Class<? extends SerializationFormat<byte[]>> headerFormat,
+			@Nullable OffsetCommitPolicy offsetCommitPolicy, @Nullable CompletionCallback completionCallback,
+			@Nullable ConnectorCallback connectorCallback, @Nullable Clock debeziumClock) {
```
artembilan, line 102: This is too much for a single factory method.
Better to point to the method based on a builder if there is need for more sophisticated configuration.
My point was only about simple use-cases where we have props and are OK with the rest as defaults.

So, why do you introduce this complex method?

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+public final class Debezium {
+
+	/**
+	 * Create an instance of {@link DebeziumMessageProducerSpec} for the provided {@link DebeziumEngine.Builder}.
+	 * @param debeziumEngineBuilder the {@link DebeziumEngine.Builder} to use.
+	 * @return the spec.
+	 */
+	public static DebeziumMessageProducerSpec inboundChannelAdapter(
+			Builder<ChangeEvent<byte[], byte[]>> debeziumEngineBuilder) {
```
artembilan, line 50: I would prefer a `DebeziumEngine.Builder` style: better for review and readability.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/DebeziumMessageProducerSpec.java
```diff
@@ -0,0 +1,125 @@
...
+import io.debezium.engine.format.SerializationFormat;
+
+import org.springframework.integration.debezium.inbound.DebeziumMessageProducer;
+import org.springframework.integration.debezium.support.DefaultDebeziumHeaderMapper;
+import org.springframework.integration.dsl.MessageProducerSpec;
+import org.springframework.messaging.support.HeaderMapper;
+import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
+
+/**
+ *
```
artembilan, line 36: Need some general class Javadoc.

src/reference/asciidoc/debezium.adoc
```diff
@@ -53,12 +53,26 @@ The following Spring Boot application shows an example of how to configure the i
...
+import org.springframework.boot.WebApplicationType;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.context.annotation.Bean;
+import org.springframework.integration.annotation.ServiceActivator;
+import org.springframework.integration.channel.DirectChannel;
+import org.springframework.integration.core.MessageProducer;
+import org.springframework.integration.debezium.inbound.DebeziumMessageProducer;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
```
artembilan, line 68: We never use imports for the code snippets in the doc.

src/reference/asciidoc/debezium.adoc
```diff
@@ -79,8 +93,8 @@ public class DebeziumJavaApplication {
     }
 
     @ServiceActivator(inputChannel = "debeziumInputChannel")
-    public void handler(String changeEventData) {
-        System.out.println(changeEventData);
+    public void handler(Message<?> changeEventData) {
+        System.out.println(new String((byte[]) changeEventData.getPayload()));
```
artembilan, line 97: And why did you change this back to such a style?
Why just plain `handler(String changeEventData)` is not enough anymore?
Or at least `byte[] changeEventData` if you think that there is no automatic conversion to string...

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslTests.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ *
+ * @author Christian Tzolov
+ *
+ * @since 6.2
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class DebeziumDslTests {
+
```
artembilan, line 31: Are you still going to add a content into this test class?

artembilan (changes requested):

src/reference/asciidoc/debezium.adoc
```diff
@@ -160,7 +164,7 @@ public void handler(List<ChangeEvent<Object, Object>> payload) {
 ----
 ====
 
-=== [[configuring-with-the-java-dsl]]Configuring with the Java DSL
+=== [[debezium-java-dsl]]Debezium Java DSL Support
```
artembilan, line 167: Something wrong with this code flow.
Must be like this:
```
[[debezium-java-dsl]]
=== Debezium Java DSL Support
```

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
```diff
@@ -0,0 +1,97 @@
...
+
+	/**
+	 * Create an instance of {@link DebeziumMessageProducerSpec} for the provided native debezium {@link Properties} and
+	 * JSON serialization formats.
+	 * @param debeziumConfig {@link Properties} with required debezium engine and connector properties.
+	 * @return the spec.
+	 */
+	public static DebeziumMessageProducerSpec inboundChannelAdapter(Properties debeziumConfig) {
+
+		return new DebeziumMessageProducerSpec(builder(
```
artembilan, line 56: Why just not delegate to `inboundChannelAdapter(Properties, Class, Class)`?

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
```diff
@@ -0,0 +1,97 @@
...
+	 * @param debeziumConfig {@link Properties} with required debezium engine and connector properties.
+	 * @param messageFormat {@link SerializationFormat} format for the {@link ChangeEvent} key and payload.
+	 * @param headerFormat {@link SerializationFormat} format for the {@link ChangeEvent} headers.
+	 * @return the spec.
+	 */
+	public static DebeziumMessageProducerSpec inboundChannelAdapter(Properties debeziumConfig,
+			Class<? extends SerializationFormat<byte[]>> messageFormat,
+			Class<? extends SerializationFormat<byte[]>> headerFormat) {
+
+		return new DebeziumMessageProducerSpec(builder(debeziumConfig, messageFormat, headerFormat));
```
artembilan, line 74: And this one could simply delegate to the `inboundChannelAdapter(DebeziumEngine.Builder)`

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	private static Builder<ChangeEvent<byte[], byte[]>> builder(Properties debeziumConfig,
```
artembilan, line 77: I believe that this return type should be as a `DebeziumEngine.Builder`, too.

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/DebeziumTestUtils.java
```diff
@@ -0,0 +1,61 @@
...
+
+		config.put("topic.prefix", "my-topic");
+		config.put("name", "my-connector");
+		config.put("database.server.id", "85744");
+		config.put("database.server.name", "my-app-connector");
+
+		config.put("key.converter.schemas.enable", "false");
+		config.put("value.converter.schemas.enable", "false");
+
+		config.put("connector.class", "io.debezium.connector.mysql.MySqlConnector");
```
artembilan, line 49: I think this util class content (just this method) can simply go into that `DebeziumMySqlTestContainer` since both of them are very connected at the moment.

While `DebeziumMySqlTestContainer` might live without this useful config, this `DebeziumTestUtils` doesn't look like it can be used without that Docker container.

src/reference/asciidoc/debezium.adoc
```diff
@@ -25,10 +23,47 @@ compile "org.springframework.integration:spring-integration-debezium:{project-ve
...
+
 [[debezium-inbound]]
 === Inbound Debezium Channel Adapter
 
-The Debezium adapter expects a pre-configured `DebeziumEngine.Builder<ChangeEvent<byte[], byte[]>>` bean instance.
+The Debezium adapter expects a pre-configured `DebeziumEngine.Builder<ChangeEvent<byte[], byte[]>>` instance.
+
+[TIP]
+====
+The https://github.com/spring-cloud/stream-applications/tree/main/functions/supplier/debezium-supplier[debezium-supplier] provides an out of the box `DebeziumEngine.Builder` SpringBoot auto-configuration with a handy https://github.com/spring-cloud/stream-applications/blob/main/functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumProperties.java[DebeziumProperties] configuration abstraction.
```
artembilan, line 59: Typo: Spring Boot

src/reference/asciidoc/debezium.adoc
```diff
@@ -25,10 +23,47 @@ compile "org.springframework.integration:spring-integration-debezium:{project-ve
...
+====
+
+[TIP]
+====
+The <<debezium-java-dsl,Debezium Java DSL>> can create a `DebeziumMessageProducer` instance from a provided `DebeziumEngine.Builder`, as well as from a plain Debezium configuration (e.g. `java.util.Properties`). Later can be handy for some common use-cases with opinionated configuration and serialization formats.
```
artembilan, line 64: You still didn't got used to the `One sentence per line` rule 😄

artembilan (changes requested):

src/reference/asciidoc/debezium.adoc
artembilan: OK. Good to hear that it works, but it is not readable with that style.
Probably the anchor `[[]]` works from any point of text, even in the middle of the sentence.
but does it really make sense to do that?

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslWithBuilderTests.java
```diff
@@ -0,0 +1,105 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Christian Tzolov
+ *
+ * @since 6.2
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class DebeziumDslWithBuilderTests implements DebeziumMySqlTestContainer {
```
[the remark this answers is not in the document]
tzolov: Each teast uses different Spring @Configuration and integrationflows

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
tzolov: Personally, I condider the telescopic constructors for anti-pattern. For me it is the same as to telescopic public methods in same class.

artembilan: I assume you talk about this this one: https://www.vojtechruzicka.com/avoid-telescoping-constructor-pattern/.

But I don't see a relevance: you build a `Builder` based on a provided props and default formats and then you delegated to the method which is already based on a `Builder` pattern.
How is that telescoping? We don't stack method calls here for extending arguments number.

What do I miss, please?

Good to know that it is an anti-pattern, but sometimes we really have to enforce end-user with our API for this or that choice, where something is so important that it cannot be extracted to a setter.

artembilan: And how about DRY principle?

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslWithBuilderTests.java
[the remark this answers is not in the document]
tzolov: I see, you have a point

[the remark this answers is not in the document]
tzolov: I'm taking my words back. 
Indeed you can configure two IntegrationFlow beans in the same configuration. But how can you ensure that only one will be active during its corresponding test is running? Using conditional annotations?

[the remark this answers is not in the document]
artembilan: Well, that's kinda the point: to ensure with one test class and complex configuration that everything is ran and works really as expected.
We indeed start an application context once and therefore all the channel adapters are started, but then in every test method we verify result of each of them.
If several Debezium channel adapters cannot be started in a single application context and they cannot work well against the same CDC data, then something is wrong with our solution or so.

artembilan (changes requested):

artembilan (changes requested):
@tzolov ,

Let me know your final thoughts about remaining review comments - and I'll go from there with a merge or so!

Thanks

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslWithBuilderTests.java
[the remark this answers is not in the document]
tzolov: Not sure what you mean. Perahps you can show it code?

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/Debezium.java
tzolov: I guess here we should agree to disagree. It is still a bad practice in my book. 
But this is your project and your responsibility so please feel free to rewrite and amend it you think is appropriate before merge.

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslTests.java
```diff
@@ -0,0 +1,98 @@
...
+@DirtiesContext
+public class DebeziumDslTests implements DebeziumMySqlTestContainer {
+
+	@Autowired
+	private Config config;
+
+	@Test
+	void dslBatch() throws InterruptedException {
+		assertThat(this.config.latch.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(config.payloads).hasSize(52);
```
artembilan, line 56: This batch stuff has a flaw: not a fact that all the CDC records are pulled at once for a first try.
An existing `DebeziumBatchTests` fails sporadically because the batch size produces with a first message is not `52` records.

I'm not sure by the logic of that Debezium sample what exactly are those `52` records, but probably we an adjust functionality to assert `at least` or have some more controllable value.

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslWithBuilderTests.java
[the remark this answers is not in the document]
artembilan: Not sure why is the question: all your tests are configured against the same container, so you just move the other DSL-related test and its configuration to this class. 
As I said before: we need to be sure that several Debezium channel adapters can work in the same application in parallel and even against the same data.

artembilan (changes requested):

artembilan (changes requested):

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/dsl/DebeziumDslTests.java
```diff
@@ -0,0 +1,151 @@
...
+			DebeziumMessageProducerSpec dsl = Debezium
+					.inboundChannelAdapter(
+							DebeziumMySqlTestContainer.connectorConfig(DebeziumMySqlTestContainer.mysqlPort()))
+					.headerNames("*")
+					.contentType("application/json")
+					.enableBatch(true);
+
+			return IntegrationFlow.from(dsl)
+					.handle(m -> {
+						List<ChangeEvent<byte[], byte[]>> batch = (List<ChangeEvent<byte[], byte[]>>) m.getPayload();
```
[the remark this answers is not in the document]
artembilan: This request has not been resolved yet.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/dsl/DebeziumMessageProducerSpec.java
```diff
@@ -0,0 +1,126 @@
...
+ * A {@link org.springframework.integration.dsl.MessageProducerSpec} for {@link DebeziumMessageProducer}.
+ *
+ * @author Christian Tzolov
+ *
+ * @since 6.2
+ */
+public class DebeziumMessageProducerSpec
+		extends MessageProducerSpec<DebeziumMessageProducerSpec, DebeziumMessageProducer> {
+
+	protected DebeziumMessageProducerSpec(Builder<ChangeEvent<byte[], byte[]>> debeziumEngineBuilder) {
```
artembilan, line 45:
```suggestion
	protected DebeziumMessageProducerSpec(DebeziumEngine.Builder<ChangeEvent<byte[], byte[]>> debeziumEngineBuilder) {
```

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/it/DebeziumStreamTests.java
```diff
@@ -47,14 +48,16 @@
 @DirtiesContext
 public class DebeziumStreamTests implements DebeziumMySqlTestContainer {
 
+	static final int EXPECTED_DB_TX_COUNT = 52;
```
artembilan, line 51: Can this constant go to that `DebeziumMySqlTestContainer` since it is really a part of that container environment.
I understand that we may have other tests where we add more transactions, but at lest this one is about initial DB setup.

artembilan (changes requested):

artembilan (approved):
LGTM.
Will pull locally for final review when CI build is green.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/c9023d114b653d185f8cce3cce2ff9b376249fe1 after some minor clean up.

PR #8636, opened 2023-05-31
Introduce HeaderFilterSpec to streamline DSL API

artembilan:
The concern has been driven by the discussion from: https://github.com/spring-projects/spring-integration/issues/8625 The point is that Java method arguments are not so descriptive when we read the code. Therefore, it is better to design DSL the way it would be cleaner from reading perspective. Plus less choice of methods to chain would give a better end-user experience from coding.

* Add a `HeaderFilterSpec` which can accept `headersToRemove` and `patternMatch` as individual options instead of top-level deprecated `headerFilter(headersToRemove, patternMatch)` `IntegrationFlow` method. This way Kotlin and Groovy DSLs get a gain from their "inner section" style.
* Such a `Consumer<HeaderFilterSpec>` way to configure an endpoint is similar to already existing `aggregate(Consumer<AggregatorSpec>)`, `resequence(Consumer<ResequencerSpec>)` etc. In other words those components which has a dedicated `ConsumerEndpointSpec` extension are OK from an idiomatic DSL style perspective
* Expose a `HeaderFilter.setHeadersToRemove()` to make it working smoothly with this new DSL requirements
* Apply a new `headerFilter()` style into Kotlin and Groovy DSLs

This is just an initial work to surface an idea.
If it is OK, I'll slow continue with others to realign and simplify the paradox of choice.


garyrussell (changes requested):
Otherwise LGTM; do you want this merged or will you add more work to this PR?

artembilan:
> do you want this merged 

Yes, I want this merged.
It is probably would be easier for you to review small chunks of changes in a scope of specific component.
Apparently this way of API refactoring going to affect those Kotlin and Groovy DSLs as well.
So, just one by one would be great. More over no push since it is not a breaking change (yet) and just nice addition to the existing API.

FYI: The failed test with Debezium batch mode is not related to the PR: I let @tzolov know that we have some race condition over there with batching or so.

PR #8639, opened 2023-06-07
Fix DSL for inner bean names generation

artembilan:
The `IntegrationFlowBeanPostProcessor.processIntegrationComponentSpec()` uses a wrong `generateBeanName()` for component to register making the provided `id` as a prefix

* Use `generateBeanName(Object instance, String prefix, @Nullable String fallbackId, boolean useFlowIdAsPrefix)` instead to properly "fallback" to the provided name

**Cherry-pick to `6.1.x`**


garyrussell:
...and cherry-picked to 6.1.x as 1a8450b1d8c260a2d51328f3de6b167ecb916954

PR #8640, opened 2023-06-07
Optimize synchronized in PartitionedDispatcher

artembilan:
Even if the `PartitionedDispatcher.populatedPartitions()` is fast, in-memory, non-blocking operation, its active call from the `dispatch()` on every message sent to the channel may pin the virtual thread.

* Optimize the `populatedPartitions()` for double `if` where we will step into a `synchronized` block only for first several concurrent messages

**Cherry-pick to `6.1.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
@@ -151,10 +151,16 @@ public class PartitionedDispatcher extends AbstractDispatcher {
 		return partitionDispatcher.dispatch(message);
 	}
 
-	private synchronized void populatedPartitions() {
+	private void populatedPartitions() {
 		if (this.partitions.isEmpty()) {
-			for (int i = 0; i < this.partitionCount; i++) {
-				this.partitions.put(i, newPartition());
+			synchronized (this.partitions) {
+				if (this.partitions.isEmpty()) {
+					Map<Integer, UnicastingDispatcher> partitionsToUse = new HashMap<>();
+					for (int i = 0; i < this.partitionCount; i++) {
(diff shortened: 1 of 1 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
```diff
@@ -151,10 +151,16 @@ public boolean dispatch(Message<?> message) {
 		return partitionDispatcher.dispatch(message);
 	}
 
-	private synchronized void populatedPartitions() {
+	private void populatedPartitions() {
 		if (this.partitions.isEmpty()) {
```
garyrussell, line 155: You will probably need a `// NOSONAR` here (and other unsynchronized uses) to prevent Sonar from complaining.

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
artembilan: I'm not sure how that suppose to complain since this property is `final`.
Plus our Sonar instance it out of order for a while.
I'd prefer to look into this when we come back into business with Sonar and it indeed will complain respectively.

garyrussell (approved):

garyrussell:
...and cherry picked as 0350a64df1

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
garyrussell: I don't think being final matters, I believe Sonar will report a smell for inconsistent synchronization of fields.

Issue #8638, opened 2023-06-07
KafkaProducerMessageHandler send failure messages for immediate exceptions


garyrussell:
### Discussed in https://github.com/spring-projects/spring-integration/discussions/8637

<div type='discussions-op-text'>

<sup>Originally posted by **yevsh** June  1, 2023</sup>
Sometimes there are  "error" network/kafka cluster errors - that can be only detected when we see in logs something like:

`"INFO","logger":"org.apache.kafka.clients.NetworkClient","message":"[Producer clientId=producer-1] Node 2 disconnected`.

The problem KafkaTemplate/KafkaProducerMessageHandler can't identify it as error, and data can be lost.

Need some reliable solution for this.</div>

PR #8641, opened 2023-06-08
GH-8638: Kafka: Send All Fails to Failure Channel

garyrussell:
Resolveshttps://github.com/spring-projects/spring-integration/issues/8638

Previously, immediate failures (e.g. timeout getting metadata) were only thrown as exceptions, and not sent to the failure channel, if present.

**cherry-pick to all supported branches**


artembilan:
... and cherry-picked to `6.1.x`, `6.0.x` & `5.5.x`

Issue #8642, opened 2023-06-08
Revise an ExecutorService injections in favor of injected TaskExecutor


artembilan:
Samples are `PostgresChannelMessageTableSubscriber`, `org.springframework.integration.hazelcast.leader.LeaderInitiator`.
These threads potentially can be created as virtual threads with Java 21.

artembilan:
What I don't like with a single managed `TaskExecutor` that threads in different components are going to have same prefix in their names.
It is a bit inconvenient when you trace the logic in the logs.
With the name of a thread dedicated to the specific component it is very clear what is going on.

Probably the whole paradigm of managed executor doesn't bring any value when we switch to virtual threads.
It might really be convenient from microservice perspective when Spring Boot gives us this or that `TaskExecutor` automatically, but for more complex and asynchronous applications like with Spring Integration, it is really misleading to use same prefix name for all virtual threads.

We can continue discussion though since I don't claim to be expert in this area: just sharing my pain over years with single managed executor.

Out of context, but still might be an argument for `TaskExecutor` injection in some places, @tzolov . See, there is an `org.springframework.core.task.support.ExecutorServiceAdapter` for our convenience when we would like to deal with `Future<?> submit()`.

garyrussell:
I didn't mean a single shared TE, each component can get its own.

In any case, as I've said before, unless something has changed, VTs don't get a `name` for logging purposes; I added this property to the listener container:

https://github.com/spring-projects/spring-kafka/issues/2501

artembilan:
Yeah...
The point is that we talk about those components which uses only a single thread, like `LockRegistryLeaderInitiator`.
But yeah, for consistency, a single-threaded `TE` could be used as well.

PR #8646, opened 2023-06-12
Executor instead of ThreadFactory for DebeziumMP

artembilan:
Related to https://github.com/spring-projects/spring-integration/issues/8642

For consistency with other Spring requirements and realignment with virtual threads, it is better to require a `TaskExecutor` injection instead of `ThreadFactory`

* Fix `DebeziumMessageProducer` to rely on a `TaskExecutor` API instead of `ThreadFactory` and `ExecutorService`

### DIFF
--- a/spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
+++ b/spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
@@ -21,9 +21,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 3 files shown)


spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
```diff
@@ -213,19 +202,6 @@ protected void doStop() {
 		}
 	}
 
-	@Override
-	public void destroy() {
```
tzolov, line 217: Aren't you supposed to gracefully stop the task executor on exit?

artembilan: See a `destroy()` impl in the super class. It does call that `stop();` for us.

tzolov:
@artembilan, 
iirc we picked the `ThreadFactory` initially, because we assumed that it will be easy to support virtual threads. 
Is the TaskExecutor switch affecting this?

artembilan:
> Is the `TaskExecutor` switch affecting this?

As far as I know that's the move Spring Framework (provides a `VirtualThreadTaskExecutor`) and Spring Boot (will auto-configure that `VirtualThreadTaskExecutor`) expect from us.
Plus asking for `TaskExecutor` injection is really what we do in many other places. Therefore for consistency I switch this to the same style as it is everywhere else in Spring.
See also what we have discussed with Gary in the issue related to this change: https://github.com/spring-projects/spring-integration/issues/8642

Well, the `PartitionedChannel` is an exception since that one really requires a dedicated thread for particular partition.
I won't mind, though, to migrate that one to a `TaskExecutor` as well, if there is the way to stick to the specific thread.

tzolov:
Ok, LGTM

PR #8647, opened 2023-06-13
GH-8642: Revise executors in the project

### LINKED ISSUE #8642: Revise an ExecutorService injections in favor of injected TaskExecutor

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8642

* Rework some `Executors.newSingleThreadExecutor()` to `ExecutorServiceAdapter(new SimpleAsyncTaskExecutor())`
* Expose `TaskExecutor` setters; deprecate `ExecutorService`-based
* Some other code clean up in the effected classes: `LogAccessor`, no `synchronized` in critical blocks
* Give a meaningful prefix for default threads in the context of components, e.g. `SubscribableRedisChannel` - `getBeanName() + "-"`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
@@ -18,25 +18,24 @@ package org.springframework.integration.support.leader;
 
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.context.SmartLifecycle;
+import org.springframework.core.log.LogAccessor;
+import org.springframework.core.task.SimpleAsyncTaskExecutor;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.core.task.support.ExecutorServiceAdapter;
 import org.springframework.integration.leader.Candidate;
 import org.springframework.integration.leader.Context;
 import org.springframework.integration.leader.DefaultCandidate;
 import org.springframework.integration.leader.event.DefaultLeaderEventPublisher;
 import org.springframework.integration.leader.event.LeaderEventPublisher;
 import org.springframework.integration.support.locks.LockRegistry;
-import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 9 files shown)


garyrussell (changes requested):
I started looking at this but I stopped after the first class; I don't think using the `ExecutorServiceAdapter` is correct - why do we need it? Why not use the TE directly?

The default SATE will never be stopped.

garyrussell:
Then force the TE to be an `AsyncTaskExecutor` instead (that's what I do in SK).

```java
	/**
	 * The executor for threads that poll the consumer.
	 */
	private AsyncTaskExecutor listenerTaskExecutor;
```

I don't see any way to stop the default TE otherwise.

artembilan:
> Then force the TE to be an `AsyncTaskExecutor`

Right. Missed that one. Thanks. Let me rework it then!

> I don't see any way to stop the default TE otherwise.

Right. With `SATE` as default  we really don't need to worry about its stop, since it does not cache threads and we really don't need them to be cached in those places.
The goal of this PR is to get rid of an `ExecutorService` as an injection point.

garyrussell:
```
org.springframework.beans.NotWritablePropertyException: Invalid property 'executorService' of bean class [org.springframework.integration.support.leader.LockRegistryLeaderInitiator]: Bean property 'executorService' has no matching field.
	at org.springframework.beans.DirectFieldAccessor.createNotWritablePropertyException(DirectFieldAccessor.java:96)
	at org.springframework.beans.AbstractNestablePropertyAccessor.processLocalProperty(AbstractNestablePropertyAccessor.java:431)
	at org.springframework.beans.AbstractNestablePropertyAccessor.setPropertyValue(AbstractNestablePropertyAccessor.java:279)
	at org.springframework.beans.AbstractNestablePropertyAccessor.setPropertyValue(AbstractNestablePropertyAccessor.java:247)
	at org.springframework.integration.support.leader.LockRegistryLeaderInitiatorTests.testGracefulLeaderSelectorExit(LockRegistryLeaderInitiatorTests.java:255)
```

spring-integration-core/src/main/java/org/springframework/integration/support/leader/LockRegistryLeaderInitiator.java
```diff
@@ -190,12 +180,23 @@ public LockRegistryLeaderInitiator(LockRegistry locks, Candidate candidate) {
 	/**
 	 * Set the {@link ExecutorService}, where is not provided then a default of
 	 * single thread Executor will be used.
-	 * @param executorService the executor service
+	 * @param taskExecutor the executor service
 	 * @since 5.0.2
+	 * @deprecated since 6.2 in favor of {@link #setTaskExecutor(AsyncTaskExecutor)}
+	 */
+	@Deprecated(since = "6.2", forRemoval = true)
+	public void setTaskExecutor(ExecutorService taskExecutor) {
```
garyrussell, line 188: Why the name change (and param)? Need to deprecate the existing method.

garyrussell (changes requested):

garyrussell (approved):

PR #8648, opened 2023-06-13
[DO NOT Merge YET] Debezium incremental snapshot tests

tzolov:
- Testing the incremental snapshot support in Debezium.
 - Tests for MySQL, Postgres and SQL Server.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -589,7 +589,11 @@ project('spring-integration-debezium') {
         }
 
         testImplementation "io.debezium:debezium-connector-mysql:$debeziumVersion"
-        testImplementation 'org.testcontainers:mysql'
+        testImplementation "io.debezium:debezium-connector-postgres:$debeziumVersion"
+        testImplementation "io.debezium:debezium-connector-sqlserver:$debeziumVersion"
+        testImplementation "org.springframework:spring-jdbc"
+        testImplementation "com.zaxxer:HikariCP:4.0.3"
+        testImplementation "org.springframework.boot:spring-boot-autoconfigure:3.1.0"
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -592,8 +592,6 @@ project('spring-integration-debezium') {
         testImplementation "io.debezium:debezium-connector-postgres:$debeziumVersion"
         testImplementation "io.debezium:debezium-connector-sqlserver:$debeziumVersion"
         testImplementation "org.springframework:spring-jdbc"
-        testImplementation "com.zaxxer:HikariCP:4.0.3"
-        testImplementation "org.springframework.boot:spring-boot-autoconfigure:3.1.0"
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -175,6 +175,8 @@ allprojects {
             mavenBom "org.springframework.data:spring-data-bom:$springDataVersion"
             mavenBom "org.springframework:spring-framework-bom:$springVersion"
             mavenBom "org.springframework.ws:spring-ws-bom:$springWsVersion"
+            mavenBom "io.debezium:debezium-bom:$debeziumVersion"
+
         }
 
     }
@@ -581,16 +583,17 @@ project('spring-integration-debezium') {
     description = 'Spring Integration Debezium Support'
     dependencies {
         api project(':spring-integration-core')
-        api("io.debezium:debezium-embedded:$debeziumVersion") {
+        api("io.debezium:debezium-embedded") {
             exclude group: 'org.glassfish.jersey.containers', module: 'jersey-container-servlet'
             exclude group: 'org.glassfish.jersey.inject', module: 'jersey-hk2'
             exclude group: 'jakarta.xml.bind', module: 'jakarta.xml.bind-api'
             exclude group: 'jakarta.activation', module: 'jakarta.activation-api'
         }
 
-        testImplementation "io.debezium:debezium-connector-mysql:$debeziumVersion"
-        testImplementation "io.debezium:debezium-connector-postgres:$debeziumVersion"
-        testImplementation "io.debezium:debezium-connector-sqlserver:$debeziumVersion"
+        testImplementation 'org.testcontainers:mysql'
+        testImplementation "io.debezium:debezium-connector-mysql"
+        testImplementation "io.debezium:debezium-connector-postgres"
+        testImplementation "io.debezium:debezium-connector-sqlserver"
         testImplementation "org.springframework:spring-jdbc"
     }
 }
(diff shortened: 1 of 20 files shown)


build.gradle
```diff
@@ -589,6 +589,9 @@ project('spring-integration-debezium') {
         }
 
         testImplementation "io.debezium:debezium-connector-mysql:$debeziumVersion"
+        testImplementation "io.debezium:debezium-connector-postgres:$debeziumVersion"
+        testImplementation "io.debezium:debezium-connector-sqlserver:$debeziumVersion"
```
artembilan, line 593: This is what I meant about Debezium BOM: https://repo1.maven.org/maven2/io/debezium/debezium-bom/2.3.0.CR1/debezium-bom-2.3.0.CR1.pom

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/snapshots/AbstractIncrementalSnapshotTest.java
```diff
@@ -0,0 +1,340 @@
...
+						+ config.messagesPerTableMap.get(destination).size());
+
+		config.messagesPerTableMap.get(destination).stream()
+				.forEach(m -> assertThat(operationType(m)).isEqualTo(expectedOperationType));
+
+	}
+
+	@Configuration
+	@EnableIntegration
+	static class StreamTestConfiguration {
```
artembilan, line 302: Why is it  called `Stream`, please?

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/snapshots/MySqlIncrementalSnapshotTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+import static org.awaitility.Awaitility.await;
+
+/**
+ * @author Christian Tzolov
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class MySqlIncrementalSnapshotTest extends AbstractIncrementalSnapshotTest implements MySqlTestContainer {
```
artembilan, line 43: I'm OK with `*Test` suffix for an `abstract` class, but, please, call top-level test classes as `*Tests`

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/snapshots/PostgresIncrementalSnapshotTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * @author Christian Tzolov
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class PostgresIncrementalSnapshotTest extends AbstractIncrementalSnapshotTest implements PostgresTestContainer {
+
+	@BeforeAll
+	public static void beforeEach() throws InterruptedException {
```
artembilan, line 46: The name of method does not reflect its purpose.

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/snapshots/PostgresIncrementalSnapshotTest.java
```diff
@@ -0,0 +1,169 @@
...
+		Thread.sleep(5000);
```
artembilan, line 47: I'm not sure why we have to sleep for 5 seconds in the beginning.
Cannot we just have some short-sleeping loop until some condition is fulfilled?

spring-integration-debezium/src/test/java/org/springframework/integration/debezium/snapshots/PostgresIncrementalSnapshotTest.java
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	protected void debeziumReadyCheck() {
+	}
+
+	protected void insertCustomer(String firstName, String lastName, String email) {
+		jdbcTemplate.update("INSERT INTO inventory.customers VALUES (default,?,?,?)", firstName, lastName, email);
+		// int found = jdbcTemplate.queryForObject( "SELECT count(*) FROM inventory.customers WHERE first_name like ?",
+		// Integer.class, firstName);
```
artembilan, line 56: Why do we need to have this commented code here?

spring-integration-debezium/src/test/resources/docker/mysql/inventory.sql
```diff
@@ -0,0 +1,124 @@
...
+	g GEOMETRY NOT NULL,
+	h GEOMETRY);
+
+INSERT INTO geom
+VALUES(default, ST_GeomFromText('POINT(1 1)'), NULL),
+      (default, ST_GeomFromText('LINESTRING(2 1, 6 6)'), NULL),
+      (default, ST_GeomFromText('POLYGON((0 5, 2 5, 2 7, 0 7, 0 5))'), NULL);
+
+# Create a signalling table
+CREATE TABLE dbz_signal (
```
artembilan, line 121: Looks like we just add this table into DB.
Everything else is exactly the same what they have so far with their standard `debezium/example-mysql` image.
Why cannot we just call such a DDL manually after starting Testcontainer or even propagate there some extra script like they do in their Docker file: https://github.com/debezium/container-images/blob/main/examples/mysql/2.3/Dockerfile ?

artembilan (changes requested):

spring-integration-debezium/src/test/resources/docker/mysql/inventory.sql
tzolov: - Already tried to use MySQLContainer + debezium/example-mysql, but it complain that that the example-mysql is not a mysql image. 
- I don't like messing with DDLs inside the test code, given that we have applly customesd docker image.

artembilan: > but it complain that that the example-mysql is not a mysql image.

So, you didn't pay attention that I had that `.asCompatibleSubstituteFor("mysql")` in my change.

> I  don't like messing with DDLs inside the test code

And I don't like having files in the project which are really not related to the project.
That is going to be just a short script for extra table and same official image.

And from here I might prefer to not have that MSSQL integration test since there is no Docker image for it from Debezium repository.
Same snapshoting functionality is covered with other two databases.

The burden of future support of this staff may neglect the effort we spend on it giving the fact that there is no any tool in the project for end-users (yet).

I just want to be reasonable in what we are doing and why.

tzolov:
Decided to maintain those tests outside the SI project. 
One can find more about the Debezium Incremental snapshots [here](https://github.com/tzolov/spring-debezium-demos)

IT tests are here: https://github.com/tzolov/spring-debezium-demos/tree/main/src/test/java/com/example/sidebeziumdemo/it/snapshots2

PR #8651, opened 2023-06-15
Optimise reading node maybeIndex in Json Property Accessor

fvlad:
I have a service which is actively evaluating SpEL expressions on top of JsonNode. 10s k messages/second are processed upon SpEL expressions filtering.
When analysing its performance I found that a large amount of CPU is dedicated to handling NumberFormatException inside SpEL.
It seems that for every read/canRead operation JsonPropertyAccessor is trying to check whether the property is a number by parsing it with Integer.valueOf inside maybeIndex method. This approach generates a lot of hidden exceptions, which degrade performance significantly.
The following fix allowed me to improve CPU consumption about X3 times on my service. In the screenshot you can see CPU consumption of the service for the last 24h period with a fix deployed at about 11:00.
image

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/json/JsonPropertyAccessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/json/JsonPropertyAccessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2022 the original author or authors.
+ * Copyright 2013-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ import org.springframework.expression.PropertyAccessor;
 import org.springframework.expression.TypedValue;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * A SpEL {@link PropertyAccessor} that knows how to read properties from JSON objects.
@@ -41,6 +42,7 @@ import org.springframework.util.Assert;
  * @author Paul Martin
  * @author Gary Russell
  * @author Pierre Lakreb
+ * @author Vladislav Fefelov
  *
  * @since 3.0
  */
@@ -109,12 +111,31 @@ public class JsonPropertyAccessor implements PropertyAccessor {
 	 * Return an integer if the String property name can be parsed as an int, or null otherwise.
 	 */
 	private static Integer maybeIndex(String name) {
+		if (!isNumeric(name)) {
+			return null;
+		}
 		try {
 			return Integer.valueOf(name);
 		}
 		catch (NumberFormatException e) {
 			return null;
 		}
+	}
+
+	/**
+	 * Check if the string is a numeric representation (all digits) or not.
+	 */
+	private static boolean isNumeric(String str) {
+		if (!StringUtils.hasLength(str)) {
+			return false;
+		}
+		final int length = str.length();
+		for (int i = 0; i < length; i++) {
+			if (!Character.isDigit(str.charAt(i))) {
+				return false;
+			}
+		}
+		return true;
 	}
 
 	@Override


spring-integration-core/src/main/java/org/springframework/integration/json/JsonPropertyAccessor.java
```diff
@@ -117,6 +120,22 @@ private static Integer maybeIndex(String name) {
 		}
 	}
 
+	/**
+	 * Check if the string is a numeric representation (all digits) or not.
+	 */
+	private static boolean isNumeric(String str) {
+		if (str == null || str.isEmpty()) {
```
artembilan, line 127: This can be replaced with `org.springframework.util.StringUtils.hasText()`

spring-integration-core/src/main/java/org/springframework/integration/json/JsonPropertyAccessor.java
```diff
@@ -117,6 +120,22 @@ private static Integer maybeIndex(String name) {
...
+			return false;
+		}
+		final int length = str.length();
```
artembilan, line 130: We don't need to extract this variable: can simply be done inside `for()` expression.

artembilan (changes requested):
I left a couple suggestions.
And one more: add your name to `@author` list.

Thank you!

spring-integration-core/src/main/java/org/springframework/integration/json/JsonPropertyAccessor.java
fvlad: Good point. But I think using `org.springframework.util.StringUtils.hasLength()` will be even better. Because `hasText()` will check for non empty string and additionally iterate over it looking for non-whitespace characters. Which is a waste of CPU in cases when we get all-whitespace strings here like "_____".
Replaced with `hasLength()`.

spring-integration-core/src/main/java/org/springframework/integration/json/JsonPropertyAccessor.java
artembilan: OK. Found another way to improve its performance:
```

	@Test
	void stringIterationPerformance() {
		int tmp = 0;
		String s = new String(new byte[64*1024]);

		{
			long st = System.nanoTime();
			int n = s.length();
			for(int i = 0; i < n; i++) {
				tmp += s.charAt(i);
			}
			st = System.nanoTime() - st;
			System.out.println("1 " + st);
		}

		{
			long st = System.nanoTime();
			for(char c : s.toCharArray()) {
				tmp += c;
			}
			st = System.nanoTime() - st;
			System.out.println("2 " + st);
		}
	}
```
The result is:
```
1 1680300
2 807000
```

So, `for(char c : s.toCharArray()) {` is twice faster then `for(int i = 0; i < n; i++) {`

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/json/JsonPropertyAccessor.java
fvlad: I'm curious in your results :)

Even if we iterate string using toCharArray first, it will create an additional memory allocation in size of the whole string every time. But in many cases in this method we will break at the first character and return. We are not trying to sum characters, we are looking for the first non-digit. On practice the first character in JSON key here is usually either digit (and the rest are as well) or not (and we return immediately). So this performance optimisation (if it does optimise something) will be relevant only for all digit strings, while significantly harming the performance of non-digit keys check.

But even if you want to test the performance of your alternative, I think the proper way would be to run it at least with something like JMH. For example:
```
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Benchmark)
@Fork(value = 1)
@Warmup(iterations = 3, time = 1)
@Measurement(iterations = 5, time = 1)
public class StringIterationBenchmark {

    private static final String s = new String(new byte[64*1024]);


    public static void main(String[] args) throws RunnerException {
        Options opt = new OptionsBuilder()
            .include(StringIterationBenchmark.class.getSimpleName())
            .build();

        new Runner(opt).run();
    }

    @Benchmark
    public int iterateWithCharAt() {
        int tmp = 0;

        int n = s.length();
        for(int i = 0; i < n; i++) {
            tmp += s.charAt(i);
        }

        return tmp;
    }

    @Benchmark
    public int iterateWithToCharArray() {
        int tmp = 0;

        for(char c : s.toCharArray()) {
            tmp += c;
        }

        return tmp;
    }

}
```

Which on my machine (M1 MacBook Pro) doesn't look like an improvement:
```
Benchmark                                        Mode  Cnt  Score    Error  Units
StringIterationBenchmark.iterateWithCharAt       avgt    5  0.021 ±  0.001  ms/op
StringIterationBenchmark.iterateWithToCharArray  avgt    5  0.032 ±  0.001  ms/op
```

artembilan: OK. Don't have any objections any more.
I'm not familiar with JMH: will learn it other day.

For now I'm accepting your fix as is and back-porting it respectively to all supported versions.

Thank you for education!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/8a29c1e60bc186b8fced4abed7e6b73d45db9416 and cherry-picked to `6.1.x`, `6.0.x` & `5.5.x`.

@fvlad ,

FYI: we have releases next Tuesday for `6.1.1` and `6.0.6`: https://calendar.spring.io/.
Does it work for you?
We may also schedule `5.5.19` if that is what you would expect to have a fix in.

fvlad:
Thanks! I patched it locally for myself, so no rush on my end.

PR #8653, opened 2023-06-16
GH-8626: Provide cleaner `transform()` DSL

### LINKED ISSUE #8626: Add DSL transform() variant overload for accepting a bean name
As discussed on thread for Issue #8625, there is a utility to have IntegrationFlows methods that accept a bean name, as is the case for 

```split(String beanName, @Nullable String methodName)``` 

The ```transform()``` method does not have such a bean-name overload.  It can be useful to have this in order to avoid the boilerplate injection of the bean via construction or bean method.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8626

* Add missed `transform(String beanName, @Nullable String methodName)` API
* Introduce a `TransformerSpec` to expose a strict API to configure transformer variants.
* Introduce `transformWith(Consumer<TransformerSpec>)` as a single point of all possible transformer and its endpoint options
* Deprecate those `IntegrationFlowDefinition.transform()` variants which are harder to configure as several lambda arguments

This change will make Kotlin & Groovy DSL more readable and straightforward

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/BaseIntegrationFlowDefinition.java
@@ -538,12 +538,18 @@ public abstract class BaseIntegrationFlowDefinition<B extends BaseIntegrationFlo
 	/**
 	 * Populate the {@code Transformer} EI Pattern specific {@link MessageHandler} implementation
 	 * for the SpEL {@link Expression}.
+	 * Shortcut for:
+	 * <pre class="code">
+	 * {@code
+	 *  .transformWith((transformerSpec) -> transformerSpec.expression(expression))
+	 * }
+	 * </pre>
 	 * @param expression the {@code Transformer} {@link Expression}.
 	 * @return the current {@link BaseIntegrationFlowDefinition}.
 	 * @see ExpressionEvaluatingTransformer
 	 */
 	public B transform(String expression) {
-		return transform(expression, (Consumer<GenericEndpointSpec<MessageTransformingHandler>>) null);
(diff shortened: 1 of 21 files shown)


artembilan:
@garyrussell ,

I have a naming problem 😄 

my new DSL method is `transformWith(Consumer<TransformerSpec> transformerConfigurer)`.
If I leave it just as `transform()`, the "typeless" lambda style configuration would clash with an existing `transform(GenericTransformer<S, T> genericTransformer)`.
Therefore only way to satisfy compiler is to give the method a different name.

So, is that `transformWith` OK? Or do you see any other possible names?

NOTE: The change in not done yet.
1. Revise tests in other modules.
2. Fix docs
3. The Kotlin & Groovy DSLs might be done in the separate PR.

I don't deprecate many other `transform()` methods because I call them shortcuts. It is really more straightforward just to type `.<String, String>transform(String::toUpperCase)`, then try something like `.transformWith(t -> t.<String, String>transformer(String::toUpperCase))`.
Or similar for just expression, or bean method call.
Probably for Kotlin & Groovy we won't have them since it looks like their DSL is more about options and their values rather then method calls.

Thanks

garyrussell:
`transformWith()` is fine for me. Should I wait for other changes before reviewing?

artembilan:
Yes, please.
I've already applied the change for Groovy DSL to justify it.
So, only Kotlin is left and docs.

Thanks

garyrussell (approved):
LGTM; needs to be removed from Draft mode for merge.

garyrussell:
Removed from draft mode - let me know if that is not what you intended.

artembilan:
> Removed from draft mode

Yes. That is exactly what we need now.
Feel free to merge.
Thanks.

PR #8657, opened 2023-06-26
GH-8654: Fix bean deps for messaging annotations

### LINKED ISSUE #8654: Fix dependency registration for Messaging annotation parsers
**Expected Behavior**

I'm building an application which transfers/transforms messages between TCP and MQTT.
I would like to monitor the connections, to be able to configure some alerts when connections stay down for a certain amount of time/consecutive negative checks.

**Current Behavior**

I implemented multiple @Scheduled methods with a fixed delay of x seconds to check both the status of Tcp and Mqtt connections. For Tcp I'm checking if there are open connections by calling `getOpenConnectionIds()` on either `AbstractClientConnectionFactory` or `AbstractServerConnectionFactory` which works fine and does the job.

I'm trying to do something similar for checking the connection (NOT subscription) with the Mqtt-broker but without success. The most obvious way would be to call the `isConnected()` method on `IMqttAsyncClient` in either `Mqttv5PahoMessageDrivenChannelAdapter` or `Mqttv5PahoMessageHandler`, but since this client has private access, I'm not able to check the connection.

**Context**

Is there any way of achieving this with the existing codebase? If not, would it be possible to:

- provide a getter which returns if the client is connected in both `Mqttv5PahoMessageDrivenChannelAdapter` and `Mqttv5PahoMessageHandler`
```
	public boolean isConnected() {
		return mqttClient.isConnected();
	}

```
- publish internal application events similar to the `TcpConnectionOpenEvent` once the Mqtt client is connected successfully to the broker?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8654

Spring Framework has an ability to start dependant beans automatically when we start the current one.

The `AbstractMethodAnnotationPostProcessor` is missing a bean dependency registration causing errors in target applications when Messaging Annotations configuration is used.

* Add `registerDependentBean()` into an `AbstractMethodAnnotationPostProcessor` when we generate and register a `ConsumerEndpointFactoryBean`
* Change one of the `ClientManagerBackToBackTests` configuration to rely on a `@ServiceActivator` for `Mqttv5PahoMessageHandler` bean to ensure that change in the `AbstractMethodAnnotationPostProcessor` has a proper effect

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/AbstractMethodAnnotationPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/AbstractMethodAnnotationPostProcessor.java
@@ -270,6 +270,7 @@ public abstract class AbstractMethodAnnotationPostProcessor<T extends Annotation
 		String endpointBeanName =
 				generateHandlerBeanName(beanName, mergedAnnotations)
 						.replaceFirst("\\.(handler|source)$", "");
+		this.beanFactory.registerDependentBean(beanName, endpointBeanName);
 		this.definitionRegistry.registerBeanDefinition(endpointBeanName, endpointBeanDefinition);
 	}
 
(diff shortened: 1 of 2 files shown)


garyrussell (approved):

garyrussell:
...and cherry picked to 6.1.x 8c0ffe3dd9 and 6.0.x 2b724c8a5d

PR #8660, opened 2023-06-27
GH-8611: Extract `IntegrationConfigurationReport`

### LINKED ISSUE #8611: Decouple `BeanDefinitionRegistryPostProcessor` implementations
While reviewing [an issue in Spring Framework](https://github.com/spring-projects/spring-framework/issues/30411#issuecomment-1532798212), I've noticed that Spring Integration has BeanDefinitionRegistryPostProcessor implementations that are also doing something else. 

We're considering reviewing this contract so that such post-processors are automatically excluded from AOT processing (as they operation on the bean factory that's pre-processed by it). Can you please review `DefaultConfiguringBeanFactoryPostProcessor` and `MessagingAnnotationPostProcessor` so that they only perform bean definition registry post-processing and nothing else?

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8611

According to the latest Spring Framework requirements related to AOT the `BeanDefinitionRegistryPostProcessor` must not do anything but only bean registrations

* Extract an `IntegrationConfigurationReport` component and move `IntegrationProperties` printing over here out from the `DefaultConfiguringBeanFactoryPostProcessor`
* Remove a `SmartInitializingSingleton` impl from the `DefaultConfiguringBeanFactoryPostProcessor`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
@@ -16,17 +16,14 @@
 
 package org.springframework.integration.config;
 
-import java.io.PrintWriter;
-import java.io.StringWriter;
+import java.beans.Introspector;
 import java.util.HashSet;
-import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ThreadPoolExecutor.CallerRunsPolicy;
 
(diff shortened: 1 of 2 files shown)


artembilan:
This is part 1 since there is yet that `MessagingAnnotationPostProcessor` which is much complicated to decouple.
So, I'm considering that as a separate  PR since it is going to be much involved and would over complicate review for this simpler part.

garyrussell (approved):

Issue #8659, opened 2023-06-27
New file is not detected when renaming the sub-folder of the watched folder (Java 7 WatchService)


pziobron:
**In what version(s) of Spring Integration are you seeing this issue?**
6.0.1.RELEASE

**Describe the bug**
The bug is related to the **_FileReadingMessageSource_** when using it for watching the directory with the (Java 7) WatchService feature enabled.
More specifically, the new files are not detected in the sub-directory of the watched folder after renaming it.
This issue has been discovered on Windows when using File Explorer to create a new directory ('New Folder' temporary name is used initially and user edits the name after that). In such a case 2 file events are generated. 

**To Reproduce**
1. Add the following logger in the logback-spring.xml to see the debug messages of the FileReadingMessageSource class
```
<logger name="org.springframework.integration.file" level="DEBUG" additivity="false">
   <appender-ref ref="Console" />
  <appender-ref ref="RollingFile" />
</logger>
```
2. Use the following FileReadingMessageSource instance to watch the directory [C:\sandbox\test]:
 ```
FileReadingMessageSourceImproved messageSource = new FileReadingMessageSourceImproved();
messageSource.setDirectory(new File(inputDirectory));
messageSource.setUseWatchService(true);
messageSource.setWatchEvents(CREATE, MODIFY);
```
3. [case A - New Folder] Create new sub-directory (PL) using Windows File Explorer in the watched folder and observe the logs:
 ```
Watch event [ENTRY_CREATE] for file [C:\sandbox\test\New folder]
registering: C:\sandbox\test\New folder for file events
Watch event [ENTRY_CREATE] for file [C:\sandbox\test\PL]
registering: C:\sandbox\test\PL for file events
```
4. [case A - New Folder] Copy file (6017405.eml) into the new PL directory and observe the logs (**The file is ignored !**):
```
Watch event [ENTRY_CREATE] for file [C:\sandbox\test\New folder\6017405.eml]
A file [C:\sandbox\test\New folder\6017405.eml] for the event [ENTRY_CREATE] doesn't exist. Ignored.
```
5. [case B - Renaming the folder] Now, create a new empty directory(**de**) in the watched folder and observe the logs:
```
Watch event [ENTRY_CREATE] for file [C:\sandbox\test\de]
registering: C:\sandbox\test\de for file events
```
6. [case B - Renaming the folder] Rename the folder to **ee** and observe the logs:
```
[...]

PR #8662, opened 2023-06-28
GH-8659: Fix the WatchServiceDirectoryScanner to detect new file afte…

pziobron:
Fix the WatchServiceDirectoryScanner to detect new file after renaming the sub-folder of the watched folder (Java 7 WatchService)

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
@@ -533,6 +533,10 @@ public class FileReadingMessageSource extends AbstractMessageSource<File> implem
 			logger.debug(() -> "Watch event [" + event.kind() + "] for file [" + file + "]");
 
 			if (StandardWatchEventKinds.ENTRY_DELETE.equals(event.kind())) {
+				if (this.pathKeys.containsKey(file.toPath())) {
+					WatchKey watchKey = this.pathKeys.remove(file.toPath());
+					watchKey.cancel();
+				}
 				if (getFilter() instanceof ResettableFileListFilter<File> resettableFileListFilter) {
 					resettableFileListFilter.remove(file);
 				}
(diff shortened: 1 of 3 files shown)


artembilan (changes requested):
All looks good, however I would like to suggest some change to the `file.adoc`.
Go to the `[[watch-service-directory-scanner]]` section and explain there in the end that we need to track also `WatchEventType.MODIFY` and `WatchEventType.DELETE` event types if sub-directories can be renamed.

artembilan (changes requested):
The test has failed, probably due to your change: 
```
WatchServiceDirectoryScannerTests > testWatchServiceDirectoryScanner() FAILED
    java.lang.AssertionError at WatchServiceDirectoryScannerTests.java:220
```

Please, revise.

pziobron:
The fix seems to solve the problem on Windows - the files are detected properly in the sub-folder (of the watched folder) after renaming it.
I've tried to write a test for this scenario but it is hard to emulate it in Java (creating empty folder in File Explorer and renaming it). I am not sure if that's even possible.
I will experiment with that a bit :)

pziobron:
I've tried to replicate the issue in Java in order to write the Integration test which covers that issue.
Unfortunately, I haven't managed to do that.
I've been using: 

1. File.mkdir()
2. File.renameTo(...)/Files.move(...)
3. File.createTempFile(...)

with no success - the file in sub-directory is detected properly.

The issue can be replicated in Windows with the File Explorer or with the Command Line by using the following commands:

1. md test1
2. move test1 test2
3. copy someFile test2

Theoretically, the Runtime.getRuntime().exec(...) could be used for the integration test but it would work obviously only for Windows (or for all Os'es when mapping md -> mkdir, copy -> cp etc....).
Nevertheless, I am not sure if that is a good idea (this could be tricky) ?

I've updated the documentation as You requested.

artembilan (changes requested):

artembilan:
... and cherry-picked to `6.1.x`, `6.0.x` & `5.5.x`.

@pziobron ,

PR #8665, opened 2023-06-29
GH-8664: Do not use broken `observeWithContext()`

### LINKED ISSUE #8664: Spring integration error when using MessageRequestReplyReceiverContext data in observation predicate

`id 'org.springframework.boot' version '3.1.1'`

`implementation("org.springframework.integration:spring-integration-ip")`

**Describe the bug**

When using the `ObservationPredicate` with Spring boot micrometer, I created some rules in order to observe certain contexts. But when using the context `MessageRequestReplyReceiverContext` and using some of its data in order to create a predicate, the rest of the chain of the application just fails with the exception:
```
java.lang.ClassCastException: class io.micrometer.observation.Observation$Context cannot be cast to class org.springframework.integration.support.management.observation.MessageRequestReplyReceiverContext (io.micrometer.observation.Observation$Context and org.springframework.integration.support.management.observation.MessageRequestReplyReceiverContext are in unnamed module of loader 'app')
```

Our application flow is just a simple TCP nio inbound, with a broker outbound. But it seems to be going wrong before we receive it in the the TCP inbound.

```
        observationRegistry.observationConfig()
                .observationPredicate((name, context) -> shouldBeObserved(context));

```

Example of usage in observation predicate when using headers: 
```
        ...
        
        if (context instanceof MessageRequestReplyReceiverContext messageRequestReplyReceiverContext) {
            return Optional.ofNullable(messageRequestReplyReceiverContext.getCarrier())
                    .map(Message::getHeaders)
                    .map(headers -> !headers.containsKey("SOMEKEY"))
                    .orElse(true);
        }

       ...
```

Example of usage in observation predicate when using payload: 

```
        ...
        
        if (context instanceof MessageRequestReplyReceiverContext messageRequestReplyReceiverContext) {
            return Optional.ofNullable(messageRequestReplyReceiverContext.getCarrier())
                    .map(Message::getPayload)
                    .map(payload -> new String((byte[]) payload, StandardCharsets.UTF_8))
                    .map(payload -> !payload.equals(SOMEDATA))
                    .orElse(true);
        }

       ...
```

Exception:
image

**Expected behavior**

Expected to not break the application when simply casting the payload or using the headers in the generic message.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8664

When an `Observation` is turned to `NoopObservation`, the `observeWithContext()` fails with `ClassCastException` since `NoopObservation` serves just plain `Context` not the one we supplied

* Fix `MessagingGatewaySupport.sendAndReceiveWithObservation()` same way as it is in version `6.0.x`
* Modify `IntegrationObservabilityZipkinTests` to reject some `Observation` via `observationPredicate()` configuration

**Cherry-pick to `6.1.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
@@ -595,14 +595,16 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 	private Message<?> sendAndReceiveWithObservation(MessageChannel requestChannel, Object object,
 			Message<?> requestMessage) {
 
+		MessageRequestReplyReceiverContext context =
+				new MessageRequestReplyReceiverContext(requestMessage, getComponentName());
+
 		return IntegrationObservation.GATEWAY.observation(this.observationConvention,
 						DefaultMessageRequestReplyReceiverObservationConvention.INSTANCE,
-						() -> new MessageRequestReplyReceiverContext(requestMessage, getComponentName()),
-						this.observationRegistry)
-				.<MessageRequestReplyReceiverContext, Message<?>>observeWithContext((ctx) -> {
+						() -> context, this.observationRegistry)
+				.observe(() -> {
 					Message<?> replyMessage = doSendAndReceive(requestChannel, object, requestMessage);
 					if (replyMessage != null) {
-						ctx.setResponse(replyMessage);
+						context.setResponse(replyMessage);
 					}
 					return replyMessage;
 				});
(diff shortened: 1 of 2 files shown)


artembilan:
This is just a workaround for the current Micrometer version.
The change might be rejected if the real fix in the Micrometer will make it into the next release train in a couple weeks.

garyrussell (commented):
Just one comment.

garyrussell (approved):

garyrussell:
...and cherry-picked to 6.1.x as 41f41ed2ca921ea1ae6b9f655f7c2e9262761b8c

PR #8669, opened 2023-07-07
Fix Groovy DSL for new splitter style

artembilan:
* Introduce a new `GroovyIntegrationFlowDefinition.splitWith()` to avoid clashing with existing method
* Deprecate all other `split()` variants, but left one without arguments for a default splitting options
* Fix doc for new splitter style

### DIFF
--- a/src/reference/asciidoc/groovy-dsl.adoc
+++ b/src/reference/asciidoc/groovy-dsl.adoc
@@ -89,8 +89,15 @@ functionFlow() {
 		            transformer { it.toUpperCase() }
                     expectedType String
                 }
-                split Message<?>, { it.payload }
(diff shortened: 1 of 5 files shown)


src/reference/asciidoc/groovy-dsl.adoc
```diff
@@ -89,8 +89,15 @@ functionFlow() {
...
-                split Object, { it }, { id 'splitterEndpoint' }
+                splitWith {
+				    expectedType Message<?>
+				    function { it.payload }
+				}
+				splitWith {
+				    expectedType Object
+				    id 'splitterEndpoint'
+				    function { it }
+				}
```
garyrussell, line 100: Is this indentation correct?

PR #8670, opened 2023-07-10
Fix ImapIdleChA for scheduling race condition

artembilan:
The IMAP IDLE is long-lived process and can be blocked waiting for any reply from the server.
This way it is not suited to be used in a `TaskScheduler` especially when it has only one thread in its pool in Spring Boot by default. Another concurrent scheduled task is exactly an `ImapMailReceiver.IdleCanceller`. With a single thread in a `TaskScheduler` pool it cannot be reached therefore we never cancel and IDLE task and cannot react to the connection loss properly

* Rework the `ImapIdleChannelAdapter` logic to use a regular `Executor` and `while()` loop with a `Thread.sleep()` when we lose connection
* Clean up the `ImapMailReceiverTests` from `TaskScheduler` not used anymore.
* Expose new `taskExecutor` option in the `ImapIdleChannelAdapterSpec` for Java DSL
* Enable `ImapMailReceiverTests.testIdleWithMessageMapping()` with an attempt to see if this fix covers an unclear problem exposed before

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapIdleChannelAdapter.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapIdleChannelAdapter.java
@@ -17,10 +17,9 @@
 package org.springframework.integration.mail;
 
 import java.io.Serial;
-import java.time.Instant;
 import java.util.List;
+import java.util.concurrent.Executor;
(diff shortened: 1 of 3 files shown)


artembilan:
For more context: https://stackoverflow.com/questions/76611354/imapidlechanneladapter-doesnt-reconnect-to-idle-after-timeout

PR #8686, opened 2023-07-21
GH-8685: Re-fetch group after setting condition

### LINKED ISSUE #8685: ForceReleaseProcessor will never be run when group condition is used and MongoDbMessageStore

6.1.0

**To Reproduce**
Use `MongoDbMessageStore` or `ConfigurableMongoDbMessageStore` as the `MessageGroupStore` implementation.
Create aggregator:
- Provide the `groupConditionSupplier` for the `AbstractCorrelatingMessageHandler`.
- Provide the `groupTimeoutExpression` for the `AbstractCorrelatingMessageHandler` with a positive value of 5-10 seconds.

**Describe the bug**

Let's take a look at this function from AbstractCorrelatingMessageHandler:
```java
	private void processForceRelease(Object groupId, long timestamp, long lastModified) {
		MessageGroup messageGroup = this.messageStore.getMessageGroup(groupId);
		if (messageGroup.getTimestamp() == timestamp && messageGroup.getLastModified() == lastModified) {
			this.forceReleaseProcessor.processMessageGroup(messageGroup);
		}
	}
``` 
To run the `forceReleaseProcessor` the `timestamp` and `lastModified` must be the same as the corresponding properties from the message group that was fetched at the beginning of this function.
The `timestamp` and `lastModified` parameters come from the message group that was fetched in the `processMessageForGroup` function.

The problem is that after fetching the message group in the `processMessageForGroup` we do a call to `setGroupConditionIfAny(message, messageGroup)`. And if we are using the group condition, this function will call `messageStore.setGroupCondition()` which will update the message group condition and its `lastModifiedTime` property in MongoDB implementation.
```java
public void setGroupCondition(Object groupId, String condition) {
	updateGroup(groupId, lastModifiedUpdate().set("condition", condition));
}
``` 

So we end up in a situation where the `lastModifiedTime` is newer in the database but the message group fetched in the `processMessageForGroup` has an old value. This means that the condition in the `forceReleaseProcessor` will never be true if we are using the group condition.

**Expected behavior**

The message group instance in the `processMessageForGroup` should be updated after a call to `setGroupConditionIfAny`.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8685

The `AbstractCorrelatingMessageHandler` updates the group metadata in DB not only for provided `condition`, but also a `lastModified` field. A subsequent scheduling for group timeout takes the `lastModified` to compare with the value in the store after re-fetching group in task. This does not reflect reality since adding `condition` modifies the data in DB, but in-memory state remains the same.

* Re-fetch a group from the store in the `AbstractCorrelatingMessageHandler.setGroupConditionIfAny()`.
* Verify expected behavior via new `ConfigurableMongoDbMessageGroupStoreTests.groupIsForceReleaseAfterTimeoutWhenGroupConditionIsSet()`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -566,7 +566,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 			this.logger.trace(() -> "Adding message to group [ " + messageGroupToLog + "]");
 			messageGroup = store(correlationKey, message);
 
-			setGroupConditionIfAny(message, messageGroup);
+			messageGroup = setGroupConditionIfAny(message, messageGroup);
 
 			if (this.releaseStrategy.canRelease(messageGroup)) {
 				Collection<Message<?>> completedMessages = null;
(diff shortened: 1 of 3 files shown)


garyrussell:
...and cherry-picked to all supported (resolved test conflicts in 5.5.x).

PR #8688, opened 2023-07-26
Connect build to ge.spring.io to benefit from deep build insights and faster builds

erichaagdev:
This pull request connects the build to the Gradle Enterprise instance at https://ge.spring.io/. This allows the Spring Integration project to benefit from deep build insights provided by build scans and faster build speeds for all contributors as a result of remote build caching. 

This Gradle Enterprise instance has all features and extensions enabled and is freely available for use by Spring Integration and all other Spring projects. On this Gradle Enterprise instance, Spring Integration will have access not only to all of the published build scans, but other aggregate data features such as:

- Dashboards to view all historical build scans, along with performance trends over time
- Build failure analytics for enhanced investigation and diagnosis of build failures
- Test failure analytics to better understand trends and causes around slow, failing, and flaky tests

[Spring Boot](https://ge.spring.io/scans?search.rootProjectNames=spring-boot-build), [Spring Framework](https://ge.spring.io/scans?search.rootProjectNames=spring), and [Spring Security](https://ge.spring.io/scans?search.rootProjectNames=spring-security) are already connected to https://ge.spring.io/ and are benefiting from these features. 

For these features to be made possible, appropriate access must be configured to publish build scans and to write to the remote build cache. To ensure a consistent experience for all Spring contributors, the [Spring Gradle Enterprise Conventions](https://github.com/spring-io/gradle-enterprise-conventions) plugin is used. Its documentation details how to configure credentials [to publish build scans](https://github.com/spring-io/gradle-enterprise-conventions#build-scan-publishing-credentials) and [for writing to the remote build cache](https://github.com/spring-io/gradle-enterprise-conventions#credentials).

Please let me know if there are any questions about the value of Gradle Enterprise or the changes in this pull request and I’d be happy to address them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 <img align="right" width="250" height="250" src="https://spring.io/img/projects/spring-integration.svg?v=2">
 
-Spring Integration [<img src="https://build.spring.io/plugins/servlet/wittified/build-status/INT-MAIN">](https://build.spring.io/browse/INT-MAIN) [![Join the chat at https://gitter.im/spring-projects/spring-integration](https://badges.gitter.im/spring-projects/spring-integration.svg)](https://gitter.im/spring-projects/spring-integration?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-==================
+# Spring Integration
+
+[<img src="https://build.spring.io/plugins/servlet/wittified/build-status/INT-MAIN">](https://build.spring.io/browse/INT-MAIN) [![Join the chat at https://gitter.im/spring-projects/spring-integration](https://badges.gitter.im/spring-projects/spring-integration.svg)](https://gitter.im/spring-projects/spring-integration?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![Revved up by Gradle Enterprise](https://img.shields.io/badge/Revved%20up%20by-Gradle%20Enterprise-06A0CE?logo=Gradle&labelColor=02303A)](https://ge.spring.io/scans?search.rootProjectNames=spring-integration)
 
 Extends the Spring programming model to support the well-known Enterprise Integration Patterns. 
 Spring Integration enables lightweight messaging within Spring-based applications and supports integration with external systems via declarative adapters. 
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,3 +1,8 @@
+plugins {
+	id 'com.gradle.enterprise' version '3.13.4'
+	id 'io.spring.ge.conventions' version '0.0.13'
+}
+
 rootProject.name = 'spring-integration'
 
 rootDir.eachDir { dir ->


artembilan:
Hi @erichaagdev !

thank you for contribution, but I have a couple questions:
1. According to the `gradle-enterprise-conventions` plugin docs it looks like we have all required keys and credentials in our environments where we build PRs (GH actions) and our CI plans (Bamboo). Only looks like I need to do `provisionGradleEnterpriseAccessKey` locally to expose respective ENV var.
2. How to trigger the task for this feature?

In Spring for Apache Kafka we also have this extra config: https://github.com/spring-projects/spring-kafka/blob/main/settings.gradle#L14-L18.
There is also this file https://github.com/spring-projects/spring-kafka/blob/main/gradle/build-scan-user-data.gradle. Not sure if that is used yet somehow...

Do we need to change anything for our [`burrunan/gradle-cache-action`](https://github.com/spring-projects/spring-integration/blob/main/.github/workflows/pr-build-workflow.yml) to make this working with GH actions?

erichaagdev:
Hi @artembilan! These are very good questions. See my answers below.

> According to the gradle-enterprise-conventions plugin docs it looks like we have all required keys and credentials in our environments where we build PRs (GH actions) and our CI plans (Bamboo).

Bamboo should already be configured with the appropriate access keys to publish build scans and push to the remote build cache. Assuming that the main branch is being built in Bamboo, upon merging this PR you will begin to see build scans for Spring Integration at https://ge.spring.io/scans?search.rootProjectNames=spring-integration and contributors will benefit from the remote build cache. 

GitHub Actions **_must not_** be configured with credentials since pull requests are built there, as is [stated in the plugin docs](https://github.com/spring-io/gradle-enterprise-conventions#credentials). This is an unfortunate caveat that does not have a solution at this time. 

> Only looks like I need to do provisionGradleEnterpriseAccessKey locally to expose respective ENV var.

Correct. All you need to do is run `./gradlew provisionGradleEnterpriseAccessKey` locally. This only has to be done once and does not have to be done again for any other projects.

> How to trigger the task for this feature?

There are no additional tasks that need to be invoked. Build scans are automatically published for all builds and tasks will not need to execute when a remote build cache entry exists. 

> In Spring for Apache Kafka we also have this extra config: https://github.com/spring-projects/spring-kafka/blob/main/settings.gradle#L14-L18.

Nice catch. That isn't explicitly required since the convention plugin [enables publishing always](https://github.com/spring-io/gradle-enterprise-conventions/blob/79c558ab9ba07ec1d072b9d62e199fae14c17db0/src/main/java/io/spring/ge/conventions/gradle/BuildScanConventions.java#L83).

> There is also this file https://github.com/spring-projects/spring-kafka/blob/main/gradle/build-scan-user-data.gradle. Not sure if that is used yet somehow...

That is also not required as the convention plugin [already adds custom tags, links, and values](https://github.com/spring-io/gradle-enterprise-conventions/blob/79c558ab9ba07ec1d072b9d62e199fae14c17db0/src/main/java/io/spring/ge/conventions/gradle/BuildScanConventions.java#L93-L170) out of the box. You can however add further customizations as you see fit. 

> Do we need to change anything for our [burrunan/gradle-cache-action](https://github.com/spring-projects/spring-integration/blob/main/.github/workflows/pr-build-workflow.yml) to make this working with GH actions?

You should not need to change anything to `burrunan/gradle-cache-action`.

artembilan:
Thanks for answers.
That clears a lot.

Re. Bamboo and local are fully OK.
Not sure, though, about GH actions yet.
I see we have all the required ENV vars at our org. Probably that is there for those projects which CI system is exactly GH actions.
How we may be sure that they are not involved when we build PRs? Just according to your recommendation and that plugin docs I don't want to have build scans triggered accidentally.

I'm merging this one as is since there is nothing else to be fixed.

I'll may come back to you when I see something suspicious afterwards.

Thank you!

erichaagdev:
> How we may be sure that they are not involved when we build PRs? Just according to your recommendation and that plugin docs I don't want to have build scans triggered accidentally.

Workflows triggered by a fork do not have access to secrets so there is no need to worry: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

garyrussell:
@erichaagdev It sounds like we should just remove https://github.com/spring-projects/spring-kafka/blob/main/gradle/build-scan-user-data.gradle (I think it predated us moving the project to the plugin.

Also, the code in `settings.gradle` seems to be ignored anyway (we have been publishing successful builds as well as failures), so it seems I should just remove that as well.

https://ge.spring.io/scans?search.relativeStartTime=P90D&search.rootProjectNames=spring-kafka-dist&search.timeZoneId=America/New_York

artembilan:
So, here is a CI build against `main` with this fix: https://build.spring.io/browse/INT-MAIN-767/log.

The report doesn't show anything: https://ge.spring.io/scans?search.rootProjectNames=spring-integration&search.timeZoneId=America/New_York#.

And I also don't see any extra Gradle tasks in the build logs 🤷 

What do we miss yet, please?

artembilan:
I see `spring-kafka` build has this info in logs:
```
24-Jul-2023 20:41:20 Publishing build scan...
24-Jul-2023 20:41:21 https://ge.spring.io/s/grkme3gq57l5s
```

garyrussell:
I see this in your log

`error	26-Jul-2023 19:26:38	Unable to publish artifact [gradle]: `

erichaagdev:
> It sounds like we should just remove https://github.com/spring-projects/spring-kafka/blob/main/gradle/build-scan-user-data.gradle (I think it predated us moving the project to the plugin).

Correct. I don't see where the script is even being applied to the build so the file can be safely deleted.

> Also, the code in settings.gradle seems to be ignored anyway (we have been publishing successful builds as well as failures), so it seems I should just remove that as well.

Yes, [the `publishIfAuthenticated` from the convention plugin is taking precedence here](https://github.com/spring-io/gradle-enterprise-conventions/blob/79c558ab9ba07ec1d072b9d62e199fae14c17db0/src/main/java/io/spring/ge/conventions/gradle/BuildScanConventions.java#L82C1-L82C1).

> What do we miss yet, please?

@artembilan can you confirm the following are true:

For remote build caching:

> Pushing to the remote cache requires authentication. The necessary credentials can be provided using the `GRADLE_ENTERPRISE_CACHE_USERNAME` and `GRADLE_ENTERPRISE_CACHE_PASSWORD` environment variables.
> 
> **Bamboo**
> 
> The username and password environment variables should be set using `${bamboo.ge.cache.username}` and `${bamboo.ge.cache.password}` respectively.

From: https://github.com/spring-io/gradle-enterprise-conventions#bamboo

And for build scan publishing:

> Publishing to [ge.spring.io](https://ge.spring.io/) requires authentication via an access key. When running on CI, the access key should be made available via the `GRADLE_ENTERPRISE_ACCESS_KEY` environment variable.
>
> **Bamboo**
>
> The environment variable should be set to `${bamboo.gradle_enterprise_secret_access_key}`.

From: https://github.com/spring-io/gradle-enterprise-conventions#bamboo-1

garyrussell:
>Correct. I don't see where the script is even being applied to the build so the file can be safely deleted.

Looks like I should have removed it with this commit: https://github.com/spring-projects/spring-kafka/commit/c0dbc6eec8073b44b4fbab1ed07c5456f505e19d

PR #8689, opened 2023-07-26
GH-3696: DeserializationEx support for KafkaMS

### LINKED ISSUE #3696: Support ErrorHandlingDeserializer in KafkaMessageSource
**Expected Behavior**

Similar logic to the listener container - send deserialization exceptions to the error channel.

**Current Behavior**

User has to manually move past the bad record by setting the offset.

**Context**

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/3696

* Add an internal logic into `KafkaMessageSource` to react properly for the `ErrorHandlingDeserializer` configuration and re-throw `DeserializationException`


artembilan:
In general just copy/pasted the logic from the `KafkaMessageListenerContainer`.
Not sure what could be done better though...

Thanks

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
```diff
@@ -257,11 +271,68 @@ public Collection<TopicPartition> getAssignedPartitions() {
...
+		this.checkNullKeyForExceptions =
+				this.consumerProperties.isCheckDeserExWhenKeyNull() ||
+						checkDeserializer(findDeserializerClass(props, kafkaConsumerProperties, false));
+		this.checkNullValueForExceptions =
+				this.consumerProperties.isCheckDeserExWhenValueNull() ||
+						checkDeserializer(findDeserializerClass(props, kafkaConsumerProperties, true));
+	}
+
+	@Nullable
+	private Object findDeserializerClass(Map<String, Object> props, Properties consumerOverrides, boolean isValue) {
```
garyrussell, line 296: I will move most of the duplicated code to `SK.ErrorHandlingUtils`; then you can use it from there.

artembilan: Sounds good!

garyrussell: https://github.com/spring-projects/spring-kafka/pull/2756

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageSourceIntegrationTests.java
```diff
@@ -122,4 +138,66 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 		template.destroy();
 	}
 
+	@Test
+	void deserializationErrorIsThrownFromSource() throws Exception {
+		Map<String, Object> consumerProps = KafkaTestUtils.consumerProps(brokers, "testErrorChannelSource", "false");
+		consumerProps.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
garyrussell, line 144: FYI, this is now the default in KTU.

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageSourceIntegrationTests.java
```diff
@@ -122,4 +138,66 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
+		source.start();
+
+		Map<String, Object> producerProps = KafkaTestUtils.producerProps(brokers);
+		DefaultKafkaProducerFactory<Object, Object> producerFactory = new DefaultKafkaProducerFactory<>(producerProps);
+		KafkaTemplate<Object, Object> template = new KafkaTemplate<>(producerFactory);
+
+		String testData = "test data";
+		template.send(TOPIC2, testData);
+
+		source.receive(); // Trigger Kafka Consumer creation and poll()
```
garyrussell, line 180: I am not entirely sure, but the record might be returned by this initial poll.

It would be safer to remove this and move the following assertion into the `untilAsserted() -> {...}` block.

garyrussell (changes requested):

garyrussell (approved):

PR #8695, opened 2023-08-04
Initial Antora docs-build branch

rwinch:


.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,50 @@
...
+      uses: spring-io/spring-doc-actions/rsync-antora-reference@v0.0.11
+      with:
+        docs-username: ${{ secrets.DOCS_USERNAME }}
+        docs-host: ${{ secrets.DOCS_HOST }}
+        docs-ssh-key: ${{ secrets.DOCS_SSH_KEY }}
+        docs-ssh-host-key: ${{ secrets.DOCS_SSH_HOST_KEY }}
+    - name: Bust Clouflare Cache
+      uses: spring-io/spring-doc-actions/bust-cloudflare-antora-cache@v0.0.11
+      with:
+        context-root: spring-graphql
```
artembilan, line 48: I think this is a copy/paste artifact and must be `spring-integration`.

artembilan (changes requested):
Hey, Rob!

Sorry, I'm not fully understand what is going on so I left a couple minor comments on lines and here is one more major question:

why do we need a dedicated branch for this stuff? Why `antora` plugin cannot be as a part of trunk?

.github/workflows/deploy-docs.yml
rwinch: Good catch! This is pushed to the branch and fixed in 2d447677bf7222224ee54e1f6ae0500adbb3362d

.idea/checkstyle-idea.xml
```diff
@@ -0,0 +1,15 @@
...
+    <option name="thirdPartyClasspath" />
+    <option name="activeLocationIds" />
+    <option name="locations">
+      <list>
+        <ConfigurationLocation id="bundled-sun-checks" type="BUNDLED" scope="All" description="Sun Checks">(bundled)</ConfigurationLocation>
+        <ConfigurationLocation id="bundled-google-checks" type="BUNDLED" scope="All" description="Google Checks">(bundled)</ConfigurationLocation>
+      </list>
+    </option>
+  </component>
+</project>
```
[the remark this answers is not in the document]
rwinch: I was using my standard .gitignore which allows for checkstyle to be added. I've removed this and added it to .gitignore.

rwinch:
It can be added as a part of the default branch, but it doesn't belong there. This is the same reason that gh-pages is a detached branch. The docs-build branch builds for every version of your documentation that is deployed. As you add tags and additional branches, this branch is in charge of using the content from the tags and branches and then building that with the latest information (eg what versions are available, which UI to use, etc).

PR #8696, opened 2023-08-04
Migrate to Antora Documentation

rwinch:


artembilan (changes requested):
Why `modules/ROOT`?
Why docs stuff cannot live in some reasonable clear for contributors directory ?

Plus pay attention that we, essentially, build now docs from `build` dir after copying src there plus some generated stuff for Micrometer docs. So, this `ROOT` migration makes it much complicated.

Thanks for understanding.

rwinch:
@artembilan  - Thanks for the feedback!

> Why modules/ROOT?
> Why docs stuff cannot live in some reasonable clear for contributors directory ?

I've moved everything to be relative to `src/reference/antora` to better align with how your Asciidoc files were setup. Does this help?

> Plus pay attention that we, essentially, build now docs from build dir after copying src there plus some generated stuff for Micrometer docs. So, this ROOT migration makes it much complicated.

Thanks for pointing this out. I have [added](https://github.com/rwinch/spring-integration/blob/fecface9059e282f73d78f282cb56b67e7242255/src/reference/antora/antora-playbook.yml#L9) and [configured](https://github.com/rwinch/spring-integration/blob/fecface9059e282f73d78f282cb56b67e7242255/src/reference/antora/antora.yml#L9) the [Antora Collector Extension](https://gitlab.com/antora/antora-collector-extension) to invoke the [generateAntoraResources](https://github.com/rwinch/spring-integration/blob/fecface9059e282f73d78f282cb56b67e7242255/gradle/docs.gradle#L32) task. This will ensure that any generated content (including the output of `filterMetricsDocsContent`) will be available to Antora. You can see the generated adoc files are successfully [included in the preview](https://rwinch.github.io/spring-integration/metrics.html#observability-metrics).

Does that help alleviate your concern?

artembilan (changes requested):
Well,  it still shows `modules/ROOT/pages`.
But I guess that's how Antora works.
We probably can live with this.
My only concern here that external contributors might have some inconvenience in searching our docs sources...

rwinch:
> Well, it still shows modules/ROOT/pages.

Yes. This is the standard Antora layout & changing it would be difficult and confusing to those that have used Antora.

PR #8702, opened 2023-08-15
Fix Issue of removeLockKey() Method of RedisLock

EddieChoCho:
### Issue
In the current code, an IllegalStateException might be thrown from the try block while invoking the `removeLockKeyInnerUnlink()` method, especially when caused by key expiration (resulting in unlinkResult == false).

This triggers the catch block, which incorrectly sets the unlinkAvailable flag to false, even if the Redis server supports the unlink operation.

As a consequence, the subsequent `removeLockKeyInnerDelete()` method is invoked when it should not be.

### Solution

To address this issue, a Boolean flag should be introduced to accurately represent the outcome of the key unlink operation:

If the flag is null, it signifies that an exception was thrown during the execution of `the removeLockKeyInnerUnlink()` method.
If the flag is true, it indicates successful unlocking of the lock key.
If the flag is false, it indicates that the lock key expired, leading to the throwing of the IllegalStateException.

By implementing this flag-based approach, the behavior can be more precisely controlled, ensuring that the flow is correctly handled and that exceptions do not lead to incorrect state changes.

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -476,13 +476,9 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 
 		private void removeLockKey() {
 			if (RedisLockRegistry.this.unlinkAvailable) {
+				Boolean unlinkResult = null;
 				try {
-					boolean unlinkResult = removeLockKeyInnerUnlink();
-					if (!unlinkResult) {
-						throw new IllegalStateException("Lock was released in the store due to expiration. " +
-								"The integrity of data protected by this lock may have been compromised.");
-					}
-					return;
+					unlinkResult = removeLockKeyInnerUnlink();
 				}
 				catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):

artembilan (changes requested):
Also, please, fix Checkstyle violations:
```
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java:495:33: 'if' is not followed by whitespace. [WhitespaceAfter]
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java:495:33: 'if' is not followed by whitespace. [WhitespaceAround]
```
You can be sure about that locally with this command:
```
gradlew :spring-integration-redis:check
```

EddieChoCho:
Thank you for your feedback and the helpful tip! I've just pushed another commit.

artembilan:
... and cherry-picked to `6.1.x` & `6.0.x`.

Thank you again!

Please, consider next time to add all of that useful info into a commit message.
The first commit message of the PR becomes a PR title and description.

PR #8706, opened 2023-08-17
GH-8704: Add global property for `defaultTimeout`

### LINKED ISSUE #8704: Provide properties to block threads by default (again)
**Expected Behavior**

Users of Spring Integration can configure the new timeout defaults (for details, see below) globally instead of sprinkling new boilerplate timeout code throughout the application.

Here's an example of how such a configuration could look in application.properties (or yaml).
- spring.integration.thread-blocking-timeout=-1s
- spring.integration.thread-blocking-timeout-enabled=false

**Current Behavior**

According to https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-6.0-to-6.1-Migration-Guide#do-not-block-by-default, the default timeout for various SI operations has been changed from disabled to 30s. This breaks multiple SI projects in our team/company.

**Context**

We have several Spring Integration projects which are negatively affected by this change. We currently have 2 possibilities to fix our code. 
- Either we look for every `gateway()` and every `scatterGather()` call and pre-emptively disable the timeout in all these places. Today we looked at an application with 28 gateway() calls. Disadvantage: Lots of boilerplate code -> not Spring-like.
- Or we take the time to analyse each `gateway()` and each `scatterGather()` call and only disable the timeout where we have decided the call could take more than 30s. Disadvantage: requires more time to do.

Here's an example of the kind of code I'm talking about.

```
.scatterGather(
  scatterer -> scatterer
    .recipientFlow(flow -> flow
      .gateway(flow -> flow
        .gateway(subflow, 
          spec -> spec.replyTimeout(-1L)),
        spec -> spec.replyTimeout(-1L))
  gatherer -> gatherer.doSomething(),
  spec -> spec.gatherTimeout(-1L)
)
```

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8704

The default timeout for requests and replies in the integration endpoints is 30 seconds to avoid indefinite blocking in threads. Sometime those 30 seconds is not enough.

* Introduce a `spring.integration.endpoints.defaultTimeout` global property to allow overriding all the timeouts to desired value. The negative number indicates an indefinite waiting time: similar to what was there before introducing 30 seconds by default

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
@@ -35,7 +35,6 @@ import org.springframework.integration.IntegrationPattern;
 import org.springframework.integration.IntegrationPatternType;
 import org.springframework.integration.MessageTimeoutException;
 import org.springframework.integration.channel.ReactiveStreamsSubscribableChannel;
-import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.core.MessagingTemplate;
 import org.springframework.integration.endpoint.AbstractEndpoint;
 import org.springframework.integration.endpoint.EventDrivenConsumer;
@@ -124,7 +123,9 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 
 	private String errorChannelName;
 
-	private long replyTimeout = IntegrationContextUtils.DEFAULT_TIMEOUT;
+	private boolean requestTimeoutSet;
+
+	private boolean replyTimeoutSet;
 
 	private InboundMessageMapper<Object> requestMapper = new DefaultRequestMapper();
 
@@ -167,8 +168,6 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 	public MessagingGatewaySupport(boolean errorOnTimeout) {
 		ConvertingMessagingTemplate template = new ConvertingMessagingTemplate();
 		template.setMessageConverter(this.messageConverter);
-		template.setSendTimeout(IntegrationContextUtils.DEFAULT_TIMEOUT);
-		template.setReceiveTimeout(this.replyTimeout);
 		this.messagingTemplate = template;
 		this.errorOnTimeout = errorOnTimeout;
 	}
@@ -252,6 +251,7 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 	 */
 	public void setRequestTimeout(long requestTimeout) {
 		this.messagingTemplate.setSendTimeout(requestTimeout);
+		this.requestTimeoutSet = true;
 	}
 
 	/**
@@ -260,8 +260,8 @@ public abstract class MessagingGatewaySupport extends AbstractEndpoint
 	 * @param replyTimeout the timeout value in milliseconds
 	 */
 	public void setReplyTimeout(long replyTimeout) {
-		this.replyTimeout = replyTimeout;
 		this.messagingTemplate.setReceiveTimeout(replyTimeout);
+		this.replyTimeoutSet = true;
 	}
 
 	/**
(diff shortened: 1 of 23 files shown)


spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -406,6 +406,13 @@ protected void onInit() {
 			}
 			this.messageConverter.setBeanFactory(beanFactory);
 		}
+		long endpointsDefaultTimeout = getIntegrationProperties().getEndpointsDefaultTimeout();
+		if (!this.requestTimeoutSet) {
+			this.messagingTemplate.setSendTimeout(endpointsDefaultTimeout);
```
garyrussell, line 411: Is this intentional? Previously, the send timeout still defaulted to -1. Based on the docs, I assume yes. It makes sense to apply it, but I wonder if two properties are needed?

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -217,6 +216,9 @@ protected void onInit() {
 		}
 		this.messagingTemplate.setDestinationResolver(getChannelResolver());
 		setAsyncIfCan();
+		if (!this.sendTimeoutSet) {
+			this.messagingTemplate.setSendTimeout(getIntegrationProperties().getEndpointsDefaultTimeout());
+		}
```
garyrussell, line 221: ditto.

src/reference/antora/modules/ROOT/pages/configuration/global-properties.adoc
```diff
@@ -51,11 +53,18 @@ See xref:scatter-gather.adoc#scatter-gather-error-handling[Error Handling] for m
 <9> A boolean flag to indicate that default global `errorChannel` must ignore dispatching errors and pass the message to the next handler.
 Since version 5.5.
 
+<10> A milliseconds for request and reply timeouts in endpoints.
```
garyrussell, line 56:
```suggestion
<10> The default number of milliseconds for request and reply timeouts in endpoints.
```

src/reference/antora/modules/ROOT/pages/configuration/global-properties.adoc
```diff
@@ -51,11 +53,18 @@ See xref:scatter-gather.adoc#scatter-gather-error-handling[Error Handling] for m
...
+Default value is 30 seconds to avoid indefinite blocking.
+Can be configured to a negative value to restore blocking behaviour in endpoints.
```
garyrussell, line 58:
```suggestion
Can be configured to a negative value to restore infinite blocking behavior in endpoints.
```

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -33,6 +33,9 @@ See xref:endpoint.adoc#endpoint-pollingconsumer[Polling Consumer] for more infor
 - Java, Groovy and Kotlin DSLs have now context-specific methods in the `IntegrationFlowDefinition` with a single `Consumer` argument to configure an endpoint and its handler with one builder and readable options.
 See, for example, `transformWith()`, `splitWith()` in xref:dsl.adoc#java-dsl[ Java DSL Chapter].
 
+- A new `spring.integration.endpoints.defaultTimeout` global property has been introduced to override a default 30 seconds timeout for all the endpoints in the application.
```
garyrussell, line 36:
```suggestion
- A new `spring.integration.endpoints.defaultTimeout` global property has been introduced to override the default 30 seconds timeout for all the endpoints in the application.
```

garyrussell (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
artembilan: It was like this in the ctor:
```
template.setSendTimeout(IntegrationContextUtils.DEFAULT_TIMEOUT);
```

Yes, it was `-1` before `6.1`.

Not sure what you are asking about two props, though...

garyrussell: I was just wondering if we need a default send timeout as well as a default reply timeout property; since they are two different things; the send timeout is, for example, if a bounded `QueueChannel` is full whereas the reply timeout is for a request/reply.

artembilan: I see what you mean, but from end-user waiting perspective it looks the same, therefore I believe just a single default property is enough.
You can always override per endpoint and per property if there is a different requirements for send and receive.

garyrussell (approved):

PR #8712, opened 2023-08-22
GH-8705 Expose errorOnTimeout on MessagingGateway

### LINKED ISSUE #8705: Send the gateway Timeout Exception into the errorChannel and not a TemporaryReplyChannel
**Expected Behavior**

Gateway timeout exceptions, like any other exceptions thrown in the flow, are send to the errorChannel.
Also a custom errorMessage handling is possible.

**Current Behavior**

A gateway timeout exception is not send to the errorChannel (like any other exceptions in the flow), but a TemporaryReplyChannel.
In this case a custom errorChannel exception-handling is not impossible.

**Context**
Because of the breaking-change in Version 6.1.0. (https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-6.0-to-6.1-Migration-Guide#do-not-block-by-default).

Code example, with this behavior:

```
 return IntegrationFlow
                .from(Http.inboundGateway(...))
                .channel("gatewayTimeoutTest")
                .get();

return IntegrationFlow.from("gatewayTimeoutTest")
                .enrichHeaders(Map.of(MessageHeaders.ERROR_CHANNEL, "MY_ERROR_CHANNEL"))
                .gateway(longRunningSubflow())
                .get();


return IntegrationFlow.from("MY_ERROR_CHANNEL")
                 // .custom-log/handle/etc.
                 .get(); 
```

Because a custom errorChannel was defined, I expect that all exceptions from the longRunningSubflow(), are passed in payload to 'MY_ERROR_CHANNEL' ().
It works fine, but in case the longRunningSubflow() needs more then 30 Seconds, a timeout exception is thrown, ignoring the 
errorChannel configuration, instead it is passed to a dynamically created TemporaryReplyChannel.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8705

an internal `MethodInvocationGateway` is a `MessagingGatewaySupport` extension with all the logic available there.
One of the option introduced in `5.2.2` to be able to throw a `MessageTimeoutException` instead of returning `null` when no reply received in time from downstream flow

* Expose an `errorOnTimeout` on the `@MessagingGateway` and `GatewayEndpointSpec`
* Propagate this option from a `GatewayProxyFactoryBean` down to its internal `MethodInvocationGateway` implementation
* Modify couple tests to react for `errorOnTimeout` set to `true`
* Document the feature

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/annotation/MessagingGateway.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/annotation/MessagingGateway.java
@@ -167,7 +167,7 @@ public @interface MessagingGateway {
 	boolean proxyDefaultMethods() default false;
 
 	/**
-	 * If errorOnTimeout is true, the null won't be returned as a result of a gateway method invocation.
+	 * If errorOnTimeout is true, null won't be returned as a result of a gateway method invocation when a timeout occurs.
 	 * Instead, a {@link org.springframework.integration.MessageTimeoutException} is thrown
 	 * or an error message is published to the error channel.
 	 * @since 6.2
--- a/spring-integration-core/src/main/java/org/springframework/integration/annotation/MessagingGateway.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/annotation/MessagingGateway.java
@@ -23,6 +23,7 @@ import java.lang.annotation.Target;
 
 import org.springframework.core.annotation.AliasFor;
 import org.springframework.integration.context.IntegrationContextUtils;
+import org.springframework.integration.gateway.MessagingGatewaySupport;
 
 /**
  * A stereotype annotation to provide an Integration Messaging Gateway Proxy
@@ -164,5 +165,14 @@ public @interface MessagingGateway {
 	 * @since 5.3
 	 */
 	boolean proxyDefaultMethods() default false;
+
+	/**
+	 * If errorOnTimeout is true, the null won't be returned as a result of a gateway method invocation.
+	 * Instead, a {@link org.springframework.integration.MessageTimeoutException} is thrown
+	 * or an error message is published to the error channel.
+	 * @since 6.2
+	 * @see MessagingGatewaySupport#setErrorOnTimeout(boolean)
+	 */
+	boolean errorOnTimeout() default false;
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayEndpointSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayEndpointSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -95,6 +95,18 @@ public class GatewayEndpointSpec extends ConsumerEndpointSpec<GatewayEndpointSpe
 	 */
 	public GatewayEndpointSpec replyTimeout(Long replyTimeout) {
 		this.handler.setReplyTimeout(replyTimeout);
+		return this;
+	}
+
+	/**
+	 * Set a error on timeout flag.
+	 * @param errorOnTimeout true to produce an error in case on reply timeout.
+	 * @return the spec
+	 * @since 6.2
+	 * @see org.springframework.integration.gateway.GatewayProxyFactoryBean#setErrorOnTimeout(boolean)
+	 */
+	public GatewayEndpointSpec errorOnTimeout(boolean errorOnTimeout) {
+		this.handler.setErrorOnTimeout(errorOnTimeout);
 		return this;
 	}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayEndpointSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayEndpointSpec.java
@@ -100,8 +100,8 @@ public class GatewayEndpointSpec extends ConsumerEndpointSpec<GatewayEndpointSpe
 
 	/**
 	 * Set a error on timeout flag.
-	 * @param errorOnTimeout true to produce an error in case on reply timeout.
-	 * @return the spec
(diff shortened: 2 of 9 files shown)


spring-integration-core/src/main/java/org/springframework/integration/annotation/MessagingGateway.java
```diff
@@ -165,4 +166,13 @@
 	 */
 	boolean proxyDefaultMethods() default false;
 
+	/**
+	 * If errorOnTimeout is true, the null won't be returned as a result of a gateway method invocation.
```
garyrussell, line 170:
```suggestion
	 * If errorOnTimeout is true, null won't be returned as a result of a gateway method invocation when a timeout occurs.
```

spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayEndpointSpec.java
```diff
@@ -98,4 +98,16 @@ public GatewayEndpointSpec replyTimeout(Long replyTimeout) {
 		return this;
 	}
 
+	/**
+	 * Set a error on timeout flag.
+	 * @param errorOnTimeout true to produce an error in case on reply timeout.
```
garyrussell, line 103:
```suggestion
	 * @param errorOnTimeout true to produce an error in case of a reply timeout.
```

spring-integration-core/src/main/java/org/springframework/integration/dsl/GatewayEndpointSpec.java
```diff
@@ -98,4 +98,16 @@ public GatewayEndpointSpec replyTimeout(Long replyTimeout) {
...
+	 * @return the spec
```
garyrussell, line 104:
```suggestion
	 * @return the spec.
```

spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayProxyFactoryBean.java
```diff
@@ -455,6 +457,18 @@ public void registerMetricsCaptor(MetricsCaptor metricsCaptorToRegister) {
 		this.gatewayMap.values().forEach(gw -> gw.registerMetricsCaptor(metricsCaptorToRegister));
 	}
 
+	/**
+	 * If errorOnTimeout is true, the null won't be returned as a result of a gateway method invocation.
```
garyrussell, line 461:
```suggestion
	 * If errorOnTimeout is true, null won't be returned as a result of a gateway method invocation, when a timeout occurs.
```

src/reference/antora/modules/ROOT/pages/gateway.adoc
```diff
@@ -794,6 +794,8 @@ If a component downstream is still running (perhaps because of an infinite loop
 However, if the timeout has been reached before the actual reply was produced, it could result in a 'null' return from the gateway method.
 You should understand that the reply message (if produced) is sent to a reply channel after the gateway method invocation might have returned, so you must be aware of that and design your flow with it in mind.
 
+Also see an `errorOnTimeout` option to throw a `MessageTimeoutException` instead of returning `null`.
```
garyrussell, line 797:
```suggestion
Also see the `errorOnTimeout` property to throw a `MessageTimeoutException` instead of returning `null`, when a timeout occurs.
```

src/reference/antora/modules/ROOT/pages/gateway.adoc
```diff
@@ -838,4 +840,8 @@ At that time, the calling thread starts waiting for the reply.
 If the flow was completely synchronous, the reply is immediately available.
 For asynchronous flows, the thread waits for up to this time.
 
+Starting with version 6.2, the `errorOnTimeout` option of an internal `MethodInvocationGateway` extension of the `MessagingGatewaySupport` is exposed to the `@MessagingGateway` and `GatewayEndpointSpec`.
```
garyrussell, line 843:
```suggestion
Starting with version 6.2, the `errorOnTimeout` property of the internal `MethodInvocationGateway` extension of the `MessagingGatewaySupport` is exposed on the `@MessagingGateway` and `GatewayEndpointSpec`.
```

src/reference/antora/modules/ROOT/pages/gateway.adoc
```diff
@@ -838,4 +840,8 @@ At that time, the calling thread starts waiting for the reply.
...
+This option has exactly the same meaning as for any inbound gateway explained in the end of xref:endpoint-summary.adoc#endpoint-summary[Endpoint Summary] chapter.
+In other words, setting this option to `true`, would lead to the `MessageTimeoutException` being thrown from a send-and-receive gateway operation instead of returning `null` when receive timeout is exhausted.
```
garyrussell, line 845:
```suggestion
In other words, setting this option to `true`, would lead to the `MessageTimeoutException` being thrown from a send-and-receive gateway operation instead of returning `null` when the receive timeout is exhausted.
```

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -33,6 +33,9 @@ See xref:endpoint.adoc#endpoint-pollingconsumer[Polling Consumer] for more infor
 - Java, Groovy and Kotlin DSLs have now context-specific methods in the `IntegrationFlowDefinition` with a single `Consumer` argument to configure an endpoint and its handler with one builder and readable options.
 See, for example, `transformWith()`, `splitWith()` in xref:dsl.adoc#java-dsl[ Java DSL Chapter].
 
+- The `@MessagingGateway` and `GatewayEndpointSpec` for Java DSL expose now an `errorOnTimeout` option of an internal `MethodInvocationGateway` extension of the `MessagingGatewaySupport`.
```
garyrussell, line 36:
```suggestion
- The `@MessagingGateway` and `GatewayEndpointSpec` provided by the Java DSL now expose the `errorOnTimeout` property of the internal `MethodInvocationGateway` extension of the `MessagingGatewaySupport`.
```

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
Please rebase and feel free to merge.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4e310bbad62e09578f9b7cecc8510a914ebed3e2 after fixing conflicts.

Issue #8692, opened 2023-07-31
MongoDbChannelMessageStore ignores Auto Index Creation property


sdobberstein:
**In what version(s) of Spring Integration are you seeing this issue?**

6.1.1

**Describe the bug**

When starting up the Application the MongoDbChannelMessageStore ignores the `spring.data.mongodb.auto-index-creation` property and goes ahead and automatically creates the indexes.   Looking at the code for `MongoDbChannelMessageStore` and `AbstractConfigurableMongoDbMessageStore` it looks like they both create the indexes in the `afterPropertiesSet()` method without any regard to that property.

The reason it is an issue for us is that we are trying to use it with AWS DocumentDB.   In that case the auto-generated name for some of the indexes ends up being too long resulting in the following error:
```
{
    "ok": 0,
    "code": 67,
    "errmsg": "namespace name generated from index name is too long",
    "operationTime": {
        "$timestamp": {
            "t": 1690812708,
            "i": 1
        }
    }
}
```

**To Reproduce**

1.  When setting up your Spring Integration configuration, create a channel that is backed by a MongoDbChannelMessageStore.
```
@Bean
public MongoDbChannelMessageStore mongoDbChannelMessageStore(MongoDatabaseFactory databaseFactory)
{
	return new MongoDbChannelMessageStore(databaseFactory);
}

@Bean
public MessageChannel myExampleChannel(MongoDbChannelMessageStore mongoDbChannelMessageStore)
{
	MessageGroupQueue messageGroupQueue = new MessageGroupQueue(mongoDbChannelMessageStore, "myExample");
	return new QueueChannel(messageGroupQueue);
}
```
2.  Turn off auto-index creation.
```
spring.data.mongodb.auto-index-creation=false
```
3.  Start up your Application, then check the indexes on the collection.
```
db.channelMessages.getIndexes();
```
4.  You will see that the indexes have been created.

**Expected behavior**

If we turn off the `spring.data.mongodb.auto-index-creation` property, it should be applied consistently across the project.

**Sample**

Example repository:
https://github.com/sdobberstein/spring-integration-mongodb-issue

artembilan:
Spring integration cannot use those Spring Boot properties, but we really can expose a `createIndex(boolean)` option on those components, which, in the end, can be customized after Spring Boot.

Feel free to contribute such a simple solution: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc !

AdamaSorho:
[...]

PR #8716, opened 2023-08-28
GH-8692 implementing MongoDbChannelMessageStore ignores auto index creation

AdamaSorho:
#8692 

- added `createIndex` field, setCreateIndex and getCreateIndex methods to `MongoDbChannelMessageStore` class
- updated `afterPropertiesSet` method of `MongoDbChannelMessageStore`
- Updating `testWithCustomConverter` of `ConfigurableMongoDbMessageGroupStoreTests`

### DIFF
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2022 the original author or authors.
+ * Copyright 2014-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -60,6 +60,7 @@ import org.springframework.util.Assert;
  * for implementations of this class.
  *
  * @author Artem Bilan
+ * @author Adama Sorho
  *
  * @since 4.0
  */
@@ -86,6 +87,8 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 
 	private MessageBuilderFactory messageBuilderFactory = new DefaultMessageBuilderFactory();
 
+	private boolean createIndex = true;
+
 	public AbstractConfigurableMongoDbMessageStore(MongoTemplate mongoTemplate, String collectionName) {
 		Assert.notNull(mongoTemplate, "'mongoTemplate' must not be null");
 		Assert.hasText(collectionName, "'collectionName' must not be empty");
@@ -105,6 +108,15 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 		this.collectionName = collectionName;
 		this.mongoDbFactory = mongoDbFactory;
 		this.mappingMongoConverter = mappingMongoConverter;
+	}
+
+	/**
+	 * Define the option to auto create indexes or not.
+	 * @param createIndex a boolean.
+	 * @since 6.0.8.
+	 */
+	public void setCreateIndex(boolean createIndex) {
+		this.createIndex = createIndex;
 	}
 
 	@Override
@@ -128,6 +140,10 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 		return this.messageBuilderFactory;
 	}
 
+	protected boolean getCreateIndex() {
+		return this.createIndex;
+	}
+
 	@Override
 	public void afterPropertiesSet() {
 		if (this.mongoTemplate == null) {
@@ -145,7 +161,9 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 		}
 
 		this.messageBuilderFactory = IntegrationUtils.getMessageBuilderFactory(this.applicationContext);
+	}
 
+	protected void createIndexes() {
 		IndexOperations indexOperations = this.mongoTemplate.indexOps(this.collectionName);
 
 		indexOperations.ensureIndex(new Index(MessageDocumentFields.MESSAGE_ID, Sort.Direction.ASC));
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
@@ -87,7 +87,7 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 
 	private MessageBuilderFactory messageBuilderFactory = new DefaultMessageBuilderFactory();
 
-	private boolean createIndex = true;
+	private boolean createIndexes = true;
 
 	public AbstractConfigurableMongoDbMessageStore(MongoTemplate mongoTemplate, String collectionName) {
 		Assert.notNull(mongoTemplate, "'mongoTemplate' must not be null");
@@ -112,11 +112,11 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 
 	/**
 	 * Define the option to auto create indexes or not.
-	 * @param createIndex a boolean.
+	 * @param createIndexes a boolean.
 	 * @since 6.0.8.
 	 */
-	public void setCreateIndex(boolean createIndex) {
-		this.createIndex = createIndex;
+	public void setCreateIndexes(boolean createIndexes) {
+		this.createIndexes = createIndexes;
 	}
 
 	@Override
@@ -140,10 +140,6 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 		return this.messageBuilderFactory;
 	}
 
-	protected boolean getCreateIndex() {
-		return this.createIndex;
-	}
-
 	@Override
 	public void afterPropertiesSet() {
 		if (this.mongoTemplate == null) {
@@ -162,8 +158,8 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 
 		this.messageBuilderFactory = IntegrationUtils.getMessageBuilderFactory(this.applicationContext);
 
-		if (getCreateIndex()) {
-			this.createIndexes();
+		if (this.createIndexes) {
+			createIndexes();
 		}
 	}
 
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
@@ -161,6 +161,10 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 		}
 
 		this.messageBuilderFactory = IntegrationUtils.getMessageBuilderFactory(this.applicationContext);
+
+		if (getCreateIndex()) {
+			this.createIndexes();
+		}
 	}
 
 	protected void createIndexes() {
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2020 the original author or authors.
+ * Copyright 2014-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -43,6 +43,7 @@ import org.springframework.util.Assert;
  * {@code priorityEnabled = true} option.
  *
  * @author Artem Bilan
+ * @author Adama Sorho
  *
  * @since 4.0
  */
@@ -55,6 +56,8 @@ public class MongoDbChannelMessageStore extends AbstractConfigurableMongoDbMessa
 	public static final String DEFAULT_COLLECTION_NAME = "channelMessages";
 
 	private boolean priorityEnabled;
+
+	private boolean createIndex = true;
 
 	public MongoDbChannelMessageStore(MongoTemplate mongoTemplate) {
 		this(mongoTemplate, DEFAULT_COLLECTION_NAME);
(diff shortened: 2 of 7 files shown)


spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
```diff
@@ -96,12 +99,22 @@ public boolean isPriorityEnabled() {
...
+			getMongoTemplate()
+					.indexOps(this.collectionName)
+					.ensureIndex(new Index(MessageDocumentFields.GROUP_ID, Sort.Direction.ASC)
+							.on(MessageDocumentFields.PRIORITY, Sort.Direction.DESC)
+							.on(MessageDocumentFields.LAST_MODIFIED_TIME, Sort.Direction.ASC)
+							.on(MessageDocumentFields.SEQUENCE, Sort.Direction.ASC));
+		}
+	}
+
+	public void setCreateIndex(boolean createIndex) {
```
artembilan, line 112: This is a `public` API, therefore it needs a Javadoc including `@since 6.0.8`.
Right, I know, we have many missed Javadocs in other places, but let's don't introduce more!

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/store/ConfigurableMongoDbMessageGroupStoreTests.java
```diff
@@ -134,6 +138,11 @@ void testWithCustomConverter() {
 						.getClass());
 		context.refresh();
 
+		MongoTemplate mongoTemplate = new MongoTemplate(MONGO_DATABASE_FACTORY);
+		List<IndexInfo> indexesInfo = mongoTemplate.indexOps(MongoDbChannelMessageStore.DEFAULT_COLLECTION_NAME).getIndexInfo();
+
+		assertThat(indexesInfo.size()).isEqualTo(4);
```
artembilan, line 144: See my comment in the issue about this problem: https://github.com/spring-projects/spring-integration/issues/8692#issuecomment-1696156878.
So, apparently we need to remove indexes on that collection before creating an application context in this test.
Or we can just remove collection altogether.

There is also a nice fluent API in AssertJ for collections:
```
assertThat(indexesInfo).hasSize(0);
```

artembilan (changes requested):

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
```diff
@@ -96,12 +99,23 @@ public boolean isPriorityEnabled() {
 	@Override
 	public void afterPropertiesSet() {
 		super.afterPropertiesSet();
-		getMongoTemplate()
-				.indexOps(this.collectionName)
-				.ensureIndex(new Index(MessageDocumentFields.GROUP_ID, Sort.Direction.ASC)
-						.on(MessageDocumentFields.PRIORITY, Sort.Direction.DESC)
-						.on(MessageDocumentFields.LAST_MODIFIED_TIME, Sort.Direction.ASC)
-						.on(MessageDocumentFields.SEQUENCE, Sort.Direction.ASC));
+		if (this.createIndex) {
```
AdamaSorho, line 102: Is this implementation, okay? The parent `AbstractConfigurableMongoDbMessageStore` still creates indexes. Should I check in the parent class or is it okay for the parent class to create indexes even if the createIndex is set to false?

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/store/ConfigurableMongoDbMessageStore-CustomConverter.xml
```diff
@@ -29,6 +29,7 @@
 
 	<beans:bean id="messageStore" parent="abstractMessageStore">
 		<beans:constructor-arg name="mappingMongoConverter" ref="customConverter"/>
+		<beans:property name="createIndex" value="false" />
```
AdamaSorho, line 32: When I add the property, the createIndex value is still true after the application context creation. However, adding to the `abstractMessageStore` set the value to corresponding value in this case to false.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
artembilan: Ok. I missed the fact that super class does create indexes as well. Thank for the pointer !
Then we need to push this property to super class and then we need to have a protected `createIndexes()` method which can be overridden in impels for their own indexes.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/store/ConfigurableMongoDbMessageStore-CustomConverter.xml
artembilan: That’s because a super class creates its own indexes and we don’t check for this property there. See our discussion in that class

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
```diff
@@ -96,6 +97,14 @@ public boolean isPriorityEnabled() {
 	@Override
 	public void afterPropertiesSet() {
 		super.afterPropertiesSet();
+		if (this.getCreateIndex()) {
```
artembilan, line 100: This check and new method call has to be done in a super class. Probably this `afterPropertiesSet()` can be removed from here altogether then.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/store/ConfigurableMongoDbMessageStore-CustomConverter.xml
```diff
@@ -17,6 +17,7 @@
 	<beans:bean id="abstractMessageStore" class="org.springframework.integration.mongodb.store.MongoDbChannelMessageStore"
 				abstract="true">
 		<beans:constructor-arg name="mongoDbFactory" ref="mongoDbFactory"/>
+		<beans:property name="createIndex" value="false" />
```
artembilan, line 20: Doesn’t this still work if we move it to concrete extension of this bean?

artembilan (changes requested):
Take a look, please, to all the impls of that abstract class if they need to implement our new protected method.

when you done with that, please, consider to mention such a new feature in the `whats-new.adoc`, as well as in the `mongodb.adoc`

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
AdamaSorho: In the previous implement `MongoDbChannelMessageStore` subclass creates an index as well. Here is the index creates by the subclass `MongoDbChannelMessageStore`: `getMongoTemplate()
				.indexOps(this.collectionName)
				.ensureIndex(new Index(MessageDocumentFields.GROUP_ID, Sort.Direction.ASC)
						.on(MessageDocumentFields.PRIORITY, Sort.Direction.DESC)
						.on(MessageDocumentFields.LAST_MODIFIED_TIME, Sort.Direction.ASC)
						.on(MessageDocumentFields.SEQUENCE, Sort.Direction.ASC));`

So, I suppose this class must create an index after the super class created its indexes. It's Why I overrode the `createIndexes()` in the subclass and check in that `afterPropertiesSet()` which will call the `createIndexes()` in the subclass. The `createIndex()` will then call its super `createIndex()`. You can see the implementation of `createIndexes()` in `MongoDbChannelMessageStore`.

If I remove `afterPropertiesSet()` from subclass, I don't know how we will create the subclass `MongoDbChannelMessageStore` index.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/store/ConfigurableMongoDbMessageStore-CustomConverter.xml
AdamaSorho: Yes! I don't know why.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
```diff
@@ -107,6 +110,15 @@ public AbstractConfigurableMongoDbMessageStore(MongoDatabaseFactory mongoDbFacto
 		this.mappingMongoConverter = mappingMongoConverter;
 	}
 
+	/**
+	 * Define the option to auto create indexes or not.
+	 * @param createIndex a boolean.
+	 * @since 6.0.8.
```
AdamaSorho, line 116: I am wondering if this is the good version. The issue was found in the `version 6.1.1`.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbChannelMessageStore.java
artembilan: That is called `polymorphism` in Java: when super class calls the method, the actual instance is called. So, if we override `createIndexes()` in the sub-class then exactly this one is called. Therefore no need to call it from sub-class if we call it from super class.
And since we move all the logic to that `createIndexes()` extension, there is nothing left in its `afterPropertiesSet()`, therefore we don't need it any more.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
artembilan: The issue is still apply for an older version as well.
Since it is in support, we still have to back port it.
Therefore we need to use as lower version for `@since` as we back port the fix.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
```diff
@@ -145,7 +161,9 @@ public void afterPropertiesSet() {
 		}
 
 		this.messageBuilderFactory = IntegrationUtils.getMessageBuilderFactory(this.applicationContext);
+	}
 
+	protected void createIndexes() {
```
artembilan, line 166: This method has to be called from this `afterPropertiesSet()`. And thanks to polymorphism an overridden  method is going to be called. Therefore we don't need to check for `createIndex` in the target implementation of this class.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/store/ConfigurableMongoDbMessageStore-CustomConverter.xml
artembilan: OK. I'll look into this when you are done with other my concerns in the production classes.

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/mongodb.adoc
```diff
@@ -231,6 +231,22 @@ To configure that scenario, you can extend one message store bean from the other
     <int:priority-queue message-store="priorityStore"/>
 </int:channel>
 ----
+[[abstract-configurable-mongodb-message-store-with-auto-index-creation-disable]]
+=== Using AbstractConfigurableMongoDbMessageStore with auto index creation disable
```
artembilan, line 235: I don't think we need a dedicated section just for this feature.
Please, consider to move the content into the end of the `mongodb-message-store` section.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -38,6 +38,10 @@ See xref:configuration/global-properties.adoc[Global Properties] for more inform
 
 - The `@MessagingGateway` and `GatewayEndpointSpec` provided by the Java DSL now expose the `errorOnTimeout` property of the internal `MethodInvocationGateway` extension of the `MessagingGatewaySupport`.
 See xref:gateway.adoc#gateway-no-response[ Gateway Behavior When No response Arrives] for more information.
+
+- A new method `setCreateIndex(boolean)` has been introduced in `org.springframework.integration.mongodb.store.AbstractConfigurableMongoDbMessageStore` to disable or enable (default) the auto index creation.
```
artembilan, line 42: This is a MongoDB support change, so it deserves its own `[[x6.2-mongodb]]` section in the end of this doc. Similar to `[[x6.2-jdbc]]` over there.

Also: we don't need to use the package for the class name in the docs.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
```diff
@@ -128,6 +140,10 @@ protected MessageBuilderFactory getMessageBuilderFactory() {
 		return this.messageBuilderFactory;
 	}
 
+	protected boolean getCreateIndex() {
```
artembilan, line 143: No it turns out that we don't need this `protected` getter altogether.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
```diff
@@ -86,6 +87,8 @@ public abstract class AbstractConfigurableMongoDbMessageStore extends AbstractMe
 
 	private MessageBuilderFactory messageBuilderFactory = new DefaultMessageBuilderFactory();
 
+	private boolean createIndex = true;
```
artembilan, line 90: Didn't we agree that property name should be a `createIndexes` since we really might create several of them, not just one as it is assumed by this singular current name?

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
```diff
@@ -146,6 +162,12 @@ public void afterPropertiesSet() {
 
 		this.messageBuilderFactory = IntegrationUtils.getMessageBuilderFactory(this.applicationContext);
 
+		if (getCreateIndex()) {
+			this.createIndexes();
```
artembilan, line 166: No `this.` prefix on method calls.

artembilan (changes requested):

artembilan (changes requested):
```
Conflicting files
src/reference/antora/modules/ROOT/pages/whats-new.adoc 
```
Please, consider to switch to `main` branch, perform `git pull upstream main`.
Then switch back to your branch and perform `git rebase main`.
It is probably not going to rebase clean - you will have to fix conflicts manually.

artembilan (changes requested):
Please, perform `git rebase main`.

It look like your PR has now unwanted commits.

artembilan (approved):
However I see that all the changes are in, so I'm pulling this locally for final review and some clean up.

If you don't have anything else to add, we go from here with a merge.

Thank you!

AdamaSorho:
I don't have anything to add. 

If merged, I would like to work on a new issue if you have any suggestions.

Thank you for your help and time

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/store/ConfigurableMongoDbMessageStore-CustomConverter.xml
```diff
@@ -17,6 +17,7 @@
 	<beans:bean id="abstractMessageStore" class="org.springframework.integration.mongodb.store.MongoDbChannelMessageStore"
 				abstract="true">
 		<beans:constructor-arg name="mongoDbFactory" ref="mongoDbFactory"/>
+		<beans:property name="createIndexes" value="false" />
```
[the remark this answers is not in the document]
AdamaSorho: Ok, I see now. Perfect!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/32eba4ecb9537d064311025397e955aad47897c4. Pay attention to the final changes. Especially the test environment.

Back-ported to `6.1.x` & `6.0.x` as https://github.com/spring-projects/spring-integration/commit/ba82efd1caa093012a83432383b432ab6b484f30 after removing docs changes since in the `main` we handle docs via new model for Antora: https://docs.spring.io/spring-integration/reference/index.html.

I'll ping you from an issue when I find one for you to make another contribution.

AdamaSorho:
Ok, Thank you!

Issue #8711, opened 2023-08-21
Use SKIP LOCKED in MySqlChannelMessageStoreQueryProvider and PostgresChannelMessageStoreQueryProvider


ckalan:
Mysql 8 supports SKIP LOCKED and Postgres also supports it in its recent versions.

Allow query providers of these databases to either conditionally support SKIP LOCKED or introduce DB version specific variants that support it.

**Current Behavior**

Currently, `OracleChannelMessageStoreQueryProvider` appends SKIP LOCKED clause to its queries but `MySqlChannelMessageStoreQueryProvider` and `PostgresChannelMessageStoreQueryProvider` don't. 

**Context**
I am planning to use Mysql or Postgres as a persistent queue in a HA cluster, but based on my understanding, without SKIP LOCKED, multiple pollers in different JVMs may receive the same message from the same JDBC backed queue simultaneously. `PostgresChannelMessageStoreQueryProvider` uses FOR UPDATE clause which I guess prevents retrieval of same message at the same time but not in an ideal way. 

Currently, developers can write their own query provider as a workaround which I plan to but I am not fully aware of the consequences. Based on Mysql and Postgres documentations, the behavior looks similar to Oracle's but I am not a DB expert and there may be other developers like me who may share the same hesitation. 

Even if it is not possible to use SKIP LOCKED as I suggest here for any reason, this ticket would probably be a guide to those who have the similar question

artembilan:
I have scheduled this for the next milestone, however no one stops you to extend that `PostgresChannelMessageStoreQueryProvider` and override respective methods to add that `SKIP LOCKED` in the end of query:
```
public class SkipLokcedPostgresChannelMessageStoreQueryProvider extends PostgresChannelMessageStoreQueryProvider {

	@Override
	public String getPollFromGroupExcludeIdsQuery() {
		return super.getPollFromGroupExcludeIdsQuery() + " SKIP LOCKED";
	}

	@Override
	public String getPollFromGroupQuery() {
		return super.getPollFromGroupQuery() + " SKIP LOCKED";
	}

	@Override
	public String getPriorityPollFromGroupExcludeIdsQuery() {
		return super.getPriorityPollFromGroupExcludeIdsQuery() + " SKIP LOCKED";
	}

	@Override
	public String getPriorityPollFromGroupQuery() {
		return super.getPriorityPollFromGroupQuery() + " SKIP LOCKED";
	}
	
}
```

artembilan:
[...]

PR #8719, opened 2023-08-30
GH-8711: Allowing SKIP LOCKED in MySqlChannelMessageStoreQueryProvider and PostgresChannelMessageStoreQueryProvider

AdamaSorho:
#8711 

- Added `useSkipLocked` property, `setUseSkipLocked(boolean)` and `getUseSkipLocked()` methods in `AbstractChannelMessageStoreQueryProvider`. I added `setUseSkipLocked(boolean)` in case there are some MySql and PostgresSql version which not support `SKIP LOCKED`. Let me know if we don't need it.
- Added `addSkipLocked(query)` and updated all methods in `PostgresChannelMessageStoreQueryProvider` and `MySqlChannelMessageStoreQueryProvider` classes.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/AbstractChannelMessageStoreQueryProvider.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/AbstractChannelMessageStoreQueryProvider.java
@@ -20,26 +20,9 @@ package org.springframework.integration.jdbc.store.channel;
 /**
  * @author Gunnar Hillert
  * @author Artem Bilan
- * @author Adama Sorho
  * @since 2.2
  */
 public abstract class AbstractChannelMessageStoreQueryProvider implements ChannelMessageStoreQueryProvider {
-
-	private boolean useSkipLocked = false;
-
-	/**
-	 * Enable the use of `SKIP LOCKED`.
-	 * Be sure your database support this functionality
-	 * @param useSkipLocked a boolean
-	 * @since 6.0.8
-	 */
-	public void setUseSkipLocked(boolean useSkipLocked) {
-		this.useSkipLocked = useSkipLocked;
-	}
-
-	protected boolean getUseSkipLocked() {
-		return this.useSkipLocked;
-	}
 
 	public String getCountAllMessagesInGroupQuery() {
 		return "SELECT COUNT(MESSAGE_ID) from %PREFIX%CHANNEL_MESSAGE where GROUP_KEY=? and REGION=?";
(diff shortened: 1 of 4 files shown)


artembilan (changes requested):
Thank you for quick turn around! 
I don’t think we need that configuration option. This type of classes are very simple to override . Plus it is really recommended to use the latest DBs. 

Now, imaging there are other hints. Should we expose props for them as well?  I just don’t see a reason to overcomplicate the logic when plain string can be easily used.

AdamaSorho:
The build has failed for `:spring-integration-file:test`, but I can't see the error. I've tried to test locally with `./gradlew clean :spring-integration-file:check`, I'm getting an error. See attached screenshot. 
testOS spring-integration-file test case failed

Did I miss something to test `:spring-integration-file`?

Note: `spring-integration-jdbc` tests passed successfully.

artembilan:
I re-ran your PR build because that test has noting to do with your changes and it really fails sporadically, especially on Linux where file system events are a bit slower to be emitted.

artembilan (changes requested):
Now it fails like this:
```
 PostgresChannelMessageTableSubscriberTests > testMessagesDispatchedInTransaction() FAILED
    org.opentest4j.AssertionFailedError at PostgresChannelMessageTableSubscriberTests.java:189
```
And this is probably related to the changes in your PR...

artembilan (changes requested):
It looks like we need to call `postgresChannelMessageTableSubscriber.stop()` in that test just after `assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();`.
This way that `PostgresChannelMessageTableSubscriber` won't poll DB any more making those records we try to verify next locked.

artembilan (approved):
Merging after some clean up locally.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/571ed20905059c2c52f22e74e305f69320e244ad.

Forgot to mention before, although I believe we have that in the [Contribution Guidelines](https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc):

1. First commit message becomes a PR description
2. The head line of the first commit must be not longer than 50 symbols
3. The next line must be a link to the issue with a `Fixed` keyword. This way the issue on GitHub going to be closed automatically.
4. The rest of text in this an subsequent commit messages must describe not just what has been changes, but also why.

See that Chris Beam's article in the end: https://cbea.ms/git-commit/

If you would like to contribute more, I have this pretty straightforward task for you.
Deprecate that `AbstractChannelMessageStoreQueryProvider` in favor of `default` method impls in the `ChannelMessageStoreQueryProvider` interface.
And remove its usage from those impls since the interface will carry on all the required default behavior.
Not need in the issue - just open PR and we will go from there.

Thank you for your great effort!

AdamaSorho:
Okay, thank you for these notes.

I'll open a PR.

PR #8721, opened 2023-08-31
Deprecate AbstractChannelMessageStoreQueryProvider

AdamaSorho:
Deprecate that `AbstractChannelMessageStoreQueryProvider` in favor of default method impls in the `ChannelMessageStoreQueryProvider` interface.

Take advantage of default interface methods.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/AbstractChannelMessageStoreQueryProvider.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/AbstractChannelMessageStoreQueryProvider.java
@@ -20,35 +20,11 @@ package org.springframework.integration.jdbc.store.channel;
 /**
  * @author Gunnar Hillert
  * @author Artem Bilan
+ * @author Adama Sorho
  *
  * @since 2.2
+ * @deprecated in favor of default methods in ChannelMessageStoreQueryProvider
  */
 @Deprecated
 public abstract class AbstractChannelMessageStoreQueryProvider implements ChannelMessageStoreQueryProvider {
-
-	public String getCountAllMessagesInGroupQuery() {
-		return "SELECT COUNT(MESSAGE_ID) from %PREFIX%CHANNEL_MESSAGE where GROUP_KEY=? and REGION=?";
-	}
-
-	public String getMessageQuery() {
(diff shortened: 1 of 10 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/AbstractChannelMessageStoreQueryProvider.java
```diff
@@ -23,6 +23,7 @@
  *
  * @since 2.2
  */
+@Deprecated
```
artembilan, line 26: Please, also add a `@deprecated in favor of default methods in ChannelMessageStoreQueryProvider` into a JavaDoc of this class after `@since`

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/AbstractChannelMessageStoreQueryProvider.java
```diff
@@ -23,6 +23,7 @@
...
 public abstract class AbstractChannelMessageStoreQueryProvider implements ChannelMessageStoreQueryProvider {
 
 	public String getCountAllMessagesInGroupQuery() {
```
artembilan, line 29: Since all methods have been moved to the interface, there is no need to keep them in this class any more.

artembilan (changes requested):
Please, also find this deprecated class in the `message-store.adoc` and fix respective sentences to use an interface instead.

artembilan:
```
PostgresChannelMessageTableSubscriberTests > testMessagesDispatchedInTransaction() FAILED
    org.opentest4j.AssertionFailedError at PostgresChannelMessageTableSubscriberTests.java:193
```
So, there is still a race condition between transaction in the `PostgresSubscribableChannel` and our queries in the this test.
I think I see what is going on and that is indeed a bug.
We have there this condition `while (dispatchedMessage.isPresent())` which is never false if we rollback transaction in this loop.
Therefore even if we stop a publisher it doesn't mean that we stop this loop.
And even worth: when we fully unsubscribed from this channel, the loop doesn't stop when TX is rolled back.

I'm going to open PR with a fix: nothing to do for your one.
This will be merged shortly.

Thank you!

artembilan:
Here is the fix if you are interested: https://github.com/spring-projects/spring-integration/pull/8723

AdamaSorho:
Ok perfect! Thank you

artembilan:
See also a Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-6.1-to-6.2-Migration-Guide#deprecation-of-the-abstractchannelmessagestorequeryprovider

PR #8725, opened 2023-09-02
GH-8691: Support age FTP, SMB over time filters

### LINKED ISSUE #8691: Support age (s/ftp) and size over time filters (both file and s/ftp)
**Expected Behavior**

Enhance available file filtering strategies out-of-the-box.

**Current Behavior**

Filtering files based on a last-modified strategy isn't currently possible out-of-the-box.
There are publically-available [possible implementations](https://stackoverflow.com/questions/64628653/lastmodifiedfilelistfilter-for-sftp-inbound-adapter) to consider.
Same deal with watching file size over time.  

**Context**

Such functionality is desirable especially for the numerous cases where one doesn't control the producer side and thus can't use a marker file.
The size-based strategy is, as always, a guard against slow producers and, in the case of SFTP, extreme edge case network slowdowns.
I would expect the size-based strategy to make use of a metadata store, probably.

AdamaSorho:
fixes #8691 

Add `LastModifiedFTPFileListFilter` and `LastModifiedSmbFileListFilter` respectively for FTP and SMB to enhance available files filtering based on a last-modified strategy.


AdamaSorho:
I ran the tests before starting to work on the issue, but I am getting an error for `sftp`. So, I did not touch this module. Let me know if I must configure something to pass this test. I will add age support filtering to `sftp` module once I pass the tests. 

See screenshot below:
SFTP test error

artembilan:
Thank you for reporting that @AdamaSorho !

Please, consider to add `testRuntimeOnly 'net.i2p.crypto:eddsa:0.3.0'` into `spring-integration-sftp` module. See `build.gradle`.
As part of your PR is OK. Turned out you are not only one who faced this problem for non-default `known_hosts` file.

Also: please, stop posting screenshots for stacktraces: it would much easier for me to copy/paste info for search 😄

AdamaSorho:
Understood, sorry about that.

The tests passed now for `sftp` module. Could you tell me why we need `net.i2p.crypto:eddsa:0.3.0`? I don't know understand 🙈

Thank you

artembilan:
We need that library to be able to parse that Ed25519 encrypted key in your known_hosts file: https://github.com/str4d/ed25519-java

AdamaSorho:
Ok I see. Thank you

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/LastModifiedSmbFileListFilter.java
```diff
@@ -0,0 +1,139 @@
...
+		}
+		else if (this.discardCallback != null) {
+			this.discardCallback.accept(file);
+		}
+
+		return false;
+	}
+
+	private boolean fileIsAged(SmbFile file, long now) {
+		return file.getLastModified() / ONE_SECOND + this.age <= now;
```
artembilan, line 132: This SMB API does not provide an `Instant` representation, so, we have to stick with millis.
However those blank lines around members an removal of `volatile` are still valid concerns.

spring-integration-smb/src/test/java/org/springframework/integration/smb/filters/LastModifiedSmbFileListFilterTests.java
```diff
@@ -0,0 +1,76 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Adama Sorho
+ * @since 6.2
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class LastModifiedSmbFileListFilterTests extends SmbTestSupport {
```
artembilan, line 42: And this one is more drastic than others: it starts a Docker container for Samba server.
Therefore I'd prefer to have all these new tests against mocks and fully in-memory.

artembilan (changes requested):

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/filters/LastModifiedFTPFileListFilter.java
```diff
@@ -0,0 +1,139 @@
...
+ * <p>
+ *     The resolution is done in seconds.
+ * </p>
+ * When {@link #discardCallback} is provided, it called for all the rejected files.
+ *
+ * @author Adama Sorho
+ *
+ * @since 6.2
+ */
+public class LastModifiedFTPFileListFilter implements DiscardAwareFileListFilter<FTPFile> {
```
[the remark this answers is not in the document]
AdamaSorho: Only this should be renamed or the other classes too?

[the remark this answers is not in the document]
artembilan: Right. Others, too. Just follow same pattern for consistency.

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/filters/FtpLastModifiedFileListFilter.java
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * Set the age that the files have to be before being passed by this filter.
+	 * If {@link FTPFile#getTimestampInstant()} plus {@link #age} is before the {@link Instant#now()}, the file
+	 * is filtered.
+	 * Defaults to 60 seconds.
+	 * @param age the age in seconds.
+	 * @param unit the timeUnit.
+	 */
+	public void setAge(long age, TimeUnit unit) {
```
artembilan, line 79: I don't think we need a `TimeUnit`-based API since we have a `Duration`-one. It really covers any possible time units.

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/filters/FtpLastModifiedFileListFilterTests.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @author Adama Sorho
+ * @since 6.2
+ */
+public class FtpLastModifiedFileListFilterTests {
+
+	@Test
+	public void testAge() {
+		FtpLastModifiedFileListFilter filter = new FtpLastModifiedFileListFilter();
+		filter.setAge(60, TimeUnit.SECONDS);
```
artembilan, line 36: Why do we need to set this value if it is exactly what we have by default?

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/filters/FtpLastModifiedFileListFilter.java
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright 2015-2023 the original author or authors.
```
artembilan, line 2: This class was really created in the `2023`. Therefore copyright is just for the current year.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/filters/SftpLastModifiedFileListFilter.java
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright 2013-2023 the original author or authors.
```
artembilan, line 2: DITTO

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/filters/SftpLastModifiedFileListFilter.java
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * Set the age that the files have to be before being passed by this filter.
+	 * If {@link FileTime#toInstant()} plus {@link #age} is before the {@link Instant#now()}, the file
+	 * is filtered.
+	 * Defaults to 60 seconds.
+	 * @param age the age in seconds.
+	 * @param unit the timeUnit.
+	 */
+	public void setAge(long age, TimeUnit unit) {
```
artembilan, line 75: DITTO

artembilan (changes requested):
This is very close to the merge.

Now, let's consider to mentioned these new filters in the respective  FTP, SFTP and SMB chapters of the docs.
The change could make it into the `ftp/inbound.adoc`, `sftp/inbound.adoc`, and `smb.adoc`, respectively.

The `whats-new.adoc` also must have a new simple section like `=== Remote Files Support Changes` and mention all new filter there.

Thank you!

artembilan (changes requested):
Looks like something is wrong with your Git workflow.

1. Switch to `main`
2. Perform `git pull upstream main`
3. Switch back to your branch
4. Perform `git rebase main`
5. `git push origin -f`

Otherwise your PR contains changes you didn't do. Plus it suggests them as new changes somehow...

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbLastModifiedFileListFilter.java
```diff
@@ -0,0 +1,142 @@
...
+
+	/**
+	 * Set the age that the files have to be before being passed by this filter.
+	 * If {@link SmbFile#getLastModified()} plus age is greater than the current time, the file
+	 * is filtered. The resolution is seconds.
+	 * Defaults to 60 seconds.
+	 * @param age the age in seconds.
+	 * @param unit the timeUnit.
+	 */
+	public void setAge(long age, TimeUnit unit) {
```
artembilan, line 74: I believe this one is still has to be remove: the `Duration` does the trick for us anyway.

artembilan (approved):
OK. Just one nit-pick.
Will take care about that on merge.

Forgot to mention: when we `force push` to the PR, GitHub doesn't notify us about changes.
So, consider in the future to leave some comment that it ready for review.

Thanks

src/reference/antora/modules/ROOT/pages/ftp/inbound.adoc
```diff
@@ -98,6 +98,23 @@ You should also understand that the FTP inbound channel adapter is a polling con
...
+The age defaults to 60 seconds, but you should choose an age that is large enough to avoid picking up a file early (due to, say, network glitches).
+The following example shows how to configure a `FtpLastModifiedFileListFilter`:
+
+[source, java]
+----
+@Bean
+public FtpLastModifiedFileListFilter ftpLastModifiedFileListFilter() {
+	FtpLastModifiedFileListFilter filter = new FtpLastModifiedFileListFilter();
+
+	filter.setAge(Duration.ofSeconds(180));
```
artembilan, line 112: We have ctor for this. Therefore this sample would end up with a single line.
More over I think we don't this examples at all: the API of the class is pretty straightforward and there is a good Javadoc.
We may say though, please, look into its Javadoc for more info.

artembilan (changes requested):
OK. After looking into this closely, I suggest this evolution.

How about extracting an `AstractLastModifiedFileListFilter<F>` with all the common logic and only leave something like `abstract Instant getLastModified(F remoteFile);` for the implementors?
I believe this way even that existing `LastModifiedFileListFilter` could benefit from new super class.

Yes, some of the implementors would need to use `Instant.ofEpochSecond()` to convert their `long` to expected format by super class.

AdamaSorho:
Yeah, that will be perfect!

spring-integration-file/src/main/java/org/springframework/integration/file/filters/AbstractLastModifiedFileListFilter.java
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	@Override
+	public boolean supportsSingleFileFiltering() {
+		return true;
+	}
+
+	protected abstract Instant getLastModified(F remoteFile);
+
+	protected Duration getCurrentAge() {
```
AdamaSorho, line 121: I added this method because we have a `long getAge()` in `LastModifiedFileListFilter` for backward compatibility.

I think we should deprecate `long getAge()`.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/LastModifiedFileListFilter.java
```diff
@@ -74,74 +66,16 @@ public LastModifiedFileListFilter(long age) {
...
-	 * is filtered. The resolution is seconds.
-	 * Defaults to 60 seconds.
-	 * @param age the age
-	 */
-	public void setAge(long age) {
-		setAge(age, TimeUnit.SECONDS);
+		setAge(unit.toSeconds(age));
 	}
 
 	public long getAge() {
```
AdamaSorho, line 72: We should deprecate this.

artembilan (changes requested):
This looks very cool!
Just a couple minor comments.

You may not be agreed with me on some of them, so just comment with your arguments.
Otherwise I believe it can be merged after those fixes.

spring-integration-file/src/main/java/org/springframework/integration/file/filters/LastModifiedFileListFilter.java
```diff
@@ -74,74 +66,16 @@ public LastModifiedFileListFilter(long age) {
 	 * @param unit the timeUnit.
 	 */
 	public void setAge(long age, TimeUnit unit) {
```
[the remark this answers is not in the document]
AdamaSorho: Okay! I think we can remove `..(long age)` constructors and deprecate `LastModifiedFileListFilter(long age)` in favor of `Duration` constructors as well.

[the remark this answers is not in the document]
artembilan: No, `long` ctor, as well as setter, are good . They are pretty simple convenient. I just don’t see a reason in a `TimeUnit`-based.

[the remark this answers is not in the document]
AdamaSorho: Okay

artembilan (approved):
Pulling locally for the final review, clean up and merge...

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/73ed3eeebd5b919df852bc484af9a40dc0281307 after some clean up.

Thank you again for contributions!

AdamaSorho:
Thank you!

Don't hesitate to ping me on any issue at anytime.

PR #8726, opened 2023-09-05
Bump Debezium to 2.3.3.Final

tzolov:


build.gradle
```diff
@@ -62,7 +62,7 @@ ext {
     commonsIoVersion = '2.13.0'
     commonsNetVersion = '3.9.0'
     curatorVersion = '5.5.0'
-    debeziumVersion = '2.3.0.Final'
+    debeziumVersion = '2.3.3.Final'
```
artembilan, line 65: I don't see this version in Maven Central: https://mvnrepository.com/artifact/io.debezium/debezium-embedded

What is their release plan?
Probably we can move on to `2.4` in our `main` and it is is going to be GA for our GA in November?

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/outbox/DefaultOutboxEventSender.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * {@link OutboxEventSender} interface implementation that converts the {@link ExportEvent} into {@link OutboxEvent}
+ * entity and persists it in the outbox table via the {@link OutboxEventRepository} repository.
+ *
+ * @author Christian Tzolov
+ *
+ * @since 6.2
+ */
+public class DefaultOutboxEventSender implements OutboxEventSender {
```
artembilan, line 30: I don't remember that we have agreed to apply these changes.
Perhaps you have messed up some branches, so unexpected files has slipped to your current proposal?..

artembilan (changes requested):

PR #8729, opened 2023-09-08
GH-2971: Add `LockRegistry.executeLocked()` API

### LINKED ISSUE #2971: add a `LockTemplate` to simplify working with distributed Locks in Spring Integration
Please add a `LockTemplate` or something to make using the `LockRegistry` easier. It took a lot of attempts before I figured out the right combination of objects was. I've encapsulated it into a `LockTemplate`, as shown below, but I'm sure might see use cases I haven't yet addressed here.  

``` java

import org.springframework.integration.support.locks.LockRegistry;
import org.springframework.util.ReflectionUtils;

import java.util.concurrent.Callable;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.locks.Lock;

public class LockTemplate {

	private final LockRegistry registry;

	public LockTemplate(LockRegistry registry) {
		this.registry = registry;
	}

	public <T> T executeWithLock(String key, int timeoutDuration, TimeUnit tu, Callable<T> callable) {
		return this.doExecuteWithLock(key, lock -> lock.tryLock(timeoutDuration, tu), callable);
	}

	public <T> T executeWithLock(String key, Callable<T> callable) {
		return this.doExecuteWithLock(key, Lock::tryLock, callable);
	}

	private <T> T doExecuteWithLock(
		String key, ExceptionSwallowingFunction<Lock, Boolean> lockProducer, Callable<T> callable) {

		try {
			Lock lock = registry.obtain(key);
			boolean lockAcquired = lockProducer.apply(lock);
			if (lockAcquired) {
				try {
					return callable.call();
				}
				finally {
					lock.unlock();
				}
			}
		}
		catch (Exception e) {
			ReflectionUtils.rethrowRuntimeException(e);
		}
		return null;
	}

	private interface ExceptionSwallowingFunction<I, O> {
		O apply(I i) throws Exception;
	}
}
```

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/2971

* Following best practice and well-known patterns with `Jdbc`, `Rest` or `Jms` templates, introduce `default` methods into `LockRegistry` interface to make it easier to perform tasks when within a lock.
* Since all the required logic is now covered by those `LockRegistry.executeLocked()` methods, there is no need in the dedicated abstract `WhileLockedProcessor` class. Deprecated it for removal in the next version
* Use a new `LockRegistry.executeLocked()` API in the `FileWritingMessageHandler` instead of just deprecated `WhileLockedProcessor`
* To satisfy Java limitations for checked lambdas, introduce `CheckedCallable` and `CheckedRunnable` utilities similar to interfaces in the `io.micrometer.observation.Observation`
* Change existing `CheckedFunction` to expose extra generic argument for `Throwable`
* Add dedicated chapter for distributed lock into docs
* Fix some links and typos in the docs

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/locks/LockRegistry.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/locks/LockRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,13 +16,20 @@
 
 package org.springframework.integration.support.locks;
 
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.locks.Lock;
+
+import org.springframework.integration.util.CheckedCallable;
+import org.springframework.integration.util.CheckedRunnable;
 
 /**
  * Strategy for maintaining a registry of shared locks.
  *
  * @author Oleg Zhurakousky
  * @author Gary Russell
+ * @author Artem Bilan
  *
  * @since 2.1.1
  */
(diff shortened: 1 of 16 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/locks/LockRegistry.java
```diff
@@ -16,24 +16,119 @@
...
+		try {
+			return callable.call();
+		}
+		finally {
+			lock.unlock();
+		}
+	}
+
+	/**
+	 * Perform a provided task when lock for the key is locked.
```
garyrussell, line 88:
```suggestion
	 * Perform the provided task when the lock for the key is locked.
```

spring-integration-core/src/main/java/org/springframework/integration/support/locks/LockRegistry.java
```diff
@@ -16,24 +16,119 @@
...
+
+		executeLocked(lockKey, waitLockDuration,
+				() -> {
+					runnable.run();
+					return null;
+				});
+	}
+
+	/**
+	 * Perform a provided task when lock for the key is locked.
```
garyrussell, line 108:
```suggestion
	 * Perform the provided task when the lock for the key is locked.
```

spring-integration-core/src/main/java/org/springframework/integration/util/CheckedRunnable.java
```diff
@@ -0,0 +1,52 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.util;
+
+/**
+ * A Runnable-like interface which allows throwing Error.
```
garyrussell, line 20:
```suggestion
 * A Runnable-like interface which allows throwing any Throwable.
 * Checked exceptions are wrapped in an IllegalStateException.
```

src/reference/antora/modules/ROOT/pages/distributed-locks.adoc
```diff
@@ -0,0 +1,37 @@
+[[distributed-locks]]
+= Distributed Locks
+
+In many situations the action against some context (or even single message) has to be performed in an exclusive manner.
+One of the example is an aggregator component where we have to check the message group state for the current message to device whether we can release group or just add that message for future consideration.
+For this purpose Java provides a good API with `java.util.concurrent.locks.Lock` implementations.
+However, the problem becomes more complex when our system is distributed and/or run in the cluster.
+The locking in this case is challenging and requires some shared and its specific approach to achieve an exclusivity requirement.
```
garyrussell, line 8:
```suggestion
The locking in this case is challenging and requires some shared state and its specific approach to achieve the exclusivity requirement.
```

garyrussell (changes requested):

garyrussell (approved):

Issue #8730, opened 2023-09-10
Improve performance of "obtainLock" blocking during "expireUnusedOlderThan" runtime of "RedisLockRegistry"


mhyeon-lee:
**In what version(s) of Spring Integration are you seeing this issue?**

6.0.7.RELEASE

**Describe the bug**

When running "expireUnusedOlderThan" on "RedisLockRegistry", too many target lock objects are holding locks for too long.
The "obtain" operation on another thread will have a longer blocking time.

The Redis call is made at "lock.isAcquiredInThisProcess" in the link below, so if RedisLockRegistry has a lot of lock objects, this can take a long time.

https://github.com/spring-projects/spring-integration/blob/863c525790444204f34ec7846002a5512c036d66/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L252

**To Reproduce**

Steps to reproduce the behavior.

**Expected behavior**

To avoid holding a lock for too long, the scopes that hold the lock should be broken up into individual objects.

artembilan:
Fixed via: https://github.com/spring-projects/spring-integration/commit/50c53e08aacc84f64e9d6266d331e9bc1fa32891

PR #8731, opened 2023-09-10
GH-8730: split lock scope each lock objects in expireUnusedOlderThan to avoid holding a lock for too long

mhyeon-lee:
PR for: #8730

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -55,6 +55,7 @@ import org.springframework.util.Assert;
  * @author Fran Aranda
  * @author Unseok Kim
  * @author Christian Tzolov
+ * @author Myeonghyeon Lee
  *
  * @since 4.3
  */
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Thank you for reporting this and looking into possible fix!

I think that `isAcquiredInThisProcess()` impl is wrong:
```
		public final boolean isAcquiredInThisProcess() {
			return RedisLockRegistry.this.clientId.equals(
					RedisLockRegistry.this.redisTemplate.boundValueOps(this.lockKey).get());
		}
```
It's contract says "in this process". Therefore we must not look into Redis for a lock state, but rather just check `this.localLock.isLocked()`. 

The operation `expireUnusedOlderThan()` we try to fix is really just about cache for locks in the current process. We cannot effect any other processes with this cache. And even if we remove our local instance while it is locked in other process, that doesn't mean that on the next `obtain()` call, when we got a fresh local instance, we will be able to acquire a lock in Redis.

WDYT?

artembilan:
Yeah... Looks like that is really a correct way to go.
This functionality was modeled after `ZookeeperLockRegistry`, where it does this:
```
		public boolean isAcquiredInThisProcess() {
			return this.mutex.isAcquiredInThisProcess();
		}
```
And that one is:
```

    /**
     * Returns true if the mutex is acquired by a thread in this JVM
     *
     * @return true/false
     */
    @Override
    public boolean isAcquiredInThisProcess()
```

So, please, fix this `RedisLockRegistry` to check a local lock in its `isAcquiredInThisProcess()`.

Same, please, do for the `JdbcLockRegistry`.

mhyeon-lee:
Ok I'll try fix it.
Thank you for your review

flex-myeonghyeon:
@artembilan I fixed it with localLock

artembilan (changes requested):
Then I'll merge.
Thanks

flex-myeonghyeon:
@artembilan I added @author 
Thank you!

artembilan:
... and cherry-picked to `6.1.x` & `6.0.x`.

PR #8733, opened 2023-09-12
GH-8732 Don't remove JDBC message if other groups

### LINKED ISSUE #8732: Messages deleted too early in JdbcMessageStore when using a message in multiple groups

6.1.2 

**Describe the bug**

When using a `JdbcMessageStore` and one group completes (or expires), the group is removed. This removal also removes the messages that were linked to that group, even when the same message is still linked to other groups.  
The link (`int_group_to_message`) for that other group is not removed, so there's a mapping entry to a message that no longer exists in the database. Also, the size attribute of the other group is not updated, so it still shows a count that suggests the message is still present, while it can't be retrieved via `getMessages()` or `streamMessages()`

**To Reproduce**

Set up two flows listening on different channels. Both flows have an aggregator and use the `JdbcMessageStore`. Define different release strategies for each flow, but make sure there is an overlap in the "required messages" for each flow.  
Now send the message that is required in both flows to both channels. After that, send a message that completes one of the groups. Then send a message that should complete the other group. It will not complete, since the first message is now missing from the group.

**Expected behavior**

The message should only be deleted from the `int_message` table, when it's not linked to any other group. 

**Sample**
https://github.com/nschoellhorn/spring-integration-jdbc-repro

In [SpringIntegrationJdbcReproApplication](https://github.com/nschoellhorn/spring-integration-jdbc-repro/blob/master/src/main/kotlin/de/kopf3/springintegrationjdbcrepro/SpringIntegrationJdbcReproApplication.kt#L27), I have implemented a CommandLineRunner that simulates the flow of events/messages as described above.  
The flow setup can be found in [Flows.kt](https://github.com/nschoellhorn/spring-integration-jdbc-repro/blob/master/src/main/kotlin/de/kopf3/springintegrationjdbcrepro/Flows.kt)

The Release strategy has some logging that should visualize the problem:
image

If my assumption is correct that this is a bug, I would be happy to assist in providing a patch to achieve the expected behavior.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8732

When same message is added into different groups,
its record in the `INT_MESSAGE` must remain until the last group is removed

* Improve `JdbcMessageStore.DELETE_MESSAGES_FROM_GROUP` SQL to ignore those messages for removal which has other group records in the `INT_GROUP_TO_MESSAGE` table

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/MessageStore.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/MessageStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -62,11 +62,12 @@ public interface MessageStore {
 	<T> Message<T> addMessage(Message<T> message);
 
 	/**
-	 * Remove the Message with the given id from the MessageStore, if present, and return it. If no Message with that id
-	 * is present in the store, this will return <i>null</i>.
-	 *
-	 * @param id THe message identifier.
-	 * @return The message.
+	 * Remove the Message with the given id from the MessageStore, if present, and return it.
+	 * If no Message with that id is present in the store, this will return {@code null}.
+	 * If this method is implemented on a {@link MessageGroupStore},
+	 * the message is removed from the store only if no groups holding this message.
+	 * @param id the message identifier.
+	 * @return the message (if any).
 	 */
 	Message<?> removeMessage(UUID id);
 
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
@@ -199,7 +199,12 @@ public class JdbcMessageStore extends AbstractMessageGroupStore
 
 		DELETE_MESSAGES_FROM_GROUP("""
 				DELETE from %PREFIX%MESSAGE
-				where MESSAGE_ID in (SELECT MESSAGE_ID from %PREFIX%GROUP_TO_MESSAGE where GROUP_KEY = ? and REGION = ?)
+				where MESSAGE_ID in (
+								SELECT MESSAGE_ID from %PREFIX%GROUP_TO_MESSAGE where GROUP_KEY = ? and REGION = ?
+												and MESSAGE_ID not in (
+																SELECT MESSAGE_ID from %PREFIX%GROUP_TO_MESSAGE
+																			where GROUP_KEY != ? and REGION = ?)
+								)
 					and REGION = ?
 				"""),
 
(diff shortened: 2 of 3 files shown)


garyrussell:
I think we'll need some javadocs on `removeMessage()`. It is an implementation of the `MessageStore` method, not a `MessageGroupStore` method.

Therefore, it will need to override the parent javadoc...

```java
	 * Remove the Message with the given id from the MessageStore, if present, and return it. If no Message with that id
	 * is present in the store, this will return <i>null</i>.
```
...to indicate that it will not be removed if the message is present in one or more groups.

garyrussell:
Or, update the parent javadocs to mention that if the implementation is on a `MessageGroupStore` then...

garyrussell:
Also, is this just a JDBC problem or can it happen with any MGS?

artembilan:
>  or can it happen with any MGS?

I thought about that, but let's see if we are correct about the fix for JDBC only and let's really decide how deep we would like to back-port it!

I believer we have a couple workarounds:
1. Use different MS instances with different `region`.
2. Don't send the same message to different aggregators with the same MS.

garyrussell:
Yes, I don't think it was anticipated that the same MGS would be used for different aggregators, and certainly not if those aggregators each get the same message ids.

However, I can see that with large groups (and/or messages), this would be a benefit, avoiding duplication.

garyrussell:
Perhaps a different issue, but we should consider optimizing `addMessage` too.

Currently, we detect (and ignore) duplicates after conversion and attempting to update. We might want to check if it already exists first.

artembilan:
So, I'm OK to have this as a fix only on `main` and `6.1.x`.
Or if you insist I can try to back-port down to `5.5.x` as we all agreed that this is a long-standing bug. 

However I still in doubts in a query for `DELETE_MESSAGES_FROM_GROUP`.

Other stores might be fixed in other PR to not make this too complicate.

The `addMessage()` is indeed a separate issue since it does not donate to the bug and could land just only in the `main` as an improvement.

garyrussell:
>However I still in doubts in a query ...

? 

Do you want it merged, or not?

artembilan:
Sure! Let's merge it!
But let's cherry-pick only to `6.1.x`!
Thanks

garyrussell:
...and cherry-picked as ccaa853d68

artembilan:
@garyrussell ,
regarding `addMessage()`.
I don't think that it is going to be a good optimization to perform one more `SELECT` before `INSERT` attempt: it is one more network hope, so not so optimal unlike plain `this.serializer.convert(message);` in memory.

artembilan:
Doesn't look like `ConfigurableMongoDbMessageStore` suffers from the same problem.
The messages are stored in the collection in in a linear way and every single is supplied with `groupId`.
So, removing one group does not effect others.
We might consider to improve plain `removeMessage()` which does not consult a presence (not) `groupId`...

artembilan:
Looks like an `AbstractKeyValueMessageStore` has that problem though.
We store plain message ids in the group and every message as a key.
When we remove the group, we just take those ids and iterate them for remove-by-key operation.
Perhaps those message ids in group have to be enhanced with the group id as well: there is no way to have a one-to-many in key-value stores.

PR #8735, opened 2023-09-14
Fix ThreadSPropagationChInterceptor for stacking

artembilan:
Related SO thread: https://stackoverflow.com/questions/77058188/multiple-threadstatepropagationchannelinterceptors-not-possible

The current `ThreadStatePropagationChannelInterceptor` logic is to wrap one message to another (`MessageWithThreadState`), essentially stacking contexts. The `postReceive()` logic is to unwrap a `MessageWithThreadState`, therefore we deal with the latest pushed context which leads to the `ClassCastException`

* Rework `ThreadStatePropagationChannelInterceptor` logic to reuse existing `MessageWithThreadState` and add the current context to its `stateQueue`.
Therefore, the `postReceive()` will `poll()` the oldest context which is, essentially, the one populated by this interceptor before, according to the interceptors order
* Fix `AbstractMessageChannel.setInterceptors()` to not modify provided list of interceptors
* The new `ThreadStatePropagationChannelInterceptorTests` demonstrates the problem described in that mentioned SO question and verifies that context are propagated in the order they have been populated

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.channel;
 
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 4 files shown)


artembilan:
OK. Thinking about this one more time, I convince myself that this wrapping approach is not so good. Especially when we send the same message to several handlers like in case of publish-subscribe channel.
And it looks like the header approach like I take for Spring Security fix is much easier and more robust: https://github.com/spring-projects/spring-security/pull/12532.

We may still have this `ThreadStatePropagationChannelInterceptor` but it should create a new message with the context populated into provided header.

Probably that should be done as a new feature just for current version since it is a bit breaking change and this PR still makes sense as bug fix for previous version.

garyrussell (approved):

garyrussell:
...and cherry-picked.

PR #8736, opened 2023-09-15
GH-8713: Support custom SftpClient

### LINKED ISSUE #8713: Not possible to add handlers for channels requests to DefaultSftpSessionFactory
**Expected Behavior**
It should be possible to add handlers for channel requests to DefaultSftpSessionFactory.

**Current Behavior**
It is not possible to add handlers for channels requests to DefaultSftpSessionFactory.

**Context**
See https://stackoverflow.com/questions/76906857/migration-to-spring-boot-3-how-can-i-handle-keepaliveopenssh-com-requests-f

AdamaSorho:
Fixes #8713 

Introduce `protected createSftpClient(...)` to support custom sftpClient. It can be overridden by the end-user

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
@@ -53,6 +53,11 @@ import org.springframework.util.Assert;
 
 /**
  * Factory for creating {@link SftpSession} instances.
+ * <p>
+ *     The {@link #createSftpClient(ClientSession, SftpVersionSelector, SftpErrorDataHandler)} can be overridden
+ *     to provide a custom {@link SftpClient}.
+ * </p>
+ * The {@link ConcurrentSftpClient} is used by default.
  *
  * @author Josh Long
  * @author Mario Gray
@@ -398,7 +403,21 @@ public class DefaultSftpSessionFactory implements SessionFactory<SftpClient.DirE
 		this.sharedSftpClient = null;
 	}
 
-	protected SftpClient createSftpClient(ClientSession clientSession, SftpVersionSelector initialVersionSelector, SftpErrorDataHandler errorDataHandler) throws IOException {
+	/**
+	 * This method allows the end-user to provide a custom {@link SftpClient}
+	 * to {@link #getSession()}.
+	 * @param clientSession the {@link ClientSession}
+	 * @param initialVersionSelector the initial {@link SftpVersionSelector}
+	 * @param errorDataHandler the {@link SftpErrorDataHandler} to handle incoming data
+	 *                         through the error stream.
+	 * @return {@link SftpClient}
+	 * @throws IOException if failed to initialize
+	 *
+	 * @since 6.2
+	 */
+	protected SftpClient createSftpClient(
+			ClientSession clientSession, SftpVersionSelector initialVersionSelector,
+			SftpErrorDataHandler errorDataHandler) throws IOException {
 
 		return new ConcurrentSftpClient(clientSession, initialVersionSelector, errorDataHandler);
 	}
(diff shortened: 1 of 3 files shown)


artembilan (changes requested):
This change also requires mentioning in the `session-factory.adoc`  to explain what this new method is about. Probably with a sample how to call `ChannelSubsystem.addRequestHandler()` in the custom `SftpClient`.

Of course, a new `[[x6.2-sftp-changes]]` is needed to be added to the `whats-new.adoc` as well.

I also mentioned before a convenient way what to do with the fist commit message in the PR: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc#provide-a-link-to-the-github-issue-in-the-associated-pull-request. What I mean that you just shape that commit message they way that you would not need to modify a PR description and that commit message will appear in the history with all the traceable information.

Thanks

AdamaSorho:
Oh  Sorry, I got it now.

Thanks

AdamaSorho:
I am trying to add a sample of how to call `ChannelSubsystem.addRequestHandler()` with this issue use case, but I am having some hard time getting it.  I don't know how to get an instance of `RequestHandler<Channel>` for `keepalive` request.

artembilan:
There is no out-of-the-box `RequestHandler<Channel>` instance for `keepalive` request.

So, our ample in the doc must be as simple as this:
```
		@Override
		protected ChannelSubsystem createSftpChannelSubsystem(ClientSession clientSession) {
			ChannelSubsystem sftpChannelSubsystem = super.createSftpChannelSubsystem(clientSession);
			sftpChannelSubsystem.addRequestHandler((channel, request, wantReply, buffer) -> ...);
			return sftpChannelSubsystem;
		}
```

and we say that custom `SftpClient` can override the `createSftpChannelSubsystem()` method to add, for example, some custom `RequestHandler` for SFTP sub-system requests and replies.

src/reference/antora/modules/ROOT/pages/sftp/session-factory.adoc
```diff
@@ -45,6 +45,21 @@ Now all you need to do is inject this SFTP session factory into your adapters.
 
 NOTE: A more practical way to provide values for the SFTP session factory is to use Spring's https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#beans-factory-placeholderconfigurer[property placeholder support].
 
+Starting with version 6.2, spring integration introduces a `createSftpClient(...)` to support a custom `SftpClient`.
+See a sample below of how to override `createSftpChannelSubsystem()` method in your custom `SftpClient` to add, for example, some custom RequestHandler for SFTP sub-system requests and replies:
```
artembilan, line 49:
```suggestion
See a sample below of how to override `createSftpChannelSubsystem()` method in your custom `SftpClient` to add, for example, some custom `RequestHandler` for SFTP sub-system requests and replies:
```

src/reference/antora/modules/ROOT/pages/sftp/session-factory.adoc
```diff
@@ -45,6 +45,21 @@ Now all you need to do is inject this SFTP session factory into your adapters.
...
+----
+@Override
+protected ChannelSubsystem createSftpChannelSubsystem(ClientSession clientSession) {
+    ChannelSubsystem sftpChannelSubsystem = super.createSftpChannelSubsystem(clientSession);
+    sftpChannelSubsystem.addRequestHandler((channel, request, wantReply, buffer) -> ...);
+    return sftpChannelSubsystem;
+}
+----
+
+NOTE: `createSftpClient(...)` can be overridden to create your custom `SftpClient` which will be used by `getSession()`.
```
artembilan, line 61: We don't need this sentence: it is just duplication of the first one in this paragraph.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
```diff
@@ -397,15 +403,34 @@ public void resetSharedSession() {
 		this.sharedSftpClient = null;
 	}
 
+	/**
+	 * This method allows the end-user to provide a custom {@link SftpClient}
```
artembilan, line 407: I don't think there is a reason to say on a method that it is a method.
Why just not: `Can be overridden to provide a custom`.
To be honest: I try to pursue "personless" docs as much as possible.
So, something like `you`, `we`, `end-user` are not appropriate words in our docs.

More over the `end-user` in this doc is misleading: the class extension and method override actions are indeed not for users of the compiled programs.

artembilan (changes requested):

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/10e79ba618e10f39b15631503b0b3732e328a32b after some clean up.
Back-ported to `6.1.x` as https://github.com/spring-projects/spring-integration/commit/78b09ed6102f1ca1a203e3a5d75220334257427e after some conflict resolutions.

Thank you for contribution!

PR #8743, opened 2023-09-20
Solving SSHD's issue, when the verify method is not found when passin…

Bloodika:
Solving SSHD's issue, when the verify method is not found when passing duration from version 2.10
java.lang.Object org.apache.sshd.client.future.ConnectFuture.verify(java.time.Duration)'

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/DefaultSftpSessionFactory.java
@@ -313,7 +313,7 @@ public class DefaultSftpSessionFactory implements SessionFactory<SftpClient.DirE
 
 		initClient();
 
-		Duration verifyTimeout = this.timeout != null ? Duration.ofMillis(this.timeout) : null;
+		long verifyTimeout = this.timeout != null ? this.timeout : null;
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Sorry, I'm not sure in the fix.
The current `main` is based on `org.apache.sshd:sshd-sftp:2.10.0`.
That `ConnectFuture` has these methods:
```
    default T verify(Duration timeout, CancelOption... options) throws IOException {
        return timeout != null ? verify(timeout.toMillis(), options) : verify(options);
    }
    
    T verify(long timeoutMillis, CancelOption... options) throws IOException;
```

Apparently you use some stable `6.1.x` version of Spring Integration which is based on the `2.9.x` yet and we cannot simply replace `sshd-sftp` to `2.10.0` there because the code of Spring Integration has to be compiled against that version.
Even if we keep it as a simple `long`, it fails with non-compatible byte code.

See more discussion here: https://stackoverflow.com/questions/76757768/spring-integration-sftp-v6-1-2-throws-nosuchmethoderror-org-apache-sshd-client-f.

The problem that they added that optional `CancelOption...` vararg into those methods and that makes our byte code incompatible with newer version.

Please, see in that SO thread that this might be a false positive vulnerability report because not everyone uses MINA SFTP server in production.

artembilan:
Closed as `Invalid` according a discussion above.

PR #8747, opened 2023-10-03
Fix maxMessagesPerPoll for SourcePollingChAdapter

artembilan:
The `AbstractMethodAnnotationPostProcessor` does not check for `PollerMetadata.MAX_MESSAGES_UNBOUNDED` before setting `maxMessagesPerPoll` into a `SourcePollingChannelAdapter` which in this case must be `1`

Also fix `SourcePollingChannelAdapterFactoryBean` to not mutate the provided `PollerMetadata` (which might be global default) with a new `maxMessagesPerPoll`


garyrussell (approved):

garyrussell:
...and cherry-picked.

PR #8758, opened 2023-10-10
GH-8757 let ObjectStringMessageConverter.toMessage() accept arbitrary payload

### LINKED ISSUE #8757: ObjectStringMessageConverter.toMessage() can't be used with arbitrary payload
I am not quite sure, what the original intention of `ObjectStringMessageConverter` was, but currently using
```java
ObjectStringMessageConverter converter = new ObjectStringMessageConverter();
Integer payload = 1234;
Message<?> converted = converter.toMessage(payload, null);
assertThat(converted).isNotNull();
```
fails, as `Message<?> converted` is always `null` when `payload` is anything else than `String.class` as `AbstractMessageConverter.canConvertTo()`

https://github.com/spring-projects/spring-framework/blob/80646591367b93d220b9dc0b0a126b203edc140c/spring-messaging/src/main/java/org/springframework/messaging/converter/AbstractMessageConverter.java#L231-L235

always returns `null` due to `StringMessageConverter.supports()`
https://github.com/spring-projects/spring-framework/blob/80646591367b93d220b9dc0b0a126b203edc140c/spring-messaging/src/main/java/org/springframework/messaging/converter/StringMessageConverter.java#L50-L55
returns `false` in this case.

The question arises, whether this is intentional. If `ObjectStringMessageConverter.toMessage()` should not be used with anything else than `String payload` explicitly throwing an `IllegalArgumentException` would probably be a better option. For the opposite case I will provide a PR.

mrpiggi:
Fixes #8757

* enhance ObjectStringMessageConverter to process arbitrary payload
* add appropriate tests

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
@@ -31,6 +31,7 @@ import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.integration.support.json.Jackson2JsonObjectMapper;
 import org.springframework.integration.support.json.JacksonPresent;
 import org.springframework.integration.support.utils.IntegrationUtils;
+import org.springframework.lang.Nullable;
 import org.springframework.messaging.converter.ByteArrayMessageConverter;
 import org.springframework.messaging.converter.CompositeMessageConverter;
 import org.springframework.messaging.converter.GenericMessageConverter;
@@ -59,6 +60,7 @@ public class ConfigurableCompositeMessageConverter extends CompositeMessageConve
 
 	private final boolean registerDefaults;
 
+	@Nullable
 	private BeanFactory beanFactory;
 
 	/**
@@ -97,6 +99,7 @@ public class ConfigurableCompositeMessageConverter extends CompositeMessageConve
 	@Override
 	public void afterPropertiesSet() {
 		if (this.registerDefaults) {
+			// TODO how to deal with this.beanFactory == null? just use DefaultConversionService.getSharedInstance()?
 			ConversionService conversionService = IntegrationUtils.getConversionService(this.beanFactory);
 			if (conversionService == null) {
 				conversionService = DefaultConversionService.getSharedInstance();
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
@@ -31,7 +31,6 @@ import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.integration.support.json.Jackson2JsonObjectMapper;
 import org.springframework.integration.support.json.JacksonPresent;
 import org.springframework.integration.support.utils.IntegrationUtils;
-import org.springframework.lang.Nullable;
 import org.springframework.messaging.converter.ByteArrayMessageConverter;
 import org.springframework.messaging.converter.CompositeMessageConverter;
 import org.springframework.messaging.converter.GenericMessageConverter;
@@ -60,7 +59,6 @@ public class ConfigurableCompositeMessageConverter extends CompositeMessageConve
 
 	private final boolean registerDefaults;
 
-	@Nullable
 	private BeanFactory beanFactory;
 
 	/**
@@ -99,7 +97,6 @@ public class ConfigurableCompositeMessageConverter extends CompositeMessageConve
 	@Override
 	public void afterPropertiesSet() {
 		if (this.registerDefaults) {
-			// TODO how to deal with this.beanFactory == null? just use DefaultConversionService.getSharedInstance()?
 			ConversionService conversionService = IntegrationUtils.getConversionService(this.beanFactory);
 			if (conversionService == null) {
 				conversionService = DefaultConversionService.getSharedInstance();
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/converter/DefaultDatatypeChannelMessageConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/converter/DefaultDatatypeChannelMessageConverter.java
@@ -59,8 +59,8 @@ public class DefaultDatatypeChannelMessageConverter implements MessageConverter,
 	}
 
 	@Override
-	public void setBeanFactory(@Nullable BeanFactory beanFactory) throws BeansException {
-		if (!this.conversionServiceSet && beanFactory != null) {
+	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
+		if (!this.conversionServiceSet) {
 			ConversionService integrationConversionService = IntegrationUtils.getConversionService(beanFactory);
 			if (integrationConversionService != null) {
 				this.conversionService = integrationConversionService;
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/converter/DefaultDatatypeChannelMessageConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/converter/DefaultDatatypeChannelMessageConverter.java
@@ -22,6 +22,7 @@ import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.integration.support.utils.IntegrationUtils;
+import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.converter.MessageConverter;
@@ -58,7 +59,7 @@ public class DefaultDatatypeChannelMessageConverter implements MessageConverter,
 	}
 
 	@Override
-	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
+	public void setBeanFactory(@Nullable BeanFactory beanFactory) throws BeansException {
 		if (!this.conversionServiceSet && beanFactory != null) {
 			ConversionService integrationConversionService = IntegrationUtils.getConversionService(beanFactory);
 			if (integrationConversionService != null) {
@@ -81,7 +82,7 @@ public class DefaultDatatypeChannelMessageConverter implements MessageConverter,
 	}
 
 	@Override
-	public Message<?> toMessage(Object payload, MessageHeaders header) {
+	public Message<?> toMessage(Object payload, @Nullable MessageHeaders header) {
 		throw new UnsupportedOperationException("This converter does not support this method");
 	}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
@@ -48,7 +48,6 @@ public class MapMessageConverter implements MessageConverter, BeanFactoryAware {
 
 	private volatile boolean filterHeadersInToMessage;
 
-	@Nullable
 	private volatile BeanFactory beanFactory;
 
 	private volatile MessageBuilderFactory messageBuilderFactory = new DefaultMessageBuilderFactory();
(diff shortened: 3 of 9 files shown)


mrpiggi:
As https://github.com/spring-projects/spring-integration/blob/22cbe14586a29aafc4d444c12a80c881ed8764d8/spring-integration-core/src/main/java/org/springframework/integration/support/converter/ObjectStringMessageConverter.java#L55-L59
is now not guarded by
https://github.com/spring-projects/spring-integration/blob/22cbe14586a29aafc4d444c12a80c881ed8764d8/spring-integration-core/src/main/java/org/springframework/integration/support/converter/ObjectStringMessageConverter.java#L50-L53
anymore, it is debatable, how `Class<?> targetClass` should be treated. It could either be enforced to `String.class` (currently implemented) or be unconditionally ignored or something in between like `Assert.isTrue(targetClass == null || targetClass == String.class)`

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ObjectStringMessageConverter.java
```diff
@@ -43,4 +65,10 @@ protected Object convertFromInternal(Message<?> message, Class<?> targetClass, O
 		}
 	}
 
+	@Override
+	@Nullable
+	protected Object convertToInternal(
+			Object payload, @Nullable MessageHeaders headers, @Nullable Object conversionHint) {
+		return super.convertToInternal(payload.toString(), headers, conversionHint);
```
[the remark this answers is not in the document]
mrpiggi: I already suspected that, as `ObjectStringMessageConverter` appears a bit odd. Maybe another name like `FallbackStringMessageConverter` would have been a better name.

mrpiggi:
As this boils down to adding only ctor `ObjectStringMessageConverter(Charset defaultCharset)` should I close this PR and create a new one with a sinble commit?

artembilan:
Bo need in a fresh PR: we squash commits on merge . Yes, you can add nullability for this package just in this PR.

mrpiggi:
squashed current state as I had to rebase on upstream anyways. will add `@NonNullApi ` on this PR this evening.

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ObjectStringMessageConverter.java
```diff
@@ -24,14 +26,30 @@
  * <p>
  * Delegates to super when payload is {@code byte[]} or {@code String}.
  * Performs {@link Object#toString()} in other cases.
+ * <p>
+ * This class meant to be used as a fallback converter internally when deserializing messages. Therefore, only
+ * {@link org.springframework.messaging.converter.AbstractMessageConverter#fromMessage(Message, Class) fromMessage}
+ * method should be called with {@code String.class} as {@code targetClass}, obviously.To be explicit, using method
```
artembilan, line 32: Sorry, I don't find words like `obviously` and `To be explicit` as formal technical documentation language.
I also don't this that we need to mention `toMessage()` in this Javadoc at all.
It is really the superclass responsibility to express its nature.
we override only `convertFromInternal()` and we should explain for what reason.
Nothing more to make the reader to feel silly.

spring-integration-core/src/test/java/org/springframework/integration/support/converter/ObjectStringMessageConverterTests.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+/**
+ * @author Falk Hanisch
+ * @since 6.2
+ *
+ */
+class ObjectStringMessageConverterTests {
+	private static final ArbitraryPayload PAYLOAD = new ArbitraryPayload(1234, "ÄÖÜ");
```
artembilan, line 35: We prefer to wrap every class member into blank lines, even the last method (or inner class, record) before final `}` should have a blank line after its declaration.

This way the code is much readable, especially during review here on GH.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
```diff
@@ -97,6 +99,7 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
 	@Override
 	public void afterPropertiesSet() {
 		if (this.registerDefaults) {
+			// TODO how to deal with this.beanFactory == null? just use DefaultConversionService.getSharedInstance()?
 			ConversionService conversionService = IntegrationUtils.getConversionService(this.beanFactory);
 			if (conversionService == null) {
 				conversionService = DefaultConversionService.getSharedInstance();
```
mrpiggi, line 105: If we consider `this.beanFactory` as `@Nullable` should we set `conversionService = DefaultConversionService.getSharedInstance();` as well or is some other behavior required?

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
```diff
@@ -43,10 +43,12 @@
  */
 public class MapMessageConverter implements MessageConverter, BeanFactoryAware {
 
-	private volatile String[] headerNames;
+	// TODO is initializing with an empty array correct or should toMessage() return 'null' if never set?
+	private volatile String[] headerNames = {};
```
mrpiggi, line 47: This is probably not desired. Should I add something like `Assert.notEmpty(this.headerNames, "at least one header name must be set with setHeaderNames()");` for `toMessage()` and `fromMessage()`

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
artembilan: The `BeanFactoryAware.setBeanFactory(BeanFactory)` is not nullable.
Not sure why do yo u make such an assumption over here then...

spring-integration-core/src/main/java/org/springframework/integration/support/converter/DefaultDatatypeChannelMessageConverter.java
```diff
@@ -58,7 +59,7 @@ public void setConversionService(ConversionService conversionService) {
 	}
 
 	@Override
-	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
+	public void setBeanFactory(@Nullable BeanFactory beanFactory) throws BeansException {
```
artembilan, line 62: Same here: we need to rework the logic with an assumption that it is not null.

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
artembilan: I think we need to make it nullable and implement respective logic.
It is going to be more efficient to avoid some extra calls if this one is just null.

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
```diff
@@ -43,10 +43,12 @@
...
 
 	private volatile boolean filterHeadersInToMessage;
 
+	@Nullable
```
artembilan, line 51: Same here: it is not supposed to be `null`

artembilan (changes requested):
I will look into failure of the MongoDB module myself.

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
mrpiggi: Well, the field `private BeanFactory beanFactory` is checked for `null` in several places throughout the whole framework although it is annotated with `@Nullable` only at very few places. I am fine with removing the annotation although it seems inconsistent regarding `@NonNullFields`

spring-integration-core/src/main/java/org/springframework/integration/support/converter/DefaultDatatypeChannelMessageConverter.java
mrpiggi: well in this case, we could just drop the check in the following if statement

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
mrpiggi: The question still remains, what logic should be implemented. Should `this.headerNames == null` be just handled like an empty array or is a different behavior needed?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
artembilan: I think what you did is enough.
It really looked like a bug by default where this `headerNames` is null and it is not handled properly.
Now it is OK. 
Let's remove that TODO!
Thank you!

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
mrpiggi: Fine. I adapted `setHeaderNames(String... headerNames)` to use an empty array if called with `null` or an empty array

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
mrpiggi: With inconsistency I mean that  field `private BeanFactory beanFactory` as actually marked as `@NonNull` via `@NonNullFields` but actually can be `null` until `setBeanFactory` was invoked. But as I said, this holds for the major part of the framework.

mrpiggi: Actually, for `ConfigurableCompositeMessageConverter` I did not change anything. Currently, there is no check `this.beanFactory != null` in `main`.

I just removed the check from `DefaultDatatypeChannelMessageConverter.setBeanFactory()` as you told me, it must not be called with `null` as an argument. In this case, getting a NPE is the expected behavior, isn't it?

artembilan: That's correct, but since we talk about Spring, it is expected that these classes have to be registered as beans in the application context.
Therefore that `setBeanFactory()` is always called and really never with `null`.
So, we may assume that it is null by some end-user mistake or we can just bite a bullet and let it fly for time being!

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
```diff
@@ -77,12 +78,16 @@ protected MessageBuilderFactory getMessageBuilderFactory() {
 	 * @param headerNames The header names.
 	 */
 	public void setHeaderNames(String... headerNames) {
-		Assert.notEmpty(headerNames, "at least one header name is required");
-		this.headerNames = Arrays.copyOf(headerNames, headerNames.length);
+		if (ObjectUtils.isEmpty(headerNames)) {
+			this.headerNames = new String[] {};
```
artembilan, line 82: Why? With `Assert.notEmpty()` we just disallow to set a null into this property.
I don't see anything what really has to be fixed in this method.
The default one is empty array, but if it is set, it must not be empty.
That's it. What wrong do you see with that logic?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/converter/MapMessageConverter.java
mrpiggi: With the current implementation, it can be changed to anything but the default state during runtime. So setting back to an empty array is not possible.

artembilan: OK. That makes sense.
I'll pull your PR locally shortly for final review and possible merge.

Thank you!

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
mrpiggi: I am totally aware of that. `@NonNullFields` is more or less a tool for developers and actually no part of the public API. I am fine with every call you make.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/29186e2b4609d30342c19e467b11e9eb45b1324d

@mrpiggi ,

PR #8766, opened 2023-10-16
Streamline UUIDConverter

tzolov:
- Use StandardCharsets.UTF_8 as a charset configuration instead of string. Later remove the necessity of handling encoder errors.
 - Use regular expressions to validate the UUID string standard representation. Later obsolete the need for try/catch exceptions.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
@@ -36,7 +36,7 @@ import org.springframework.util.ClassUtils;
 public class UUIDConverter implements Converter<Object, UUID> {
 
 	/**
-	 * @deprecated since 6.2 as it is not used internally by the UUIDConverter. The internal implementation relies, now,
+	 * @deprecated since 6.0.8 as it is not used internally by the UUIDConverter. The internal implementation relies, now,
 	 * on StandardCharsets.UTF_8 instead.
 	 */
 	@Deprecated
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
@@ -19,7 +19,6 @@ package org.springframework.integration.util;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectOutputStream;
-import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.UUID;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
```diff
@@ -30,15 +32,17 @@
  *
  * @author Dave Syer
  * @author Gary Russell
+ * @author Christian Tzolov
  */
 public class UUIDConverter implements Converter<Object, UUID> {
 
-	public static final String DEFAULT_CHARSET = "UTF-8";
+	public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
```
artembilan, line 39: We cannot do this change since it is a `public` property.
Since it is `final` and really cannot be changed, I suggest the change like:
1. Deprecate this constant: there is really just that `StandardCharsets.UTF_8`. This one in this class is just out of use.
2. Use the mentioned `StandardCharsets.UTF_8` in the code below instead of this constant.

spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
```diff
@@ -30,15 +32,17 @@
...
 
+	private static final Pattern UUID_REGEX = Pattern
+			.compile("^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
 
 	/**
-	 * Convert the input to a UUID using the convenience method
-	 * {@link #getUUID(Object)}.
+	 * Convert the input to a UUID using the convenience method {@link #getUUID(Object)}.
 	 *
```
artembilan, line 46: Consider to remove blank lines from method Javadocs since we are already here in changes for this class.

spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
```diff
@@ -115,4 +103,14 @@ private static byte[] serialize(Object object) {
 		return stream.toByteArray();
 	}
 
+	/**
+	 * Verifies if the provided UUID string complies with the standard representation.
```
artembilan, line 107: I don't see a reason in Javadocs for private methods, but if you still would like to have it here, please, consider to make it consistent we are rules:
1. Imperative (not infinitive) - therefore `Verify`.
2. No blank lines in method Javadocs.

https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

artembilan (changes requested):
Thank you for quick turn around!
I think we can accept this change and back-port it respectively.
Just see if my review makes sense!

spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
```diff
@@ -30,16 +31,22 @@
  *
  * @author Dave Syer
  * @author Gary Russell
+ * @author Christian Tzolov
  */
 public class UUIDConverter implements Converter<Object, UUID> {
 
+	/**
+	 * @deprecated since 6.2 as it is not used internally by the UUIDConverter. The internal implementation relies, now,
```
artembilan, line 39:
```suggestion
	 * @deprecated since 6.0.8 as it is not used internally by the UUIDConverter. The internal implementation relies, now,
```

artembilan:
Subsequent commit does some clean up: https://github.com/spring-projects/spring-integration/commit/ea8d917278f3e1f5c94e84e58a08a2d2da18ff64

Cherry-picked to `6.1.x` & `6.0.x` as https://github.com/spring-projects/spring-integration/commit/68f977a5887f881577abe4243dec14e5fe546028

PR #8771, opened 2023-10-20
KafkaInbounds should not mutate an externally provided RetryTemplate.

kurthong:
In order to achieve custom retry in batch mode, we may to use a RetryTemplate in listener itself. But if the RetryTemplate is shared with another KafkaMessageDrivenChannelAdapter, batch mode's ATTRIBUTES_HOLDER might be over-written by another KafkaMessageDrivenChannelAdapter's IntegrationRecordMessageListener. 

The situation is like shown below.
- There is only one RetryTemple bean in the application.
- There are two KafkaMessageDrivenChannelAdapters(A,B) in the application.
- A KafkaMessageDrivenChannelAdapter is batch mode and utilizing the retryTemplate in the listener.
- B KafkaMessageDrivenChannelAdapter is record mode and using the retryTemplate itself.
- (B KafkaMessageDrivenChannelAdapter's recordListener is registered in the retryTemplate.)
- When A retry is attempted in the listener, it will trigger B KafkaMessageDrivenChannelAdapter's recordListener.
- B KafkaMessageDrivenChannelAdapter's recordListener will overwrite A KafkaMessageDrivenChannelAdapter's ATTRIBUTES_HOLDER.

### DIFF
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 the original author or authors.
+ * Copyright 2022-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package org.springframework.integration.kafka.inbound;
 
 import org.apache.kafka.clients.consumer.Consumer;
 
+import org.springframework.core.AttributeAccessor;
 import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.support.Acknowledgment;
 import org.springframework.retry.RecoveryCallback;
@@ -64,16 +65,23 @@ public interface KafkaInboundEndpoint {
 
 		try {
 			template.execute(context -> {
-				context.setAttribute(CONTEXT_RECORD, data);
-				context.setAttribute(CONTEXT_ACKNOWLEDGMENT, acknowledgment);
-				context.setAttribute(CONTEXT_CONSUMER, consumer);
+				if (context.getRetryCount() == 0) {
+					context.setAttribute(CONTEXT_RECORD, data);
+					context.setAttribute(CONTEXT_ACKNOWLEDGMENT, acknowledgment);
+					context.setAttribute(CONTEXT_CONSUMER, consumer);
+					getAttributesHolder().set(context);
+				}
 				runnable.run();
 				return null;
 			}, callback);
 		}
 		catch (Exception ex) {
 			throw new KafkaException("Failed to execute runnable", ex);
+		} finally {
+			getAttributesHolder().remove();
 		}
 	}
+
+	ThreadLocal<AttributeAccessor> getAttributesHolder();
 
 }
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
@@ -17,8 +17,8 @@
 package org.springframework.integration.kafka.inbound;
 
 import org.apache.kafka.clients.consumer.Consumer;
-
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+
 import org.springframework.core.AttributeAccessor;
 import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.support.Acknowledgment;
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
@@ -77,7 +77,8 @@ public interface KafkaInboundEndpoint {
 		}
 		catch (Exception ex) {
 			throw new KafkaException("Failed to execute runnable", ex);
-		} finally {
+		}
+		finally {
 			getAttributesHolder().remove();
 		}
 	}
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
@@ -33,7 +33,7 @@ import org.springframework.retry.support.RetryTemplate;
  * @since 6.0
  *
  */
-public interface KafkaInboundEndpoint<K, V> {
+public interface KafkaInboundEndpoint {
 
 	/**
 	 * {@link org.springframework.retry.RetryContext} attribute key for an acknowledgment
@@ -63,7 +63,7 @@ public interface KafkaInboundEndpoint<K, V> {
 	 * @param consumer the consumer.
 	 * @param runnable the runnable.
 	 */
-	default void doWithRetry(RetryTemplate template, RecoveryCallback<?> callback, ConsumerRecord<K, V> record,
+	default void doWithRetry(RetryTemplate template, RecoveryCallback<?> callback, ConsumerRecord<?, ?> record,
 			Acknowledgment acknowledgment, Consumer<?, ?> consumer, Runnable runnable) {
 
 		try {
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
@@ -18,6 +18,7 @@ package org.springframework.integration.kafka.inbound;
 
 import org.apache.kafka.clients.consumer.Consumer;
 
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.springframework.core.AttributeAccessor;
 import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.support.Acknowledgment;
@@ -32,7 +33,7 @@ import org.springframework.retry.support.RetryTemplate;
  * @since 6.0
  *
  */
-public interface KafkaInboundEndpoint {
+public interface KafkaInboundEndpoint<K, V> {
 
 	/**
 	 * {@link org.springframework.retry.RetryContext} attribute key for an acknowledgment
@@ -51,6 +52,8 @@ public interface KafkaInboundEndpoint {
 	 */
 	String CONTEXT_RECORD = "record";
 
+	ThreadLocal<AttributeAccessor> ATTRIBUTES_HOLDER = new ThreadLocal<>();
+
 	/**
 	 * Execute the runnable with the retry template and recovery callback.
 	 * @param template the template.
@@ -60,16 +63,16 @@ public interface KafkaInboundEndpoint {
 	 * @param consumer the consumer.
 	 * @param runnable the runnable.
 	 */
-	default void doWithRetry(RetryTemplate template, RecoveryCallback<?> callback, Object data,
+	default void doWithRetry(RetryTemplate template, RecoveryCallback<?> callback, ConsumerRecord<K, V> record,
 			Acknowledgment acknowledgment, Consumer<?, ?> consumer, Runnable runnable) {
 
 		try {
 			template.execute(context -> {
 				if (context.getRetryCount() == 0) {
-					context.setAttribute(CONTEXT_RECORD, data);
+					context.setAttribute(CONTEXT_RECORD, record);
 					context.setAttribute(CONTEXT_ACKNOWLEDGMENT, acknowledgment);
 					context.setAttribute(CONTEXT_CONSUMER, consumer);
-					getAttributesHolder().set(context);
+					ATTRIBUTES_HOLDER.set(context);
 				}
 				runnable.run();
 				return null;
@@ -79,10 +82,8 @@ public interface KafkaInboundEndpoint {
 			throw new KafkaException("Failed to execute runnable", ex);
 		}
 		finally {
-			getAttributesHolder().remove();
+			ATTRIBUTES_HOLDER.remove();
 		}
 	}
-
-	ThreadLocal<AttributeAccessor> getAttributesHolder();
 
 }
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
@@ -74,9 +74,7 @@ import org.springframework.util.Assert;
  *
  */
 public class KafkaInboundGateway<K, V, R> extends MessagingGatewaySupport
-		implements KafkaInboundEndpoint, Pausable, OrderlyShutdownCapable {
-
-	private static final ThreadLocal<AttributeAccessor> ATTRIBUTES_HOLDER = new ThreadLocal<>();
+		implements KafkaInboundEndpoint<K, V>, Pausable, OrderlyShutdownCapable {
 
 	private final IntegrationRecordMessageListener listener = new IntegrationRecordMessageListener();
 
@@ -244,11 +242,6 @@ public class KafkaInboundGateway<K, V, R> extends MessagingGatewaySupport
 	@Override
 	public int afterShutdown() {
 		return getPhase();
-	}
-
-	@Override
-	public ThreadLocal<AttributeAccessor> getAttributesHolder() {
-		return ATTRIBUTES_HOLDER;
 	}
 
 	/**
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2022 the original author or authors.
+ * Copyright 2018-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


artembilan:
Would you mind to explain how one thread may effect a `ThreadLocal` in the other?
I see the flaw anyway.
With shared `RetryTemplate`, we just register those listeners form different channel adapters.
Therefore you must not share same `RetryTemplate`.

Not sure yet how to fix though ...

kurthong:
> Would you mind to explain how one thread may effect a `ThreadLocal` in the other? I see the flaw anyway. With shared `RetryTemplate`, we just register those listeners form different channel adapters. Therefore you must not share same `RetryTemplate`.
> 
> Not sure yet how to fix though ...

@artembilan I mean, when we use a shared RetryTemplate within the listener(output channel) itself by ourselves manually, a retry attempt causes those RetryTemplate listeners form different channel adapters in the current thread.

kurthong:
> As I said before: I don't see how `ThreadLocal` in one thread can be modified from another.

@artembilan hmm... I wonder we are on the same page. the issue is that `IntegrationRecordMessageListener` can be executed in unexpectred threads.

artembilan:
So, that `IntegrationRecordMessageListener` does only this:
```
		public <T, E extends Throwable> boolean open(RetryContext context, RetryCallback<T, E> callback) {
			if (KafkaMessageDrivenChannelAdapter.this.retryTemplate != null) {
				ATTRIBUTES_HOLDER.set(context);
			}
			return true;
		}
```

The `context` in this for a batch channel adapter and it is set to its `ThreadLocal`.
What is wrong with that? What problem may it cause?

I'm really looking into our current code, and I believe that `IntegrationBatchMessageListener` also must be registered as a listener into a `RetryTemplate`. To retry the whole batch.
Even if we have several those listeners in the same `RetryTemplate`, all of them are going to do the same `open()` action mentioned above.
Technically they are just idempotent and the content of the `ATTRIBUTES_HOLDER` is exactly what need in the current channel adapter scope.

What do I miss? How that is painful?

kurthong:
> So, that `IntegrationRecordMessageListener` does only this:
> 
> ```
> 		public <T, E extends Throwable> boolean open(RetryContext context, RetryCallback<T, E> callback) {
> 			if (KafkaMessageDrivenChannelAdapter.this.retryTemplate != null) {
> 				ATTRIBUTES_HOLDER.set(context);
> 			}
> 			return true;
> 		}
> ```
> 
> The `context` in this for a batch channel adapter and it is set to its `ThreadLocal`. What is wrong with that? What problem may it cause?

Yes. 
1. The batch channel adapter stores two attributes for later use in `org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter#setAttributesIfNecessary`
2. Unexpectedly `org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter.IntegrationRecordMessageListener#open` method overwrite the `ThreadLocal` with `RetryContext`. 
3. Then `org.springframework.integration.kafka.support.RawRecordHeaderErrorMessageStrategy#buildErrorMessage` fails to set header properly.
4. And `StaticMessageHeaderAccessor.getSourceData(errorMessage)` gets failure to retrieve source data.

I'm not clearly sure for that the usage of `StaticMessageHeaderAccessor.getSourceData(errorMessage)` in batch mode is proper way to retrieve the original message as I'm not the author of the code. But at least [`spring cloud stream` is utilizing it this way](https://github.com/sobychacko/spring-cloud-stream/blob/0133a95ebd086942b0ab2c6bc776906b211270a9/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L1165). 

> 
> I'm really looking into our current code, and I believe that `IntegrationBatchMessageListener` also must be registered as a listener into a `RetryTemplate`. To retry the whole batch. Even if we have several those listeners in the same `RetryTemplate`, all of them are going to do the same `open()` action mentioned above. Technically they are just idempotent and the content of the `ATTRIBUTES_HOLDER` is exactly what need in the current channel adapter scope.
> 
> What do I miss? How that is painful?

garyrussell:
It looks to me like we don't need to implement `RetryListener` at all; we could just move the code from `open` to before the `doWithRetry()` call and the code from close to a `finally` block after it.

We certainly shouldn't be mutating an externally provided `RetryTemplate`.

kurthong:
> It looks to me like we don't need to implement `RetryListener` at all; we could just move the code from `open` to before the `doWithRetry()` call and the code from close to a `finally` block after it.
> 
> We certainly shouldn't be mutating an externally provided `RetryTemplate`.

Yes, you are right. I thought the same but I chose just easier way as I was not that confidence:-(
Thank you for your advice:-)
Please review my 2nd commit.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -76,7 +74,7 @@
  *
  */
 public class KafkaInboundGateway<K, V, R> extends MessagingGatewaySupport
-		implements KafkaInboundEndpoint, Pausable, OrderlyShutdownCapable {
```
artembilan, line 79: I don't see a reason in this change. It is definitely a contract for endpoints, but not specific listener.
Would you mind explaining why did you go this way?

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -557,33 +536,10 @@ public void onMessage(List<ConsumerRecord<K, V>> records, Acknowledgment acknowl
 				message = toMessage(records, acknowledgment, consumer);
 			}
 			if (message != null) {
-				RetryTemplate template = KafkaMessageDrivenChannelAdapter.this.retryTemplate;
-				if (template != null) {
-					doWIthRetry(records, acknowledgment, consumer, message, template);
```
artembilan, line 562: I don't think it is too bad to have a retry for the whole batch. And I believe that's why that `KafkaInboundEndpoint` API is so generic.
Any explanations why would you like to get rid of this?
Indeed we still can use a built-in retry feature from Spring for Apache Kafka: https://docs.spring.io/spring-kafka/reference/kafka/annotation-error-handling.html#default-eh.

But then we just don't set a `RetryTemplate` on this channel adapter.

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
```diff
@@ -281,6 +281,65 @@ protected boolean doSend(Message<?> message, long timeout) {
...
+		containerProps.setIdleEventInterval(100L);
+		KafkaMessageListenerContainer<Integer, String> container =
+				new KafkaMessageListenerContainer<>(cf, containerProps);
+		KafkaMessageDrivenChannelAdapter<Integer, String> adapter = new KafkaMessageDrivenChannelAdapter<>(container,
+				ListenerMode.batch);
+		MessageChannel out = new DirectChannel() {
+
+			@Override
+			protected boolean doSend(Message<?> message, long timeout) {
+				return retryTemplate.execute(retryContext -> {
```
artembilan, line 314: Why do you use `RetryTemplate` in this piece of code?

artembilan (changes requested):
The overall change looks reasonable.
I have failed before to come up with that `if (context.getRetryCount() == 0) {` to avoid extra context interaction. 👍

artembilan:
I believe when you restore retry around batch and bring back that `KafkaInboundEndpoint`, we will be able to back-port the fix to other supported versions.
Even if we have a simple workaround with non-shared `RetryTemplate`, I'd like to have a code as stable as possible.

kurthong77:
> restore retry around batch and bring back that `KafkaInboundEndpoint`

I restored retry around batch and bring back `KafkaInboundEndpoint` to endpoints itself.
Thank you:-)

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
kurthong77: Yes, you are right. But, if we do that, `ATTRIBUTES_HOLDER` become publicly accessible via `KafkaMessageDrivenChannelAdapter#getAttributesHolder` method, and that is not good in terms of Encapsulation. So I tried to hide it from users.
Another option could be to move the methods within KafkaInboundEndpoint into MessagingMessageListenerAdapter or somewhere else.

Anyway, it might not be a big deal at the moment.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
kurthong77: `KafkaMessageDrivenChannelAdapter.retryTemplate` cannot be set in batch mode. Therefore, template != null will never be true.
```
	public void setRetryTemplate(RetryTemplate retryTemplate) {
		Assert.isTrue(retryTemplate == null || this.mode.equals(ListenerMode.record),
				"Retry is not supported with mode=batch");
		this.retryTemplate = retryTemplate;
	}
```
Is there another way to set a RetryTemplate for the channel adapter? Please correct me if I'm wrong.

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
kurthong77: That usage would call `RetryListener.open()` method and reset  `ATTRIBUTES_HOLDER` finally.
 
This is what I tried to explain with https://github.com/spring-projects/spring-integration/pull/8771#issuecomment-1775707880, https://github.com/spring-projects/spring-integration/pull/8771#issuecomment-1773585667.

It's an edge case and a very specific and complex scenario. It will be a kind of burden to maintenance and the subject of this fix is changed. So, I removed the test.

Shall I add several lines to `MessageDrivenAdapterTests#testInboundRecordRetryRecover` as shown below?
```
	@Test
	void testInboundRecordRetryRecover() {
                .....
		adapter.setOutputChannel(out);
		RetryTemplate retryTemplate = spy(new RetryTemplate()); <-- this line
		.....
                assertThat(receivedMessageHistory.get().toString()).isEqualTo("myNullChannel");
		verify(retryTemplate, never()).registerListener(any()); <-- this line
		verify(retryTemplate, never()).registerListener(any(),anyInt()); <-- this line
                .....
	}
```
Or add a test similar to the one below?
```
	@Test
	void testShouldNotMutateRetryTemplate() {
		Map<String, Object> props = KafkaTestUtils.consumerProps(EMBEDDED_BROKERS, "test4", "true");
		props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<>(props);
		ContainerProperties containerProps = new ContainerProperties(topic4);
		KafkaMessageListenerContainer<Integer, String> container =
				new KafkaMessageListenerContainer<>(cf, containerProps);
		KafkaMessageDrivenChannelAdapter<Integer, String> adapter = new KafkaMessageDrivenChannelAdapter<>(container);
		RetryTemplate retryTemplate = spy(new RetryTemplate());
		adapter.setRetryTemplate(retryTemplate);
		adapter.afterPropertiesSet();
		adapter.start();
		ContainerTestUtils.waitForAssignment(container, 1);
		
		verify(retryTemplate, never()).registerListener(any());
		verify(retryTemplate, never()).registerListener(any(),anyInt());

		adapter.stop();
	}
```

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
```diff
@@ -281,6 +281,65 @@ protected boolean doSend(Message<?> message, long timeout) {
...
+
+		Map<String, Object> props1 = KafkaTestUtils.consumerProps(EMBEDDED_BROKERS, "test4", "true");
+		props1.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+		DefaultKafkaConsumerFactory<Integer, String> cf1 = new DefaultKafkaConsumerFactory<>(props1);
+		ContainerProperties containerProps1 = new ContainerProperties(topic4);
+		KafkaMessageListenerContainer<Integer, String> container1 =
+				new KafkaMessageListenerContainer<>(cf1, containerProps1);
+		KafkaMessageDrivenChannelAdapter<Integer, String> adapter1 = new KafkaMessageDrivenChannelAdapter<>(container1);
+		adapter1.setRetryTemplate(retryTemplate);
+		adapter1.afterPropertiesSet();
```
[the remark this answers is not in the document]
kurthong77: `adapter1.afterPropertiesSet()` adds the `RetryListener` to the `RetryTemplate`, which means mutating the RetryTemplate. This is a precondition.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
```diff
@@ -64,16 +65,24 @@ default void doWithRetry(RetryTemplate template, RecoveryCallback<?> callback, O
 
 		try {
 			template.execute(context -> {
-				context.setAttribute(CONTEXT_RECORD, data);
-				context.setAttribute(CONTEXT_ACKNOWLEDGMENT, acknowledgment);
-				context.setAttribute(CONTEXT_CONSUMER, consumer);
+				if (context.getRetryCount() == 0) {
```
artembilan, line 68: I wonder if we can bring back a `ConsumerRecord` param type instead of that `Object data` since in the end we agree that we don't support retry for batch.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
```diff
@@ -64,16 +65,24 @@ default void doWithRetry(RetryTemplate template, RecoveryCallback<?> callback, O
...
+				}
 				runnable.run();
 				return null;
 			}, callback);
 		}
 		catch (Exception ex) {
 			throw new KafkaException("Failed to execute runnable", ex);
 		}
+		finally {
+			getAttributesHolder().remove();
```
artembilan, line 82: I wonder if instead of this method implementation requirement we can move that `ATTRIBUTES_HOLDER` into this interface.
All the impls of this one works in their onw thread and there really not possible to have several of them in the same thread.
So, I don't see any problems to have that constant on the interface.

artembilan (changes requested):
Please, consider also to bring back your change for the `IntegrationBatchMessageListener` since I admit my mistake and we indeed does not support retry for batch. Therefore that code is dead as you pointed before.

Thank you!

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
kurthong: Yes, that's good.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
kurthong: Yes, that looks more concise.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
```diff
@@ -31,7 +33,7 @@
  * @since 6.0
  *
  */
-public interface KafkaInboundEndpoint {
+public interface KafkaInboundEndpoint<K, V> {
```
artembilan, line 36: This would be a breaking change for the version we are going, but at the same time the `doWithRetry()` contract change to the `ConsumerRecord` is a breaking change anyway.
So, yeah, probably let's stick with the current version where we can introduce breaking change, but let's get rid of these generics! The `doWithRetry()` doesn't care what are those key and value types.
So, simple `ConsumerRecord<?, ?>` as a param should be enough. Similar to what we have so far with a `consumer` param.

artembilan (changes requested):
Everything else is good.
Thank you!

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
kurthong: Yes, it's done.

artembilan (changes requested):
There is a Checkstyle failure for your change:
```
 > Task :spring-integration-kafka:checkstyleMain
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java:21:1: Extra separation in import group before 'org.apache.kafka.clients.consumer.ConsumerRecord' [ImportOrder]
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java:22:1: 'org.springframework.core.AttributeAccessor' should be separated from previous imports. [ImportOrder]
```
Consider to run locally `./gradlew :spring-integration-kafka:check`

kurthong:
> There is a Checkstyle failure for your change:
> 
> ```
>  > Task :spring-integration-kafka:checkstyleMain
> Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java:21:1: Extra separation in import group before 'org.apache.kafka.clients.consumer.ConsumerRecord' [ImportOrder]
> Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java:22:1: 'org.springframework.core.AttributeAccessor' should be separated from previous imports. [ImportOrder]
> ```
> 
> Consider to run locally `./gradlew :spring-integration-kafka:check`

Yes, I got it. I don't know why my check was successful previously though.

artembilan:
@kurthong ,

PR #8777, opened 2023-10-25
GH-8770: Add `PostgresSubsChannel.errorHandler`

### LINKED ISSUE #8770: PostgresSubscribableChannel initial message polling dies due to exception
Currently, on the first subscribe of a MessageHandler, `notifyUpdate()` is called to process messages already in the message store.
The problem with `notifyUpdate()` is that the try-catch block is outside the loop, so the loop will die on an exception, leaving further messages unprocessed.
When it receives notifications for inserted messages, this is not a problem because those notifications will submit a new task to poll further messages.

We could move the try-catch inside the loop to prevent it from dying. (If we don't do this, we might as well remove the try-catch, as it's good for nothing) 
This would be fine for **non-transactional** usage. Messages are all processed and removed, and eventually, the loop stops.

This solution is a problem for **transactional** usage: The loop will never terminate, and the same message will be read repeatedly. For incoming notifications, we'd start further tasks that will go looping and exhaust the thread pool.
As of today, with transactional usage and without any other measures, messages producing an exception will also effectively block the channel (but not exhaust the thread pool)

[Here is a reproducing test case and the mentioned solution](https://github.com/joshiste/spring-integration/commit/08839725100c9aaf62c4ae4384b234876ff1c566). I'm not very lucky with that solution but currently have no better idea.

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8770

The problem with the `PostgresSubscribableChannel.notifyUpdate()` is that the try-catch block is outside the loop, so the loop will die on an exception, leaving further messages unprocessed.

* Add ``PostgresSubscribableChannel.errorHandler` option to be invoked after a `RetryTemplate` and for every failed message.
* The `askForMessage()` new logic is to catch an exception on a message and call `errorHandler` returning a `FALLBACK_STUB` to continue an outer loop in the `notifyUpdate()`

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
@@ -180,14 +180,14 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 		this.executor.execute(() -> {
 			Optional<?> dispatchedMessage;
 			do {
-				dispatchedMessage = askForMessage();
+				dispatchedMessage = pollAndDispatchMessage();
 			} while (dispatchedMessage.isPresent());
 		});
 	}
 
-	private Optional<?> askForMessage() {
+	private Optional<?> pollAndDispatchMessage() {
 		try {
-			return doAskForMessage();
+			return doPollAndDispatchMessage();
 		}
 		catch (Exception ex) {
 			try {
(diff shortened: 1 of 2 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -160,19 +178,29 @@ protected boolean doSend(Message<?> message, long timeout) {
...
 	public void notifyUpdate() {
 		this.executor.execute(() -> {
+			Optional<?> dispatchedMessage;
+			do {
+				dispatchedMessage = askForMessage();
+			} while (dispatchedMessage.isPresent());
+		});
+	}
+
+	private Optional<?> askForMessage() {
```
garyrussell, line 188: These method names seem to be misleading; we are actually dispatching; maybe `dispatchMessageIfPresent`?

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
...and cherry-picked after resolving minor conflict.

PR #8788, opened 2023-11-03
GH-8786: Make `FtpSession.finalizeRaw()` robust

### LINKED ISSUE #8786: FtpSession readRaw + finalizeRaw blocking until 421 timeout

6.1.2 - 6.1.4, and probably older too

**Describe the bug**

I am using the spring integration ftp package to connect to an ftp server to try and download some files.
Before downloading the actual files I created a check to see if the file is actually downloadable in the following way:

1. Open inputStream using FtpSession readRaw(String source) method. 
2. Check if inputStream is not null and if no Exception is thrown
3. Close inputStream
4. call FtpSession finalizeRaw() to complete the transaction

However, when I do this for a file to which my ftpuser does not have access by OS permissions, the readRaw method will throw an Exception because of a 550 FTP error. This makes sense.
Then when I call finalizeRaw to allow to check or download another file afterwards, the readingRaw atomic boolean is correctly set to false. However when the finalizeRaw method calls this.client.completePendingCommand() it hangs until it receives a 421 (timeout) error.

Therefore my assumption that it seems the completePendingCommand should not be called when readRaw throws an Exception.
However the atomic boolean readingRaw is still set to true, so I am not able to skip the finalizeRaw call since this would not allow me to do another readRaw call.

**To Reproduce**

See above, when using readraw + finalizeraw on a path to a file to which the ftp user does not have os permissions

**Expected behavior**

Either of 3 possibilities:
1. The readRaw method should put the atomicBoolean readingRaw to false because the process is no longer active
2. There should be a (different) way to call finalize when readRaw threw an IOException
3. The finalizeRaw method should not block upon completePendingCommand

artembilan:
Fixes https://github.com/spring-projects/spring-integration/issues/8786

If `FtpSession.readRaw()` fails, the next `FtpSession.finalizeRaw()` call would lead to `FTPClient.completePendingCommand()` failure since there is no command to finish.

* Fix `FtpSession.finalizeRaw()` to exit earlier positively in case of `FTPReply.isNegativePermanent()` for the current reply code set by the failure from a previous `FtpSession.readRaw()`

### DIFF
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


tzolov:
Merged at fd6fb71ce77b9b8ff4274c41274de76f5624c3f2
Also above commit was cherry-picked to 6.1.x, 6.0.x 5.5.x along with ef327e98f6163a29018fc8e5833267dbb1994388

PR #8790, opened 2023-11-06
Fix docs ambiguity.

ahrytsiuk:
- Replaced pharase 'String with version' to 'Starting with version'
- Fixed formatting for code block


artembilan:
@ahrytsiuk ,

Issue #8792, opened 2023-11-08
Problem when filtering with max fetch size > 1


abraham-macias:
**In what version(s) of Spring Integration are you seeing this issue?**

Spring integration SFTP v6.1.3

**Describe the bug**

If you combine SftpSimplePatternFileListFilter (so you use *.trg files) and SftpPersistentAcceptOnceFileListFilter (so you only accept one file once) by using a ChainFileListFilter, the resulting filter has "supportsSingleFileFiltering" to false.
This causes the filtering to be done and rollback (necessary for SftpPersistentAcceptOnceFileListFilter) in AbstractRemoteFileStreamingMessageSource.listFiles() method.

When you set max fetch size to a number bigger than one (for example 5) and at a certain point it is necessary to "this.toBeReceived.clear()" inside "AbstractRemoteFileStreamingMessageSource.doReceive()", those removed elements from toBeReceived are not rolled back.

**To Reproduce**

1) Use a SftpStreamingMessageSource with a maxFetchSize = 5 and a ChainFileListFilter filter composed by SftpSimplePatternFileListFilter and SftpPersistentAcceptOnceFileListFilter.
2) Put 2 files in the folder and invoke SftpStreamingMessageSource.receive() method twice.
3) Put 5 files in the folder and invoke SftpStreamingMessageSource.receive() method five times.
4) The last two files won't be received.

**Expected behavior**

All the 7 files should be received.

**Sample**

See the related pull request.

artembilan:
And the fix has been cherry-picked to all requested branches.

PR #8793, opened 2023-11-08
Fix problem with max fetch size > 1

abraham-macias:
This patch aims to solve the problem of the issue #8792 where when using max fetch size > 1 in combination with a supportsSingleFileFiltering = false file filter that needs to be rollback with unread messages can produce problems.


abraham-macias:
“@pivotal-cla This is an Obvious Fix”.

artembilan:
@abraham-macias ,

this does not look like an "Obvious Fix".
Please, consider to sign the CLA: you have added (and changed) a lot of code.
Thanks

artembilan:
I agree with the problem and fix, but since you have not signed CLA, we cannot accept your PR.
Will be replaced with other fix variant.

Thanks for understanding.

artembilan:
Superseded with: https://github.com/spring-projects/spring-integration/commit/22c4db291b5ec02714d279cb13a0b4f2d6019a8c

PR #8801, opened 2023-12-06
GH-8800: Fix SMB Gateway for recursive MGET

artembilan:
Fixes: gh-8800

The `SmbFile.getName()` returns plain file name, not with a sub-directory as it is expected by the `AbstractRemoteFileOutboundGateway` for (S)FTP protocols

* Introduce `AbstractRemoteFileOutboundGateway.getFullFileName()` with dir and file name concatenation by default (as it was before) and override it in the `SmbOutboundGateway` to return the full path according this protocol.
* Fix `SmbOutboundGateway.enhanceNameWithSubDirectory()` to return an `SmbFile` instance since it has all the info about remote file, include requested directory
* Fix `SmbOutboundGateway.getFilename()` to remove the trailing `/` in the dir name since this is not what expected by the `AbstractRemoteFileOutboundGateway` logic

**Cherry-pick to `6.1.x`**


artembilan:
Merged and cherry-picked to `6.1.x`

PR #8880, opened 2024-02-01
MQTT: Use subscription identifiers. Fixes #8879.

### LINKED ISSUE #8879: Spring Integration hits a bug in Paho client with $share/ MQTT subscriptions
Spring Integration hits a bug in Paho client with $share/ MQTT subscriptions

For example:

6.1.5

**Describe the bug**

Spring Integration hits a bug (eclipse/paho.mqtt.java/issues/827) in Paho MQTT client with `$share/` MQTT subscriptions. This bug can be worked around by using subscription ID by providing it in `MqttProperties` when calling, e.g., `org.eclipse.paho.mqttv5.client.IMqttAsyncClient#subscribe(org.eclipse.paho.mqttv5.common.MqttSubscription[], java.lang.Object, org.eclipse.paho.mqttv5.client.MqttActionListener, org.eclipse.paho.mqttv5.client.IMqttMessageListener, org.eclipse.paho.mqttv5.common.packet.MqttProperties)`. Unfortunately, Spring Integration currently calls only `org.eclipse.paho.mqttv5.client.IMqttAsyncClient#subscribe(org.eclipse.paho.mqttv5.common.MqttSubscription, org.eclipse.paho.mqttv5.client.IMqttMessageListener)` from `org.springframework.integration.mqtt.inbound.Mqttv5PahoMessageDrivenChannelAdapter#addTopic` and does not provide subscription ID.

**To Reproduce**

Subscribe on `$share/sharegroup/topic` using `Mqttv5PahoMessageDrivenChannelAdapter` and send message to `topic`. The message will be received but it will be rejected by Paho itself because it is comparing the filter and the message topic textually without removing the `$share/sharegroup/` prefix.

**Expected behavior**

`Mqttv5PahoMessageDrivenChannelAdapter` should be able to subscribe to shared MQTT topics.

**Sample**

N/A

wilx:


artembilan:
Would you mind to issue this PR against `main` instead, please?

We still need to apply the change over there as well.
And then we will back-port it to all the supported versions.

Thanks

artembilan (changes requested):
Thank you for the test!
However I still would like to see the PR agains `main`.

PR #8881, opened 2024-02-01
MQTT: Use subscription identifiers. Fixes #8879.

### LINKED ISSUE #8879: Spring Integration hits a bug in Paho client with $share/ MQTT subscriptions

wilx:
This works around eclipse/paho.mqtt.java/issues/827 issue because it avoids textual comparison of subscription topics.

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.IntStream;
@@ -106,6 +107,8 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 	private HeaderMapper<MqttProperties> headerMapper = new MqttHeaderMapper();
 
 	private volatile boolean readyToSubscribeOnStart;
+
+	private final AtomicInteger subIdCounter = new AtomicInteger(0);
 
 	/**
 	 * Create an instance based on the MQTT url, client id and subscriptions.
@@ -336,8 +339,10 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 				this.subscriptions.add(subscription);
 			}
 			if (this.mqttClient != null && this.mqttClient.isConnected()) {
+				MqttProperties subsProperties = new MqttProperties();
+				subsProperties.setSubscriptionIdentifier(this.subIdCounter.incrementAndGet());
 				this.mqttClient.subscribe(new MqttSubscription[] { subscription },
-								null, null, new IMqttMessageListener[] { this::messageArrived }, new MqttProperties())
+								null, null, new IMqttMessageListener[] { this::messageArrived }, subsProperties)
 						.waitForCompletion(getCompletionTimeout());
 			}
 		}
@@ -467,7 +472,11 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 			IMqttMessageListener[] listeners = IntStream.range(0, mqttSubscriptions.length)
 					.mapToObj(t -> listener)
 					.toArray(IMqttMessageListener[]::new);
-			this.mqttClient.subscribe(mqttSubscriptions, null, null, listeners, new MqttProperties())
+			MqttProperties subsProperties = new MqttProperties();
+			subsProperties.setSubscriptionIdentifiers(IntStream.range(0, mqttSubscriptions.length)
+					.mapToObj(i -> this.subIdCounter.incrementAndGet())
+					.toList());
+			this.mqttClient.subscribe(mqttSubscriptions, null, null, listeners, subsProperties)
 					.waitForCompletion(getCompletionTimeout());
 			String message = "Connected and subscribed to " + Arrays.toString(mqttSubscriptions);
 			logger.debug(message);
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -339,10 +339,10 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 				this.subscriptions.add(subscription);
 			}
 			if (this.mqttClient != null && this.mqttClient.isConnected()) {
-				MqttProperties subsProperties = new MqttProperties();
-				subsProperties.setSubscriptionIdentifier(this.subIdCounter.incrementAndGet());
+				MqttProperties subscriptionProperties = new MqttProperties();
+				subscriptionProperties.setSubscriptionIdentifier(this.subIdCounter.incrementAndGet());
 				this.mqttClient.subscribe(new MqttSubscription[] { subscription },
-								null, null, new IMqttMessageListener[] { this::messageArrived }, subsProperties)
+								null, null, new IMqttMessageListener[] { this::messageArrived }, subscriptionProperties)
 						.waitForCompletion(getCompletionTimeout());
 			}
 		}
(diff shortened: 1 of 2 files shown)


vhbcm:
Fixes #8879

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -107,6 +108,8 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 
 	private volatile boolean readyToSubscribeOnStart;
 
+	private final AtomicInteger subIdCounter = new AtomicInteger(0);
```
artembilan, line 111: I'm still not sure in this incrementing approach.
Let's imaging we have several `Mqttv5PahoMessageDrivenChannelAdapter` instances!
Each of them with its own client id, but still same set of topics to subscribe.
However all of them are started with the same initial count.
So, different instances on the same topics might get same identifiers.

Now another scenario: we have several instances in the cluster with the same `clientId`, but according to their different JVMs they might get different identifiers.

So, how it is really supposed to work correctly if there is no correlation between subscriber instance and that identifier?

I need more proves that a plain counter which produces same number for different instances is really what is enough for those shared subscriptions to work.

Thanks

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
wilx: Subscription identifier is meaningful only to the client that does the SUBSCRIBE operation. 

From the spec 3.8.2.1.2 Subscription Identifier: _[..]The Subscription Identifier is associated with any subscription created or modified as the result of this SUBSCRIBE packet. If there is a Subscription Identifier, it is stored with the subscription. If this property is not specified, then the absence of a Subscription Identifier is stored with the subscription._

From the spec 3.8.4 SUBSCRIBE Actions: _The Subscription Identifiers are part of the Session State in the Server and are returned to the Client receiving a matching PUBLISH packet. They are removed from the Server’s Session State when the Server receives an UNSUBSCRIBE packet, when the Server receives a SUBSCRIBE packet from the Client for the same Topic Filter but with a different Subscription Identifier or with no Subscription Identifier, or when the Server sends Session Present 0 in a CONNACK packet._

artembilan: I'm sorry, I'm still failing to understand that description.

Would you mind to rephrase it some how to confirm that plain counter is enough for us for any use-cases I described before?

Thanks

wilx: The subscription identifiers are part of the server side session state for the client. From [4.1 Session State](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Session_State):

> The Session State in the Server consists of:
> [..]
> The Clients subscriptions, including any Subscription Identifiers.

So the scope of Subscription identifiers is the session.

> Let's imaging we have several Mqttv5PahoMessageDrivenChannelAdapter instances!
> Each of them with its own client id, but still same set of topics to subscribe.
> However all of them are started with the same initial count.
> So, different instances on the same topics might get same identifiers.

If they add the topics in the same order each then they will generate the same Subscription identifiers. But that is fine because those Subscription identifiers are per session.

> Now another scenario: we have several instances in the cluster with the same clientId, but according to their different JVMs they might get different identifiers.

You cannot have the same client ID for multiple connections. If the server gets a connection with the same client ID, it has to terminate the existing connection and use the new connection. From [3.1.4 CONNECT Actions](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901073):

> [..]If the ClientID represents a Client already connected to the Server, the Server sends a DISCONNECT packet to the existing Client with Reason Code of 0x8E (Session taken over) as described in [section 4.13](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#S4_13_Errors) and MUST close the Network Connection of the existing Client.

[ActiveMQ Artemis](https://activemq.apache.org/components/artemis/documentation/latest/mqtt.html#link-stealing) allows to change this and reject the new connection in favour of the existing one.

> I need more proves that a plain counter which produces same number for different instances is really what is enough for those shared subscriptions to work.

It works because the Subscription identifier is per session and because you cannot have multiple connection with the same client ID.

artembilan: Sounds good!

Reviewing, merging and back-porting then.

Thank you!

artembilan: Since we do an Open Source here, it is better to be as readable as possible.
Therefore we prefer more clear names even for local variables.
My request here is to make this property as `subscriptionIdentifierCounter`

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -336,8 +339,10 @@ public void addTopic(String topic, int qos) {
 				this.subscriptions.add(subscription);
 			}
 			if (this.mqttClient != null && this.mqttClient.isConnected()) {
+				MqttProperties subsProperties = new MqttProperties();
```
artembilan, line 342: And this one name, please, as `subscriptionProperties`.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -467,7 +472,11 @@ private void subscribe() {
 			IMqttMessageListener[] listeners = IntStream.range(0, mqttSubscriptions.length)
 					.mapToObj(t -> listener)
 					.toArray(IMqttMessageListener[]::new);
-			this.mqttClient.subscribe(mqttSubscriptions, null, null, listeners, new MqttProperties())
+			MqttProperties subsProperties = new MqttProperties();
```
artembilan, line 475: DITTO

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
wilx: Fixed.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
wilx: Fixed.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
wilx: Fixed.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
artembilan: No need to say `Fixed` if you agree with me and going to incorporate that in the next commit for the PR.
I'll see those changes this or other way.
Otherwise this sounds like a spam in my email box 😄 

Only exclusion is a force-push where GitHub don't generate email notification for some reason.

artembilan:
... and back-ported to `6.2.x` & `6.1.x`.

@wilx ,

And if I understood your correctly, the `6.1.6` is scheduled for February, 20th.

aditya-pathak-dc:
@wilx  and @artembilan I am using 6.3.0 version of spring-mqtt-integration, but still facing the same issue. Unable to receive message on shared topics.
Here are the details: https://stackoverflow.com/questions/78605910/spring-mqtt-integration-shared-subscritopn-is-not-working

Issue #8898, opened 2024-02-11
[SFTP Adapters]Allow clearing unprocessed files in AbstractRemoteFileStreamingMessageSource


kkosiacki:
**Expected Behavior**

When I'm using RotatingServerAdvice when pooling files from different remote, I want to have ability to remove unprocessed files from queue, to avoid trying to get file from remote when Session is rotated. 

**Current Behavior**

When you have `DelegatingSessionFactory `with `RotatingServerAdvice `and there are unprocessed files in queue (see `toBeReceived ` in `AbstractRemoteFileStreamingMessageSource`), when you call `doReceive` MessageSource will try to retrieve file from queue and it will be success or not depends on `Session `return from `DelegatingSessionFactory`. If rotation happened and 'new' session is not able to retrieve file from remote, exception is thrown (access denied) that clears that internal queue.

**Context**
I want to add public method on `AbstractRemoteFileStreamingMessageSource` thats allow developer to decide if he wants to clean this queue (before `doReceive`). It means that you can write own `RotatingPolicy` that will clear `toBeReceived ` queue if necessary.

PR #8899, opened 2024-02-11
GH-8898: method to clear queue in AbstractRemoteFileStreamingMessageS…

kkosiacki:
This Pull Request is fixing problem describe in [GH-8898 ](https://github.com/spring-projects/spring-integration/issues/8898)

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -184,17 +184,7 @@ public abstract class AbstractRemoteFileStreamingMessageSource<F>
 	@Override
 	public void stop() {
 		if (this.running.compareAndSet(true, false)) {
-			if (this.filter == null || this.filter.supportsSingleFileFiltering()) {
-				this.toBeReceived.clear();
-			}
-			else {
-				// remove unprocessed files from the queue (and filter)
-				AbstractFileInfo<F> file = this.toBeReceived.poll();
-				while (file != null) {
-					resetFilterIfNecessary(file);
-					file = this.toBeReceived.poll();
-				}
-			}
+			clearUnprocessedFilesQueue();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java
```diff
@@ -328,4 +318,21 @@ protected void rollbackFromFileToListEnd(List<F> filteredFiles, F file) {
 
 	protected abstract boolean isDirectory(F file);
 
+	/**
+	 * Clear internal queue of listed files..
+	 */
+	public void clearUnprocessedFilesQueue() {
```
artembilan, line 324: OK. I cannot imaging the scenario when it is necessary, but let's give it a more meaningful name like `clearFetchedCache()` and some good Javadoc, plus `@since 6.3` and your name to the `@author` list of the affected classes.

I think another fix (separate issue/PR) must be done in the `StandardRotationPolicy` to enforce `setMaxFetchSize(1)` when mode is `fair`. Otherwise that fetched cache really does not make sense since we have just rotated to another dir and (possible) session factory. So, we would expect fresh file for that condition.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/RemoteFileStreamingMessageSourceTests.java
```diff
@@ -31,13 +31,14 @@
...
+import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
+import org.springframework.messaging.Message;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
artembilan, line 41: No asterisk imports.

See if `spring-framework.xml` in the `src/idea` or similar in `eclipse` can help you to re-align your IDE with our code style.

artembilan (changes requested):
This new API has also to be mentioned in the `whats-new.adoc`.
You also can talk about it in the `rotating-server-advice.adoc`.

Thanks

artembilan (changes requested):
See the fix for this issue: https://github.com/spring-projects/spring-integration/issues/8967.

Might be that is enough for your use-case and we don't need to do anything else?

Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/753916ca2432b88bf4587791dab3700e4a0ba0a5 after some cleanup and docs.

Thank you for contribution anyway; looking forward for more!

PR #8980, opened 2024-03-04
GH-7925: Make message history header as mutable

### LINKED ISSUE #7925: Consider a Mechanism to Avoid Message Rebuilding on Every Track [INT-3983]
**[Gary Russell](https://jira.spring.io/secure/ViewProfile.jspa?name=grussell)** opened **[INT-3983](https://jira.spring.io/browse/INT-3983?redirect=false)** and commented

See the discussion on https://github.com/spring-projects/spring-integration/pull/1764

---
No further details from [INT-3983](https://jira.spring.io/browse/INT-3983?redirect=false)

artembilan:
Fixes: #7925

The `MessageHistory.write()` creates not only a new instance of the `MessageHistory`, but also a new copy of the whole message.
This significantly impacts the performance when we have too many components to track

* Make `MessageHistory` as append-only container and create a new instance (plus message) only on the first track when no prior history is present
* Change `WireTap`, `BroadcastingDispatcher`, `AbstractMessageRouter` and `AbstractMessageSplitter` to use a new `AbstractIntegrationMessageBuilder.cloneMessageHistoryIfAny()` API for every branch a message is produced.
Essentially, create a new message with copy of the message history to let that downstream sub-flow have its own trace
* Modify failed unit tests for a new logic where message history header is not immutable anymore
* This also fixes an `AbstractMessageSplitter` for propagating its track into messages it emits

### DIFF
--- a/src/reference/antora/modules/ROOT/pages/message-history.adoc
+++ b/src/reference/antora/modules/ROOT/pages/message-history.adoc
@@ -129,7 +129,9 @@ The MBean's object name is `<domain>:name=messageHistoryConfigurer,type=MessageH
 IMPORTANT: Only one `@EnableMessageHistory` (or `<message-history/>`) must be declared in the application context as single source for components tracking configuration.
 Do not use a generic bean definition for the `MessageHistoryConfigurer`.
 
-NOTE: By definition, the message history header is immutable (you cannot re-write history).
-Therefore, when writing message history values, the components either create new messages (when the component is an origin) or they copy the history from a request message, modifying it and setting the new list on a reply message.
-In either case, the values can be appended even if the message itself is crossing thread boundaries.
-That means that the history values can greatly simplify debugging in an asynchronous message flow.
+NOTE: Prior to version 6.3, the message history header was immutable (you cannot re-write history): every single track created not only new instance of the `MessageHistory`, but a fully new message copy.
+Now it works in append-only mode: the first track creates a new message with a new `MessageHistory` container.
+All the rest `MessageHistory.write()` calls add new entries to existing header - and no new message created.
+This significantly improves an application performance.
+All the components in the framework, where same message can be sent to several consumers (`PublishSubscribeChannel`, `AbstractMessageRouter`, `WireTap` etc.) (or splitter produces several outputs based on input message), are now cloning an existing `MessageHistory` header into those new messages.
+For any other multi-producing use-cases, outside the framework scope, the `AbstractIntegrationMessageBuilder.cloneMessageHistoryIfAny()` API is recommended to ensure that parallel downstream sub-flows contribute their own message history traces.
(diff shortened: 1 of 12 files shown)


src/reference/antora/modules/ROOT/pages/message-history.adoc
```diff
@@ -129,7 +129,9 @@ The MBean's object name is `<domain>:name=messageHistoryConfigurer,type=MessageH
...
 Do not use a generic bean definition for the `MessageHistoryConfigurer`.
 
-NOTE: By definition, the message history header is immutable (you cannot re-write history).
-Therefore, when writing message history values, the components either create new messages (when the component is an origin) or they copy the history from a request message, modifying it and setting the new list on a reply message.
-In either case, the values can be appended even if the message itself is crossing thread boundaries.
-That means that the history values can greatly simplify debugging in an asynchronous message flow.
+NOTE: Prior to version 6.3, the message history header was immutable (you cannot re-write history): every single track created not only new instance of the `MessageHistory`, but a fully new message copy.
+Now it works in append-only mode: the first track creates a new message with a new `MessageHistory` container.
+All the rest `MessageHistory.write()` calls add new entries to existing header - and no new message created.
+This significantly improves an application performance.
```
sobychacko, line 135: `this` instead of `an`?

src/reference/antora/modules/ROOT/pages/message-history.adoc
```diff
@@ -129,7 +129,9 @@ The MBean's object name is `<domain>:name=messageHistoryConfigurer,type=MessageH
...
+All the components in the framework, where same message can be sent to several consumers (`PublishSubscribeChannel`, `AbstractMessageRouter`, `WireTap` etc.) (or splitter produces several outputs based on input message), are now cloning an existing `MessageHistory` header into those new messages.
```
sobychacko, line 136: I don't see the need for the parenthesis around the sentence, `or splitter produces...`.  Also ` or where splitter produces...`?

sobychacko (changes requested):

sobychacko:
@artembilan Looks great based on our code walkthrough. Some very minor nit-picks on the docs. That's all.

src/reference/antora/modules/ROOT/pages/message-history.adoc
artembilan: Is it OK to have several `where` in the sequence?
I mean we already have one `where` in the beginning.
This `or` is just one more in the sequence. 
So, might just no parenthesis and that's it?

src/reference/antora/modules/ROOT/pages/message-history.adoc
sobychacko: oh ya, i meant `the`. sorry.

src/reference/antora/modules/ROOT/pages/message-history.adoc
sobychacko: ya, thats fine.

sobychacko (approved):

PR #8982, opened 2024-03-08
GH-8981: Add `UnicastingDispatcher.failoverStrategy` option

### LINKED ISSUE #8981: Conditional failover in Direct Channel - Introduce Predicate<Exception> failoverStrategy into DirectChannel and its UnicastingDispatcher ( and in ExecutorChannel if possible)
**Expected Behavior**

  If I use DirectChannel and subscribe to it many ServiceActivators I would like to handle error - by failover strategy on DirectChannel level - not on ErrorChannel.  Reason of having many subscribes on one channel is having load balancing.  

  So if there will be expected Exception I could stop processing and pass it to error channel instead of passing message to other subscriber.

  Example - to one Direct channel I have connected two JmsOutboundGateway - both have same logic but they are connected to different mq servers - to have load balancing.

  I have request that I process in my app and if I'm sending something via JmsOutboundGateway in case of MessageTimeoutException - due to long timeout I want to stop processing whole request in my app ( timeout for whole request is short). But in case its different error for example problem with connection JmsOutboundGateway to mq I want to use other subscribers connected to different server (use failover strategy).

**Current Behavior**
  
Currently I can't do conditional failover - if something fails other subscribers will be involved. I can turn off failover in channel and build some flows with ErrorMessageExceptionTypeRouter  but then I will have to implement by my own round robin and failover strategy.

**Context**

A lot of custom development.

What are you trying to accomplish?
App with high volume of request with load balancing.

Coding everything myself.

Yes. I'd like to avoid them.

artembilan:
Fixes: #8981

Sometime the simple `boolean failover` on the `MessageChannel` (default `true`) is not enough to be sure that we can dispatch to the next handler or not. Such a decision can be made using `ErrorMessageExceptionTypeRouter`, but that would require an overhaul for the whole integration flow

* Introduce a simple `Predicate<Exception> failoverStrategy` into `UnicastingDispatcher` and all its `MessageChannel` implementation consumers to allow to make a decision about next failover according to a thrown exception from the current `MessageHandler`
* Expose `failoverStrategy` on the `DirectChannel`, `ExecutorChannel` & `PartitionedChannel`, and add it into respective Java DSL specs
* Fix involved tests to rely on the `failoverStrategy` property from now on
* Document the new feature

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/channel/config/DispatchingChannelParserTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/channel/config/DispatchingChannelParserTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,9 +19,9 @@ package org.springframework.integration.channel.config;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.function.Predicate;
 
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.Test;
 
 import org.springframework.beans.DirectFieldAccessor;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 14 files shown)


spring-integration-core/src/test/java/org/springframework/integration/channel/config/DispatchingChannelParserTests.java
```diff
@@ -37,18 +37,20 @@
...
  * @author Mark Fisher
  * @author Oleg Zhurakousky
+ * @author Artem Bilan
+ *
  * @since 1.0.3
  */
-@RunWith(SpringJUnit4ClassRunner.class)
-@ContextConfiguration
+@SpringJUnitConfig
 public class DispatchingChannelParserTests {
```
sobychacko, line 54: One general comment: since you are modifying a test, why don't we clean this up as JUnit 5 style tests? i.e - keeping the test class and the test methods with default access?

spring-integration-core/src/test/java/org/springframework/integration/dispatcher/FailOverDispatcherTests.java
```diff
@@ -188,15 +166,26 @@ public void allHandlersReturnFalse() {
...
-		catch (Exception e) {
-			// ignore
-		}
+		assertThatExceptionOfType(AggregateMessageDeliveryException.class)
+				.isThrownBy(() -> dispatcher.dispatch(new GenericMessage<>("test")));
 		assertThat(counter.get()).as("each target should have been invoked").isEqualTo(3);
 	}
 
+	@Test
+	public void failoverStrategyRejects() {
```
sobychacko, line 175: Junit 5 style tests?

sobychacko (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/channel/config/DispatchingChannelParserTests.java
artembilan: This is not enforced style, so I leave it up personal preferences. Since this is an existing class, I wouldn’t bother removing those `public` words. 
The new test is just a copy/paste artifact , so still no care 😅

sobychacko (approved):

PR #9046, opened 2024-03-22
I11 spring integration nats

Pratiyush:


artembilan:
@Pratiyush ,

Can we understand, please, why have you closed this PR immediately after opening it?
The feature looks promising, just let's follow with standard review procedure and so on!

Thanks

Pratiyush:
@artembilan We are working on it and would create a PR in couple of weeks.

PR #9053, opened 2024-03-30
GH-3444: Add Custom TTL support for RedisLock, and JdbcLock

### LINKED ISSUE #3444: Custom TTL per LOCK in LockRegistry
Current default (and only) implementation of `LockRepository` based on JDBC (`DefaultLockRepository`) allows to set static TTL for all locks, controlled by this repository.

But not all locks are equal. Some locks require bigger TTL than others. It would be good to support custom TTL per lock.

From DB point of view, I think this is easy task. Currently DB stores `CREATED_DATE` field. If new `EXPIRATION_DATE` field will be added to the table, it would be easy to expire records based on this date instead of date of creation + TTL.

From repository point of view it is not so easy. It requires to change the interface or create a new interface that extends current `LockRepository` and adds new method:

```java
boolean acquire(String lock, Duration ttl)
```

From registry point of view it is not so easy too. The only method of `LockRegistry` allows to obtain a lock only using `lockKey` (`Lock obtain(Object lockKey)`). But to support custom TTL per Lock we need to add another method, like:

```java
Lock obtain(Object lockKey, Duration ttl)
```

that would set an explicit TTL on `JdbcLock`. And `JdbcLock` then will pass this TTL into it's `LockRepository` from `doLock()` method.

So, for JDBC implementation it looks totally possible to implement.
But I didn't investigate other implementations of `LockRepository` and `LockRegistry` (not based on JDBC). I'm not familiar with Redis or Zookeeper, so it would be good if someone first evaluate the idea of this feature request.

**Context**

Not all business processes are the same. Some usually finish in 2 secs. Other in 20 mins. So, having a static TTL for all locks of repository is impractical. Setting it to intentionally big value is impractical too, as this means that fast business process, that is killed before lock release, will not be restarted until this big TTL ends.

The workaround to this problem is to use multiple instances of `LockRepository` (and wrapping `LockRegistry`) with multiple different `INT_LOCK` tables. In this case each instance of `LockRepository` can be configured with explicit TTL. This is only partial workaround as it's hard to predict all possible TTLs for each possible business process. But it cloud be generalized to, say, 3 tables (with TTL=10 secs, 10 mins and 10 hours).

The other workaround, I think, is to periodically call `lock()` methods again on currently held lock. As I see in the code, it should update the `CREATED_DATE` field in the table to current datetime, so expiration date will be prolonged too. I actually think, this is the best approach, as it allows to not set too long TTL and at the same time to not mistakenly acquire the lock, that other process still holds.

But it's not always easy to implement this in business logic. For example, if some external service is called with big timeout (and if timeout must be big for this service for some reason) and this external service usually responds in 2 secs, but sometimes in 2 mins. In this case, calling thread is blocked until external service respond or timeout is reached, so no `lock()` method could be called again in this timeframe.

EddieChoCho:
Fixes: #3444

* Add `DistributedLock` interface which is implemented by `RedisLock` and `JdbcLock`.
* Modify `LockRegistry`, `ExpirableLockRegistry`, `RenewableLockRegistry` interfaces.
* Modify implementation of `DefaultLockRepository, `JdbcLockRegistry`, `RedisLockRegistry`
* Modify ddl of `INT_LOCK` table.
* Maintain test cases and documents.


spring-integration-core/src/main/java/org/springframework/integration/support/locks/CustomTtlLock.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.concurrent.locks.Lock;
+
+/**
+ * A {@link Lock} implementing this interface supports the spring distributed locks with custom time-to-live value per lock
+ *
+ * @author Eddie Cho
+ *
+ * @since 6.3
+ */
+public interface CustomTtlLock extends Lock {
```
EddieChoCho, line 29: In this PR, the `CustomTtlLock` interface is only implemented by `RedisLock` and `JdbckLock`.
It can be renamed to `DistributedLock` in the future if all the implementations of Spring distributed lock have implemented the interface.

artembilan (changes requested):
Thanks for looking into this!

I still think it has to be called a `DistributedLock` from the beginning: we may come up eventually with other API, not only TTL.

I think the `LockRegistry` must be modified for a generic arg: `LockRegistry<L extends Lock>`.
And no need in the extra `CustomTtlLockRegistry` abstraction.

Can you elaborate, please, why do we need a result from `delete()`?

Since this is a huge braking change (as it is expected according to the discussion in the issue), we are going to look into this in details in the next `6.4` version, which we are going to start this June.

EddieChoCho:
> Can you elaborate, please, why do we need a result from `delete()`?

If the result(row count) is 0, which means that the lock is not owned by current process.

For example:
ProcessA and processB are in the same region. 
ProcessA acquired the distributed lock. //{REGION: r, LOCK_KEY: lock, CLIENT_ID: processA}
But the work takes too long and the ownership of the lock has been expired.
Before processA release the lock, processB acquired the same distributed lock.  //{REGION: r, LOCK_KEY: lock, CLIENT_ID: processB}
When processA try to release the distributed lock. //DELETE FROM INT_LOCK WHERE REGION='r' AND LOCK_KEY='lock' AND CLIENT_ID='processA'
The result from `delete()` would be 0.

In this case, an IllegalStateException should be thrown to info the process that the integrity of data protected by this lock may have been compromised.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -305,12 +351,21 @@ public void unlock() {
...
+						else {
+							throw new IllegalStateException();
+							// the lock is no longer owned by current process, the exception should be handle and rollback the execution result
+						}
 					}
 					catch (TransientDataAccessException | TransactionTimedOutException | TransactionSystemException e) {
 						// try again
 					}
+					catch (IllegalStateException e) {
+						throw new IllegalStateException("Lock was released in the store due to expiration. " +
```
artembilan, line 366: OK. I see why you use `boolean` for `delete()` now.
I wonder if we can use a `ConcurrentModificationException` instead.
Or something from the `org.springframework.dao` package, like `ConcurrencyFailureException`

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: I see. Well, that was just a suggestion if that is makes sense to rely on some specific exception instead of this common.
Might be the fact that `ConcurrentModificationException` does not fit in our locks scenario.
However that `ConcurrencyFailureException` might be OK.

EddieChoCho:
```
catch (TransientDataAccessException | TransactionTimedOutException | TransactionSystemException e) {
      // try again
}
```
In the current implementation, `TransientDataAccessException`, `TransactionTimedOutException`, and `TransactionSystemException` will be caught, and then the `delete()` method will be invoked again.

Since `ConcurrencyFailureException` extends `TransientDataAccessException`, I think we should not throw it for this scenario.

I will change it to throw `ConcurrentModificationException` instead if you think `ConcurrentModificationException` fits better.

artembilan: I see. Thanks. Missed that.

So, yeah, this `ConcurrentModificationException` fits in my brain because its name means exactly what happens here for us.
But I'm not sure if this is a correct scenario where we can use it from Java perspective.
That's why I'm reaching you for other opinion, arguments and so on.
Just want to be sure that we are on the same page with a decision to move on.

EddieChoCho: I see, 
I agree that changing from `IllegalStateException` to `ConcurrentModificationException` can improve the readability.
But I'm also not sure if this is a correct scenario where we can use it from Java perspective.

How about we create our own exception -`LockOwnershipExpiredException` for Spring distributed locks?
It can be used by both `RedisLock`, and `JdbcLock`.

artembilan:
This is good stuff. 
Thank you very much for looking into that!

As I expected this is really huge breaking change in the API, so I'm afraid we cannot accept it right now according to our migration policies.
We have to wait until the next major version `7.0` next year.
Of course we might be able to start it in the end of this year, when Spring Framework `6.2` release train is out this November.

Please, keep in touch meanwhile.

Thank you for understanding!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
```diff
@@ -389,23 +375,23 @@ public void close() {
 	}
 
 	@Override
-	public void delete(String lock) {
-		this.defaultTransactionTemplate.executeWithoutResult(
-				transactionStatus -> this.template.update(this.deleteQuery, this.region, lock, this.id));
+	public boolean delete(String lock) {
```
Pavankn18, line 378: This also presents an opportunity to check for `template.update() > 1` to throw exception, this will help prevent incorrect constraint on Table. 

This is possible only if table is created without `constraint INT_LOCK_PK primary key (LOCK_KEY, REGION)`.

Pavankn18:
Any new on this? Spring 7 will be out soon right? Will these changes be included 7.0?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
EddieChoCho: This is actually outside the scope of the PR. (my bad, I should not include this change). 
The method has been maintained in PR #9292. 
I think business logic should not address cases of misconfiguration or incorrect settings. 
If you'd like to discuss this method further, it would be great to create a new issue.

artembilan:
Thank you both for keeping an eye on this! So, in November we are releasing `6.4`. After that we are good to start `7.0`. But that is going to be released only next November 2025. So, not soon enough, but we still don’t have a choice with such a breaking change.

artembilan:
We are now at `7.0` development timeline.
I wonder if it is still OK for you to reinstate the work over here.
Thanks

EddieChoCho:
I have rebased the branch, here is the summary of the changes:
* `RedisLock` and `JdbcLock` now implement `DistributedLock` interface to support the feature of customized time-to-live for the lock status data.
* `RenewableLockRegistry` now provides new `renewLock(Object lockKey, long customTtl, TimeUnit customTtlTimeUnit)` method, allowing user to renew the lock with a custom time-to-live value.
* The APIs of `LockRepository` and `DefaultLockRepository` are also modified to support the feature.
* The `CREATED_DATE` column in the `INT_LOCK` table has been replaced with `EXPIRED_AFTER` to support the feature.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -341,12 +390,16 @@ public boolean isAcquiredInThisProcess() {
 		}
 
 		public boolean renew() {
+			return renew(JdbcLockRegistry.this.ttl);
+		}
+
+		public boolean renew(Duration ttl) {
```
cppwfs, line 400: Nitpick.   
Could you please add a javadoc here.   Thank you!

src/reference/antora/modules/ROOT/pages/jdbc/lock-registry.adoc
```diff
@@ -58,4 +58,12 @@ lockRepository.setInsertQuery(lockRepository.getInsertQuery() + " ON CONFLICT DO
...
-And the `JdbcLockRegistry.JdbcLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
\ No newline at end of file
+And the `JdbcLockRegistry.JdbcLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
+
+Starting with version 7.0, the `JdbcLock` implements `DistributedLock` interface to support the feature of customized time-to-live (TTL) for the lock status data.
+The `JdbcLockRegistry` now provides new `renewLock()` method, allowing you to renew the lock with a custom time-to-live value.
+The APIs of `LockRepository` and `DefaultLockRepository` are also modified to support the feature.
+[IMPORTANT]
+====
+The `CREATED_DATE` column in the `INT_LOCK` table has been replaced with `EXPIRED_AFTER` to support the custom time-to-live feature.
```
cppwfs: Need to have a note that states that the users will have to update their table definitions, if they are already using this feature.

cppwfs (changes requested):

Only had a couple of small requests.

src/reference/antora/modules/ROOT/pages/jdbc/lock-registry.adoc
artembilan: I wonder if `CREATED_DATE` column could still be useful...

spring-integration-core/src/main/java/org/springframework/integration/support/locks/DistributedLock.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.locks;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+
+/**
+ * A {@link Lock} implementing for spring distributed locks
```
artembilan: How about just to say:
```
A distributed {@link Lock} extension.
```
No need to mention `Spring` especially not from capital latter.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/DistributedLock.java
```diff
@@ -0,0 +1,53 @@
...
+ * @since 7.0
+ */
+public interface DistributedLock extends Lock {
+
+	/**
+	 * Attempt to acquire a lock with a specific time-to-live
+	 * @param customTtl the specific time-to-live for the lock status data
+	 * @param customTtlTimeUnit the time unit of the {@code customTtl} argument
+	 */
+	void lock(long customTtl, TimeUnit customTtlTimeUnit);
```
artembilan: Can we use a `Duration` API instead?

spring-integration-core/src/main/java/org/springframework/integration/support/locks/DistributedLock.java
```diff
@@ -0,0 +1,53 @@
...
+	 * @param customTtl the specific time-to-live for the lock status data
+	 * @param customTtlTimeUnit the time unit of the customTtl argument
+	 * @return {@code true} if the lock was acquired and {@code false}
+	 *         if the waiting time elapsed before the lock was acquired
+	 *
+	 * @throws InterruptedException if the current thread is interrupted
+	 *         while acquiring the lock (and interruption of lock
+	 *         acquisition is supported)
+	 */
+	boolean tryLock(long time, TimeUnit unit, long customTtl, TimeUnit customTtlTimeUnit) throws InterruptedException;
```
artembilan: Especially here where we have so many method arguments...
I mean `Duration` instead.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/RenewableLockRegistry.java
```diff
@@ -37,6 +41,16 @@ public interface RenewableLockRegistry extends LockRegistry {
...
 
+	/**
+	 * Renew the time to live of the lock is associated with the parameter object with a specific value.
+	 * The lock must be held by the current thread
+	 * @param lockKey The object with which the lock is associated.
+	 * @param customTtl the specific time-to-live for the lock status data
+	 * @param customTtlTimeUnit the time unit of the {@code customTtl} argument
+	 *
+	 */
+	void renewLock(Object lockKey, long customTtl, TimeUnit customTtlTimeUnit);
```
artembilan: Same here about `Duration`.
I know `tryLock(long time, TimeUnit unit)` is there in `Lock` API, but that one has been done long time ago before `java.time` API.
Make sense?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
```diff
@@ -189,14 +182,6 @@ public void setPrefix(String prefix) {
 		this.prefix = prefix;
 	}
 
-	/**
-	 * Specify the time (in milliseconds) to expire deadlocks.
-	 * @param timeToLive the time to expire deadlocks.
-	 */
-	public void setTimeToLive(int timeToLive) {
```
artembilan, line 196: What is a justification of the removal of this public API?
If we are really strong enough then let's deprecate it first!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -85,12 +86,25 @@ protected boolean removeEldestEntry(Entry<String, JdbcLock> eldest) {
...
 	/**
 	 * Construct an instance based on the provided {@link LockRepository}.
 	 * @param client the {@link LockRepository} to rely on.
 	 */
 	public JdbcLockRegistry(LockRepository client) {
 		this.client = client;
+		this.ttl = DEFAULT_TTL;
+	}
+
+	public JdbcLockRegistry(LockRepository client, long expireAfter) {
```
artembilan: Let's see if we can rely on `Duration` instead!

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-db2.sql
```diff
@@ -30,7 +30,7 @@ CREATE TABLE INT_LOCK  (
 	LOCK_KEY CHAR(36) NOT NULL,
 	REGION VARCHAR(100) NOT NULL,
 	CLIENT_ID CHAR(36),
-	CREATED_DATE TIMESTAMP NOT NULL,
+    EXPIRED_AFTER TIMESTAMP NOT NULL,
```
artembilan: Something is off with indents.
Has to be tab.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/lock/DefaultLockRepositoryTests.java
```diff
@@ -41,27 +45,33 @@
...
+	void testNewTransactionIsStartedWhenTransactionIsAlreadyActive() {
 		// Make sure a transaction is active
 		assertThat(TransactionSynchronizationManager.isActualTransactionActive()).isTrue();
 
 		TransactionSynchronization transactionSynchronization = spy(TransactionSynchronization.class);
 		TransactionSynchronizationManager.registerSynchronization(transactionSynchronization);
 
-		this.client.acquire("foo"); // 1
-		this.client.renew("foo"); // 2
+		this.client.acquire("foo", Duration.ofMillis(10000)); // 1
```
artembilan: Feels like `Duration.ofSeconds(10)`.
Much easier to read and less typing.
No? 😉

src/reference/antora/modules/ROOT/pages/jdbc/lock-registry.adoc
```diff
@@ -58,4 +58,12 @@ lockRepository.setInsertQuery(lockRepository.getInsertQuery() + " ON CONFLICT DO
 ----
 
 Starting with version 6.4, the `LockRepository.delete()` method return the result of removing ownership of a distributed lock.
-And the `JdbcLockRegistry.JdbcLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
\ No newline at end of file
+And the `JdbcLockRegistry.JdbcLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
+
+Starting with version 7.0, the `JdbcLock` implements `DistributedLock` interface to support the feature of customized time-to-live (TTL) for the lock status data.
```
artembilan, line 63: I think it is also worth mentioning a new convenient API from that `DistributedLock` abstraction.
And give some example from how you get one from the registry and use  a new API to lock with some TTL.

artembilan (changes requested):
This is really great!
Please, take a look into `distributed-locks.adoc` to mention this our new `DistributedLock` abstraction and fix code snippet for a new style: `LockRegistry<Lock>` and so.

Thank you!

src/reference/antora/modules/ROOT/pages/jdbc/lock-registry.adoc
cppwfs: I think so as well.   i.e. how long has this lock been around (Not just renewed).

EddieChoCho: If we keep the `CREATED_DATE` column, it now provides information about the time the lock was acquired, rather than the TTL. 
Then we need to consider these two cases:
1. A process acquires same lock multiple times before it releases the lock.
    In this case, once the value of `CREATED_DATE` column has been set during the first acquisition and should not be updated by subsequent acquisitions before the lock is released.
2. The TTL of a lock is expires and the lock is now acquired by another process.
   In this case, sine the expired lock's data still existed in the table, we should not only update the `CLIENT_ID` but also the value of `CREATED_DATE` column.

I will modify the implementation of `acquire()` to cover this scenario.

artembilan: I think this one is typical for any RDBD data: when it was inserted. And we really don’t update it on subsequent acquisitions.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
EddieChoCho: I moved the default TTL constant from `DefaultLockRepository` to `JdbcLockRegistry`.
The TTL value can now be set through the constructor of `JdbcLockRegistry`, which aligns with the design of `RedisLockRegistry`.
As a result, I removed the `setTimeToLive()` method from `DefaultLockRepository`.

However, should I also keep a default TTL in `DefaultLockRepository` to support developers who use `DefaultLockRepository` directly without using `JdbcLockRegistry`?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
EddieChoCho: Let's also update the constructor of `RedisLockRegistry` from
`public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, long expireAfter)`
to
`public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, Duration expireAfter)`
to align with the API design.

Do you agree?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
artembilan: That's the point.
The `DefaultLockRepository` really could be used (not sure why) out of `JdbcLockRegistry`.
This one is a public API, so as less distraction as possible.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: Sure! We can do that, but as a new constructor and deprecation of existing one.
Remember, we develop here a library, therefore compatibility in API should be honored as much as possible.
Since we can still keep an old constructor here with simple adaptation, no reason to introduce another breaking change.

EddieChoCho:
I just pushed another commit, please review

spring-integration-core/src/main/java/org/springframework/integration/support/locks/DistributedLock.java
```diff
@@ -0,0 +1,52 @@
...
+public interface DistributedLock extends Lock {
+
+	/**
+	 * Attempt to acquire a lock with a specific time-to-live
+	 * @param ttl the specific time-to-live for the lock status data
+	 */
+	void lock(Duration ttl);
+
+	/**
+	 * Acquires the lock with a specific time-to-live if it is free within the
```
artembilan: Method JavaDocs have to be imperative, like a command: do this, do that.
And no blank lines in method JavaDocs, please.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/DistributedLock.java
```diff
@@ -0,0 +1,52 @@
...
+	 * @param unit the time unit of the {@code time} argument
+	 * @param ttl the specific time-to-live for the lock status data
+	 * @return {@code true} if the lock was acquired and {@code false}
+	 *         if the waiting time elapsed before the lock was acquired
+	 *
+	 * @throws InterruptedException if the current thread is interrupted
+	 *         while acquiring the lock (and interruption of lock
+	 *         acquisition is supported)
+	 */
+	boolean tryLock(long time, TimeUnit unit, Duration ttl) throws InterruptedException;
```
artembilan: Why the first arg is not a `Duration`?
We always can covert it back to millis internally.
This is a new API, so let's make it consistent with end-user expectations!

spring-integration-core/src/main/java/org/springframework/integration/support/locks/RenewableLockRegistry.java
```diff
@@ -37,6 +41,15 @@ public interface RenewableLockRegistry extends LockRegistry {
...
 	void renewLock(Object lockKey);
 
+	/**
+	 * Renew the time to live of the lock is associated with the parameter object with a specific value.
+	 * The lock must be held by the current thread
+	 * @param lockKey The object with which the lock is associated.
+	 * @param ttl the specific time-to-live for the lock status data
+	 *
+	 */
+	void renewLock(Object lockKey, Duration ttl);
```
artembilan, line 52: Since you have added a new API to the class, please, add your name to the `@author` list.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -85,16 +86,29 @@ protected boolean removeEldestEntry(Entry<String, JdbcLock> eldest) {
...
 	/**
 	 * Construct an instance based on the provided {@link LockRepository}.
 	 * @param client the {@link LockRepository} to rely on.
 	 */
 	public JdbcLockRegistry(LockRepository client) {
 		this.client = client;
+		this.ttl = DEFAULT_TTL;
+	}
+
+	public JdbcLockRegistry(LockRepository client, Duration expireAfter) {
```
artembilan, line 112: New public API, therefore `@since 7.0` and some simple JavaDoc.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -85,16 +86,29 @@ protected boolean removeEldestEntry(Entry<String, JdbcLock> eldest) {
 
 	private int cacheCapacity = DEFAULT_CAPACITY;
 
+	/**
+	 * Default value for the time-to-live property.
```
artembilan, line 90: `@since 7.0` here as well, please

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -184,6 +185,16 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
...
 	 */
 	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, long expireAfter) {
+		this(connectionFactory, registryKey, Duration.ofMillis(expireAfter));
+	}
+
+	/**
+	 * Constructs a lock registry with the supplied lock expiration.
+	 * @param connectionFactory The connection factory.
+	 * @param registryKey The key prefix for locks.
+	 * @param expireAfter The expiration in {@link Duration}.
```
artembilan, line 195: `@since 7.0`

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -184,6 +185,16 @@ public RedisLockRegistry(RedisConnectionFactory connectionFactory, String regist
 	 * @param expireAfter The expiration in milliseconds.
 	 */
 	public RedisLockRegistry(RedisConnectionFactory connectionFactory, String registryKey, long expireAfter) {
+		this(connectionFactory, registryKey, Duration.ofMillis(expireAfter));
+	}
+
+	/**
+	 * Constructs a lock registry with the supplied lock expiration.
```
artembilan: Imperative: 
```
Create a lock registry with the supplied lock expiration.
```

src/reference/antora/modules/ROOT/pages/jdbc/lock-registry.adoc
```diff
@@ -58,4 +58,19 @@ lockRepository.setInsertQuery(lockRepository.getInsertQuery() + " ON CONFLICT DO
...
+The `JdbcLockRegistry` now provides new `renewLock(Object lockKey, Duration ttl)` method, allowing you to renew the lock with a custom time-to-live value.
+The APIs of `LockRepository` and `DefaultLockRepository` are also modified to support the feature.
+[IMPORTANT]
+====
+If you're already using an earlier version of `JdbcLockRegistry` or `DefaultLockRepository`, please execute the necessary DDL to modify the `INT_LOCK` table before upgrading to this version.
+
+Here is an example of the Postgres DDL for adding the new column to the lock table:
+[source,sql]
+----
+ALTER TABLE INT_LOCK ADD EXPIRED_AFTER TIMESTAMP NOT NULL;
```
artembilan, line 75: This is great tip!
We will add it into a Migration Guide: https://github.com/spring-projects/spring-integration/wiki

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -23,4 +23,16 @@ Junit 4 Based Support Components are deprecated.
 == New Components
 
 The JDBC module now provides a Java DSL API via its dedicated `org.springframework.integration.jdbc.dsl.Jdbc` factory.
-The xref:jdbc/dsl.adoc[] chapter provides more details.
\ No newline at end of file
+The xref:jdbc/dsl.adoc[] chapter provides more details.
```
artembilan, line 26: Please, add a sentence in this `New Components` section about a `DistributedLock` API and a link to respective `distributed-locks.adoc` chapter.

artembilan (changes requested):
I don't see too many problem now.
Thank you for the great update!
After the next round of changes according to our review, we are going to look into merging this.

EddieChoCho:
Thanks for your feedbacks, I pushed another commit.

cppwfs:
LGTM!    Thank you for your contribution!!!!

cppwfs:
Squashed, Rebased, Merged

Thank you for the hard work on this contribution!

PR #9062, opened 2024-04-01
fix: re-new connection in PostgresChannelMessageTableSubscriber

### LINKED ISSUE #9061: PostgresChannelMessageTableSubscriber not properly handling Database Failover

6.x

**Describe the bug**

When the database is doing a failover the `PostgresChannelMessageTableSubscriber` doesn't notice and will never receive new notifications.
For example on AWS RDS instance with multi-AZ standby, when performing a failover the IP address for the DNS does change. But `PostgresChannelMessageTableSubscriber` never re-establishes a connection, nor has some other measurement in place to validate that the connection is still valid.

**To Reproduce**

Perform a DB failover. After this no more messages will be processed.

**Expected behavior**

Eventually the `PostgresChannelMessageTableSubscriber` obtains a new connection after failover, and message processing will work again.

**Sample**

Sorry, not being able to provide a complete sample, as this involves infrastructure 🤷‍♂️

joshiste:
PostgresChannelMessageTableSubscriber currently never renews the connection. This causes problems on DB failover. With this change the connection is renewed when notifications are not received for a certain time.

fixes #9061

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
@@ -18,6 +18,7 @@ package org.springframework.integration.jdbc.channel;
 
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.time.Duration;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -62,6 +63,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Igor Lovich
  * @author Christian Tzolov
+ * @author Johannes Edmeier
  *
  * @since 6.0
  */
@@ -85,6 +87,8 @@ public final class PostgresChannelMessageTableSubscriber implements SmartLifecyc
 
 	@Nullable
 	private volatile PgConnection connection;
+
+	private Duration notificationTimeout = Duration.ofSeconds(60);
 
 	/**
 	 * Create a new subscriber using the {@link JdbcChannelMessageStore#DEFAULT_TABLE_PREFIX}.
@@ -114,6 +118,19 @@ public final class PostgresChannelMessageTableSubscriber implements SmartLifecyc
 	public void setTaskExecutor(AsyncTaskExecutor taskExecutor) {
 		Assert.notNull(taskExecutor, "A 'taskExecutor' must not be null.");
 		this.taskExecutor = taskExecutor;
+	}
+
+	/**
+	 * Set the timeout for the notification polling.
+	 * If for the specified duration no notificiation are received the underlying connection is closed and re-established.
+	 * Setting a value of {@code Duration.ZERO} will disable the timeout and wait forever.
+	 * This might cause problems in DB failover scenarios.
+	 * @param notificationTimeout the timeout for the notification polling.
+	 * @since 6.1.8
+	 */
+	public void setNotificationTimeout(Duration notificationTimeout) {
+		Assert.notNull(notificationTimeout, "'notificationTimeout' must not be null.");
+		this.notificationTimeout = notificationTimeout;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -278,6 +285,19 @@ private static String getKey(Object input) {
 		return input == null ? null : UUIDConverter.getUUID(input).toString();
 	}
 
+	/**
+	 * Set the timeout for the notification polling.
+	 * If for the specified duration no notificiation are received the underlying connection is closed and re-established.
+	 * Setting a value of {@code Duration.ZERO} will disable the timeout and wait forever.
+	 * This might cause problems in DB failover scenarios.
+	 *
```
artembilan, line 293: No blank lines in the method Javadocs.
Please, move this new setter to the setters sections, which usually is after constructor.
Also add your name to the `@author` list.

artembilan (changes requested):
I wonder if we can simulate the reconnection somehow in the unit test.
See `PostgresChannelMessageTableSubscriberTests`.

Also note: the version `6.0.x` is out of Open Source Support.
So, we accept the fix and back-port it down to `6.1.x` only.

Thanks

joshiste:
> I wonder if we can simulate the reconnection somehow in the unit test.
See PostgresChannelMessageTableSubscriberTests.

I've added a test that verifies for reconnection after certain time without notifications

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -261,6 +263,26 @@ void testRetryOnErrorDuringDispatch(boolean transactionsEnabled) throws Interrup
 		assertThat(payloads).containsExactly("1");
 	}
 
+	@Test
+	public void testRenewConnection() throws Exception {
+		CountDownLatch latch = new CountDownLatch(2);
+		List<Object> payloads = new ArrayList<>();
+		CountDownLatch connectionLatch = new CountDownLatch(2);
+		connectionSupplier.onGetConnection = connectionLatch::countDown;
```
artembilan, line 271: We need to set it to `null` in the end of test.
There is just no guarantee in test method executions order.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -300,4 +322,19 @@ public JdbcChannelMessageStore jdbcChannelMessageStore(DataSource dataSource) {
 
 	}
 
+	private static class ConnectionSupplier implements PgConnectionSupplier {
+		Runnable onGetConnection;
```
artembilan, line 326: Every member in the class must be surrounded with blank lines.
Including the last method in the class. 

Run `gradlew :spring-integration-jdbc:check`  locally.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -116,6 +120,18 @@ public void setTaskExecutor(AsyncTaskExecutor taskExecutor) {
 		this.taskExecutor = taskExecutor;
 	}
 
+	/**
+	 * Set the timeout for the notification polling.
+	 * If for the specified duration no notificiation are received the underlying connection is closed and re-established.
+	 * Setting a value of {@code Duration.ZERO} will disable the timeout and wait forever.
+	 * This might cause problems in DB failover scenarios.
+	 * @param notificationTimeout the timeout for the notification polling.
```
artembilan, line 128: `@since 6.1.8`

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
joshiste: fixed

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriberTests.java
joshiste: It is initialized in the `BeforeEach` there is no need in unsetting,  as each test will have a new instance.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriberTests.java
joshiste: fixed

artembilan (changes requested):
Something is off with a new test:
```
PostgresChannelMessageTableSubscriberTests > testRenewConnection() FAILED
    org.opentest4j.AssertionFailedError at PostgresChannelMessageTableSubscriberTests.java:282
```

artembilan:
... and cherry-picked to `6.2.x` & `6.1.x`.

joshiste:
What is the schedule for the next release?

artembilan:
April, 16

PR #9224, opened 2024-06-10
Be able to set SmartLifeCycle Phase on PostgresChannelMessageTableSubscriber

ansgarschulte:
There is actually no way to modify the SmartLifeCycle Phase for the PostgresChannelMessageTableSubscriber but we need this.

Could please also backport the change to 6.2.x?


artembilan:
I wonder what is the use-case like that.
The default phase of `Integer.MAX_VALUE` means to start this component as late as possible.
This way we would be able to initialize the rest of our integration before we start polling data from DB in this component.
So, the real question is why would one change the phase for this component to something what would lead to unexpected behavior because of race condition triggered by this phase modification?
Thanks

ansgarschulte:
It is about the shutdown.
We have some processes running during the shutdown that rely on this component.
And the smartlifecycle shuts down in reverse order. So Integer.MAX_VALUE will be the first. 
And there would be no way to do something before that.

ansgarschulte:
But. We discussed that again internally. And you are right.
Thank you.

PR #9229, opened 2024-06-11
GH-9228: ZeroMQ - provide binding to ZeroMqMessageHandler

### LINKED ISSUE #9228: ZeroMQ: ZeroMQMessageHandler with bind capability
**Expected Behavior**

The expected behaviour is that the ZeroMqMessageHandler could bind a port when the other part (a ZeroMqMessageProducer or whatever out of Spring world) connects to that port.

**Current Behavior**

As for now, the ZeroMqMessageHandler can only connect to a specific URL, while [no binding is allowed](https://github.com/spring-projects/spring-integration/blob/afb87d45e82cd6ad3ea02a87758701310b08e473/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/outbound/ZeroMqMessageHandler.java#L53).

**Context**

As per ZeroMQ philosophy, binding or connecting to a socket is completely decoupled from sender/receiver abstraction.

My question is: Is there any reason behind the current limitation? Something I'm missing out?
If you agree, I could make a PR to add this functionality

alessiomatricardi:
Closes #9228 

This PR provides the `bind` capability to the outbound channel adapter `ZeroMqMessageHandler`.

> Consider to implement that logic in the Lifecycle.start()

I tried, but the current implementation instantiate the socket within the class constructor, and some old tests relies on this mechanism.

### DIFF
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
@@ -23,17 +23,14 @@ import org.zeromq.ZMQ;
  *
  * @author Alessio Matricardi
  *
- * @since 6.4
  */
 public final class ZeroMqUtils {
 
 	/**
-	 * Binds the ZeroMq socket to the given port when different from 0
-	 * or to a random port if not.
+	 * Bind the ZeroMq socket to the given port over the TCP transport protocol.
 	 * @param socket the ZeroMq socket
-	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @param port the port to bind ZeroMq socket to over TCP. If equal to 0, the socket will bind to a random port.
 	 * @return the effectively bound port
-	 * @since 6.4
 	 */
 	public static int bindSocket(ZMQ.Socket socket, int port) {
 		if (port == 0) {
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
@@ -23,6 +23,8 @@ import org.zeromq.ZMQ;
  *
  * @author Alessio Matricardi
  *
+ * @since 6.4
+ *
  */
 public final class ZeroMqUtils {
 
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2020-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.zeromq;
+
+import org.zeromq.ZMQ;
+
+/**
+ * Module that wraps common methods of ZeroMq integration classes
+ *
+ * @author Alessio Matricardi
+ *
+ * @since 6.4
+ */
+public final class ZeroMqUtils {
+
+	/**
+	 * Binds the ZeroMq socket to the given port when different from 0
+	 * or to a random port if not.
+	 * @param socket the ZeroMq socket
+	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @return the effectively bound port
+	 * @since 6.4
+	 */
+	public static int bindSocket(ZMQ.Socket socket, int port) {
+		if (port == 0) {
+			return socket.bindToRandomPort("tcp://*");
+		}
+		else {
+			boolean bound = socket.bind("tcp://*:" + port);
+			if (!bound) {
+				throw new IllegalArgumentException("Cannot bind ZeroMQ socket to port: " + port);
+			}
+			return port;
+		}
+	}
+
+	private ZeroMqUtils() {
+	}
+
+}
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMq.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMq.java
@@ -25,6 +25,7 @@ import org.zeromq.ZContext;
  * Factory class for ZeroMq components DSL.
  *
  * @author Artem Bilan
+ * @author Alessio Matricardi
  *
  * @since 5.4
  */
@@ -59,6 +60,17 @@ public final class ZeroMq {
 	}
 
 	/**
+	 * Create an instance of {@link ZeroMqMessageHandlerSpec} for the provided {@link ZContext} and binding port.
+	 * @param context the {@link ZContext} to use.
+	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @return the spec.
+	 * @since 6.4
+	 */
+	public static ZeroMqMessageHandlerSpec outboundChannelAdapter(ZContext context, int port) {
+		return new ZeroMqMessageHandlerSpec(context, port);
+	}
+
+	/**
 	 * Create an instance of {@link ZeroMqMessageHandlerSpec} for the provided {@link ZContext}
 	 * and connection URL supplier.
 	 * @param context the {@link ZContext} to use.
@@ -82,6 +94,20 @@ public final class ZeroMq {
 			SocketType socketType) {
 
 		return new ZeroMqMessageHandlerSpec(context, connectUrl, socketType);
+	}
+
+	/**
+	 * Create an instance of {@link ZeroMqMessageHandlerSpec} for the provided {@link ZContext}, binding port
+	 * and {@link SocketType}.
+	 * @param context the {@link ZContext} to use.
+	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @param socketType the {@link SocketType} for ZeroMq socket.
+	 * @return the spec.
+	 * @since 6.4
+	 */
+	public static ZeroMqMessageHandlerSpec outboundChannelAdapter(ZContext context, int port,
+			SocketType socketType) {
+		return new ZeroMqMessageHandlerSpec(context, port, socketType);
 	}
 
 	/**
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMq.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMq.java
@@ -97,6 +97,29 @@ public final class ZeroMq {
 	}
 
 	/**
+	 * Create an instance of {@link ZeroMqMessageHandlerSpec} for the provided {@link ZContext}.
+	 * The created socket will be bound to a random port.
+	 * @param context the {@link ZContext} to use.
+	 * @return the spec.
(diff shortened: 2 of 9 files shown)


artembilan (changes requested):
Please, also add your name to the `@author` list of all the affected classes.
And verify changes locally like `./gradlew :spring-integration-zeromq:check`.

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/outbound/ZeroMqMessageHandler.java
```diff
@@ -122,6 +151,26 @@ public ZeroMqMessageHandler(ZContext context, String connectUrl, SocketType sock
...
+	 * @since 6.2.6
+	 */
+	public ZeroMqMessageHandler(ZContext context, int port, SocketType socketType) {
+		this(context, socketType);
+		this.bindPort.set(port);
+		this.socketMono =
+				Mono.just(this.context.createSocket(this.socketType))
+						.publishOn(this.publisherScheduler)
+						.doOnNext((socket) -> this.socketConfigurer.accept(socket))
+						.doOnNext((socket) -> this.bindPort.set(bindSocket(socket, this.bindPort.get())))
```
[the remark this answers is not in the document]
artembilan: Well, actually we do this in the `start()`: 
```
this.socketMonoSubscriber = this.socketMono.subscribe();
```
So, all good, however looking into this one more time I believe that we have to always renew this `socketMono` in the `start()`. After the `stop()` the environment might be changed, so or now way to connect or bind.
Therefore every time when we start this component we have to renew socket.


I'm also thinking about DRY principle here.
Looks like we now have two socket `Mono` which has difference only in these lines:
```
.doOnNext((socket) -> socket.connect(connectUrl.get()))
```
in opposed to:
```
.doOnNext((socket) -> this.bindPort.set(bindSocket(socket, this.bindPort.get())))
```
Essentially `if..else` according to the server/client preferences.
This could be just extracted into a `private` method where we would do `connect` or `bind`. 
And `monoSocket` would be just a single instance in the `start` without any duplication.

WDYT?

artembilan (changes requested):
Also this new functionality has to be mention in the `whats-new.adoc`. Like a new section in the end:
```
[[x6.4-zeromq-changes]]
=== ZeroMQ Changes

```

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMqMessageHandlerSpec.java
```diff
@@ -52,6 +52,16 @@ protected ZeroMqMessageHandlerSpec(ZContext context, String connectUrl) {
 		this(context, () -> connectUrl);
 	}
 
+	/**
+	 * Create an instance based on the provided {@link ZContext} and binding port.
+	 * @param context the {@link ZContext} to use for creating sockets.
+	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @since 6.2.6
```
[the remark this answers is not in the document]
alessiomatricardi: Done!

src/reference/antora/modules/ROOT/pages/zeromq.adoc
```diff
@@ -146,7 +146,9 @@ ZeroMqMessageProducer zeroMqMessageProducer(ZContext context, MessageChannel out
 
 The `ZeroMqMessageHandler` is a `ReactiveMessageHandler` implementation to produce publish messages into a ZeroMQ socket.
 Only `SocketType.PAIR`, `SocketType.PUSH` and `SocketType.PUB` are supported.
-The `ZeroMqMessageHandler` only supports connecting the ZeroMQ socket; binding is not supported.
+This component can connect to the remote socket or bind onto TCP protocol with the provided or random port.
```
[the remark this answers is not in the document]
alessiomatricardi: The logic now follows the one implemented in ZeroMqMessageProducer: if not provided, a random port will be chosen

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/outbound/ZeroMqMessageHandler.java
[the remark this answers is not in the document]
alessiomatricardi: Completely agree with you. I was trying to touch less code as possible, but as you said this is the best approach

alessiomatricardi:
Addressed almost every comment. I also added a typo in `ZeroMqHeaders`, but I've not added myself as author because no code has been touched. Let me know if I should do it

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
```diff
@@ -0,0 +1,54 @@
...
+	 * Binds the ZeroMq socket to the given port when different from 0
+	 * or to a random port if not.
+	 * @param socket the ZeroMq socket
+	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @return the effectively bound port
+	 * @since 6.4
+	 */
+	public static int bindSocket(ZMQ.Socket socket, int port) {
+		if (port == 0) {
+			return socket.bindToRandomPort("tcp://*");
```
artembilan, line 40: I wonder if it always has to be `tcp` protocol: http://api.zeromq.org/4-2:zmq-bind.

Or at least, let's document that in this method Javadoc!

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/outbound/ZeroMqMessageHandler.java
```diff
@@ -206,6 +246,47 @@ public void wrapTopic(boolean wrapTopic) {
...
+		this.connectUrl = connectUrl;
+	}
+
+	/**
+	 * Configure a port for TCP protocol binding via {@link org.zeromq.ZMQ.Socket#bind(String)}.
+	 * Mutually exclusive with the {@link #setConnectUrl(String)}.
+	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @since 6.4
+	 */
+	public void setBindPort(int port) {
```
artembilan, line 275: We have these options covered by constructors.
I don't see a reason in extra setters on the matter: might cause some paradox of choice.

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/outbound/ZeroMqMessageHandler.java
```diff
@@ -91,6 +94,27 @@ public class ZeroMqMessageHandler extends AbstractReactiveMessageHandler
...
+	private final AtomicInteger bindPort = new AtomicInteger();
+
+	@Nullable
+	private Supplier<String> connectUrl;
+
+	public ZeroMqMessageHandler(ZContext context) {
+		this(context, SocketType.PAIR);
+	}
+
+	public ZeroMqMessageHandler(ZContext context, SocketType socketType) {
```
artembilan, line 110: These two new constructors need some Javadoc.
I guess they are about random port.

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/outbound/ZeroMqMessageHandler.java
```diff
@@ -214,6 +295,8 @@ public String getComponentType() {
 	@Override
 	protected void onInit() {
 		super.onInit();
+		Assert.state(this.connectUrl == null || this.bindPort.get() == 0,
+				"Only one of the 'connectUrl' or 'bindPort' must be provided");
```
artembilan, line 299: If we remove those new setters, this condition won't be necessary.
So, you would only have a choice to give `connectUrl` or `port` via those constructors.

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMq.java
```diff
@@ -58,6 +59,17 @@ public static ZeroMqMessageHandlerSpec outboundChannelAdapter(ZContext context,
...
 	}
 
+	/**
+	 * Create an instance of {@link ZeroMqMessageHandlerSpec} for the provided {@link ZContext} and binding port.
+	 * @param context the {@link ZContext} to use.
+	 * @param port the port to bind ZeroMq socket to over TCP.
+	 * @return the spec.
+	 * @since 6.4
+	 */
+	public static ZeroMqMessageHandlerSpec outboundChannelAdapter(ZContext context, int port) {
```
artembilan, line 69: I wonder if DSL has to expose extra factories to explicitly bind to random port.
Something similar what you have with those new ctors in the `ZeroMqMessageHandler`.

artembilan (changes requested):
This is great!
Just couple follow up suggestions.

Thank you!

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
alessiomatricardi: Fully agree with you.
I only merged the behaviour between `ZeroMqMessageProducer` and `ZeroMqMessageHandler`.

I believe that the support of the other transport protocols is out of the scope of this PR, if you don't mind I will make a specific one for this purpose.
Additionally, some protocols like `inproc` don't need a port but only an arbitrary string endpoint, so some rework is needed.

BTW, I will remark the fact that bind is actually possible only over TCP.

alessiomatricardi:
Addressed your comments!

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.springframework.integration.zeromq;
+
+import org.zeromq.ZMQ;
+
+/**
+ * Module that wraps common methods of ZeroMq integration classes
+ *
+ * @author Alessio Matricardi
+ *
```
artembilan, line 25: Sorry, I might was not right: the `@since` is required for new classes. And after that for new `public` methods added to existing classes.
So, bring back `@since 6.4` over here, please

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/ZeroMqUtils.java
artembilan: Yeah... Let's look into that when concern arises!

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMqMessageHandlerSpec.java
```diff
@@ -52,6 +52,25 @@ protected ZeroMqMessageHandlerSpec(ZContext context, String connectUrl) {
 		this(context, () -> connectUrl);
 	}
 
+	/**
+	 * Create an instance based on the provided {@link ZContext}.
```
artembilan, line 56: Probably better to say that socket is going to be bound to random port.

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMqMessageHandlerSpec.java
```diff
@@ -73,6 +92,29 @@ protected ZeroMqMessageHandlerSpec(ZContext context, String connectUrl, SocketTy
 		this(context, () -> connectUrl, socketType);
 	}
 
+	/**
+	 * Create an instance based on the provided {@link ZContext} and {@link SocketType}.
```
artembilan, line 96: DITTO

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/dsl/ZeroMq.java
artembilan: Or you have missed to push the change, or just didn't implement yet.
Hence the conversion is not resolved yet 😄

artembilan (changes requested):
Nice progress!
I believe we are very close for merge.
Please, take a look to the latest review.
Honestly, nothing critical and I can address them on merge.
Just let me know your preferences!

alessiomatricardi:
Hi, sorry for the delay but I'm very busy with work.. hope now looks good for you :)

artembilan:
@alessiomatricardi ,

Issue #9272, opened 2024-06-25
SFTP: ClientSession associated with SftpSession is never closed. 


darrylsmithUGA:
**In what version(s) of Spring Integration are you seeing this issue?**

6.3.1

**Describe the bug**

ClientSession associated with SftpSession is never closed. If the SFTP server is not setup to timeout these connections eventually no more connections can be made to the SFTP server. 

Also described here: [https://stackoverflow.com/questions/78226325/spring-integration-open-a-new-ssh-connection-after-each-sftpexception](https://stackoverflow.com/questions/78226325/spring-integration-open-a-new-ssh-connection-after-each-sftpexception)

PR #9273, opened 2024-06-25
GH-9272 SFTP: ClientSession associated with SftpSession is never closed

darrylsmithUGA:
Fixes for #9272

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
@@ -161,7 +161,7 @@ public class SftpSession implements Session<SftpClient.DirEntry> {
 
 		try {
 			ClientSession session = this.sftpClient.getSession();
-			if(session != null && session.isOpen()) {
+			if (session != null && session.isOpen()) {
 				session.close();
 			}
 		}
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
@@ -25,6 +25,7 @@ import java.time.Duration;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import org.apache.sshd.client.session.ClientSession;
 import org.apache.sshd.common.util.net.SshdSocketAddress;
 import org.apache.sshd.sftp.SftpModuleProperties;
 import org.apache.sshd.sftp.client.SftpClient;
@@ -47,6 +48,7 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Christian Tzolov
+ * @author Darryl Smith
  * @since 2.0
  */
 public class SftpSession implements Session<SftpClient.DirEntry> {
@@ -155,6 +157,16 @@ public class SftpSession implements Session<SftpClient.DirEntry> {
 		}
 		catch (IOException ex) {
 			throw new UncheckedIOException("failed to close an SFTP client", ex);
+		}
+
+		try {
+			ClientSession session = this.sftpClient.getSession();
+			if(session != null && session.isOpen()) {
+				session.close();
+			}
+		}
+		catch (IOException ex) {
+			throw new UncheckedIOException("failed to close an SFTP client (session)", ex);
 		}
 	}
 
--- a/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
+++ b/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
@@ -249,13 +249,13 @@ public class SftpSessionFactoryTests {
 			SftpSession session = sftpSessionFactory.getSession();
 			ClientSession clientSession = session.getClientInstance().getClientSession();
 
-			assertThat(session.isOpen()).isEqualTo(true);
-			assertThat(clientSession.isOpen()).isEqualTo(true);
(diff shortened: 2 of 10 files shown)


spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
```diff
@@ -156,6 +158,16 @@ public void close() {
 		catch (IOException ex) {
 			throw new UncheckedIOException("failed to close an SFTP client", ex);
 		}
+
+		try {
+			ClientSession session = this.sftpClient.getSession();
+			if (session != null && session.isOpen()) {
```
artembilan, line 164: I don't think this fix is correct.
The `DefaultSftpClient` is based on the externally provided `ClientSession`.
We cannot mutate that one from our volatile `SftpSession`.

Let me double check what is really the problem!

artembilan (changes requested):

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
artembilan: OK. I confirm the problem, but sounds more like we have to fix `DefaultSftpSessionFactory` to keep cache for the `ClientSession` and reuse it for every `ConcurrentSftpClient` created.
If that is closed, we just have renew it.
Then we have to implement `DisposableBean` over there to close this shared `ClientSession` properly.
And apparently we also have to `SshClient.stop()` if it `isInnerClient`.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
```diff
@@ -218,4 +220,30 @@ void customTimeoutIsApplied() throws IOException {
 		}
 	}
 
+	@Test
+	void clientSessionIsClosedOnSessionClose() throws IOException {
```
artembilan, line 224: This test proves nothing.
Please, consider to revise it when you agree with my proposal.

Thanks

artembilan (changes requested):

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
darrylsmithUGA: For the simple use cases for example connecting to a SFTP server every five hours to retrieve files, I don't feel the ```ClientSession``` should remain open after the after the files have been synchronized locally. This fix was an attempt to solve those simple use cases.

How would you recommend handling the simpler use cases?

artembilan: After more thinking, I'll accept your suggestion.
Looks like it works well in combination of `DefaultSftpSessionFactory` and `CachingSessionFactory`.
So, if our session is not valid, it will close all its relevant resources.
The next new session will come with new resources.
If it is valid, then it is just going to be cached with all of its relevant resources.

Confirm, please, that it is what you are intended for.

However I still believe that `isInnerClient` has to be closed in the `destroy()` of the `DefaultSftpSessionFactory`.
Would you mind addressing that as a part of your fix?

darrylsmithUGA: Yes, that's the intent. 

I'll can add stop method call in ```DisposableBean.destory()``` implementation as well.

Thanks

darrylsmithUGA: Adding the ```DisposableBean``` interface to DefaultSftpSessionFactory adds the requirement that objects of this type need to be created be created by the BeanFactory. This is not the case in the test suite and cannot be assumed to be the case in existing applications.

Should I proceed with adding the interface?

Thanks

artembilan: The assumption that components from the framework might be used outside of `BeanFactory` is error-prone.
When we talk about Spring, the component must be a bean for the proper dependency injection lifecycle.
So, yes, please, implement that interface and fix respective tests to call `destroy()` directly, if that unit test is not based on `BeanFactory`.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
```diff
@@ -215,7 +225,39 @@ void customTimeoutIsApplied() throws IOException {
...
+			SftpSession session = sftpSessionFactory.getSession();
+			ClientSession clientSession = session.getClientInstance().getClientSession();
+
+			assertThat(session.isOpen()).isEqualTo(true);
+			assertThat(clientSession.isOpen()).isEqualTo(true);
+
+			session.close();
+
+			assertThat(session.isOpen()).isEqualTo(false);
+			assertThat(clientSession.isClosed()).isEqualTo(true);
```
artembilan, line 258: Nit-pick.
Isn't there something like `isTrue()/isFalse()` convenient assertions?

artembilan (changes requested):
Just one comment for the test code.
Thanks

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpSessionFactoryTests.java
darrylsmithUGA: Yes, there is will push with that change.

Thanks

artembilan:
@darrylsmithUGA ,

Issue #9271, opened 2024-06-25
Refactor to Eliminate Repetitive Mock Object Creation in TcpMessageMapperTests.java


gzhao9:
Hi there!

While working with the `TcpMessageMapperTests` class, I noticed two mock variables repeatedly created across various tests. To simplify the code, I propose a small refactor to eliminate these redundancies, which could reduce the code by 41 lines.

1. **Repetitive Mock TcpConnection Creation**: The creation of a mock `TcpConnection` object repeated in 7 test cases.

2. **Repetitive Mock Socket Creation**: The creation of a mock `Socket` object repeated in 3 test cases.

Specifically, here is the refactoring method I propose:

Creating a mock for `TcpConnection` currently looks like this:

```java
TcpConnection connection = mock(TcpConnection.class);
//other code...
when(connection.getPayload()).thenReturn(TEST_PAYLOAD.getBytes());
when(connection.getHostName()).thenReturn("MyHost");
when(connection.getHostAddress()).thenReturn("1.1.1.1");
when(connection.getPort()).thenReturn(1234);
```

I introduced a method, `CreatMockTcpConnection`:

```java
public TcpConnection CreatMockTcpConnection(Object bytes, String hostName, String ipAdress, int port){
    TcpConnection connection = mock(TcpConnection.class);
    when(connection.getPayload()).thenReturn(bytes);
    when(connection.getHostName()).thenReturn(hostName);
    when(connection.getHostAddress()).thenReturn(ipAdress);
    when(connection.getPort()).thenReturn(port);
    return connection;
}
```

With this method, creating a mock `TcpConnection` becomes:

```java
TcpConnection connection = CreatMockTcpConnection(map, "someHost", "1.1.1.1", 1234);
```

Similarly, for `Socket`, I introduced a method, `creatMockSocket` as follows:

```java
public Socket creatMockSocket(InetAddress local){
    Socket socket = mock(Socket.class);
    when(socket.getLocalAddress()).thenReturn(local);
    return socket;
}
```

This allows us to create mocks on one line as well：

```java
Socket socket = creatMockSocket(local);
```

I created a draft of the PR, you can see the details file changes  [here](https://github.com/gzhao9/spring-integration/pull/1/files).

---

The refactor reduced the test cases by 41 lines of code, and I believe these changes will improve code readability.

artembilan:
Thank you for looking into that!
[...]

PR #9277, opened 2024-06-26
Refactor to Eliminate Repetitive Mock Object Creation in TcpMessageMapperTests.java (for issue #9271)

gzhao9:
### Pull Request Title:
Refactor `TcpMessageMapperTests` to Reduce Redundancy and Simplify Mock Creation

### Pull Request Description:
This pull request addresses issue #9271 by refactoring the `TcpMessageMapperTests` class to eliminate redundant mock creations and simplify the testing code. Specifically, it introduces helper methods to centralize the creation and configuration of mocks for `TcpConnection` and `Socket`, reducing code duplication and enhancing readability.

#### Changes:

**Added Methods**:
  - `createMockTcpConnection(Object bytes, String hostName, String ipAddress, int port)`: Creates and configures a mock `TcpConnection` instance.
  - `createMockSocket(InetAddress local)`: Creates and configures a mock `Socket` instance.

With these helper methods, the test cases are simplified, making the code more maintainable and easier to read. The refactor reduces the test cases by 41 lines of code.

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpMessageMapperTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpMessageMapperTests.java
@@ -74,15 +74,10 @@ public class TcpMessageMapperTests {
 	@Test
 	public void testToMessage() {
 		TcpMessageMapper mapper = new TcpMessageMapper();
-		TcpConnection connection = mock(TcpConnection.class);
-		Socket socket = mock(Socket.class);
+		TcpConnection connection = CreatMockTcpConnection(TEST_PAYLOAD.getBytes(), "MyHost", "1.1.1.1", 1234);
 		InetAddress local = mock(InetAddress.class);
+		Socket socket = creatMockSocket(local);
 		SocketInfo info = new SocketInfo(socket);
-		when(socket.getLocalAddress()).thenReturn(local);
-		when(connection.getPayload()).thenReturn(TEST_PAYLOAD.getBytes());
-		when(connection.getHostName()).thenReturn("MyHost");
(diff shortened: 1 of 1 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpMessageMapperTests.java
```diff
@@ -458,5 +427,19 @@ public void testWithBytesMapper() {
 		assertThat(message.getHeaders().get(IpHeaders.REMOTE_PORT)).isEqualTo(1234);
 		assertThat(message.getHeaders().get(IpHeaders.CONNECTION_ID)).isEqualTo("someId");
 	}
+	public TcpConnection CreatMockTcpConnection(Object bytes, String hostName, String ipAdress, int port){
```
artembilan, line 430: These new methods are not supposed to be `public`. They are concern of the logic just of this class.
So, `private` and `static`.

And our convention is to use camel case for method names.
Not sure from where you took an idea to have it capitalized.
More over it is not consistent since the next method is indeed in camel case...

Please, also run `./gradlew :spring-integration-ip:check` to be sure that you don't break any Checkstyle rules.
In the end, please, add your name to the `@author` list of this class.

artembilan (changes requested):

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpMessageMapperTests.java
gzhao9: Thank you very much for your correction. I have finished these modifications. If there are any problems, please feel free to let me know.

artembilan:
@gzhao9 ,

PR #9282, opened 2024-06-28
Refactor to Eliminate Repetitive Mock Object Creation(more exmple with #9271)

gzhao9:
### Pull Request Description:
This pull request addresses issue #9283 , which is essentially doing the same thing as issue #9271. It refactors the 4 classes to eliminate redundant mock creations and simplify the testing code. Specifically, it introduces helper methods to centralize the creation and configuration of mocks for `AbstractClientConnectionFactory`, and `BeanFactory`, reducing code duplication and enhancing readability.

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/CachingClientConnectionFactoryTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/CachingClientConnectionFactoryTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -88,6 +88,7 @@ import static org.mockito.Mockito.when;
 /**
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Gengwu Zhao
  *
  * @since 2.2
  *
@@ -505,17 +506,13 @@ public class CachingClientConnectionFactoryTests {
 	@Test
 	public void testCachedFailover() throws Exception {
 		// Failover
-		AbstractClientConnectionFactory factory1 = mock(AbstractClientConnectionFactory.class);
-		AbstractClientConnectionFactory factory2 = mock(AbstractClientConnectionFactory.class);
+		TcpConnectionSupport mockConn1 = makeMockConnection();
+		TcpConnectionSupport mockConn2 = makeMockConnection();
+		AbstractClientConnectionFactory factory1 = createFactoryWithMockConnection(mockConn1);
+		AbstractClientConnectionFactory factory2 = createFactoryWithMockConnection(mockConn2);
 		List<AbstractClientConnectionFactory> factories = new ArrayList<>();
 		factories.add(factory1);
 		factories.add(factory2);
-		TcpConnectionSupport mockConn1 = makeMockConnection();
-		TcpConnectionSupport mockConn2 = makeMockConnection();
-		when(factory1.getConnection()).thenReturn(mockConn1);
-		when(factory2.getConnection()).thenReturn(mockConn2);
-		when(factory1.isActive()).thenReturn(true);
-		when(factory2.isActive()).thenReturn(true);
 		doThrow(new UncheckedIOException(new IOException("fail"))).when(mockConn1).send(Mockito.any(Message.class));
 		doAnswer(invocation -> null).when(mockConn2).send(Mockito.any(Message.class));
 		FailoverClientConnectionFactory failoverFactory = new FailoverClientConnectionFactory(factories);
@@ -820,6 +817,13 @@ public class CachingClientConnectionFactoryTests {
 		TcpConnectionSupport connection = mock(TcpConnectionSupport.class);
 		when(connection.isOpen()).thenReturn(true);
 		return connection;
+	}
+
+	private static AbstractClientConnectionFactory createFactoryWithMockConnection(TcpConnectionSupport mockConn) throws Exception {
+		AbstractClientConnectionFactory factory = mock(AbstractClientConnectionFactory.class);
+		when(factory.getConnection()).thenReturn(mockConn);
+		when(factory.isActive()).thenReturn(true);
+		return factory;
 	}
 
 }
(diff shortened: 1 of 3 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/CachingClientConnectionFactoryTests.java
```diff
@@ -505,17 +506,13 @@ private void testCloseOnTimeoutGuts(AbstractClientConnectionFactory cf) throws E
 	@Test
 	public void testCachedFailover() throws Exception {
 		// Failover
-		AbstractClientConnectionFactory factory1 = mock(AbstractClientConnectionFactory.class);
-		AbstractClientConnectionFactory factory2 = mock(AbstractClientConnectionFactory.class);
+		TcpConnectionSupport mockConn1 = makeMockConnection();
+		TcpConnectionSupport mockConn2 = makeMockConnection();
+		AbstractClientConnectionFactory factory1 = FailoverClientConnectionFactoryTests.makeMockactory(mockConn1);
+		AbstractClientConnectionFactory factory2 = FailoverClientConnectionFactoryTests.makeMockactory(mockConn2);
```
artembilan, line 512: First of all there is a typo in the method name.
Secondly I'd prefer to mix test classes with each other.
Even if right now they have similar logic for mock it might be hard in the future to refactor them.
My point is that those tests are essentially closed boxed and it is better to design them without tangling with other test classes.
It might look right now as a readability improvement, but it will be harder to support them individually in the future when the logic one of them might change.

In the end these are just tests, so might not worse pursuing such a refactoring if you cannot find a solution to not tangle test classes with each other.

artembilan (changes requested):
```
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-jms/src/test/java/org/springframework/integration/jms/OutboundGatewayConnectionTests.java:36:8: Unused import - org.springframework.integration.context.IntegrationContextUtils. [UnusedImports]

Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-jms/src/test/java/org/springframework/integration/jms/OutboundGatewayConnectionTests.java:41:8: Unused import - org.springframework.scheduling.TaskScheduler. [UnusedImports]
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-jms/src/test/java/org/springframework/integration/jms/OutboundGatewayConnectionTests.java:45:15: Unused import - org.mockito.Mockito.mock. [UnusedImports]
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-jms/src/test/java/org/springframework/integration/jms/OutboundGatewayConnectionTests.java:46:15: Unused import - org.mockito.Mockito.when. [UnusedImports]
```
Please, consider to run `./gradlew check` locally before pushing to pull request.

artembilan:
FYI, we prefer to use `rebase` instead of `merge` . This way there won’t be extra commits in the PR and noise in mail box. Plus rebase helps to keep linear history. Otherwise it is easy to break what is not yours if there is a conflict .

Will review the rest on Monday.
Thank you update!

artembilan (commented):
```

Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/MockFactory.java:33:17: 'METHOD_DEF' should be separated from previous line. [EmptyLineSeparator]
```
Please, use `gradlew check` command locally before pushing changes to PR.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/MockFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.springframework.integration.ip.tcp.connection;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Gengwu Zhao
+ *
+ * @since 6.3.1
```
artembilan, line 25: This version has been release already.
And we are now in the `6.4` phase.
Since the change you are doing is just plain cosmetic refactoring, we are not going to back-port it into `6.3.x`  branch.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/MockFactory.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+public class MockFactory {
```
artembilan, line 29: The name for the class is too generic to be as a public API.
Plus it really does not reflect what is going on in the class.
The code style in the class does not follow our rules.
Please, use `check` Gradle command locally to catch all the code style problems before pushing.

I commented before that I prefer to not have a mixes between test classes since their logic might diverse in the future.
So, while the change you are doing is OK now and makes less code lines, it can become burden in the future to support such a generic code.

Therefore I'm OK with extracting common code into static methods of the same test class.
This will ensure that change into one test class won't effect others.

Plus it is better to be more specific with names.
This `MockFactory` class name is not about what it does.
The name of the `makeMockFactory()` method is also not fully about what it is doing.

spring-integration-jms/src/test/java/org/springframework/integration/jms/MockBeanFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Gengwu Zhao
+ *
+ * @since 6.3.1
+ *
+ */
+
+public class MockBeanFactory {
```
artembilan, line 34: This name is also wrong for public API.
And it also does not explain what it does to let other classes to use it.
I also wonder if we can use `TestUtils.createTestApplicationContext()` instead which has an expected bean registered.
Right, we might need to think about closing that ctx in the end of test, but that is indeed an existing API for the goal of your change: refactor code to minimize duplication.

artembilan (changes requested):
Let me know if my comments make sense to you.
Thank you for your effort!

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/MockClientConnectionFactoryBuilder.java
```diff
@@ -26,12 +26,12 @@
  *
  */
 
-public class MockFactory {
-	private MockFactory() {
+public class MockClientConnectionFactoryBuilder {
```
artembilan, line 29: I am still against these new classes.
Please, make the logic as `private static` in those test classes affected.
Thanks

artembilan (changes requested):

artembilan:
> Merge branch 'spring-projects:main' into refactor-CachingClientConnectionFactoryTests

I told you to use `rebase` not `merge`.

Issue #3866, opened 2022-07-28
duplicate key value violates unique constraint "int_lock_pk"


dkwakkel:
**In what version(s) of Spring Integration are you seeing this issue?**

For example:

5.5.14

**Describe the bug**

When using DefaultLockRepository The following error is shown in postgres log under high load:

STATEMENT:  INSERT INTO INT_LOCK (REGION, LOCK_KEY, CLIENT_ID, CREATED_DATE) VALUES ($1, $2, $3, $4)
ERROR:  duplicate key value violates unique constraint "int_lock_pk"

**To Reproduce**

Use DefaultLockRepository with high load with postgres

**Expected behavior**

No error

** Workaround **
We use below workaround, but would be nice if this can be solved somewhere in spring integration itself:
```
		DefaultLockRepository repository = new DefaultLockRepository(datasource) {

			@Override
			public void afterPropertiesSet() {
				// Postgres specific fix to have no 'duplicate key value violates unique constraint "int_lock_pk"' in the log.
				if(asList(environment.getActiveProfiles()).contains("postgres")) {
					addOnConflictDoNothing();
				}
				super.afterPropertiesSet();
			}

			private void addOnConflictDoNothing() {
				String fieldName = "insertQuery";
				try {
					Field field = DefaultLockRepository.class.getDeclaredField(fieldName);
					field.setAccessible(true);
					field.set(this, field.get(this) + " ON CONFLICT DO NOTHING");
				} catch(Exception e) {
					log.warn("Unable to change spring integration '" + fieldName + "' query", e);
				}
			}
		};
		return repository;
	}
```

artembilan:
May we see the whole stack trace, please?
Doesn't look like that `duplicate key value violates unique constraint` is translated to the `DataIntegrityViolationException` some way...
This is not the first time I see that PostgreSQL drive throws non-standard error which can be translated properly to respective exception in the `JdbcTemplate`.

I think we can fix that `DefaultLockRepository` exposing all the statements as setters, so you will be able to do something like:
```
repository.setInsertQuery("INSERT INTO %sLOCK (REGION, LOCK_KEY, CLIENT_ID, CREATED_DATE) VALUES (?, ?, ?, ?) ON CONFLICT DO NOTHING");
```
If that is really what is recommended for handling duplicate keys in PostgreSQL ...

Another way would be to let this `DefaultLockRepository` to be extendable, so you could override that `acquire(String lock)`.

In `6.0`, though, we added this:
```
[...]

PR #9292, opened 2024-06-30
GH-9291: Enhanced unlock() method of JdbcLock to verify successful unlocking

### LINKED ISSUE #9291: Enhanced unlock() method of JdbcLock to verify successful unlocking
**Expected Behavior**

* The `unlock()` method of `JdbcLock` should verify the execution result of releasing the lock ownership.
* If the ownership can not be removed due to data expiration, a `ConcurrentModificationException` should be thrown.
* `The RedisLock` implementation should also throw a `ConcurrentModificationException` in similar scenarios.

**Current Behavior**

* The `unlock()` method does not verify the execution result of releasing the lock ownership. 
* The `RedisLock` throws an `IllegalStateException` when attempting to release the lock after the ownership has expired.

**Context**

If the `unlock()` method does not verify the execution result of releasing the lock ownership, users might face concurrency issues without being notified. For example:
1. Process A and process B are operating in the same region. Process A acquires the distributed lock with lock_key: 'key'.
2. Due to prolonged work, the ownership of the lock held by process A expires.
3. Before process A releases the lock, process B also acquires the distributed lock with same lock_key.
4. When process A attempts to release the distributed lock, the current implementation does not inform the user that the integrity of data protected by this lock may have been compromised.

To address this, it is proposed that the `unlock()` method be enhanced to verify if the ownership of lock has been removed successfully.  If the ownership cannot be removed due to data expiration, a `ConcurrentModificationException` will be thrown.  This approach maintains the integrity of the distributed lock mechanism.

EddieChoCho:
Fixes: #9291
* Modify `unlock()` method of `JdbcLock`: if the lock ownership can not be removed due to data expiration, a `ConcurrentModificationException` should be thrown.
* Modify `unlock()` method of `RedisLock`: if the lock ownership can not be removed due to data expiration, a `ConcurrentModificationException` should be thrown.
* Maintain test cases

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.integration.jdbc.lock;
 
 import java.time.Duration;
+import java.util.ConcurrentModificationException;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -56,6 +57,7 @@ import org.springframework.util.Assert;
  * @author Unseok Kim
  * @author Christian Tzolov
  * @author Myeonghyeon Lee
+ * @author Eddie Cho
  *
  * @since 4.3
  */
@@ -305,11 +307,20 @@ public class JdbcLockRegistry implements ExpirableLockRegistry, RenewableLockReg
 			try {
 				while (true) {
 					try {
-						this.mutex.delete(this.path);
-						return;
+						if (this.mutex.delete(this.path)) {
+							return;
+						}
+						else {
+							throw new ConcurrentModificationException();
+							// the lock is no longer owned by current process, the exception should be handle and rollback the execution result
+						}
 					}
 					catch (TransientDataAccessException | TransactionTimedOutException | TransactionSystemException e) {
 						// try again
+					}
+					catch (ConcurrentModificationException e) {
+						throw new ConcurrentModificationException("Lock was released in the store due to expiration. " +
+								"The integrity of data protected by this lock may have been compromised.");
 					}
 					catch (Exception e) {
 						throw new DataAccessResourceFailureException("Failed to release mutex at " + this.path, e);
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -311,16 +311,15 @@ public class JdbcLockRegistry implements ExpirableLockRegistry, RenewableLockReg
 							return;
 						}
 						else {
-							throw new ConcurrentModificationException();
-							// the lock is no longer owned by current process, the exception should be handle and rollback the execution result
+							throw new ConcurrentModificationException("Lock was released in the store due to expiration. " +
+									"The integrity of data protected by this lock may have been compromised.");
 						}
 					}
 					catch (TransientDataAccessException | TransactionTimedOutException | TransactionSystemException e) {
 						// try again
 					}
 					catch (ConcurrentModificationException e) {
-						throw new ConcurrentModificationException("Lock was released in the store due to expiration. " +
-								"The integrity of data protected by this lock may have been compromised.");
+						throw e;
 					}
 					catch (Exception e) {
 						throw new DataAccessResourceFailureException("Failed to release mutex at " + this.path, e);
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/LockRepository.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/LockRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import java.io.Closeable;
  * @author Dave Syer
  * @author Alexandre Strubel
  * @author Artem Bilan
+ * @author Eddie Cho
  *
  * @since 4.3
  */
@@ -41,8 +42,9 @@ public interface LockRepository extends Closeable {
 	/**
 	 * Remove a lock from this repository.
 	 * @param lock the lock to remove.
+	 * @return deleted or not.
 	 */
-	void delete(String lock);
+	boolean delete(String lock);
 
 	/**
 	 * Remove all the expired locks.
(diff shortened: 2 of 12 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/LockRepository.java
```diff
@@ -41,8 +42,9 @@ public interface LockRepository extends Closeable {
 	/**
 	 * Remove a lock from this repository.
 	 * @param lock the lock to remove.
+	 * @return deleted or not.
 	 */
-	void delete(String lock);
+	boolean delete(String lock);
```
artembilan, line 47: This is a breaking change in the public API.
Can we keep an old version of this method as `@Deprecated`?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -305,12 +307,21 @@ public void unlock() {
...
+						else {
+							throw new ConcurrentModificationException();
+							// the lock is no longer owned by current process, the exception should be handle and rollback the execution result
+						}
 					}
 					catch (TransientDataAccessException | TransactionTimedOutException | TransactionSystemException e) {
 						// try again
 					}
+					catch (ConcurrentModificationException e) {
+						throw new ConcurrentModificationException("Lock was released in the store due to expiration. " +
```
artembilan, line 322: I'm not sure what is the reason to catch it here and re-throw immediately.
Why we just cannot include the message in the original one?

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/lock/JdbcLockRegistryTests-context.xml
```diff
@@ -24,8 +24,6 @@
 
 	<bean id="lockClient" class="org.springframework.integration.jdbc.lock.DefaultLockRepository">
 		<constructor-arg name="dataSource" ref="dataSource"/>
-		<property name="insertQuery"
-				  value="INSERT INTO INT_LOCK (REGION, LOCK_KEY, CLIENT_ID, CREATED_DATE) VALUES (?, ?, ?, ?)"/>
```
artembilan, line 28: There was a reason to have this in the test: https://github.com/spring-projects/spring-integration/issues/3866.

Even if it is the same value we have by default, it still covers a respective setter call.
So, please, consider to revert your change.
It is still not relevant to the purpose of this PR.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/lock/JdbcLockRegistryTests.java
```diff
@@ -270,7 +272,12 @@ public void testTwoThreadsDifferentRegistries() throws Exception {
 					Thread.currentThread().interrupt();
 				}
 				finally {
-					lock2.unlock();
+					try {
+						lock2.unlock();
+					}
+					catch (ConcurrentModificationException ignored) {
```
artembilan, line 278: Why do we need to catch it here, please?

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/lock/JdbcLockRegistryTests.java
```diff
@@ -502,6 +509,51 @@ void noTableThrowsExceptionOnStart() {
 		}
 	}
 
+	@Test
+	void testUnlock_lockStatusIsExpired_lockHasBeenAcquiredByAnotherProcess_ConcurrentModificationExceptionWillBeThrown() throws Exception {
```
artembilan, line 513: We prefer the line length not more than `120` symbols.
Please, consider to shorten the name of this test method.
Plus we prefer a camel case for names.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -500,12 +501,12 @@ private void removeLockKey() {
 					return;
 				}
 				else if (Boolean.FALSE.equals(unlinkResult)) {
-					throw new IllegalStateException("Lock was released in the store due to expiration. " +
+					throw new ConcurrentModificationException("Lock was released in the store due to expiration. " +
```
artembilan, line 504: This is breaking change in the behavior.
We will have to mention this in the Migration Guide when we done with this PR: https://github.com/spring-projects/spring-integration/wiki

artembilan (changes requested):
This also requires respective changes in the `lock-registry.adoc`, `redis.adoc` and `whats-new.adoc`.

The change is aimed for new `6.4` version.

Thanks

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
EddieChoCho: There is a catch block for catching general `Exception` in this method which will covert exceptions to `DataAccessResourceFailureException`.
To avoid `ConcurrentModificationException` be convert to `DataAccessResourceFailureException`, it needs to has it's own catch block.
Yes, we can include the message in the original one.

artembilan: Yeah... I see.
OK.
Let it be then like:
```
catch (ConcurrentModificationException e) {
    throw e;
}
```
And move that message to the original `throw` point.
With this `throw e;` we just don't create a new object.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/LockRepository.java
artembilan: After more thinking even if it is a breaking change from the API perspective, it is still save to call this method without checking its return.
So, having the fact that we still are changing behavior of the `RedisLockRegistry`, I probably will be OK to leave this as is and just mention in the Migration Guide.

artembilan (changes requested):

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/lock/JdbcLockRegistryTests.java
artembilan: Pushed.
Please, rebase your branch against the latest `main`.

src/reference/antora/modules/ROOT/pages/jdbc/lock-registry.adoc
```diff
@@ -57,3 +57,4 @@ For example, an insert query for PostgreSQL hint can be configured like this:
 lockRepository.setInsertQuery(lockRepository.getInsertQuery() + " ON CONFLICT DO NOTHING");
 ----
 
+Starting with version 6.4, the `LockRepository.delete()` method return the result of removing ownership of a distributed lock. And the `JdbcLockRegistry.JdbcLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
```
artembilan, line 60: We prefer `One Sentence per Line` best practice: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -32,7 +32,14 @@ The `LobHandler` (and respective API) has been deprecated for removal in Spring
 Respective option on `JdbcMessageStore` (and similar) have been deprecated as well.
 The byte array handling for serialized message is fully deferred to JDBC driver.
 
+The `LockRepository.delete()` method return the result of removing ownership of a distributed lock. And the `JdbcLockRegistry.JdbcLock.unlock()` method throws ConcurrentModificationException if the ownership of the lock is expired.
```
artembilan, line 35: DITTO.
Plus the `ConcurrentModificationException` is supposed to be a a code snippet like you do in other places.

artembilan (changes requested):
So, all good. Just a couple nit-pics for the doc.

Thanks

EddieChoCho:
Thank you for your feedback and the good reference! I've just pushed another commit.

artembilan:
@EddieChoCho ,

I know, we have with you a `DistributedLock` feature outstanding 😄

artembilan:
Here is a Migration Guide page we have talked about before: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-6.3-to-6.4-Migration-Guide

PR #9370, opened 2024-08-08
GH-9369: Fix MutableMessageHeaders not being editable after deserialization

### LINKED ISSUE #9369: Serializing MutableMessageHeaders containing a non-serializable header property results in serializer incorrectly serializing the 'headers' property on the class

v6.2.7

**Describe the bug**

When MutableMessageHeaders is serialized using Spring Cores DefaultSerializer with a header that is not serializable it can result in the 'headers' property on the class being serialized as the superclass (MessageHeaders) instead of a HashMap. When deserialized this causes the 'headers' property to be an instance of MessageHeaders which prevents callers from adding properties to the MutableMessageHeaders class (When they should be able to). The following error is observed

```
Exception in thread "main" java.lang.UnsupportedOperationException: MessageHeaders is immutable
	at org.springframework.messaging.MessageHeaders.put(MessageHeaders.java:274)
	at org.springframework.messaging.MessageHeaders.put(MessageHeaders.java:73)
	at org.springframework.integration.support.MutableMessageHeaders.put(MutableMessageHeaders.java:63)
```

This looks to be related to the `writeObject` on the MessageHeaders class which will instantiate a new instance of the MessageHeaders class if there is a non-serializable object in the headers. Its likely that similar functionality needs to be overridden in the MutableMessageHeaders class in order to serialize it correctly. (See https://github.com/spring-projects/spring-framework/blob/main/spring-messaging/src/main/java/org/springframework/messaging/MessageHeaders.java#L307)

If there is no non-serializable objects in the headers then the MutableMessageHeaders works as expected when deserialized, this is because it just writes the object directly in the `writeObject` method rather than creating a new instance of MessageHeaders

This issue was originally observed when using the RedisMessageStore with an aggregator which under the hood makes use of the DefaultSerializer class, when I enabled observations it attempted to add the traceparent to the message header after it had been deserialized which then resulted in a failure.

**To Reproduce**

1. Create a MutableMessageHeaders instance with a header value containing a non-serializable instance
2. Serialize the class using the SerializingConverter
3. Deserialize the class using the DeserializingConverter using the output from step 2
4. Attempt to put a header into the deserialized MutableMessageHeaders instance

**Expected behavior**

Its expected that when MutableMessageHeaders is deserialized back into the object that it should be able to behave as normal and allow headers to be added using the classes 'put' method.

**Sample**

Link to sample showing bug: https://github.com/mitchmcd18/spring-messageheader-bug

**Related links**
https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/serializer/support/SerializingConverter.java
https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/serializer/support/DeserializingConverter.java
https://github.com/spring-projects/spring-integration/blob/main/spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java

mitchmcd18:
Fixes: #9369

* Fix MutableMessageHeaders not being editable after deserialization

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/MutableMessageHeaders.java
@@ -16,6 +16,8 @@
 
 package org.springframework.integration.support;
 
+import java.io.ObjectStreamException;
+import java.io.Serial;
 import java.nio.ByteBuffer;
 import java.util.Map;
 import java.util.UUID;
@@ -31,6 +33,7 @@ import org.springframework.messaging.MessageHeaders;
  * @author David Turanski
  * @author Artem Bilan
  * @author Nathan Kurtyka
+ * @author Mitchell McDonald
  *
  * @since 4.2
  */
@@ -71,6 +74,11 @@ public class MutableMessageHeaders extends MessageHeaders {
 	@Override
 	public Object remove(Object key) {
 		return super.getRawHeaders().remove(key);
+	}
+
+	@Serial
+	private Object readResolve() throws ObjectStreamException {
+		return new MutableMessageHeaders(this);
 	}
 
 	@Nullable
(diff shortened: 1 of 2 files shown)


artembilan (approved):
I'm fine with the solution you propose.

Will be merged when PR build is green.

Thank you!

PR #9382, opened 2024-08-14
GH-9368: Support for adding MqttMessageDrivenChannelAdapter at runtime

### LINKED ISSUE #9368: Adding Mqttv5PahoMessageDrivenChannelAdapter to running Mqttv5ClientManager
version 6.3.1

**Description**
I want to add multiple instances of Mqttv5PahoMessageDrivenChannelAdapter (on demand, when application is running) using already running client in Mqttv5ClientManager as constructor parameter. When I try, it does not work. 
Cause seams to be MqttCallback:: connectComplete ,which sets the readyToSubscribeOnStart property.
That obviously cant trigger because client is already connected.

**To Reproduce**
1.	Create an instance of Mqttv5ClientManager
2.	Start clientManager
3.	create an instance of Mqttv5PahoMessageDrivenChannelAdapter with clientManager
4.	register adapter in IntegrationFlow using IntegrationFlowContext
5.	Try to send messages to it

**Possible Fix**
https://github.com/Ricore72/spring-integration
The fix does not work completely.

It only works if Handlers and adapters using same clientManager have the same Qos.

**Note** 
I am not sure if this is the proper way to do this. This is my first issue to public project. So sorry if it is incorrect in any way.

Ricore72:
Fixes: #9368 

Currently, it is not possible to add `MqttMessageDrivenChannelAdapter` after `MqttClient` is connected, using `ClientManager` in constructor.

* Exposed `isConnection()` in `ClientManager` 
* Implemented `isConnection()` in `Mqttv3ClientManager` and `Mqttv5ClientManager`
* Modified `OnInit()` in `AbstractMqttMessageDrivenChannelAdapter` to set `readyToSubscribeOnStart` = true
* Added tests to `ClientManagerBackToBackTests` to check functionality (`testV3ClientManagerStarted`, `testV5ClientManagerStarted`)

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
@@ -74,7 +74,7 @@ public interface ClientManager<T, C> extends SmartLifecycle, MqttComponent<C> {
 	 * @return the managed clients isConnected.
 	 * @since 6.4
 	 */
-	boolean isConnection();
+	boolean isConnected();
 
 	/**
 	 * A contract for a custom callback on {@code connectComplete} event from the client.
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
@@ -28,6 +28,7 @@ import org.springframework.context.SmartLifecycle;
  *
  * @author Artem Vozhdayenko
  * @author Artem Bilan
+ * @author Jiri Soucek
  *
  * @since 6.0
  */
@@ -71,6 +72,7 @@ public interface ClientManager<T, C> extends SmartLifecycle, MqttComponent<C> {
 	/**
 	 * Return the managed clients isConnected.
 	 * @return the managed clients isConnected.
+	 * @since 6.4
 	 */
 	boolean isConnection();
 
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2023 the original author or authors.
+ * Copyright 2022-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -67,6 +67,12 @@ public interface ClientManager<T, C> extends SmartLifecycle, MqttComponent<C> {
 	 * @return true if callback was removed.
 	 */
 	boolean removeCallback(ConnectCallback connectCallback);
+
+	/**
+	 * Return the managed clients isConnected.
+	 * @return the managed clients isConnected.
+	 */
+	boolean isConnection();
 
 	/**
 	 * A contract for a custom callback on {@code connectComplete} event from the client.
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
@@ -200,7 +200,9 @@ public class Mqttv3ClientManager
 
 	@Override
 	public boolean isConnection() {
-		if(getClient()!=null) return getClient().isConnected();
+		if (getClient() != null) {
+			return getClient().isConnected();
+		}
 		return false;
 	}
 }
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
@@ -198,4 +198,9 @@ public class Mqttv3ClientManager
 		// nor this manager concern
 	}
 
+	@Override
+	public boolean isConnection() {
+		if(getClient()!=null) return getClient().isConnected();
+		return false;
+	}
 }
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
@@ -38,6 +38,7 @@ import org.springframework.util.Assert;
  * @author Artem Vozhdayenko
  * @author Artem Bilan
  * @author Christian Tzolov
+ * @author Jiri Soucek
  *
  * @since 6.0
  */
@@ -200,9 +201,17 @@ public class Mqttv3ClientManager
 
 	@Override
 	public boolean isConnection() {
-		if (getClient() != null) {
-			return getClient().isConnected();
+		this.lock.lock();
+		try {
+			IMqttAsyncClient client = getClient();
+			if (client != null) {
+				return client.isConnected();
+			}
+			return false;
 		}
-		return false;
+		finally {
+			this.lock.unlock();
+		}
+
 	}
 }
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
@@ -200,7 +200,7 @@ public class Mqttv3ClientManager
 	}
 
 	@Override
-	public boolean isConnection() {
+	public boolean isConnected() {
 		this.lock.lock();
 		try {
 			IMqttAsyncClient client = getClient();
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv5ClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv5ClientManager.java
@@ -40,6 +40,7 @@ import org.springframework.util.Assert;
  * @author Artem Vozhdayenko
  * @author Artem Bilan
  * @author Christian Tzolov
+ * @author Jiri Soucek
  *
  * @since 6.0
  */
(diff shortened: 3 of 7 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
```diff
@@ -68,6 +68,12 @@ public interface ClientManager<T, C> extends SmartLifecycle, MqttComponent<C> {
 	 */
 	boolean removeCallback(ConnectCallback connectCallback);
 
+	/**
+	 * Return the managed clients isConnected.
+	 * @return the managed clients isConnected.
```
artembilan, line 73: `@since 6.4`

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv3ClientManager.java
```diff
@@ -198,4 +198,11 @@ public void deliveryComplete(IMqttDeliveryToken token) {
 		// nor this manager concern
 	}
 
+	@Override
+	public boolean isConnection() {
+		if (getClient() != null) {
+			return getClient().isConnected();
```
artembilan, line 204: Better to extract into a local variable.
I mean `IMqttAsyncClient client = getClient()`.
Also wrap the logic into a `this.lock.lock();`.
See `start()` for example.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/Mqttv5ClientManager.java
```diff
@@ -206,4 +206,11 @@ public void mqttErrorOccurred(MqttException exception) {
 		logger.error("MQTT error occurred", exception);
 	}
 
+	@Override
+	public boolean isConnection() {
+		if (getClient() != null) {
+			return getClient().isConnected();
```
artembilan, line 212: DITTO

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -203,6 +203,9 @@ protected void onInit() {
 		super.onInit();
 		if (this.clientManager != null) {
 			this.clientManager.addCallback(this);
+			if (this.clientManager.isConnection()) {
+				connectComplete(true);
```
artembilan, line 207: I think has to be `false` since we indeed not reconnecting.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/ClientManagerBackToBackTests.java
```diff
@@ -177,6 +189,42 @@ public IntegrationFlow mqttInFlow(Mqttv3ClientManager mqttv3ClientManager) {
...
+			MqttConnectOptions connectionOptions = new MqttConnectOptions();
+			connectionOptions.setServerURIs(new String[] {MosquittoContainerTest.mqttUrl()});
+			connectionOptions.setAutomaticReconnect(true);
+			Mqttv3ClientManager manager = new Mqttv3ClientManager(connectionOptions, "client-manager-client-id-v3");
+			return manager;
+		}
+
+		@Bean
+		public IntegrationFlow mqttOutFlow(Mqttv3ClientManager mqttv3ClientManager) {
+			return f -> f.handle(new MqttPahoMessageHandler(mqttv3ClientManager));
```
artembilan, line 216: This configuration does not reflect the purpose of the issue.
I believe these new tests are going to work even without the change in the client managers and channel adapters.
Please, consider to provide the test configuration according to your experience: register channel adapters at runtime via `IntegrationFlowContext`.

artembilan (changes requested):
Please, add your name to the `@author` list of all the affected class.

Would be also great to add a `whats-new.adoc` for this new feature and explain in in the target `mqtt.adoc` where we describe `Shared MQTT Client Support`.

Thanks

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/ClientManagerBackToBackTests.java
Ricore72: yes, sorry I modified it incorrectly. `Mqttv3ClientManager` was meaned to be started in before returning. Didnt noticed olny one test is failing on `main`

I added test with `IntegrationFlowContext` in the modifications and fixed the this configuration. 

4 tests failing on `main` now

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
```diff
@@ -68,6 +69,13 @@ public interface ClientManager<T, C> extends SmartLifecycle, MqttComponent<C> {
...
+	 * @since 6.4
+	 */
+	boolean isConnection();
```
artembilan, line 77: I think the name of the method is slightly wrong.
I believe it has to be `isConnected()`.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -61,8 +61,16 @@ See xref:redis.adoc[Redis Support] for more information.
...
 [[x6.4-sftp-changes]]
 === SFTP Support Changes
 
 The `DefaultSftpSessionFactory` now exposes a `Consumer<SshClient>` configurer property to further customize an internal `SshClient`.
-See xref:sftp/session-factory.adoc[SFTP Session Factory] for more information.
\ No newline at end of file
+See xref:sftp/session-factory.adoc[SFTP Session Factory] for more information.
+
+[[x6.4-mqtt-changes]]
+=== mqtt Changes
```
artembilan, line 72: `MQTT Support Changes`, please.

src/reference/antora/modules/ROOT/pages/mqtt.adoc
```diff
@@ -548,3 +548,19 @@ public IntegrationFlow mqttOutFlow(
     return f -> f.handle(new Mqttv5PahoMessageHandler(clientManager));
 }
 ----
+
+NOTE: Starting with version 6.4, multiple instances of `MqttPahoMessageDrivenChannelAdapter` and `Mqttv5PahoMessageDrivenChannelAdapter` can now be added at runtime using corresponding `ClientManager` through `IntegrationFlowContext`
+
+[source,java]
+----
+private void addAddRuntimeAdapter(IntegrationFlowContext flowContext,Mqttv5ClientManager clientManager,
```
artembilan, line 556: Whitespace is supposed to be after each comma.
Also double check, please, that there are no tab indents in all those lines of code.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/ClientManagerBackToBackTests.java
```diff
@@ -102,6 +131,40 @@ private void testSubscribeAndPublish(Class<?> configClass, String topicName, Cou
...
+			Message<?> receive = output.receive(20_000);
+
+			// then
+			assertThat(receive).isNotNull();
+			Object payload = receive.getPayload();
+			if (payload instanceof String sp) {
+				assertThat(sp).isEqualTo(testPayload);
+			}
+			else {
+				assertThat(payload).isEqualTo(testPayload.getBytes(StandardCharsets.UTF_8));
```
artembilan, line 163: Why does this happen?
Why do we need this `if..else` logic?
Why the payload cannot always be string to simplify the things?

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/ClientManagerBackToBackTests.java
```diff
@@ -102,6 +131,40 @@ private void testSubscribeAndPublish(Class<?> configClass, String topicName, Cou
...
+			throws Exception {
+
+		try (var ctx = new AnnotationConfigApplicationContext(configClass)) {
+			// given
+			var input = ctx.getBean("mqttOutFlow.input", MessageChannel.class);
+			var flowContext = ctx.getBean(IntegrationFlowContext.class);
+			var clientManager = ctx.getBean(ClientManager.class);
+			var output = new QueueChannel();
+			Class<?>[] parameterTypes = {ClientManager.class, String[].class};
+			Constructor<?> declaredConstructor = adapter.getConstructor(parameterTypes);
```
artembilan, line 144: I don't see a reason in this reflection.
You just can have a factory method in the `Mqttv5ConfigRuntime` which would return us a respective `MessageProducerSupport` instance.
The `@Configuration` class is registered as a bean as well.
So, you might just need to extract some common interface for both of your `Mqttv5ConfigRuntime` and `Mqttv3ConfigRuntime` to have that factory method implemented respectively.

artembilan (changes requested):

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/ClientManagerBackToBackTests.java
Ricore72: To be honest I have just copied previously used tests but It is not possible, because `MqttPahoMessageDrivenChannelAdapter` returns `String`  and `Mqttv5PahoMessageDrivenChannelAdapter` returns `byte[]`. I think that it is due to changes between `MQTT v3` and `MQTT v5`, have not researched it.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/ClientManagerBackToBackTests.java
```diff
@@ -247,6 +378,76 @@ public IntegrationFlow mqttInFlow(Mqttv5ClientManager mqttv5ClientManager) {
...
+
+		@EventListener
+		public void onSubscribed(MqttSubscribedEvent e) {
+			subscribedLatch.countDown();
+		}
+
+		@Bean
+		public Mqttv5ClientManager mqttv5ClientManager() {
+			Mqttv5ClientManager manager = new Mqttv5ClientManager(MosquittoContainerTest.mqttUrl(), "client-manager-client-id-v5");
+			manager.start();
```
artembilan, line 397: You still didn't address this.
One more time: it is an anti-pattern to have component started too early.
Therefore the whole purpose of these `...WithStartedManager` configurations is wrong.
I suggestion to remove them altogether with their dedicated tests.
The regular (recommended) application context start is covered by other tests.
What we need for this change is really that based on the `IntegrationFlowContext` and beans added at runtime.
Therefore I'd expect only two new tests in your PR.

Thanks

artembilan (changes requested):

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/ClientManagerBackToBackTests.java
Ricore72: I am sorry.
Removed the tests and Messed up the history.  Thank you for your feedback and time. I think it must be hard to cooperate with someone like me. Hope everything is ok now.

Issue #9383, opened 2024-08-16
Replace `JsonPropertyAccessor` with `JsonIndexAccessor`


sbrannen:
Spring Framework 6.2 introduces dedicated support for indexing into custom structures via a new `IndexAccessor` SPI, with a corresponding `CompilableIndexAccessor` SPI and a `ReflectiveIndexAccessor` implementation.

Spring Integration should consider replacing (and potentially deprecating) the existing `JsonPropertyAccessor` with an `JsonIndexAccessor`.

Documentation on the new features in Spring Framework can be found in the [reference manual](https://docs.spring.io/spring-framework/reference/6.2/core/expressions/language-ref/properties-arrays.html#expressions-indexing-custom).

In addition, a proof-of-concept `JacksonArrayNodeIndexAccessor` can be found in Spring Framework's test suite:

https://github.com/spring-projects/spring-framework/blob/a9a1798012242facac264cc468a23d01155f3e26/spring-expression/src/test/java/org/springframework/expression/spel/IndexingTests.java#L813-L850

sbrannen:
If you decide to implement a `JsonIndexAccessor`, I'd be happy to assist you in case you run into any issues.

And... it would be great if you could investigate the possibility before Spring Framework 6.2 GA is released, so that we have time to make any necessary changes before the SPIs and implementations in Spring Framework are set in stone.

artembilan:
Hi @sbrannen !

Thank you for sharing this!

Looking into the code I'm not totally sure why to replace.
Looks like that `JacksonArrayNodeIndexAccessor` deals only with an `ArrayNode` which is just a part of the logic of this `JsonPropertyAccessor ` in its inner `ArrayNodeAsList`.

I will try to replace it blindly with the suggested `JsonIndexAccessor`, but I'm not sure if that is enough.

I'll let you know later today or next week.
Looks like our `SpelPropertyAccessorRegistrar` should be fixed for the `TargetedAccessor` instead.
Then its name is a bit misleading. So, deprecate that one as well in favor of `SpelAccessorRegistrar`  😄

artembilan:
So, this a failure in my test for the `<int:transformer expression="payload.firstName"/>` when I remove `JsonPropertyAccessor` and just add `JsonIndexAccessor`:
```
[...]

PR #9429, opened 2024-08-29
Introduce `JsonIndexAccessor`

sbrannen:
## Overview

This PR contains two commits.

The first polishes `JsonPropertyAccessor` and associated tests.

The second introduces a `JsonIndexAccessor` as a complement to the existing `JsonPropertyAccessor`.

When a `JsonIndexAccessor` is registered with the SpEL `EvaluationContext` alongside a registered `JsonPropertyAccessor`, JSON arrays can be consistently indexed via integer literals (e.g., `[1]`) instead of string literals representing integers (e.g., `['1']`).

## Related Issues

- #9383

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/json/JsonIndexAccessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/json/JsonIndexAccessor.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2013-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.json;
+
+import com.fasterxml.jackson.databind.node.ArrayNode;
+
+import org.springframework.expression.AccessException;
+import org.springframework.expression.EvaluationContext;
+import org.springframework.expression.IndexAccessor;
+import org.springframework.expression.TypedValue;
+import org.springframework.lang.Nullable;
+
+/**
+ * A SpEL {@link IndexAccessor} that knows how to read indexes from JSON arrays, using
+ * Jackson's {@link ArrayNode} API.
+ *
+ * <p>Supports indexes supplied as an integer literal &mdash; for example, {@code myJsonArray[1]}.
+ * Also supports negative indexes &mdash; for example, {@code myJsonArray[-1]} which equates
+ * to {@code myJsonArray[myJsonArray.length - 1]}. Furthermore, {@code null} is returned for
+ * any index that is out of bounds (see {@link ArrayNode#get(int)} for details).
+ *
+ * @author Sam Brannen
+ * @since 6.4
+ * @see JsonPropertyAccessor
+ */
+public class JsonIndexAccessor implements IndexAccessor {
+
+	private static final Class<?>[] SUPPORTED_CLASSES = { ArrayNode.class };
+
+	@Override
+	public Class<?>[] getSpecificTargetClasses() {
+		return SUPPORTED_CLASSES;
+	}
+
+	@Override
+	public boolean canRead(EvaluationContext context, Object target, Object index) {
+		return (target instanceof ArrayNode && index instanceof Integer);
+	}
+
+	@Override
+	public TypedValue read(EvaluationContext context, Object target, Object index) throws AccessException {
+		ArrayNode arrayNode = (ArrayNode) target;
+		Integer intIndex = (Integer) index;
+		if (intIndex < 0) {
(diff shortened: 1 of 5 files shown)


artembilan (changes requested):
Hi @sbrannen !

I'd like to have a couple clarifications:

1. You say in your commit message ` instead of string literals representing integers (e.g., ['1']).`. But that's not fully true according to our existing tests, e.g. `JsonPropertyAccessorTests.testArrayLookup()` or the next one.
I might miss something and this sentence is for another use-case.
2. Would you mind to give me some explanation for docs to justify this new `JsonIndexAccessor` and what benefits it has when it is used in combination with `JsonPropertyAccessor`? 
Or feel free to make a respective change into `spel.adoc`. 😄

sbrannen:
Hi @artembilan,

> Thank you very much for the contribution.

You're very welcome. 😎 

> You say in your commit message `instead of string literals representing integers (e.g., ['1']).`. But that's not fully true according to our existing tests, e.g. `JsonPropertyAccessorTests.testArrayLookup()` or the next one.
> I might miss something and this sentence is for another use-case.

On `main`, `JsonPropertyAccessorTests.testArrayLookup()` explicitly **wraps** the `ArrayNode`, which means that it is no longer an `ArrayNode`. Instead, it becomes an `ArrayNodeAsList` which is a `java.util.List`, and you can always index into a `List` with integer indexes.

In my PR, I renamed that method to `arrayLookupWithIntegerIndexAndExplicitWrapping()` to make it clear that explicit, manual wrapping is occurring in the test method.

Whereas, users would likely not manually wrap an `ArrayNode` that they make available in a SpEL expression, meaning that users would have to supply `['1']` instead of `[1]` as the expression.

Without the `JsonIndexAccessor`, the first index must always be a `String` if you attempt to index directly into an `ArrayNode`. 

I made this clear in the new `nestedArrayLookupWithStringIndexAndThenIntegerIndex()` test method, which works with or without the `JsonIndexAccessor`; whereas, **all** of the tests in `JsonPropertyAccessorTests.ArrayNodeTests` assert that integer indexing is not supported with only a `JsonPropertyAccessor`. On the other hand, all of the tests in `JsonIndexAccessorTests.ArrayNodeTests` pass with integer indexes.

Does that clarify things for you?

sbrannen:
> Would you mind to give me some explanation for docs to justify this new `JsonIndexAccessor` and what benefits it has when it is used in combination with `JsonPropertyAccessor`?

To be honest, I originally thought that introducing a `JsonIndexAccessor` would provide greater benefit to Spring Integration (SI) users; however, after thoroughly diving into the current feature set in SI, I realized that the _wrapping_ functionality in `JsonPropertyAccessor` already covers a lot of ground, and I was able to reuse that wrapping functionality in `JsonIndexAccessor` (indirectly via `JsonPropertyAccessor.typedValue()`).

For example, as mentioned in my previous comment, the "wrapping" of an `ArrayNode` in a `List` allows for integer indexing into the list.

And that wrapping occurs automatically whenever a property is accessed on a `JsonNode` which returns an `ArrayNode` or whenever indexing into an `ArrayNodeAsList` returns an `ArrayNode` wrapped in an `ArrayNodeAsList`.

The only time an `ArrayNode` is not automatically wrapped in an `ArrayNodeAsList` is when the root context object is an `ArrayNode` or when a property or method invocation returns an `ArrayNode` directly.

In summary, the primary (perhaps only) benefit of `JsonIndexAccessor` is that it allows JSON arrays to be consistently indexed via integer literals, with the key word being "consistently" (meaning "always").

Having said that, I still think it's beneficial to have the `JsonIndexAccessor` since it allows people to write idiomatic expressions like `[1][2]` instead of `['1']['2']`.

> Or feel free to make a respective change into `spel.adoc`. 😄

Oh... you want me to write documentation, too? 🤣

Kidding aside, I'll see if I find time to look at / edit the `spel.adoc`.

sbrannen:
> 2\. Would you mind to give me some explanation for docs to justify this new `JsonIndexAccessor` and what benefits it has when it is used in combination with `JsonPropertyAccessor`?
>         Or feel free to make a respective change into `spel.adoc`.

I took a look at the [Spring Expression Language (SpEL)](https://docs.spring.io/spring-integration/reference/6.4-SNAPSHOT/spel.html) section of the reference manual, and based on the status quo, I don't think it makes sense to add documentation for the `JsonIndexAccessor` _yet_.

Rationale: you'll first need to decide if you want to introduce first-class configuration support for `IndexAccessor`s in general.

For example, you can add a `PropertyAccessor` to `IntegrationEvaluationContextFactoryBean` and `IntegrationSimpleEvaluationContextFactoryBean` (which is not currently documented in the reference manual), via the `<spel-property-accessors>` XML namespace element, and via a `spelPropertyAccessorRegistrar` `@Bean` method. But... there's currently no easy way for SI users to register an `IndexAccessor` (either a custom one of their own or the `JsonIndexAccessor`.

sbrannen:
> Rationale: you'll first need to decide if you want to introduce first-class configuration support for `IndexAccessor`s in general.

I created https://github.com/spring-projects/spring-integration/issues/9436, so that that issue can be addressed separately, independent of the fate of this PR.

artembilan (approved):
OK, Sam.
Nothing to complain about.
Merging as is.
I'll take care about `IndexAccessor` injection support and respective docs with the issue you have created.

Thank you for contribution!

sbrannen:
> Merging as is. 

Awesome!

> I'll take care about `IndexAccessor` injection support and respective docs with the issue you have created.

Sounds good. I'll keep an eye on the status of that issue.

> Thank you for contribution!

🙇

PR #9454, opened 2024-09-12
GH-9453: Correct separator when checking smb path

### LINKED ISSUE #9453: Spring Integration Smb upload fails to create missing remote directory when running on Windows
https://github.com/spring-projects/spring-integration/blob/775bfd6168925c8fee94a22deec11ca84926550e/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java#L341

Hi, the method linked above checks the remote path (which is a smb path) for `FILE_SEPARATOR` which is initialized as
```
private static final String FILE_SEPARATOR = FileSystems.getDefault().getSeparator();
```
Clearly on Windows that is "\\" which is never found in a smb path. So the need to create possibly missing remote dirs is never detected.

pfosser:
Fixes: #9453
Issue link: https://github.com/spring-projects/spring-integration/issues/9453

On a smb file upload, to detect the need for new remote directories, the remote path was checked for the local filesystem path separator. On Windows that is \\ which is never found in a smb path, so the necessary remote path was not created and the operation failed.

Use the correct separator which was already available as a constant.

@pivotal-cla This is an Obvious Fix

### DIFF
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URL;
-import java.nio.file.FileSystems;
 import java.util.Arrays;
 
 import jcifs.smb.SmbException;
@@ -54,14 +53,13 @@ import org.springframework.util.StringUtils;
  * @author Prafull Kumar Soni
  * @author Gregory Bragg
  * @author Adam Jones
+ * @author Paolo Fosser
  *
  * @since 6.0
  */
 public class SmbSession implements Session<SmbFile> {
 
 	private static final LogAccessor logger = new LogAccessor(SmbSession.class);
-
-	private static final String FILE_SEPARATOR = FileSystems.getDefault().getSeparator();
 
 	private static final String SMB_FILE_SEPARATOR = "/";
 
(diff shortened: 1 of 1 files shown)


spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
```diff
@@ -23,19 +23,18 @@
...
-
 import org.springframework.core.log.LogAccessor;
 import org.springframework.integration.file.remote.session.Session;
 import org.springframework.util.Assert;
 import org.springframework.util.FileCopyUtils;
 import org.springframework.util.StringUtils;
 
+import jcifs.smb.SmbException;
+import jcifs.smb.SmbFile;
+import jcifs.smb.SmbFileOutputStream;
```
artembilan, line 36: This is not correct change.
Please, consider to revert imports order.

artembilan (changes requested):
Add your name to the `@author` list of that affected class.

Thanks

artembilan (changes requested):
The build fails clearly with that import order problem:
```
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java:34:1: Wrong order for 'jcifs.smb.SmbException' import. [ImportOrder]
```
Consider to run `./gradlew :spring-integration-smb:check` before pushing changes to the PR.

Thanks

artembilan:
@pfosser ,

PR #9493, opened 2024-09-20
GH-9455: Introduce `IntegrationKeepAlive`

### LINKED ISSUE #9455: Application shuts down immediately after connecting to a websocket
I wrote a simple Spring Integration app that connects to a websocket, using `spring-integration-websocket`.

**Expected Behavior**

The app connects to the websocket and keeps listening for new messages.

**Current Behavior**

Once connected the application shuts down immediately, since there is no non-daemon thread running, as pointed out by @artembilan on stackoverflow (https://stackoverflow.com/questions/78961574/why-does-my-spring-integration-app-just-shutdown-without-listening/) and the canonical way to keep this from happening is to use the property `spring.main.keep-alive=true`.

**Context**

It seems like a strange convention to keep this as the default. I think it would make sense to make the default the inverse of the current behavior.

A working example:

```
import org.springframework.integration.channel.PublishSubscribeChannel;
import org.springframework.integration.handler.LoggingHandler;
import org.springframework.integration.websocket.ClientWebSocketContainer;
import org.springframework.integration.websocket.inbound.WebSocketInboundChannelAdapter;
import org.springframework.messaging.MessageChannel;
import org.springframework.web.socket.client.standard.StandardWebSocketClient;

@SpringBootApplication
@EnableIntegration
public class BinanceListenerApplication {

    @Bean
    MessageChannel rawAggTradeChannel(){
        return new PublishSubscribeChannel();
    }

    @Bean
    IntegrationFlow logging(){
        return IntegrationFlow.from(rawAggTradeChannel()).handle(new LoggingHandler(LoggingHandler.Level.INFO)).get();
    }

    @Bean
    IntegrationFlow binanceWebsocketAggTradeStream() {
        var clientWebSocketContainer = new ClientWebSocketContainer(new StandardWebSocketClient(),
                "wss://stream.binance.com:9443/ws/avaxusdt@aggTrade");
        var websocketInboundChannelAdapter = new WebSocketInboundChannelAdapter(clientWebSocketContainer);
        return IntegrationFlow
                .from(websocketInboundChannelAdapter)
                .channel(rawAggTradeChannel())
                .get();
    }

        SpringApplication application = new SpringApplication(BinanceListenerApplication.class);
        application.setWebApplicationType(WebApplicationType.NONE);
        application.run(args);
    }

}
```

artembilan:
Fixes: #9455
Issue link: https://github.com/spring-projects/spring-integration/issues/9455

* Add an `IntegrationKeepAlive` infrastructure bean to initiate a long-lived non-daemon thread to keep application alive when it cannot be kept like that for various reason, but has to.
* Expose `spring.integration.keepAlive` global property to disable an `IntegrationKeepAlive` auto-startup

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
@@ -44,6 +44,7 @@ import org.springframework.integration.channel.PublishSubscribeChannel;
 import org.springframework.integration.config.xml.IntegrationNamespaceUtils;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.context.IntegrationProperties;
+import org.springframework.integration.endpoint.management.IntegrationKeepAlive;
 import org.springframework.integration.handler.LoggingHandler;
 import org.springframework.integration.handler.support.IntegrationMessageHandlerMethodFactory;
 import org.springframework.integration.json.JsonPathUtils;
@@ -129,6 +130,7 @@ public class DefaultConfiguringBeanFactoryPostProcessor implements BeanDefinitio
 		registerListMessageHandlerMethodFactory();
 		registerIntegrationConfigurationReport();
 		registerControlBusCommandRegistry();
+		registerKeepAlive();
 	}
 
 	@Override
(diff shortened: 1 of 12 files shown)


artembilan:
@sobychacko ,

when this is merged, let's chat about Spring Cloud Stream.
If I recall correctly, there is some hook in the framework making an application keep-alive.
Since this feature is now in Spring Integration, no need to manage it over there any more after upgrading to this latest version.

sobychacko (approved):

PR #9494, opened 2024-09-21
Use isEmpty() for String type instead

ngocnhan-tran1996:


artembilan (changes requested):
Please, add you name to the `&author` list of the affected classes.
And run `./gradlew check` locally before pushing.

ngocnhan-tran1996:
I updated

artembilan:
> I updated

Please, consider in the future to not comment about the changes you have agreed with me to apply.
I'll see new changes in the next commit pushed to the PR.
Feel free to discuss whatever you are not agree.
Also: please, don't squash changes into a single commit: easier for me to review just only new changes and that would preserve a history of PR evolution.
We are squashing on merge anyway.
Extra: see what is our preference for commit message: https://cbea.ms/git-commit/

artembilan:
@ngocnhan-tran1996 ,

Would you mind sharing with us your experience how have you spotted these minor code smells even in different modules?

ngocnhan-tran1996:
@artembilan 

I search keyword `length() == 0` in IDE and check each match line in file. 
The reason is I also have a task use `String#isEmpty` instead and it's more easier beacause code smells in files are suggested by sonar

artembilan:
@ngocnhan-tran1996 ,

Can you tell us what drives you to do such a search and contribute the fix?
What is an original goal?
We do these changes as well, but not as a dedicated task - rather as side-effect of other task when we touch this or that class and they have smells like you are fixing.
Therefore I'm curious what how have you decided to make this changes since they don't feel very interesting subjects to play with.
Thanks

ngocnhan-tran1996:
@artembilan

I want to be become a better developer, and I think I can improve myself by experience reviewer advice when contribute PR. One more thing, changing old code may be break the present behavior and that must be avoid in my company, includes fix code smells, maybe they don't want deploy a feature just only fix code smells.

Code smells is an very interesting subject to me. It was a better code in the past, but now the techology has changed, it got something smells and I think fixing code smells is also a part that I can apply my knowledges.

PR #9505, opened 2024-09-22
Polish

ngocnhan-tran1996:
This PR includes:

- Polish diamond operator
- Use pattern matching
- Improve `#isEmpty` condition

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
@@ -37,6 +37,7 @@ import org.springframework.messaging.support.ErrorMessage;
  * @author Gary Russell
  * @author Kazuki Shimizu
  * @author Christian Tzolov
+ * @author Ngoc Nhan
  *
  * @since 2.0
  */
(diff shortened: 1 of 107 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
```diff
@@ -249,8 +250,8 @@ public void removeDeadConnection(TcpConnection connection) {
 				return;
 			}
 			this.removed = true;
-			if (this.theConnection instanceof TcpConnectionInterceptorSupport && !this.theConnection.equals(this)) {
-				((TcpConnectionInterceptorSupport) this.theConnection).removeDeadConnection(this);
+			if (this.theConnection instanceof TcpConnectionInterceptorSupport tcpConnectionInterceptorSupport && !this.theConnection.equals(this)) {
```
artembilan, line 253: I probably will accept this PR as is.
Due to its size and simplicity, but for the future contributions, please, consider to make code lines not more than `120` symbols.

artembilan:
@ngocnhan-tran1996 ,

Thank you for very much for the contribution; looking forward for more!

Please, consider these rules for commit messages in the future: https://cbea.ms/git-commit/

Issue #9492, opened 2024-09-19
Better inheritance support for @Nested when using @SpringIntegration


mtomik:
**Expected Behavior**

I have @SpringIntegration(noAutoStartup = {"*"}) defined in abstract class that holds all integration tests to disable all IntegrationFlows by default

When @Nested class is used in some test, same configuration (@SpringIntegration(noAutoStartup = {"*"})) is applied to that nested class as well

**Current Behavior**

class annotated with @Nested doesn't have that annotation inherited, that leads to starting all IntegrationFlows during test, that is not intended

**Context**

Without this feature, it leads to unexpected IntegrationFlow startup

sample for demonstration:
```
@SpringIntegrationTest(noAutoStartup = {"*"})
class AbstractIntegrationTest {}


class SomeIntegrationTest extends AbstractIntegrationTest {

    // all beans autoStarted just because of this  
    @Nested
    class SomeNestedTest {
        
        @Test 
        void test() {}
    }
} 
```

**workaround:**
1. Have IT profile defined and used when running integration tests
2. Define IntegrationFlow beans with non-"IT" profile  ( @Profile("!it") )

Thanks guys for any feedback on this.

artembilan:
Can you share, please, the whole test class to play with?
Or, if you have in mind the solution, consider to contribute it alongside with the mentioned test class to cover: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc
Thanks

artembilan:
Looks like there is something like `@NestedTestConfiguration` in Spring Framework, but I'm not sure how it works: https://docs.spring.io/spring-framework/reference/testing/annotations/integration-junit-jupiter.html#integration-testing-annotations-nestedtestconfiguration.

So, would be great to have some real test to play with.
Thanks again.

mtomik:
Hey,
sorry for the late reply.

I have spent some time looking to this and there are few things that I found out:

1.  issue is `AnnotatedElementUtils` used in `SpringIntegrationTestExecutionListener`. There should be used TestContextAnnotationUtils that is able to find annotations in enclosing class. In provided zip, you can find test `FixAnnotationFindIntegrationTest` that is demonstrating that difference   
2.  Null check for MockIntegrationContext (NPE is thrown) in  `SpringIntegrationTestExecutionListener`. I didn't check it further, why that context is missing there.

In provided zip, I have overridden that listener and added both fixes
[...]

PR #9506, opened 2024-09-24
GH-9492: Fix @SpringIntegration in @Nested test

mtomik:
Hi,
this Pull Request aims to fix issue with using @SpringIntegration test annotation in conjuction with @Nested classes
more information can be found here  #9492

### DIFF
--- a/spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
+++ b/spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
@@ -18,10 +18,11 @@ package org.springframework.integration.test.context;
 
 import java.util.Arrays;
 
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
-import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.integration.endpoint.AbstractEndpoint;
 import org.springframework.test.context.TestContext;
+import org.springframework.test.context.TestContextAnnotationUtils;
 import org.springframework.test.context.TestExecutionListener;
 import org.springframework.util.PatternMatchUtils;
 
@@ -39,12 +40,16 @@ class SpringIntegrationTestExecutionListener implements TestExecutionListener {
 	@Override
 	public void prepareTestInstance(TestContext testContext) {
 		SpringIntegrationTest springIntegrationTest =
-				AnnotatedElementUtils.findMergedAnnotation(testContext.getTestClass(), SpringIntegrationTest.class);
+				TestContextAnnotationUtils.findMergedAnnotation(testContext.getTestClass(), SpringIntegrationTest.class);
 
 		String[] patterns = springIntegrationTest != null ? springIntegrationTest.noAutoStartup() : new String[0];
 
 		ApplicationContext applicationContext = testContext.getApplicationContext();
-		MockIntegrationContext mockIntegrationContext = applicationContext.getBean(MockIntegrationContext.class);
+		MockIntegrationContext mockIntegrationContext = handledGetMockIntegrationContext(applicationContext);
+		if (mockIntegrationContext == null) {
+			return;
+		}
+
 		mockIntegrationContext.getAutoStartupCandidates()
 				.stream()
 				.filter(endpoint -> !match(endpoint.getBeanName(), patterns))
@@ -55,7 +60,11 @@ class SpringIntegrationTestExecutionListener implements TestExecutionListener {
 	@Override
 	public void afterTestClass(TestContext testContext) {
 		ApplicationContext applicationContext = testContext.getApplicationContext();
-		MockIntegrationContext mockIntegrationContext = applicationContext.getBean(MockIntegrationContext.class);
+		MockIntegrationContext mockIntegrationContext = handledGetMockIntegrationContext(applicationContext);
+		if (mockIntegrationContext == null) {
+			return;
+		}
+
 		mockIntegrationContext.getAutoStartupCandidates()
 				.forEach(AbstractEndpoint::stop);
 	}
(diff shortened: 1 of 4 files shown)


spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
```diff
@@ -66,4 +75,15 @@ private boolean match(String name, String[] patterns) {
 						.anyMatch(pattern -> PatternMatchUtils.simpleMatch(pattern, name));
 	}
 
+	/**
+	 * to allow nested class execution
+	 */
+	private MockIntegrationContext handledGetMockIntegrationContext(ApplicationContext applicationContext) {
```
artembilan, line 81: See the `MockIntegrationContextCustomizerFactory` for similar `TestContextAnnotationUtils` fix instead.

artembilan (changes requested):
Thank you for looking into this! The problem is really not  so easy to spot.
We ended up with @onobc for the same conclusion as you did.

We will probably ask @sbrannen to improve docs for the `@NestedTestCinfiguration` to mention that this `TestContextAnnotationUtils` is important to use in custom annotation processors if we’d like make them working with `@Nested` as well.

We still would like to see some test suite to cover this functionality and supplement the fix.

You are right, something like AMQP cannot be used in this module, but that still is out of the annotation in the subject responsibility. The listener container you are mentioning in your test is definitely not an `AbstractEndpoint` we worry here about.

spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
onobc: Got it [here](https://github.com/spring-projects/spring-integration/pull/9506/files#diff-7c2fcf66d0bd0e26c5966321b6dd529e60067a1d1c36200d3ca82fdabbeb6e47R42)

onobc:
> Thank you for looking into this! The problem is really not so easy to spot. We ended up with @onobc for the same conclusion as you did.

I agree w/ @artembilan in that It was not easy to spot and you did a great job @mtomik!
 
> We still would like to see some test suite to cover this functionality and supplement the fix.

I went ahead and contributed this in commit 2. Me and @artembilan paired this afternoon and had the beginnings of the tests so I went ahead and wrapped them up and added to this code proposal. I verified the tests fail as expected w/o the fixes.

> 
> You are right, something like AMQP cannot be used in this module, but that still is out of the annotation in the subject responsibility. The listener container you are mentioning in your test is definitely not an `AbstractEndpoint` we worry here about.

sbrannen:
> We will probably ask @sbrannen to improve docs for the `@NestedTestCinfiguration` to mention that this `TestContextAnnotationUtils` is important to use in custom annotation processors if we’d like make them working with `@Nested` as well.

I've created the following issue to address that.

- https://github.com/spring-projects/spring-framework/issues/33586

mtomik:
Thanks guys for quick feedback :+1:

spring-integration-test/src/test/java/org/springframework/integration/test/context/AbstractIntegrationTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.integration.endpoint.ReactiveMessageSourceProducer;
+import org.springframework.messaging.support.GenericMessage;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * Base integration test that specifies a default {@link SpringIntegrationTest}
+ * to be inherited by concrete subclasses.
+ *
+ * @author Chris Bono
+ * @since 6.4.0
```
artembilan, line 32: I think we can treat this as a bug, plus this is just a testing feature, so I'm going to back port the fix down to `6.2.x`.
Therefore `@since 6.2.10` over here.

spring-integration-test/src/test/java/org/springframework/integration/test/context/NestedSpringIntegrationTestAnnotationTests.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Concrete specialization of {@link AbstractIntegrationTest} used to
+ * test inherit and override behavior of {@link SpringIntegrationTest}
+ * when used with {@link Nested} and {@link NestedTestConfiguration}.
+ *
+ * @author Chris Bono
+ * @since 6.4.0
```
artembilan, line 35: DITTO

spring-integration-test/src/test/java/org/springframework/integration/test/context/NestedSpringIntegrationTestAnnotationTests.java
```diff
@@ -0,0 +1,86 @@
...
+	class NestedTestWithoutEnclosingConfiguration {
+
+		@Test
+		void annotationDefinedOnParentOfEnclosingIsInheritedByDefault(@Autowired AbstractEndpoint mockEndpoint) {
+			assertThat(mockEndpoint.isRunning()).isFalse();
+		}
+	}
+
+	@Nested
+	@NestedTestConfiguration(NestedTestConfiguration.EnclosingConfiguration.INHERIT)
```
artembilan, line 55: I don't think we need this specific test since `INHERIT` is there by default.

spring-integration-test/src/test/java/org/springframework/integration/test/context/NestedSpringIntegrationTestAnnotationTests.java
```diff
@@ -0,0 +1,86 @@
...
+	class NestedTestWithInheritEnclosingConfiguration {
+
+		@Test
+		void annotationDefinedOnParentOfEnclosingIsInherited(@Autowired AbstractEndpoint mockEndpoint) {
+			assertThat(mockEndpoint.isRunning()).isFalse();
+		}
+	}
+
+	@Nested
+	@NestedTestConfiguration(NestedTestConfiguration.EnclosingConfiguration.INHERIT)
```
artembilan, line 65: This line is redundant since `INHERIT` is by default.

artembilan (changes requested):
Thank you all for contribution!
Not clear from now how is going to fix the rest of review concerns, but I can do that myself on merge to make all our lives easier :smile.

artembilan (approved):
I'm taking this PR locally for final touches.

Thank you all for contribution!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/3a8e3abc7d2f25fe39290a9758e9d556cc0faf22.
And back-ported to `6.3.x` & `6.2.x` respectively.
Thank you again for the contribution!

PR #9541, opened 2024-10-08
GH-9540: Add idleBetweenTries property to RedisLockRegistry

### LINKED ISSUE #9540: Add idleBetweenTries property to RedisLockRegistry
Hello,
We are using `RedisLockRegistry ` in our project and noticed that, there is a Redis query every 100 milliseconds to check for the lock:

```
@Override
protected boolean tryRedisLockInner(long time) throws InterruptedException {
    long now = System.currentTimeMillis();
    if (time == -1L) {
        while (!obtainLock()) {
            Thread.sleep(100); //NOSONAR
        }
        return true;
    } else {
        long expire = now + TimeUnit.MILLISECONDS.convert(time, TimeUnit.MILLISECONDS);
        boolean acquired;
        while (!(acquired = obtainLock()) && System.currentTimeMillis() < expire) { //NOSONAR
            Thread.sleep(100); //NOSONAR
        }
        return acquired;
    }
}		
```

The issue is that we can't configure the sleep duration between these requests (`Thread.sleep(100); //NOSONAR`), resulting in a large number of Redis requests.
In `JdbcLockRegistry`, there is a similar property, [idleBetweenTries](https://github.com/spring-projects/spring-integration/blob/30afdffb7215ca92fabbb6f500cc0b18989c0572/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java#L102), that allows this configuration.

My suggestion is to add a similar configuration option for `RedisLockRegistry`. Specifically:

1. Add a field:
`private Duration idleBetweenTries = Duration.ofMillis(DEFAULT_IDLE);`
2. Provide a setter method:
```
public void setIdleBetweenTries(Duration idleBetweenTries) {
    Assert.notNull(idleBetweenTries, "'idleBetweenTries' must not be null");
    this.idleBetweenTries = idleBetweenTries;
}
```
3. Use the `idleBetweenTries` in the `tryRedisLockInner()` method:
```
@Override
protected boolean tryRedisLockInner(long time) throws InterruptedException {
    long now = System.currentTimeMillis();
    if (time == -1L) {
        while (!obtainLock()) {
            Thread.sleep(idleBetweenTries.toMillis()); //NOSONAR
        }
        return true;
    } else {
        long expire = now + TimeUnit.MILLISECONDS.convert(time, TimeUnit.MILLISECONDS);
        boolean acquired;
        while (!(acquired = obtainLock()) && System.currentTimeMillis() < expire) { //NOSONAR
            Thread.sleep(idleBetweenTries.toMillis()); //NOSONAR
        }
        return acquired;
    }
}
```

Ichanskiy:
Resolves https://github.com/spring-projects/spring-integration/issues/9540

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.redis.util;
 
 import java.text.SimpleDateFormat;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.ConcurrentModificationException;
 import java.util.Date;
@@ -100,7 +101,11 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 
 	private static final int DEFAULT_CAPACITY = 100_000;
 
+	private static final int DEFAULT_IDLE = 100;
+
 	private final Lock lock = new ReentrantLock();
+
+	private Duration idleBetweenTries = Duration.ofMillis(DEFAULT_IDLE);
 
 	private final Map<String, RedisLock> locks =
 			new LinkedHashMap<>(16, 0.75F, true) {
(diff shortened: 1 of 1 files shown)


artembilan:
Thank you, Olexander, for contribution; looking forward for more!

PR #9542, opened 2024-10-09
GH-9540: Add RedisLockRegistry.idleBetweenTries property

Ichanskiy:
GH-9540: Add RedisLockRegistry.idleBetweenTries property

Fixes: https://github.com/spring-projects/spring-integration/issues/9540


Ichanskiy:
In our microservices, we are using versions 6.2.9 and 6.3.0 of Spring Integration. It would be great to have this improvement in both the 6.2.x and 6.3.x versions accordingly.

PR #9546, opened 2024-10-11
Flag and maybe delete messages after messages have been copied

filiphr:
The reason for this pull request is due to the fact that we have noticed that sometimes the IMAP connection breaks just after the message flags have been set, but the message has not been copied yet. This then leads to the message never being received by (as it has been flagged and the next search will not return it).

e.g.

```
2024-08-08T13:16:02.888-04:00 ERROR 5820 --- [able-mail-30971] c.f.p.e.i.e.mail.MailMessageHandler : Unexpected error occurred in when handling messages for channel processinMailChannel

org.springframework.messaging.MessagingException: failure occurred while polling for mail
at org.springframework.integration.mail.MailReceivingMessageSource.doReceive(MailReceivingMessageSource.java:74) ~[spring-integration-mail-6.1.8.jar:6.1.8]
at org.springframework.integration.endpoint.AbstractMessageSource.receive(AbstractMessageSource.java:142) ~[spring-integration-core-6.1.8.jar:6.1.8]
at org.springframework.integration.endpoint.SourcePollingChannelAdapter.receiveMessage(SourcePollingChannelAdapter.java:212) ~[spring-integration-core-6.1.8.jar:6.1.8]
at org.springframework.integration.endpoint.AbstractPollingEndpoint.doPoll(AbstractPollingEndpoint.java:443) ~[spring-integration-core-6.1.8.jar:6.1.8]
at org.springframework.integration.endpoint.AbstractPollingEndpoint.pollForMessage(AbstractPollingEndpoint.java:412) ~[spring-integration-core-6.1.8.jar:6.1.8]
at org.springframework.integration.endpoint.AbstractPollingEndpoint.lambda$createPoller$4(AbstractPollingEndpoint.java:348) ~[spring-integration-core-6.1.8.jar:6.1.8]
at org.springframework.integration.util.ErrorHandlingTaskExecutor.lambda$execute$0(ErrorHandlingTaskExecutor.java:57) ~[spring-integration-core-6.1.8.jar:6.1.8]
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
at java.base/java.lang.Thread.run(Thread.java:840) ~[na:na]
Caused by: jakarta.mail.MessagingException: IOException while copying message
at jakarta.mail.internet.MimeMessage.<init>(MimeMessage.java:254) ~[jakarta.mail-api-2.1.3.jar:na]
at org.springframework.integration.mail.AbstractMailReceiver$IntegrationMimeMessage.<init>(AbstractMailReceiver.java:658) ~[spring-integration-mail-6.1.8.jar:6.1.8]
at org.springframework.integration.mail.AbstractMailReceiver.postProcessFilteredMessages(AbstractMailReceiver.java:513) ~[spring-integration-mail-6.1.8.jar:6.1.8]
at org.springframework.integration.mail.AbstractMailReceiver.searchAndFilterMessages(AbstractMailReceiver.java:427) ~[spring-integration-mail-6.1.8.jar:6.1.8]
at org.springframework.integration.mail.AbstractMailReceiver.receive(AbstractMailReceiver.java:380) ~[spring-integration-mail-6.1.8.jar:6.1.8]
at org.springframework.integration.mail.MailReceivingMessageSource.doReceive(MailReceivingMessageSource.java:62) ~[spring-integration-mail-6.1.8.jar:6.1.8]
... 9 common frames omitted
Caused by: java.io.IOException: jakarta.mail.FolderClosedException
at com.sun.mail.imap.IMAPInputStream.fill(IMAPInputStream.java:141) ~[angus-mail-1.1.0.jar:na]
at com.sun.mail.imap.IMAPInputStream.read(IMAPInputStream.java:245) ~[angus-mail-1.1.0.jar:na]
at com.sun.mail.imap.IMAPInputStream.read(IMAPInputStream.java:272) ~[angus-mail-1.1.0.jar:na]
at com.sun.mail.imap.IMAPMessage.writeTo(IMAPMessage.java:892) ~[angus-mail-1.1.0.jar:na]
at jakarta.mail.internet.MimeMessage.<init>(MimeMessage.java:246) ~[jakarta.mail-api-2.1.3.jar:na]
... 14 common frames omitted
Caused by: jakarta.mail.FolderClosedException: null
... 19 common frames omitted
```

I could reproduce a similar exception while debuging in the `IntegrationMimeMessage` constructor. If I put a break point there, wait a minute and then continue I would get this exception.

While debugging I could find some things such as

```
java.io.IOException: Connection dropped by server?
	at org.eclipse.angus.mail.iap.ResponseInputStream.readResponse(ResponseInputStream.java:96)
	at org.eclipse.angus.mail.iap.Response.<init>(Response.java:113)
```

I know that it says Spring Integration 6.1.8 here. However, I could reproduce the same with 6.3.5 as well.

I've added a test trying to mimic the behaviour (an `IOException`) when copying the message contents.

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
@@ -503,17 +503,27 @@ public abstract class AbstractMailReceiver extends IntegrationObjectSupport impl
 	}
 
 	private void postProcessFilteredMessages(Message[] filteredMessages) throws MessagingException {
-		setMessageFlags(filteredMessages);
-
-		if (shouldDeleteMessages()) {
-			deleteMessages(filteredMessages);
-		}
 		// Copy messages to cause an eager fetch
 		if (this.headerMapper == null && (this.autoCloseFolder || this.simpleContent)) {
+			Message[] originalMessages = new Message[filteredMessages.length];
 			for (int i = 0; i < filteredMessages.length; i++) {
-				MimeMessage mimeMessage = new IntegrationMimeMessage((MimeMessage) filteredMessages[i]);
+				Message originalMessage = filteredMessages[i];
+				originalMessages[i] = originalMessage;
+				MimeMessage mimeMessage = new IntegrationMimeMessage((MimeMessage) originalMessage);
 				filteredMessages[i] = mimeMessage;
 			}
+			setMessageFlagsAndMaybeDeleteMessages(originalMessages);
+		}
+		else {
+			setMessageFlagsAndMaybeDeleteMessages(filteredMessages);
+		}
+	}
+
+	private void setMessageFlagsAndMaybeDeleteMessages(Message[] messages) throws MessagingException {
+		setMessageFlags(messages);
+
+		if (shouldDeleteMessages()) {
+			deleteMessages(messages);
 		}
 	}
 
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
@@ -504,26 +504,19 @@ public abstract class AbstractMailReceiver extends IntegrationObjectSupport impl
 	}
 
 	private void postProcessFilteredMessages(Message[] filteredMessages) throws MessagingException {
-		// It is more intuitive use a local variable Message[] messages = filteredMessages;
-		// and then call setMessageFlagsAndMaybeDeleteMessages(messages); after the if, i.e. remove the else.
-		// However, in setMessageFlagsAndMaybeDeleteMessages we are calling Message#setFlag and Message#setFlags
-		// which have different implementations in different implementations of Message.
-		// e.g. IMAPMessage has a different implementation of those two methods.
-
 		// Copy messages to cause an eager fetch
+		Message[] messages = filteredMessages;
 		if (this.headerMapper == null && (this.autoCloseFolder || this.simpleContent)) {
-			Message[] originalMessages = new Message[filteredMessages.length];
+			messages = new Message[filteredMessages.length];
 			for (int i = 0; i < filteredMessages.length; i++) {
 				Message originalMessage = filteredMessages[i];
-				originalMessages[i] = originalMessage;
+				messages[i] = originalMessage;
 				MimeMessage mimeMessage = new IntegrationMimeMessage((MimeMessage) originalMessage);
 				filteredMessages[i] = mimeMessage;
 			}
-			setMessageFlagsAndMaybeDeleteMessages(originalMessages);
 		}
-		else {
-			setMessageFlagsAndMaybeDeleteMessages(filteredMessages);
-		}
+
+		setMessageFlagsAndMaybeDeleteMessages(messages);
 	}
 
 	private void setMessageFlagsAndMaybeDeleteMessages(Message[] messages) throws MessagingException {
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
@@ -68,6 +68,7 @@ import org.springframework.util.ObjectUtils;
  * @author Dominik Simmen
  * @author Yuxin Wang
  * @author Ngoc Nhan
+ * @author Filip Hrisafov
  */
 public abstract class AbstractMailReceiver extends IntegrationObjectSupport implements MailReceiver, DisposableBean {
 
@@ -503,6 +504,12 @@ public abstract class AbstractMailReceiver extends IntegrationObjectSupport impl
 	}
 
 	private void postProcessFilteredMessages(Message[] filteredMessages) throws MessagingException {
+		// It is more intuitive use a local variable Message[] messages = filteredMessages;
+		// and then call setMessageFlagsAndMaybeDeleteMessages(messages); after the if, i.e. remove the else.
+		// However, in setMessageFlagsAndMaybeDeleteMessages we are calling Message#setFlag and Message#setFlags
+		// which have different implementations in different implementations of Message.
+		// e.g. IMAPMessage has a different implementation of those two methods.
+
 		// Copy messages to cause an eager fetch
 		if (this.headerMapper == null && (this.autoCloseFolder || this.simpleContent)) {
 			Message[] originalMessages = new Message[filteredMessages.length];
--- a/spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
+++ b/spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
@@ -1005,10 +1005,12 @@ public class ImapMailReceiverTests {
 				.hasMessage("Simulated exception");
 		assertThat(msg1.getFlags().contains(Flag.SEEN)).isFalse();
 		assertThat(msg2.getFlags().contains(Flag.SEEN)).isFalse();
+		assertThat(msg2.getMyTestFlags().contains(Flag.SEEN)).isFalse();
 
 		receiver.receive();
 		assertThat(msg1.getFlags().contains(Flag.SEEN)).isTrue();
 		assertThat(msg2.getFlags().contains(Flag.SEEN)).isTrue();
+		assertThat(msg2.getMyTestFlags().contains(Flag.SEEN)).isTrue();
 		verify(receiver, times(0)).deleteMessages(Mockito.any());
 	}
 
@@ -1051,6 +1053,8 @@ public class ImapMailReceiverTests {
 
 		protected final AtomicInteger exceptionsBeforeWrite;
 
+		protected final Flags myTestFlags = new Flags();
+
 		private TestThrowingMimeMessage(MimeMessage source, int exceptionsBeforeWrite) throws MessagingException {
 			super(source);
 			this.exceptionsBeforeWrite = new AtomicInteger(exceptionsBeforeWrite);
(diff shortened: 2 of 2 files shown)


spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
```diff
@@ -503,17 +503,27 @@ private Object byteArrayToContent(Map<String, Object> headers, ByteArrayOutputSt
...
-				MimeMessage mimeMessage = new IntegrationMimeMessage((MimeMessage) filteredMessages[i]);
+				Message originalMessage = filteredMessages[i];
+				originalMessages[i] = originalMessage;
+				MimeMessage mimeMessage = new IntegrationMimeMessage((MimeMessage) originalMessage);
 				filteredMessages[i] = mimeMessage;
 			}
+			setMessageFlagsAndMaybeDeleteMessages(originalMessages);
+		}
+		else {
+			setMessageFlagsAndMaybeDeleteMessages(filteredMessages);
```
artembilan, line 518: May we consider a local method variable for `Message[]` instead of `else` branch with the same method call?

I mean something like:
```
Message[] messages = filteredMessages;
if (this.headerMapper == null && (this.autoCloseFolder || this.simpleContent)) {
    messages = new Message[filteredMessages.length];
    ...
}
setMessageFlagsAndMaybeDeleteMessages(messages);
```

artembilan (changes requested):
Please, consider to run `gradlew spring-integration-mail:check` locally.
Otherwise we have those Checkstyle failures:
```
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java:516:17: '}' at column 3 should be alone on a line. [RightCurly]

> Task :spring-integration-mail:checkstyleMain

> Task :spring-integration-mail:checkstyleMain FAILED
> Task :spring-integration-stream:compileTestJava
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java:1053:17: Redundant 'public' modifier. [RedundantModifier]
```

Also, ad your name to all the affected classes.

We will consider to back-port it, but that is not going to happen for `6.1.x`.
That one is out of Open Source support already.

spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
filiphr: I was wanted to do that initially as well. However, in the `if` we are creating a new `Message` of the `IntegrationMimeMessage`. This means that when we call `setMessageFlagsAndMaybeDeleteMessages` and we invoke `Message#setFlags` the actual implementation differs. If we look into the `IMAPMessage#setFlags` the implementation looks like

```java
synchronized (getMessageCacheLock()) {
    try {
        IMAPProtocol p = getProtocol();
        checkExpunged(); // Insure that this message is not expunged
        p.storeFlags(getSequenceNumber(), flag, set);
    } catch (ConnectionException cex) {
        throw new FolderClosedException(folder, cex.getMessage());
    } catch (ProtocolException pex) {
        throw new MessagingException(pex.getMessage(), pex);
    }
}
```

whereas the `MimeMessage#setFlags` looks like

```java
if (set)
    flags.add(flag);
else
    flags.remove(flag);
```

I am going to add a comment there to explain why it is done in the way I did it

filiphr:
Thanks for the quick review @artembilan.

>Please, consider to run `gradlew spring-integration-mail:check` locally.

That was my mistake. The checkstyle errors are already fixed.

>Also, add your name to all the affected classes.

Done

>We will consider to back-port it, but that is not going to happen for 6.1.x.
>That one is out of Open Source support already.

We are already upgraded to Spring Boot 3.3 and using Spring Integration 6.3, so back-porting it to at least 6.3 would be more than appreciated for us.

spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
artembilan: I’m not sure why does that matter since you call `setMessageFlagsAndMaybeDeleteMessages` in both cases? So, one local variable for both cases feels right.
Ok, I’ll look into that locally on merge.

filiphr: @artembilan it matters a lot because what will be invoked in the end is not the same. The behaviour of the `setFlags` will change due to the fact that there will be different implementations. 

Prior to this PR the `setFlags` was invoked on the original messages received from the folder. The goal of the PR is to still invoke the `setFlags` on the original messages. However, instead of doing it before copying it, to do it after it has been copied.

I have an idea for a test case. I'll add it to illustrate the problem.

artembilan (changes requested):

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
```diff
@@ -1063,6 +1067,21 @@ public void writeTo(OutputStream os) throws IOException, MessagingException {
 			}
 			super.writeTo(os);
 		}
+
+		@Override
+		public synchronized void setFlags(Flags flag, boolean set) throws MessagingException {
```
artembilan, line 1072: Why do we need to override this if there is that `Message.getFlags()`?

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
```diff
@@ -1015,4 +1047,22 @@ public void close() throws SecurityException {
...
+		protected final AtomicInteger exceptionsBeforeWrite;
+
+		private TestThrowingMimeMessage(MimeMessage source, int exceptionsBeforeWrite) throws MessagingException {
+			super(source);
+			this.exceptionsBeforeWrite = new AtomicInteger(exceptionsBeforeWrite);
+		}
+
+		@Override
+		public void writeTo(OutputStream os) throws IOException, MessagingException {
+			if (this.exceptionsBeforeWrite.decrementAndGet() >= 0) {
```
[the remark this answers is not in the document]
filiphr: You are right. I've adjusted it

artembilan (changes requested):

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
filiphr: This is to show that we are calling the `setFlags` on the original message and not the copied one. The `Flags` can be shared. The `MimeMessage` has

```java
flags = source.getFlags();
if (flags == null)    // make sure flags is always set
    flags = new Flags();
```

So if the source has `Flags` and we call the `setFlags` on the `filteredMessages` then the test will still pass. The goal of the change is to show that the `setFlags` is called on the original message.

artembilan: Doesn't look like you validate somehow in the test if message was original or `IntegrationMimeMessage`.
Therefore this extra precaution is redundant.

artembilan (changes requested):

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
filiphr: There is

```java
assertThat(msg2.getMyTestFlags().contains(Flag.SEEN)).isTrue();
```

In the test.

artembilan: But looks like it is exactly the same as previous one:
```
assertThat(msg2.getFlags().contains(Flag.SEEN)).isTrue();
```

That's why I believe that is really enough.
Because you are not going to deal with mock messages in the real application.

artembilan (changes requested):

spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
filiphr: I added this to validate that the `setFlags` is called on the original message and not on the `IntegrationMimeMessage`. Currently, `GreenMailUtil#newMimeMessage` does not set the flags. However, if you want I can set it up in such case that if the `setMessageFlagsAndMaybeDeleteMessages` is called on the `IntegrationMimeMessage` the test would fail.

```java
assertThat(msg2.getFlags().contains(Flag.SEEN)).isTrue();
```

The test above passes as the `Flags` are shared. But

```java
assertThat(msg2.getMyTestFlags().contains(Flag.SEEN)).isTrue();
```

would fail as it has it's own flags. If you think it is too much, I can remove it

artembilan: OK. You probably can `spy()` on the `GreenMailUtil#newMimeMessage` result to verfiy that exactly its `setFlags()` is called.
It is still look redundant to have our own property when it is there in the super class.

filiphr: good point with the `spy()`. I'll add that

filiphr: I've replaced that with a spy. Hope it is good now.

artembilan:
@filiphr ,

filiphr:
Thanks a lot for the speedy review and integration @artembilan

PR #9556, opened 2024-10-14
Remove not used private variable from ImapIdleChannelAdapter

filiphr:
In e4c1851c070e05ba14979851509c6312ec88d932 the logic for the `ImapIdleChannelAdapter` was changed. When that was done the `ScheduledFuture<?> receivingTask` from the `ImapIdleChannelAdapter` was not removed. This PR just removes the no longer used private variable `receivingTask`


artembilan:
Great catch, @filiphr !

Will be merged when PR build is green.

Thank you!

Issue #3636, opened 2021-09-23
Can  spring-integration provide renewable distributed locks in the future? 


maozhen520:
This is not a bug: 

Can  spring-integration  provide renewable distributed locks in the future? Similar to redisson, it can renew the expiration time of the lock. I know that although I can directly use redisson as a redis client, our project has already used spring-integration, It is good, but we need to use renewable distributed locks, so ask, will spring-integration support renewable lock-related operations later, or why not , I am curious about your ideas

artembilan:
Starting with version `5.4` the `JdbcLockRegistry` already implements such a `RenewableLockRegistry`: https://docs.spring.io/spring-integration/docs/current/reference/html/jdbc.html#jdbc-lock-registry.

You request is really valid and if we can come up with some contract for specific `LockRegistry` we can add this hook over there. Like you brought `RedisLockRegistry` for example. 
I mean we are planning to do all of this, but on demand and when resources permit. 
The contribution is welcome: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc !

Plus we can treat this as a duplicate of this one and close respectively: https://github.com/spring-projects/spring-integration/issues/3272

Issue #2894, opened 2019-04-11
Is this a bug when lock is timeout it throws an excption in spring-integration-redis?


Farwell-Liu:
I use spring-integration-redis for distibution lock and I found that if the lock key is expired, it will throw an exception when I unlock the lock. here is my code:
```
RedisLockRegistry registry = new RedisLockRegistry(getConnectionFactory(), this.registryKey);
            Lock lock = registry.obtain("foo");
            lock.lock();
            try {
                //do something that may be expired
            }
           finally {
               lock.unlock();//it throws an exception if the redis key is expired
           }
}
```

I read the source code of 'unlock' method in RedisLockRegistry class I found that:
```
if (!this.isAcquiredInThisProcess()) {
                        throw new IllegalStateException("Lock was released in the store due to expiration. The integrity of data protected by this lock may have been compromised.");
                    }
```

It seems that it is not a bug. I think it mislead the developers likely, at least it mislead me. so, my question is that why not just ignore case that the redis key is expired. Isn't it more friendly to developers. And then I needn't to embeded a try catch block in finally block above. 

My English is not good enough probably, Did I describe my question clearly?
Could you please explain that for me?  thank you very much!

garyrussell:
The problem is that if a lock expires while the application has it locked, it means that the data protected by the lock is compromised - some other process could have changed it.

The exception is to tell the application that it has happened so that it can take some action to verify the data is ok.

If we just exit silently, the application would not know that the data has been compromised.

artembilan:
We have a ticket to react to the event in case of key expiration: https://jira.spring.io/browse/INT-4286.

But that doesn't mean that we are going to change the logic in this `unlock()` implementation.

We'll think about something internally, but I'm not sure that we can interfere somehow into the locking thread...

Farwell-Liu:
Thank you garyrussell and artembilan! I think that I have undertood the reason.But I am thinking that should you provide method like this:
```
Lock lock = registry.obtain("foo");
lock.lock();
try {
    //do something that may be expired
}
finally {
   ResuntEnum result= lock.unlock();//normally it returns Success,if expired then it returns KeyExpired
   if(result==ResultEnum.KeyExpried){
      //do something to deal with compromised data
   }
}
```

artembilan, did you mean that RedisLockRegistry provide method overload:
`lock.unlock(new KeyExpirationEventMessageListener(){....});//to implement a interface`
Or:
[...]

PR #9557, opened 2024-10-15
add a new redis lock type that renewal expire

NaccOll:
link https://github.com/spring-projects/spring-integration/issues/3636

add a new redis lock type that renewal expire like redission.

Although RenewableLockRegistry provides a renew interface, it is inconvenient for users. Developers hope to have a lock that can be automatically renewed. On the one hand, it can avoid subsequent failures caused by locks that will not expire when abnormal exits, and on the other hand, it can avoid unlock failures caused by lock expired.

In earlier issues, such as https://github.com/spring-projects/spring-integration/issues/2894, some users mentioned this problem, but because it needs to follow the constraints of java.util.concurrent.locks.Lock, and in order to maintain the abstract interface, different lock implementations are not exposed to the outside world (they are all private)

So I hope to add a Redis Lock Type to solve this problem

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/locks/RenewableLockRegistry.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/locks/RenewableLockRegistry.java
@@ -16,12 +16,15 @@
 
 package org.springframework.integration.support.locks;
 
+import org.springframework.scheduling.TaskScheduler;
+
 /**
  * A {@link LockRegistry} implementing this interface supports the renewal
  * of the time to live of a lock.
  *
  * @author Alexandre Strubel
  * @author Artem Bilan
+ * @author Youbin Wu
  *
  * @since 5.4
  */
@@ -33,5 +36,15 @@ public interface RenewableLockRegistry extends LockRegistry {
 	 * @param lockKey The object with which the lock is associated.
 	 */
 	void renewLock(Object lockKey);
+
+	/**
+	 * Set the {@link TaskScheduler} to use for the renewal task.
+	 * When renewalTaskScheduler is set, it will be used to periodically renew the lock to ensure that
+	 * the lock does not expire while the thread is working.
+	 * @param renewalTaskScheduler renew task scheduler
+	 * @since 6.4.0
+	 */
+	default void setRenewalTaskScheduler(TaskScheduler renewalTaskScheduler) {
+	}
 
 }
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -32,6 +32,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.locks.Condition;
@@ -54,6 +55,8 @@ import org.springframework.data.redis.listener.ChannelTopic;
 import org.springframework.data.redis.listener.RedisMessageListenerContainer;
 import org.springframework.data.redis.listener.Topic;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
+import org.springframework.integration.support.locks.RenewableLockRegistry;
+import org.springframework.scheduling.TaskScheduler;
 import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
 import org.springframework.util.Assert;
 import org.springframework.util.ReflectionUtils;
@@ -89,11 +92,12 @@ import org.springframework.util.ReflectionUtils;
  * @author Myeonghyeon Lee
  * @author Roman Zabaluev
  * @author Alex Peelman
+ * @author Youbin Wu
  *
  * @since 4.0
  *
  */
-public final class RedisLockRegistry implements ExpirableLockRegistry, DisposableBean {
+public final class RedisLockRegistry implements ExpirableLockRegistry, DisposableBean, RenewableLockRegistry {
 
 	private static final Log LOGGER = LogFactory.getLog(RedisLockRegistry.class);
 
@@ -137,6 +141,8 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 	 */
 	private Executor executor =
 			Executors.newCachedThreadPool(new CustomizableThreadFactory("redis-lock-registry-"));
+
+	private TaskScheduler renewalTaskScheduler;
 
 	/**
 	 * Flag to denote whether the {@link ExecutorService} was provided via the setter and
@@ -205,6 +211,12 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 	public void setExecutor(Executor executor) {
 		this.executor = executor;
 		this.executorExplicitlySet = true;
+	}
+
+	@Override
+	public void setRenewalTaskScheduler(TaskScheduler renewalTaskScheduler) {
+		Assert.notNull(renewalTaskScheduler, "'renewalTaskScheduler' must not be null");
+		this.renewalTaskScheduler = renewalTaskScheduler;
 	}
 
 	/**
@@ -291,6 +303,26 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 		}
 	}
 
+	@Override
+	public void renewLock(Object lockKey) {
+		String path = (String) lockKey;
+		RedisLock redisLock;
+		this.lock.lock();
+		try {
+			redisLock = this.locks.computeIfAbsent(path, getRedisLockConstructor(this.redisLockType));
+		}
+		finally {
+			this.lock.unlock();
+		}
+		if (redisLock == null) {
+			throw new IllegalStateException("Could not renew mutex at " + path);
+		}
+
+		if (!redisLock.renew()) {
+			throw new IllegalStateException("Could not renew mutex at " + path);
+		}
+	}
+
 	/**
 	 * The mode in which this registry is going to work with locks.
 	 */
@@ -328,14 +360,27 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 				return false
 				""";
 
+		private static final String RENEW_SCRIPT = """
+				if (redis.call('GET', KEYS[1]) == ARGV[1]) then
+					redis.call('PEXPIRE', KEYS[1], ARGV[2])
+					return true
+				end
+				return false
+				""";
+
 		protected static final RedisScript<Boolean>
 				OBTAIN_LOCK_REDIS_SCRIPT = new DefaultRedisScript<>(OBTAIN_LOCK_SCRIPT, Boolean.class);
+
+		public static final RedisScript<Boolean>
+				RENEW_REDIS_SCRIPT = new DefaultRedisScript<>(RENEW_SCRIPT, Boolean.class);
 
 		protected final String lockKey;
 
 		private final ReentrantLock localLock = new ReentrantLock();
 
 		private volatile long lockedAt;
+
+		private volatile ScheduledFuture<?> renewFuture;
 
 		private RedisLock(String path) {
 			this.lockKey = constructLockKey(path);
@@ -454,6 +499,10 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 					LOGGER.debug("Acquired lock; " + this);
 				}
 				this.lockedAt = System.currentTimeMillis();
+				if (RedisLockRegistry.this.renewalTaskScheduler != null) {
+					Duration delay = Duration.ofMillis(RedisLockRegistry.this.expireAfter / 3);
+					this.renewFuture = RedisLockRegistry.this.renewalTaskScheduler.scheduleWithFixedDelay(this::renew, delay);
+				}
 			}
 			return acquired;
 		}
@@ -515,6 +564,7 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 
 				if (Boolean.TRUE.equals(unlinkResult)) {
 					// Lock key successfully unlinked
+					this.stopRenew();
 					return;
 				}
 				else if (Boolean.FALSE.equals(unlinkResult)) {
@@ -525,6 +575,26 @@ public final class RedisLockRegistry implements ExpirableLockRegistry, Disposabl
 			if (!removeLockKeyInnerDelete()) {
 				throw new ConcurrentModificationException("Lock was released in the store due to expiration. " +
 						"The integrity of data protected by this lock may have been compromised.");
+			}
+			else {
+				this.stopRenew();
+			}
+		}
+
+		protected final boolean renew() {
+			boolean res = Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
+					RENEW_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
+					RedisLockRegistry.this.clientId, String.valueOf(RedisLockRegistry.this.expireAfter)));
+			if (!res) {
+				this.stopRenew();
+			}
+			return res;
+		}
+
+		protected final void stopRenew() {
+			if (this.renewFuture != null) {
+				this.renewFuture.cancel(true);
+				this.renewFuture = null;
 			}
 		}
 
--- a/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
+++ b/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
@@ -51,8 +51,10 @@ import org.springframework.data.redis.core.StringRedisTemplate;
 import org.springframework.integration.redis.RedisContainerTest;
 import org.springframework.integration.redis.util.RedisLockRegistry.RedisLockType;
 import org.springframework.integration.test.util.TestUtils;
+import org.springframework.scheduling.concurrent.SimpleAsyncTaskScheduler;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.Mockito.mock;
@@ -66,6 +68,7 @@ import static org.mockito.Mockito.mock;
  * @author Artem Vozhdayenko
  * @author Anton Gabov
  * @author Eddie Cho
+ * @author Youbin Wu
  *
  * @since 4.0
  *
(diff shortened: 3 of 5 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -311,10 +320,11 @@ private Function<String, RedisLock> getRedisLockConstructor(RedisLockType redisL
 		return switch (redisLockType) {
 			case SPIN_LOCK -> RedisSpinLock::new;
 			case PUB_SUB_LOCK -> RedisPubSubLock::new;
+			case SPIN_LOCK_WITH_RENEWAL -> RedisSpinLockWithRenewal::new;
 		};
 	}
 
-	private abstract class RedisLock implements Lock {
+	public abstract class RedisLock implements Lock {
```
artembilan, line 327: Why did you make this `public`?

artembilan (changes requested):
OK. I got an idea.
Thank you!
However I don't think it has to be that complex internally.

Let's see if we can come up with a `setTaskScheduler(TaskScheduler)` on the `RenewableLockRegistry`! With `default` to nothing.
And implement this contract on the `RedisLockRegistry`.
So, if `TaskScheduler` is provided, we just schedule that `renewLock()` at a `scheduleWithFixedDelay()`.

I don't think we need anything else like a new type, tracking those threads and so on. Every `RedisLock` could have that `ScheduledFuture` for itself to be cancelled on unlock and so.

NaccOll:
> OK. I got an idea. Thank you! However I don't think it has to be that complex internally.
> 
> Let's see if we can come up with a `setTaskScheduler(TaskScheduler)` on the `RenewableLockRegistry`! With `default` to nothing. And implement this contract on the `RedisLockRegistry`. So, if `TaskScheduler` is provided, we just schedule that `renewLock()` at a `scheduleWithFixedDelay()`.
> 
> I don't think we need anything else like a new type, tracking those threads and so on. Every `RedisLock` could have that `ScheduledFuture` for itself to be cancelled on unlock and so.

The implementation of RenewableLockRegistry#renewLock should be a single renewal, which is just a wrapper for the renewal of Lock. But this does not solve the problem of calling timing.

redission generates a task when acquiring the lock and delegates it to ExpirationEntry, and removes the task when unlocking or the lock fails.

The complexity cannot be reduced by scheduleAtFixedRate. It is still necessary to start the scheduled task when acquiring the lock, delegate it to ExpirationEntry, and remove it when unlocking or the lock fails. It just changes the loop task to fixed rate. But I feel that loop task is a better solution, which can ensure that each delayed task is new and orderly. If an error occurs, the worst result is that the lock does not exist when unlocking, which is consistent with the previous result. Using scheduleAtFixedRate, I am not sure whether its scheduling order is as I imagined.

Another point you mentioned is whether a new RedisType is needed. The answer is no. But this is just a stopgap measure. A gradual improvement that does not affect the existing logic is more likely to be accepted without facing "why did you change the design". If you guys want to use RenewableLockRegistry instead of the new RedisType, I can modify the code, but I will still use `schedule` instead of `scheduleAtFixedRate`

artembilan:
Pay attention, I said `scheduleWithFixedDelay`, not `Rate`. 
My point was exactly start this schedule when we lock and scheduler is provided . It will do it trick on background until we cancel it with unlock or so. The beauty of renewal lock contract that api is already there and does not affect anything else.
We may make it opposite : I’ll implement my simple ideas and you review. I just really don’t see reason in such a complexity with those extra internal objects a loops.

NaccOll:
@artembilan  Could you review my latest commit to confirm whether it complies with the discussion

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -922,6 +951,14 @@ private void waitForExpire(String key) throws Exception {
 		assertThat(n < 100).as(key + " key did not expire").isTrue();
 	}
 
+	private void waitForExpire2(String key) throws Exception {
+		StringRedisTemplate template = createTemplate();
+		int n = 0;
+		while (n++ < 50 && template.keys(this.registryKey + ":" + key).size() > 0) {
+			Thread.sleep(100);
+		}
```
artembilan, line 959: Doesn't look like this new method has some value, unless it introduces unnecessary thread blocking, including Redis call.

I believe in your second new test we could just use existing `waitForExpire()`.
However, at the same time your `testRenewalExceptionOnExpire()` is fully similar to existing `testExceptionOnExpire()`.
So, that one is redundant.

You might think about adding a new test for the new `renewLock()` API and model it after  `JdbcLockRegistryTests.testLockRenew()` and `testLockRenewLockNotOwned()`, respectively.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -427,6 +429,33 @@ void testExceptionOnExpire(RedisLockType testRedisLockType) throws Exception {
...
 
+	@ParameterizedTest
+	@EnumSource(RedisLockType.class)
+	void testRenewalOnExpire(RedisLockType redisLockType) throws Exception {
+		RedisLockRegistry registry = new RedisLockRegistry(redisConnectionFactory, this.registryKey, 3_000L);
+		registry.setRenewalScheduledExecutor(new ScheduledThreadPoolExecutor(10, new CustomizableThreadFactory("renewal-schedule-thread")));
+		registry.setRedisLockType(redisLockType);
+		Lock lock1 = registry.obtain("foo");
+		assertThat(lock1.tryLock()).isTrue();
+		waitForExpire2("foo");
```
artembilan, line 440: I think this test has to be modeled with a shorter `expireAfter` and instead of Redis call, we indeed should just `Thread.sleep(expireAfter * 2)` and then verify that lock was not expired because we have that renewal task on background.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -454,6 +503,13 @@ private boolean tryRedisLock(long time) throws ExecutionException, InterruptedEx
 					LOGGER.debug("Acquired lock; " + this);
 				}
 				this.lockedAt = System.currentTimeMillis();
+				if (RedisLockRegistry.this.renewalScheduledExecutor != null) {
+					long delay = RedisLockRegistry.this.expireAfter / 3;
+					Runnable task = () -> {
+						renewLock(this.lockKey.substring((RedisLockRegistry.this.registryKey + ':').length()));
+					};
+					RedisLockRegistry.this.renewalScheduledExecutor.scheduleWithFixedDelay(task, delay, delay, TimeUnit.MILLISECONDS);
```
artembilan, line 511: You are missing extracting a `volatile ScheduledFuture<?>` for this call to `cancel(true)` when we `unlock()`.

artembilan (changes requested):

You may also think about adding such an automatic renewal into the `JdbcLockRegistry` where we implement that `RenewableLockRegistry` already.

In the end it would be great to mention this new feature in docs. See `lock-registry.adoc`.
And then a `whats-new.adoc` entry for the whole picture.

Thank you so far for the contribution!

FYI: we have scheduled `6.4.0-RC1` next Tuesday.
So, if you cannot make it for these last day, let us know and we will take it over from where you have left!

NaccOll:
I have modified most of the problems according to the comments, but I do not intend to implement automatic renewal of JdbcLockRegistry.

There are several reasons for this.

1. I don't think it's a good idea to use Jdbc as a distributed lock. In fact, I think it's a temporary solution with only a database component.

2. The current JdbcLock does not have an automatic expiration mechanism. It only stores a creation time in the database and does not expire. Its expiration should be called by the developer on demand ExpirableLockRegistry#expireUnusedOlderThan, which is different from the key expiration of Redis.

3. Introducing an expiration time in JdbcLock will destroy the original design. It is not known how many people rely on this mechanism that does not expire, and there is a lot of new work involved.

4. Without introducing an expiration time, it is just a simple renewal, and the benefits are unclear. Because people who need renewal often call expireUnusedOlderThan themselves, if a global automatic renewal is introduced, it may also cause their expireUnusedOlderThan to be invalidated.

To implement this function, more scenarios need to be considered, but I can't do this at the moment.

I hope the repository can accept this PR, or quickly implement a similar function. As I mentioned in the comments, a few years ago, a user faced the same problem as me. In fact, I initiated this PR because I was limited by the implementation mechanism of RedisLock. Its implementation is based on enumeration and private. If RedisLockType is an interface instead of an enumeration, I will make a custom implementation in my own project. In fact, my current temporary solution is to implement a RedisLockAdator to achieve similar functions, but it is problematic because it can only complete the creation and cancellation of tasks after the actual lock and unlock and cannot obtain the redis value for renewal comparison. Therefore, when the lock of the same key is acquired alternately by different threads, the renewal task of the second acquisition of the lock may be canceled by the release of the first lock.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/RenewableLockRegistry.java
```diff
@@ -16,12 +16,15 @@
...
 
+import org.springframework.scheduling.TaskScheduler;
+
 /**
  * A {@link LockRegistry} implementing this interface supports the renewal
  * of the time to live of a lock.
  *
  * @author Alexandre Strubel
  * @author Artem Bilan
+ * @author NaccOll
```
artembilan, line 27: Forgive me for my official tone, but this does not sound like a legal name to be applied for the license.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/RenewableLockRegistry.java
```diff
@@ -34,4 +37,11 @@ public interface RenewableLockRegistry extends LockRegistry {
 	 */
 	void renewLock(Object lockKey);
 
+	/**
+	 * Set the {@link TaskScheduler} to use for the renewal task.
+	 * when renewalTaskScheduler is set, it will be used to periodically renew the lock to ensure that the lock does not expire while the thread is working.
```
artembilan, line 42: The sentence suppose to start from a capital letter.
Plus our rule that line length not more than `120` symbols.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/RenewableLockRegistry.java
```diff
@@ -34,4 +37,11 @@ public interface RenewableLockRegistry extends LockRegistry {
...
+	 * @param renewalTaskScheduler renew task scheduler
+	 * @since 6.4.0
+	 */
+	default void setRenewalTaskScheduler(TaskScheduler renewalTaskScheduler){}
```
artembilan, line 46: Every member of the class has to be surrounded with blank lines.
I also believe that our `check` Gradle command would not let that `{}` to be pass.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -138,6 +142,11 @@ protected boolean removeEldestEntry(Entry<String, RedisLock> eldest) {
 	private Executor executor =
 			Executors.newCachedThreadPool(new CustomizableThreadFactory("redis-lock-registry-"));
 
+	/**
+	 * An {@link TaskScheduler} to call {@link RenewableLockRegistry#renewLock(Object)} after acquire lock
```
artembilan, line 146: No need in JavaDocs for `private` members.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -207,6 +216,17 @@ public void setExecutor(Executor executor) {
 		this.executorExplicitlySet = true;
 	}
 
+	/**
+	 * Set the {@link TaskScheduler} to use for the renewal task.
+	 * when renewalTaskScheduler is set, it will be used to periodically renew the lock to ensure that the lock does not expire while the thread is working.
+	 * @param renewalTaskScheduler renew task scheduler
+	 * @since 6.4.0
```
artembilan, line 223: No need in duplicating JavaDoc from the super class.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -207,6 +216,17 @@ public void setExecutor(Executor executor) {
...
+	 */
+	@Override
+	public void setRenewalTaskScheduler(TaskScheduler renewalTaskScheduler) {
+		this.renewalTaskScheduler = renewalTaskScheduler;
```
artembilan, line 227: `Assert.notNull(renewalTaskScheduler, "'renewalTaskScheduler' must not be null")`

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -291,6 +311,27 @@ public void destroy() {
...
+		finally {
+			this.lock.unlock();
+		}
+		if (redisLock == null) {
+			throw new IllegalStateException("Could not renew mutex at " + path);
+		}
+
+		if (!redisLock.renew()) {
+			redisLock.stopRenew();
+			throw new IllegalStateException("Could not renew mutex at " + path);
```
artembilan, line 331: I think this logic has to go to the `redisLock.renew()` at all.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -454,6 +511,10 @@ private boolean tryRedisLock(long time) throws ExecutionException, InterruptedEx
 					LOGGER.debug("Acquired lock; " + this);
 				}
 				this.lockedAt = System.currentTimeMillis();
+				if (RedisLockRegistry.this.renewalTaskScheduler != null) {
+					Duration delay = Duration.ofMillis(RedisLockRegistry.this.expireAfter / 3);
+					this.renewFuture = RedisLockRegistry.this.renewalTaskScheduler.scheduleWithFixedDelay(() -> renewLock(this.rawKey), delay);
```
artembilan, line 516: I think we should just call `renew()` from here with the mention above change to `stopRenew()` over there, respectively.
This way we would not need that extra `rawKey` property.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -526,6 +588,22 @@ else if (Boolean.FALSE.equals(unlinkResult)) {
...
+			}
+		}
+
+		public final boolean renew() {
+			return Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
+					RENEW_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
+					RedisLockRegistry.this.clientId, String.valueOf(RedisLockRegistry.this.expireAfter)));
+		}
+
+		public final void stopRenew() {
```
artembilan, line 602: I don't believe these new methods has to be `public`.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -526,6 +588,22 @@ else if (Boolean.FALSE.equals(unlinkResult)) {
...
+			if (this.renewFuture != null) {
+				this.renewFuture.cancel(false);
```
artembilan, line 604: Why don't use `true` instead?
We definitely are looking for the way to break the currently in the flight request.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -427,6 +430,20 @@ void testExceptionOnExpire(RedisLockType testRedisLockType) throws Exception {
 		registry.destroy();
 	}
 
+	@ParameterizedTest
+	@EnumSource(RedisLockType.class)
+	void testRenewalOnExpire(RedisLockType redisLockType) throws Exception {
+		long expireAfter = 3_000L;
```
artembilan, line 436: I don't want the test to be so long. 
How about something like just `300` or even `100`?

src/reference/antora/modules/ROOT/pages/redis.adoc
```diff
@@ -856,4 +856,6 @@ Default.
 The pub-sub is preferred mode - less network chatter between client Redis server, and more performant - the lock is acquired immediately when subscription is notified about unlocking in the other process.
 However, the Redis does not support pub-sub in the Master/Replica connections (for example in AWS ElastiCache environment), therefore a busy-spin mode is chosen as a default to make the registry working in any environment.
 
-Starting with version 6.4, instead of throwing `IllegalStateException`, the `RedisLockRegistry.RedisLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
\ No newline at end of file
+Starting with version 6.4, instead of throwing `IllegalStateException`, the `RedisLockRegistry.RedisLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
+
+Starting with version 6.4, a method `RedisLockRegistry.setRenewalTaskScheduler` is added to configure the scheduler for automatic renewal of locks. When it is set, the lock will be automatically renewed every 1/3 of the expiration time after the lock is successfully acquired, until unlocked or the redis key is removed
```
artembilan, line 861: We prefer every sentence to be on its own line: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line.
Plus officially sentences have to be finished with a period.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -61,6 +61,8 @@ See xref:zeromq.adoc[ZeroMQ Support] for more information.
 Instead of throwing `IllegalStateException`, the `RedisLockRegistry.RedisLock.unlock()` method throws `ConcurrentModificationException` if the ownership of the lock is expired.
 See xref:redis.adoc[Redis Support] for more information.
 
+A method `RedisLockRegistry.setRenewalTaskScheduler` is added to configure the scheduler for automatic renewal of locks. When it is set, the lock will be automatically renewed every 1/3 of the expiration time after the lock is successfully acquired, until unlocked or the redis key is removed
```
artembilan, line 64: No need to go in details in the `whats-new`.
The link above would lead us to the target doc.
So, logically this sentence has to be moved before that `See xref:redis.adoc[Redis Support] for more information.`

artembilan (changes requested):
Thank you for great work so far and such a thorough explanation about JDBC!

Let me know if that is still OK with you to address those latest reviews!

NaccOll:
@artembilan I have modified the code as per the comments and request re-review.

artembilan (approved):
Pulling locally for final review, some clean ups and merge.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/4d08e119032f38f8ee599a52319f5321f11fd454.

@NaccOll ,

PR #9619, opened 2024-10-30
Improve readability of `delete` and `renew` methods of DefaultLockRepository

EddieChoCho:
* Modify `delete` method of DefaultLockRepository: improve readability 
* Modify `renew` method of DefaultLockRepository: improve readability

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
@@ -392,7 +392,7 @@ public class DefaultLockRepository
 	@Override
 	public boolean delete(String lock) {
 		return this.defaultTransactionTemplate.execute(
-				transactionStatus -> this.template.update(this.deleteQuery, this.region, lock, this.id)) > 0;
+				transactionStatus -> this.template.update(this.deleteQuery, this.region, lock, this.id)) == 1;
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
```diff
@@ -392,7 +392,7 @@ public void close() {
 	@Override
 	public boolean delete(String lock) {
 		return this.defaultTransactionTemplate.execute(
-				transactionStatus -> this.template.update(this.deleteQuery, this.region, lock, this.id)) > 0;
+				transactionStatus -> this.template.update(this.deleteQuery, this.region, lock, this.id)) == 1;
```
artembilan, line 395: I am not sure what difference with existing style.
How whatever you have done here helps with readability?

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
EddieChoCho: I think reads can easily understand that only one lock ownership is expected to be deleted.

artembilan: That's debatable. Because you see: for me this does nothing.
However to save both our time discussing this I'll go ahead and merge.

Issue #5123, opened 2010-05-03
Add locking strategy to AbstractMessageGroupStore [INT-1117]


spring-operator:
**[Dave Syer](https://jira.spring.io/secure/ViewProfile.jspa?name=david_syer)** opened **[INT-1117](https://jira.spring.io/browse/INT-1117?redirect=false)** and commented

Global locks might be an interesting feature in general, but the most pressing need is for message store use cases.  It only makes sense if the store uses back-end persistence, so it seems like there is an obvious tie in to the store implementation itself,

---

**Issue Links:**
- #6182 Consider improving message polling in JdbcMessageStore

- #7321 Implement RedisLockRegistry

- #6730 Improve MessageGroupStoreReaper Algorithms

- #5182 Add CorrelatingMessageBarrier

- #6131 Consider refactoring the scope of synchronized blocks in AbstractCorrelatingMessageHandler
  (_**"supersedes"**_)

1 votes, 4 watchers

PR #9680, opened 2024-11-23
fix(RedisMessageStore): RedisMessageStore add lock

NaccOll:
I found that when RedisMessageStore adds and removes messages, it operates on two keys separately, which may cause problems in multi-threading due to non-atomic operations.

Although using redis to delay messages is not a good idea, the abnormal loss of messages in the logs alerted me when the number of requests was not large. By comparing the logs, I found the problem that the message group representing the metadata is not consistent with the actual message.

I discovered this problem at least three years ago, but not many people use redis as a message storage, and there is little feedback from the community. I solved this problem by inheriting RedisMessageStore in my project.

My current solution is to add a lock like SimpleMessageStore, which is also the method used in this pull request. After several years of production environment verification, this is feasible. But this is bound to bring some performance loss, which is why I have not yet initiated a pull request to spring-integrate.  But now I think it is troublesome to duplicate this class between multiple projects, and not all project members are aware of this problem, so I hope to solve this problem at the framework level.

like https://github.com/spring-projects/spring-integration/pull/463


spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -69,12 +82,27 @@ public RedisMessageStore(RedisConnectionFactory connectionFactory) {
...
+	 * Construct {@link RedisMessageStore} based on the provided
+	 * {@link RedisConnectionFactory} and prefix.
+	 * @param connectionFactory the RedisConnectionFactory to use
+	 * @param prefix the key prefix to use, allowing the same broker to be used for
+	 * multiple stores.
+	 * @param lockRegistry the {@link LockRegistry} to use.
+	 * @since 6.4.1
+	 * @see AbstractKeyValueMessageStore#AbstractKeyValueMessageStore(String)
+	 */
+	public RedisMessageStore(RedisConnectionFactory connectionFactory, String prefix, LockRegistry lockRegistry) {
```
artembilan, line 98: I'm fine with `DefaultLockRegistry` as default, but let's make this as a setter! It is optional from object creation perspective, so no need in extra ctor argument!

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -183,6 +211,83 @@ protected Collection<?> doListKeys(String keyPattern) {
...
+			return null;
+		});
+	}
+
+	public <T> T lockExecute(Object groupId, Supplier<T> supplier) {
+		Lock lock = this.lockRegistry.obtain(groupId);
+		try {
+			lock.lockInterruptibly();
+			try {
+				return supplier.get();
```
artembilan, line 279: Let me know, please, why `LockRegistry.executeLocked()` API doesn't work here?
I think you have extracted a common code from the `SimpleMessageStore`, but I guess the `LockRegistry.executeLocked()` could help us a lot. 

I don't say that `SimpleMessageStore` must not be fixed as well, but that is different story for different PR.

artembilan (changes requested):
I think this change somehow related to: https://github.com/spring-projects/spring-integration/issues/5123. At least by the title of the issue: the rest of the discussion went sideways. 

Can we also think about moving this locking logic into a super class `AbstractKeyValueMessageStore`?
This way any impl would benefit immediately.

I also wonder if it is really safe to say that this is not a new feature which should go to the next `6.5` instead of making new bugs from existing...

Thank you for bringing this up!

artembilan (changes requested):
Well, actually, it sounds more like we have to pull the lock usage up to the `AbstractMessageGroupStore`. So, all the impls would benefit.
Including that `SimpleMessageStore` refactoring I have mentioned.

Let me know if you are OK to work on all of that!
Otherwise I'll take it from here.
But again: that is going to be as part of the next `6.5` version.

artembilan:
So, I would say it is official: this PR is about to fix https://github.com/spring-projects/spring-integration/issues/5123.

NaccOll:
@artembilan II'll be working on this for a week or so. If still cannot meet the pull request by then, please you continue with the remaining work.

artembilan:
No problem , @NaccOll , take your time! We have not planned `6.5.0-M1` yet. But that is definitely somewhere next month or so.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
```diff
@@ -206,7 +205,7 @@ public MessageGroupMetadata getGroupMetadata(Object groupId) {
 	}
 
 	@Override
-	public void addMessagesToGroup(Object groupId, Message<?>... messages) {
+	public void addMessagesToGroupInner(Object groupId, Message<?>... messages) {
```
artembilan, line 208: We prefer to call those delegating methods as `do..`.
Therefore `doAddMessagesToGroup()`.
Plus I believe it must not be `public` here.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
```diff
@@ -359,7 +358,7 @@ public void setLastReleasedSequenceNumberForGroup(Object groupId, int sequenceNu
 	}
 
 	@Override
-	public Message<?> pollMessageFromGroup(Object groupId) {
+	public Message<?> pollMessageFromGroupInner(Object groupId) {
```
artembilan, line 361: DITTO about all the new method with `do..` instead, plus their `public` must be `protected`.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
```diff
@@ -195,12 +221,98 @@ public void removeMessagesFromGroup(Object key, Message<?>... messages) {
...
+
+	@Override
+	public Message<?> pollMessageFromGroup(Object groupId) {
+		Assert.notNull(groupId, GROUP_ID_MUST_NOT_BE_NULL);
+		return lockExecute(groupId, () -> pollMessageFromGroupInner(groupId));
+	}
+
+	protected abstract Message<?> pollMessageFromGroupInner(Object groupId);
+
+	protected <T, E extends RuntimeException> T lockExecute(Object groupId, CheckedCallable<T, E> runnable) {
```
artembilan, line 296: I would still prefer `executeLocked()`.

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -37,6 +37,7 @@
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Youbin Wu
```
artembilan, line 40: Looks like no changes in this class, so no need in extra `@author` for nothing.
You have moved the logic in other super classes, and that is really great.

artembilan (changes requested):
This is really cool!
Please, consider to mention such an improvement in the `message-store.adoc`.
I'll come back to you for the `whats-new.adoc` change request as well, when I switch `main` to `6.5.0` - somewhere in the end of month or so. When we done with follow-up bug fixes for `6.4.x` generation.

Thanks

NaccOll:
@artembilan 

In addition to the Lock mechanism, SimpleMessageStore also introduces a semaphore mechanism. My code will cause an error in the `org.springframework.integration.store.SimpleMessageStoreTests#shouldWaitIfGroupCapacity` test case, but I don’t know how to solve this problem. I look forward to your subsequent solution. My work to this location.

spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
```diff
@@ -268,56 +266,37 @@ protected MessageGroup copy(MessageGroup group) {
 	}
 
 	@Override
-	public void addMessagesToGroup(Object groupId, Message<?>... messages) {
```
artembilan, line 271: I think it is OK to revert the previous version of the method and just don't use `doAddMessagesToGroup()` contract.
This way that custom logic for the lock and `UpperBound` would be met and test will work again.

spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
```diff
@@ -162,6 +159,7 @@ public void setCopyOnGet(boolean copyOnGet) {
 		this.copyOnGet = copyOnGet;
 	}
 
+	@Override
 	public void setLockRegistry(LockRegistry lockRegistry) {
```
artembilan, line 163: We don't need this method since it is there in the super class.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
```diff
@@ -56,11 +68,15 @@ public abstract class AbstractMessageGroupStore extends AbstractBatchingMessageG
 
 	private boolean timeoutOnIdle;
 
+	protected LockRegistry lockRegistry;
```
artembilan, line 71: This must be `private` and whenever we use outside of this class has to be replaced by the `protected getLockRegistry()`.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
NaccOll: `isUsed` is the private variable of SimpleMessageStore.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
NaccOll: You are right, even LockRegister does not need to be exposed to subclasses. But SimpleMessageStore is special, it needs lockRegister and has its own semaphore mechanism. I hope you can handle the follow-up work, because I feel there will be many special considerations in it. I will not update this PR again

spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
NaccOll: The code revert is ok, but this leads to the question of whether `doAddMessagesToGroup` is a common design

NaccOll: The code revert is ok, but this leads to the question of whether `doAddMessagesToGroup` is a common design?

spring-integration-core/src/main/java/org/springframework/integration/store/SimpleMessageStore.java
artembilan: Well, I don't think that `Assert.isTrue(!(this.isUsed), "Cannot change the lock registry after the store has been used");` really makes sense.
An unofficial rule for Spring beans is to have them setup only once during configuration phase.
So, calling setters at runtime is a bit like making a trap for yourself.
Therefore having such a big refactoring in this `MessageStore` hierarchy, I would say let's just remove this `isUsed` altogether!

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
artembilan: Please, confirm that you are OK that I'll pick it up whenever you have left.

I'll do that, thought, in new year, when I switch to `6.5` already.

Happy holidays!

artembilan:
Pulling locally for further improvements and possible merge.
Thank you!

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/9962ee49a2e0804ed11c3fe3b66e1e766900e5a4 after some cleanups and fix for the `SimpleMessageStore.addMessagesToGroup()`.

Thank you for contribution as always!

PR #9696, opened 2024-12-05
GH-9694: Remove duplicate SocketChannel initialization

### LINKED ISSUE #9694: Close SocketChannel in TcpNioClientConnectionFactory.buildNewConnection()
**Expected Behavior**
- When a connection attempt fails, the SocketChannel should be properly closed, and resources should be released.
- Socket management should prevent the number of open sockets from increasing indefinitely during retries.

**Current Behavior**
- When the target server is down and `TcpNioClientConnectionFactory.buildNewConnection()` attempts to connect, `java.net.ConnectException: Connection refused: no further information` is thrown. However, the socket connection is not closed, and the connection retry interval causes the number of sockets to increase indefinitely.
- When using `singleUse=false`, `clientMode=true`, and `retryInterval`, the number of open sockets increases indefinitely until the target server is up.

``` java
// TcpNioClientConnectionFactory class

@Override
protected TcpConnectionSupport buildNewConnection() {
	try {
		SocketChannel socketChannel = SocketChannel.open();
		setSocketAttributes(socketChannel.socket());
		connect(socketChannel);
		TcpNioConnection connection =
				this.tcpNioConnectionSupport.createNewConnection(socketChannel, false, isLookupHost(),
						getApplicationEventPublisher(), getComponentName());
		connection.setUsingDirectBuffers(this.usingDirectBuffers);
		connection.setTaskExecutor(getTaskExecutor());
		Integer sslHandshakeTimeout = getSslHandshakeTimeout();
		if (sslHandshakeTimeout != null && connection instanceof TcpNioSSLConnection) {
			((TcpNioSSLConnection) connection).setHandshakeTimeout(sslHandshakeTimeout);
		}
		TcpConnectionSupport wrappedConnection = wrapConnection(connection);
		if (!wrappedConnection.equals(connection)) {
			connection.setSenders(getSenders());
		}
		initializeConnection(wrappedConnection, socketChannel.socket());
		if (getSoTimeout() > 0) {
			connection.setLastRead(System.currentTimeMillis());
		}
		this.channelMap.put(socketChannel, connection);
		wrappedConnection.publishConnectionOpenEvent();
		this.newChannels.add(socketChannel);
		this.selector.wakeup();
		return wrappedConnection;
	}
	catch (IOException e) {
		throw new UncheckedIOException(e);
	}
		throw new UncheckedIOException(new IOException(e));
	}
}

private void connect(SocketChannel socketChannel) throws IOException, InterruptedException {
	socketChannel.configureBlocking(false);
	socketChannel.connect(new InetSocketAddress(getHost(), getPort()));
	boolean connected = socketChannel.finishConnect();
	long timeLeft = getConnectTimeout().toMillis();
	while (!connected && timeLeft > 0) {
		Thread.sleep(5); // NOSONAR Magic #
		connected = socketChannel.finishConnect();
		timeLeft -= 5; // NOSONAR Magic #
	}
	if (!connected) {
	throw new IOException("Not connected after connectTimeout");
		}
}
```

**Context**
- I believe this may cause resource leakage, as each failed connection attempt opens a new socket that is never closed.
- I think that when the server is down and the connection fails, the socket should be properly closed, and resources should be released.
- I'm wondering if adding `socketChannel.close()` in the event of a connection failure would be a good solution, or if I might be misunderstanding the intended behavior of `TcpNioClientConnectionFactory.buildNewConnection()`.

anthologia:
Fixes: https://github.com/spring-projects/spring-integration/issues/9694
Issue link: https://github.com/spring-projects/spring-integration/issues/9694

- Remove duplicate socketChannel initialization
- Properly close the SocketChannel when connection fails
- Prevents socket resource leakage during connection retries

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
@@ -116,8 +116,11 @@ public class TcpNioClientConnectionFactory extends
 		}
 		catch (IOException e) {
 			try {
-				socketChannel.close();
-			} catch (IOException e2) {
+				if (socketChannel != null) {
+					socketChannel.close();
+				}
+			}
+			catch (IOException e2) {
 				logger.error(e2, "Error closing socket channel");
 			}
 			throw new UncheckedIOException(e);
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
@@ -40,6 +40,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Christian Tzolov
  * @author Ngoc Nhan
+ * @author Jooyoung Pyoung
  *
  * @since 2.0
  *
@@ -85,8 +86,9 @@ public class TcpNioClientConnectionFactory extends
 
 	@Override
 	protected TcpConnectionSupport buildNewConnection() {
+		SocketChannel socketChannel = null;
 		try {
-			SocketChannel socketChannel = SocketChannel.open();
+			socketChannel = SocketChannel.open();
 			setSocketAttributes(socketChannel.socket());
 			connect(socketChannel);
 			TcpNioConnection connection =
@@ -113,6 +115,11 @@ public class TcpNioClientConnectionFactory extends
 			return wrappedConnection;
 		}
 		catch (IOException e) {
+			try {
+				socketChannel.close();
+			} catch (IOException e2) {
+				logger.error(e2, "Error closing socket channel");
+			}
 			throw new UncheckedIOException(e);
 		}
 		catch (InterruptedException e) {


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
```diff
@@ -113,6 +115,11 @@ protected TcpConnectionSupport buildNewConnection() {
 			return wrappedConnection;
 		}
 		catch (IOException e) {
+			try {
+				socketChannel.close();
```
artembilan, line 119: Don't we need to check if `socketChannel != null` first?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
```diff
@@ -113,6 +115,11 @@ protected TcpConnectionSupport buildNewConnection() {
...
+			} catch (IOException e2) {
```
artembilan, line 120: Our code style is to have `catch` on a new line.
You can run locally `./gradlew :spring-integration-ip:check`.

artembilan (changes requested):

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
anthologia: Ah right, I missed the null check. Thanks for catching that!

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpNioClientConnectionFactory.java
anthologia: Checked the style guide and fixed it accordingly. Thanks for the feedback :)

artembilan:
@anthologia ,

Issue #9711, opened 2024-12-12
SMB components fail to access files with spaces in its path due to duplicate URL encoding


smitsjelle:
**In what version(s) of Spring Integration are you seeing this issue?**
6.4.0 & 6.2.1 (likely affects older versions too, but tested on these)

**Describe the bug**

When attempting to connect to an SMB share where either the 'share and dir', the 'remote directory' or the file name contains a whitespace, this will be URL-encoded to `%20`. It appears that the URL encoding is applied twice, both by Spring integration as well as by the underlying codelibs/jcifs library. 

Essentially, in [SmbShare](https://github.com/spring-projects/spring-integration/blob/4d84220dadbdaca1592afab5a4725875fe9d33cc/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbShare.java#L66), Spring calls the constructor of its parent [jcifs.smb.SmbFile](https://github.com/codelibs/jcifs/blob/8024c9779bab238c40e3a5546c9628d7a3c70e67/src/main/java/jcifs/smb/SmbFile.java#L385) with the `String url` argument (sidenote: this contructor appears to be deprecated). To obtain the URL string, [SmbConfig#209](https://github.com/spring-projects/spring-integration/blob/4d84220dadbdaca1592afab5a4725875fe9d33cc/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbConfig.java#L209) creates an URI from it, an returns the ASCIIString that maintains the URL encoding. The parent class subsequently creates an URL object out of this already URL-encoded String. 

This appears to cause double encoding, resulting in e.g. an outbound channel adapter with auto create directory `true` creating the directory `space%20directory` instead of the provided argument `space directory`

**To Reproduce**

Using the existing Spring integration unit tests, this can easiliy be reproduced. In [SmbTestSupport](https://github.com/spring-projects/spring-integration/blob/4d84220dadbdaca1592afab5a4725875fe9d33cc/spring-integration-smb/src/test/java/org/springframework/integration/smb/SmbTestSupport.java#L71), change `SHARE_AND_DIR` to `smb share`, and error will occur. Subsequently change the output of `SmbConfig#toUrl()` to a non-url-encoded String (e.g. by replacing line 209 with `return "smb://%s@%s/%s".formatted(domainUserPass, getHostPort(), path);` and it will succeed. 

[...]

PR #9795, opened 2025-01-28
GH-9794: Allow usage of '@' in SMB domain, username and password

### LINKED ISSUE #9794: `@` character is no longer supported in SMB domain, username or password
6.4.1

**Describe the bug**
With the introduction of the fix of [GH-9711](https://github.com/spring-projects/spring-integration/issues/9711) that uses the new `rawUrl` method (thanks for the solution there by the way), I noticed that some existing connections suddenly were unauthorized. On investigation, it appeared that the common denominator is that in the cases of the connection breaking, the password contained an `@` character. With the way the underlying JCIFS library calls the URL constructor, it will trigger the `java.net.Url` code to look for `indexOf('@')`. When the URL resulting from the `rawUrl` method would be `smb://sambaguest:samb@guest@127.0.0.1:445/smb share`, `java.net.URL` will interpret the host part as empty string (see screencap below). This (obviously) subsequently yields a `UnknownHostException`.

Image

**To Reproduce**

Add the following lines to the existing `SmbMessageHistoryTests#testMessageHistory` method, it will fail on the assertion:
```
CIFSContext context = new BaseContext(new PropertyConfiguration(new Properties()));
URL rawUrl = new URL(null, smbSessionFactory.rawUrl(true), context.getUrlHandler());
assertThat(rawUrl.getHost()).isEqualTo("localhost");
```

**Expected behavior**

The `@` character can be used in the domain, username or password as it could before the fix.

smitsjelle:
Fixes: #9794

The `SmbConfig.rawUrl()` method doesn't apply any URL encoding, resulting in parts of the `domainUserPass` that may contain a `@` character to break the URL logic in regard to determining the hostname.

* Modify `SmbConfig.getDomainUserPass(_includePassword)` to conditionally encode the variables. Makes sure to encode the individual parts to not undesirably encode the `;` and `:` characters, breaking other logic.
* Modify `SmbConfig.rawUrl(_includePassword)` and `SmbConfig.createUri(_includePassword)` to call the modified method with the correct `_urlEncode` variable

### DIFF
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbConfig.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2024 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,8 @@ package org.springframework.integration.smb.session;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 
 import jcifs.DialectVersion;
 
@@ -163,16 +165,19 @@ public class SmbConfig {
 		this.smbMaxVersion = _smbMaxVersion;
 	}
 
-	String getDomainUserPass(boolean _includePassword) {
+	String getDomainUserPass(boolean _includePassword, boolean _urlEncode) {
 		String domainUserPass;
+		String username = _urlEncode ? URLEncoder.encode(this.username, StandardCharsets.UTF_8) : this.username;
+		String password = _urlEncode ? URLEncoder.encode(this.password, StandardCharsets.UTF_8) : this.password;
 		if (StringUtils.hasText(this.domain)) {
-			domainUserPass = String.format("%s;%s", this.domain, this.username);
+			String domain = _urlEncode ? URLEncoder.encode(this.domain, StandardCharsets.UTF_8) : this.domain;
+			domainUserPass = String.format("%s;%s", domain, username);
 		}
 		else {
-			domainUserPass = this.username;
+			domainUserPass = username;
 		}
-		if (StringUtils.hasText(this.password)) {
-			domainUserPass += ":" + (_includePassword ? this.password : "********");
+		if (StringUtils.hasText(password)) {
+			domainUserPass += ":" + (_includePassword ? password : "********");
 		}
 		return domainUserPass;
 	}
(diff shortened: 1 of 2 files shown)


spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbConfig.java
```diff
@@ -211,20 +216,24 @@ public final String rawUrl() {
...
 	public final String rawUrl(boolean _includePassword) {
-		String domainUserPass = getDomainUserPass(_includePassword);
+		String domainUserPass = getDomainUserPass(_includePassword, true);
 		String path = cleanPath();
 		return "smb://%s@%s%s".formatted(domainUserPass, getHostPort(), path);
 	}
 
 	private URI createUri(boolean _includePassword) {
-		String domainUserPass = getDomainUserPass(_includePassword);
+		String domainUserPass = getDomainUserPass(_includePassword, false);
```
artembilan, line 236: I wonder if we ever need this part as unencoded.
I even remember that we recommend in docs for e-mail support to have credential encoded before setting them into respective beans.

artembilan (changes requested):
Please, also add your name as `@author` to all the effected classes.

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbConfig.java
smitsjelle: I am not fully familiar with how email is handled (does it also use URL-style connection?) but it feels like doing so would open the door to reintroducing the issue (that GH-9711 fixed) in the first place. Shouldn't we rely on `java.net.Uri` to properly handle the encoding instead of doing it twice on purpose? I took a quick glance at the `Uri` class and it doesn't seem like it attempts to see if encoding is needed, it just always applies it.

artembilan: Right. I see. We use it only like this:
```
public final String getUrl(boolean _includePassword) {
		return createUri(_includePassword).toASCIIString();
	}
```
I guess since it goes to a string in the end and we now have that encoding manually. it feels safe to remove this `createUri()` method altogether.

Unfortunately, we cannot rely on the `URI` in here.
The underlying library has some manipulation with it before creating `URL`.
That's really what the fix was.
I believe I have kept this `createUri()` here really because of its encoding logic.
But now with your fix we do that manually, so no need in anything else.

Makes sense?

smitsjelle: I see where you're coming from. On that note, wouldn't it make more sense to put the current behavior of `rawUrl` to `createUri` and remove the `rawUrl` method? As with this change it isn't really 'raw' anymore anyways?

artembilan: Yeah... I see what you mean.
Unfortunately the `rawUrl()` is a `public` API.
We cannot just drop it off without deprecation and replacement.

I'm very surprised that `URL` requires credentials to be encoded, but at the same time it does encoding on the path and params.

Thinking about it more, I now believe that we can leave that fix as you have done originally.

One `rawUrl()` for SMB client API, and another `getUrl()` for end-user consideration.

I'll pull your changes locally for final review and possible merge.

Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/36e4012a5693f1605be6c1a049807dfcad782dc2.

PR #9824, opened 2025-02-13
Removed deprecated config from test

mohammad1ta:
Removed unused and deprecated sessionConfig from test config.


artembilan:
@mohammad1ta ,

Issue #9878, opened 2025-03-05
ConcurrentModificationException for late-registered AbstractEndpoints


proggler23:
**In what version(s) of Spring Integration are you seeing this issue?**

6.4.2.RELEASE

**Describe the bug**

After executing a test, annotated with `@SpringBootTest `and `@SpringIntegrationTest`, we saw a `ConcurrentModificationException` happening in `SpringIntegrationTestExecutionListener`.

The underlying issue was, that some logic added a new integrationFlow late (after application started). With bad timings, this happened exactly during `afterTestClass` execution currently looping through all AutoStartupCandidates, resulting in the exception.

**To Reproduce**

Register a new bean, extending `AbstractEndpoint `with `autostartup = true` exactly on test shutdown.

**Expected behavior**

Probably only handle Autostartup candidates, that are received during the ApplicationContext initialization (stop adding Autostartup candidates after ApplicationStarted).

**Sample**

[Sample project](https://github.com/proggler23/spring-integration-test-issue-sample)

proggler23:
probably it's already sufficient to just return a copy of the list for `MockIntegrationContext.getAutoStartupCandidates`?

artembilan:
I think your:
```
  override fun doStop() {
        integrationFlowContext.registration(
```
is still valid use-case and it has nothing to do with testing or application shutdown.
We indeed can stop some endpoint manually at runtime and have respective logic over there.
The fact that same `stop()` is used when we shutdown application, is totally different story and has to be mitigated in the target project, respectively.

Such a bean registration during application shutdown has to fails somehow, e.g. with a:
```
/**
 * Exception thrown in case of a bean being requested despite
 * bean creation currently not being allowed (for example, during
 * the shutdown phase of a bean factory).
 *
 * @author Juergen Hoeller
 * @since 2.0
 */
@SuppressWarnings("serial")
public class BeanCreationNotAllowedException extends BeanCreationException {
```
But again: that is different story.

I think we should rework `MockIntegrationContext` to deal only with endpoints registered before application context is fully initialized.
In other words: do not react to those provided at runtime, e.g. via the mentioned `register()` API from the `IntegrationFlowContext`.
That's first.

Another side of the problem if we really need that `SpringIntegrationTestExecutionListener.afterTestClass()`.
Since we talk about the application context shutdown, so all beans will be stopped anyway.
[...]

PR #9879, opened 2025-03-06
GH-9878: Fix ConcurrentModificationException in SpringIntegrationTest…

proggler23:
…ExecutionListener

    Fixes: gh-9878

    * return new list in MockIntegrationContext.getAutoStartupCandidates()

### DIFF
--- a/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
+++ b/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
@@ -19,6 +19,7 @@ package org.springframework.integration.test.context;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
 import reactor.core.publisher.Mono;
@@ -59,6 +60,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Artem Bilan
  * @author Yicheng Feng
+ * @author Alexander Hain
  *
  * @since 5.0
  *
@@ -111,7 +113,7 @@ public class MockIntegrationContext implements BeanPostProcessor, SmartInitializ
 	}
 
 	List<AbstractEndpoint> getAutoStartupCandidates() {
-		return this.autoStartupCandidates;
+		return Collections.unmodifiableList(this.autoStartupCandidates);
 	}
 
 	/**


spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -111,7 +111,7 @@ public void afterSingletonsInstantiated() {
 	}
 
 	List<AbstractEndpoint> getAutoStartupCandidates() {
-		return this.autoStartupCandidates;
+		return new ArrayList<>(this.autoStartupCandidates);
```
artembilan, line 114: I think I'm OK with what you are suggesting, but this has to be like `Collections.unmodifiableList()` instead.
We indeed do like that in many other places where we return a collection from the object.
I missed this here because it is not public API, but as we see turns out it does not matter if we may deal with concurrent modifications. 😄

spring-integration-test/src/test/java/org/springframework/integration/test/context/MockIntegrationContextTests.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.integration.dsl.context.IntegrationFlowContext;
+import org.springframework.integration.endpoint.AbstractEndpoint;
+import org.springframework.integration.gateway.MessagingGatewaySupport;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+@SpringJUnitConfig
+@ContextConfiguration(classes = MockIntegrationContextTests.Config.class)
+@SpringIntegrationTest
+public class MockIntegrationContextTests {
```
artembilan, line 17: Let's just remove this new void class altogether!
I'm really not sure that registering beans in  `start()/stop()` is OK idea, so I don't want to promote such a pattern with our tests.
Plus the fix is really simple and it is obvious that it is going to work in your project.

artembilan (changes requested):

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
proggler23: ok, I was just not returning unmodifiableList to not break any user, who is in fact modifying. but you're right, it's not public anyways

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -16,10 +16,7 @@
 
 package org.springframework.integration.test.context;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
+import java.util.*;
```
artembilan, line 19: Sorry, no asterisk imports.
Also, please, add your name to the `@author` list of this class.

artembilan (changes requested):

artembilan:
@proggler23 ,

PR #9945, opened 2025-04-07
doc: update code example for Deferred Acknowledgment Pollable Message.

mjd507:
… Source

The current example is misleading, FunctionalInterface can only have one abstract method.


artembilan:
@mjd507 ,

PR #9946, opened 2025-04-08
Apply Junit5BestPractices in spring-integration-mail

mjd507:
use open-rewrite Junit5BestPractices recipe to rewrite test for spring-integration-mail module

### DIFF
--- a/spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
+++ b/spring-integration-mail/src/test/java/org/springframework/integration/mail/ImapMailReceiverTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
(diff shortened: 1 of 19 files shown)


mjd507:
discussion: https://github.com/spring-projects/spring-integration/discussions/9947

artembilan (changes requested):
Would you mind to include some details into commit message saying "what" and "why".

Because what I see so far is just removal of `public` word from those test methods.
While it is convenient to not use it and less words to ready, I find this as silly best practice.
And discussing it takes more time then just leave it as is.
Such a change just don't bring any benefits.

We definitely can accept your change and merge it since you have already dedicated some effort to bring this into life.
But, please, consider in the future to not waste your time for such a pointless change.
Unless that was an automatic pass and we indeed can go ahead and apply it everywhere in the project.

This is not my first discussion about applying OpenRewrite and I think that applying its rules is OK to satisfy community expectations. But that has to be a reasonable amount of the effort.

Thanks for understanding.

mjd507:
Totally agreed with you, I will dig more into this.

PR #9997, opened 2025-04-28
Fix exception message for method `doPoll` in JpaExecutor

mjd507:
using `@ParameterizedTest` to aggregate similar unit test cases

### DIFF
--- a/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
+++ b/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
```diff
@@ -569,7 +569,7 @@ else if (this.entityClass != null) {
 		else {
 			throw new IllegalStateException("For the polling operation, one of "
 					+ "the following properties must be specified: "
-					+ "query, namedQuery or entityClass.");
+					+ "jpaQuery, nativeQuery, namedQuery or entityClass.");
```
artembilan, line 572: This is good catch.
Fine with this.

artembilan (changes requested):
Sorry, I cannot accept those `@ParameterizedTest` changes.
Doesn't look like it makes our life easier: so much new `if...else` logic to satisfy.
I would prefer to have duplication in different test method, but as linear as possible, then some extra logic to meet `@ParameterizedTest` set requirements.
This also does not make it more flexible. If we add a new scenario, and would try to cover it in the generic `@ParameterizedTest` method, that may lead to failure for existing scenarios.
We may think that there is a bug in the main code, but it could simply be a problem of the test method. Just because it is too complex to deal with all the  `@ParameterizedTest` variants.

Just my opinion where I would prefer to not do such a refactoring.

Thanks

mjd507:
Thank you, actually I also have this feeling, removed

PR #10002, opened 2025-05-02
Remove direct initialization of IntegrationProperties In IntegrationObjectSupport

mjd507:
In class `IntegrationObjectSupport`,  the field `integrationProperties` is directly initialized when the class is instantiated.
`private IntegrationProperties integrationProperties = new IntegrationProperties();`

I think it is redundant,  it will be replaced in method `afterPropertiesSet()`.
```
@Override
public final void afterPropertiesSet() {
    this.integrationProperties = IntegrationContextUtils.getIntegrationProperties(this.beanFactory);
    ....
}
```

based on this, I made the changes. and fix the broken unit tests after this change.

I assume any instance of IntegrationObjectSupport will be a Bean in Spring ApplicationContext.  

correct me if wrong.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
@@ -81,7 +81,7 @@ public abstract class IntegrationObjectSupport implements ComponentSourceAware, 
 
 	private TaskScheduler taskScheduler;
 
-	private IntegrationProperties integrationProperties = new IntegrationProperties();
+	private IntegrationProperties integrationProperties;
 
 	private ConversionService conversionService;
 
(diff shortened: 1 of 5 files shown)


mjd507:
similarly, for `messageBuilderFactory`, it will be initialized in `afterPropertiesSet()`,  but we also have a null check and initialization in the `getMessageBuilderFactory()` 
```
	protected MessageBuilderFactory getMessageBuilderFactory() {
		if (this.messageBuilderFactory == null) {
			this.messageBuilderFactory = new DefaultMessageBuilderFactory();
		}
		return this.messageBuilderFactory;
	}
```

confused whether this method is also intended to be called without first calling the `afterPropertiesSet()` method.

artembilan (changes requested):
> I assume any instance of IntegrationObjectSupport will be a Bean in Spring ApplicationContext.

That's exactly why we have it initialized by default.
No guarantee that all the instances are beans.
Plus it really doesn't cause any harm.

Same conclusion for `messageBuilderFactory` property.

Therefore I don't see a reason in the change you propose.

mjd507:
thanks for clarification, closed this.

PR #10009, opened 2025-05-06
Migrate `spring-integration-core` tests from Junit 4 to Junit Jupiter

mjd507:

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/channel/DatatypeChannelTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/channel/DatatypeChannelTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +21,7 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.Set;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
@@ -45,6 +45,7 @@ import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.messaging.support.GenericMessage;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * @author Mark Fisher
@@ -61,10 +62,11 @@ public class DatatypeChannelTests {
 		assertThat(channel.send(new GenericMessage<String>("test"))).isTrue();
 	}
 
-	@Test(expected = MessageDeliveryException.class)
+	@Test
 	public void unsupportedTypeAndNoConversionService() {
 		MessageChannel channel = createChannel(Integer.class);
-		channel.send(new GenericMessage<String>("123"));
+		assertThatThrownBy(() -> channel.send(new GenericMessage<String>("123")))
+				.isInstanceOf(MessageDeliveryException.class);
 	}
 
 	@Test
@@ -77,14 +79,15 @@ public class DatatypeChannelTests {
 		assertThat(channel.send(new GenericMessage<String>("123"))).isTrue();
 	}
 
-	@Test(expected = MessageDeliveryException.class)
+	@Test
 	public void unsupportedTypeAndConversionServiceDoesNotSupport() {
 		QueueChannel channel = createChannel(Integer.class);
 		ConversionService conversionService = new DefaultConversionService();
 		DefaultDatatypeChannelMessageConverter converter = new DefaultDatatypeChannelMessageConverter();
 		converter.setConversionService(conversionService);
 		channel.setMessageConverter(converter);
-		assertThat(channel.send(new GenericMessage<Boolean>(Boolean.TRUE))).isTrue();
+		assertThatThrownBy(() -> assertThat(channel.send(new GenericMessage<Boolean>(Boolean.TRUE))).isTrue())
+				.isInstanceOf(MessageDeliveryException.class);
 	}
 
 	@Test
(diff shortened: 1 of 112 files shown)


spring-integration-core/src/test/java/org/springframework/integration/filter/MessageFilterTests.java
```diff
@@ -133,18 +136,21 @@ public void filterDiscardsMessageAndThrowsException() throws Exception {
...
+				throw e;
+			}
+			finally {
+				Message<?> reply = discardChannel.receive(0);
+				assertThat(reply).isNotNull();
+				assertThat(reply).isEqualTo(message);
+				assertThat(outputChannel.receive(0)).isNull();
+			}
+
+		}).isInstanceOf(MessageRejectedException.class);
```
artembilan, line 153: This is a bit awkward situation.
I guess it is hard to get rid of `finally` block, so let's see if extracting exception into a variable from that `catch` block, and then assert it in the end would make the test easier to digest!

spring-integration-core/src/test/java/org/springframework/integration/message/ExpressionEvaluatingMessageHandlerTests.java
```diff
@@ -76,20 +77,23 @@ public void validExpressionWithSomeArgs() {
...
+		handler.afterPropertiesSet();
+		assertThatThrownBy(() -> {
+			try {
+				handler.handleMessage(message);
+			}
+			catch (MessagingException e) {
+				assertThat(message).isEqualTo(e.getFailedMessage());
+				throw e;
+			}
+		}).isInstanceOf(MessagingException.class);
```
artembilan, line 95: There  is an AssertJ API to extract object properties smooth way without this composition which maybe lost.
Something like:
```
		assertThatExceptionOfType(MessagingException.class)
				.isThrownBy(() -> handler.handleMessage(message))
				.extracting(MessagingException::getFailedMessage)
				.isEqualTo(message);
```

spring-integration-core/src/test/java/org/springframework/integration/resource/ResourceInboundChannelAdapterParserTests.java
```diff
@@ -82,9 +83,10 @@ public void testDefaultConfig() {
 		context.close();
 	}
 
-	@Test(expected = BeanCreationException.class)
+	@Test
 	public void testDefaultConfigNoLocationPattern() {
-		new ClassPathXmlApplicationContext("ResourcePatternResolver-config-fail.xml", getClass()).close();
+		assertThatThrownBy(() -> new ClassPathXmlApplicationContext("ResourcePatternResolver-config-fail.xml", getClass()).close())
```
artembilan, line 88: I think with this AssertJ API we don't need that `close()` any more.
Previously it was a code quality question to have `Closable` resource closed.
The fact that it ended up with an exception was not visible for that tool.

spring-integration-core/src/test/java/org/springframework/integration/router/RecipientListRouterTests.java
```diff
@@ -120,20 +121,22 @@ public void sendFailureOnFirstRecipientTriggersExceptionByDefault() {
...
+		assertThatThrownBy(() -> {
+			try {
+				router.handleMessage(message);
+			}
+			catch (RuntimeException e) {
+				Message<?> result1a = channelA.receive(0);
+				assertThat(result1a).isNotNull();
+				assertThat(result1a.getPayload()).isEqualTo("blocker");
+				assertThat(channelB.receive(0)).isNull();
+				assertThat(channelC.receive(0)).isNull();
```
artembilan, line 133: Doesn't look like this logic has to be done inside `catch` block.
We simply can move all of these assertions after `assertThatExceptionOfTyle()`.
This way the code in this test would be much easier to read.

spring-integration-core/src/test/java/org/springframework/integration/router/RecipientListRouterTests.java
```diff
@@ -346,17 +353,21 @@ public void applySequenceEnabled() {
...
 	}
 
-	@Test(expected = IllegalArgumentException.class)
+	@Test
 	public void nullChannelListRejected() {
 		RecipientListRouter router = new RecipientListRouter();
-		router.setChannels(null);
+		assertThatThrownBy(() -> {
+			router.setChannels(null);
+		}).isInstanceOf(IllegalArgumentException.class);
```
artembilan, line 361: Doesn't look like we need body for the lambda. I mean `{}` and `;` in the end of sentence are redundant.

artembilan (changes requested):
This is huge, @mjd507 !
Thank you very much!

Honestly, all my review is just nit-pick and we simply can accept your changes as is.
Those are just recommendations for the future.
Let me know if you are going to address them or let's move on for other more important tasks!

I wonder if you automated such a migration somehow or what 😄

mjd507:
hi @artembilan 

Yes, please accept as is.

I will address the pr comments this evening, probably your tomorrow morning time.

Not fully automated, I utilize the open-rewrite tool but only applies its simplest recipe, which is change the @Test to Jupiter package, then I change all changed files manually, really huge😉

spring-integration-core/src/test/java/org/springframework/integration/channel/DatatypeChannelTests.java
```diff
@@ -77,14 +79,15 @@ public void unsupportedTypeButConversionServiceSupports() {
...
+	@Test
 	public void unsupportedTypeAndConversionServiceDoesNotSupport() {
 		QueueChannel channel = createChannel(Integer.class);
 		ConversionService conversionService = new DefaultConversionService();
 		DefaultDatatypeChannelMessageConverter converter = new DefaultDatatypeChannelMessageConverter();
 		converter.setConversionService(conversionService);
 		channel.setMessageConverter(converter);
-		assertThat(channel.send(new GenericMessage<Boolean>(Boolean.TRUE))).isTrue();
+		assertThatExceptionOfType(MessageDeliveryException.class)
+				.isThrownBy(() -> assertThat(channel.send(new GenericMessage<Boolean>(Boolean.TRUE))).isTrue());
```
artembilan, line 90: The assertion in the lambda does not make sense.
If we expect a `MessageDeliveryException`, then `send` is never going to be `true`.
It is just suppressed by that exception:
```suggestion
				.isThrownBy(() -> channel.send(new GenericMessage<Boolean>(Boolean.TRUE)));
```

artembilan (approved):
Will apply my suggestion on merge.
Thanks

PR #10019, opened 2025-05-13
GH-9988: Add FileExistsMode expression support

### LINKED ISSUE #9988: Support fileExistModeExpression() in Ftp.outboundGateway
**Expected Behavior**
The `Ftp.outboundGateway` should support a `fileExistModeExpression()` method that accepts a SpEL expression to dynamically determine the FileExistsMode at runtime. This would allow users to conditionally choose between `REPLACE`, `APPEND`, `FAIL`, or other file exist modes based on message content or context variables.

**Current Behavior**
Currently, the `Ftp.outboundGateway` only supports `fileExistMode()` which accepts a static FileExistsMode value. This mode is fixed at configuration time and cannot be changed dynamically during message processing.

**Context**
I'd be willing to implement this feature if the team agrees it would be valuable to the Spring Integration community.

anthologia:
Fixes https://github.com/spring-projects/spring-integration/issues/9988

In `AbstractRemoteFileOutboundGateway`:
- Add `fileExistsModeExpression` field and setter methods
- Use `resolveFileExistsMode` in put and get operations
- Remove logic that disabled temporary filenames when setting `APPEND` mode

In `RemoteFileTemplate`:
- Remove exception validation when APPEND mode is used with temporary filenames
- Modify logic to skip applying temporaryFileSuffix in APPEND mode

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
@@ -388,6 +388,20 @@ public abstract class RemoteFileOutboundGatewaySpec<F, S extends RemoteFileOutbo
 	}
 
 	/**
+	 * Specify a {@link Function} to determine the action to take when files already exist.
+	 * Expression evaluation should return a {@link FileExistsMode} or a String representation.
+	 * Used for GET and MGET operations when the file already exists locally,
+	 * or PUT and MPUT when the file exists on the remote system.
+	 * @param fileExistsModeFunction the {@link Function} to use.
+	 * @param <P> the expected payload type.
+	 * @return the Spec.
+	 * @since 6.5
+	 */
+	public <P> S fileExistsModeFunction(Function<Message<P>, String> fileExistsModeFunction) {
+		return remoteDirectoryExpression(new FunctionExpression<>(fileExistsModeFunction));
+	}
+
+	/**
 	 * Determine whether the remote directory should automatically be created when
 	 * sending files to the remote system.
 	 * @param autoCreateDirectory true to create the directory.
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
@@ -397,8 +397,8 @@ public abstract class RemoteFileOutboundGatewaySpec<F, S extends RemoteFileOutbo
 	 * @return the Spec.
 	 * @since 6.5
 	 */
-	public <P> S fileExistsModeFunction(Function<Message<P>, String> fileExistsModeFunction) {
-		return remoteDirectoryExpression(new FunctionExpression<>(fileExistsModeFunction));
+	public <P> S fileExistsModeFunction(Function<Message<P>, Object> fileExistsModeFunction) {
+		return fileExistsModeExpression(new FunctionExpression<>(fileExistsModeFunction));
 	}
 
 	/**
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 the original author or authors.
+ * Copyright 2016-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -44,6 +44,7 @@ import org.springframework.messaging.Message;
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Jooyoung Pyoung
  *
  * @since 5.0
  */
@@ -355,6 +356,34 @@ public abstract class RemoteFileOutboundGatewaySpec<F, S extends RemoteFileOutbo
 	 */
 	public S fileExistsMode(FileExistsMode fileExistsMode) {
 		this.target.setFileExistsMode(fileExistsMode);
+		return _this();
+	}
+
+	/**
+	 * Specify a SpEL expression to determine the action to take when files already exist.
+	 * Expression evaluation should return a {@link FileExistsMode} or a String representation.
+	 * Used for GET and MGET operations when the file already exists locally,
+	 * or PUT and MPUT when the file exists on the remote system.
+	 * @param fileExistsModeExpression a SpEL expression to evaluate the file exists mode
+	 * @return the Spec.
+	 * @since 6.5
+	 */
+	public S fileExistsModeExpression(Expression fileExistsModeExpression) {
+		this.target.setFileExistsModeExpression(fileExistsModeExpression);
+		return _this();
+	}
+
+	/**
+	 * Specify a SpEL expression to determine the action to take when files already exist.
+	 * Expression evaluation should return a {@link FileExistsMode} or a String representation.
+	 * Used for GET and MGET operations when the file already exists locally,
+	 * or PUT and MPUT when the file exists on the remote system.
+	 * @param fileExistsModeExpression the String in SpEL syntax.
+	 * @return the Spec.
+	 * @since 6.5
+	 */
+	public S fileExistsModeExpression(String fileExistsModeExpression) {
+		this.target.setFileExistsModeExpressionString(fileExistsModeExpression);
 		return _this();
 	}
 
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
@@ -565,8 +565,8 @@ public class RemoteFileTemplate<F> implements RemoteFileOperations<F>, Initializ
 		// write remote file first with temporary file extension if enabled
 
 		String tempFilePath = tempRemoteFilePath;
-		if (!FileExistsMode.APPEND.equals(mode)) {
-			tempFilePath += this.useTemporaryFileName ? this.temporaryFileSuffix : "";
+		if (!FileExistsMode.APPEND.equals(mode) && this.useTemporaryFileName) {
+			tempFilePath += this.temporaryFileSuffix;
 		}
 
 		if (this.autoCreateDirectory) {
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2024 the original author or authors.
+ * Copyright 2013-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -63,6 +63,7 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Alen Turkovic
+ * @author Jooyoung Pyoung
  *
  * @since 3.0
  *
@@ -303,8 +304,6 @@ public class RemoteFileTemplate<F> implements RemoteFileOperations<F>, Initializ
 
 	private String send(Message<?> message, String subDirectory, FileExistsMode mode) {
 		Assert.notNull(this.directoryExpressionProcessor, "'remoteDirectoryExpression' is required");
-		Assert.isTrue(!FileExistsMode.APPEND.equals(mode) || !this.useTemporaryFileName,
-				"Cannot append when using a temporary file name");
 		Assert.isTrue(!FileExistsMode.REPLACE_IF_MODIFIED.equals(mode),
 				"FilExistsMode.REPLACE_IF_MODIFIED can only be used for local files");
 		final StreamHolder inputStreamHolder = payloadToInputStream(message);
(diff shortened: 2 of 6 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
```diff
@@ -565,7 +564,10 @@ private void sendFileToRemoteDirectory(InputStream inputStream, String temporary
 		String tempRemoteFilePath = temporaryRemoteDirectory + fileName;
 		// write remote file first with temporary file extension if enabled
 
-		String tempFilePath = tempRemoteFilePath + (this.useTemporaryFileName ? this.temporaryFileSuffix : "");
+		String tempFilePath = tempRemoteFilePath;
+		if (!FileExistsMode.APPEND.equals(mode)) {
+			tempFilePath += this.useTemporaryFileName ? this.temporaryFileSuffix : "";
```
artembilan, line 569: Why do we need to append empty string to the end if not `this.useTemporaryFileName`?
How about this instead:
```
if (!FileExistsMode.APPEND.equals(mode) &&  this.useTemporaryFileName) {
			tempFilePath += this.temporaryFileSuffix;
}
```

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
```diff
@@ -1351,6 +1378,15 @@ protected String getRemoteFilename(String remoteFilePath) {
 		}
 	}
 
+	private FileExistsMode resolveFileExistsMode(Message<?> message) {
+		if (this.fileExistsModeExpression != null) {
+			EvaluationContext evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
+			evaluationContext.setVariable("fileExistsMode", this.fileExistsMode);
```
artembilan, line 1384: Why do we need this variable in the expression?
In most cases no one is going to care about whatever is there by default in the gateway setting.
And with that we could use a global `EvaluationContext` created only once in the `doInit()`.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
```diff
@@ -1351,6 +1378,15 @@ protected String getRemoteFilename(String remoteFilePath) {
...
+			return this.fileExistsModeExpression.getValue(evaluationContext, message, FileExistsMode.class);
```
artembilan, line 1385: I also think we should be able to allow to evaluate expression to just string and then try to resolve enum manually.
This way our contract for this expression would be much friendly for end-users.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
```diff
@@ -303,8 +304,6 @@ public String send(Message<?> message, String subDirectory, FileExistsMode... mo
 
 	private String send(Message<?> message, String subDirectory, FileExistsMode mode) {
 		Assert.notNull(this.directoryExpressionProcessor, "'remoteDirectoryExpression' is required");
-		Assert.isTrue(!FileExistsMode.APPEND.equals(mode) || !this.useTemporaryFileName,
-				"Cannot append when using a temporary file name");
```
artembilan, line 307: I suggest to make a warning message in the `doInit()` when this condition is `true`.
We still gong to ignore, but users would know that it does not make sense to use both `useTemporaryFileName` and `FileExistsMode.APPEND`

artembilan (changes requested):
Please, consider to describe this new feature in the: https://docs.spring.io/spring-integration/reference/ftp/rft.html#page-title. (`src/reference/antora/modules/ROOT/pages/sftp/rft.adoc` file)

And also mention new feature in the `whats-new.adoc`.

And as a bonus we also can consider to expose this new option on the `RemoteFileOutboundGatewaySpec` for Java DSL.

Thanks

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
anthologia: I've removed the unnecessary variable and implemented a global `EvaluationContext` that's created once in `doInit()`. 

Additionally, I was thinking that in the long term, it might be beneficial to move the Expression evaluation logic to the `IntegrationObjectSupport` class. This could standardize expression evaluation across Spring Integration and reduce duplicate code in each component. Would this be something worth considering?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
anthologia: I should have considered user-friendliness more! The expression can now return either a `FileExistsMode` enum value or a String representation (case-insensitive).  Would it be better not to support case-insensitive matching?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileTemplate.java
anthologia: I've added a warning message in `doInit()` when `APPEND` mode is configured with `useTemporaryFileName` is `true`. Thanks for the good suggestion!

anthologia:
I've implemented all these suggestions:
- Added documentation to the `rft.adoc` and `whats-new.adoc`
- Implemented Java DSL support in `RemoteFileOutboundGatewaySpec` with both Expression and String parameter variants

Thank you for the helpful guidance! Please let me know if any further adjustments are needed.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
artembilan: Feel free to raise a GH issue with more info or even draft some PoC.
And we will go from there this or other way.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
```diff
@@ -1351,12 +1387,39 @@ protected String getRemoteFilename(String remoteFilePath) {
...
+			}
+		}
+		return this.fileExistsMode;
+	}
+
 	private File generateLocalDirectory(Message<?> message, String remoteDirectory) {
-		EvaluationContext evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
 		if (remoteDirectory != null) {
-			evaluationContext.setVariable("remoteDirectory", remoteDirectory);
+			this.standardEvaluationContext.setVariable("remoteDirectory", remoteDirectory);
```
artembilan, line 1420: This is not correct.
We cannot mutate a global context.
That's why the logic in this method uses its own local context instance.
We are going to mutate it with this specific variable.
The other thread has its own context instance, so it can mutate it with its own local variable.
when we do that on a global context, we make mistake.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
```diff
@@ -1366,9 +1429,8 @@ private File generateLocalDirectory(Message<?> message, String remoteDirectory)
 
 	private String generateLocalFileName(Message<?> message, String remoteFileName) {
 		if (this.localFilenameGeneratorExpression != null) {
-			EvaluationContext evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
-			evaluationContext.setVariable("remoteFileName", remoteFileName);
-			return this.localFilenameGeneratorExpression.getValue(evaluationContext, message, String.class);
+			this.standardEvaluationContext.setVariable("remoteFileName", remoteFileName);
```
artembilan, line 1432: Same here

src/reference/antora/modules/ROOT/pages/ftp/rft.adoc
```diff
@@ -37,3 +37,23 @@ This is useful when you need to perform several high-level operations of the `Re
 For example, `AbstractRemoteFileOutboundGateway` uses it with the `mput` command implementation, where we perform a `put` operation for each file in the provided directory and recursively for its sub-directories.
 See the https://docs.spring.io/spring-integration/api/org/springframework/integration/file/remote/RemoteFileOperations.html#invoke[Javadoc] for more information.
 
+Starting with version 6.5, the `AbstractRemoteFileOutboundGateway` supports dynamic resolution of `FileExistsMode`
+at runtime via SpEL expressions. This allows you to determine the action to take when files already
+exist based on message content or other conditions.
```
artembilan, line 42: We prefer a `One sentence per line` style: https://asciidoctor.org/docs/asciidoc-recommended-practices/

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -79,6 +79,13 @@ The `AbstractRecentFileListFilter` strategy has been introduced to accept only t
 The respective implementations are provided: `RecentFileListFilter`, `FtpRecentFileListFilter`, `SftpRecentFileListFilter` and `SmbRecentFileListFilter`.
 See xref:file/reading.adoc[Reading Files] for more information.
 
+[[x6.5-file-exists-mode-expression]]
+== FileExistsMode Expression Support
+
+The remote file gateways (`AbstractRemoteFileOutboundGateway`) now support dynamic resolution of `FileExistsMode` at runtime via SpEL expressions. This allows determining the action to take when files already exist based on message content or other conditions.
```
artembilan, line 85: This is a `What's New` chapter.
Just thesis extracts for the rest the docs.
So, there is just enough to say one sentence, like your first one, and then link to the target chapter.

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
anthologia: Actually, I haven't yet fully understood the usage pattern of EvaluationContext..
However, when following the logic, I determined that the return value of `ExpressionUtils.createStandardEvaluationContext(getBeanFactory())` retrieves the `integrationEvaluationContext` Bean registered in `DefaultConfiguringBeanFactoryPostProcessor`.
So I thought it would be the same instance as the global context set in `doInit()`.
I would greatly appreciate if you could clarify where my understanding might be incorrect on this matter.

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
```diff
@@ -358,6 +359,48 @@ public S fileExistsMode(FileExistsMode fileExistsMode) {
...
+	 * Specify a {@link Function} to determine the action to take when files already exist.
+	 * Expression evaluation should return a {@link FileExistsMode} or a String representation.
+	 * Used for GET and MGET operations when the file already exists locally,
+	 * or PUT and MPUT when the file exists on the remote system.
+	 * @param fileExistsModeFunction the {@link Function} to use.
+	 * @param <P> the expected payload type.
+	 * @return the Spec.
+	 * @since 6.5
+	 */
+	public <P> S fileExistsModeFunction(Function<Message<P>, String> fileExistsModeFunction) {
```
artembilan, line 400: `String` is not correct here since we also expect a `FileExistsMode`.
Therefore just `Object`

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
```diff
@@ -573,7 +608,7 @@ private void populateBeanFactoryIntoComponentsIfAny() {
 	private void setupLocalDirectory() {
 		File localDirectory =
 				ExpressionUtils.expressionToFile(this.localDirectoryExpression,
-						ExpressionUtils.createStandardEvaluationContext(getBeanFactory()), null,
+						this.standardEvaluationContext, null,
```
artembilan, line 611: If we do this, then `doInit()` has to be fixed respectively.
Pay attention to its logic: this `setupLocalDirectory()` method is called before you initialize `this.standardEvaluationContext`

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
artembilan: Right. Just because the `EvaluationContext` could be mutated, the bean producing it from the `BeanFactory` is a `prototype`. See `IntegrationEvaluationContextFactoryBean` and its `isSingleton()` implementation.
Therefore `ExpressionUtils.createStandardEvaluationContext()` always returns for us a fresh `EvaluationContext` instance. And since sometimes we don't need to mutate it, that is OK to have a global one for the whole component lifecycle.

artembilan (changes requested):

spring-integration-file/src/main/java/org/springframework/integration/file/remote/gateway/AbstractRemoteFileOutboundGateway.java
anthologia: Thank you for explaining the `EvaluationContext` behavior! I understand now that `ExpressionUtils.createStandardEvaluationContext()` always returns a fresh instance due to the `prototype` scope of `IntegrationEvaluationContextFactoryBean`. This clarifies my confusion and will help me better understand Spring Integration going forward!

spring-integration-file/src/main/java/org/springframework/integration/file/dsl/RemoteFileOutboundGatewaySpec.java
anthologia: Fixed the `Function` parameter type to `Object` as suggested. I also noticed and fixed an incorrect method call in the `fileExistsModeFunction()` implementation (it was incorrectly calling `remoteDirectoryExpression()` instead of `fileExistsModeExpression()`).

anthologia:
I've fixed all the issues you pointed out in your reviews, as well as additional mistake I noticed. I think I made these mistakes because I was rushing with the 6.5 release timeline in mind. Please let me know if there's anything else I might have missed! Thanks.

artembilan:
@anthologia ,

PR #10020, opened 2025-05-17
Migrate few modules to Junit Jupiter

mjd507:
including: `spring-integration-amqp`, `spring-integration-file`, `spring-integration-groovy`

for `RabbitAvailable` test cases, I am not sure they are working or not,  in local all skipped due to no broker available.


artembilan (changes requested):
```
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/inbound/ManualAckTests.java:50:74: Using a static member import should be avoided - org.springframework.integration.amqp.inbound.ManualAckTests.ManualAckConfig. [AvoidStaticImport]
```
Use `check` Gradle command to verify Checkstyle as well.
Thanks

mjd507:
Hi @artembilan I found one test `org.springframework.integration.file.tail.FileTailingMessageProducerTests#testOS` annotated with `@TailAvailable` always failed in my machine (MacOs - support `tail` command), but always success in our CI pipeline test, not sure our CI container support `tail` or not.

local failed at this assert `assertThat(tailEventLatch.await(20, TimeUnit.SECONDS)).isTrue();`  which expect the latch to be released, but actually not.  as per the test case, latch will be released if we publish an event. 

By checking the `OSDelegatingFileTailingMessageProducer`, looks we didn't publish any event if no error happened from the std error.

By checking the doc [file-tailing](https://docs.spring.io/spring-integration/reference/file/reading.html#file-tailing),  we support `FileTailingIdleEvent `, and if I try set `adapter.setIdleEventInterval(2000);` in test, and it will pass.

can confirm can I do it this way, or any better idea. 

Thank you!

artembilan:
Hey, @mjd507 !

Thank you for digging more about those tests!

The CI is Linux, so ha to pass I guess, e..g. latest SNAPSHOT build: https://ge.spring.io/s/xefov3l7yanqy/tests/task/:spring-integration-file:test/details/org.springframework.integration.file.tail.FileTailingMessageProducerTests/testOS?top-execution=1

I use Windows for development. Still works with my WSL2.
Yes, we observed some problems on MacOS without any clues why.

Please, raise a new PR with whatever you think could be as a fix for that problem.

spring-integration-file/src/test/java/org/springframework/integration/file/tail/TailAvailable.java
```diff
@@ -21,13 +21,23 @@
...
  */
+@ExtendWith(TailCondition.class)
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD})
 public @interface TailAvailable {
 
+	/**
+	 * the options for the 'tail' command.
+	 * @return the options.
+	 *
```
artembilan, line 39: The sentence has to start from capital letter.
No blank lines in method Javadocs, please.
Your name to the `@author` list.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlJdbcMessageStoreMultipleChannelTests.java
```diff
@@ -55,10 +53,9 @@
  * @author Gunnar Hillert
  *
  */
-@ContextConfiguration
-@RunWith(SpringJUnit4ClassRunner.class)
+@SpringJUnitConfig
 @DirtiesContext(classMode = ClassMode.AFTER_EACH_TEST_METHOD)
-@Ignore
+@Disabled
```
artembilan, line 58: There is a `MySqlContainerTest` contract to test with Testcontainers.
Feel free to raise a new GH issue to enable this test eventually.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/DerbyTxTimeoutMessageStoreTests.java
```diff
@@ -26,7 +26,7 @@
  * @author Artem Bilan
  *
  */
-@Ignore
+@Disabled
```
artembilan, line 29: Would be interesting to see if we can enable it back.
Separate issue, though.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/MySqlTxTimeoutMessageStoreTests.java
```diff
@@ -26,7 +26,7 @@
  * @author Artem Bilan
  *
  */
-@Ignore
+@Disabled
```
artembilan, line 29: Same here to use `MySqlContainerTest`

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/OracleTxTimeoutMessageStoreTests.java
```diff
@@ -26,7 +26,7 @@
  * @author Artem Bilan
  *
  */
-@Ignore
+@Disabled
```
artembilan, line 29: For this use-case we have `OracleContainerTest`

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/PostgresTxTimeoutMessageStoreTests.java
```diff
@@ -26,7 +26,7 @@
  * @author Artem Bilan
  *
  */
-@Ignore
+@Disabled
```
artembilan, line 29: And here `PostgresContainerTest`

artembilan (changes requested):
This is great.
Just some minor comments over there in the code.

Do you know already, @mjd507 , how many modules left with some JUnit 4 tests?

Thank you for great contribution!

mjd507:
Thank you @artembilan , I always expect the comments you leave on my changes, Even though I was doing the relatively easy part, I always learned something useful.

> how many modules left with some JUnit 4 tests?

by search `org.junit.Test`, I see 55 match files, and if I exclude `spring-integration-hazelcast`, which contains 15 matched files, so only 40 files, most are in `spring-integration-jmx` module,  we don't need migrate `spring-integration-hazelcast` right?

spring-integration-file/src/test/java/org/springframework/integration/file/config/FileOutboundGatewayParserTests.java
```diff
@@ -52,12 +49,11 @@
...
  * @author Tony Falabella
  */
-@RunWith(SpringRunner.class)
+@SpringJUnitConfig
 @DirtiesContext
 public class FileOutboundGatewayParserTests {
 
-	@ClassRule
-	public static final TemporaryFolder tempFolder = new TemporaryFolder();
+	public static File tempFolder = new File(System.getProperty("java.io.tmpdir"), "FileOutboundGatewayParserTests");
```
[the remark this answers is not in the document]
mjd507: yes, we can use. I should have referred to another test case at that time.

spring-integration-file/src/test/java/org/springframework/integration/file/tail/TailAvailable.java
mjd507: updated, is there anyway to make a warning in local for JavaDoc?

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlJdbcMessageStoreMultipleChannelTests.java
mjd507: nice, another new thing for me. will raise new issue

spring-integration-file/src/test/java/org/springframework/integration/file/tail/TailAvailable.java
artembilan: I don’t think it does validation if natural languages. Plus, the formatting of Javadoc is also hard to verify. So, no way to catch such an error. 
We might think about migrating to https://github.com/spring-io/spring-javaformat, but I’m not very happy how it treats method chain style.

artembilan:
No, we still need to migrate Hazelcast. We have deprecated only some of the components which require CP subsystem. The rest remains as is. And if you say there are some JUnit 4 tests , then we need to migrate them.
No rush, though, because they are just tests.
But tomorrow I have `6.5` GA release . And after that I’m switching to `7.0` where real “hardcore” begins, including the mentioned JUnit 4 utils deprecation.

So, thank you for reporting that, I’ll try to migrate renaming tests myself before releasing tomorrow .

Issue #10045, opened 2025-05-20
Enable JDBC test cases with TestContainers approach


mjd507:
There are few JDBC test cases currently are Disabled, 
Now I want to enable them back with TestContainers approach like @artembilan mentioned [MySqlContainerTest](https://github.com/spring-projects/spring-integration/pull/10020#discussion_r2096066016), [OracleContainerTest](https://github.com/spring-projects/spring-integration/pull/10020#discussion_r2096070381), [PostgresContainerTest](https://github.com/spring-projects/spring-integration/pull/10020#discussion_r2096070727)

artembilan:
> I didn't find the official TestContainer for Derby, so I use the Embedded one. we already have its dependency.

That's correct. Why do we need a test container for such a simple embedded database?


I added this into the test:
```

	@BeforeAll
	static void setUp() {
		System.setProperty("derby.locks.deadlockTrace", "true");
	}
```
and got this in logs:
```
2025-05-27 12:02:21,218 ERROR [pool-2-thread-1] [org.springframework.integration.jdbc.store.channel.DerbyTxTest] - IdCache race condition.
org.springframework.dao.PessimisticLockingFailureException: PreparedStatementCallback; SQL [	SELECT INT_CHANNEL_MESSAGE.MESSAGE_ID, INT_CHANNEL_MESSAGE.MESSAGE_BYTES
	from INT_CHANNEL_MESSAGE
	where INT_CHANNEL_MESSAGE.GROUP_KEY = ? and INT_CHANNEL_MESSAGE.REGION = ?
and INT_CHANNEL_MESSAGE.MESSAGE_ID not in (?) order by CREATED_DATE, MESSAGE_SEQUENCE FETCH FIRST ROW ONLY]; A lock could not be obtained within the time requested.  The lockTable dump is: 
Tue May 27 12:02:21 EDT 2025
XID       |TYPE         |MODE|LOCKCOUNT|LOCKNAME                                                                        |STATE|TABLETYPE / LOCKOBJ                   |INDEXNAME / CONTAINER_ID / (MODE for LATCH only)  |TABLENAME / CONGLOM_ID                |
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
*** The following row is the victim ***
251       |ROW          |S   |0        |(1,8)                                                                           |WAIT |T                                     |NULL                                              |INT_CHANNEL_MESSAGE                   |
*** The above row is the victim ***
[...]

PR #10059, opened 2025-05-30
#10045: Enable JDBC test cases with TestContainers approach

mjd507:
I really don't have any better ways to enable those JDBC concurrent test cases back :(

for `AbstractTxTimeoutMessageStoreTests`, there are two pollers impact the performance, first poller using custom task-executor with 100 threads run with fixed-rate 200, second poller with default executor run with fixed-delay 100.

for Mysql/Oracle, I increase the fixed-rate from 200 to 2000, and fixed-delay from 100 to 1000.

### DIFF
--- a/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlJdbcMessageStoreMultipleChannelTests-context.xml
+++ b/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlJdbcMessageStoreMultipleChannelTests-context.xml
@@ -19,14 +19,13 @@
 	<tx:annotation-driven/>
 	<int:annotation-config/>
 
-	<bean id="dataSource" class="org.apache.commons.dbcp2.BasicDataSource" destroy-method="close">
-		<property name="driverClassName" value="com.mysql.jdbc.Driver"/>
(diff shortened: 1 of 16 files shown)


mjd507:
thank you, all those test cases passed in my local, both individual test classes and whole `jdbc` integration tests.  I didn't add `LongRunningTest` for them now, but I think we need. possibly add `@Disabled` back again .

after reviewing or testing, if any changes required, can help me to change it directly ? 
thanks again

mjd507:
> That would cause some routine when this is merged and you would need to force update your local main.

Sure, nice suggestion！
Currently after every PR merged, I manually reset --hard to a previous upstream version in my local, and pull latest version from upstream again. hmm.. Just only want to keep only one branch.  regarding my remote branch, I often push with -f , it's sometimes dangerous. 

Will try new a feature branch in future and see.

artembilan (approved):
Merging.
The timing is OK for these new tests: no need for `long-running`.
We will disable those tests which are not stable according to our experience.
For now we can live with whatever we have so far.

Thank you!

artembilan:
FYI, the snapshot against your change is green: https://github.com/spring-projects/spring-integration/actions/runs/15350770171/job/43198145399 🤷

PR #10061, opened 2025-05-31
Streaming support for `JdbcOutboundGateway`

### LINKED ISSUE #3963: Streaming support into a JdbcPollingChannelAdapter
`JdbcPollingChannelAdapter` used by `JdbcOutboundGateway` does NOT stream the resultset, whereas returns all the results as a list. This could result in OOM if there are large number of records returned.

Would be better if streaming support is added to `JdbcPollingChannelAdapter` used by `JdbcOutboundGateway`

**Reference from SO:** 
https://stackoverflow.com/questions/74729365/spring-integration-jdbc-outboundgateway-returning-1-record-only-even-with-maxrow

mjd507:
Fixes: #3963

Introduce a `queryForStream` variable and a `streamConsumer` function in `JdbcOutboundGateway`. along with the underlying method `JdbcPollingChannelAdapter.doPollForStream`.

when produce reply messages, if in stream mode, gateways always receives a empty list, process logic are moved to `streamConsumer`.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,8 @@ package org.springframework.integration.jdbc;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
 
 import javax.sql.DataSource;
 
@@ -36,6 +38,7 @@ import org.springframework.util.StringUtils;
  * @author Gunnar Hillert
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Jiandong Ma
  *
  * @since 2.0
  */
@@ -52,6 +55,10 @@ public class JdbcOutboundGateway extends AbstractReplyProducingMessageHandler {
 	private boolean keysGenerated;
 
 	private Integer maxRows;
+
+	private boolean queryForStream = false;
+
+	private Consumer<Object> streamConsumer;
 
 	/**
 	 * Construct an instance based on the provided {@link DataSource} and update SQL.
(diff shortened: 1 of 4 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
```diff
@@ -53,6 +56,10 @@ public class JdbcOutboundGateway extends AbstractReplyProducingMessageHandler {
 
 	private Integer maxRows;
 
+	private boolean queryForStream = false;
+
+	private Consumer<Object> streamConsumer;
```
mjd507, line 61: here given Object type for `Consumer`, user have to do  the type cast , which I am not very satisfied with this,  because when set `RowMapper`, they already give return type.
I don't have a good way. I tried to put `?`, but looks it can not be applied to Stream<?> .

mjd507 (commented):
regarding `transaction`, I think no impact ?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
```diff
@@ -218,7 +248,15 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
 					sqlQueryParameterSource = this.sqlParameterSourceFactory.createParameterSource(list);
 				}
 			}
-			list = this.poller.doPoll(sqlQueryParameterSource);
+			if (this.queryForStream) {
+				try (Stream<?> stream = this.poller.doPollForStream(sqlQueryParameterSource)) {
+					stream.forEach(this.streamConsumer);
+				}
+				return Collections.emptyList();
```
artembilan, line 255: Thank you for the effort, but this is not correct.
This is not a messaging way and this is not a goal of Spring Integration channel adapter.
If we go the of callback instead of sending a message to the channel, then there is no reason in using channel adapters.
Simple `JdbcTemplate` with its streaming API would do exactly the same without extra messaging layer to introduce for nothing.

Having a `Stream` as a message payload to produce from the channel adapter it totally fine.
However still all those questions are opened: the `Stream` has to be closed, not clear how to update entries of this stream to satisfy concurrent polling from the table. At the same time how do transactions behave with these opened (and possible updated) streams.

I general I suggest to close this PR since we are not going the way you suggest.
Sorry about that and thank you again for your effort!

mjd507: ohh, make sense, thank you for making me think in messaging way, I was wrong from the beginning.

PR #10079, opened 2025-06-03
Disable multicast tests when buiding on MacOS

cppwfs:
Multicast tests fail when running on MacOS.  So these tests will be disabled when building on macOS based machines.

I did not include the Oracle tests because M4 processors do support the oracle testcontainer based tests.


cppwfs:
Closing PR.   Will open a new one with proper credentials.

PR #10087, opened 2025-06-09
GH-9996: Migrating remaining tests to Junit Jupiter and Deprecate Junit4 utilities

### LINKED ISSUE #9996: Deprecate JUnit 4 utilities in the project
Also, migrate JUnit 4 tests to Jupiter.
More info in: https://github.com/spring-projects/spring-framework/issues/34794

mjd507:
first commit: migrate remaining tests to Junit Jupiter
second one: deprecate Junit 4 utilities

fixes: https://github.com/spring-projects/spring-integration/issues/9996

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/support/MessageScenariosTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/support/MessageScenariosTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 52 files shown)


spring-integration-core/src/test/java/org/springframework/integration/support/MessageScenariosTests.java
```diff
@@ -37,7 +37,7 @@
  */
 @ContextConfiguration
 @DirtiesContext
-public class MessageScenariosTests extends AbstractRequestResponseScenarioTests {
+public class MessageScenariosTests extends AbstractRequestResponseScenarioTest {
```
mjd507, line 40: not very good at naming, so remove `s` to create a new copy in our test-support.

cppwfs (approved):

cppwfs (changes requested):
Great Job!

In this case let's leave these SingleRequestResponseScenarioTests and its parent as plural.   While there is only a single test at present, it does not mean that ones can be added in the future.

artembilan:
> In this case let's leave these SingleRequestResponseScenarioTests and its parent as plural.

I think the change in this PR on the matter is OK. 
Those `scenario` classes are abstract and supposed to be used in the implementation.
That is a fully-blown test class has to have a `Tests` suffix to satisfy our code style.

Thanks

Issue #10083, opened 2025-06-06
Apply Nullability to the project


artembilan:
More info in the Spring Framework `7.0` docs: https://docs.spring.io/spring-framework/reference/7.0/core/null-safety.html.

We can treat this issue an Epic and have sub-issues for specific tasks or dedicated module fixes.

This can be done in phase:

1. Migrate from deprecated `org.springframework.lang.Nullable` on properties methods to the `org.jspecify.annotations.Nullable`.
2. Migrate `@org.springframework.lang.NonNullApi` and `@org.springframework.lang.NonNullFields` pair to the `@org.jspecify.annotations.NullMarked`. However this would trigger the third phase.
3. Fix nullabilty errors in those `@org.jspecify.annotations.NullMarked` packages.
4. Add more and more `@org.jspecify.annotations.NullMarked`  and fix respective nullability vulnerabilities in those package and modules.

Any other thoughts and possible ways to fix this huge task are welcome!

list (checked means someone has taken it )
- [x] spring-integration-amqp
- [x] spring-integration-camel
- [x] spring-integration-cassandra
- [x] spring-integration-core
- [x] spring-integration-debezium
- [x] spring-integration-event
- [x] spring-integration-feed
- [x] spring-integration-file
- [x] spring-integration-ftp
- [x] spring-integration-graphql
- [x] spring-integration-groovy
- [x] spring-integration-hazelcast
- [x] spring-integration-http
- [x] spring-integration-ip
- [x] spring-integration-jdbc
- [x] spring-integration-jms
- [x] spring-integration-jmx
- [x] spring-integration-jpa
- [x] spring-integration-kafka
- [x] spring-integration-mail
- [x] spring-integration-mongodb
- [x] spring-integration-r2dbc
- [x] spring-integration-mqtt
- [x] spring-integration-redis
- [x] spring-integration-rsocket
- [x] spring-integration-scripting
- [x] spring-integration-sftp
- [x] spring-integration-smb
- [x] spring-integration-stomp
- [x] spring-integration-stream
- [x] spring-integration-syslog
- [x] spring-integration-test
- [x] spring-integration-test-support
- [x] spring-integration-webflux
- [x] spring-integration-websocket
- [x] spring-integration-ws
- [x] spring-integration-xml
- [x] spring-integration-xmpp
- [x] spring-integration-zeromq
- [x] spring-integration-zip
- [x] spring-integration-zookeeper

anayonkars:
[...]

PR #10091, opened 2025-06-09
GH-10083: Migrate ZeroMQ module to Jspecify

anthologia:

**Changes:**
- Replaced `org.springframework.lang.Nullable` with `org.jspecify.annotations.Nullable`

Should I create separate PRs for each module?

### DIFF
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
@@ -142,7 +142,7 @@ public class ZeroMqChannel extends AbstractMessageChannel implements Subscribabl
 		this.subscriberData = prepareSubscriberDataFlux();
 	}
 
-	@SuppressWarnings("this-escape")
+	@SuppressWarnings({"this-escape", "NullAway"})
 	private Mono<Integer> prepareProxyMono() {
 		return Mono.defer(() -> {
 					if (this.zeroMqProxy != null) {
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
+import org.jspecify.annotations.Nullable;
 import org.zeromq.SocketType;
 import org.zeromq.ZContext;
 import org.zeromq.ZMQ;
@@ -36,7 +37,6 @@ import org.springframework.integration.channel.AbstractMessageChannel;
 import org.springframework.integration.mapping.BytesMessageMapper;
 import org.springframework.integration.support.json.EmbeddedJsonHeadersMessageMapper;
 import org.springframework.integration.zeromq.ZeroMqProxy;
-import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.messaging.SubscribableChannel;
--- a/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
+++ b/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
@@ -66,6 +66,7 @@ import org.springframework.util.Assert;
  * concurrency primitives for multi-publisher(subscriber) communication within the same application.
  *
  * @author Artem Bilan
+ * @author Jooyoung Pyoung
  *
  * @since 5.4
  */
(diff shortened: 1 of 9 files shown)


artembilan:
> Fixes: https://github.com/spring-projects/spring-integration/issues/10083

Is not correct in the PR since it does not fully fix the whole issue.
Please, don't use it for this kind of PRs again.
Better like `Related to: `

> Should I create separate PRs for each module?

Yes, please.
That will be easier for us to review.

Thanks

artembilan (changes requested):
I'm not sure if this was automated, but can we also look into migrating those `package-info.java` files in this module to the `org.jspecify.annotations.NullMarked` instead of those deprecated over there?
Then it would be great to run `gradlew :spring-integration-zeromq:check` and see if your change passes NullAway.

Thanks

artembilan:
I know the `org.jspecify.annotations.NullMarked` is a phase 2 in our issue roadmap, but that is OK to do that for such a small module in one go, so we would know where we are for the rest of the project.

Thanks

artembilan:
We have just merged one for Camel module: https://github.com/spring-projects/spring-integration/commit/b07d04f9e5dafe54819ff47ec361a75c6c0e244e.

Feel free to follow its experience!

anthologia:
Thanks for the feedback @artembilan ! 😁

I've updated the PR:
- Migrated all `package-info.java` files to `@NullMarked`
- Added `@SuppressWarnings("NullAway.Init")` for lifecycle-initialized fields
- Updated existing commit message from `Fixes:` to `Related to:` via force push (apologies for the extra notifications)

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
```diff
@@ -142,7 +142,7 @@ public ZeroMqChannel(ZContext context, boolean pubSub) {
 		this.subscriberData = prepareSubscriberDataFlux();
 	}
 
-	@SuppressWarnings("this-escape")
+	@SuppressWarnings({"this-escape", "NullAway"})
```
artembilan: Not sure in this `NullAway`.
What is an error you got without it?

artembilan (changes requested):

spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java
anthologia: I think `NullAway` cannot track the null check for `this.zeroMqProxy` inside the lambda expression.
Since `zeroMqProxy` could potentially be modified, I believe adding a `@SuppressWarnings("NullAway")` would be more appropriate than caching to a local variable.

```
> Task :spring-integration-zeromq:compileJava FAILED
/Users/jooyoung/spring-integration/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java:149: error: [NullAway] dereferenced expression this.zeroMqProxy is @Nullable
                                                return Mono.fromCallable(() -> this.zeroMqProxy.getBackendPort())
                                                                                               ^
/Users/jooyoung/spring-integration/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java:153: error: [NullAway] dereferenced expression this.zeroMqProxy is @Nullable
                                                                                setConnectUrl("tcp://localhost:" + this.zeroMqProxy.getFrontendPort() +
                                                                                                                                   ^
2 errors

[Incubating] Problems report is available at: file:///Users/jooyoung/spring-integration/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spring-integration-zeromq:compileJava'.
> Compilation failed; see the compiler output below.
  /Users/jooyoung/spring-integration/spring-integration-zeromq/src/main/java/org/springframework/integration/zeromq/channel/ZeroMqChannel.java:149: error: [NullAway] dereferenced expression this.zeroMqProxy is @Nullable
                                                return Mono.fromCallable(() -> this.zeroMqProxy.getBackendPort())
                                                                                               ^
  2 errors
```

artembilan: Right. So, how about indeed to follow a recommendation and extract into a local variable before this `return Mono.defer(()` ?

artembilan: Sorry, we cannot do that before `Mono.defer(()` since this `prepareProxyMono()` is called from ctor.
So, we have to extract a local `ZeroMqProxy zeroMqProxyToUse = this.zeroMqProxy;` in that `Mono.defer(()` block and go from there. 
Perhaps the `this-escape` won't be necessary as well.

artembilan (changes requested):

anthologia:
I've applied the feedback. All tests pass locally, but Kafka module tests are failing in CI. Could you re-trigger the build please?

artembilan:
As far as the PR build is green, this is going to be merged.
Thank you!

If we miss something, we can address it later when we got less those warnings in the logs to spot leftovers easier.
Yes, that Kafka test fail sporadically.
have no idea yet why we go same record again when no rebalance.

anthologia:
Thanks ! But build keeps failing... I wonder what might be causing this? Perhaps something related to JDK 24 migration?

artembilan:
Most of them are related to sockets from libraries we rely on, so it is hard to diagnose.

PR #10092, opened 2025-06-09
GH-10083: Replace springframework Nullable with jspecify Nullable in spring-integration-camel

anayonkars:
This change is applicable only to `spring-integration-camel`. 
Replaced `org.springframework.lang.Nullable` with `org.jspecify.annotations.Nullable`.

### DIFF
--- a/spring-integration-camel/src/main/java/org/springframework/integration/camel/outbound/CamelMessageHandler.java
+++ b/spring-integration-camel/src/main/java/org/springframework/integration/camel/outbound/CamelMessageHandler.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.camel.outbound;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
 import org.apache.camel.CamelContext;
@@ -68,6 +69,7 @@ import org.springframework.util.StringUtils;
  */
 public class CamelMessageHandler extends AbstractReplyProducingMessageHandler {
 
+	@Nullable
 	private ProducerTemplate producerTemplate;
 
 	private Expression exchangePatternExpression = new ValueExpression<>(ExchangePattern.InOnly);
@@ -83,6 +85,7 @@ public class CamelMessageHandler extends AbstractReplyProducingMessageHandler {
 	@Nullable
 	private Expression exchangePropertiesExpression;
 
+	@Nullable
 	private StandardEvaluationContext evaluationContext;
 
 	public CamelMessageHandler() {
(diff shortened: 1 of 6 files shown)


artembilan (changes requested):
Same here, @anayonkars ,
Please, migrate `package-info.java` in this module to the `@org.jspecify.annotations.NullMarked` and run `gradlew :spring-integration:check` locally.

artembilan (changes requested):
I mean that is essentially what `Apply Nullability` is about.
And, please, be more specific in the title of your PR (commit headline, to be precise).
I mean you don't apply the change for the whole project, but rather specific module, so would be great to reflect that in the PR title.

Let me know how that `@NullMarked` goes!

Thanks

anayonkars:
@artembilan I've made changes to PR's title.
Also made relevant changes in `package-info.java`.
Had to do few more changes in `CamelMessageHandler` due to nullable related issues.
Kindly review.

spring-integration-camel/src/main/java/org/springframework/integration/camel/outbound/CamelMessageHandler.java
```diff
@@ -83,6 +85,7 @@ public class CamelMessageHandler extends AbstractReplyProducingMessageHandler {
 	@Nullable
 	private Expression exchangePropertiesExpression;
 
+	@Nullable
```
artembilan: This has to be mitigated as `@SuppressWarnings("NullAway.Init")`.
It is initialized from the `afterPropertiesSet()`.

spring-integration-camel/src/main/java/org/springframework/integration/camel/outbound/CamelMessageHandler.java
```diff
@@ -197,46 +200,57 @@ public void configure() throws Exception {
 	}
 
 	@Override
+	@Nullable
 	protected Object handleRequestMessage(Message<?> requestMessage) {
+		StandardEvaluationContext localEvaluationContext = Objects.requireNonNull(this.evaluationContext,
+				"'evaluationContext' must not be null");
```
artembilan: With the `NullAway.Init` we don't need this anymore.

spring-integration-camel/src/main/java/org/springframework/integration/camel/outbound/CamelMessageHandler.java
```diff
@@ -197,46 +200,57 @@ public void configure() throws Exception {
...
 		ExchangePattern exchangePattern =
-				this.exchangePatternExpression.getValue(this.evaluationContext, requestMessage, ExchangePattern.class);
+				this.exchangePatternExpression.getValue(localEvaluationContext, requestMessage, ExchangePattern.class);
 
 		Assert.notNull(exchangePattern, "'exchangePatternExpression' must not evaluate to null");
 
 		Endpoint endpoint = resolveEndpoint(requestMessage);
 		Exchange exchange = prepareInExchange(endpoint, exchangePattern, requestMessage);
 
+		ProducerTemplate localProducerTemplate = Objects.requireNonNull(this.producerTemplate,
```
artembilan: Same with `this.producerTemplate`

artembilan (changes requested):

spring-integration-camel/src/main/java/org/springframework/integration/camel/outbound/CamelMessageHandler.java
anayonkars: Thanks! I've handled this as per review comments.

artembilan: Thanks, but you don't have to say this since I got GH notification about a new commit in this PR.
I believe if you agreed with me than I would see the requested change in that commit.
If you don't agree, then feel free to comment and start discussion about the possible path.

Sometime I have like one hundred notifications about updates in my projects and simple `Done` in my mail box brings extra unnecessary noise 😄

PR #10096, opened 2025-06-11
GH-10083: Apply Nullability to `spring-integration-mail`

mjd507:
Related to: #10083


mjd507:
I choose maybe the laziest way,  add `@SuppressWarnings("NullAway")` on class level.

should I try to add more `@Nullable` on Fields ?  actually I tried on `AbstractMailReceiver`,  but too many dereferenced expression have to be solved by adding extra `@SuppressWarnings("NullAway")
` on methods, which made me not sure adding `@Nullable` on Fields is correct or not.

spring-integration-mail/src/main/java/org/springframework/integration/mail/config/package-info.java
```diff
@@ -1,4 +1,7 @@
 /**
  * Provides classes for configuration - parsers, namespace handlers.
  */
+@NullMarked
 package org.springframework.integration.mail.config;
+
+import org.jspecify.annotations.NullMarked;
```
artembilan: Every class has to end with blank line, please.

spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
```diff
@@ -70,6 +70,7 @@
  * @author Ngoc Nhan
  * @author Filip Hrisafov
  */
+@SuppressWarnings("NullAway")
```
artembilan: Yes, this is not OK. 
We really have to implement null-away in our classes.
Thanks

artembilan (changes requested):

spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
```diff
@@ -116,8 +119,10 @@ public abstract class AbstractMailReceiver extends IntegrationObjectSupport impl
 
 	private boolean flaggedAsFallback = true;
 
+	@SuppressWarnings("NullAway.Init")
 	private volatile Store store;
 
+	@SuppressWarnings("NullAway.Init")
 	private volatile Folder folder;
```
mjd507: I tried to make both `store` and `folder` with `@Nullable`, but not that easy.
and seems it is never null at mail receiving runtime? so I suppress warnings here, and remove the `this.folder = null;  this.store = null` in `destroy` method

spring-integration-mail/src/main/java/org/springframework/integration/mail/MailSendingMessageHandler.java
```diff
@@ -190,12 +194,14 @@ private void applyHeadersToMailMessage(MailMessage mailMessage, MessageHeaders h
...
+	private String @Nullable [] retrieveHeaderValueAsStringArray(MessageHeaders headers, String key) {
 		Object value = headers.get(key);
 		String[] returnedHeaders = null;
 		if (value != null) {
-			if (value instanceof String[]) {
-				returnedHeaders = (String[]) value;
+			if (value instanceof String[] strArr) {
+				returnedHeaders = Arrays.stream(strArr)
+						.filter(Objects::nonNull)
+						.toArray(String[]::new);
```
mjd507, line 204: here I filter null elements from String[], because `new String {"a", "b", null}` really could happen, but not sure in this real case.

spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapIdleChannelAdapter.java
```diff
@@ -252,8 +256,8 @@ private void delayNextIdleCall() {
 		}
 	}
 
-	@Nullable
-	private static jakarta.mail.MessagingException getJakartaMailMessagingExceptionFromCause(Throwable cause) {
+	@SuppressWarnings("NullAway")
```
artembilan: Why `NullAway` if this method really can return `null`?

spring-integration-mail/src/main/java/org/springframework/integration/mail/MailSendingMessageHandler.java
artembilan: Fine with me.
Thanks

spring-integration-mail/src/main/java/org/springframework/integration/mail/dsl/package-info.java
```diff
@@ -1,6 +1,7 @@
 /**
  * Provides Mail Components for the Java DSL.
  */
-@org.springframework.lang.NonNullApi
-@org.springframework.lang.NonNullFields
+@NullMarked
 package org.springframework.integration.mail.dsl;
+
+import org.jspecify.annotations.NullMarked;
```
artembilan: No imports, please, in `package-info.java` files.
Use the FQCN in the annotation place.
See my PR for `event` module: #10099

spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
artembilan: I think they have to be `@Nullable`.
In those places where we have a problem, we should consider to mitigate them with local variables instead of constant access to the property.
I believe null checks are around there, so should be OK to move on with local copy variable in those methods.

artembilan (changes requested):

spring-integration-mail/src/main/java/org/springframework/integration/mail/ImapIdleChannelAdapter.java
mjd507: ohh, I forgot to leave a comment on this. Actually I tried put '@Nullable' on the return type, but looks jspecify doesn't allow to put nullable on a Fully Qualified Class Name return type, 

let me try to see if we have ways to solve it today

artembilan: See my comment on another PR for similar problem: https://github.com/spring-projects/spring-integration/pull/10098#discussion_r2140471773

PR #10097, opened 2025-06-11
Introduce NullMarked to some packages in spring-integration-core module

cppwfs:
This is a first pass at the module to make sure it is being converted properly.

There are 3 commits that comprise this PR.  Each commit handles a specific series of package(s) on the model.  
This was done to simplify the review process so each package could be reviewed indepently.
* Commit 1 - org.springframework.integration.container
* Commit 2 - org.springframework.integration.messaging.core
* Commit 3 - org.springframework.integration.aggregator & org.springframework.integration.acks

> [!NOTE]  
> Highlights information that users should take into account, even when skimming. in this commit that there cases where  methods are called where the parameters passed in may contain a null.
> In those cases the attribute that contains the potential null is tested with an `Assert.notNull`.  This may not be the best approach, let's discuss.

> [!NOTE]  
> Also note that in FluxAggregatorMessageHandler line 112 `@NullUnmarked` is used on the the applyWindowOptions method.
> This is because the consumer.apply on line 121 can not be null according to JSpecify.   But we can't make that assumption for how the user implemented the fuction.
>There maybe a better way of handling this besides marking the method as `NullUnmarked`.   Let's discuss.
> Similarly in FluxAggregatorMessageHandler the way in which sequenceSizeHeader is utilized it had to be `@NullUnmarked`.   Let's discuss

Related to https://github.com/spring-projects/spring-integration/issues/10083

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractAggregatingMessageGroupProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractAggregatingMessageGroupProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.util.function.Function;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
@@ -56,6 +57,7 @@ public abstract class AbstractAggregatingMessageGroupProcessor implements Messag
 
 	private boolean messageBuilderFactorySet;
 
+	@SuppressWarnings("NullAway.Init")
 	private BeanFactory beanFactory;
 
 	@Override
@@ -117,6 +119,7 @@ public abstract class AbstractAggregatingMessageGroupProcessor implements Messag
 		return getHeadersFunction().apply(group);
 	}
 
+	@Nullable
 	protected abstract Object aggregatePayloads(MessageGroup group, Map<String, Object> defaultHeaders);
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractAggregatingMessageGroupProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractAggregatingMessageGroupProcessor.java
@@ -21,7 +21,6 @@ import java.util.function.Function;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
@@ -119,7 +118,6 @@ public abstract class AbstractAggregatingMessageGroupProcessor implements Messag
 		return getHeadersFunction().apply(group);
 	}
 
-	@Nullable
 	protected abstract Object aggregatePayloads(MessageGroup group, Map<String, Object> defaultHeaders);
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -937,7 +937,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 			this.logger.debug(() -> "Completing group with correlationKey [" + correlationKey + "]");
 
 			result = this.outputProcessor.processMessageGroup(group);
-			Assert.notNull(result, "the group returned a null result");
+			Assert.state(result != null, "The processorMessageGroup returned a null result. Null result is not expected.");
 			if (isResultCollectionOfMessages(result)) {
 				partialSequence = (Collection<Message<?>>) result;
 			}
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -65,6 +65,7 @@ import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHandlingException;
 import org.springframework.messaging.core.DestinationResolutionException;
 import org.springframework.messaging.support.GenericMessage;
+import org.springframework.scheduling.TaskScheduler;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
@@ -125,8 +126,10 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	private boolean releaseStrategySet;
 
+	@SuppressWarnings("NullAway.Init")
 	private MessageChannel discardChannel;
 
+	@Nullable
 	private String discardChannelName;
 
 	private boolean sendPartialResultOnExpiry;
@@ -143,18 +146,23 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	private boolean releasePartialSequences;
 
+	@Nullable
 	private Expression groupTimeoutExpression;
 
+	@Nullable
 	private List<Advice> forceReleaseAdviceChain;
 
 	private long expireTimeout;
 
+	@Nullable
 	private Duration expireDuration;
 
 	private MessageGroupProcessor forceReleaseProcessor = new ForceReleaseMessageGroupProcessor();
 
+	@SuppressWarnings("NullAway.Init")
 	private EvaluationContext evaluationContext;
 
+	@Nullable
 	private ApplicationEventPublisher applicationEventPublisher;
 
 	private boolean expireGroupsUponTimeout = true;
@@ -165,10 +173,11 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	private volatile boolean running;
 
+	@Nullable
 	private BiFunction<Message<?>, String, String> groupConditionSupplier;
 
 	public AbstractCorrelatingMessageHandler(MessageGroupProcessor processor, MessageGroupStore store,
-			CorrelationStrategy correlationStrategy, ReleaseStrategy releaseStrategy) {
+			@Nullable CorrelationStrategy correlationStrategy, @Nullable ReleaseStrategy releaseStrategy) {
 
 		Assert.notNull(processor, "'processor' must not be null");
 		Assert.notNull(store, "'store' must not be null");
@@ -504,6 +513,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		return this.discardChannel;
 	}
 
+	@Nullable
 	protected String getDiscardChannelName() {
 		return this.discardChannelName;
 	}
@@ -532,6 +542,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		return this.releasePartialSequences;
 	}
 
+	@Nullable
 	protected Expression getGroupTimeoutExpression() {
 		return this.groupTimeoutExpression;
 	}
@@ -641,8 +652,10 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 	}
 
 	private void removeEmptyGroupAfterTimeout(UUID groupId, long timeout) {
+		TaskScheduler taskScheduler = getTaskScheduler();
+		Assert.notNull(taskScheduler, "'taskScheduler' must not be null");
 		ScheduledFuture<?> scheduledFuture =
-				getTaskScheduler()
+				taskScheduler
 						.schedule(() -> {
 							Lock lock = this.lockRegistry.obtain(groupId.toString());
 
@@ -699,8 +712,10 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 				Object groupId = messageGroup.getGroupId();
 				long timestamp = messageGroup.getTimestamp();
 				long lastModified = messageGroup.getLastModified();
+				TaskScheduler taskScheduler = getTaskScheduler();
+				Assert.notNull(taskScheduler, "'taskScheduler' must not be null");
 				ScheduledFuture<?> scheduledFuture =
-						getTaskScheduler()
+						taskScheduler
 								.schedule(() -> {
 									try {
 										processForceRelease(groupId, timestamp, lastModified);
@@ -753,7 +768,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 	 * @param group The group.
 	 * @param completedMessages The completed messages.
 	 */
-	protected abstract void afterRelease(MessageGroup group, Collection<Message<?>> completedMessages);
+	protected abstract void afterRelease(MessageGroup group, @Nullable Collection<Message<?>> completedMessages);
 
 	/**
 	 * Subclasses may override if special action is needed because the group was released or discarded
@@ -912,14 +927,12 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 	}
 
 	protected void completeGroup(Object correlationKey, MessageGroup group, Lock lock) {
-		Message<?> first = null;
-		if (group != null) {
-			first = group.getOne();
-		}
+		Message<?> first = group.getOne();
 		completeGroup(first, correlationKey, group, lock);
 	}
 
 	@SuppressWarnings("unchecked")
+	@Nullable
 	protected Collection<Message<?>> completeGroup(Message<?> message, Object correlationKey, MessageGroup group,
 			Lock lock) {
 
@@ -929,6 +942,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 			this.logger.debug(() -> "Completing group with correlationKey [" + correlationKey + "]");
 
 			result = this.outputProcessor.processMessageGroup(group);
+			Assert.notNull(result, "the group returned a null result");
 			if (isResultCollectionOfMessages(result)) {
 				partialSequence = (Collection<Message<?>>) result;
 			}
@@ -988,6 +1002,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		return false;
 	}
 
+	@Nullable
 	protected Object obtainGroupTimeout(MessageGroup group) {
 		if (this.groupTimeoutExpression != null) {
 			Object timeout = this.groupTimeoutExpression.getValue(this.evaluationContext, group);
@@ -1024,7 +1039,9 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 			if (this.expireTimeout > 0) {
 				purgeOrphanedGroups();
 				if (this.expireDuration != null) {
-					getTaskScheduler()
+					TaskScheduler taskScheduler = getTaskScheduler();
+					Assert.notNull(taskScheduler, "'taskScheduler' must not be null");
+					taskScheduler
 							.scheduleWithFixedDelay(this::purgeOrphanedGroups, this.expireDuration);
 				}
 			}
@@ -1062,6 +1079,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	protected static class SequenceAwareMessageGroup extends SimpleMessageGroup {
 
+		@Nullable
 		private final SimpleMessageGroup sourceGroup;
 
 		public SequenceAwareMessageGroup(MessageGroup messageGroup) {
@@ -1124,6 +1142,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		}
 
 		@Override
+		@Nullable
 		public Object processMessageGroup(MessageGroup group) {
 			forceComplete(group);
 			return null;
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -65,7 +65,6 @@ import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHandlingException;
 import org.springframework.messaging.core.DestinationResolutionException;
 import org.springframework.messaging.support.GenericMessage;
-import org.springframework.scheduling.TaskScheduler;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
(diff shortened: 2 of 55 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractAggregatingMessageGroupProcessor.java
```diff
@@ -117,6 +119,7 @@ protected Map<String, Object> aggregateHeaders(MessageGroup group) {
 		return getHeadersFunction().apply(group);
 	}
 
+	@Nullable
```
artembilan: I don't think this is OK.
The `payload` cannot be `null` for the message we are going to produce from an aggregator.
If it is potentially null in the specific implementation, let's address there, e.g. with an `Assert.state()`!

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -641,8 +652,10 @@ protected boolean isExpireGroupsUponCompletion() {
 	}
 
 	private void removeEmptyGroupAfterTimeout(UUID groupId, long timeout) {
+		TaskScheduler taskScheduler = getTaskScheduler();
+		Assert.notNull(taskScheduler, "'taskScheduler' must not be null");
```
artembilan: We probably need to address this in the `IntegrationObjectSupport.getTaskScheduler()`.
Make it really non-null and have this `Assert.notNull` over there in the end of method.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/ExpressionEvaluatingMessageGroupProcessor.java
```diff
@@ -58,6 +60,7 @@ public void setExpectedType(Class<?> expectedType) {
 	 * {@link org.springframework.integration.core.MessagingTemplate} to send downstream.
 	 */
 	@Override
+	@Nullable
 	protected Object aggregatePayloads(MessageGroup group, Map<String, Object> headers) {
 		return this.processor.process(group.getMessages());
```
artembilan: Yes, that's what I mean. The SpEL may return null, indeed.
But that is not what we expect here.
Let's assert its result before returning!
Or do you see some use-case where it is deliberately returned as null?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
```diff
@@ -62,18 +64,24 @@ public class FluxAggregatorMessageHandler extends AbstractMessageProducingHandle
 	private CorrelationStrategy correlationStrategy =
 			new HeaderAttributeCorrelationStrategy(IntegrationMessageHeaderAccessor.CORRELATION_ID);
 
+	@Nullable
 	private Predicate<Message<?>> boundaryTrigger;
 
-	private Function<Message<?>, Integer> windowSizeFunction = FluxAggregatorMessageHandler::sequenceSizeHeader;
+	@Nullable
+	public Function<Message<?>, Integer> windowSizeFunction = FluxAggregatorMessageHandler::sequenceSizeHeader;
```
artembilan: This is not correct. The `@Nullable` on the property type would mean that `FluxAggregatorMessageHandler::sequenceSizeHeader` lambda is null.
However that is not the case.
Here we are talking about result of `message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_SIZE, Integer.class)`.
Therefore it is `null` for function invocation.
Hence, has to be like this:
```
private Function<Message<?>, @Nullable Integer> windowSizeFunction = FluxAggregatorMessageHandler::sequenceSizeHeader;
```

I'm not sure why have you made this property as `public`, though...

spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
```diff
@@ -99,13 +109,15 @@ private Flux<Message<?>> releaseBy(Flux<Message<?>> groupFlux) {
 				.flatMap((windowFlux) -> windowFlux.transform(this.combineFunction));
 	}
 
+	@NullUnmarked
```
artembilan: When you do `Function<Message<?>, @Nullable Integer>`, we won't need this anymore.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/MessageListProcessor.java
```diff
@@ -27,6 +29,7 @@
 @FunctionalInterface
 public interface MessageListProcessor {
 
+	@Nullable
```
artembilan: I think this one has to be non-null as well.

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -128,6 +131,7 @@ public static MetadataStore getMetadataStore(BeanFactory beanFactory) {
 	 * @param beanFactory BeanFactory for lookup, must not be null.
 	 * @return The {@link MessageChannel} bean whose name is "errorChannel".
 	 */
+	@Nullable
 	public static MessageChannel getErrorChannel(BeanFactory beanFactory) {
 		return getBeanOfType(beanFactory, ERROR_CHANNEL_BEAN_NAME, MessageChannel.class);
```
artembilan: This bean is always there in the Integration infrastructure.
So, let's put an `Assert.notNull` in the end of this method instead!

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -136,6 +140,7 @@ public static MessageChannel getErrorChannel(BeanFactory beanFactory) {
 	 * @param beanFactory BeanFactory for lookup, must not be null.
 	 * @return The {@link TaskScheduler} bean whose name is "taskScheduler" if available.
 	 */
+	@Nullable
 	public static TaskScheduler getTaskScheduler(BeanFactory beanFactory) {
 		return getBeanOfType(beanFactory, TASK_SCHEDULER_BEAN_NAME, TaskScheduler.class);
```
artembilan: Same for `TaskScheduler`.
See `DefaultConfiguringBeanFactoryPostProcessor` for those infra beans we populate when `@EnableIntegration` is in charge.

artembilan (changes requested):
This is very cool!
Let's see if we can agree on my feedback.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
cppwfs: Even after the change has been applied the following line in the method is getting flagged:
https://github.com/spring-projects/spring-integration/blob/main/spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java#L109

This is because the ` R apply(T t);` from the Function.java has not been annotated with @Nullable.   
Let's discuss to see if their is a smoother resolution than the `@NullMarked`

spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
```diff
@@ -277,6 +289,7 @@ private Mono<Message<?>> messageForWindowFlux(Flux<Message<?>> messageFlux) {
 								.build());
 	}
 
+	@NullUnmarked
 	private static Integer sequenceSizeHeader(Message<?> message) {
```
artembilan: The result of this method can be null, so:
```
private static @Nullable Integer sequenceSizeHeader(Message<?> message)
```
instead of `@NullUnmarked`

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
cppwfs: With the change applied, It still exhibits the same behavior as enumerated here: https://github.com/spring-projects/spring-integration/pull/10097#discussion_r2141065674

artembilan (changes requested):
See my last commit into your PR for the `FluxAggregatorMessageHandler` fix.
At least that what I think about those nulls.

spring-integration-syslog/src/test/java/org/springframework/integration/syslog/inbound/SyslogReceivingChannelAdapterTests.java
```diff
@@ -272,4 +263,12 @@ public void testUdpRFC5424() throws Exception {
 		adapter.stop();
 	}
 
+	private BeanFactory getBeanFactory() {
```
cppwfs, line 266: Some tests needed the bean factory with Mocked components.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -929,6 +937,7 @@ protected Collection<Message<?>> completeGroup(Message<?> message, Object correl
 			this.logger.debug(() -> "Completing group with correlationKey [" + correlationKey + "]");
 
 			result = this.outputProcessor.processMessageGroup(group);
+			Assert.notNull(result, "the group returned a null result");
```
artembilan: I think that is processor is responsible for null result, not group. But again: it really can return null. Might be something is off with our logic here?
Not saying to do that right now with such a huge pr: let’s just have a note to come back to this method!
For now, fix, please, this error message .

spring-integration-core/src/main/java/org/springframework/integration/aggregator/ExpressionEvaluatingMessageListProcessor.java
```diff
@@ -102,7 +105,9 @@ public void setExpectedType(Class<?> expectedType) {
 	 */
 	@Override
 	public Object process(Collection<? extends Message<?>> messages) {
-		return this.evaluateExpression(this.expression, messages, this.expectedType);
+		Object object = this.evaluateExpression(this.expression, messages, this.expectedType);
+		Assert.state(object != null, "Failed to evaluate expression: " + this.expression);
```
artembilan: I don’t think it is really correct . The expression has been just evaluated to null, but we don’t expect null here. Let’s say exactly that condition: the expression evaluated to null and therefore this exception .

spring-integration-core/src/main/java/org/springframework/integration/aggregator/ExpressionEvaluatingMessageGroupProcessor.java
```diff
@@ -59,7 +60,9 @@ public void setExpectedType(Class<?> expectedType) {
 	 */
 	@Override
 	protected Object aggregatePayloads(MessageGroup group, Map<String, Object> headers) {
-		return this.processor.process(group.getMessages());
+		Object object = this.processor.process(group.getMessages());
+		Assert.notNull(object, "Result from processor must not be null");
```
artembilan: Use `state()` instead. This really not an illegal arg.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/MethodInvokingMessageGroupProcessor.java
```diff
@@ -87,7 +88,9 @@ public void setBeanFactory(BeanFactory beanFactory) {
 	@Override
 	protected final Object aggregatePayloads(MessageGroup group, Map<String, Object> headers) {
 		final Collection<Message<?>> messagesUpForProcessing = group.getMessages();
-		return this.processor.process(messagesUpForProcessing, headers);
+		Object object = this.processor.process(messagesUpForProcessing, headers);
+		Assert.notNull(object, "Result from processor must not be null");
```
artembilan: Same here: `state()`

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
artembilan: Was not addressed

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
```diff
@@ -208,6 +224,7 @@ public final void setPrimaryExpression(Expression expression) {
 	}
 
 	@Override
+	@SuppressWarnings("NullAway")
```
artembilan, line 227: We need to revise this latter.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/TcpReceivingChannelAdapterTests.java
```diff
@@ -442,47 +445,53 @@ public void testNioSingleSharedMany() throws Exception {
...
 	@Test
 	public void testNetInterceptors() throws Exception {
-		AbstractServerConnectionFactory scf = new TcpNetServerConnectionFactory(0);
+		AbstractServerConnectionFactory scf = getDefaultServerConnectionFactory();
 		noopPublisher(scf);
 		interceptorsGuts(scf);
 		scf.stop();
 	}
 
+	private AbstractServerConnectionFactory getDefaultServerConnectionFactory() {
```
artembilan: `static` and probably move to the end of class

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/CachingClientConnectionFactoryTests.java
```diff
@@ -820,5 +822,4 @@ private static AbstractClientConnectionFactory createFactoryWithMockConnection(T
 		when(factory.isActive()).thenReturn(true);
 		return factory;
 	}
-
```
artembilan, line 823: See if you can use https://github.com/spring-projects/spring-integration/blob/main/src/idea/spring-framework.xml in your IDEA to avoid this kind of change.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
```diff
@@ -698,5 +699,18 @@ private static AbstractClientConnectionFactory createFactoryWithMockConnection(T
 		return factory;
 	}
 
+	private TcpNetServerConnectionFactory getTcpNetServerConnectionFactory(int port) {
```
artembilan, line 702: `static`?

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
```diff
@@ -698,5 +699,18 @@ private static AbstractClientConnectionFactory createFactoryWithMockConnection(T
...
+		TcpNetServerConnectionFactory result = new TcpNetServerConnectionFactory(port);
+		result.setTaskScheduler(new SimpleAsyncTaskScheduler());
+
+		return result;
+	}
+
+	private TcpNetClientConnectionFactory getTcpNetServerConnectionFactory(String host, int port) {
```
artembilan: DITTO

spring-integration-syslog/src/test/java/org/springframework/integration/syslog/inbound/SyslogReceivingChannelAdapterTests.java
```diff
@@ -251,4 +263,12 @@ public void testUdpRFC5424() throws Exception {
...
+		BeanFactory beanFactory = mock(BeanFactory.class);
+		TaskScheduler taskScheduler = mock(TaskScheduler.class);
+		when(beanFactory.getBean(eq("taskScheduler"), any(Class.class)))
+				.thenReturn(taskScheduler);
+		when(beanFactory.containsBean("taskScheduler")).thenReturn(true);
+		return beanFactory;
+	}
```
artembilan, line 273: Empty line before the last `}`vin the class.
In other words: every member in the class has to be surrounded with blank lines. 
I think the editor config for IDEA I suggested before should do the trick when you reformat files.

artembilan (changes requested):
This is cool!
We can chat on Monday if you wish.
I also hope you do not update Copyright year manually . We have a Gradle task to do that for us 😉

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -125,8 +125,10 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	private boolean releaseStrategySet;
 
+	@SuppressWarnings("NullAway.Init")
```
artembilan, line 128: This is not correct on this property.
It is initialized in the `getDiscardChannel()`.
So, it has to be `@Nullable` and that getter has to be used instead in other places.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/ExpressionEvaluatingMessageListProcessor.java
```diff
@@ -102,7 +105,9 @@ public void setExpectedType(Class<?> expectedType) {
 	 */
 	@Override
 	public Object process(Collection<? extends Message<?>> messages) {
-		return this.evaluateExpression(this.expression, messages, this.expectedType);
+		Object object = this.evaluateExpression(this.expression, messages, this.expectedType);
+		Assert.state(object != null, "The evaluation of the expression returned a null. Null result is not expected." + this.expression);
```
artembilan, line 109: Probably you mean: `Null result is not expected: " + this.expression` ?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
```diff
@@ -106,7 +115,8 @@ private Flux<Flux<Message<?>>> applyWindowOptions(Flux<Message<?>> groupFlux) {
 		return groupFlux
 				.switchOnFirst((signal, group) -> {
 					if (signal.hasValue()) {
-						Integer maxSize = this.windowSizeFunction.apply(signal.get());
+						Assert.notNull(this.windowSizeFunction, "'windowSizeFunction' must not be null");
```
artembilan, line 118: How can that be null, please?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/FluxAggregatorMessageHandler.java
```diff
@@ -277,8 +288,7 @@ private Mono<Message<?>> messageForWindowFlux(Flux<Message<?>> messageFlux) {
 								.build());
 	}
 
-	private static Integer sequenceSizeHeader(Message<?> message) {
+	private static @Nullable Integer sequenceSizeHeader(Message<?> message) {
 		return message.getHeaders().get(IntegrationMessageHeaderAccessor.SEQUENCE_SIZE, Integer.class);
 	}
-
```
artembilan, line 283: Blank line in the end of class before its last `}`

spring-integration-core/src/main/java/org/springframework/integration/aggregator/MessageListProcessor.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
```
artembilan, line 2: This looks like redundant change for now.
But that's OK.
We will move to `present` instead soon enough 😄

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -129,15 +132,19 @@ public static MetadataStore getMetadataStore(BeanFactory beanFactory) {
 	 * @return The {@link MessageChannel} bean whose name is "errorChannel".
 	 */
 	public static MessageChannel getErrorChannel(BeanFactory beanFactory) {
-		return getBeanOfType(beanFactory, ERROR_CHANNEL_BEAN_NAME, MessageChannel.class);
+		MessageChannel channel = getBeanOfType(beanFactory, ERROR_CHANNEL_BEAN_NAME, MessageChannel.class);
+		Assert.state(channel != null, "Error Channel was not found");
```
artembilan, line 136: Maybe `"No such bean '" + ERROR_CHANNEL_BEAN_NAME + "'"` here as well?

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -156,6 +163,7 @@ public static TaskScheduler getRequiredTaskScheduler(BeanFactory beanFactory) {
 	 * @return the instance of {@link StandardEvaluationContext} bean whose name is
 	 * {@value #INTEGRATION_EVALUATION_CONTEXT_BEAN_NAME}.
 	 */
+	@Nullable
```
artembilan, line 166: This cannot be null.
Another infrastructure bean always present there by Integration infrastructure.

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationContextUtils.java
```diff
@@ -166,11 +174,13 @@ public static StandardEvaluationContext getEvaluationContext(BeanFactory beanFac
 	 * {@value #INTEGRATION_SIMPLE_EVALUATION_CONTEXT_BEAN_NAME}.
 	 * @since 4.3.15
 	 */
+	@Nullable
```
artembilan, line 177: Same here: never null.
Therefore similar to that:
```
Assert.state(taskScheduler != null, "No such bean '" + INTEGRATION_SIMPLE_EVALUATION_CONTEXT_BEAN_NAME + "'");
```
And in the previous, too.

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
artembilan: Any comments why this is still here, please?

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/CachingClientConnectionFactoryTests.java
artembilan: Was not addressed, please

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/FailoverClientConnectionFactoryTests.java
artembilan: Was not addressed

artembilan (changes requested):
Sorry, there are still some review feedback.
But this is great to be merged soon enough.

Thanks

artembilan:
Hey, @cppwfs !

I'm sorry to cause more headache, but we have just pushed a breaking change which might not merge clean into your branch.
That one is about changing Copyright years from let's say `2020-2025` to `2020-present`.
Let's me know and I may help you to mitigate the pain!

artembilan:
Actually it won't be too bad if you run this script before merged with `main`:
```
echo Updating copyright headers in Java, Kotlin, and Groovy source code

for file in $(find . -type f -iregex '.*.\(java\|groovy\|kt\)' | uniq); do
	sed -i 's/Copyright \([0-9]\{4\}\)\(-[0-9]\{4\}\)\? the original author or authors\./Copyright \1-present the original author or authors./g' "$file";
done
```
Amend the changes with one of your previous commit, then merge.
Then I hope no merge conflicts 😄

artembilan:
Consider to rebase to the latest `main`.
Have just pushed redundant `grgit` Gradle plugin usage.
Since we use now `-present` pattern for Copyright, we don't need to track modified files anymore.
That should speed up a local development loop a little bit 😄 .

cppwfs:
Closing this PR and will reopen under a new branch.
The number of changes that have occured because of refactoring and merge conflicts make this branch invalid.

PR #10098, opened 2025-06-11
GH-10083: Migrate `spring-integration-ws` module to Jspecify

anthologia:
- Replace `org.springframework.lang.Nullable` with `org.jspecify.annotations.Nullable`
- Migrate `package-info.java` files to use `@NullMarked` annotation
- Add `@SuppressWarnings("NullAway.Init")` for fields initialized in lifecycle methods

For the array fields `BaseWsOutboundGatewaySpec.messageSenders` and `BaseWsOutboundGatewaySpec.gatewayInterceptors`, I applied `@SuppressWarnings("NullAway.Init")` instead of `@Nullable` because adding `@Nullable` still produces the same NullAway warnings.

### DIFF
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,8 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import javax.xml.transform.TransformerException;
+
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.expression.Expression;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
@@ -68,13 +70,14 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 
 	private final String uri;
 
-	private final DestinationProvider destinationProvider;
+	private final @Nullable DestinationProvider destinationProvider;
 
 	private final Map<String, Expression> uriVariableExpressions = new HashMap<>();
 
+	@SuppressWarnings("NullAway.Init")
 	private StandardEvaluationContext evaluationContext;
 
-	private WebServiceMessageCallback requestCallback;
+	private @Nullable WebServiceMessageCallback requestCallback;
 
 	private WebServiceTemplate webServiceTemplate;
 
@@ -84,15 +87,17 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 
 	private boolean webServiceTemplateExplicitlySet;
 
-	public AbstractWebServiceOutboundGateway(final String uri, WebServiceMessageFactory messageFactory) {
+	@SuppressWarnings("NullAway")
+	public AbstractWebServiceOutboundGateway(final String uri, @Nullable WebServiceMessageFactory messageFactory) {
 		Assert.hasText(uri, "URI must not be empty");
 		this.webServiceTemplate = new WebServiceTemplate(messageFactory);
 		this.destinationProvider = null;
 		this.uri = uri;
 	}
 
+	@SuppressWarnings("NullAway")
 	public AbstractWebServiceOutboundGateway(DestinationProvider destinationProvider,
-			WebServiceMessageFactory messageFactory) {
+			@Nullable WebServiceMessageFactory messageFactory) {
 
 		Assert.notNull(destinationProvider, "DestinationProvider must not be null");
 		this.webServiceTemplate = new WebServiceTemplate(messageFactory);
@@ -165,7 +170,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 		this.webServiceTemplate.setMessageFactory(messageFactory);
 	}
 
-	public void setRequestCallback(WebServiceMessageCallback requestCallback) {
+	public void setRequestCallback(@Nullable WebServiceMessageCallback requestCallback) {
 		this.requestCallback = requestCallback;
 	}
 
@@ -199,7 +204,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 	}
 
 	@Override
-	public final Object handleRequestMessage(Message<?> requestMessage) {
+	public final @Nullable Object handleRequestMessage(Message<?> requestMessage) {
 		URI uriWithVariables = prepareUri(requestMessage);
 		if (uriWithVariables == null) {
 			throw new MessageDeliveryException(requestMessage, "Failed to determine URI for " +
@@ -215,7 +220,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 		return null;
 	}
 
-	private URI prepareUri(Message<?> requestMessage) {
+	private @Nullable URI prepareUri(Message<?> requestMessage) {
 		if (this.destinationProvider != null) {
 			return this.destinationProvider.getDestination();
 		}
@@ -229,17 +234,17 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 		return this.uriFactory.expand(this.uri, uriVariables);
 	}
 
-	protected abstract Object doHandle(String theUri, Message<?> requestMessage,
-			WebServiceMessageCallback reqCallback);
+	protected abstract @Nullable Object doHandle(String theUri, Message<?> requestMessage,
+			@Nullable WebServiceMessageCallback reqCallback);
 
 	protected abstract class RequestMessageCallback extends TransformerObjectSupport
 			implements WebServiceMessageCallback {
 
-		private final WebServiceMessageCallback reqCallback;
+		private final @Nullable WebServiceMessageCallback reqCallback;
 
 		private final Message<?> requestMessage;
 
-		public RequestMessageCallback(WebServiceMessageCallback requestCallback, Message<?> requestMessage) {
+		public RequestMessageCallback(@Nullable WebServiceMessageCallback requestCallback, Message<?> requestMessage) {
 			this.reqCallback = requestCallback;
 			this.requestMessage = requestMessage;
 		}
@@ -266,7 +271,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 			implements WebServiceMessageExtractor<Object> {
 
 		@Override
-		public Object extractData(WebServiceMessage message)
+		public @Nullable Object extractData(WebServiceMessage message)
 				throws IOException, TransformerException {
 
 			Object resultObject = this.doExtractData(message);
@@ -284,7 +289,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 			}
 		}
 
-		public abstract Object doExtractData(WebServiceMessage message) throws IOException, TransformerException;
+		public abstract @Nullable Object doExtractData(WebServiceMessage message) throws IOException, TransformerException;
 
 	}
 
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
@@ -69,7 +69,7 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 
 	protected final DefaultUriBuilderFactory uriFactory = new DefaultUriBuilderFactory(); // NOSONAR - final
 
-	private final String uri;
+	private final @Nullable String uri;
 
 	private final @Nullable DestinationProvider destinationProvider;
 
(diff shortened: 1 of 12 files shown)


anthologia:
I found some packages don't have `@org.jspecify.annotations.NullMarked` applied.. I'll add them in a follow-up commit !

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
```diff
@@ -51,26 +53,36 @@ public abstract class BaseWsOutboundGatewaySpec<
...
 	private boolean ignoreEmptyResponses = true;
 
+	@Nullable
 	private WebServiceMessageCallback requestCallback;
 
+	@Nullable
 	protected FaultMessageResolver faultMessageResolver; // NOSONAR
 
+	@SuppressWarnings("NullAway.Init")
 	protected WebServiceMessageSender[] messageSenders; // NOSONAR
```
artembilan: This has to be like this:
```
protected WebServiceMessageSender @Nullable [] messageSenders;
```
See more info in docs: https://docs.spring.io/spring-framework/reference/7.0/core/null-safety.html#null-safety-guidelines-jpecify

> `Object @Nullable [] array` means individual elements can’t be null but the array itself can.

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
```diff
@@ -51,26 +53,36 @@ public abstract class BaseWsOutboundGatewaySpec<
...
 
+	@Nullable
 	protected WebServiceMessageFactory webServiceMessageFactory; // NOSONAR
 
+	@Nullable
 	private SoapHeaderMapper headerMapper;
 
-	private DefaultUriBuilderFactory.EncodingMode encodingMode;
+	@Nullable
+	private EncodingMode encodingMode;
```
artembilan: I prefer to use it with an outer class yet.
Therefore this one has to be:
```
private DefaultUriBuilderFactory.@Nullable EncodingMode encodingMode;
```
See more info in docs: https://docs.spring.io/spring-framework/reference/7.0/core/null-safety.html#null-safety-guidelines-jpecify

> The Java specifications also enforces that annotations defined with `@Target(ElementType.TYPE_USE)` like JSpecify `@Nullable` should be specified after the last `.` with inner or fully qualified types:

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsOutboundGatewaySpec.java
```diff
@@ -63,8 +63,10 @@ public static class MarshallingWsOutboundGatewayNoTemplateSpec
 			extends BaseWsOutboundGatewaySpec<MarshallingWsOutboundGatewayNoTemplateSpec,
 			MarshallingWebServiceOutboundGateway> {
 
+		@SuppressWarnings("NullAway.Init")
 		protected Marshaller gatewayMarshaller; // NOSONAR
 
+		@SuppressWarnings("NullAway.Init")
```
artembilan: This has to be `@Nullable`.
See target `MarshallingWebServiceOutboundGateway`.
Looks like `Marshaller` cannot be.
Let's see if we can set it into a `Jaxb2Marshaller` by default!

artembilan (changes requested):

artembilan:
> I'll add them in a follow-up commit !

Right. Let's finish with the current change you have so far and then we will go from there for others.
If that is more convenient for you, we may even go separate PRs.

Thanks

anthologia:
I've applied all the requested changes!
- Set `gatewayMarshaller` to use `Jaxb2Marshaller` as default since it's an implementation of `Marshaller`. Thanks!
- Updated all `@Nullable` annotations to be placed between access modifier and type for consistency.
  - Should we standardize `@Nullable` placement convention project-wide? 🤔 Currently there seems to be mixed usage.
  
I'll add commits tomorrow for other packages in the `ws` module to this PR rather than separate PRs! Is that fine?

artembilan:
> Currently there seems to be mixed usage.

I know, but the easier way to proceed is to replace imports and leave as is.
We can clean up the proper placement as we will go eventually.
If that is fine for you, feel free to make it as it is recommended by JSpecify, but I'm not picky on that.

It is fine to continue in this PR.

Thanks.

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/MarshallingWsOutboundGatewaySpec.java
```diff
@@ -63,9 +66,9 @@ public static class MarshallingWsOutboundGatewayNoTemplateSpec
 			extends BaseWsOutboundGatewaySpec<MarshallingWsOutboundGatewayNoTemplateSpec,
 			MarshallingWebServiceOutboundGateway> {
 
-		protected Marshaller gatewayMarshaller; // NOSONAR
+		protected Marshaller gatewayMarshaller = new Jaxb2Marshaller(); // NOSONAR
```
artembilan: Therefore we can remove that noisy `// NOSONAR`.

Honestly we discontinued to use SonarQube in our projects.
So, all those comments are redundant from now on.

Thanks

artembilan (changes requested):

spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
```diff
@@ -84,15 +87,17 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 
 	private boolean webServiceTemplateExplicitlySet;
 
-	public AbstractWebServiceOutboundGateway(final String uri, WebServiceMessageFactory messageFactory) {
+	@SuppressWarnings("NullAway")
+	public AbstractWebServiceOutboundGateway(final String uri, @Nullable WebServiceMessageFactory messageFactory) {
```
artembilan: This is not correct. 
The `WebServiceTemplate` expects from us a `WebServiceMessageFactory` as a not null.
If you see somewhere it is, let's mitigate it over there!

spring-integration-ws/src/main/java/org/springframework/integration/ws/SimpleWebServiceOutboundGateway.java
```diff
@@ -68,7 +69,7 @@ public SimpleWebServiceOutboundGateway(DestinationProvider destinationProvider,
 
 	public SimpleWebServiceOutboundGateway(DestinationProvider destinationProvider,
 			@Nullable SourceExtractor<?> sourceExtractor,
-			WebServiceMessageFactory messageFactory) {
+			@Nullable WebServiceMessageFactory messageFactory) {
```
artembilan, line 72: We have to fix those places to not provide a null for the `WebServiceMessageFactory`.

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
```diff
@@ -45,11 +46,12 @@ public class MarshallingWebServiceOutboundGateway extends AbstractWebServiceOutb
 
 	@SuppressWarnings("this-escape")
 	public MarshallingWebServiceOutboundGateway(DestinationProvider destinationProvider, Marshaller marshaller,
-			Unmarshaller unmarshaller, WebServiceMessageFactory messageFactory) {
+			@Nullable Unmarshaller unmarshaller, @Nullable WebServiceMessageFactory messageFactory) {
 		super(destinationProvider, messageFactory);
 		configureMarshallers(marshaller, unmarshaller);
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: I'm not very OK with these `@SuppressWarnings("NullAway")`, but we can look into that separately if that is too much for this change cycle.

artembilan (changes requested):

spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
anthologia: then.. when ctors pass null for `WebServiceMessageFactory`, I'll replace null with `SaajSoapMessageFactory` as the default. Does this sound good?

Based on the [WebServiceTemplate.properties](https://github.com/spring-projects/spring-ws/blob/main/spring-ws-core/src/main/resources/org/springframework/ws/client/core/WebServiceTemplate.properties), `SaajSoapMessageFactory` is indeed the default strategy.

artembilan: Why would we allow to pass null at all?
The `WebServiceTemplate` does not expect that from us, so let's comeback to not-null for this ctors!

anthologia: Right, I meant that I'll remove `@Nullable` from the constructor parameters and fix the calling sites to pass `SaajSoapMessageFactory` instance instead of null.

For example:
```java
// Instead of
super(uri, null);

// I'll change to  
super(uri, new SaajSoapMessageFactory());
```

Am I still misunderstanding something?

artembilan: I see:
```
	public MarshallingWebServiceOutboundGateway(DestinationProvider destinationProvider,
			WebServiceTemplate webServiceTemplate) {
		super(destinationProvider, null);
		doSetWebServiceTemplate(webServiceTemplate);
	}
```
I think it is better to introduce a `protected` ctor in the `AbstractWebServiceOutboundGateway` based on the `WebServiceTemplate` then.

artembilan:
Hey @anthologia !

Any updates, please?

anthologia:
@artembilan

Sorry for the delay ! I just pushed the updates 😊

I noticed something while working on this - the `WebServiceTemplate(WebServiceMessageFactory messageFactory)` ctor param is marked as non-null, but internally it calls `initDefaultStrategies()` which handles null cases by initializing default strategies.

Should the `messageFactory` param actually be `@Nullable` in the WebServiceTemplate ctor? It seems like the implementation already supports null values through the default initialization mechanism.

What do you think?

artembilan:
The logic there is like this:
```
	/** Creates a new {@code WebServiceTemplate} using default settings. */
	public WebServiceTemplate() {
		initDefaultStrategies();
	}

	/**
	 * Creates a new {@code WebServiceTemplate} based on the given message factory.
	 * @param messageFactory the message factory to use
	 */
	public WebServiceTemplate(WebServiceMessageFactory messageFactory) {
		setMessageFactory(messageFactory);
		initDefaultStrategies();
	}
```

So, the contract enforces us to provide not-null if we chose that specific ctor.

You may raise a GH issue in [Spring WS](https://github.com/spring-projects/spring-ws) if you still think that it is OK to have it nullable in the second ctor since the first one simply assumes that it really can be null.

spring-integration-ws/src/main/java/org/springframework/integration/ws/AbstractWebServiceOutboundGateway.java
```diff
@@ -84,13 +88,46 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
...
+	public AbstractWebServiceOutboundGateway(final String uri) {
+		Assert.hasText(uri, "URI must not be empty");
+		this.webServiceTemplate = new WebServiceTemplate();
+		this.destinationProvider = null;
+		this.uri = uri;
+	}
+
 	public AbstractWebServiceOutboundGateway(final String uri, WebServiceMessageFactory messageFactory) {
 		Assert.hasText(uri, "URI must not be empty");
 		this.webServiceTemplate = new WebServiceTemplate(messageFactory);
```
artembilan: While having your comment about `WebServiceTemplate`, I think it is fine to come back to your original request to make this `messageFactory` param as `@Nullable`, then here we would do:
```
this.webServiceTemplate = messageFactory != null ? new WebServiceTemplate(messageFactory) : new WebServiceTemplate();
```

And probably we would not need all those new ctors you have introduced.

Thanks.

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
artembilan: This was not addressed.

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
```diff
@@ -60,17 +62,19 @@ public MarshallingWebServiceOutboundGateway(DestinationProvider destinationProvi
 		this(destinationProvider, marshaller, null, messageFactory);
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: DITTO

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
```diff
@@ -60,17 +62,19 @@ public MarshallingWebServiceOutboundGateway(DestinationProvider destinationProvi
...
 
 	@SuppressWarnings("this-escape")
-	public MarshallingWebServiceOutboundGateway(String uri, Marshaller marshaller, Unmarshaller unmarshaller,
+	public MarshallingWebServiceOutboundGateway(String uri, Marshaller marshaller, @Nullable Unmarshaller unmarshaller,
 			WebServiceMessageFactory messageFactory) {
 		super(uri, messageFactory);
 		configureMarshallers(marshaller, unmarshaller);
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: DITTO

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
```diff
@@ -80,6 +84,7 @@ public MarshallingWebServiceOutboundGateway(String uri, Marshaller marshaller,
 		this(uri, marshaller, null, messageFactory);
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: DITTO

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
```diff
@@ -51,27 +53,30 @@ public abstract class BaseWsOutboundGatewaySpec<
...
 	protected WebServiceTemplate template; // NOSONAR
 
-	protected DestinationProvider destinationProvider; // NOSONAR
+	protected @Nullable DestinationProvider destinationProvider; // NOSONAR
 
+	@SuppressWarnings("NullAway.Init")
 	protected String uri; // NOSONAR
 
+	@SuppressWarnings("NullAway.Init")
 	protected WebServiceMessageFactory webServiceMessageFactory; // NOSONAR
```
artembilan: The `template` comes from the ctor in the impl classes.
So, I guess, not-null and `final`.
But these two are neither initialized in ctors, nor in the `afterPropertiesSet()`, therefore they have to be marked as `@Nullable`.

spring-integration-ws/src/test/java/org/springframework/integration/ws/config/WebServiceOutboundGatewayParserTests.java
```diff
@@ -382,7 +382,7 @@ public void simpleGatewayWithDestinationProvider() {
 		DirectFieldAccessor accessor = new DirectFieldAccessor(gateway);
 		assertThat(accessor.getPropertyValue("destinationProvider")).as("Wrong DestinationProvider")
 				.isEqualTo(stubProvider);
-		assertThat(accessor.getPropertyValue("uri")).isNull();
+		assertThat(accessor.getPropertyValue("uri")).isEqualTo("");
```
artembilan: I think `uri` must be `@Nullable` in the `AbstractWebServiceOutboundGateway`.
I know there is a `prepareUri()` method in the `AbstractWebServiceOutboundGateway` which expects `this.uri` to not be null, but exactly that one has to have an `Assert.notNull(this.uri)` before calling `this.uriFactory.expand()`.

artembilan (changes requested):
Thank you for an update.
Please, give me some feedback on those comments I have left.
As a fix or reject with respective arguments.

spring-integration-ws/src/main/java/org/springframework/integration/ws/dsl/BaseWsOutboundGatewaySpec.java
anthologia: I agree that `webServiceMessageFactory` should be `@Nullable`.

But I'm not sure if `template` should be `@Nullable`. Because excluding the `*NoTemplateSpec`, two meaningful impls (`SimpleWsOutboundGatewaySpec` and `MarshallingWsOutboundGatewaySpec`) both receive `template` as a non-null param in their ctors.

And, if `template` becomes `@Nullable`, the setter would end up looking like this, which feels awkward to me. 😞

```java
// or SupressWarning()
protected final void doSetWebServiceTemplate(@Nullable WebServiceTemplate template) {
    Assert.notNull(template, "'webServiceTemplate' must not be null");
    this.webServiceTemplate = template;
    this.webServiceTemplateExplicitlySet = true;
}
```

artembilan: Sorry I meant like this:
```
protected final WebServiceTemplate template;

protected @Nullable String uri;

protected @Nullable WebServiceMessageFactory webServiceMessageFactory;
```

Therefore `template` is not null as you've just explained.
And that is exactly what I meant with original comment on this line.

anthologia: Ah, got it! That makes perfect sense. Thank you.

spring-integration-ws/src/test/java/org/springframework/integration/ws/config/WebServiceOutboundGatewayParserTests.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
```
artembilan, line 2: This is not the first time.
So, after merging your PR, I'm going to apply our new `-present` style for Copyright.
That would make our life easier 😄

PR #10102, opened 2025-06-12
GH-10100: Fix NPE in `AbstractMailReceiver.obtainFolderInstance()`.

### LINKED ISSUE #10100: Mail: Proposal for removing constructors in `AbstractMailReceiver` which support `url` to be null
current in `AbstractMailReceiver`, we have 3 constructors.

```java
        private final URLName url;

	public AbstractMailReceiver() {
		this.url = null;
	}

	public AbstractMailReceiver(URLName urlName) {
		Assert.notNull(urlName, "urlName must not be null");
		this.url = urlName;
	}

	public AbstractMailReceiver(String url) {
		if (url != null) {
			this.url = new URLName(url);
		}
		else {
			this.url = null;
		}
	}
``` 

it says `url` could be null in some cases. 

however, inside the `AbstractMailReceiver.receive()`, we will call below:
```java
	private Folder obtainFolderInstance() throws MessagingException {
		return this.store.getFolder(this.url);
	}
```
here, if `this.url` is null, `getFolder` will throw an NPE.

So, I think it's safe to remove this constructors.

mjd507:
Fixes: #10100

if `url` is not provided, use `store.getDefaultFolder()` instead.

### DIFF
--- a/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
+++ b/spring-integration-mail/src/main/java/org/springframework/integration/mail/AbstractMailReceiver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


mjd507:
wrong fixes in commit, @artembilan can you help me to modify the fixes in commit from 10083 to 10100

PR #10106, opened 2025-06-12
GH-10101: Mitigate constant access to the nullable fields in AbstractMailReceiver

### LINKED ISSUE #10101: Mail: Mitigate constant access to the nullable fields (`folder` & `store`) in `AbstractMailReceiver`
both `folder` & `store` fields are nullable in `AbstractMailReceiver`.

since we are going to apply `jspecify` for nullability check, in order to minimize the scope of impact,

I want to mitigate them with local variables instead of constant access. 

so it will be clear for me when it's null, when it's not null

mjd507:
Fixes: #10101


artembilan (changes requested):
Well, I disagree with such a bit breaking change.
Isn't it easier to extract local variable in the beginning of those methods instead?
We develop here a library and that is too destructive to introduce those method arguments.

mjd507:
all right, after more thinking, I decide to close this, will be implement in mail nullability PR

PR #10107, opened 2025-06-13
gh-10083: Scope: spring-integration-cassandra; Replace springframework Nullable with jspecify Nullable

anayonkars:
1. Replaced springframework Nullable with jsepcify Nullable.
2. Replaced springframewor NonNullApi and NonNullFields with jspecify NullMarked

### DIFF
--- a/spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
+++ b/spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
@@ -163,14 +163,16 @@ public class CassandraMessageHandler extends AbstractReplyProducingMessageHandle
 		this.parameterExpressions.putAll(parameterExpressions);
 	}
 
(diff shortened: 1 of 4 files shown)


spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
```diff
@@ -161,6 +163,7 @@ public void setParameterExpressions(Map<String, Expression> parameterExpressions
 		this.parameterExpressions.putAll(parameterExpressions);
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: Can we look into fixing the null problem instead of suppressing it?
This Cassandra module is short enough to spend more time on its nullability, rather than suppress it and forget eventually.
There is also a `config.xml` package which comes without `@org.jspecify.annotations.NullMarked`.
Would you mind addressing that as well?

artembilan (changes requested):

PR #10109, opened 2025-06-14
GH-10083: Apply Nullability for spring-integration-core

anayonkars:
This is applicable only for `spring-integration-core`
1. Replaced `org.springframework.lang.Nullable` with `org.jspecify.annotations.Nullable`
2. Replaced `org.springframework.lang.NonNull` with `org.jspecify.annotations.NonNull`

**TODO:** Need to work on Nullability for package-info


artembilan (changes requested):
Well, actually I can accept this PR. This is pretty clean for the purpose of its goal .
We can work on the rest later.
Thank you!

PR #10138, opened 2025-06-17
Apply Nullability to `spring-integration-kafka`

mjd507:
related to #10083

Suppress warning for child classes which parent class (`RecordMessagingMessageListenerAdapter` or `BatchMessagingMessageListenerAdapter`) are in `spring-kafka` module, from the parent constructor, `(Object bean, Method method)`  are non null, but we give(null, null) in `spring-integration-kafka`.  not sure how to solve it.


artembilan:
> `(Object bean, Method method)` are non `null`, but we give `(null, null)`.

Please, raise a GH issue in https://github.com/spring-projects/spring-kafka.
Doesn't look like those args required to be not-null.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/PollableKafkaChannel.java
```diff
@@ -195,6 +196,7 @@ public boolean hasExecutorInterceptors() {
 		return this.executorInterceptorsSize > 0;
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: Why this one, please?

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
```diff
@@ -176,15 +179,16 @@ public boolean unsubscribe(MessageHandler handler) {
 		return this.dispatcher.removeHandler(handler);
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: May not on the whole class, but just that ctor?

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -281,6 +281,7 @@ protected AttributeAccessor getErrorMessageAttributes(Message<?> message) {
 		}
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: Same here: let's see if we can have it only on ctor!

artembilan (changes requested):
Just couple nit-picks.
Overall this is great.
Thank you!

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/PollableKafkaChannel.java
mjd507: nice catch, fixed

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
mjd507: I removed this warning, opened a [PR](https://github.com/spring-projects/spring-kafka/pull/3980) in `spring-kafka`, once that merged & published, we can try re-build this.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
mjd507: removed, waiting for `spring-kafka`

mjd507:
in local, I add `mavenLocal=true` in `gradle.properties`, to use local `spring-kafka` changes,  `check` build successful, 
so I think once `spring-kafka` publish my changes, this build will be green.

artembilan:
>  I add `mavenLocal=true` in `gradle.properties`

Can you point us to the documentation on the matter, please?
This is nice trick but you also could use a composite Gradle build: https://docs.gradle.org/current/userguide/composite_builds.html

artembilan:
Ah! I see.
That is not official doc, but rather our own arrangement in the Gradle config 😄 :
```
    repositories {
        if (project.hasProperty('mavenLocal')) {
            mavenLocal()
        }
```

artembilan:
Hey, @mjd507 !

Any updates?
The related PR in Spring Kafka has been merged.

Thanks

mjd507:
merge `main` to trigger build. now its green.

PR #10157, opened 2025-06-22
Apply Nullability to `spring-integration-jpa`

mjd507:
related to: #10083

### DIFF
--- a/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
+++ b/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
@@ -160,7 +160,7 @@ public class JpaExecutor implements InitializingBean, BeanFactoryAware {
 	}
 
 	/**
-	 * deprecated setIntegrationEvaluationContext, in favor of the one obtained from the application context.
+	 * @deprecated in favor of the one obtained from the application context.
 	 */
 	@Deprecated(since = "7.0", forRemoval = true)
 	public void setIntegrationEvaluationContext(EvaluationContext evaluationContext) {
@@ -613,9 +613,8 @@ public class JpaExecutor implements InitializingBean, BeanFactoryAware {
 	}
 
 	private ParameterSource determineParameterSource(final Message<?> requestMessage) {
-		Assert.state(this.usePayloadAsParameterSource != null, "'usePayloadAsParameterSource' must not be null");
 		Assert.state(this.parameterSourceFactory != null, "'parameterSourceFactory' must not be null");
-		if (this.usePayloadAsParameterSource) {
+		if (Boolean.TRUE.equals(this.usePayloadAsParameterSource)) {
 			return this.parameterSourceFactory.createParameterSource(requestMessage.getPayload());
 		}
 		else {
--- a/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
+++ b/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
@@ -20,6 +20,7 @@ import java.util.List;
 
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.EntityManagerFactory;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
@@ -35,7 +36,6 @@ import org.springframework.integration.jpa.support.parametersource.BeanPropertyP
 import org.springframework.integration.jpa.support.parametersource.ExpressionEvaluatingParameterSourceFactory;
 import org.springframework.integration.jpa.support.parametersource.ParameterSource;
 import org.springframework.integration.jpa.support.parametersource.ParameterSourceFactory;
-import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
 import org.springframework.util.Assert;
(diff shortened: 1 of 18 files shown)


spring-integration-jpa/src/main/java/org/springframework/integration/jpa/outbound/JpaOutboundGatewayFactoryBean.java
```diff
@@ -38,6 +39,7 @@
  */
 public class JpaOutboundGatewayFactoryBean extends AbstractSimpleMessageHandlerFactoryBean<JpaOutboundGateway> {
 
+	@SuppressWarnings("NullAway") // will be initialized in xml parser
```
artembilan: We cannot assume how this class is going to be used. So, make it nullable and assert in the `createHandler()`

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
```diff
@@ -111,11 +111,11 @@ public class JpaExecutor implements InitializingBean, BeanFactoryAware {
...
 	 * used for the sqlParameterSourceFactory property.
 	 */
-	private Boolean usePayloadAsParameterSource = null;
+	private @Nullable Boolean usePayloadAsParameterSource = null;
 
-	private BeanFactory beanFactory;
+	private @Nullable BeanFactory beanFactory;
 
-	private EvaluationContext evaluationContext;
+	private @Nullable EvaluationContext evaluationContext;
```
artembilan: The `BeanFactory` and this should be set from `afterPropertiesSet()`, there suppress warning with `Init`.

artembilan (changes requested):

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
mjd507: changed, but I have a query, in `afterPropertiesSet()`,  we have below:
```java
if (this.evaluationContext == null) {
    this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(this.beanFactory);
}
```
here we have a warning in `IntelliJ IDEA`: **Condition 'this.evaluationContext == null' is always 'false'**
but I don't know how to fix this warning in `IDEA`?  do you know that?

Also, in `AbstractJpaOperations`, I changed field `entityManager` to use `@SuppressWarnings("NullAway.Init")` as well.  it is also initialized in `afterPropertiesSet()`, and similarly has same warning: **Condition 'this.entityManager == null' is always 'false'**

In few modules I worked earlier, I may face this but forget to raise it, I use idea community version.

artembilan: I think we need to deprecated that `JpaExecutor.setIntegrationEvaluationContext()` in favor of the one obtained from the application context.
And then remove that `if (this.evaluationContext == null) {` from its `afterPropertiesSet()`.

artembilan: The `entityManager` in the `AbstractJpaOperations` has to be `@Nullable` since there is a choice between setting an `EntityManager` or `EntityManagerFactory`.

artembilan: Then that `getEntityManager()` should use `Objects.requireNonNull()`.

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
```diff
@@ -398,10 +404,7 @@ public void afterPropertiesSet() {
 		else if (this.flush) {
 			this.flushSize = 1;
 		}
-
-		if (this.evaluationContext == null) {
```
mjd507, line 402: is it harmful to remove in this version?

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
```diff
@@ -157,6 +159,10 @@ public JpaExecutor(JpaOperations jpaOperations) {
 		this.jpaOperations = jpaOperations;
 	}
 
+	/**
+	 * deprecated setIntegrationEvaluationContext, in favor of the one obtained from the application context.
```
artembilan: Has to be `@deprecated`.

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
artembilan: I don't think so.
This is major version, so we can have some breaking changes.
But I don't see how this one could be.

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
```diff
@@ -609,6 +613,8 @@ else if (evaluationResult instanceof String) {
 	}
 
 	private ParameterSource determineParameterSource(final Message<?> requestMessage) {
+		Assert.state(this.usePayloadAsParameterSource != null, "'usePayloadAsParameterSource' must not be null");
+		Assert.state(this.parameterSourceFactory != null, "'parameterSourceFactory' must not be null");
```
artembilan, line 616: I think we need to check here only for `this.parameterSourceFactory`.
And I guess an `Assert.notNull()` is OK here.
The next one has to be `if (Boolean.TRUE.equals(this.usePayloadAsParameterSource)) {`

artembilan (changes requested):
Thank you for the update!
This is very close to the merge.

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
mjd507: okay, so instead of depredated method `setIntegrationEvaluationContext()`, can we simply remove that method?

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
mjd507: tried `Assert.notNull()` not ok, `Assert.state()` is ok.  don't know why

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/core/JpaExecutor.java
artembilan: No. We never remove a public API in our classes just because.
Has to be deprecated first.

Issue #10058, opened 2025-05-28
Add Jackson 3 support and deprecate Jackson 2 APIs


artembilan:
This is where we use it:
* `ConfigurableCompositeMessageConverter`
* `WebSocketInboundChannelAdapter`
* `WebSocketOutboundMessageHandler`
* `StompInboundChannelAdapterWebSocketIntegrationTests`

Has to be migrated to a new `JacksonJsonMessageConverter`.

anthologia:
Hi! Would it be okay if I work on this issue?

Is this related to https://github.com/spring-projects/spring-framework/issues/33798?

artembilan:
That's correct.
Feel free to open Pull Request.
However we need to be careful if this has to be done before: https://github.com/spring-projects/spring-framework/issues/34963.

artembilan:
Repurpose an issue to a broader subject.
Essentially, we have to provide Jackson 3 support and deprecate existing Jackson 2 API.
More info in: https://github.com/spring-projects/spring-framework/issues/33798

anthologia:
Hi @artembilan!

I see the scope got expanded. Are we looking at migrating Jackson 2 to 3 across all Spring Integration modules for this issue?

artembilan:
Thank you for looking into this, @anthologia !

Well, it is not about migration any more.
We have to add Jackson 3 dependency, implement Jackson 3 mirrored API where we Jackson 2 and deprecate those Jackson 2 API.
In between we can migrate to a new `JacksonJsonMessageConverter` where we see deprecation warnings for the `MappingJackson2MessageConverter`.

The most important part that Jackson 3 has to be managed as dependency to make everything working.

anthologia:
Thanks for clarifying! I'll fix the compile warning issues (in this milestone) first, then get into this.

artembilan:
>  I'll fix the compile warning issues

Thanks.
I hope you mean warnings related to this issue, not many others like JUnit 4 deprecations, or `this-espace` because of Java 24 upgrade.

anthologia:
Oh, I meant I'd handle the `this-escape` warnings in #10069 first, then come back to this Jackson issue. I've already started on the `this-escape` issue. I wasn't planning to mix them together..!

artembilan:
Good. 
Would be great if you would comment on that issue that you are working on it.
This way anyone interested would not do the same effort 😄

artembilan:
[...]

PR #10158, opened 2025-06-22
GH-10058: Introduce Jackson 3 in the Gradle build

anthologia:

First part of the Jackson 3 support implementation for issue. I've split this work for easier review and management:
- This PR (Work 1): Jackson 3 dependency management
- Work 2: Jackson 3 API implementation
- Work 3: Migration (e.g., from MappingJackson2MessageConverter to JacksonJsonMessageConverter)
- Work 4: Deprecation Jackson2 APIs

Does this approach work for you?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,6 +79,7 @@ ext {
     hsqldbVersion = '2.7.4'
     h2Version = '2.3.232'
     jacksonVersion = '2.19.1'
+    jackson3Version = '3.0.0-rc5'
     jaxbVersion = '4.0.5'
     jcifsVersion = '2.1.40'
     jeroMqVersion = '0.6.0'
@@ -165,6 +166,7 @@ allprojects {
 
         imports {
             mavenBom "com.fasterxml.jackson:jackson-bom:$jacksonVersion"
+            mavenBom "tools.jackson:jackson-bom:$jackson3Version"
             mavenBom "io.micrometer:micrometer-bom:$micrometerVersion"
             mavenBom "io.micrometer:micrometer-tracing-bom:$micrometerTracingVersion"
             mavenBom "io.projectreactor:reactor-bom:$reactorVersion"
@@ -453,6 +455,7 @@ project('spring-integration-amqp') {
         testImplementation 'org.springframework:spring-web'
         testImplementation 'org.testcontainers:rabbitmq'
         testImplementation 'com.fasterxml.jackson.core:jackson-databind'
+        testImplementation 'tools.jackson.core:jackson-databind'
     }
 }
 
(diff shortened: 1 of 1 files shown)


artembilan:
Thank you, @anthologia , for looking into this!
However I think the Work 3 has to be done second. Kinda migrate from deprecated upstream.
The part 2 might be divided because we have a lot of places using Jackson 2.
Another problem: we might need to be sure that upstream Spring projects come with respective new API, like Spring Kafka, Spring AMQP, Spring Data. However this could be a good trigger to check with those dependencies if they are planning already to do that migration to Jackson 3.

build.gradle
```diff
@@ -497,6 +500,11 @@ project('spring-integration-core') {
         api 'io.projectreactor:reactor-core'
         api 'io.micrometer:micrometer-observation'
 
+        optionalApi 'tools.jackson.core:jackson-databind'
+        optionalApi 'tools.jackson.datatype:jackson-datatype-joda'
```
artembilan, line 504: I’m not familiar with Jackson 3 and how it manages time API, but is Joda still a thing?
From where did you take this dependencies arrangement ?
Why there are no more for time, for example and parameters?

artembilan (changes requested):

anthologia:
Thanks for the feedback! Let me clarify my approach for Work 2 and Work 3:

Work 2 would implement the Jackson 3 counterparts like:
- `Jackson3JsonObjectMapper` 
- `Jackson3JsonMessageParser`
- `JacksonJsonUtils`, `JacksonPresent`
- ...

Work 3 would then use these new components in places like `ConfigurableCompositeMessageConverter`:
```java
if (JacksonPresent.isJackson3Present()) { // here
    ObjectMapper objectMapper = new Jackson3JsonObjectMapper().getObjectMapper(); // here
    JacksonJsonMessageConverter jacksonJsonMessageConverter = new JacksonJsonMessageConverter(objectMapper);
    // ...
}

**Comment by artembilan** [comment]
Thanks for clarification!
Sounds like a plan !
Let’s give it a shot!

File: build.gradle (hunk 1 above)
Comment (anthologia) [reply build.gradle:504]: In Jackson 3, "parameter names" and "datatypes" modules are now built-in.
JSTEP documentation: https://github.com/FasterXML/jackson-future-ideas/wiki/Jackson3-Changes#maven-modules-jars-java-packages

For Joda, I added it by referencing our `Jackson2JsonObjectMapper`. I tested it as shown below, and without adding `JodaModule`, Joda couldn't be supported.
```java
public static class JodaData {
    public String name;
    public org.joda.time.DateTime createdAt;

    public JodaData(String name, DateTime createdAt) {
       this.name = name;
       this.createdAt = createdAt;
    }
}

@Test
@DisplayName("[Jackson3] With JodaModule")
public void testJackson3JodaSerializationWithJodaModule() throws Exception {
    tools.jackson.databind.ObjectMapper mapper = JsonMapper.builder()
          .addModule(new JodaModule())
          .build();

    DateTime jodaDateTime = new DateTime();
    JodaData data = new JodaData("John", jodaDateTime);

    String json = mapper.writeValueAsString(data);
    mapper.readValue(json, JodaData.class);
    System.out.println("✅ Joda works with JodaModule in Jackson 3");
}

@Test
@DisplayName("[Jackson3] Without JodaModule")
public void testJackson3JodaSerializationWithoutJodaModule() throws Exception {
    tools.jackson.databind.ObjectMapper mapper = JsonMapper.builder()
          .build();

    DateTime jodaDateTime = new DateTime();
    JodaData data = new JodaData("John", jodaDateTime);

    Assert.assertThrows(tools.jackson.core.JacksonException.class, () -> mapper.writeValueAsString(data));
    System.out.println("❌ Joda doesn't work without JodaModule");
}
```

There are many changes in default configurations and other aspects. To save our time, I'll leave review comments with references to these changes in upcoming PRs!

PR #10160, opened 2025-06-23
GH-10058: Add Jackson 3 ObjectMapper and MessageParser

anthologia:

* Add `Jackson3JsonObjectMapper`, `Jackson3JsonMessageParser` to prepare for Jackson 2 to 3 migration (Work3).

## [Jackson 3 Exception Model Changes](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-4#replace-checked-jsonprocessingexception-with-unchecked-jacksonexception)
Jackson 3 changed its exception hierarchy from checked to unchecked.

Now, `JacksonException` extends `RuntimeException` not `IOException`.

## [Jackson 3 default setting changes](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-2#jackson-3-default-settings-behavior-changes)
Jackson 3.0 has different default settings compared to 2.x that may affect serialization behavior.

For example, these timestamp-related settings have changed:
- `WRITE_DATES_AS_TIMESTAMPS(false)` // was true in 2.x
- `WRITE_DURATIONS_AS_TIMESTAMPS(false)` // was true in 2.x
- `ONE_BASED_MONTHS(true)` // was false in 2.x

Should we maintain Jackson 2.x compatibility, or adopt Jackson 3 defaults?

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonMessageParser.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonMessageParser.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2025-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.json;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.jspecify.annotations.Nullable;
+import tools.jackson.core.JacksonException;
+import tools.jackson.core.JsonParser;
+import tools.jackson.core.JsonToken;
+import tools.jackson.core.ObjectReadContext;
+import tools.jackson.core.json.JsonFactory;
+
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * {@link JsonInboundMessageMapper.JsonMessageParser} implementation that parses JSON messages
+ * and builds a {@link Message} with the specified payload type from provided {@link JsonInboundMessageMapper}.
+ * Uses <a href="https://github.com/FasterXML">Jackson JSON Processor</a>.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ */
+public class Jackson3JsonMessageParser extends AbstractJacksonJsonMessageParser<JsonParser> {
+	private static final JsonFactory JSON_FACTORY = JsonFactory.builder().build();
+
+	public Jackson3JsonMessageParser() {
+		this(new Jackson3JsonObjectMapper());
+	}
+
+	public Jackson3JsonMessageParser(Jackson3JsonObjectMapper objectMapper) {
+		super(objectMapper);
+	}
+
+	@Override
+	protected JsonParser createJsonParser(String jsonMessage) throws JacksonException {
+		return JSON_FACTORY.createParser(ObjectReadContext.empty(), jsonMessage);
+	}
+
+	@Override
+	protected Message<?> parseWithHeaders(JsonParser parser, String jsonMessage,
+			@Nullable Map<String, Object> headersToAdd) throws JacksonException {
+
+		String error = AbstractJsonInboundMessageMapper.MESSAGE_FORMAT_ERROR + jsonMessage;
+		Assert.isTrue(JsonToken.START_OBJECT == parser.nextToken(), error);
+		Map<String, Object> headers = null;
+		Object payload = null;
+		while (JsonToken.END_OBJECT != parser.nextToken()) {
+			Assert.isTrue(JsonToken.PROPERTY_NAME == parser.currentToken(), error);
+			String currentName = parser.currentName();
+			boolean isHeadersToken = "headers".equals(currentName);
+			boolean isPayloadToken = "payload".equals(currentName);
+			Assert.isTrue(isHeadersToken || isPayloadToken, error);
+			if (isHeadersToken) {
+				Assert.isTrue(parser.nextToken() == JsonToken.START_OBJECT, error);
+				headers = readHeaders(parser, jsonMessage);
+			}
+			else {
+				parser.nextToken();
+				payload = readPayload(parser, jsonMessage);
+			}
+		}
+		Assert.notNull(headers, error);
+
+		return getMessageBuilderFactory()
+				.withPayload(payload)
+				.copyHeaders(headers)
+				.copyHeadersIfAbsent(headersToAdd)
+				.build();
+	}
+
+	private Map<String, Object> readHeaders(JsonParser parser, String jsonMessage) throws JacksonException {
+		Map<String, Object> headers = new LinkedHashMap<>();
+		while (JsonToken.END_OBJECT != parser.nextToken()) {
+			String headerName = parser.currentName();
+			parser.nextToken();
+			Object headerValue = readHeader(parser, headerName, jsonMessage);
+			headers.put(headerName, headerValue);
+		}
+		return headers;
+	}
+
+}
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
@@ -1,0 +1,210 @@
+/*
+ * Copyright 2025-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.support.json;
+
+import java.io.File;
+import java.io.InputStream;
+import java.io.Reader;
+import java.io.Writer;
+import java.lang.reflect.Type;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import tools.jackson.core.JacksonException;
+import tools.jackson.core.JsonParser;
+import tools.jackson.databind.JacksonModule;
+import tools.jackson.databind.JavaType;
+import tools.jackson.databind.JsonNode;
+import tools.jackson.databind.ObjectMapper;
+import tools.jackson.databind.json.JsonMapper;
+
+import org.springframework.integration.mapping.support.JsonHeaders;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+
+/**
+ * Jackson 3 JSON-processor (@link https://github.com/FasterXML)
+ * {@linkplain JsonObjectMapper} implementation.
+ * Delegates {@link #toJson} and {@link #fromJson}
+ * to the {@linkplain ObjectMapper}
+ * <p>
+ * It customizes Jackson's default properties with the following ones:
+ * <ul>
+ * <li>The well-known modules are registered through the classpath scan</li>
+ * </ul>
+ *
+ * See {@code tools.jackson.databind.json.JsonMapper.builder} for more information.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ *
+ */
+public class Jackson3JsonObjectMapper extends AbstractJacksonJsonObjectMapper<JsonNode, JsonParser, JavaType> {
+
+	private static final boolean JODA_MODULE_PRESENT =
+			ClassUtils.isPresent("tools.jackson.datatype.joda.JodaModule", null);
+
+	private static final boolean KOTLIN_MODULE_PRESENT =
+			ClassUtils.isPresent("kotlin.Unit", null) &&
+					ClassUtils.isPresent("tools.jackson.module.kotlin.KotlinModule", null);
+
+	private final ObjectMapper objectMapper;
+
+	public Jackson3JsonObjectMapper() {
+		List<JacksonModule> jacksonModules = collectWellKnownModulesIfAvailable();
+		this.objectMapper = JsonMapper.builder()
+				.addModules(jacksonModules)
+				.build();
+	}
+
+	public Jackson3JsonObjectMapper(ObjectMapper objectMapper) {
+		Assert.notNull(objectMapper, "objectMapper must not be null");
+		this.objectMapper = objectMapper;
+	}
+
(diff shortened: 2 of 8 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonMessageParser.java
```diff
@@ -0,0 +1,100 @@
...
+	@Override
+	protected Message<?> parseWithHeaders(JsonParser parser, String jsonMessage,
+			@Nullable Map<String, Object> headersToAdd) throws JacksonException {
+
+		String error = AbstractJsonInboundMessageMapper.MESSAGE_FORMAT_ERROR + jsonMessage;
+		Assert.isTrue(JsonToken.START_OBJECT == parser.nextToken(), error);
+		Map<String, Object> headers = null;
+		Object payload = null;
+		while (JsonToken.END_OBJECT != parser.nextToken()) {
+			Assert.isTrue(JsonToken.PROPERTY_NAME == parser.currentToken(), error);
```
anthologia: It changes from `FIELD_NAME`.

https://javadoc.io/static/tools.jackson.core/jackson-core/3.0.0-rc5/tools.jackson.core/tools/jackson/core/JsonToken.html#PROPERTY_NAME

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonMessageParser.java
```diff
@@ -0,0 +1,100 @@
...
+ * {@link JsonInboundMessageMapper.JsonMessageParser} implementation that parses JSON messages
+ * and builds a {@link Message} with the specified payload type from provided {@link JsonInboundMessageMapper}.
+ * Uses <a href="https://github.com/FasterXML">Jackson JSON Processor</a>.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ */
+public class Jackson3JsonMessageParser extends AbstractJacksonJsonMessageParser<JsonParser> {
+	private static final JsonFactory JSON_FACTORY = JsonFactory.builder().build();
```
artembilan: Every member in the class has to be surrounded with blank lines.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
```diff
@@ -0,0 +1,210 @@
...
+			ClassUtils.isPresent("kotlin.Unit", null) &&
+					ClassUtils.isPresent("tools.jackson.module.kotlin.KotlinModule", null);
+
+	private final ObjectMapper objectMapper;
+
+	public Jackson3JsonObjectMapper() {
+		List<JacksonModule> jacksonModules = collectWellKnownModulesIfAvailable();
+		this.objectMapper = JsonMapper.builder()
+				.addModules(jacksonModules)
+				.build();
```
artembilan: I see Spring Framework does this instead:
```
 JsonMapper.builder().findAndAddModules(Jackson3JsonObjectMapper.class.getClassLoader()).build();
```
Maybe we check if really those well-known modules are picked up properly by Jackson 3 native API?
I mean this our new class might be a bit lighter without those inner classes and `collectWellKnownModulesIfAvailable()` ?

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonMessageParser.java
```diff
@@ -0,0 +1,100 @@
...
+	public Jackson3JsonMessageParser() {
+		this(new Jackson3JsonObjectMapper());
+	}
+
+	public Jackson3JsonMessageParser(Jackson3JsonObjectMapper objectMapper) {
+		super(objectMapper);
+	}
+
+	@Override
+	protected JsonParser createJsonParser(String jsonMessage) throws JacksonException {
```
artembilan: You said this `JacksonException`  is an unchecked exception, so no need to mention it here I believe.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
```diff
@@ -0,0 +1,210 @@
...
+		Assert.notNull(objectMapper, "objectMapper must not be null");
+		this.objectMapper = objectMapper;
+	}
+
+	public ObjectMapper getObjectMapper() {
+		return this.objectMapper;
+	}
+
+	@Override
+	public String toJson(Object value) throws JacksonException {
```
artembilan: I think to support this abstraction contract we have to re-throw an `IOException` instead.

artembilan (changes requested):
Thank you!
This is great start!
Any chances to have some tests on the matter?
Especially I'm interested in those loaded by the Jackson 3 modules if we remove our custom `collectWellKnownModulesIfAvailable()`.

Re. defaults in v3 VS v2: I have raised concern with the team.
Will come back to ASAP.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
```diff
@@ -0,0 +1,210 @@
...
+ * </ul>
+ *
+ * See {@code tools.jackson.databind.json.JsonMapper.builder} for more information.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ *
+ */
+public class Jackson3JsonObjectMapper extends AbstractJacksonJsonObjectMapper<JsonNode, JsonParser, JavaType> {
```
artembilan: After team discussion, we prefer to have this as a `JacksonJsonObjectMapper` instead.
And try to keep the rest of API without a number as long as we have that `Jackson2` variants.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonMessageParser.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * {@link JsonInboundMessageMapper.JsonMessageParser} implementation that parses JSON messages
+ * and builds a {@link Message} with the specified payload type from provided {@link JsonInboundMessageMapper}.
+ * Uses <a href="https://github.com/FasterXML">Jackson JSON Processor</a>.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ */
+public class Jackson3JsonMessageParser extends AbstractJacksonJsonMessageParser<JsonParser> {
```
artembilan: Same here: just `JacksonJsonMessageParser`

artembilan (changes requested):
The team decision is to keep Jackson 3 defaults since this is a major version it is better to dictate a modern protocols. Luckily, we expose an API to inject a customized `ObjectMapper`.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
anthologia: By test results, it works as we expected.

But I think `JacksonJsonObjectMapper.class.getClassLoader()` would be the `ApplicationClassLoader`. This works because our Jackson dependencies only have these two modules, but if a user's application has other Jackson modules beyond Spring Integration... the situation might be different.

I'm a bit worried about this case. WDYT?

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
anthologia: I've all re-throw as IOException to comply with the abstraction contract !

But I saw Spring Framework plans to remove Jackson 2 in version 7.2. In that case, would we be able to change `JsonObjectMapper`'s `IOException` to `RuntimeException` at that time?

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
anthologia: Just curious - when you say team, do you mean the Spring team? Or is there a separate Spring Integration team internally?

artembilan: Yes, there is a separate number of people who work for this specific project, but of course we all here is Spring team and we collaborate with each other on daily basis.
The "team" in this case was a broader scope since we'd like to have consistency between all the Spring projects.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
artembilan: > our Jackson dependencies only have these two modules

Those are optional in our case, so end-user has to add even these two in their project explicitly.
I'm not sure what problem you see since in the end any custom Jackson module has to be added into final product classpath to make them available.

Do you see something off with our ClassLoader arrangement?

We probably can use `ClassUtils.getDefaultClassLoader()` instead.
I always had a problem with understand how all those class loaders work 😢

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
artembilan: Yeah, in the past we had one for [Boon](https://github.com/spring-projects/spring-integration/blob/5.2.x/spring-integration-core/src/main/java/org/springframework/integration/support/json/BoonJsonObjectMapper.java) in the past, but then it was removed due to target library low maintenance.
There is also a request for [GSON](https://github.com/spring-projects/spring-integration/issues/8498), but didn't go anywhere yet.

Well, I think we should really remove that `IOException` from the contract and re-throw our own generic `RuntimeException`.
But that's different story.
Let's come back to that when we done here in the separate GH issue!

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson3JsonObjectMapper.java
anthologia: Ah, I understood that in the existing `Jackson2JsonObjectMapper`, no matter how many Jackson dependencies a user's product has, when created with the default ctor, only `wellKnownModules()` would be added.

This time, since it's changed to find modules from the loader, I thought that in such situations, when created with the default ctor, an `ObjectMapper` with additional `JacksonModules` beyond `wellKnownModules()` could be created.

I might be misunderstanding something though - please let me know if I'm thinking about this incorrectly! 🥲

artembilan: > only wellKnownModules() would be added.

Correct. That's the current behavior.

> with additional `JacksonModules` beyond `wellKnownModules()` could be created.

That's correct, too. Those "wellKnownModules" are just placeholder in our code to simplify class loading.
With this `findAndAddModules()` we definitely are going to load everything what would fit Jackson expectations, essentially whatever is there on classpath.

Having this discussion with you I have to go and double check a new code from Jackson.
Might be the case that this `findAndAddModules()` is not OK by default and we should comeback to `wellKnownModules()` instead.
Again: team consultation :wink:

artembilan: OK. Looks like it uses a `ServiceLoader`:
```
    /**
     * Method for locating available methods, using JDK {@link ServiceLoader}
     * facility, along with module-provided SPI.
     *<p>
     * Note that method does not do any caching, so calls should be considered
     * potentially expensive.
     */
    public static List<JacksonModule> findModules(ClassLoader classLoader)
    {
        ArrayList<JacksonModule> modules = new ArrayList<>();
        ServiceLoader<JacksonModule> loader = secureGetServiceLoader(JacksonModule.class, classLoader);
        for (JacksonModule module : loader) {
            modules.add(module);
        }
        return modules;
    }
```
So, we should be good as long as the `ClassLoader` is correct.

anthologia: Thank you ! Actually I didn't know that.. Sorry for making you overthink this. 😞

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonObjectMapper.java
```diff
@@ -0,0 +1,194 @@
...
+			}
+		}
+		catch (JacksonException e) {
+			if (!(json instanceof String) && !(json instanceof byte[])) {
+				throw new IOException(e);
+			}
+			// Otherwise the input might not be valid JSON, fallback to TextNode with ObjectMapper.valueToTree()
+		}
+
+		return this.objectMapper.valueToTree(json);
```
artembilan: Why this one is not part of the `try..catch`?

spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
```diff
@@ -43,6 +44,9 @@ private JsonObjectMapperProvider() {
 		if (JacksonPresent.isJackson2Present()) {
 			return new Jackson2JsonObjectMapper();
 		}
+		else if (JacksonPresent.isJackson3Present()) {
```
artembilan, line 48: Let's flip this conditions and make Jackson 3 as default one!
If both versions are present, we prefer Jackson 3.

spring-integration-core/src/test/java/org/springframework/integration/support/json/JacksonJsonObjectMapperTest.java
```diff
@@ -0,0 +1,303 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ */
+class JacksonJsonObjectMapperTest {
```
artembilan: Our style is `*Tests` independently of number of methods in the class.

spring-integration-core/src/test/java/org/springframework/integration/support/json/JacksonJsonObjectMapperTest.java
```diff
@@ -0,0 +1,303 @@
...
+
+	private static final boolean JODA_MODULE_PRESENT =
+			ClassUtils.isPresent("tools.jackson.datatype.joda.JodaModule", null);
+
+	private static final boolean KOTLIN_MODULE_PRESENT =
```
artembilan: I don't think we need these conditions here since we deliberately have those dependencies in our project even if they are optional.
There are still there in classpath for our tests.
So, you may simplify this test environment a little bit.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapperProvider.java
anthologia: I understand making Jackson 3 the default, but it might be better to migrate related classes to Jackson 3 APIs first. Currently, `JsonPropertyAccessor` only supports Jackson 2, so SpEL doesn't work.

Should we track this as a follow-up after the migration is complete?

artembilan: I see. Let’s make it as a task on the original issue !
Thanks

anthologia: This migration turned out to be quite a bit more complex than expected. 😂 But I'm happy to contribute to such work! Thank you for always providing thorough and detailed reviews.

artembilan: Yeah… when I raised an issue, I hadn’t looked in details and filly not expected a Jackson 3 migration.
Sorry about that and thank you for a great contribution!

PR #10161, opened 2025-06-24
GH-10083: Implement Nullability in XML module

EddieChoCho:
Related to: #10083

* Add @org.jspecify.annotations.NullMarked to all the packages in the XML module
* Apply JSpecify null safety annotation improvements

### DIFF
--- a/spring-integration-xml/src/main/java/org/springframework/integration/xml/transformer/XsltPayloadTransformer.java
+++ b/spring-integration-xml/src/main/java/org/springframework/integration/xml/transformer/XsltPayloadTransformer.java
@@ -33,6 +33,7 @@ import javax.xml.transform.dom.DOMResult;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamSource;
 
+import org.jspecify.annotations.Nullable;
 import org.w3c.dom.Document;
 
 import org.springframework.beans.factory.BeanClassLoaderAware;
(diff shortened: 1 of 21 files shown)


spring-integration-xml/src/main/java/org/springframework/integration/xml/transformer/XsltPayloadTransformer.java
```diff
@@ -230,6 +233,7 @@ protected void onInit() {
 		this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
 		if (this.templates == null) {
```
artembilan, line 234: You marked it as `@SuppressWarnings("NullAway.Init")`, and I understand why.
But doesn't this `if` complain that the property is never `null`?
Thanks

artembilan (changes requested):
Just one doubt.
Everything else is great.

spring-integration-xml/src/main/java/org/springframework/integration/xml/transformer/XsltPayloadTransformer.java
artembilan: Yeah, looks like that work.
Just did something similar in Zip module: https://github.com/spring-projects/spring-integration/commit/e21d8433245152d52d305bf869ec187d5cb6b00e#diff-9f1c53dcaacbf6eb4a502eeb67a286da470aabb2b9401c05c50f5bc0788fa76dR64
And only what I see is a complain from my IntelliJ IDEA.
JSpecify compiler plugin is good with that.
So, merging your contribution as is.

Thank you!

PR #10164, opened 2025-06-24
Set default sendFailureChannelName to "errorChannel"

### LINKED ISSUE #10163: RabbitMQ Streams Client Errors Getting Lost
5.3.1.RELEASE

**Describe the bug**

**To Reproduce**

```java
@Bean
                .sync(true)
            .get();
}
```

```
...
```

**Expected behavior**

**Sample**

rrileyca:
Closes #10163.

the `RabbitStreamMessageHandler` previously did not set a `sendFailureChannelName`. If `sync` was set to false (as it is in the default constructor for `RabbitStreamMessageHandler`), then errors thrown by the Rabbit Stream client would be silently discarded.

This PR changes the default `sendFailureChannelName` to be `"errorChannel"`.

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpOutboundChannelAdapterSpec.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/dsl/AmqpOutboundChannelAdapterSpec.java
@@ -22,7 +22,7 @@ import org.springframework.amqp.core.AmqpTemplate;
  * Spec for an outbound AMQP channel adapter.
  *
  * @author Gary Russell
- * @author Artme Bilan
+ * @author Artem Bilan
  *
  * @since 5.3
  *
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/MethodInvokingMessageGroupProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/MethodInvokingMessageGroupProcessor.java
@@ -35,7 +35,7 @@ import org.springframework.util.Assert;
  * @author Mark Fisher
  * @author Dave Syer
  * @author Gary Russell
- * @author Artme Bilan
+ * @author Artem Bilan
  *
  * @since 2.0
  */
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/MethodInvokingReleaseStrategy.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/MethodInvokingReleaseStrategy.java
@@ -29,7 +29,7 @@ import org.springframework.integration.support.management.ManageableLifecycle;
  *
  * @author Marius Bogoevici
  * @author Dave Syer
- * @author Artme Bilan
+ * @author Artem Bilan
  */
 public class MethodInvokingReleaseStrategy implements ReleaseStrategy, BeanFactoryAware, ManageableLifecycle {
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractSubscribableChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractSubscribableChannel.java
@@ -33,7 +33,7 @@ import org.springframework.util.Assert;
  * @author Mark Fisher
  * @author Oleg Zhurakousky
  * @author Gary Russell
- * @author Artme Bilan
+ * @author Artem Bilan
  */
 public abstract class AbstractSubscribableChannel extends AbstractMessageChannel
 		implements SubscribableChannel, SubscribableChannelManagement {
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/xml/ServiceActivatorParser.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/xml/ServiceActivatorParser.java
@@ -28,7 +28,7 @@ import org.springframework.integration.config.ServiceActivatorFactoryBean;
  * @author Mark Fisher
  * @author Oleg Zhurakousky
  * @author Gary Russell
- * @author Artme Bilan
+ * @author Artem Bilan
  */
 public class ServiceActivatorParser extends AbstractDelegatingConsumerEndpointParser {
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/UniqueExpiryCallback.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/UniqueExpiryCallback.java
@@ -23,7 +23,7 @@ package org.springframework.integration.store;
  * class is present in the expire callbacks.
  *
  * @author Meherzad Lahewala
- * @author Artme Bilan
+ * @author Artem Bilan
  *
  * @since 5.0.10
  */
--- a/spring-integration-core/src/test/java/org/springframework/integration/aggregator/CorrelatingMessageHandlerTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/aggregator/CorrelatingMessageHandlerTests.java
@@ -48,7 +48,7 @@ import static org.mockito.Mockito.when;
 /**
  * @author Iwein Fuld
  * @author Dave Syer
- * @author Artme Bilan
+ * @author Artem Bilan
  */
 public class CorrelatingMessageHandlerTests implements TestApplicationContextAware {
 
--- a/spring-integration-core/src/test/java/org/springframework/integration/aggregator/TimeoutCountSequenceSizeReleaseStrategyTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/aggregator/TimeoutCountSequenceSizeReleaseStrategyTests.java
@@ -26,7 +26,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * @author Dave Syer
- * @author Artme Bilan
+ * @author Artem Bilan
  * @author Peter Uhlenbruck
  */
 public class TimeoutCountSequenceSizeReleaseStrategyTests {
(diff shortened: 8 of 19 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -77,6 +78,7 @@ public class RabbitStreamMessageHandler extends AbstractMessageHandler {
 	public RabbitStreamMessageHandler(RabbitStreamOperations streamOperations) {
 		Assert.notNull(streamOperations, "'streamOperations' cannot be null");
 		this.streamOperations = streamOperations;
+		this.sendFailureChannelName = IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME;
```
artembilan: And why don't you reset it back to `null`, when the mode is `sync`?
That's not what we have discussed with you.
The default `null` has still be there for `sync` mode when we re-throw exception from the `Future.get()`.

Any chances to have something covered with tests in the `RabbitStreamMessageHandlerTests`, please?

artembilan (changes requested):

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
rrileyca: Sorry for the delay. I will add the features you wish, with some comments, but I find it unexpected and odd that a `setSync()` nullifies the channel name. What if a user does this:

```java
IntegrationFlow.from(outboundChannel)
            .sendFailureChannel("errorChannel")
            .sync(true))
            .get();
```

This will remove their configured `sendFailureChannel` configuration.

artembilan: Correct . That’s why I suggested to fallback to `errorChannel` only in the `getFailureChannel()` and only if async mode. Don’t set this property default value !

rrileyca: Ignore latest commit please, I will fix as you describe

rrileyca:
Ready for review @artembilan. I've implemented as you suggest. Thank you for your patience, I was having some issues with my dev environment on my end.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
artembilan: You have not removed this default . That’s not what we agreed upon

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -174,11 +178,15 @@ protected MessageChannel getSendFailureChannel() {
...
-		}
-		return null;
+		this.sendFailureChannel = getChannelResolver()
+				.resolveDestination(getSendFailureChannelNameOrDefault());
+		return this.sendFailureChannel;
+	}
+
+	protected String getSendFailureChannelNameOrDefault() {
+		return this.sendFailureChannelName == null ?
+				IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME :
```
artembilan: This is not what we have discussed .
Where is a condition that fallback to this channel is only for async mode?

artembilan (changes requested):

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
```diff
@@ -117,4 +125,108 @@ void sendNative() throws InterruptedException {
...
+		}
+		env.streamCreator().stream("stream.stream").create();
+		RabbitStreamTemplate streamTemplate = new RabbitStreamTemplate(env, "stream.stream");
+
+		RabbitStreamMessageHandler handler = RabbitStream.outboundStreamAdapter(streamTemplate)
+				.getObject();
+		handler.setSendFailureChannelName("SomethingElse");
+
+		String failureChannelName = handler.getSendFailureChannelNameOrDefault();
+		Assertions.assertEquals("SomethingElse", failureChannelName);
```
artembilan: We prefer AssertJ API.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
```diff
@@ -117,4 +125,108 @@ void sendNative() throws InterruptedException {
...
+		ArgumentCaptor<ErrorMessage> captor = ArgumentCaptor.forClass(ErrorMessage.class);
+		Mockito.doReturn(true).when(errorChannel).send(captor.capture());
+		RabbitStreamMessageHandler handler = RabbitStream.outboundStreamAdapter(streamTemplate)
+				.sync(false)
+				.sendFailureChannel(errorChannel)
+				.getObject();
+
+		SecureRandom random = new SecureRandom();
+		// Default stream frame size is 1mb
+		byte[] messageTooBig = new byte[1100000];
```
artembilan: I’m not sure that we should exhaust the memory to test failure channel.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
```diff
@@ -117,4 +125,108 @@ void sendNative() throws InterruptedException {
...
+				.port(RabbitTestContainer.streamPort())
+				.build();
+		try {
+			env.deleteStream("stream.stream");
+		} catch (Exception e) {
+		}
+		env.streamCreator().stream("stream.stream").create();
+		RabbitStreamTemplate streamTemplate = new RabbitStreamTemplate(env, "stream.stream");
+
+		MessageChannel errorChannel = Mockito.mock(MessageChannel.class);
```
artembilan: This channel for testing could be a `QueueChannel` with very convenient `receive(timeout)` API.
So, anything you are mocking here and verify is fully redundant.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
```diff
@@ -117,4 +125,108 @@ void sendNative() throws InterruptedException {
...
+		Environment env = Environment.builder()
+				.lazyInitialization(true)
+				.port(RabbitTestContainer.streamPort())
+				.build();
+		try {
+			env.deleteStream("stream.stream");
+		} catch (Exception e) {
+		}
+		env.streamCreator().stream("stream.stream").create();
+		RabbitStreamTemplate streamTemplate = new RabbitStreamTemplate(env, "stream.stream");
```
artembilan: I think for this test we don’t need to go to the broker to get failure back.
Consider to mock this template to throw something when we interact with it from the handler.

artembilan (changes requested):

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
rrileyca: I have mocked the error response instead

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
rrileyca: Apologies this was a mistake

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
rrileyca: You are right, I made a mistake

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -118,7 +121,9 @@ public void setSendSuccessChannelName(String sendSuccessChannelName) {
 	}
 
 	/**
-	 * Set to true to wait for a confirmation.
+	 * Set to true to wait for a confirmation. Note that if sync is set to {@code true}, this sets both
```
artembilan: Please, revise this JavaDoc change.
It really does not reflect reality since we don't reset anything anywhere.
But async mode indeed has to be mentioned with its fallback to the `errorChannel` if not set.

rrileyca:
I made a total mess of this branch syncing the fork. I will work to fix it.

artembilan:
>  Files changed 358 

Something wrong happened with your branch.
Please, consider to create a new branch from a fresh `main`. Usually we use something like `GH-10163` to point to the issue we are trying to fix.
Right now your `main` is out-dated with what is there in the `upstream`.

artembilan (changes requested):

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
rrileyca: This was not intentional. I tried using IntelliJ's Git > GitHub > Sync Fork, and it made quite the mess. I will resolve this.

PR #10165, opened 2025-06-26
Add JSpecify Nullability checks to SI

cppwfs:
This is a refresh of the the PR https://github.com/spring-projects/spring-integration/pull/10097.  The 10097 was closed due its age and some restructuring. 

This PR picks up with requested changes from the last review (so you will see the original commits).  The last commit contains the changes that were requested in the last review of 10097 as well as a rebase to the latest main.


artembilan:
My Nullability fix has been pushed for AMQP module.
That touched a little bit some of the core classes, what has led us to fixes in the WS module as well.
Please, let me know if that rebases OK on your side.

Thank you!

spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
```diff
@@ -103,12 +110,14 @@ private Object getLock(Object correlationKey) {
...
 	@SuppressWarnings("unchecked")
 	@Override
+	@Nullable
 	public Message<Object> receive() {
 		for (Object key : this.correlationLocks.keySet()) {
 			Object lock = getLock(key);
 			synchronized (lock) {
 				MessageGroup group = this.store.getMessageGroup(key);
 				//group might be removed by another thread
+				Assert.notNull(this.releaseStrategy, "'releaseStrategy' must not be null");
```
artembilan: Not the task right now, but I suggestion to optimize this a bit and have this assert and the one in the `handleMessageInternal()` for  the `correlationStrategy` in the `onInit()` impl.
And have both pros as `	@SuppressWarnings("NullAway.Init")`.
Something to do on a spare time 😄

spring-integration-core/src/main/java/org/springframework/integration/aggregator/ExpressionEvaluatingCorrelationStrategy.java
```diff
@@ -52,7 +54,7 @@ public ExpressionEvaluatingCorrelationStrategy(Expression expression) {
 	}
 
 	public Object getCorrelationKey(Message<?> message) {
-		return this.processor.processMessage(message);
+		return Objects.requireNonNull(this.processor.processMessage(message));
```
artembilan: This contradicts with a `CorrelationStrategy` contract.
Why this method here cannot return null and be marked respectively?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/ExpressionEvaluatingMessageListProcessor.java
```diff
@@ -102,7 +105,9 @@ public void setExpectedType(Class<?> expectedType) {
 	 */
 	@Override
 	public Object process(Collection<? extends Message<?>> messages) {
-		return this.evaluateExpression(this.expression, messages, this.expectedType);
+		Object object = this.evaluateExpression(this.expression, messages, this.expectedType);
+		Assert.state(object != null, "The evaluation of the expression returned a null. Null result is not expected:" + this.expression);
```
artembilan: `is not expected for: "`
And pay attention to the space after colon.
Otherwise that expression is sticking to the colon 😄

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
```diff
@@ -209,6 +224,7 @@ public final void setPrimaryExpression(Expression expression) {
 	}
 
 	@Override
+	@SuppressWarnings("NullAway")
```
artembilan: Do we still need this here after all our hard work?

cppwfs:
PR has been rebased against main and looks good!

spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
cppwfs: I'll create an issue.

artembilan (changes requested):
This is great!
I'm sorry for causing you this headache.
I hope after all of these we would not have more big problems.

I'll give you a chance to answer to my last review and then I'll merge it tomorrow.

Thank you!

spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
cppwfs: I submitted an new commit to resolve this issue.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
```diff
@@ -86,9 +91,11 @@ public void setReleaseStrategy(ReleaseStrategy releaseStrategy) {
 		this.releaseStrategy = releaseStrategy;
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan, line 94: Instead of this those properties has to be marked as `@SuppressWarnings("NullAway.Init")`.
The point is that bean cannot be in the state where those properties are null.
Therefore the check in the `onInit()` and that's it.

artembilan (changes requested):
Thank you for an update!

spring-integration-core/src/main/java/org/springframework/integration/aggregator/CorrelatingMessageBarrier.java
cppwfs: The `SuppressWarnings` were applied to the  `handleMessageInternal` and `receive` methods  after moving the null checks to the onInit, because JSpecify isn't smart enough to see the optimization as a viable way to check for nullability, thus it flags it during compilation time..    i.e checking at `onInit` time and the null protection offered on the setters are valid methods to prevent a NPE, it does's see that.

artembilan: Correct. Then instead of `@Nullable` on those properties, use `@SuppressWarnings("NullAway.Init")`.

cppwfs: In this case "NullAway.Init" does not suppress the failure for these methods.

artembilan: Good.
Let me double check locally on merge!
I think I had something similar in other places.
Your PR is too big and already good enough to move on.
Thanks

artembilan:
Merged after rebase and squash as: https://github.com/spring-projects/spring-integration/commit/a27c74c6e68b6e824ab3e92570d370f8f782cf74.

Some clean up will be done in the separate commit.

Thanks again!

PR #10166, opened 2025-06-27
Apply Nullability to `spring-integration-jmx` module

mjd507:
related to: #10083

### DIFF
--- a/spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
+++ b/spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
@@ -62,7 +62,8 @@ public class NotificationPublishingMessageHandler extends AbstractMessageHandler
 
 	private @Nullable String defaultNotificationType;
 
-	private @Nullable OutboundMessageMapper<Notification> notificationMapper;
+	@SuppressWarnings("NullAway.Init")
+	private OutboundMessageMapper<Notification> notificationMapper;
 
 	/**
 	 * Construct an instance based on the provided object name.
@@ -94,7 +95,7 @@ public class NotificationPublishingMessageHandler extends AbstractMessageHandler
 	 * will be passed as the 'userData' of the Notification.
 	 * @param notificationMapper The notification mapper.
 	 */
-	public void setNotificationMapper(@Nullable OutboundMessageMapper<Notification> notificationMapper) {
+	public void setNotificationMapper(OutboundMessageMapper<Notification> notificationMapper) {
 		this.notificationMapper = notificationMapper;
 	}
 
@@ -140,7 +141,6 @@ public class NotificationPublishingMessageHandler extends AbstractMessageHandler
 
 	@Override
 	protected void handleMessageInternal(Message<?> message) {
-		Assert.state(this.notificationMapper != null, "'notificationMapper' must not be null");
 		this.delegate.publish(Objects.requireNonNull(this.notificationMapper.fromMessage(message)));
 	}
 
--- a/spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
+++ b/spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
@@ -17,10 +17,13 @@
 package org.springframework.integration.jmx;
 
 import java.util.Map;
+import java.util.Objects;
 
 import javax.management.MalformedObjectNameException;
 import javax.management.Notification;
 import javax.management.ObjectName;
+
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryUtils;
@@ -34,7 +37,6 @@ import org.springframework.jmx.export.annotation.ManagedResource;
 import org.springframework.jmx.export.notification.NotificationPublisher;
 import org.springframework.jmx.export.notification.NotificationPublisherAware;
 import org.springframework.jmx.support.ObjectNameManager;
-import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.util.Assert;
 
@@ -58,9 +60,9 @@ public class NotificationPublishingMessageHandler extends AbstractMessageHandler
 
 	private final ObjectName objectName;
 
-	private String defaultNotificationType;
+	private @Nullable String defaultNotificationType;
 
-	private OutboundMessageMapper<Notification> notificationMapper;
+	private @Nullable OutboundMessageMapper<Notification> notificationMapper;
 
 	/**
 	 * Construct an instance based on the provided object name.
@@ -138,7 +140,8 @@ public class NotificationPublishingMessageHandler extends AbstractMessageHandler
 
 	@Override
 	protected void handleMessageInternal(Message<?> message) {
-		this.delegate.publish(this.notificationMapper.fromMessage(message));
+		Assert.state(this.notificationMapper != null, "'notificationMapper' must not be null");
+		this.delegate.publish(Objects.requireNonNull(this.notificationMapper.fromMessage(message)));
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


mjd507:
> Task :spring-integration-mqtt:test
ClientManagerBackToBackTests > testSameV3ClientIdWorksForPubAndSub() FAILED
    java.lang.AssertionError at ClientManagerBackToBackTests.java:128

local build is green, please help me to re-trigger build process. 
thank you

spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
```diff
@@ -58,9 +60,9 @@ public class NotificationPublishingMessageHandler extends AbstractMessageHandler
 
 	private final ObjectName objectName;
 
-	private String defaultNotificationType;
+	private @Nullable String defaultNotificationType;
 
-	private OutboundMessageMapper<Notification> notificationMapper;
+	private @Nullable OutboundMessageMapper<Notification> notificationMapper;
```
artembilan: This has to be `@SuppressWarnings("NullAway.Init")` instead.
And no those changes in the `handleMessageInternal()`

spring-integration-jmx/src/main/java/org/springframework/integration/jmx/config/IntegrationMBeanExportConfiguration.java
```diff
@@ -63,22 +65,28 @@ public class IntegrationMBeanExportConfiguration implements ImportAware, Environ
...
+	@SuppressWarnings("NullAway.Init")
 	private AnnotationAttributes attributes;
 
+	@SuppressWarnings("NullAway.Init")
 	private BeanFactory beanFactory;
 
 	private BeanExpressionResolver resolver = new StandardBeanExpressionResolver();
 
-	private BeanExpressionContext expressionContext;
+	private @Nullable BeanExpressionContext expressionContext;
```
artembilan: This one is also `@SuppressWarnings("NullAway.Init")`

spring-integration-jmx/src/main/java/org/springframework/integration/monitor/IntegrationJmxAttributeSource.java
```diff
@@ -46,7 +48,7 @@
  */
 public class IntegrationJmxAttributeSource extends AnnotationJmxAttributeSource {
 
-	private StringValueResolver valueResolver;
+	private @Nullable StringValueResolver valueResolver;
```
artembilan, line 51: `@SuppressWarnings("NullAway.Init")`

artembilan (changes requested):

spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
mjd507: hmm.. we have below set method, allow it to be null.
```java
public void setNotificationMapper(@Nullable OutboundMessageMapper<Notification> notificationMapper) {
  this.notificationMapper = notificationMapper;
}
```

spring-integration-jmx/src/main/java/org/springframework/integration/jmx/config/MBeanExporterHelper.java
```diff
@@ -57,7 +58,7 @@ public Object postProcessBeforeInitialization(Object bean, String beanName) thro
 			@SuppressWarnings("unchecked")
 			Collection<String> autoCreateChannelCandidatesNames =
 					(Collection<String>) new DirectFieldAccessor(bean).getPropertyValue("channelNames");
-			this.siBeanNames.addAll(autoCreateChannelCandidatesNames);
+			this.siBeanNames.addAll(Objects.requireNonNull(autoCreateChannelCandidatesNames));
```
artembilan, line 61: FYI: https://github.com/spring-projects/spring-integration/issues/10167

spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
artembilan: And then see ` onInit()` details.
So, in the end when this bean is ready, the `notificationMapper` is not null.

mjd507: yes, but compile will fail,  `[NullAway] assigning @Nullable expression to @NonNull field`

artembilan: OK. Let's remove that `@Nullable` from the setter.
Its JavaDoc does not imply a null value for this property.
Just saying `if not provided`.

artembilan (changes requested):

spring-integration-jmx/src/main/java/org/springframework/integration/jmx/config/IntegrationMBeanExportConfiguration.java
mjd507: this one is initialized inside a if condition of the method `setBeanFactory(BeanFactory beanFactory)`, 
```java
	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
		this.beanFactory = beanFactory;
		if (beanFactory instanceof ConfigurableListableBeanFactory configurableListableBeanFactory) {
			...
			this.expressionContext = new BeanExpressionContext(configurableListableBeanFactory, null);
		}
	}
```
should we remove this `if (beanFactory instanceof ConfigurableListableBeanFactory) {`
use `Assert.isInstanceOf(ConfigurableListableBeanFactory.class, beanFactory);` instead.

artembilan: Looks like that.
But better against `ConfigurableBeanFactory`.
Then also that `getBeanExpressionResolver()` could be `null`. So, better to do `new StandardBeanExpressionResolver(configurableBeanFactory.getClassLoader());`

spring-integration-jmx/src/main/java/org/springframework/integration/monitor/IntegrationJmxAttributeSource.java
mjd507: this one I am a little confused

first the initialize part.
```java
if (beanFactory instanceof ConfigurableBeanFactory) {
	this.valueResolver = new EmbeddedValueResolver((ConfigurableBeanFactory) beanFactory);
}
```
then usage part.
```java
if (this.valueResolver != null && value instanceof String) {
	value = this.valueResolver.resolveStringValue((String) value);
}
```
seems it could be null, and `beanFactory` is not always an instance of `ConfigurableBeanFactory` ?

artembilan: You are right. 
Taking that back.
Sorry for a noise.

spring-integration-jmx/src/main/java/org/springframework/integration/jmx/NotificationPublishingMessageHandler.java
```diff
@@ -138,7 +141,7 @@ public final void onInit() {
 
 	@Override
 	protected void handleMessageInternal(Message<?> message) {
-		this.delegate.publish(this.notificationMapper.fromMessage(message));
+		this.delegate.publish(Objects.requireNonNull(this.notificationMapper.fromMessage(message)));
```
artembilan, line 144: I don't think we need this `Objects.requireNonNull()` yet.

artembilan (changes requested):

artembilan (approved):
I think that's enough.
I'm taking this locally for final review an merge.
Some clean might still be included.

Thank you!

artembilan:
Merged as: https://github.com/spring-projects/spring-integration/commit/0669e3aa83b99e044a2b85c325ece4dae87f4b40.

I'll push some clean up a bit latter in the separate commit.

Thank you again!

artembilan:
Here is follow up: https://github.com/spring-projects/spring-integration/commit/f0591f63b02119d7519834f09ead2feb0f301610

PR #10168, opened 2025-06-27
Apply jspecify nullify  to spring-integration-core aop/annotation packages

cppwfs:
…kages

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aop/SimplePublisherMetadataSource.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aop/SimplePublisherMetadataSource.java
@@ -36,11 +36,11 @@ import org.springframework.expression.Expression;
  */
 public class SimplePublisherMetadataSource implements PublisherMetadataSource {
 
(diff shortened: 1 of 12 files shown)


spring-integration-core/src/main/java/org/springframework/integration/aop/SimplePublisherMetadataSource.java
```diff
@@ -34,18 +36,18 @@
...
 public class SimplePublisherMetadataSource implements PublisherMetadataSource {
 
-	private volatile String channelName;
+	private @Nullable volatile String channelName;
 
-	private volatile Expression payloadExpression;
+	private @Nullable volatile Expression payloadExpression;
 
-	private volatile Map<String, Expression> headerExpressions;
+	private @Nullable volatile Map<String, Expression> headerExpressions;
```
artembilan: Remove `volatile`, please.
It does not make sense for setters.
Plus `@Nullable` has to be next to the type :wink

artembilan (changes requested):

PR #10181, opened 2025-07-01
Fix typo in `@author` tag

anthologia:
I noticed some incorrect author names while reviewing the code 😂

Fixed typo in `@author` tag: corrected misspelling of "Artem Bilan".


artembilan (approved):
Wow!
It was not necessary to do yourself, but I'm honored 😄 

Thank you very much !

PR #10182, opened 2025-07-01
Apply JSpecify Nullify to the Channel and AOT packages

cppwfs:
* Removed `@Nullify` annotations from classes outside the target packages, where the implemented interface or parent class did not mark the method as `Nullable`.
*  Note: One `"NullAway.Init"` was used for beanName.   
*  Removed all `NOSONAR` comments in touched classes for improved code cleanliness.
* Adjusted placement of existing `Nullable` annotations: moved them from above the method declarations to precede the method return type for consistency.

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/channel/PollableAmqpChannel.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/channel/PollableAmqpChannel.java
@@ -20,6 +20,7 @@ import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.jspecify.annotations.Nullable;
@@ -322,12 +323,12 @@ public class PollableAmqpChannel extends AbstractAmqpChannel
 	}
 
 	@Override
-	public @Nullable ChannelInterceptor removeInterceptor(int index) {
+	public ChannelInterceptor removeInterceptor(int index) {
 		ChannelInterceptor interceptor = super.removeInterceptor(index);
 		if (interceptor instanceof ExecutorChannelInterceptor) {
 			this.executorInterceptorsSize--;
 		}
-		return interceptor;
+		return Objects.requireNonNull(interceptor);
 	}
 
 	@Override
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/channel/PollableAmqpChannel.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/channel/PollableAmqpChannel.java
@@ -20,7 +20,6 @@ import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.jspecify.annotations.Nullable;
@@ -328,7 +327,7 @@ public class PollableAmqpChannel extends AbstractAmqpChannel
 		if (interceptor instanceof ExecutorChannelInterceptor) {
 			this.executorInterceptorsSize--;
 		}
-		return Objects.requireNonNull(interceptor);
+		return interceptor;
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
@@ -61,10 +61,6 @@ public abstract class AbstractExecutorChannel extends AbstractSubscribableChanne
 
 	protected @Nullable Executor executor;
 
-	/**
-	 * {@code @SuppressWarnings} was used the dispatcher is initialized in each of the implementations in the {@link AbstractExecutorChannel}.   And each implementation
-	 * utilizes a unique implementation of the {@link AbstractDispatcher}.
-	 */
 	@SuppressWarnings("NullAway.Init")
 	protected AbstractDispatcher dispatcher;
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
@@ -59,13 +59,19 @@ import org.springframework.util.CollectionUtils;
 public abstract class AbstractExecutorChannel extends AbstractSubscribableChannel
 		implements ExecutorChannelInterceptorAware {
 
-	protected Executor executor; // NOSONAR
+	protected @Nullable Executor executor;
 
-	protected AbstractDispatcher dispatcher; // NOSONAR
+	/**
+	 * {@code @SuppressWarnings} was used the dispatcher is initialized in each of the implementations in the {@link AbstractExecutorChannel}.   And each implementation
+	 * utilizes a unique implementation of the {@link AbstractDispatcher}.
+	 */
+	@SuppressWarnings("NullAway.Init")
+	protected AbstractDispatcher dispatcher;
 
-	protected Integer maxSubscribers; // NOSONAR
+	@Nullable
+	protected Integer maxSubscribers;
 
-	protected int executorInterceptorsSize; // NOSONAR
+	protected int executorInterceptorsSize;
 
 	public AbstractExecutorChannel(@Nullable Executor executor) {
 		this.executor = executor;
@@ -117,7 +123,6 @@ public abstract class AbstractExecutorChannel extends AbstractSubscribableChanne
 	}
 
 	@Override
-	@Nullable
 	public ChannelInterceptor removeInterceptor(int index) {
 		ChannelInterceptor interceptor = super.removeInterceptor(index);
 		if (interceptor instanceof ExecutorChannelInterceptor) {
@@ -167,7 +172,7 @@ public abstract class AbstractExecutorChannel extends AbstractSubscribableChanne
 				if (!CollectionUtils.isEmpty(interceptorStack)) {
 					triggerAfterMessageHandled(message, ex, interceptorStack);
 				}
-				if (ex instanceof MessagingException) { // NOSONAR
+				if (ex instanceof MessagingException) {
 					throw new MessagingExceptionWrapper(message, (MessagingException) ex);
 				}
 				String description = "Failed to handle " + message + " to " + this + " in " + messageHandler;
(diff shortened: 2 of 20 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/channel/PollableAmqpChannel.java
```diff
@@ -322,12 +323,12 @@ public boolean removeInterceptor(ChannelInterceptor interceptor) {
...
 
 	@Override
-	public @Nullable ChannelInterceptor removeInterceptor(int index) {
+	public ChannelInterceptor removeInterceptor(int index) {
 		ChannelInterceptor interceptor = super.removeInterceptor(index);
 		if (interceptor instanceof ExecutorChannelInterceptor) {
 			this.executorInterceptorsSize--;
 		}
-		return interceptor;
+		return Objects.requireNonNull(interceptor);
```
artembilan: Remove this `requireNonNull` and change contract of that `super.removeInterceptor()`.
I agree it cannot be null, so let's fix that `AbstractMessageChannel.ChannelInterceptorList.remove(index)`

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractExecutorChannel.java
```diff
@@ -59,13 +59,19 @@
 public abstract class AbstractExecutorChannel extends AbstractSubscribableChannel
 		implements ExecutorChannelInterceptorAware {
 
-	protected Executor executor; // NOSONAR
+	protected @Nullable Executor executor;
 
-	protected AbstractDispatcher dispatcher; // NOSONAR
+	/**
+	 * {@code @SuppressWarnings} was used the dispatcher is initialized in each of the implementations in the {@link AbstractExecutorChannel}.   And each implementation
```
artembilan: I prefer to follow rule: One sentence per line.
However, not more that 120 symbols per line is still there.
I'm not sure, though, why `was used` since it is still here and will be here in the future.
Just present time is OK.
Or we just can remove this comment at all.
I think we may come back to this `@SuppressWarnings` when we done with nullability for all the implementations of this class.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -235,9 +235,8 @@ public boolean removeInterceptor(ChannelInterceptor interceptor) {
 	}
 
 	@Override
-	@Nullable
 	public ChannelInterceptor removeInterceptor(int index) {
-		return this.interceptors.remove(index);
+		return Objects.requireNonNull(this.interceptors.remove(index));
```
artembilan: Why is this?
Just fix that `this.interceptors.remove()` to not produce null.
According to the `List.remove(int)` contract it does not return null.
So, that `if (removed != null) {` is wrong

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -394,14 +393,15 @@ private boolean sendWithObservation(Message<?> message, long timeout) {
 				messageToSendInternal =
 						new ErrorMessage(errorMessage.getPayload(),
 								messageToSend.getHeaders(),
-								errorMessage.getOriginalMessage());
+								Objects.requireNonNull(errorMessage.getOriginalMessage()));
```
artembilan: This is not correct.
That `original` indeed could be null.
We have to look into a ternary operator to chose this or other `ErrorMessage` ctor.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -394,14 +393,15 @@ private boolean sendWithObservation(Message<?> message, long timeout) {
...
 			}
 			return sendInternal(messageToSendInternal, timeout);
 		});
 		return Boolean.TRUE.equals(observe);
 	}
 
 	private boolean sendWithMetrics(Message<?> message, long timeout) {
+		Assert.state(this.metricsCaptor != null, "The metricsCaptor must not be null");
```
artembilan: Better to use `@SuppressWarnings("NullAway") // Dataflow analysis limitation` on this method instead.
The point is that we don't go to this method at all if `this.metricsCaptor` is `null`.
On the other side this is a critical path of the whole messaging, and having an extra call in this method for `Assert` might cause a performance impact.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -469,6 +469,7 @@ private TimerFacade sendTimer(boolean sent) {
 	}
 
 	private TimerFacade buildSendTimer(boolean success, String exception) {
+		Assert.state(this.metricsCaptor != null, "The metricsCaptor must not be null");
```
artembilan: DITTO

spring-integration-core/src/main/java/org/springframework/integration/channel/MessagePublishingErrorHandler.java
```diff
@@ -47,7 +49,8 @@ public class MessagePublishingErrorHandler extends ErrorMessagePublisher impleme
 
 	private static final ErrorMessageStrategy DEFAULT_ERROR_MESSAGE_STRATEGY = (ex, attrs) -> {
 		if (ex instanceof MessagingExceptionWrapper) {
-			return new ErrorMessage(ex.getCause(), ((MessagingExceptionWrapper) ex).getFailedMessage());
+			return new ErrorMessage(Objects.requireNonNull(ex.getCause()),
+					Objects.requireNonNull(((MessagingExceptionWrapper) ex).getFailedMessage()));
```
artembilan: Please use pattern expression for the if:
```
if (ex instanceof MessagingExceptionWrapper messagingExceptionWrapper) {
```
Then we won't need this cast.

I also think that we can extract both `cause` and `getFailedMessage()` to the local variables.
And add `@SuppressWarnings("NullAway") // Dataflow analysis limitation` since the `MessagingExceptionWrapper` contract does not bring us nulls.

spring-integration-core/src/main/java/org/springframework/integration/channel/NullChannel.java
```diff
@@ -163,6 +165,7 @@ public void onComplete() {
 
 	private TimerFacade sendTimer() {
 		if (this.successTimer == null) {
+			Assert.state(this.metricsCaptor != null, "metricsCaptor not must not be null");
```
artembilan: Same here with a `@SuppressWarnings("NullAway") // Dataflow analysis limitation`.
Same reason as in the `AbstractMessageChannel`.

spring-integration-core/src/main/java/org/springframework/integration/channel/NullChannel.java
```diff
@@ -199,6 +202,7 @@ private void incrementReceiveCounter() {
 	}
 
 	private CounterFacade buildReceiveCounter() {
+		Assert.state(this.metricsCaptor != null, "metricsCaptor not must not be null");
```
artembilan: DITTO

spring-integration-core/src/main/java/org/springframework/integration/channel/PartitionedChannel.java
```diff
@@ -65,7 +66,8 @@ public class PartitionedChannel extends AbstractExecutorChannel {
 	 * sent to this channel.
 	 */
 	public PartitionedChannel(int partitionCount) {
-		this(partitionCount, (message) -> message.getHeaders().get(IntegrationMessageHeaderAccessor.CORRELATION_ID));
+		this(partitionCount, (message) -> Objects.requireNonNull(
+				message.getHeaders().get(IntegrationMessageHeaderAccessor.CORRELATION_ID)));
```
artembilan: This is just nit-pick, but for me the code is more readable when we put everything after an arrow on a new line and have that as a single line.
I mean no breaks on the `requireNonNull(`

spring-integration-core/src/main/java/org/springframework/integration/core/ErrorMessagePublisher.java
```diff
@@ -70,7 +70,7 @@ public final void setErrorMessageStrategy(ErrorMessageStrategy errorMessageStrat
 		this.errorMessageStrategy = errorMessageStrategy;
 	}
 
-	public final void setChannel(MessageChannel channel) {
+	public final void setChannel(@Nullable MessageChannel channel) {
```
artembilan: And why is this, please?
Why we cannot keep contract as non-null, but preserve nullability internally?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -392,15 +390,17 @@ private boolean sendWithObservation(Message<?> message, long timeout) {
 			Message<?> messageToSendInternal = messageToSend;
 			if (message instanceof ErrorMessage errorMessage) {
 				messageToSendInternal =
-						new ErrorMessage(errorMessage.getPayload(),
+						(errorMessage.getOriginalMessage() != null) ? new ErrorMessage(errorMessage.getPayload(),
```
artembilan: Consider to extract `errorMessage.getOriginalMessage()` into a local variable for some optimization.

spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java
```diff
@@ -676,7 +677,6 @@ public boolean remove(ChannelInterceptor interceptor) {
 			}
 		}
 
-		@Nullable
 		public ChannelInterceptor remove(int index) {
 			ChannelInterceptor removed = this.interceptors.remove(index);
 			if (removed != null) {
```
artembilan: This one has to be removed as well.
That `this.interceptors.remove(index)` never returns null and may only throw an exception.
The confuse was really around this nullable variable.
If we remove this if altogether, all good.

spring-integration-core/src/main/java/org/springframework/integration/channel/PartitionedChannel.java
```diff
@@ -65,7 +66,7 @@ public class PartitionedChannel extends AbstractExecutorChannel {
 	 * sent to this channel.
 	 */
 	public PartitionedChannel(int partitionCount) {
-		this(partitionCount, (message) -> message.getHeaders().get(IntegrationMessageHeaderAccessor.CORRELATION_ID));
+		this(partitionCount, (message) -> Objects.requireNonNull(message.getHeaders().get(IntegrationMessageHeaderAccessor.CORRELATION_ID)));
```
artembilan: Sorry for my bad English, but I think I said:

>  put everything after an arrow on a new line

Remember `120` symbols per line rule still applies.

artembilan (changes requested):

PR #10183, opened 2025-07-01
GH-10163: Fix async mode dropping errors

### LINKED ISSUE #10163: RabbitMQ Streams Client Errors Getting Lost

rrileyca:
closes #10163 

A clean branch fixing several issues from my "corrupt" previous PR #10164

This PR has the `RabbitStreamMessageHandler` fallback to a `sendFailureChannelName` of `"errorChannel"` when a channel name is not set and when it is operating in `async` mode.

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
@@ -175,26 +175,18 @@ public class RabbitStreamMessageHandler extends AbstractMessageHandler {
 	}
 
 	protected @Nullable MessageChannel getSendFailureChannel() {
-		if (this.sendFailureChannel != null) {
-			return this.sendFailureChannel;
-		}
-		if (this.sync) {
-			if (this.sendFailureChannelName != null) {
-				this.sendFailureChannel = getChannelResolver()
-						.resolveDestination(sendFailureChannelName);
+		if (sendFailureChannel == null) {
+			if (sync && sendFailureChannelName != null) {
+				this.sendFailureChannel = getChannelResolver().resolveDestination(this.sendFailureChannelName);
+			}
+			else if (!sync) {
+				String sendFailureChannelNameToUse = sendFailureChannelName == null
+						? IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME
+						: sendFailureChannelName;
+				this.sendFailureChannel = getChannelResolver().resolveDestination(sendFailureChannelNameToUse);
 			}
 		}
-		else {
-			this.sendFailureChannel = getChannelResolver()
-					.resolveDestination(getSendFailureChannelNameOrDefault());
-		}
-		return this.sendFailureChannel;
-	}
-
-	protected String getSendFailureChannelNameOrDefault() {
-		return this.sendFailureChannelName == null ?
-				IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME :
-				sendFailureChannelName;
+		return sendFailureChannel;
 	}
 
 	protected @Nullable MessageChannel getSendSuccessChannel() {
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
@@ -29,6 +29,7 @@ import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.integration.amqp.support.AmqpHeaderMapper;
 import org.springframework.integration.amqp.support.DefaultAmqpHeaderMapper;
 import org.springframework.integration.amqp.support.MappingUtils;
+import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.core.MessagingTemplate;
 import org.springframework.integration.handler.AbstractMessageHandler;
 import org.springframework.messaging.Message;
@@ -176,11 +177,23 @@ public class RabbitStreamMessageHandler extends AbstractMessageHandler {
 		if (this.sendFailureChannel != null) {
 			return this.sendFailureChannel;
 		}
-		else if (this.sendFailureChannelName != null) {
-			this.sendFailureChannel = getChannelResolver().resolveDestination(this.sendFailureChannelName);
-			return this.sendFailureChannel;
+		if (this.sync) {
+			if (this.sendFailureChannelName != null) {
+				this.sendFailureChannel = getChannelResolver()
+						.resolveDestination(sendFailureChannelName);
+			}
 		}
-		return null;
+		else {
+			this.sendFailureChannel = getChannelResolver()
+					.resolveDestination(getSendFailureChannelNameOrDefault());
+		}
+		return this.sendFailureChannel;
+	}
+
+	protected String getSendFailureChannelNameOrDefault() {
+		return this.sendFailureChannelName == null ?
+				IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME :
+				sendFailureChannelName;
 	}
 
 	protected @Nullable MessageChannel getSendSuccessChannel() {
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
@@ -46,6 +46,7 @@ import org.springframework.util.Assert;
  *
  * @author Gary Russell
  * @author Chris Bono
+ * @author Ryan Riley
  * @since 6.0
  *
  */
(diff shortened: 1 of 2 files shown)


rrileyca:
@artembilan here is the clean branch. Sorry again for the confusion.

artembilan:
> A clean branch fixing several issues from my "corrupt" previous PR

Why don't you create a top-level branch as I suggested in our previous discussion?
Why make your life so hard trying to contribute from your `main`?
The rule of thumb is to never develop in `main`.
See more info in our Contribution Guidelines: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc

rrileyca:
I must admit I don't do much fork style PRs. I will do that in future. Thanks for the advice.

Do you want me to switch to another branch still?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
```diff
@@ -176,11 +177,23 @@ public RabbitStreamOperations getStreamOperations() {
...
+			this.sendFailureChannel = getChannelResolver()
+					.resolveDestination(getSendFailureChannelNameOrDefault());
+		}
+		return this.sendFailureChannel;
+	}
+
+	protected String getSendFailureChannelNameOrDefault() {
+		return this.sendFailureChannelName == null ?
+				IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME :
+				sendFailureChannelName;
```
artembilan: I think this should do that trick instead of your extra method calls:
```
	protected @Nullable MessageChannel getSendFailureChannel() {
		if (this.sendFailureChannel == null && (this.sendFailureChannelName != null || !this.sync)) {
			String sendFailureChannelNameToUse = this.sendFailureChannelName;
			if (sendFailureChannelNameToUse == null) {
				sendFailureChannelNameToUse = IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME;
			}
			this.sendFailureChannel = getChannelResolver().resolveDestination(sendFailureChannelNameToUse);
		}
		return this.sendFailureChannel;
	}
```

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
```diff
@@ -117,4 +128,79 @@ void sendNative() throws InterruptedException {
...
+						.build())
+				.build();
+		handler.handleMessage(testMessage);
+		StreamSendException streamException = new StreamSendException("Test Error Code", 99);
+		errorFuture.completeExceptionally(streamException);
+		ErrorMessage errorMessage = (ErrorMessage) errorChannel.receive(1000);
+		assertThat(errorMessage).isNotNull();
+		Throwable caughtException = errorMessage.getPayload();
+		assertThat(caughtException).isInstanceOf(StreamSendException.class);
+		assertThat(caughtException).isEqualTo(streamException);
```
artembilan: You can combine all of these into a single chained assert:
```
assertThat(errorMessage).extracting(Message::getPayload).isEqualTo(streamException);
```
I don't see a reason in extra `isInstanceOf()` since that `isEqualTo()` does cover this as well.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
```diff
@@ -117,4 +128,79 @@ void sendNative() throws InterruptedException {
...
+				.sync(true)
+				.sendFailureChannel(errorChannel)
+				.getObject();
+		SimpleCodec codec = new SimpleCodec();
+		org.springframework.messaging.Message<Message> testMessage = MessageBuilder.withPayload(codec.messageBuilder()
+						.addData(new byte[1])
+						.build())
+				.build();
+		assertThatThrownBy(() -> handler.handleMessage(testMessage))
+				.isInstanceOf(MessageHandlingException.class);
```
artembilan: We prefer to use opposite API: `assertThatExceptionOfType()`

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java
```diff
@@ -117,4 +128,79 @@ void sendNative() throws InterruptedException {
...
+
+	@Test
+	void defaultFailureChannel() {
+		Environment env = Mockito.mock(Environment.class);
+		RabbitStreamTemplate streamTemplate = new RabbitStreamTemplate(env, "stream.stream");
+
+		RabbitStreamMessageHandler handler = RabbitStream.outboundStreamAdapter(streamTemplate)
+				.getObject();
+
+		String failureChannelName = handler.getSendFailureChannelNameOrDefault();
```
artembilan: Exactly!
I don't want to have any `protected` methods just to satisfy tests.
You technically covered this code path with that `errorChanelAsync()`.
Which indeed is kind of an integration test, but it also includes a `failureChannel` path.
Therefore I don't see a reason in extra tests.

artembilan (changes requested):

artembilan:
> Do you want me to switch to another branch still?

It depends how far we still would go out of what we have agreed before 😉

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
rrileyca: I will slightly modify this, as it doesn't allow for sync mode with a configured `sendFailureChannelName` - unless that is your intention that in sync mode, a `sendFailureChannel` shall never be used.

artembilan: I'm not sure what you are talking about.
See this condition:
```
this.sendFailureChannelName != null || !this.sync
```
So, we go down in a sync mode and provided name.
Then we do `if (sendFailureChannelNameToUse == null) {`.
Which applies would not be true in sync mode and provided name.
What do I miss?

rrileyca:
> It depends how far we still would go out of what we have agreed before 😉

There was no intention to deviate on what we agreed on. This is my first time contributing to Spring, so I have a lot to learn. Thank you for your feedback and patience.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandler.java
rrileyca: You are right, I misread this :)

artembilan: Good. So, please, incorporate whatever I've suggestion.
So, far your code is not what we have discussed.

rrileyca: Done!

artembilan (changes requested):
Please, see that contributing doc: https://github.com/spring-projects/spring-integration/blob/main/CONTRIBUTING.adoc#provide-a-link-to-the-github-issue-in-the-associated-pull-request.

We recommend to have a link to the issue like `Fixes: ` exactly in the commit message.
Whatever you say on PR won't close issue automatically. 

Plus, the first commit of the PR becomes very own PR title and description.
Therefore, here is a best practice for Git commits: https://cbea.ms/git-commit/.

And you already know that it is not very convenient to raise a PR against `main`.
Usually we pull your branch locally for final review and merge.
If that is done from `main`, then it causes some headache on merge, not only for your local project copy.

artembilan:
Now, since we are very close to the merge, tell us, please, if you are going to contribute similar fix for the `KafkaProducerMessageHandler`.
We have that discussed in the issue and that one is scheduled for Kafka module fix as well.
I don't think we need dedicated tests in that module since the functionality is exactly the same you have done here.
I'm asking for the fix over there as well, since the issue is also scheduled for back-port to previous versions.

Thank you again!

artembilan (changes requested):
The build has failed:
```
 Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/RabbitStreamMessageHandlerTests.java:43:59: Using a static member import should be avoided - org.assertj.core.api.AssertionsForClassTypes.assertThatExceptionOfType. [AvoidStaticImport]
```

Please, run `./gradlew :spring-integration-amqp:check` locally before pushing.

rrileyca:
> tell us, please, if you are going to contribute similar fix for the KafkaProducerMessageHandler.

Yes, I can add that.

> And you already know that it is not very convenient to raise a PR against main.

Apologies, I did not know. Does this mean I should create a new PR from the new branch?

artembilan:
> Does this mean I should create a new PR from the new branch?

No, it's fine so far.
Just would be some inconvenience if I would do clean up myself for you and manual merge locally.
But right now I had to ask you to do all of that boring stuff around the code perfection 😉 .

And then, when we merge, it would be a problem for your local `main` to bring those your changes back to your copy.
Right now you would need to do reset force or just remove the branch and fetch from upstream.

rrileyca:
Well, I have to squash all the commits in order to make a proper first git commit, even if I keep this fork as is (with the incorrect commits on the `main` branch. 

Plus, I hope to make more contributions in the future, so I think it is a good idea to just clean up and make a new fork, with the proper feature branch and properly written first commit.

PR #10184, opened 2025-07-02
GH-10163: Fallback to default errorChannel in RabbitStreamMessageHandler

### LINKED ISSUE #10163: RabbitMQ Streams Client Errors Getting Lost

rrileyca:
Fixes: #10163

If RabbitStreamMessageHandler was used in async mode, error messages were dropped unless a sendFailureChannel or sendFailureChannelName were provided. This adds the "errorChannel" name as a fallback.


rrileyca:
@artembilan Third time is the charm. I hope I have cleaned up everything now up to Spring's standards!

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
```diff
@@ -543,7 +544,9 @@ void testConsumeAndProduceTransaction() throws Exception {
 		KafkaMessageDrivenChannelAdapter inbound = new KafkaMessageDrivenChannelAdapter<>(container);
 		DirectChannel channel = new DirectChannel();
 		inbound.setOutputChannel(channel);
+		DirectChannel sendFailureChannel = new DirectChannel();
```
artembilan: Why do you introduce his variable if it is out of verification part?

artembilan (changes requested):

artembilan:
The PR looks really good .
Pay attention how PR description has matched from your first commit message .
Plus now with that `Fixed:` it will close issue on merge.

> up to Spring's standards

it is not about standards, but reasoning and convenience for all of us. The code style is also for easy to read. So, yeah, we can call it “standard”, but that came after years of cooperation with the community and back & forth attempts . Therefore now we can even call it “the way” 😉

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
rrileyca: I did this because without it, the call to `getChannelResolver()` would throw an error because the `beanFactory` is null. Instead of creating a BeanFactory *and* a message channel, I thought it would be harmless to just create a channel and set it in order to circumvent the error.

rrileyca:
> Therefore now we can even call it “the way” 😉

Thank you for showing me the way then! This was fun. I hope to contribute more in the future - this time, with clean PRs ;)

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
artembilan: Oh, I see! So, we probably just can use `new NullChannel()` in that setter and no any extra local variables.

rrileyca: I had forgotten about the NullChannel. Good idea. Fixed.

artembilan:
There is one more problem , @rrileyca : 

> Signed-off-by: rrileyca <rriley.ca@gmail.com>

That has to be your legal name if you are OK with legal obligations according to DCO.
If you are not, then we cannot accept contributions from you.

I took a liberty to modify your final commit message on merge, but I hope you’ll fix you Git client to populate a proper name.

Thank you again !

rrileyca:
I had made a non-global config in the previous repo, which I since deleted and recreated. Sorry about that. My legal name is Ryan Riley, as per the `@author` annotations.

Thank you!

artembilan:
I think I did it right: https://github.com/spring-projects/spring-integration/commit/f4670e58be5fe238fb9441897b917613bb78c1eb

rrileyca:
> I think I did it right: [f4670e5](https://github.com/spring-projects/spring-integration/commit/f4670e58be5fe238fb9441897b917613bb78c1eb)

You did!

PR #10187, opened 2025-07-02
Add JSpecify nullify to channel/interceptor, code, and codec/kryo packages

cppwfs:
* Missed the channel/interceptor on the last PR.
* Removed CompositeCodec(Codec defaultCodec)  because null can never be pass to the Codec.    In the past it would have thrown a NPE.

Continued effort or issue: https://github.com/spring-projects/spring-integration/issues/10083

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -37,7 +37,7 @@ public class CompositeCodec implements Codec {
 
 	private final Codec defaultCodec;
 
-	private final @Nullable Map<Class<?>, Codec> delegates;
+	private final Map<Class<?>, Codec> delegates;
 
 	public CompositeCodec(@Nullable Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		Assert.notNull(defaultCodec, "'defaultCodec' cannot be null");
@@ -96,7 +96,7 @@ public class CompositeCodec implements Codec {
 	}
 
 	private @Nullable Codec findDelegate(Class<?> type) {
-		if (this.delegates == null) {
+		if (this.delegates.isEmpty()) {
 			return null;
 		}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -39,14 +39,17 @@ public class CompositeCodec implements Codec {
 
 	private final @Nullable Map<Class<?>, Codec> delegates;
 
-	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
+	public CompositeCodec(@Nullable Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		Assert.notNull(defaultCodec, "'defaultCodec' cannot be null");
 		this.defaultCodec = defaultCodec;
+		if (delegates == null) {
+			delegates = Map.of();
+		}
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
 
 	public CompositeCodec(Codec defaultCodec) {
-		this(Map.of(), defaultCodec);
+		this(null, defaultCodec);
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -45,6 +45,10 @@ public class CompositeCodec implements Codec {
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
 
+	public CompositeCodec(Codec defaultCodec) {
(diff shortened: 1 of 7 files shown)


spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -35,18 +37,14 @@ public class CompositeCodec implements Codec {
...
 	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		Assert.notNull(defaultCodec, "'defaultCodec' cannot be null");
 		this.defaultCodec = defaultCodec;
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
 
-	public CompositeCodec(Codec defaultCodec) {
-		this(null, defaultCodec);
-	}
-
```
mrpiggi, line 49: Isn't this about ctor argument `Map<Class<?>, Codec> delegates` which actually could be `null` (at least `this.delegates` is marked as `@Nullable`) and eventually would lead to always use `this.defaultCodec`?

artembilan: I agree.
Let's revert this ctor back and have another one with `@Nullable` for the `delegates` argument!

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
mrpiggi: Alternatively, one could just use
```
	public CompositeCodec(Codec defaultCodec) {
		this(Map.of(), defaultCodec);
	}
```
and drop `@Nullable` for `this.delegates` and just return `this.defaultCodec` from `findDelegate()` if nothing found in the map with
```
	private Codec findDelegate(Class<?> type) {
		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), false);
		return this.delegates.getOrDefault(clazz, this.defaultCodec);
	}
```

artembilan: Yeah... That feels more smooth in regard to that ctor contract.
I don't feel comfortable to give a `null` for the first arg.
So, having both ctors, but with an empty map to delegate would be nice.

cppwfs: I agree with the  suggestion :
```
	public CompositeCodec(Codec defaultCodec) {
		this(Map.of(), defaultCodec);
	}
``` 
Else the user will just get a NPE downstream.

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -35,7 +37,7 @@ public class CompositeCodec implements Codec {
 
 	private final Codec defaultCodec;
 
-	private final Map<Class<?>, Codec> delegates;
+	private final @Nullable Map<Class<?>, Codec> delegates;
```
artembilan: Then this should not be null

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -90,7 +92,7 @@ public <T> T decode(byte[] bytes, Class<T> type) throws IOException {
 		return decode(new ByteArrayInputStream(bytes), type);
 	}
 
-	private Codec findDelegate(Class<?> type) {
+	private @Nullable Codec findDelegate(Class<?> type) {
 		if (this.delegates == null) {
```
artembilan: And here we would use `isEmpty()` instead.

Everything else is good.

artembilan (changes requested):
Just one concern for the latest commit.

@mrpiggi ,

Do you have any other feedback, please?

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -37,9 +39,12 @@ public class CompositeCodec implements Codec {
 
 	private final Map<Class<?>, Codec> delegates;
 
-	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
+	public CompositeCodec(@Nullable Map<Class<?>, Codec> delegates, Codec defaultCodec) {
```
artembilan: Everything else is good in this class.
But why do you still keep this `@Nullable` here?
Other ctor provides `Map.of()`, and this one should not let to pass `null` at all.
Therefore question.
Thanks

artembilan (changes requested):

mrpiggi:
@artembilan So you obviously voted against my suggestion regarding `findDelegate(Class<?> type)` to never return `null` from this method but always fall back to `this.defaultCodec` and get rid of all `null` checks after calling. Just an oversight or is there any reason?

artembilan:
>  you obviously voted against my suggestion

No, I did not voted against you.
Sorry about that: was concentrated to make those nullables OK.

Feel free to raise a PR with whatever improvements you see!
Thank you for your help!

PR #10189, opened 2025-07-03
CompositeCodec constructors require at least one delegate

cppwfs:
* In the current implementation, users could pass in an empty list to the composite codec.
* This is an CompositeCodec and if we had no codecs, then why have the class.
* All `Nullable's` were removed from the delegate attribute.
* Null returns from the findDelegate have been removed.   The default delegate will be returned in the stead.
* Update the tests to require at least one codec in the delegates

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -41,7 +41,7 @@ public class CompositeCodec implements Codec {
 	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.defaultCodec = defaultCodec;
 		Assert.notEmpty(delegates, "delegates must not be empty");
-		this.delegates = new HashMap<Class<?>, Codec>(delegates);
+		this.delegates = new HashMap<>(delegates);
 	}
 
 	/**
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -49,7 +49,8 @@ public class CompositeCodec implements Codec {
 	 */
 	@Deprecated(since = "7.0", forRemoval = true)
 	public CompositeCodec(Codec defaultCodec) {
-		this(Map.of(), defaultCodec);
+		this.defaultCodec = defaultCodec;
+		this.delegates = Map.of();
 	}
 
 	@Override
@@ -79,11 +80,7 @@ public class CompositeCodec implements Codec {
 
 	private Codec findDelegate(Class<?> type) {
 		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), false);
-		Codec codec = this.delegates.get(clazz);
-		if (codec == null) {
-			codec = this.defaultCodec;
-		}
-		return codec;
+		return clazz == null ? this.defaultCodec : this.delegates.getOrDefault(clazz, this.defaultCodec);
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -23,8 +23,6 @@ import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.Map;
 
-import org.jspecify.annotations.Nullable;
-
 import org.springframework.integration.util.ClassUtils;
 import org.springframework.util.Assert;
 
@@ -41,9 +39,20 @@ public class CompositeCodec implements Codec {
 
 	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.defaultCodec = defaultCodec;
+		Assert.notEmpty(delegates, "delegates must not be empty");
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
 
+	/**
+	 * @since 7.0
+	 *
+	 * @deprecated This constructor is deprecated because the delegates map must contain at least one entry.
+	 * Creating a {@code CompositeCodec} without any delegates is not a valid use case, as there would be
+	 * no codecs to delegate to for encoding or decoding operations. Please use the constructor that accepts
+	 * a non-empty delegates map to ensure proper delegation behavior.
+	 *
+	 */
+	@Deprecated
 	public CompositeCodec(Codec defaultCodec) {
 		this(Map.of(), defaultCodec);
 	}
@@ -52,38 +61,20 @@ public class CompositeCodec implements Codec {
 	public void encode(Object object, OutputStream outputStream) throws IOException {
 		Assert.notNull(object, "cannot encode a null object");
 		Assert.notNull(outputStream, "'outputStream' cannot be null");
-		Codec codec = findDelegate(object.getClass());
-		if (codec != null) {
-			codec.encode(object, outputStream);
-		}
-		else {
-			this.defaultCodec.encode(object, outputStream);
-		}
+		findDelegate(object.getClass()).encode(object, outputStream);
 	}
 
 	@Override
 	public byte[] encode(Object object) throws IOException {
 		Assert.notNull(object, "cannot encode a null object");
-		Codec codec = findDelegate(object.getClass());
-		if (codec != null) {
-			return codec.encode(object);
-		}
-		else {
-			return this.defaultCodec.encode(object);
-		}
+		return findDelegate(object.getClass()).encode(object);
 	}
 
 	@Override
 	public <T> T decode(InputStream inputStream, Class<T> type) throws IOException {
 		Assert.notNull(inputStream, "'inputStream' cannot be null");
 		Assert.notNull(type, "'type' cannot be null");
-		Codec codec = findDelegate(type);
-		if (codec != null) {
-			return codec.decode(inputStream, type);
-		}
-		else {
-			return this.defaultCodec.decode(inputStream, type);
-		}
+		return findDelegate(type).decode(inputStream, type);
 	}
 
 	@Override
@@ -91,13 +82,14 @@ public class CompositeCodec implements Codec {
 		return decode(new ByteArrayInputStream(bytes), type);
 	}
 
-	private @Nullable Codec findDelegate(Class<?> type) {
-		if (this.delegates.isEmpty()) {
-			return null;
-		}
-
+	private Codec findDelegate(Class<?> type) {
+		Assert.state(!this.delegates.isEmpty(), "Delegates must not be empty");
 		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), false);
-		return this.delegates.get(clazz);
+		Codec codec = this.delegates.get(clazz);
+		if (codec == null) {
+			codec = this.defaultCodec;
+		}
+		return codec;
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -29,6 +29,7 @@ import org.springframework.util.Assert;
 /**
  * A Codec that can delegate to one out of many Codecs, each mapped to a class.
  * @author David Turanski
+ * @author Glenn Renfro
  * @since 4.2
  */
 public class CompositeCodec implements Codec {
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -44,15 +44,9 @@ public class CompositeCodec implements Codec {
 	}
 
 	/**
-	 * @since 7.0
-	 *
-	 * @deprecated This constructor is deprecated because the delegates map must contain at least one entry.
-	 * Creating a {@code CompositeCodec} without any delegates is not a valid use case, as there would be
-	 * no codecs to delegate to for encoding or decoding operations. Please use the constructor that accepts
-	 * a non-empty delegates map to ensure proper delegation behavior.
-	 *
+	 * @deprecated since 7.0 in favor of {@link #CompositeCodec(Map, Codec)}.
 	 */
-	@Deprecated
+	@Deprecated(since = "7.0", forRemoval = true)
 	public CompositeCodec(Codec defaultCodec) {
 		this(Map.of(), defaultCodec);
 	}
@@ -83,7 +77,6 @@ public class CompositeCodec implements Codec {
 	}
 
 	private Codec findDelegate(Class<?> type) {
-		Assert.state(!this.delegates.isEmpty(), "Delegates must not be empty");
 		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), false);
 		Codec codec = this.delegates.get(clazz);
 		if (codec == null) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -22,7 +22,6 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 
 import org.springframework.integration.util.ClassUtils;
 import org.springframework.util.Assert;
@@ -43,7 +42,6 @@ import org.springframework.util.Assert;
  * @author Glenn Renfro
  *
  * @since 4.2
- * @see ClassUtils#findClosestMatch(Class, Set, boolean)
  */
 public class CompositeCodec implements Codec {
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -22,6 +22,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 
 import org.springframework.integration.util.ClassUtils;
 import org.springframework.util.Assert;
@@ -31,9 +32,18 @@ import org.springframework.util.Assert;
  * delegating encoding and decoding operations to the appropriate type-specific codec.
  * This implementation associates object types with their appropriate codecs while providing a fallback default codec
  * for unregistered types.
+ * This class uses {@code ClassUtils.findClosestMatch} to select the appropriate codec for a given object type.
+ * When multiple codecs match an object type, {@code ClassUtils.findClosestMatch} offers the
+ * {@code failOnTie} option. If {@code failOnTie} is {@code false}, it will return any one of the matching codecs.
+ * If {@code failOnTie} is {@code true} and multiple codecs match, it will throw an {@code IllegalStateException}.
+ * {@link CompositeCodec} sets {@code failOnTie} to {@code true}, so if multiple codecs match, an
+ * {@code IllegalStateException} is thrown.
+ *
  * @author David Turanski
  * @author Glenn Renfro
+ *
  * @since 4.2
+ * @see ClassUtils#findClosestMatch(Class, Set, boolean)
  */
 public class CompositeCodec implements Codec {
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -27,7 +27,10 @@ import org.springframework.integration.util.ClassUtils;
 import org.springframework.util.Assert;
 
 /**
- * A Codec that can delegate to one out of many Codecs, each mapped to a class.
+ * An implementation of {@link Codec} that combines multiple codecs into a single codec,
+ * delegating encoding and decoding operations to the appropriate type-specific codec.
+ * This implementation associates object types with their appropriate codecs while providing a fallback default codec
+ * for unregistered types.
  * @author David Turanski
  * @author Glenn Renfro
  * @since 4.2
@@ -42,15 +45,6 @@ public class CompositeCodec implements Codec {
 		this.defaultCodec = defaultCodec;
 		Assert.notEmpty(delegates, "delegates must not be empty");
 		this.delegates = new HashMap<>(delegates);
-	}
-
-	/**
-	 * @deprecated since 7.0 in favor of {@link #CompositeCodec(Map, Codec)}.
-	 */
-	@Deprecated(since = "7.0", forRemoval = true)
-	public CompositeCodec(Codec defaultCodec) {
-		this.defaultCodec = defaultCodec;
-		this.delegates = Map.of();
 	}
 
 	@Override
@@ -79,7 +73,7 @@ public class CompositeCodec implements Codec {
 	}
 
 	private Codec findDelegate(Class<?> type) {
-		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), false);
+		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), true);
 		return clazz == null ? this.defaultCodec : this.delegates.getOrDefault(clazz, this.defaultCodec);
 	}
 
--- a/spring-integration-core/src/test/java/org/springframework/integration/codec/kryo/CompositeCodecTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/codec/kryo/CompositeCodecTests.java
@@ -29,6 +29,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * @author David Turanski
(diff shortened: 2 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -41,9 +39,20 @@ public class CompositeCodec implements Codec {
 
 	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.defaultCodec = defaultCodec;
+		Assert.notEmpty(delegates, "delegates must not be empty");
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
 
+	/**
+	 * @since 7.0
```
artembilan: This tag means when the API was introduced.
Which is not correct in our case.
And doesn't matter no when it was introduced anyway since we are deprecating it.

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -41,9 +39,20 @@ public class CompositeCodec implements Codec {
...
+	 *
```
artembilan: No blank lines in the method JavaDoc.
I think our `src/idea/spring-framework.xml` does the trick to disable JavaDoc reformatting if that is the case.

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -41,9 +39,20 @@ public class CompositeCodec implements Codec {
...
+	 * @deprecated This constructor is deprecated because the delegates map must contain at least one entry.
+	 * Creating a {@code CompositeCodec} without any delegates is not a valid use case, as there would be
+	 * no codecs to delegate to for encoding or decoding operations. Please use the constructor that accepts
+	 * a non-empty delegates map to ensure proper delegation behavior.
```
artembilan: JavaDocs must be neutral and come with an official language without any `please` and similar.
How about this:
```
* @deprecated since 7.0 in favor of {@link #CompositeCodec(Map, Codec)} with some delegates.
```
I don't think we should say more here  about "why?".

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -41,9 +39,20 @@ public class CompositeCodec implements Codec {
...
+	 *
+	 */
+	@Deprecated
```
artembilan: Should be like: `@Deprecated(since = "7.0", forRemoval = true)`

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -52,52 +61,35 @@ public CompositeCodec(Codec defaultCodec) {
...
 		return decode(new ByteArrayInputStream(bytes), type);
 	}
 
-	private @Nullable Codec findDelegate(Class<?> type) {
-		if (this.delegates.isEmpty()) {
-			return null;
-		}
-
+	private Codec findDelegate(Class<?> type) {
+		Assert.state(!this.delegates.isEmpty(), "Delegates must not be empty");
```
artembilan: This is redundant. The `delegates` is initialized and validated in the ctor.
The state of the instance cannot be changed afterward.

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/codec/kryo/CompositeCodecTests.java
```diff
@@ -38,7 +38,8 @@ public class CompositeCodecTests {
 
 	@BeforeEach
 	public void setup() {
-		Map<Class<?>, Codec> codecs = new HashMap<>();
+		PojoCodec pojoCodec = new PojoCodec();
+		Map<Class<?>, Codec> codecs = Map.of(pojoCodec.getClass(), pojoCodec);
```
artembilan: I don't think this is right to demonstrate in our test.
I mean it look like the point of the delegating codec is to chose one by the type of the object we would like to codec.
So, this `pojoCodec.getClass()` is somehow misleading even if we rely on the default one.

Can we have a reasonable test to see those delegates in action?
Looks like none of the existing tests do that: only fallback to the default one.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -41,9 +40,14 @@ public class CompositeCodec implements Codec {
 
 	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.defaultCodec = defaultCodec;
+		Assert.notEmpty(delegates, "delegates must not be empty");
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
```
artembilan: Remove, please, generics. Just plain diamond operator is enough.

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/codec/kryo/CompositeCodecTests.java
mrpiggi: Adding to this: there is currently no test for the (now deprecated) ctor without a delegates map. Otherwise, the newly introduced assertion check
```java
Assert.notEmpty(delegates, "delegates must not be empty");
```
in the main ctor would have failed.

@cppwfs thanks a lot for picking up my suggestion

artembilan: Correct.
We didn't have such a test and since we realized that ctor is wrong we are not going to add one.
But instead let's cover remaining one with more tests!

mrpiggi: Having no test for the deprecated ctor is fine but with this assertion in place, the deprecated ctor will always fail at runtime which is probably not desired

artembilan: I think that is fine for current major version.
You will see a deprecated warning on compilation anyway.
Our point is to lead users of this class away from the impression it can be used without delegates.
Therefore make it failing is intentional, since an original choice of this empty constructor feels awkward for the goal of the composite pattern.

Or do you have something else in mind?

Thank you for feedback thought!

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -41,9 +40,14 @@ public class CompositeCodec implements Codec {
 
 	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.defaultCodec = defaultCodec;
+		Assert.notEmpty(delegates, "delegates must not be empty");
```
mrpiggi, line 54: See my discussion with Artem, this will fail if the deprecated ctor is used

mrpiggi: Instead of calling the main ctor, one could go with

	@Deprecated(since = "7.0", forRemoval = true)
	public CompositeCodec(Codec defaultCodec) {
		this.delegates = Map.of();
	}

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -52,52 +56,34 @@ public CompositeCodec(Codec defaultCodec) {
...
 		return decode(new ByteArrayInputStream(bytes), type);
 	}
 
-	private @Nullable Codec findDelegate(Class<?> type) {
-		if (this.delegates.isEmpty()) {
-			return null;
-		}
-
+	private Codec findDelegate(Class<?> type) {
 		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), false);
```
artembilan: I know, this is not the part of the problem we are fixing here, but I’d like to question this part. What does it mean and if it is explained in the JavaDocs and covered by tests.
Plus why there is no null check for this method call? Does it fail with error and that is what that your new code in the test? How does fallback to the default work then ?
Would you mind looking into that, please ?

artembilan (changes requested):

mrpiggi (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
mrpiggi: I just saw, that I started a review on the former related PR but did not submit it so here my thoughts from the last review:

As `ClassUtils.findClosestMatch()` can return `null` (through `failOnTie=false`), one could argue that `Map::get` may not be called with `null` as key (which is no concern with `HashMap` but can be harmful if implementation is changed. So a safe approach would be
```java
	private Codec findDelegate(Class<?> type) {
		Class<?> clazz = ClassUtils.findClosestMatch(type, this.delegates.keySet(), false);
		return clazz == null ? this.defaultCodec : this.delegates.getOrDefault(clazz, this.defaultCodec);
	}
```

Regarding the question from Artem:

`ClassUtils.findClosestMatch()` traverses the class hierarchy by calling `ClassUtils.getTypeDifferenceWeight()` and gives back the class itself if found in `this.delegates.keySet()`. Otherwise, the "closest" super class or interface in the hierarchy found is returned.

As `failOnTie=false` is set, there will be something returned if a match in `this.delegates` was found. But as there could be multiple interfaces on the same "hierarchy level" in `this.delegates` and `ClassUtils.findClosestMatch()` is called with an unordered set, in this case the returned interface is non-deterministic. So one could argue that setting `failOnTie=true` would be the more defensive and reliable approach, though I have no suggestion, how one should then deal with the potentially thrown `IllegalStateException`

cppwfs: Argue with me please.    We could give that option to the user  and they make the decision.   But we still support our original spec from the `Codec` interface:

```java
/**
* Finds the closest matching codec from the given set of candidate codecs based on the
* type difference weight between the specified target type and each candidate.
* Then encodes an object to a byte array using the chosen codec.
* @param object the object to encode
* @param failOnTie whether to fail with an exception if two codecs have equal minimal weight
* @return the bytes
* @throws IOException if the operation fails
*/  
  
public byte[] encode(Object object, boolean failOnTie) throws IOException {
	Assert.notNull(object, "cannot encode a null object");
	return findDelegate(object.getClass(), findOnTie).encode(object);
}

/**
* Finds the closest matching codec from the given set of candidate codecs based on the
* type difference weight between the specified target type and each candidate.
* Then encodes an object to a byte array using the chosen codec.  If 2 codes have the same weight, then one of the two will be chosen.
* @param object the object to encode
* @return the bytes
* @throws IOException if the operation fails
*/  
  
public byte[] encode(Object object) throws IOException {
	Assert.notNull(object, "cannot encode a null object");
	return findDelegate(object.getClass(), false).encode(object);
}
```

Thoughts?

artembilan: Thank you for looking into this!

I had a chance today as well:
```
			else if (failOnTie && typeDiffWeight < Integer.MAX_VALUE && (typeDiffWeight == minTypeDiffWeight)) {
				throw new IllegalStateException("Unresolvable ambiguity while attempting to find closest match for [" +
						type.getName() + "]. Candidate types [" + closestMatch.getName() + "] and [" +
						candidate.getName() + "] have equal weight.");
			}
```
And I think this is the best approach we could live so far.
There is really no guarantee which one would be returned and different types may lead to different delegates.
And in different environments (or different application runs) we could end up with different delegates.
Therefore the best approach is to `failOnTie` and let the target project to fix configuration for this component to avoid ambiguity after seeing that `IllegalStateException`.

I want that to be mentioned in class JavaDocs.

Nit-pick: the method JavaDoc has to be imperative (like a command), not infinitive 😉

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -41,63 +40,47 @@ public class CompositeCodec implements Codec {
...
 	}
 
+	/**
+	 * @deprecated since 7.0 in favor of {@link #CompositeCodec(Map, Codec)}.
+	 */
+	@Deprecated(since = "7.0", forRemoval = true)
 	public CompositeCodec(Codec defaultCodec) {
-		this(Map.of(), defaultCodec);
+		this.defaultCodec = defaultCodec;
+		this.delegates = Map.of();
```
artembilan: I don't think this is what I said in the discussion.
The current (release) `CompositeCodec` for this ctor is to fail with NPE:
```
	public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
		Assert.notNull(defaultCodec, "'defaultCodec' cannot be null");
	}

	public CompositeCodec(Codec defaultCodec) {
		this(null, defaultCodec);
	}
```

That `HashMap(Map<? extends K, ? extends V> m)` does not let to pass a `null`.
So, with our new `Assert.notEmpty(Map)`, there is no too much difference with current behavior.
We still fail, but with extra deprecation warning.

Why you have decided to change behavior?

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpMessageMapperTests.java
```diff
@@ -67,7 +67,8 @@ public class TcpMessageMapperTests {
 
 	@BeforeEach
 	public void setup() {
-		Map<Class<?>, Codec> codecs = new HashMap<>();
+		MessageCodec messageCodec = new MessageCodec();
+		Map<Class<?>, Codec> codecs = Map.of(messageCodec.getClass(), messageCodec);
```
artembilan: This ha not be addressed.
I mean saying `messageCodec.getClass()` is wrong from the context of delegates logic.

I think this test could be fixed like this:
```
this.codec = new MessageCodec();
```

As we realized for the goal of this PR.
Why would one use a `Composite` if there is no composition?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
mrpiggi: That was my suggestion, but I had no look in the `6.5.x` branch so far. You are totally right, `new HashMap<>(null)` obviously cannot work. But then, this should be considered as a (10 year old) flawed/buggy implementation. And I still think, there are only two reasonable choices:

a) deprecate the ctor but provide a sane implementation
b) remove the flawed ctor completely without deprecation as this has never worked in the first place

I don't see, why an obvious bug should be kept although it was identified. I don't see any benefit from providing a flawed implementation and would rather prefer an error on compile time. But the call is up to you.

artembilan: Yes. This is great thinking.
I totally missed the fact that we can deprecate this ctor in previous versions and remove it altogether here in `7.0`.

I see your point about reasoning now.
So, let's make it with an empty map in `6.5.x` & `6.4.x` and mark as deprecated for removal over there.
Then we have a freedom here in `main` to remove it at all.

Can we go this route?

mrpiggi: > I totally missed the fact that we can deprecate this ctor in previous versions and remove it altogether here in `7.0`.

Great idea. I missed that as well ;)

> Can we go this route?

I totally agree!

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -23,14 +23,16 @@
...
 import org.springframework.util.Assert;
 
 /**
- * A Codec that can delegate to one out of many Codecs, each mapped to a class.
+ * An implementation of {@link Codec} that combines multiple codecs into a single codec,
+ * delegating encoding and decoding operations to the appropriate type-specific codec.
+ * This implementation associates object types with their appropriate codecs while providing a fallback default codec
+ * for unregistered types.
  * @author David Turanski
+ * @author Glenn Renfro
```
artembilan, line 42: I prefer to have different sections of the class JavaDocs separate by blank lines.
This way the `@author` section is not stuck to `@since`  or general text.
And that it is more respective for a credit 😉

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -23,14 +23,16 @@
...
-
 import org.springframework.integration.util.ClassUtils;
 import org.springframework.util.Assert;
 
 /**
- * A Codec that can delegate to one out of many Codecs, each mapped to a class.
+ * An implementation of {@link Codec} that combines multiple codecs into a single codec,
+ * delegating encoding and decoding operations to the appropriate type-specific codec.
+ * This implementation associates object types with their appropriate codecs while providing a fallback default codec
+ * for unregistered types.
```
artembilan, line 33: This text is not what we have discussed.
It is find to say that this class delegates to the provided codecs.
But what is more important is to say what `ClassUtils.findClosestMatch()` does for the logic in this class.
Any chances to have that mentioned, please?
Especially the part when it fails on tie.

We may also give a `@see` tag to this JavaDocs pointing to `ClassUtils.findClosestMatch()`.

spring-integration-core/src/test/java/org/springframework/integration/codec/kryo/CompositeCodecTests.java
```diff
@@ -17,39 +17,71 @@
...
+				AnotherClassWithNoDefaultConstructors.class);
+		assertThat(foo2).isEqualTo(foo);
 	}
 
 	@Test
-	public void testPojoSerialization() throws IOException {
+	void testWithUnRegisteredClass() throws IOException {
+		// Verify that the default encodes and decodes properly
+		Codec codec = onlyDefaultCodec();
 		SomeClassWithNoDefaultConstructors foo = new SomeClassWithNoDefaultConstructors("hello", 123);
```
artembilan: Just for the future consideration.
I don't like in our code works like `foo` and `bar`.
That is really not professional language to be used in the Open Source.
Even if they are stuck in industry for years, that does not mean that we have to follow bad habit to be neutral to the people who reads our code.
No need to do anything right now, just my opinion  😄

spring-integration-core/src/test/java/org/springframework/integration/codec/kryo/CompositeCodecTests.java
```diff
@@ -17,39 +17,71 @@
...
+	private static Codec getFullyQualifiedCodec() {
+		Map<Class<?>, Codec> codecs = Map.of(SomeClassWithNoDefaultConstructors.class, new PojoCodec(
+				new KryoClassListRegistrar(SomeClassWithNoDefaultConstructors.class)));
+		return new CompositeCodec(codecs, new PojoCodec(
+				new KryoClassListRegistrar(AnotherClassWithNoDefaultConstructors.class)));
+	}
+
+	private static Codec onlyDefaultCodec() {
+		PojoCodec pojoCodec = new PojoCodec();
+		Map<Class<?>, Codec> codecs = Map.of(pojoCodec.getClass(), pojoCodec);
```
artembilan: This is still misleading.
I would prefer something like `java.util.Date.class` to be more specific for the goal of the codecs mapping.
Otherwise this code gives us an impression that we are going to (de)serialize the `PojoCodec` itself 😕

spring-integration-core/src/test/java/org/springframework/integration/codec/kryo/CompositeCodecTests.java
```diff
@@ -81,4 +113,33 @@ public int hashCode() {
 
 	}
 
+	static class AnotherClassWithNoDefaultConstructors {
+
+		private String val1;
+
+		private int val2;
+
+		AnotherClassWithNoDefaultConstructors(String val1, int val2) {
```
artembilan: This class fully looks like Java record.
Can we convert it and have less code to support?

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpMessageMapperTests.java
```diff
@@ -67,11 +67,11 @@ public class TcpMessageMapperTests {
 
 	@BeforeEach
 	public void setup() {
-		Map<Class<?>, Codec> codecs = new HashMap<>();
-		this.codec = new CompositeCodec(codecs, new MessageCodec());
+		this.codec = new MessageCodec();
 	}
 
 	@Test
+	@SuppressWarnings("NullAway")
```
artembilan: Please, remove all these `@SuppressWarnings("NullAway")` from tests.
It it does not warn into logs when we build the project, then it should not be there.
The goal of the test is to ensure the prod code works as expected.
Having complex tests and unnecessary code around makes it hard to maintain the project.

artembilan (changes requested):
And since we are here, please, consider to mention this class in the `codec.adoc`.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -22,16 +22,28 @@
 import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.Map;
-
-import org.jspecify.annotations.Nullable;
+import java.util.Set;
```
artembilan: I’m not sure how it works right now, but in the past it failed for us when is used only in JavaDocs. Or use FQCN in the link for the method, or mention that method without argument types.

artembilan (changes requested):
And add some paragraph with a sample code into `codec.adoc`, please.

src/reference/antora/modules/ROOT/pages/codec.adoc
```diff
@@ -53,6 +54,69 @@ The framework provides several custom serializers:
 The first can be used with the `PojoCodec` by initializing it with the `FileKryoRegistrar`.
 The second and third are used with the `MessageCodec`, which is initialized with the `MessageKryoRegistrar`.
 
+[[composite-codec]]
+=== CompositeCodec
```
artembilan, line 58: Blank line after title.

src/reference/antora/modules/ROOT/pages/codec.adoc
```diff
@@ -53,6 +54,69 @@ The framework provides several custom serializers:
...
+The CompositeCodec is a codec that combines multiple codecs into a single codec, delegating encoding and decoding operations to the appropriate type-specific codec.
```
artembilan: Wrap `CompositeCodec` into a code snippet.

src/reference/antora/modules/ROOT/pages/codec.adoc
```diff
@@ -53,6 +54,69 @@ The framework provides several custom serializers:
...
+This implementation associates object types with their appropriate codecs while providing a fallback default codec for unregistered types.
+
+An example implementation can be seen below:
+```java
+void encodeDecodeSample() {
+        Codec codec = getFullyQualifiedCodec();
```
artembilan, line 66: Not visible here, but we need to be sure that no tab indents in docs.

src/reference/antora/modules/ROOT/pages/codec.adoc
```diff
@@ -53,6 +54,69 @@ The framework provides several custom serializers:
...
+            new KryoClassListRegistrar(Animal.class)));
+}
+
+// Records that will be encoded and decoded in this sample
+record Dog(String name, int age, String tag) {}
+record Cat(String name, int age, int lives) {}
+record Animal(String name, int age){}
+```
+
+In some cases a single type of object may return multiple codecs. In these cases an `IllegalStateException` is thrown.
```
artembilan: One sentence per line: https://asciidoctor.org/docs/asciidoc-recommended-practices/

src/reference/antora/modules/ROOT/pages/codec.adoc
```diff
@@ -53,6 +54,69 @@ The framework provides several custom serializers:
...
+
+NOTE: This class uses `ClassUtils.findClosestMatch` to select the appropriate codec for a given object type.
+When multiple codecs match an object type, `ClassUtils.findClosestMatch` offers the `failOnTie` option.
+If `failOnTie` is `false`, it will return any one of the matching codecs.
+If `failOnTie` is `true` and multiple codecs match, it will throw an `IllegalStateException`.
+CompositeCodec` sets `failOnTie` to `true`, so if multiple codecs match, an
+`IllegalStateException` is thrown.
```
artembilan: DITTO: One sentence per line

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/codec.adoc
cppwfs: No tabs were found in code segment.

src/reference/antora/modules/ROOT/pages/codec.adoc
cppwfs: Blank line applied

artembilan (approved):
LGTM.
But would be great to have https://github.com/spring-projects/spring-integration/pull/10207 reviewed and merged.
So, we could rebase this one before merge to working upstream.

Thanks

PR #10193, opened 2025-07-06
GH-10058: Add Jackson 3 (de)serializer support

anthologia:
* Add Jackson3-compatible messaging (de)serializers
* Add Jackson 3 `messagingAwareMapper()` support
* Maintain existing Jackson 2 functionality

I thought about splitting this into separate PRs (`(de)serializers` + `messagingAwareMapper()` support), but figured it'd be easier to review as one. It's a bit large though - apologies for the extra work! 😅

Now, I have a few questions..
### Package Structure
With many Jackson 2 / Jackson 3 classes in the same package, how about reorganizing Jackson 2 into sub-package?
```
org.springframework.integration.support.json/
├── jackson2/
```

### Copyright and Authors
When refactoring the classes, I moved existing impls and created new ones, following the team's previous decision:

> After team discussion, we prefer to have this as a `JacksonJsonObjectMapper` instead.
> And try to keep the rest of API without a number as long as we have that `Jackson2` variants.
> 
> *Originally posted by @artembilan in https://github.com/spring-projects/spring-integration/pull/10160#discussion_r2166619792*

E.g., :
- `*JacksonDeserializer` → `*Jackson2Deserializer` (Jackson 2 support)
- Updated `*JacksonDeserializer` for Jackson 3 support

I've attempted to update `@author` and copyright year appropriately, but please review if they're handled correctly.

### Docs
Currently, Many docs (e.g., redis.adoc) references the default class names:
> Starting with version 4.3.10, the Framework provides Jackson serializer and deserializer implementations for Message instances and MessageHeaders instances — MessageJacksonDeserializer and MessageHeadersJacksonSerializer, respectively.

With this PR, we now have version-specific implementations:
- Jackson 2: `MessageHeadersJackson2Serializer`
- Jackson 3: `MessageHeadersJacksonSerializer`

Should we update docs to reflect current impl(Jackson 2) or keep it for 7.0 release since Jackson 3 will be the default?


artembilan (changes requested):
>  *JacksonDeserializer → *Jackson2Deserializer

thank you for the effort !
I’ll take a look tommorow in details, but I think this is a wrong direction.

we develop here a library for target project to rely on the API. We cannot break anything just because. It could be possible in target product where no one depend on your public API. With frameworks and libraries we don’t have such a liberty.
Therefore renaming existing classes, or moving them to different packages is wrong direction for us. 
my apologies that my comment about naming was not so clear. Let me try again!
We only deprecate existing classes. We name new classes without a number, as much as possible. No need in a new package since deprecated classes will be removed in next version.
Again: we need to keep in mind that our code is used to compile other projects: breaking an API so dramatically is not acceptable.
Please, revise your solution.

anthologia:
Thank you for taking the time to review this on the weekend !

I apologize for not fully considering that I'm contributing to a widely-used library. At my job, I work on middleware primarily focusing on backward compatibility for end-user functionality, but I hadn't deeply thought about the compatibility of public APIs that other projects depend on. I'll be much more careful to consider API dependencies and backward compatibility in my contributions.

For easier review, I'll revert `*Jackson2Deserializer` back to `*JacksonDeserializer` and add `*Jackson3Deserializer` instead. I'd appreciate any further guidance when you have a chance !

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonUtils.java
```diff
@@ -146,9 +146,9 @@ private static final class AllowListTypeResolverBuilder extends ObjectMapper.Def
 
 		@Override
 		protected TypeIdResolver idResolver(MapperConfig<?> config,
-				JavaType baseType,
-				PolymorphicTypeValidator subtypeValidator,
-				Collection<NamedType> subtypes, boolean forSer, boolean forDeser) {
+											JavaType baseType,
```
anthologia: I'll fix it(indentation) when applying the review feedback.

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonUtils.java
```diff
@@ -168,7 +168,7 @@ private static final class AllowListTypeIdResolver implements TypeIdResolver {
 
 		private final TypeIdResolver delegate;
 
-		private final Set<String> trustedPackages = new LinkedHashSet<>(DEFAULT_TRUSTED_PACKAGES);
+		private final Set<String> trustedPackages = new LinkedHashSet<>(JacksonJsonUtils.DEFAULT_TRUSTED_PACKAGES);
```
anthologia: DITTO

spring-integration-core/src/main/java/org/springframework/integration/support/json/AdviceMessageJackson3Deserializer.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2017-present the original author or authors.
```
anthologia: I'll fix these years when applying the review feedback 😂

spring-integration-core/src/main/java/org/springframework/integration/support/json/AdviceMessageJackson3Deserializer.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.messaging.Message;
+
+/**
+ * The {@link MessageJackson3Deserializer} implementation for the {@link AdviceMessage}.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ */
+public class AdviceMessageJackson3Deserializer extends MessageJackson3Deserializer<AdviceMessage<?>> {
```
artembilan: As we discussed before, let's try to have new classes without number!
How about to name them like `MessageJsonDeserializer`, `AdviceMessageJsonDeserializer ` etc.?

spring-integration-core/src/main/java/org/springframework/integration/support/json/AdviceMessageJackson3Deserializer.java
```diff
@@ -0,0 +1,48 @@
...
+
+	@SuppressWarnings("unchecked")
+	public AdviceMessageJackson3Deserializer() {
+		super((Class<AdviceMessage<?>>) (Class<?>) AdviceMessage.class);
+	}
+
+	@Override
+	protected AdviceMessage<?> buildMessage(MutableMessageHeaders headers, Object payload, JsonNode root,
+			DeserializationContext ctxt) throws JacksonException {
```
artembilan: Please, put a blank line after multi-line method header.

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonUtils.java
artembilan: Well, next time, please, consider to fix them before asking for review.
I see that this class will be fully reverted and disappear in the next commit.
Therefore I've just spent some time reviewing for nothing 😄

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonMessagingUtils.java
```diff
@@ -0,0 +1,233 @@
...
+	 * and throws an IllegalStateException if the class being looked up is not trusted,
+	 * does not provide an explicit mixin mappings.
+	 *
+	 * @author Jooyoung Pyoung
+	 */
+	private static final class AllowListTypeIdResolver implements TypeIdResolver {
+
+		private final TypeIdResolver delegate;
+
+		private final Set<String> trustedPackages = new LinkedHashSet<>(JacksonJsonUtils.DEFAULT_TRUSTED_PACKAGES);
```
artembilan: Jackson 2 based API will be deprecated.
And I believe that this `JacksonJsonUtils` as well.
Therefore we have to copy that constant into this new class.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/json/MessageJackson3Deserializer.java
```diff
@@ -0,0 +1,87 @@
...
+	public Object deserializeWithType(JsonParser jp, DeserializationContext ctxt, TypeDeserializer td)
+			throws JacksonException {
+
+		return super.deserialize(jp, ctxt);
+	}
+
+	@Override
+	public T convert(JsonNode root, DeserializationContext ctxt) throws JacksonException {
+		Map<String, Object> headers = this.mapper.readValue(root.get("headers").traverse(ctxt),
+				TypeFactory.createDefaultInstance().constructMapType(HashMap.class, String.class, Object.class));
```
artembilan: Use `this.mapper.getTypeFactory()` instead.

spring-integration-core/src/test/java/org/springframework/integration/support/json/JacksonMessagingUtilsTests.java
```diff
@@ -0,0 +1,145 @@
...
+	private ObjectMapper mapper;
+
+	@BeforeEach
+	void setUp() {
+		mapper = JacksonMessagingUtils.messagingAwareMapper();
+	}
+
+	@Test
+	void shouldAutoDiscoverAllWellKnownModules() {
+		Set<String> collectedModuleNames = getModuleNames(collectWellKnownModulesIfAvailable());
```
artembilan: I don't think we need that `collectWellKnownModulesIfAvailable()` dedicated method, neither those constants.
The method name does not reflect reality because those modules are really always available.
We don't need those modules as constants since they are used only in this test.
Therefore my ask is to move everything into this test method scope.

artembilan (changes requested):
Thank you for the effort!
Please, keep in mind that Jackson 2 will be deprecated.

I even wonder if we should do that right now, so then it will become more clear what to implement and where to migrate in tests. I don't say to do that in your current PR, but just an idea for the next step (if any).

Thank you!

anthologia:
Thank you for the guidance!

It seems like we're getting closer to the point where Jackson 2 deprecation makes sense. I think we maybe need to add `JsonPropertyAccessor` support first.

As you mentioned previously, we'll likely be blocked by Jackson 3 migration in other projects. I'm not entirely certain, but from what I've checked, it appears to be Spring Kafka's `DefaultKafkaHeaderMapper`.

I see you're also a Spring Kafka maintainer. If needed, I'd be happy to help with migration in other projects too. Feel free to ping me on the relevant issues if my contribution would be helpful ! 😁

artembilan:
Sure thing, @anthologia !

Here is Spring Kafka part: https://github.com/spring-projects/spring-kafka/issues/3930
And if you are interested this is Spring AMQP part: https://github.com/spring-projects/spring-amqp/pull/3117

anthologia:
I'm not entirely familiar with these projects, but I'd be happy to give both a shot based on what I've learned here!

What's the expected release timeline for AMQP?

artembilan:
> What's the expected release timeline for AMQP?

For AMQP that was a PR from me to review.
If you want.
Nothing to do over there.
We have milestone in a couple weeks.
But all the plans for Jackson 3 migration are due on this November.
So, we have a plenty of time to clean up as maximum as we could.

Thank you for your help with all of this!

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonMessagingUtils.java
```diff
@@ -0,0 +1,249 @@
...
+			SimpleModule simpleModule = new SimpleModule()
+					.addSerializer(new MessageHeadersJsonSerializer())
+					.addSerializer(new MimeTypeJsonSerializer())
+					.addDeserializer(GenericMessage.class, genericMessageDeserializer)
+					.addDeserializer(ErrorMessage.class, errorMessageDeserializer)
+					.addDeserializer(AdviceMessage.class, adviceMessageDeserializer)
+					.addDeserializer(MutableMessage.class, mutableMessageDeserializer);
+
+			ObjectMapper mapper = JsonMapper.builder()
+					.findAndAddModules(JacksonJsonObjectMapper.class.getClassLoader())
```
artembilan: This is not correct.
Unnecessary classes tangle.
Use `JacksonMessagingUtils.class.getClassLoader()` instead.

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonMessagingUtils.java
```diff
@@ -0,0 +1,249 @@
...
+		private final String[] trustedPackages;
+
+		AllowListTypeResolverBuilder(String... trustedPackages) {
+			super(
+					BasicPolymorphicTypeValidator.builder().allowIfSubType(Object.class).build(),
+					DefaultTyping.NON_FINAL, JsonTypeInfo.As.PROPERTY
+			);
+
+			this.trustedPackages =
+					trustedPackages != null ? Arrays.copyOf(trustedPackages, trustedPackages.length) : null;
```
artembilan, line 139: Looks like nullability has to be added then!
`String @Nullable ...`
and:
`private final String @Nullable [] trustedPackages;`

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonMessagingUtils.java
```diff
@@ -0,0 +1,249 @@
...
+	 *
+	 * @author Jooyoung Pyoung
+	 */
+	private static final class AllowListTypeIdResolver implements TypeIdResolver {
+
+		private final TypeIdResolver delegate;
+
+		private final Set<String> trustedPackages = new LinkedHashSet<>(DEFAULT_TRUSTED_PACKAGES);
+
+		AllowListTypeIdResolver(TypeIdResolver delegate, String... trustedPackages) {
```
artembilan: DITTO

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonMessagingUtils.java
```diff
@@ -0,0 +1,249 @@
...
+import org.springframework.messaging.support.GenericMessage;
+
+/**
+ * Utility for creating Jackson {@link ObjectMapper} instance for Spring messaging.
+ *
+ * <p>Provides custom serializers/deserializers for Spring messaging types
+ * and validates deserialization against trusted package patterns.
+ *
+ * @author Jooyoung Pyoung
+ * @since 7.0
```
artembilan, line 55: Blank line between `@author` section and `@since`.
Just to make it more cleaner for a credit! 😉

artembilan (changes requested):
Very good!
Just final nit-pick review.
Thank you!

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonMessagingUtils.java
anthologia: Based on the [JSpecify](https://jspecify.dev/docs/spec/#recognized-type-use), it looks like your suggestion (`String @Nullable ...`) is the correct: 
> `void method(@Nullable String @NonNull ... strings)`

But this triggers a checkstyle violation:
`'...' is preceded with whitespace. [NoWhiteSpaceBefore]`

Should we fix the checkstyle?

Spring Framework seems to follow the standard:
https://github.com/spring-projects/spring-framework/blob/v7.0.0-M6/spring-core/src/main/java/org/springframework/core/ResolvableType.java
`ResolvableType forClassWithGenerics(Class<?> clazz, @Nullable ResolvableType @Nullable ... generics)`

artembilan: > Should we fix the checkstyle?

Sounds about right.
See respective fix in Spring AMQP: https://github.com/spring-projects/spring-amqp/commit/0173186675f0567099c337d3366c2c9ca7f9ef45#diff-a1c2846ba2cb772df68702bddd12e6c5fa59f8f80bb879715093ddb13bea4c4cR176

PR #10194, opened 2025-07-07
Apply Nullability to JMS module

mjd507:
related to: #10083

### DIFF
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.jms;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.micrometer.observation.ObservationRegistry;
@@ -114,7 +115,8 @@ public class ChannelPublishingJmsMessageListener
 
 	private JmsHeaderMapper headerMapper = new DefaultJmsHeaderMapper();
 
-	private @Nullable BeanFactory beanFactory;
+	@SuppressWarnings("NullAway.Init")
+	private BeanFactory beanFactory;
 
 	private MessageBuilderFactory messageBuilderFactory = new DefaultMessageBuilderFactory();
 
@@ -175,8 +177,8 @@ public class ChannelPublishingJmsMessageListener
 	}
 
 	@Override
-	public @Nullable String getComponentName() {
-		return this.gatewayDelegate.getComponentName();
+	public String getComponentName() {
+		return Objects.requireNonNull(this.gatewayDelegate.getComponentName());
 	}
 
 	@Override
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
@@ -17,7 +17,6 @@
 package org.springframework.integration.jms;
 
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.micrometer.observation.ObservationRegistry;
@@ -177,8 +176,8 @@ public class ChannelPublishingJmsMessageListener
 	}
 
 	@Override
-	public String getComponentName() {
-		return Objects.requireNonNull(this.gatewayDelegate.getComponentName());
+	public @Nullable String getComponentName() {
+		return this.gatewayDelegate.getComponentName();
 	}
 
 	@Override
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
@@ -26,6 +26,7 @@ import jakarta.jms.InvalidDestinationException;
 import jakarta.jms.JMSException;
 import jakarta.jms.MessageProducer;
 import jakarta.jms.Session;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
@@ -56,7 +57,6 @@ import org.springframework.jms.support.converter.MessageConverter;
 import org.springframework.jms.support.converter.SimpleMessageConverter;
 import org.springframework.jms.support.destination.DestinationResolver;
 import org.springframework.jms.support.destination.DynamicDestinationResolver;
-import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessagingException;
@@ -96,9 +96,9 @@ public class ChannelPublishingJmsMessageListener
 
 	private boolean extractReplyPayload = true;
 
-	private Object defaultReplyDestination;
+	private @Nullable Object defaultReplyDestination;
 
-	private String correlationKey;
+	private @Nullable String correlationKey;
 
 	private long replyTimeToLive = jakarta.jms.Message.DEFAULT_TIME_TO_LIVE;
 
@@ -110,14 +110,15 @@ public class ChannelPublishingJmsMessageListener
 
 	private DestinationResolver destinationResolver = new DynamicDestinationResolver();
 
-	private Expression replyToExpression;
+	private @Nullable Expression replyToExpression;
 
 	private JmsHeaderMapper headerMapper = new DefaultJmsHeaderMapper();
 
-	private BeanFactory beanFactory;
+	private @Nullable BeanFactory beanFactory;
 
 	private MessageBuilderFactory messageBuilderFactory = new DefaultMessageBuilderFactory();
 
+	@SuppressWarnings("NullAway.Init")
 	private StandardEvaluationContext evaluationContext;
 
 	/**
@@ -174,7 +175,7 @@ public class ChannelPublishingJmsMessageListener
 	}
 
 	@Override
-	public String getComponentName() {
+	public @Nullable String getComponentName() {
 		return this.gatewayDelegate.getComponentName();
 	}
 
@@ -563,7 +564,7 @@ public class ChannelPublishingJmsMessageListener
 	 * @see #setDefaultReplyTopicName
 	 * @see #setDestinationResolver
 	 */
-	private Destination resolveDefaultReplyDestination(Session session) throws JMSException {
+	private @Nullable Destination resolveDefaultReplyDestination(Session session) throws JMSException {
 		if (this.defaultReplyDestination instanceof Destination destination) {
 			return destination;
 		}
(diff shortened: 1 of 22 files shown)


spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
```diff
@@ -110,14 +110,15 @@ public class ChannelPublishingJmsMessageListener
 
 	private DestinationResolver destinationResolver = new DynamicDestinationResolver();
 
-	private Expression replyToExpression;
+	private @Nullable Expression replyToExpression;
 
 	private JmsHeaderMapper headerMapper = new DefaultJmsHeaderMapper();
 
-	private BeanFactory beanFactory;
+	private @Nullable BeanFactory beanFactory;
```
mjd507: I think (but not very sure), this class might be used a pure Java Object, without calling the BeanFactoryAware callback.  so make `BeanFactory`  nullable here.

because I saw in `JmsMessageDrivenEndpoint#onInit`, there it directly calls this class `this.listener.afterPropertiesSet();`  without setting a beanFactory instance.

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsInboundGateway.java
```diff
@@ -156,7 +157,8 @@ public void setBeanFactory(BeanFactory beanFactory) {
 
 	@Override
 	protected void onInit() {
-		this.endpoint.setComponentName(getComponentName());
+		String componentName = getComponentName();
+		this.endpoint.setComponentName(componentName == null ? "jms-inbound-gateway" : componentName);
```
mjd507: not sure here, will `getComponentName()` from gateway always not null ?

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsMessageDrivenEndpoint.java
```diff
@@ -218,7 +219,8 @@ protected void onInit() {
 				this.listenerContainer.setSessionAcknowledgeMode(acknowledgeMode);
 			}
 		}
-		this.listener.setComponentName(getComponentName());
+		String componentName = getComponentName();
+		this.listener.setComponentName(componentName == null ? "jms-message-listener" : componentName);
```
mjd507: not sure as well, will `getComponentName()` from endpoint always not null ?

should we set parameter nullable in method `IntegrationObjectSupport.setComponentName` in `core` module

spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
artembilan: This is not correct assumption.
The class is public and it is really a bean-like, Spring-aware component.
So, this has to be `@SuppressWarnings("NullAway.Init")`.
And that `JmsMessageDrivenEndpoint#onInit` has to propagate its `BF` if it is mission.

spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
```diff
@@ -174,7 +175,7 @@ public void setShouldTrack(boolean shouldTrack) {
 	}
 
 	@Override
-	public String getComponentName() {
+	public @Nullable String getComponentName() {
```
artembilan, line 179: This must not be null.

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsInboundGateway.java
artembilan: It has to be.

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsMessageDrivenEndpoint.java
artembilan: I don't think so.
If we haven't set anything, or the component is not a bean, then yes, `getComponentName()` may be null.
But I don't think we should allow to pass `null` into `setComponentName()`

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsOutboundGateway.java
```diff
@@ -807,9 +809,10 @@ private AbstractIntegrationMessageBuilder<?> buildReply(jakarta.jms.Message jmsR
 		}
 	}
 
-	private Object sendAndReceiveWithContainer(Message<?> requestMessage) throws JMSException {
+	private @Nullable Object sendAndReceiveWithContainer(Message<?> requestMessage) throws JMSException {
 		Connection connection = createConnection(); // NOSONAR - closed in ConnectionFactoryUtils.
 		Session session = null;
+		Assert.notNull(this.replyContainer, "'replyContainer' must not be null");
```
artembilan: We don't need to do this here.
We have respective `if..else` in the `handleRequestMessage()`.
This method has to be marked like:
```
@SuppressWarnings("NullAway") // Dataflow analysis limitation
```

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsOutboundGateway.java
```diff
@@ -1071,14 +1074,15 @@ private jakarta.jms.Message retryableReceiveReply(Session session, Destination r
...
-	private Object doSendAndReceiveAsync(Destination reqDestination, jakarta.jms.Message jmsRequest, Session session,
+	private @Nullable Object doSendAndReceiveAsync(Destination reqDestination, jakarta.jms.Message jmsRequest, Session session,
 			int priority) throws JMSException {
 
 		String correlation = null;
 		MessageProducer messageProducer = null;
 		try {
 			messageProducer = session.createProducer(reqDestination);
 			correlation = this.gatewayCorrelation + "_" + this.correlationId.incrementAndGet();
+			Assert.notNull(this.correlationKey, "'correlationKey' must not be null");
```
artembilan: There is logic like this:
```
			if (isAsync() && this.correlationKey == null) {
				logger.warn("'async=true' requires a correlationKey; ignored");
				setAsync(false);
			}
		}
		else {
			if (isAsync()) {
				logger.warn("'async=true' is ignored when a reply container is not being used");
				setAsync(false);
			}
```
So, when we reach this method, `this.correlationKey` is really not null.
Therefore similar `@SuppressWarnings` for the method as before.

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsOutboundGateway.java
```diff
@@ -1109,7 +1113,7 @@ private Object doSendAndReceiveAsync(Destination reqDestination, jakarta.jms.Mes
 				return future;
 			}
 			else {
-				return obtainReplyFromContainer(correlation, replyQueue);
+				return obtainReplyFromContainer(correlation, Objects.requireNonNull(replyQueue));
```
artembilan: Same here: we don't need this assert since when we reach this code block, the `replyQueue` is not null.
The mentioned `@SuppressWarnings` on this method will cover us.

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsOutboundGateway.java
```diff
@@ -1500,12 +1504,12 @@ public void run() {
 			try {
 				if (System.currentTimeMillis() - JmsOutboundGateway.this.lastSend >
 						JmsOutboundGateway.this.idleReplyContainerTimeout
-						&& JmsOutboundGateway.this.replies.size() == 0 &&
-						JmsOutboundGateway.this.replyContainer.isRunning()) {
+						&& JmsOutboundGateway.this.replies.isEmpty() &&
+						Objects.requireNonNull(JmsOutboundGateway.this.replyContainer).isRunning()) {
```
artembilan: Looks like better to propagate `this.replyContainer` as this class ctor argument from that `handleRequestMessage()`.

spring-integration-jms/src/main/java/org/springframework/integration/jms/JmsOutboundGateway.java
```diff
@@ -1500,12 +1504,12 @@ public void run() {
...
 
 					logger.debug(() -> getComponentName() + ": Stopping idle reply container.");
 					JmsOutboundGateway.this.replyContainer.stop();
-					JmsOutboundGateway.this.idleTask.cancel(false);
+					Objects.requireNonNull(JmsOutboundGateway.this.idleTask).cancel(false);
```
artembilan: This is not OK.
We have to extract `JmsOutboundGateway.this.idleTask` into a local variable.
Guard with `if (idleTaskToCancel != null)` and so on the rest against this variable.
But yeah... `JmsOutboundGateway.this.idleTask = null;` is still correct.

spring-integration-jms/src/main/java/org/springframework/integration/jms/SubscribableJmsChannel.java
```diff
@@ -238,7 +240,9 @@ else if (this.logger.isWarnEnabled()) {
 					}
 				}
 				else {
-					throw new MessageDeliveryException(messageToSend, exceptionMessage, e);
+					throw messageToSend != null
+							? new MessageDeliveryException(messageToSend, exceptionMessage, e)
+							: new MessageDeliveryException(exceptionMessage);
```
artembilan: This is not correct.
We are in a `MessageDispatchingException` case, which can happen only from the `this.dispatcher.dispatch(messageToSend);`.
Therefore `messageToSend` is never null here when we reach this place.
So, I suggest to have a `@SuppressWarnings("NullAway")` as we discussed before.

artembilan (changes requested):
I think JMS is one of the hardest modules to apply Nullability.
Thank you for doing this @mjd507 !

spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
mjd507: I see. thank you. we do propagate the `BeanFactory` from `JmsMessageDrivenEndpoint` to this listener.
```java
	@Override
	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
		super.setApplicationContext(applicationContext);
		this.listener.setBeanFactory(applicationContext);
	}
```

spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
```diff
@@ -175,7 +178,7 @@ public void setShouldTrack(boolean shouldTrack) {
 
 	@Override
 	public String getComponentName() {
-		return this.gatewayDelegate.getComponentName();
+		return Objects.requireNonNull(this.gatewayDelegate.getComponentName());
```
artembilan: I think we need to make this and delegate getter as `@Nullable`.
We probably even need to push `@Nullable` down to the `NamedComponent`.
But that is a different story and might effect a lot in the project.

artembilan (changes requested):
Last one I suppose.
Thanks

spring-integration-jms/src/main/java/org/springframework/integration/jms/ChannelPublishingJmsMessageListener.java
mjd507: updated, so for gateway & endpoint , `getComponentName ` never null, because the spring aware method will at least set beanName,
but here the listener and its internal delegate, the `getComponentName ` method depends on whether it's caller set this componentName? am i correct

artembilan: Yes. And there is no guarantee if that is going to be called if it is not a bean.
As you see we do that manually in the `JmsMessageDrivenEndpoint.onInit()`

PR #10195, opened 2025-07-07
Apply Nullability to `spring-integration-websocket` module

anthologia:
Related to: #10083

### DIFF
--- a/spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
+++ b/spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
@@ -59,23 +59,21 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 
 	private final String[] paths;
 
-	@SuppressWarnings("NullAway.Init")
-	private HandshakeHandler handshakeHandler;
+	private @Nullable HandshakeHandler handshakeHandler;
 
-	private HandshakeInterceptor @Nullable [] interceptors;
+	private HandshakeInterceptor[] interceptors = new HandshakeInterceptor[0];
 
 	private WebSocketHandlerDecoratorFactory @Nullable [] decoratorFactories;
 
 	private @Nullable SockJsServiceOptions sockJsServiceOptions;
 
-	private String @Nullable [] origins;
+	private String[] origins = new String[0];
 
 	private boolean autoStartup = true;
 
 	private int phase = 0;
 
-	@SuppressWarnings("NullAway.Init")
-	private TaskScheduler sockJsTaskScheduler;
+	private @Nullable TaskScheduler sockJsTaskScheduler;
 
 	public ServerWebSocketContainer(String... paths) {
 		Assert.notEmpty(paths, "'paths' must not be empty");
@@ -149,7 +147,7 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 		this.sockJsTaskScheduler = sockJsTaskScheduler;
 	}
 
-	public TaskScheduler getSockJsTaskScheduler() {
+	public @Nullable TaskScheduler getSockJsTaskScheduler() {
 		return this.sockJsTaskScheduler;
 	}
 
@@ -164,6 +162,7 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 			}
 		}
 
+		Assert.notNull(this.handshakeHandler, "'handshakeHandler' must not be null");
 		WebSocketHandlerRegistration registration = registry.addHandler(webSocketHandler, this.paths)
 				.setHandshakeHandler(this.handshakeHandler);
 
@@ -240,8 +239,8 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 	public void start() {
 		this.lock.lock();
 		try {
-			if (this.handshakeHandler instanceof Lifecycle && !isRunning()) {
-				((Lifecycle) this.handshakeHandler).start();
+			if (this.handshakeHandler instanceof Lifecycle lifeCycleHandler && !isRunning()) {
+				lifeCycleHandler.start();
 			}
 		}
 		finally {
@@ -251,15 +250,15 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 
 	@Override
 	public void stop() {
-		if (isRunning()) {
-			((Lifecycle) this.handshakeHandler).stop();
+		if (this.handshakeHandler instanceof Lifecycle lifeCycleHandler && isRunning()) {
+			lifeCycleHandler.stop();
 		}
 	}
 
 	@Override
 	public void stop(Runnable callback) {
-		if (isRunning()) {
-			((Lifecycle) this.handshakeHandler).stop();
+		if (this.handshakeHandler instanceof Lifecycle lifeCycleHandler && isRunning()) {
+			lifeCycleHandler.stop();
 		}
 		callback.run();
 	}
--- a/spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
+++ b/spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
@@ -61,13 +61,13 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 
 	private @Nullable HandshakeHandler handshakeHandler;
 
-	private HandshakeInterceptor[] interceptors = new HandshakeInterceptor[0];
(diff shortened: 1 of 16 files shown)


spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -56,20 +59,22 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 
 	private final String[] paths;
 
+	@SuppressWarnings("NullAway.Init")
```
artembilan: This can be null if respective setter is not called.
Let's have `Assert.notNull()` before calling `.setHandshakeHandler(this.handshakeHandler)` in the `registerWebSocketHandlers()`!

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -56,20 +59,22 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
...
 	private HandshakeHandler handshakeHandler;
 
-	private HandshakeInterceptor[] interceptors;
+	private HandshakeInterceptor @Nullable [] interceptors;
```
artembilan: Can we have it to be initialized to empty arrays instead?
Looks like there is a check `if (!ObjectUtils.isEmpty(interceptors)) {`.
However, I agree that `WebSocketHandlerRegistration` API has to support nulls, and for the mentioned `HandshakeHandler`, too.
At the same time the `AbstractWebSocketHandlerRegistration` does support null:
```
public WebSocketHandlerRegistration setHandshakeHandler(@Nullable HandshakeHandler handshakeHandler) {
```

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -56,20 +59,22 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
...
 
-	private WebSocketHandlerDecoratorFactory[] decoratorFactories;
+	private WebSocketHandlerDecoratorFactory @Nullable [] decoratorFactories;
 
-	private SockJsServiceOptions sockJsServiceOptions;
+	private @Nullable SockJsServiceOptions sockJsServiceOptions;
 
-	private String[] origins;
+	private String @Nullable [] origins;
```
artembilan: Same here for an empty array by default.

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -56,20 +59,22 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
...
 
 	private boolean autoStartup = true;
 
 	private int phase = 0;
 
+	@SuppressWarnings("NullAway.Init")
 	private TaskScheduler sockJsTaskScheduler;
```
artembilan: Let's revise this!
It indeed can be null.
Especially its getter.

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
```diff
@@ -324,6 +327,7 @@ else if (StompCommand.RECEIPT.equals(stompCommand)) {
 		}
 		else {
 			if (this.useBroker) {
+				Assert.notNull(this.brokerHandler, "brokerHandler is required");
```
artembilan: This is redundant.
We do have such a check in the `onInit()`.
This method has to be marked with a `@SuppressWarnings("NullAway") // Dataflow analysis limitation` instead.

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
```diff
@@ -346,8 +350,9 @@ private boolean isProcessingTypeOrCommand(SimpMessageHeaderAccessor headerAccess
 				&& !checkDestinationPrefix(headerAccessor.getDestination());
 	}
 
-	private boolean checkDestinationPrefix(String destination) {
+	private boolean checkDestinationPrefix(@Nullable String destination) {
 		if (this.useBroker) {
+			Assert.notNull(this.brokerHandler, "brokerHandler is required");
```
artembilan: DITTO

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
```diff
@@ -374,7 +379,10 @@ private void produceConnectAckMessage(Message<?> message, SimpMessageHeaderAcces
 	}
 
 	private void produceMessage(Message<?> message, SimpMessageHeaderAccessor headerAccessor) {
-		Object payload = this.messageConverter.fromMessage(message, this.payloadType.get());
+		Class<?> targetType = this.payloadType.get();
+		Assert.notNull(targetType, "payloadType must not be null");
```
artembilan: This is not correct: the value of that `AtomicReference` is never null.
Actually, I'm thinking if it has to be an `AtomicReference` at all.
Can you experiment if that could be just plain `Class<?>` property?

artembilan (changes requested):

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
anthologia: Agreed. No need for `AtomicReference` since this isn't even expression.

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -56,21 +59,21 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 
 	private final String[] paths;
 
-	private HandshakeHandler handshakeHandler;
+	private @Nullable HandshakeHandler handshakeHandler;
 
-	private HandshakeInterceptor[] interceptors;
+	private HandshakeInterceptor[] interceptors = new HandshakeInterceptor[0];
```
artembilan: I wonder why this doesn't work for you:
```
private HandshakeInterceptor[] interceptors = {};
```
?

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -56,21 +59,21 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
...
 
-	private WebSocketHandlerDecoratorFactory[] decoratorFactories;
+	private WebSocketHandlerDecoratorFactory @Nullable [] decoratorFactories;
 
-	private SockJsServiceOptions sockJsServiceOptions;
+	private @Nullable SockJsServiceOptions sockJsServiceOptions;
 
-	private String[] origins;
+	private String[] origins = new String[0];
```
artembilan: Same here:
```
	private String[] origins = {};
```

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -159,14 +162,23 @@ public void registerWebSocketHandlers(WebSocketHandlerRegistry registry) {
...
 
+		configureRegistration(registration);
 		configureSockJsOptionsIfAny(registration);
 	}
 
+	private void configureRegistration(WebSocketHandlerRegistration registration) {
+		if (this.interceptors != null) {
+			registration.addInterceptors(this.interceptors);
+		}
+		if (this.origins != null) {
```
artembilan: That was exactly the reason to have an empty arrays by default instead of null.
So, we would avoid these conditions and such a method altogether.
What do I miss, please?

artembilan (changes requested):

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
anthologia: Sorry, I'll be more careful and pay attention.. Thanks for your patience. 🥲
I'll fix it tomorrow.

PR #10196, opened 2025-07-09
Add JSpecify Nullability to the xml package

cppwfs:
* Tests were updated
* Test XML files updated to handle non nullable attributes and parameters
* In some cases I reformatted the whole xml file vs the changed section.   If this is a problem I'll undo.
* In many circumstances the chances of having null, should not happen, so `Objects.requireNonNull` was used.  But wonder if we should use Assert.state and have a message.   Thoughts?


spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -91,7 +92,7 @@ public static void processOutboundTypeAttributes(Element element, ParserContext
 
 	}
 
-	public static boolean areMutuallyExclusive(String query, BeanDefinition statementExpressionDef,
+	public static boolean areMutuallyExclusive(@Nullable String query, @Nullable BeanDefinition statementExpressionDef,
```
artembilan: How that happened that `query` is `null`?
According to this class logic, the `element.getAttribute("query")` does not return null.
So, I'm not sure in the change, please.

The `statementExpressionDef` is indeed could be null.
However that is not covered by the `IntegrationNamespaceUtils.createExpressionDefIfAttributeDefined()`.
The method is not marked as `@Nullable`, therefore NullAway  compiler does not fail.

I think it is also a time to hide this `areMutuallyExclusive()` method. 
It really must not be `public`.
Maybe we should just remove that `CassandraParserUtilsTests` altogether?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelAdapterParser.java
```diff
@@ -62,6 +63,7 @@ else if (!StringUtils.hasText(id)) {
 	}
 
 	@Override
+	@SuppressWarnings("NullAway")
```
artembilan, line 65: Any explanation for this, please?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelParser.java
```diff
@@ -107,6 +110,6 @@ protected void registerBeanDefinition(BeanDefinitionHolder definition, BeanDefin
 	 * @param parserContext The parser context.
 	 * @return The bean definition builder.
 	 */
-	protected abstract BeanDefinitionBuilder buildBeanDefinition(Element element, ParserContext parserContext);
+	protected abstract @Nullable BeanDefinitionBuilder buildBeanDefinition(Element element, ParserContext parserContext);
```
artembilan: This is not correct.
This contract must not never return `null`.
Why is that?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelParser.java
```diff
@@ -44,7 +47,7 @@ public abstract class AbstractChannelParser extends AbstractBeanDefinitionParser
 	@Override
 	protected AbstractBeanDefinition parseInternal(Element element, ParserContext parserContext) {
 		BeanDefinitionBuilder builder = this.buildBeanDefinition(element, parserContext);
-		AbstractBeanDefinition beanDefinition = builder.getBeanDefinition();
+		AbstractBeanDefinition beanDefinition = Objects.requireNonNull(builder).getBeanDefinition();
```
artembilan: Since `buildBeanDefinition()` never returns null, this is extra not required.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractConsumerEndpointParser.java
```diff
@@ -78,16 +80,16 @@ protected String resolveId(Element element, AbstractBeanDefinition definition, P
 	 * @param parserContext The parser context.
 	 * @return The bean definition builder.
 	 */
-	protected abstract BeanDefinitionBuilder parseHandler(Element element, ParserContext parserContext);
+	protected abstract @Nullable BeanDefinitionBuilder parseHandler(Element element, ParserContext parserContext);
```
artembilan: Please, revise this.
I don't think it is correct: the handler can never be null for consumer endpoint.

If there is a null, like in this case:
```
		else if (hasRef && hasExpression) {
			parserContext.getReaderContext().error(
					"Only one of 'ref' or 'expression' is permitted, not both, on element " +
							IntegrationNamespaceUtils.createElementDescription(element) + ".", source);
			return null;
		}
```
then better to throw an exception.
Although it won't be reached according to that `error()` contract which throws a `BeanDefinitionParsingException` anyway.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractConsumerEndpointParser.java
```diff
@@ -140,9 +142,9 @@ protected final AbstractBeanDefinition parseInternal(Element element, ParserCont
 
 		String handlerBeanName =
 				BeanDefinitionReaderUtils.generateBeanName(handlerBeanDefinition, parserContext.getRegistry());
-		String[] handlerAlias = IntegrationNamespaceUtils.generateAlias(element);
 		parserContext.registerBeanComponent(
-				new BeanComponentDefinition(handlerBeanDefinition, handlerBeanName, handlerAlias));
+				new BeanComponentDefinition(handlerBeanDefinition, handlerBeanName,
+						IntegrationNamespaceUtils.generateAlias(element)));
```
artembilan: Why this refactoring?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractConsumerEndpointParser.java
```diff
@@ -194,13 +196,11 @@ private void registerChannelForCreation(ParserContext parserContext, String inpu
...
 			}
-
 			@SuppressWarnings("unchecked")
 			Collection<String> channelCandidateNames =
-					(Collection<String>) caValues.getArgumentValue(0, Collection.class)
+					(Collection<String>) Objects.requireNonNull(caValues.getArgumentValue(0, Collection.class))
 							.getValue(); // NOSONAR see comment above
-			channelCandidateNames.add(inputChannelName); // NOSONAR
-
+			Objects.requireNonNull(channelCandidateNames).add(inputChannelName); // NOSONAR
```
artembilan: I think all changes in this method are redundant.
See the logic:
```
			if (vh == null) { //although it should never happen if it does we can fix it
				caValues.addIndexedArgumentValue(0, new ManagedSet<String>());
			}
```

And then we call get on that ctor argument.
Therefore it is never null over there.
I would say this is a `@SuppressWarnings("NullAway") // Dataflow analysis limitation` candidate.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractDelegatingConsumerEndpointParser.java
```diff
@@ -39,7 +40,7 @@
 abstract class AbstractDelegatingConsumerEndpointParser extends AbstractConsumerEndpointParser {
 
 	@Override
-	protected final BeanDefinitionBuilder parseHandler(Element element, ParserContext parserContext) {
+	protected final @Nullable BeanDefinitionBuilder parseHandler(Element element, ParserContext parserContext) {
```
artembilan: As I said before: this is never null.
Please, fix respectively.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractPollingInboundChannelAdapterParser.java
```diff
@@ -59,6 +60,7 @@ else if (source instanceof RuntimeBeanReference runtimeBeanReference) {
 		}
 		else {
 			parserContext.getReaderContext().error("Wrong 'source' type: must be 'BeanDefinition' or 'RuntimeBeanReference'", source);
+			throw new IllegalStateException("sourceBeanName must not be null");
```
artembilan: I would just copy/paste the message from above `errro()` call.
Although this our exception will never be reached according `error()` contract, it won't hurt to short-circuit NullAway code flow expectations.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractPollingInboundChannelAdapterParser.java
```diff
@@ -79,6 +81,6 @@ else if (source instanceof RuntimeBeanReference runtimeBeanReference) {
 	 * @param parserContext The parser context.
 	 * @return The bean metadata element.
 	 */
-	protected abstract BeanMetadataElement parseSource(Element element, ParserContext parserContext);
+	protected abstract @Nullable BeanMetadataElement parseSource(Element element, ParserContext parserContext);
```
artembilan: Same here: must never be null.
Similar `throw new IllegalStateException()` is OK in all those places where `return null;` since it is, again, covered by that `error()` call.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AnnotationConfigParser.java
```diff
@@ -67,11 +68,11 @@ private static final class ExtendedAnnotationMetadata extends AnnotationMetadata
 		}
 
 		@Override
-		public Map<String, Object> getAnnotationAttributes(String annotationType) {
+		public @Nullable Map<String, @Nullable Object> getAnnotationAttributes(String annotationType) {
 			if (EnablePublisher.class.getName().equals(annotationType)) {
 				Element enablePublisherElement = DomUtils.getChildElementByTagName(this.element, "enable-publisher");
 				if (enablePublisherElement != null) {
-					Map<String, Object> attributes = new HashMap<>();
+					Map<String, @Nullable Object> attributes = new HashMap<>();
```
artembilan, line 75: Why is this?
How that happened that values could be null?
I see this is a contract:
```
    /**
     * Retrieves an attribute value by name.
     * @param name The name of the attribute to retrieve.
     * @return The <code>Attr</code> value as a string, or the empty string
     *   if that attribute does not have a specified or default value.
     */
    public String getAttribute(String name);
```

So, what made you to add this annotation for values in the map?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/ChainParser.java
```diff
@@ -65,7 +67,7 @@ public class ChainParser extends AbstractConsumerEndpointParser {
 	private final Log logger = LogFactory.getLog(this.getClass());
 
 	@Override
-	protected BeanDefinitionBuilder parseHandler(Element element, ParserContext parserContext) {
+	protected @Nullable BeanDefinitionBuilder parseHandler(Element element, ParserContext parserContext) {
```
artembilan: Same as before, not-null due to `error()` before `return null;`

spring-integration-core/src/main/java/org/springframework/integration/config/xml/ChainParser.java
```diff
@@ -123,7 +125,7 @@ protected String resolveId(Element element, AbstractBeanDefinition definition, P
 		return id;
 	}
 
-	private BeanMetadataElement parseChild(String chainHandlerId, Element element, int order, ParserContext parserContext,
+	private @Nullable BeanMetadataElement parseChild(String chainHandlerId, Element element, int order, ParserContext parserContext,
```
artembilan: Same here, please.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/ChannelInterceptorParser.java
```diff
@@ -74,6 +76,7 @@ else if ("ref".equals(localName)) {
 						parserContext.getReaderContext().error(
 								"unsupported interceptor element '" + localName + "'", childElement);
 					}
+					Assert.state(parser != null, "no parser registered for '" + localName + "'");
```
artembilan: This is redundant.
Place "dummy" `throw new IllegalStateException()` after that `error()` instead.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/DefaultInboundChannelAdapterParser.java
```diff
@@ -46,8 +49,8 @@
  */
 public class DefaultInboundChannelAdapterParser extends AbstractPollingInboundChannelAdapterParser {
 
-	@Override // NOSONAR complexity
-	protected BeanMetadataElement parseSource(Element element, ParserContext parserContext) { // NOSONAR
+	@Override
+	protected @Nullable BeanMetadataElement parseSource(Element element, ParserContext parserContext) {
```
artembilan: Not-null. For same reason as before.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/DefaultInboundChannelAdapterParser.java
```diff
@@ -76,6 +79,7 @@ protected BeanMetadataElement parseSource(Element element, ParserContext parserC
 								+ "element " + IntegrationNamespaceUtils.createElementDescription(element) + ".", source);
 				return null;
 			}
+			Assert.state(innerBeanDef != null, "No inner bean definition found"); //Required because JSpecify does not know how to use handle a preparsed boolean
```
artembilan: I think `Objects.requireNonNull()` would read better.
We definitely never going to have it as null when we reach this place.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/DefaultInboundChannelAdapterParser.java
```diff
@@ -91,6 +95,7 @@ else if (hasScriptElement) {
 								+ IntegrationNamespaceUtils.createElementDescription(element) + ".", source);
 				return null;
 			}
+			Assert.state(scriptElement != null, "No script element found");
```
artembilan: Same `Objects.requireNonNull()` here instead.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/GatewayParser.java
```diff
@@ -70,10 +72,10 @@ public class GatewayParser implements BeanDefinitionParser {
 	private static final String MAPPER_ATTR = "mapper";
 
 	@Override
-	public BeanDefinition parse(final Element element, ParserContext parserContext) {
+	public @Nullable BeanDefinition parse(final Element element, ParserContext parserContext) {
 		boolean isNested = parserContext.isNested();
 
-		final Map<String, Object> gatewayAttributes = new HashMap<>();
+		final Map<String, @Nullable Object> gatewayAttributes = new HashMap<>();
```
artembilan, line 78: I would prefer to suppress warning instead of this unnecessary complexity on us.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/GatewayParser.java
```diff
@@ -263,8 +265,8 @@ else if (StringUtils.hasText(asyncExecutor)) {
 		return new BeanDefinitionHolder(beanDefinition, id);
 	}
 
-	private static BeanDefinition getMethodMetadataBeanDefinition(String defaultPayloadExpression,
-			Map<String, Object>[] defaultHeaders) {
+	private static BeanDefinition getMethodMetadataBeanDefinition(@Nullable String defaultPayloadExpression,
+			@Nullable Map<String, Object> @Nullable [] defaultHeaders) {
```
artembilan, line 269: OK. I see.
Let modify that `headers()` to inject and empty array instead!
Looks like same for `methods()`.
This way a `gatewayAttributes` map would never return null for key we request.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/IntegrationNamespaceUtils.java
```diff
@@ -286,18 +287,19 @@ public static String getTextFromAttributeOrNestedElement(Element element, String
...
+	public static @Nullable BeanComponentDefinition parseInnerHandlerDefinition(Element element, ParserContext parserContext) {
 		// parses out the inner bean definition for concrete implementation if defined
 		List<Element> childElements = DomUtils.getChildElementsByTagName(element, "bean");
 		BeanComponentDefinition innerComponentDefinition = null;
 		if (childElements.size() == 1) {
 			Element beanElement = childElements.get(0);
 			BeanDefinitionParserDelegate delegate = parserContext.getDelegate();
 			BeanDefinitionHolder bdHolder = delegate.parseBeanDefinitionElement(beanElement);
-			bdHolder = delegate.decorateBeanDefinitionIfRequired(beanElement, bdHolder); // NOSONAR never null
+			Assert.state(bdHolder != null, "bdHolder must not be null");
```
artembilan: The `Objects.requireNonNull()` is better here.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/PointToPointChannelParser.java
```diff
@@ -44,7 +45,7 @@
 public class PointToPointChannelParser extends AbstractChannelParser {
 
 	@Override
-	protected BeanDefinitionBuilder buildBeanDefinition(Element element, ParserContext parserContext) {
+	protected @Nullable BeanDefinitionBuilder buildBeanDefinition(Element element, ParserContext parserContext) {
```
artembilan: As we discussed before: this is never null.

spring-integration-core/src/test/java/org/springframework/integration/config/ChainParserTests-context.xml
```diff
@@ -125,7 +125,7 @@
 
 	<chain id="subComponentsIdSupport1" input-channel="subComponentsIdSupport1Channel">
 		<splitter id="splitterWithinChain"/>
-		<resequencer id="resequencerWithinChain"/>
+		<resequencer id="resequencerWithinChain" correlation-strategy="errorChannel" release-strategy="errorChannel"/>
```
artembilan: I think we need to revise the parser and avoid such a change in the XML configs.
See Java DSL for example:
```
	/**
	 * Populate the
	 * {@link org.springframework.integration.aggregator.ResequencingMessageHandler} with
	 * default options.
	 * @return the current {@link BaseIntegrationFlowDefinition}.
	 */
	public B resequence() {
```
so, having `<resequencer/>` in the XML is totally fine.
If that is otherwise, the respective XML parser has to be fixed.

spring-integration-core/src/test/java/org/springframework/integration/config/xml/EndpointRoleParserTests-context.xml
```diff
@@ -1,38 +1,41 @@
...
 
-	<int:outbound-channel-adapter id="out4" channel="foo" role="${cluster.3}" auto-startup="false" ref="sink" method="foo" />
+	<int:outbound-channel-adapter id="out4" channel="foo" role="${cluster.3}" auto-startup="false" ref="sink"
+								  method="foo"/>
 
-	<bean id="sink" class="org.springframework.integration.config.xml.EndpointRoleParserTests$Sink" />
+	<bean id="sink" class="org.springframework.integration.config.xml.EndpointRoleParserTests$Sink"/>
 
-	<int:bridge id="bridge" input-channel="foo" output-channel="nullChannel" role="cluster" auto-startup="false" />
+	<int:bridge id="bridge" input-channel="foo" output-channel="nullChannel" role="cluster" auto-startup="false"/>
```
artembilan: I'm not fully sure what is the crucial part of the change in this file, please?

artembilan (changes requested):
We can discuss my review off-line if you prefer.
Thanks

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
cppwfs: I will remove the `@Nullable`.
Explanation:
This was a last minute change that I made.  
It was based on the action taken on line 98 where query has a !StringUtil.`hasText` which checks for empty string and `null`.  In this case I preferred the method make that call for Null or empty.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelParser.java
cppwfs: According to the `this.buildBeanDefinition` is marked `@Nullable`.   The `KafkaChannelParser` implementation does return a null.

Discussed above.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelParser.java
cppwfs: The KafkaChannelParser implementation does return a null.
Do you want me to revisit all implmentations and see if there is a path other than returning a 'null'.   i.e. throw an exception?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractConsumerEndpointParser.java
cppwfs: The variable `handlerAlias` is only being used by the `alias` parameter of the`BeanComponentDefinition` ctor.   So the the method that initializes the `handlerAlias` can be used in stead of the `handlerAlias` in the `BeanComponentDefinition` ctor's `alias` parameter.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AnnotationConfigParser.java
cppwfs: The contract with `AnnotatedTypeMetadata.getAnnotationAttributes` (parent of `AnnotationConfigParser` class) requires the  `@Nullable` tag on the value portion of the map.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/GatewayParser.java
cppwfs: To verify your request, the caller of this method should make sure a stubbed empty `defaultHeaders passed into this method instead of `null`?  Same for `defaultPayloadExpression`?

spring-integration-core/src/test/java/org/springframework/integration/config/ChainParserTests-context.xml
cppwfs: Let's discuss

spring-integration-core/src/test/java/org/springframework/integration/config/xml/EndpointRoleParserTests-context.xml
cppwfs: I mistakingly used the IDE to update the format.  I will reset it back to its original format.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelAdapterParser.java
cppwfs: The 	`createDirectChannel(element, parserContext);` on line 72 can return a null.  And it appears some adapters like AMQP may accept a null directChannel name.   
We can discuss.  It maybe just a test fix.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractConsumerEndpointParser.java
cppwfs: In these cases the `@Nullable` was removed as well as the `return null`.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelParser.java
artembilan: Well, looks like that was a blind fix over there with that `@Nullable`.
The null is there exactly in the place after `error()` we have discussed all around.
So, please, fix that `KafkaChannelParser` removing `@Nullable` and throwing "dummy" `IllegalStateException` as we discussed for other places.

spring-integration-core/src/test/java/org/springframework/integration/config/ChainParserTests-context.xml
artembilan: We develop here a library of API and components for target developers.
It was an intention to have `<resequencer>` without any arguments.
Having you adding those `correlation-strategy` and `release-strategy` explicitly means that something is broken.
And this API for target developers is not in the appropriate shape.
We might look into this separately from anything else.
But I'd prefer to have all those XML configs reverted since the change you propose for them is not a goal of our framework.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/xml/PointToPointChannelParser.java
```diff
@@ -78,7 +79,7 @@ else if ((queueElement = DomUtils.getChildElementByTagName(element, "rendezvous-
 						"The 'fixed-subscriber' attribute is not allowed when a <queue/> child element is present.",
 						element);
 			}
-			return builder;
+			return Objects.requireNonNull(builder);
```
artembilan: Not correct.
Has to be "dummy" exception instead of that `return null;`.
That dummy will never be reached, but with this `requireNonNull()` we do extra hops for the value which is never null.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractConsumerEndpointParser.java
artembilan: Correct, but the reason for such a local variable is to make the code more easier to read.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/PointToPointChannelParser.java
```diff
@@ -78,6 +78,7 @@ else if ((queueElement = DomUtils.getChildElementByTagName(element, "rendezvous-
 						"The 'fixed-subscriber' attribute is not allowed when a <queue/> child element is present.",
 						element);
 			}
+			Assert.state(builder != null, "builder must not be null");
```
artembilan, line 81: Why not throw our friend "dummy" exception after error instead?
I mean this extra call is redundant and misleading.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractChannelAdapterParser.java
cppwfs: This change is valid via our conversation.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/DefaultInboundChannelAdapterParser.java
```diff
@@ -121,7 +119,7 @@ else if (hasRef) {
 				result = sourceValue;
 			}
 		}
-		return result;
+		return Objects.requireNonNull(result);
```
artembilan, line 136: Same here: why not exception after `error()`?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/DefaultInboundChannelAdapterParser.java
```diff
@@ -74,13 +76,12 @@ protected BeanMetadataElement parseSource(Element element, ParserContext parserC
...
 								+ "element " + IntegrationNamespaceUtils.createElementDescription(element) + ".", source);
-				return null;
 			}
 			if (hasMethod) {
-				result = parseMethodInvokingSource(innerBeanDef, methodName, element, parserContext);
+				result = parseMethodInvokingSource(Objects.requireNonNull(innerBeanDef), methodName, element, parserContext);
 			}
 			else {
-				result = innerBeanDef.getBeanDefinition();
+				result = Objects.requireNonNull(innerBeanDef).getBeanDefinition();
```
artembilan, line 89: DITTO

spring-integration-core/src/main/java/org/springframework/integration/config/xml/ChainParser.java
```diff
@@ -142,14 +142,13 @@ private BeanMetadataElement parseChild(String chainHandlerId, Element element, i
 			BeanDefinition beanDefinition = parserContext.getDelegate().parseCustomElement(element, parentDefinition);
 			if (beanDefinition == null) {
 				parserContext.getReaderContext().error("child BeanDefinition must not be null", element);
-				return null;
```
artembilan, line 145: Why not exception?

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractPollingInboundChannelAdapterParser.java
```diff
@@ -60,7 +61,7 @@ else if (source instanceof RuntimeBeanReference runtimeBeanReference) {
 		else {
 			parserContext.getReaderContext().error("Wrong 'source' type: must be 'BeanDefinition' or 'RuntimeBeanReference'", source);
 		}
-
+		Assert.state(sourceBeanName != null, "'sourceBeanName' must not be null");
```
artembilan: Same here: exception after `error()` in the previous `else`

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractConsumerEndpointParser.java
```diff
@@ -85,9 +87,9 @@ protected String getInputChannelAttributeName() {
 	}
 
 	@Override
-	protected final AbstractBeanDefinition parseInternal(Element element, ParserContext parserContext) {
+	protected final @Nullable AbstractBeanDefinition parseInternal(Element element, ParserContext parserContext) {
 		BeanDefinitionBuilder handlerBuilder = parseHandler(element, parserContext);
-		IntegrationNamespaceUtils.setValueIfAttributeDefined(handlerBuilder, element, "output-channel",
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(Objects.requireNonNull(handlerBuilder), element, "output-channel",
```
artembilan: This is not correct: the `parseHandler()` never returns `null`

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -91,7 +92,7 @@ public static void processOutboundTypeAttributes(Element element, ParserContext
 
 	}
 
-	public static boolean areMutuallyExclusive(String query, BeanDefinition statementExpressionDef,
+	public static boolean areMutuallyExclusive(String query, @Nullable BeanDefinition statementExpressionDef,
```
artembilan: And why this method is not `private` yet since you have already removed that redundant test?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/xml/PointToPointChannelParser.java
cppwfs: I merged to aggressively and put a few changes in the 2nd commit.  
But here is the excerpt from the commit message:
`PointToPointChannelParser` requires a null check on line 82 because builder because it may not have been set.   With this assert we can throw away the previous bogus exception.

Sorry for the noise.

artembilan: I see. OK. 
We do have the whole coverage for `queueElement`, but looks like JSpecify cannot "feel".
Let's leave it here!
Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/xml/DefaultInboundChannelAdapterParser.java
cppwfs: There is a path where result is never set.   Unless we make the assumption that the check on line 118 can be removed.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/ChainParser.java
cppwfs: Adding for consistency.  NullAway did not report an error here.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractPollingInboundChannelAdapterParser.java
```diff
@@ -59,8 +59,8 @@ else if (source instanceof RuntimeBeanReference runtimeBeanReference) {
 		}
 		else {
 			parserContext.getReaderContext().error("Wrong 'source' type: must be 'BeanDefinition' or 'RuntimeBeanReference'", source);
+			throw new IllegalStateException("This dummy exception is meant to signify to  NullAway that an the error method throws an exception");
```
artembilan: I don't think it is correct and polite to say that as an exception message.
Just copy/paste from `error()` and be with it!
Plus, keep in mind the rule of not more than 120 symbols in line.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractPollingInboundChannelAdapterParser.java
```diff
@@ -59,8 +59,8 @@ else if (source instanceof RuntimeBeanReference runtimeBeanReference) {
...
 		}
-
```
artembilan, line 63: Please, bring this line back.
The code should be as readable as possible.

spring-integration-core/src/main/java/org/springframework/integration/config/xml/ChainParser.java
```diff
@@ -142,14 +142,14 @@ private BeanMetadataElement parseChild(String chainHandlerId, Element element, i
...
+				throw new IllegalStateException("This dummy exception is meant to signify to  NullAway that an the error method throws an exception");
```
artembilan: DITTO

artembilan (changes requested):

Issue #10197, opened 2025-07-09
Deprecate CompositeCodec(Codec defaultCodec)


cppwfs:
The precept is that if using a CompositeCode, you should be using `delegates` with a `defaultCodec`.

This will need to be done in the 6.4.6 release as well.

spring-builds:
Fixed via 545028162f8a087ff0844a80017fb40358e4a0f7

PR #10198, opened 2025-07-09
Mark CompositeCodec(defaultCodec) ctor for deprecation

cppwfs:
Fixes: https://github.com/spring-projects/spring-integration/issues/10197

Update deprecated ctor to use an empty map for delegates instead of null.

**Auto-cherry-pick to `6.4.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -43,8 +43,10 @@ public class CompositeCodec implements Codec {
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
 
+	@Deprecated(since = "6.4", forRemoval = true)
 	public CompositeCodec(Codec defaultCodec) {
-		this(null, defaultCodec);
+		this.defaultCodec = defaultCodec;
+		this.delegates = Map.of();
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -42,6 +42,7 @@ public class CompositeCodec implements Codec {
 		this.defaultCodec = defaultCodec;
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
+
 	/**
 	 * @param defaultCodec codec for fallback
 	 * @deprecated since 6.4.6 in favor of {@link #CompositeCodec(Map, Codec)} with provided delegates.
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
@@ -42,7 +42,10 @@ public class CompositeCodec implements Codec {
 		this.defaultCodec = defaultCodec;
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
-
+	/**
+	 * @param defaultCodec codec for fallback
+	 * @deprecated since 6.4.6 in favor of {@link #CompositeCodec(Map, Codec)} with provided delegates.
+	 */
 	@Deprecated(since = "6.4.6", forRemoval = true)
 	public CompositeCodec(Codec defaultCodec) {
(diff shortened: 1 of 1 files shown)


spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -43,8 +43,10 @@ public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
 
+	@Deprecated(since = "6.4", forRemoval = true)
```
artembilan: Please, be more specific for the version.
The `6.4` is long gone.

artembilan (changes requested):
Missed `**Auto-cherry-pick to 6.4.x**` in the commit to let respective bots to do the work for us.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
cppwfs: I can update it to 6.5.     
Explanation:
6.4 was a one of the targets mentioned in PR that suggested this strategy: https://github.com/spring-projects/spring-integration/pull/10189/files#r2190233214
So the deprecation `since` field went to the oldest version being modified which is 6.4.

cppwfs:
I will update the PR comment with the cherrypick notation.   But it is in the commit message.

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
cppwfs: So should I remove the cherry-pick and set this for 6.5?

artembilan: Sorry, I was not clear.
That has to be like: `since 6.4.6`

artembilan:
> cppwfs:SI-CODEC-CTOR-DEPRECATE

More convenient to use an issue number in the branch.
In this case that would be `GH-10197`.

No a problem at all: just sharing with you an experience how easier to manage issues and PRs.

cppwfs:
My bad I thought we were to use the link itself instead of the GH number.
i.e. https://github.com/spring-projects/spring-integration/issues/10197

Should we use something like this:
A) fixes: [GH-10197](https://github.com/spring-projects/spring-integration/issues/10197)
Or 
B) fixes: GH-10197

Will probably have to see the raw comment vs what it shows here, to see the differences 😄

artembilan:
No, I mean the local branch name:
image

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
artembilan:
```suggestion
	@Deprecated(since = "6.4.6", forRemoval = true)
```

artembilan (approved):

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -43,8 +43,10 @@ public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
```
artembilan, line 45:
```suggestion
	/**
	 * @param defaultCodec codec for fallback
	 * @deprecated since 6.4.6 in favor of {@link #CompositeCodec(Map, Codec)} with provided delegates.
	 */
```

artembilan (changes requested):

artembilan (approved):
LGTM.
I added all the required changes into this PR.

Will be merged when build is green.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/codec/CompositeCodec.java
```diff
@@ -42,9 +42,14 @@ public CompositeCodec(Map<Class<?>, Codec> delegates, Codec defaultCodec) {
 		this.defaultCodec = defaultCodec;
 		this.delegates = new HashMap<Class<?>, Codec>(delegates);
 	}
-
+	/**
```
artembilan, line 46:
```suggestion

	/**
```

PR #10203, opened 2025-07-14
GH-10058: Add SpEL JSON accessors and converter with Jackson 3

anthologia:
Related to: #10058

* Add `JsonNodePropertyAccessor` for reading JSON object properties
* Add `JsonArrayNodeIndexAccessor` for array index access
* Add `JsonNodeWrapperConverter` for type conversion between `JsonNodeWrapper` and `JsonNode`
* Add `EmbeddedJsonMessageHeadersMessageMapper` for JSON message serialization with embedded headers
* Deprecate Jackson 2 SpEL JSON accessors
---
I've prepared changes to deprecate other Jackson 2 classes and make 3 as the project default, but this involves several changes and discussion points. (I encountered an NPE in Jackson 3 logic that maybe needs to be fixed by the Jackson team, but I've implemented a workaround in our code.)

For this reason, I've kept it separate from the current Accessor PR and will address it in the upcoming PR.


build.gradle
```diff
@@ -259,7 +259,7 @@ configure(javaProjects) { subproject ->
 
     tasks.withType(Javadoc) {
         options.addBooleanOption('Xdoclint:syntax', true) // only check syntax with doclint
-        options.addBooleanOption('Werror', true) // fail build on Javadoc warnings
+//  TODO until deprecation cleanup      options.addBooleanOption('Werror', true) // fail build on Javadoc warnings
```
anthologia: Temporarily disabled Javadoc `Werror` due to deprecation warnings.

```
~/JsonNodeWrapperToJsonNodeConverter.java:27: warning: [removal] JsonPropertyAccessor in org.springframework.integration.json has been deprecated and marked for removal
import org.springframework.integration.json.JsonPropertyAccessor.JsonNodeWrapper;
```

spring-integration-core/src/test/java/org/springframework/integration/json/JsonIndexAccessorTests.java
```diff
@@ -34,6 +34,7 @@
  * @since 6.4
  * @see JsonPropertyAccessorTests
  */
+@Deprecated(forRemoval = true, since = "7.0")
 class JsonIndexAccessorTests extends AbstractJsonAccessorTests {
```
anthologia: Should we remove these deprecated class tests?

spring-integration-core/src/main/java/org/springframework/integration/json/JsonArrayNodeIndexAccessor.java
```diff
@@ -0,0 +1,76 @@
...
+ * Also supports negative indexes &mdash; for example, {@code myJsonArray[-1]} which equates
+ * to {@code myJsonArray[myJsonArray.length - 1]}. Furthermore, {@code null} is returned for
+ * any index that is out of bounds (see {@link ArrayNode#get(int)} for details).
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ * @see JsonNodePropertyAccessor
+ */
+public class JsonArrayNodeIndexAccessor implements IndexAccessor {
```
artembilan: Let's call it (and all new classes) as `JacksonIndexAccessor`!
That would be much easier for end-users to migrate.
And I anticipate a renaming back to `Json` when we remove those deprecated classes.

spring-integration-core/src/main/java/org/springframework/integration/json/JsonNodeWrapperConverter.java
```diff
@@ -0,0 +1,55 @@
...
+ * The {@link org.springframework.core.convert.converter.Converter} implementation for the conversion
+ * of {@link JsonNodeWrapper} to {@link JsonNode},
+ * when the {@link JsonNodeWrapper} can be a result of the expression
+ * for JSON in case of the {@link JsonNodePropertyAccessor} usage.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ */
+public class JsonNodeWrapperConverter implements GenericConverter {
```
artembilan, line 39: This name is fine.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/json/EmbeddedJsonMessageHeadersMessageMapper.java
```diff
@@ -0,0 +1,282 @@
...
+ * {@link JacksonMessagingUtils#messagingAwareMapper(String...)}.
+ * <p>
+ * A constructor is provided allowing the provision of such a configured object mapper.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ *
+ */
+public class EmbeddedJsonMessageHeadersMessageMapper implements BytesMessageMapper {
```
artembilan: I would call it like `EmbeddedHeadersJsonMessageMapper`

spring-integration-core/src/test/java/org/springframework/integration/json/AbstractJsonAccessorTests.java
```diff
@@ -51,6 +51,7 @@
  *
  * @since 3.0
  */
+@Deprecated(forRemoval = true, since = "7.0")
```
artembilan: No, we don't need to keep tests for deprecated API.

spring-integration-core/src/test/java/org/springframework/integration/json/JsonIndexAccessorTests.java
artembilan: Indeed we should

build.gradle
artembilan: Please, bring it back.
Probably try to use `@SuppressWarning("removal")` instead in the `JsonNodeWrapperToJsonNodeConverter`.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/json/JsonArrayNodeIndexAccessor.java
anthologia: I have a question about backward compatibility regarding the naming strategy.

If we rename `JacksonIndexAccessor` back to `JsonIndexAccessor` in the future (after removing deprecated classes), wouldn't this break backward compatibility for users who migrated to `JacksonIndexAccessor` ?

spring-integration-core/src/main/java/org/springframework/integration/json/JacksonIndexAccessor.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * <p>Supports indexes supplied as an integer literal &mdash; for example, {@code myJsonArray[1]}.
+ * Also supports negative indexes &mdash; for example, {@code myJsonArray[-1]} which equates
+ * to {@code myJsonArray[myJsonArray.length - 1]}. Furthermore, {@code null} is returned for
+ * any index that is out of bounds (see {@link ArrayNode#get(int)} for details).
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ * @see JacksonPropertyAccessor
```
artembilan, line 39: Blank line after `@since`.
I like to see sections in class JavaDoc separated.

spring-integration-core/src/test/java/org/springframework/integration/json/AbstractJacksonAccessorTests.java
```diff
@@ -34,24 +34,20 @@
...
 
 /**
- * Abstract base class for tests involving {@link JsonPropertyAccessor} and {@link JsonIndexAccessor}.
+ * Abstract base class for tests involving {@link JacksonPropertyAccessor} and {@link JacksonIndexAccessor}.
  *
- * @author Eric Bottard
- * @author Artem Bilan
- * @author Paul Martin
- * @author Pierre Lakreb
- * @author Sam Brannen
```
artembilan, line 50: Please, bring back all the names.
The class re-naming is not justification of taking authorship away.

spring-integration-core/src/test/java/org/springframework/integration/support/json/JacksonJsonObjectMapperTests.java
```diff
@@ -60,7 +60,7 @@
  *
  * @since 7.0
  */
-class JacksonJsonObjectMapperTests {
+public class JacksonJsonObjectMapperTests {
```
artembilan: Why this change, please?
JUnit Jupiter is good to work with non-public classes and methods.

artembilan (changes requested):

PR #10205, opened 2025-07-15
Apply Nullability to JDBC module

mjd507:
related to: #10083


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
```diff
@@ -172,6 +174,7 @@ public void setReplySqlParameterSourceFactory(SqlParameterSourceFactory sqlParam
 	 * @param rowMapper the {@link RowMapper} to use.
 	 */
 	public void setRowMapper(RowMapper<?> rowMapper) {
+		Assert.notNull(this.poller, "If you want to set 'rowMapper', then you must provide a 'selectQuery'.");
```
artembilan: The technical documentation and similar messages must be impersonated.
So, we cannot use words like `you`, `we`, `they` etc. in these error messages.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
```diff
@@ -113,10 +115,10 @@ protected PreparedStatementCreator getPreparedStatementCreator(String sql,
 	 * @param rowMapper the {@link RowMapper} to use.
 	 */
 	public void setRowMapper(@Nullable RowMapper<?> rowMapper) {
-		this.rowMapper = rowMapper;
 		if (rowMapper == null) {
-			this.rowMapper = new ColumnMapRowMapper();
+			rowMapper = new ColumnMapRowMapper();
```
artembilan: To re-assign method argument is bad practice.
We should always treat them as `final`.
Please, consider `if..else` instead.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * Provides lock classes for JDBC module.
+ */
+@org.jspecify.annotations.NullMarked
+package org.springframework.integration.jdbc.lock;
```
artembilan: New line in every single file, please.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageRowMapper.java
```diff
@@ -51,6 +51,7 @@ public MessageRowMapper(AllowListDeserializingConverter deserializer) {
 	}
 
 	@Override
+	@SuppressWarnings("NullAway") // See https://github.com/uber/NullAway/issues/1075
```
artembilan, line 54: I'm not sure why do we need this?
Why our `mapRow()` just cannot be `@Nullable` like that `SingleColumnRowMapper` for `queryForObject(String, Class)`?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
```diff
@@ -676,9 +681,8 @@ protected void doSetLastReleasedSequenceNumberForGroup(Object groupId, int seque
 	}
 
 	@Override
-	protected Message<?> doPollMessageFromGroup(Object groupId) {
-		String key = getKey(groupId);
-
+	protected @Nullable Message<?> doPollMessageFromGroup(Object groupId) {
+		String key = Objects.requireNonNull(getKey(groupId));
```
artembilan: I think this `Objects.requireNonNull()` should go directly to that `getKey()`.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -151,7 +153,7 @@ public DistributedLock obtain(Object lockKey) {
 		}
 	}
 
-	private String pathFor(String input) {
+	private @Nullable String pathFor(@Nullable String input) {
 		return input == null ? null : UUIDConverter.getUUID(input).toString();
```
artembilan: I think `Objects.requireNonNull()` is better here instead.
The ` input` is never null, and result of the method must not be null.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -297,7 +297,7 @@ public boolean isRunning() {
 		return this.latch.getCount() > 0;
 	}
 
-	private static String getKey(Object input) {
+	private static @Nullable String getKey(@Nullable Object input) {
```
artembilan: I believe same approach for the `Objects.requireNonNull()` as for similar places in other classes we have discussed here.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
```diff
@@ -224,6 +226,7 @@ protected List<?> doPoll(@Nullable SqlParameterSource sqlQueryParameterSource) {
 		}
 	}
 
+	@SuppressWarnings("NullAway") // Dataflow analysis limitation
```
artembilan, line 227: Can we understand what is the problem here, please?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageHandler.java
```diff
@@ -201,6 +202,7 @@ protected void handleMessageInternal(Message<?> message) {
 	 * @param keysGenerated generate key or not.
 	 * @return a generated keys for update.
 	 */
+	@SuppressWarnings("NullAway") // Dataflow analysis limitation
```
artembilan, line 206: Can we understand what is the problem here, please?

artembilan (changes requested):
Looks like hard stuff.
Thank you for such a great effort!
I appreciate this contribution.

I won't mind if you object from what I have asked in review.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageHandler.java
mjd507: I do a little change inside this huge method, and still keep it `@SuppressWarnings("NullAway") // Dataflow analysis limitation`, 
because `this.generatedKeysStatementCreator` and `this.sqlParameterSourceFactory` will not be null in their own if else branch.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcOutboundGateway.java
mjd507: ok, I refered the message from the `doInit()` method in this class, I modified that message as well.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
mjd507: make sense.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcPollingChannelAdapter.java
mjd507: errors says `error: [NullAway] passing @Nullable parameter 'this.updateSql' where @NonNull is required`
after checking, when we call this method, we already check `this.updateSql != null`. 
so I think its a Dataflow analysis limitation.

sometimes I am confused when to use `requireNonNull` and `@SuppressWarnings("NullAway") // Dataflow analysis limitation`,  
looks most places that we use `requireNonNull` can be use the `dataflow analysis limitation` as well :)

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
mjd507: updated, I had a same thought why input(groupId) could be null. in db-schema, `group_key` never null.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageRowMapper.java
mjd507: I tried like this:
```java 
public class MessageRowMapper implements RowMapper<@Nullable Message<?>> {

    public @Nullable Message<?> mapRow(ResultSet rs, int rowNum) throws SQLException {
    }
}
```

got 8 errors like this:
`error: [NullAway] Cannot pass parameter of type MessageRowMapper, as formal parameter has type RowMapper<Message<?>>, which has mismatched type parameter nullability
                        messages = namedParameterJdbcTemplate.query(query, parameters, this.messageRowMapper);`

looks like the generic type inference issue, I don't know how to fix that yet.

should we make `MessageRowMapper` a generic type like `SingleColumnRowMapper`
```java
public class MessageRowMapper<T extends Message<?>> implements RowMapper<@Nullable T> {

    public @Nullable T mapRow(ResultSet rs, int rowNum) throws SQLException {
    }
}
```

artembilan (approved):
LGTM

PR #10207, opened 2025-07-15
Fix `SpringIntegrationTestExecutionListener` for restart

artembilan:
Related to: https://github.com/spring-projects/spring-framework/issues/35168

Spring Framework has introduced recently an optimization for cached contexts. So, if one is not used, it is stopped in the cache. When we pull it next time from there, it is restarted. The `SpringIntegrationTestExecutionListener` ha a flaw to change the state of the endpoints in the application context, but never reset it back to original. The problem has become apparent because of not reset auto-startup status of endpoints and that `AC.restart()`

* Fix `SpringIntegrationTestExecutionListener` to use `setAutoStartup(false)` on `autoStartupCandidates` in the `afterTestClass()`. This is exactly a state the `MockIntegrationContext` has gathered them in its initialization phase.
* Fix `MockIntegrationContext.autoStartupCandidates` to be a `Set` instead. The point is that `postProcessBeforeInitialization()` deals with `AbstractEndpoint` beans. And then `afterSingletonsInstantiated()` also gathers those beans from the `BeanFactory`. So, with a `Set` we aim for no duplications

### DIFF
--- a/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
+++ b/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
@@ -16,11 +16,11 @@
 
 package org.springframework.integration.test.context;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
+import java.util.HashSet;
+import java.util.Set;
 
 import reactor.core.publisher.Mono;
 
@@ -79,7 +79,7 @@ public class MockIntegrationContext implements BeanPostProcessor, SmartInitializ
 
 	private final MultiValueMap<String, Object> beans = new LinkedMultiValueMap<>();
 
-	private final List<AbstractEndpoint> autoStartupCandidates = new ArrayList<>();
+	private final Set<AbstractEndpoint> autoStartupCandidates = new HashSet<>();
 
 	private ConfigurableListableBeanFactory beanFactory;
 
@@ -99,8 +99,9 @@ public class MockIntegrationContext implements BeanPostProcessor, SmartInitializ
 	}
 
 	private void addAutoStartupCandidates(AbstractEndpoint endpoint) {
-		endpoint.setAutoStartup(false);
-		this.autoStartupCandidates.add(endpoint);
+		if (this.autoStartupCandidates.add(endpoint)) {
+			endpoint.setAutoStartup(false);
+		}
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
```diff
@@ -112,8 +113,8 @@ public void afterSingletonsInstantiated() {
 				.forEach(this::addAutoStartupCandidates);
 	}
 
-	List<AbstractEndpoint> getAutoStartupCandidates() {
-		return Collections.unmodifiableList(this.autoStartupCandidates);
+	Set<AbstractEndpoint> getAutoStartupCandidates() {
```
cppwfs, line 116: Just curious why this method is at package level vs public or private?

spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
```diff
@@ -49,16 +49,17 @@ public void prepareTestInstance(TestContext testContext) {
...
 				.filter(endpoint -> !match(endpoint.getBeanName(), patterns))
 				.peek(endpoint -> endpoint.setAutoStartup(true))
 				.forEach(AbstractEndpoint::start);
 	}
 
 	@Override
 	public void afterTestClass(TestContext testContext) {
-		this.autoStartupCandidates.forEach(AbstractEndpoint::stop);
+		this.autoStartupCandidates.stream()
+				.peek(endpoint -> endpoint.setAutoStartup(false))
```
cppwfs, line 61: Peek is usually used for debugging purposes. (IIRC) 

So should we rewrite the stream as follows:
```java
	@Override
	public void afterTestClass(TestContext testContext) {
		this.autoStartupCandidates.forEach(abstractEndpoint -> {
					abstractEndpoint.setAutoStartup(false);
					abstractEndpoint.stop();
				});
	}
```

cppwfs (changes requested):
Looks good!

Just a couple of questions.

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
artembilan: The `MockIntegrationContext` is end-user API.
But these `candidates` are only for other classes in this package.
Therefor package level to have it not visible for end-user.

spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
artembilan: We could, but why make our code uglier.
And no `peek` is not for debug.
See its JavaDocs: it is like `forEach()`, but in the middle of the stream process.

cppwfs (approved):
LGTM

Thanks for the quick reply!
And thanks for working on this!

spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContext.java
cppwfs: Understood.   Thank you.

spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
cppwfs: Understood!   Thank you

PR #10208, opened 2025-07-17
Add NullaWay to package-info for o.s.i.config package

cppwfs:
- Add `@org.jspecify.annotations.NullAway` to package-info.java
- Apply `@Nullable`, `@SuppressWarnings("NullAway.Init)` as need. Tweak code as necessary in locations where NullAway can not discern that the null condition is already mitigated

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/AbstractSimpleMessageHandlerFactoryBean.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/AbstractSimpleMessageHandlerFactoryBean.java
@@ -26,7 +26,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
 
-import org.springframework.aop.framework.Advised;
 import org.springframework.aop.framework.AopProxyUtils;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
@@ -95,7 +94,8 @@ public abstract class AbstractSimpleMessageHandlerFactoryBean<H extends MessageH
 	@SuppressWarnings("NullAway.Init")
 	private String beanName;
 
-	private @Nullable ApplicationEventPublisher applicationEventPublisher;
+	@SuppressWarnings("NullAway.Init")
+	private ApplicationEventPublisher applicationEventPublisher;
 
 	private @Nullable DestinationResolver<MessageChannel> channelResolver;
 
@@ -208,16 +208,16 @@ public abstract class AbstractSimpleMessageHandlerFactoryBean<H extends MessageH
 			JavaUtils.INSTANCE
 					.acceptIfCondition(this.handler instanceof ApplicationContextAware && this.applicationContext != null,
 							this.applicationContext,
-							context -> ((ApplicationContextAware) Objects.requireNonNull(this.handler)).setApplicationContext(this.applicationContext))
+							context -> ((ApplicationContextAware) this.handler).setApplicationContext(this.applicationContext))
 					.acceptIfCondition(this.handler instanceof BeanFactoryAware && getBeanFactory() != null,
 							getBeanFactory(),
-							factory -> ((BeanFactoryAware) Objects.requireNonNull(this.handler)).setBeanFactory(factory))
+							factory -> ((BeanFactoryAware) this.handler).setBeanFactory(factory))
 					.acceptIfCondition(this.handler instanceof BeanNameAware && this.beanName != null, this.beanName,
-							name -> ((BeanNameAware) Objects.requireNonNull(this.handler)).setBeanName(this.beanName))
+							name -> ((BeanNameAware) this.handler).setBeanName(this.beanName))
 					.acceptIfCondition(this.handler instanceof ApplicationEventPublisherAware
 									&& this.applicationEventPublisher != null,
 							this.applicationEventPublisher,
-							publisher -> ((ApplicationEventPublisherAware) Objects.requireNonNull(Objects.requireNonNull(this.handler)))
+							publisher -> ((ApplicationEventPublisherAware) this.handler)
 									.setApplicationEventPublisher(publisher));
 			configureOutputChannelIfAny();
 			Object actualHandler = extractTarget(this.handler);
@@ -232,7 +232,7 @@ public abstract class AbstractSimpleMessageHandlerFactoryBean<H extends MessageH
 							this.async,
 							asyncValue -> ((AbstractMessageProducingHandler) handlerToConfigure).setAsync(asyncValue))
 					.acceptIfCondition(this.handler instanceof Orderable && this.order != null,
-							this.order, theOrder -> ((Orderable) Objects.requireNonNull(this.handler)).setOrder(theOrder));
+							this.order, theOrder -> ((Orderable) this.handler).setOrder(theOrder));
 			this.initialized = true;
 		}
 		finally {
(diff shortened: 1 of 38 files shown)


artembilan:
Pay attention how those annotations are converted GH user handles:
image
That's why I prefer to use back-ticks even in commit messages for code snippets to avoid calling those GH users for nothing 😄

cppwfs:
Thank you for catching that!    Updated the comment.

artembilan:
>  Updated the comment.

It doesn't help any more.
You have already invited those GH users to our PR review 😉 

I just mean the reasoning behind my preference to use code snippet tools in commit messages as well.

cppwfs:
Understood.

spring-integration-core/src/main/java/org/springframework/integration/config/AbstractSimpleMessageHandlerFactoryBean.java
```diff
@@ -71,29 +73,32 @@ public abstract class AbstractSimpleMessageHandlerFactoryBean<H extends MessageH
 
 	private final Lock initializationMonitor = new ReentrantLock();
 
+	@SuppressWarnings("NullAway.Init")
 	private BeanFactory beanFactory;
 
-	private H handler;
+	private @Nullable H handler;
```
artembilan: I think the main point of this class is to produce exactly `handler`.
So, please, consider `@SuppressWarnings("NullAway.Init")` instead.

spring-integration-core/src/main/java/org/springframework/integration/config/AbstractSimpleMessageHandlerFactoryBean.java
```diff
@@ -312,7 +317,7 @@ public boolean isSingleton() {
 		return true;
 	}
 
-	private static Object extractTarget(Object object) {
+	private static @Nullable Object extractTarget(@Nullable Object object) {
```
artembilan: I think this method has to be changed to be non-nullable.
And we should have something like in the body:
```
return Objects.requireNonNullElse(AopProxyUtils.getSingletonTarget(object), object);
```

I doubt that we need recursion here.
But I might be wrong if some tests would fail.
Forth checking anyway!

spring-integration-core/src/main/java/org/springframework/integration/config/CorrelationStrategyFactoryBean.java
```diff
@@ -39,8 +39,10 @@
  */
 public class CorrelationStrategyFactoryBean implements FactoryBean<CorrelationStrategy>, InitializingBean {
 
+	@SuppressWarnings("NullAway.Init")
 	private Object target;
 
+	@SuppressWarnings("NullAway.Init")
 	private String methodName;
```
artembilan, line 47: This can be null according to the class logic.
While for the `target` we need to ensure that it is not-null.

spring-integration-core/src/main/java/org/springframework/integration/config/DefaultConfiguringBeanFactoryPostProcessor.java
```diff
@@ -79,6 +81,7 @@ public class DefaultConfiguringBeanFactoryPostProcessor implements BeanDefinitio
 
 	private static final Set<Integer> REGISTRIES_PROCESSED = new HashSet<>();
 
+	@Nullable
 	private static final Class<?> XPATH_CLASS;
```
artembilan: Please, put the `@Nullable` just before the `Class<?>`.
While it works with JSpecify, but better to follow its code style recommendation.

spring-integration-core/src/main/java/org/springframework/integration/config/SpelFunctionFactoryBean.java
```diff
@@ -42,9 +44,9 @@ public class SpelFunctionFactoryBean implements FactoryBean<Method>, Initializin
 
 	private final String functionMethodSignature;
 
-	private String functionName;
+	private @Nullable String functionName;
 
-	private Method method;
+	private @Nullable Method method;
```
artembilan: The `@SuppressWarnings("NullAway.Init")` on both properties.
One of them is from `setBeanName()` and another from the `afterPropertiesSet()`.
I doubt we should pursue the case when this class is not used as a bean in the code.

artembilan (changes requested):
Very cool!
Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/AbstractSimpleMessageHandlerFactoryBean.java
```diff
@@ -317,12 +314,14 @@ public boolean isSingleton() {
...
 	}
 
-	private static @Nullable Object extractTarget(@Nullable Object object) {
+	private static Object extractTarget(Object object) {
 		if (!(object instanceof Advised)) {
 			return object;
 		}
 		else {
-			return extractTarget(AopProxyUtils.getSingletonTarget(object));
+//			return extractTarget(Objects.requireNonNull(AopProxyUtils.getSingletonTarget(object)));
```
artembilan: Why do we need this comment here?

spring-integration-core/src/main/java/org/springframework/integration/config/AbstractSimpleMessageHandlerFactoryBean.java
```diff
@@ -71,29 +73,33 @@ public abstract class AbstractSimpleMessageHandlerFactoryBean<H extends MessageH
...
+	private @Nullable String componentName;
 
+	@SuppressWarnings("NullAway.Init")
 	private ApplicationContext applicationContext;
 
+	@SuppressWarnings("NullAway.Init")
 	private String beanName;
 
-	private ApplicationEventPublisher applicationEventPublisher;
+	private @Nullable ApplicationEventPublisher applicationEventPublisher;
```
artembilan: I think this one as `SuppressWarnings("NullAway.Init")` as well because of `ApplicationEventPublisherAware` hook from the container.

spring-integration-core/src/main/java/org/springframework/integration/config/AbstractSimpleMessageHandlerFactoryBean.java
```diff
@@ -230,7 +232,7 @@ protected final H createHandlerInternal() {
 							this.async,
 							asyncValue -> ((AbstractMessageProducingHandler) handlerToConfigure).setAsync(asyncValue))
 					.acceptIfCondition(this.handler instanceof Orderable && this.order != null,
-							this.order, theOrder -> ((Orderable) this.handler).setOrder(theOrder));
+							this.order, theOrder -> ((Orderable) Objects.requireNonNull(this.handler)).setOrder(theOrder));
```
artembilan: Why of my ideas to mark `this.handler` with a `@SuppressWarnings("NullAway.Init")` is to avoid this extra `Objects.requireNonNull()`.
Any chances to ensure that we don't need them in this method?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationManagementConfigurer.java
```diff
@@ -182,6 +185,7 @@ private void setupObservationRegistry() {
 		}
 	}
 
+	@SuppressWarnings("NullAway")  // is not called if the metricsCaptor is null in  afterSingletonsInstantiated.
```
artembilan: Make the call to this method like this instead:
```
		MetricsCaptor metricsCaptorToUse = obtainMetricsCaptor();
		if (metricsCaptorToUse != null) {
			registerComponentGauges(metricsCaptorToUse);
		}
```

PR #10210, opened 2025-07-18
Add ability for a user to set mqtt's `quiescentTimeout` for forceable shutdowns

### LINKED ISSUE #10095: MQTT: Configure quiesceTimeout in disconnectForcibly()
### Discussed in https://github.com/spring-projects/spring-integration/discussions/10093

<sup>Originally posted by **leonard2901** June 10, 2025</sup>
Hello,
I use spring-integration-mqtt (phao3) to communicate with a Mosquitto instance. For integration testing, I start Mosquitto using Testcontainers.

This generally works fine, but the tests encounter a 30-second timeout when the client attempts to disconnect, as the container has already been stopped. I have looked through the code, especially the `doStop()` method in  `MqttPahoMessageDrivenChannelAdapter`. There `this.client.disconnectForcibly(getDisconnectCompletionTimeout());` is called. In this overload the default `QUIESCE_TIMEOUT` of 30 seconds is used.

In my tests I would like to just disconnect immediately and discard any potentially pending messages.

Is there a way to configure the quiesceTimeout or any other way to disconnect without a timeout?

Any help is appreciated! :)</div>

cppwfs:
* Mqttv5 updated to  also support this feature
* Fixes: https://github.com/spring-projects/spring-integration/issues/10095

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
@@ -50,6 +50,7 @@ import org.springframework.util.Assert;
  * @author Mikhail Polivakha
  * @author Artem Vozhdayenko
  * @author Jiri Soucek
+ * @author Glenn Renfro
  *
  * @since 4.0
  *
@@ -58,6 +59,8 @@ import org.springframework.util.Assert;
 @IntegrationManagedResource
 public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends MessageProducerSupport
 		implements ApplicationEventPublisherAware, ClientManager.ConnectCallback {
+
+	public static final Long QUIESCENT_TIMEOUT = 30_000L;
 
 	protected final Lock topicLock = new ReentrantLock(); // NOSONAR
 
@@ -72,6 +75,8 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends Mess
 	private long completionTimeout = ClientManager.DEFAULT_COMPLETION_TIMEOUT;
 
 	private long disconnectCompletionTimeout = ClientManager.DISCONNECT_COMPLETION_TIMEOUT;
+
+	private long quiescentTimeout = QUIESCENT_TIMEOUT;
 
 	private boolean manualAcks;
 
@@ -197,6 +202,20 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends Mess
 
 	protected long getDisconnectCompletionTimeout() {
 		return this.disconnectCompletionTimeout;
+	}
+
+	/**
+	 * Set the quiescentTimeout timeout when disconnecting.
+	 * Default is 30,000 milliseconds.
+	 * @param quiescentTimeout The timeout.
+	 * @since 7.0.0
+	 */
+	public void setQuiescentTimeout(long quiescentTimeout) {
+		this.quiescentTimeout = quiescentTimeout;
+	}
+
+	public long getQuiescentTimeout() {
+		return this.quiescentTimeout;
 	}
 
 	@Override
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
@@ -60,8 +60,6 @@ import org.springframework.util.Assert;
 public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends MessageProducerSupport
 		implements ApplicationEventPublisherAware, ClientManager.ConnectCallback {
 
-	public static final Long QUIESCENT_TIMEOUT = 30_000L;
-
 	protected final Lock topicLock = new ReentrantLock(); // NOSONAR
 
 	private final String url;
@@ -76,7 +74,7 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends Mess
 
 	private long disconnectCompletionTimeout = ClientManager.DISCONNECT_COMPLETION_TIMEOUT;
 
-	private long quiescentTimeout = QUIESCENT_TIMEOUT;
+	private long quiescentTimeout = ClientManager.QUIESCENT_TIMEOUT;
 
 	private boolean manualAcks;
 
@@ -206,7 +204,7 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends Mess
 
 	/**
 	 * Set the quiescentTimeout timeout when disconnecting.
-	 * Default is 30,000 milliseconds.
+	 * Default is {@link ClientManager#QUIESCENT_TIMEOUT} milliseconds.
 	 * @param quiescentTimeout The timeout.
 	 * @since 7.0.0
 	 */
@@ -214,7 +212,7 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends Mess
 		this.quiescentTimeout = quiescentTimeout;
 	}
 
-	public long getQuiescentTimeout() {
+	protected long getQuiescentTimeout() {
 		return this.quiescentTimeout;
 	}
 
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -37,6 +37,7 @@ import org.eclipse.paho.mqttv5.common.MqttException;
 import org.eclipse.paho.mqttv5.common.MqttMessage;
 import org.eclipse.paho.mqttv5.common.MqttSubscription;
 import org.eclipse.paho.mqttv5.common.packet.MqttProperties;
+import org.eclipse.paho.mqttv5.common.packet.MqttReturnCode;
 
 import org.springframework.beans.factory.BeanCreationException;
 import org.springframework.context.ApplicationEventPublisher;
@@ -81,6 +82,7 @@ import org.springframework.util.ObjectUtils;
  * @author Lucas Bowler
  * @author Artem Vozhdayenko
  * @author Matthias Thoma
+ * @author Glenn Renfro
  *
  * @since 5.5.5
  *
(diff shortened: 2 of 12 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -59,6 +60,8 @@
 public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends MessageProducerSupport
 		implements ApplicationEventPublisherAware, ClientManager.ConnectCallback {
 
+	public static final Long QUIESCENT_TIMEOUT = 30_000L;
```
artembilan: Move this into a `ClientManager` alongside with the `DISCONNECT_COMPLETION_TIMEOUT`.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -199,6 +204,20 @@ protected long getDisconnectCompletionTimeout() {
 		return this.disconnectCompletionTimeout;
 	}
 
+	/**
+	 * Set the quiescentTimeout timeout when disconnecting.
+	 * Default is 30,000 milliseconds.
```
artembilan:
```suggestion
	 * Default {@value ClientManager#QUIESCENT_TIMEOUT} milliseconds.
```

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -199,6 +204,20 @@ protected long getDisconnectCompletionTimeout() {
...
+	 * @param quiescentTimeout The timeout.
+	 * @since 7.0.0
+	 */
+	public void setQuiescentTimeout(long quiescentTimeout) {
+		this.quiescentTimeout = quiescentTimeout;
+	}
+
+	public long getQuiescentTimeout() {
```
artembilan: ``
```suggestion
	protected long getQuiescentTimeout() {
```

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/BackToBackAdapterTests.java
```diff
@@ -123,6 +128,33 @@ public void testSingleTopic() {
...
+		adapter.setDefaultTopic("mqtt-foo");
+		adapter.setBeanFactory(mock(BeanFactory.class));
+		adapter.afterPropertiesSet();
+		adapter.start();
+		MqttPahoMessageDrivenChannelAdapter inbound =
+				new MqttPahoMessageDrivenChannelAdapter(MosquittoContainerTest.mqttUrl(), "si-test-in", "mqtt-foo");
+		QueueChannel outputChannel = new QueueChannel();
+		inbound.setOutputChannel(outputChannel);
+		inbound.setTaskScheduler(taskScheduler);
+		inbound.setQuiescentTimeout(QUIESCENT_TIMEOUT);
```
artembilan: Let's apply it to all existing tests instead!

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackTests.java
```diff
@@ -153,6 +154,23 @@ public void testSharedTopicMqttv5Interaction() {
 		assertThat(receive.getPayload()).isEqualTo(testPayload);
 	}
 
+	@Test
+	void testSharedTopicMqttv5InteractionQuiescentTimeout() {
+		this.mqttv5MessageDrivenChannelAdapter.addTopic("$share/group/testTopicq");
+		this.mqttv5MessageDrivenChannelAdapter.setQuiescentTimeout(2000);
+		this.mqttv5MessageDrivenChannelAdapter.setDisconnectCompletionTimeout(2000);
```
artembilan: Let's apply this to existing tests instead!

artembilan (changes requested):
Also apply this feature in the `AbstractMqttClientManager` implementations.
Plus mention a new property in the `whats-new.adoc`.

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -296,7 +298,8 @@ protected void doStop() {
 
 				}
 				if (getClientManager() == null) {
-					this.mqttClient.disconnectForcibly(getDisconnectCompletionTimeout());
+					this.mqttClient.disconnectForcibly(ClientManager.QUIESCENT_TIMEOUT, getDisconnectCompletionTimeout(),
```
artembilan: Did you mean `getQuiescentTimeout()` as you do in the `MqttPahoMessageDrivenChannelAdapter` for MQTTv3?

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/BackToBackAdapterTests.java
```diff
@@ -396,6 +389,14 @@ public void onApplicationEvent(MqttSubscribedEvent event) {
 
 	}
 
+	private void initializeInboundAdapter(MqttPahoMessageDrivenChannelAdapter inbound, QueueChannel outputChannel) {
```
artembilan: `static`

artembilan (changes requested):
See `Mqttv3ClientManager` and `Mqttv5ClientManager` `stop()`.
Such a `quiesceTimeout` has to be implemented over there as well.
Plus `whats-new.adoc` entry.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
cppwfs: DOH!

artembilan:
Thanks.
See, please, my general comment to the previous review(s).

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -44,3 +44,8 @@ See xref:redis.adoc#redis-lock-registry[Redis Lock Registry] for more informatio
 === Hazelcast Changes
 
 Previously deprecated classes in the `spring-integation-hazelcast` module, such as `LeaderInitiator`, `HazelcastMembershipListener`, `HazelcastLocalInstanceRegistrar` and `HazelcastLockRegistry`, are now removed due to not supported CP-subsystem in Hazelcast library for Open Source.
+
+[[x7.0-mqtt-changes]]
+=== MQTT Changes
+
+The `disconnectForcibly` method in `MqttPahoMessageDrivenChannelAdapter`, `Mqttv5PahoMessageDrivenChannelAdapter`, `Mqttv3ClientManager`, and `Mqttv5ClientManager` now supports `getQuiescentTimeout` in addition to the existing `disconnectCompletionTimeout`.
```
artembilan: That’s not quite true. That method had it supported for years in the target client. And in fact that is not a method of our classes. Please , try to be more specific. 
let’s say something like those classes now provide such an option which is propagated down to the respective client API.
Also pay attention to the typo of new property. It is really not `get…` copy/paste artifact I guess 😉
Also link to the target chapter , please.

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -48,4 +48,5 @@ Previously deprecated classes in the `spring-integation-hazelcast` module, such
 [[x7.0-mqtt-changes]]
 === MQTT Changes
 
-The `disconnectForcibly` method in `MqttPahoMessageDrivenChannelAdapter`, `Mqttv5PahoMessageDrivenChannelAdapter`, `Mqttv3ClientManager`, and `Mqttv5ClientManager` now supports `getQuiescentTimeout` in addition to the existing `disconnectCompletionTimeout`.
+The `disconnectForcibly` method in `MqttPahoMessageDrivenChannelAdapter`, `Mqttv5PahoMessageDrivenChannelAdapter`, `Mqttv3ClientManager`, and `Mqttv5ClientManager` now expose the option to configure `quiescentTimeout`, which is available through the https://eclipse.dev/paho/files/javadoc/org/eclipse/paho/client/mqttv3/IMqttAsyncClient.html[IMqttAsyncClient] client.
```
artembilan: Sorry, I still disagree with this sentence.
How about like this:
```
```
The JavaDoc to MQTT v3 client is not quit right here since we also do the same for MQTT v5 client interaction.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -48,4 +48,5 @@ Previously deprecated classes in the `spring-integation-hazelcast` module, such
...
+See https://docs.spring.io/spring-integration/api/org/springframework/integration/mqtt/core/package-summary.html[MQTT Core Java Docs] for more information.
```
artembilan: This is not right.
We have talked about the target MQTT chapter of this reference manual, but you point to JavaDoc of the package.
It has to be like this:
```
```

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/whats-new.adoc
cppwfs: Referring to this document does not help the user, in that it does not contain any information about `disconnectForcibly`.   Do we want to add `disconnectForcibly` documentation there?  Without this documentation, it seems useful to point the user to docs that have information about the API.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
cppwfs: Should we add MQTT V3 and V5 links.   OR remove the V3?

artembilan: Remove altogether.
We may mention something like `See their JavaDocs for more information.`, though.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
artembilan: It does help because it leads to more details about MQTT channel adapters. Even if we don't mention there such a low-level function like `disconnectForcibly`, it is still good follow up from this short sentence in the `What's new`.
Yes, adding some info about `disconnectForcibly` and related adapter properties into the target chapter would be great.
Let me know if you willing to add it!
Or let's go with this final fix for the `What's new` and merge!

cppwfs: Let's go with the final fix. We can revisit the docs later on, in that our javadocs cover the updates satisfactorly.

artembilan (changes requested):
Looks like PR build is failing in the test related to the change.
```
 > Task :spring-integration-mqtt:test
ClientManagerBackToBackTests > testSameV3ClientIdWorksForPubAndSub() FAILED
    java.lang.AssertionError at ClientManagerBackToBackTests.java:132
```
Any thoughts?
Thanks

PR #10211, opened 2025-07-18
GH-10083: Apply Nullability to test-support module

artembilan:

* Add missed `package-info.java` to packages of the `test-support` module
* Add `@NullMarked` and fix all the reported nullability failures
* Move `hashCode()` and `equals()` in the `OnlyOnceTrigger` to the end of class for more logical class structure.

### DIFF
--- a/spring-integration-test-support/src/main/java/org/springframework/integration/test/condition/package-info.java
+++ b/spring-integration-test-support/src/main/java/org/springframework/integration/test/condition/package-info.java
@@ -1,0 +1,5 @@
+/**
+ * JUnit Jupiter conditions.
+ */
+@org.jspecify.annotations.NullMarked
+package org.springframework.integration.test.condition;
--- a/spring-integration-test-support/src/main/java/org/springframework/integration/test/mail/package-info.java
+++ b/spring-integration-test-support/src/main/java/org/springframework/integration/test/mail/package-info.java
@@ -1,4 +1,0 @@
-/**
- * Provides testing utils for mail support.
- */
-package org.springframework.integration.test.mail;
--- a/spring-integration-test-support/src/main/java/org/springframework/integration/test/util/OnlyOnceTrigger.java
+++ b/spring-integration-test-support/src/main/java/org/springframework/integration/test/util/OnlyOnceTrigger.java
@@ -21,6 +21,8 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.jspecify.annotations.Nullable;
+
 import org.springframework.scheduling.Trigger;
 import org.springframework.scheduling.TriggerContext;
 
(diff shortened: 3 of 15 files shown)


spring-integration-test-support/src/main/java/org/springframework/integration/test/condition/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * JUnit Jupiter conditions.
```
cppwfs, line 2: Should this be written as , "Null checks for JUnit Jupiter condition package"

spring-integration-test-support/src/main/java/org/springframework/integration/test/mail/package-info.java
```diff
@@ -1,4 +0,0 @@
-/**
```
cppwfs, line 1: Asking:
Why was this package-info removed?

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/OnlyOnceTrigger.java
```diff
@@ -61,6 +63,23 @@ public Instant nextExecution(TriggerContext triggerContext) {
...
 	}
 
+	public void reset() {
+		this.latch = new CountDownLatch(1);
+		this.hasRun.set(false);
+	}
+
+	public void await() {
+		try {
+			if (!this.latch.await(10000, TimeUnit.MILLISECONDS)) { // NOSONAR magic number
```
cppwfs: We can remove the `NOSONAR ` comment

cppwfs (changes requested):
Looks good!

Just one question and a couple of nitpicks.

spring-integration-test-support/src/main/java/org/springframework/integration/test/condition/package-info.java
artembilan: This is a `package-info.java` file for the whole `coondition` package.
I tried to be as generic for those classes in this package as possible.
However I'm not sure that any of them deals with null checks.

What do you mean, please, with such a sentence as package info?

spring-integration-test-support/src/main/java/org/springframework/integration/test/util/OnlyOnceTrigger.java
artembilan:
```suggestion
			if (!this.latch.await(10000, TimeUnit.MILLISECONDS)) {
```

spring-integration-test-support/src/main/java/org/springframework/integration/test/condition/package-info.java
cppwfs: Ok I understand.   
I was reading this comment as if it was only to be applied to the `@org.jspecify.annotations.NullMarked` addition to the package-info.   This comment is to be applied to the package declaration on on line 5.

artembilan: I see.
we just follow existing pattern from where we took such an entity: https://github.com/spring-projects/spring-framework/blob/main/spring-beans/src/main/java/org/springframework/beans/support/package-info.java

spring-integration-test-support/src/main/java/org/springframework/integration/test/mail/package-info.java
artembilan: There is no content in this `mail` package at all.
So, no need in this directory and its dangling `package-info`.

Looks like I have missed to mention that in the commit message 😄

cppwfs (approved):

cppwfs:
LGTM!

cppwfs:
Rebased, Squashed Merged

Thanks for the great work!

PR #10212, opened 2025-07-18
GH-10058: Switch to Jackson 3 as default and deprecate Jackson 2

anthologia:
Related to: #10058

* Switch to Jackson 3 classes as default
* Deprecate Jackson 2 classes
* Update tests and docs
---
I plan to address the `spring-integration-kafka` module in another PR, once Spring Kafka's Jackson 3 support is complete.


spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonObjectMapper.java
```diff
@@ -59,8 +62,13 @@ public class JacksonJsonObjectMapper extends AbstractJacksonJsonObjectMapper<Jso
 	private final ObjectMapper objectMapper;
 
 	public JacksonJsonObjectMapper() {
+		SimpleModule simpleModule = new SimpleModule()
```
anthologia: Without this, NPE occurs when calling `ObjectMapper.valueToTree(UUID)`, which is used for `MessageHeader`'s ID.

This is a workaround for Jackson 3.0.0-rc5 bug reported by my teammate. Once Jackson 3.0.0-rc6 is released, this UUID serializer should be removed.

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonObjectMapper.java
```diff
@@ -59,8 +62,13 @@ public class JacksonJsonObjectMapper extends AbstractJacksonJsonObjectMapper<Jso
...
+				.addSerializer(UUID.class, new UUIDJsonSerializer());
+
 		this.objectMapper = JsonMapper.builder()
+				.configure(DeserializationFeature.FAIL_ON_TRAILING_TOKENS, false)
```
anthologia: Jackson 3 enables `FAIL_ON_TRAILING_TOKENS` by default.

But when parsing `{"payload":"value","headers":{...}}`, we reads payload first using `JsonParser`, but Jackson detects the remaining ,`"headers":{...}` as `trailing tokens` and throws an exception.

spring-integration-http/src/main/java/org/springframework/integration/http/inbound/HttpRequestHandlingEndpointSupport.java
```diff
@@ -513,24 +519,28 @@ protected Object extractRequestBody(ServletServerHttpRequest request) throws IOE
...
-					? genericConverter.canRead(targetType, null, contentType)
-					: (converter.canRead(targetClass, contentType))) {
 
-				if (genericConverter != null) {
-					return genericConverter.read(targetType, null, request);
-				}
-				else {
-					return converter.read((Class) targetClass, request);
-				}
+			SmartHttpMessageConverter<?> smartConverter =
```
anthologia, line 530: `JacksonJsonHttpMessageConverter` now implements `SmartHttpMessageConverter` instead of `GenericHttpMessageConverter`.

artembilan (changes requested):
Something is off with Redis:
```
RedisMessageGroupStoreTests > testJsonSerialization() FAILED
    org.opentest4j.AssertionFailedError at RedisMessageGroupStoreTests.java:416
```
Feels like your change is related.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/aot/CoreRuntimeHints.java
```diff
@@ -104,11 +104,11 @@ public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader)
 		}
 
 		reflectionHints.registerType(
-				TypeReference.of("org.springframework.integration.json.JsonPropertyAccessor$ComparableJsonNode"),
+				TypeReference.of("org.springframework.integration.json.JacksonPropertyAccessor$ComparableJsonNode"),
 				MemberCategory.INVOKE_PUBLIC_METHODS);
 
 		reflectionHints.registerType(
-				TypeReference.of("org.springframework.integration.json.JsonPropertyAccessor$ArrayNodeAsList"),
+				TypeReference.of("org.springframework.integration.json.JacksonPropertyAccessor$ArrayNodeAsList"),
```
artembilan, line 115: Since we talk about about GraalVM and native images, no one said that we should drop off Jackson 2 support from there.
Even if it is deprecated, it is still OK to have support for that.
So, please, bring old entries back with respective `removal` suppression, and add new entries.

spring-integration-core/src/main/java/org/springframework/integration/config/ConverterRegistrar.java
```diff
@@ -72,7 +74,10 @@ private void registerConverters(GenericConversionService conversionService) {
 						.values()
 						.stream().map(IntegrationConverterRegistration::converter)
 						.collect(Collectors.toSet());
-		if (JacksonPresent.isJackson2Present()) {
+		if (JacksonPresent.isJackson3Present()) {
+			converters.add(new JsonNodeWrapperConverter());
+		}
+		else if (JacksonPresent.isJackson2Present()) {
```
artembilan: I think this is similar to the previous one.
We do not give a choice to end-user here.
Better to have both converters registered for time being until we remove those deprecated classes.
But yeah, Jackson 3 has to be first as priority.

spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
```diff
@@ -247,6 +248,9 @@ protected int obtainSizeIfPossible(Iterable<?> iterable) {
 		if (iterable instanceof Collection<?> collection) {
 			return collection.size();
 		}
+		else if (JacksonPresent.isJackson3Present() && JacksonTreeNodeHelper.isNode(iterable)) {
+			return JacksonTreeNodeHelper.nodeSize(iterable);
+		}
 		else if (JacksonPresent.isJackson2Present() && JacksonNodeHelper.isNode(iterable)) {
 			return JacksonNodeHelper.nodeSize(iterable);
```
artembilan, line 256: Let's deprecate this one, so we won't forget to remove it altogether with Jackson 2 support!

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonObjectMapper.java
artembilan: Good. 
Please, raise a new GH issue, so we don't forget to remove this when we got a fix from Jackson.

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonObjectMapper.java
artembilan: Fine with me, but let's use `disable()` API instead!

spring-integration-core/src/main/java/org/springframework/integration/support/json/UUIDJsonSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Simple {@link StdSerializer} extension to represent a {@link UUID} object in the
+ * target JSON as a plain string.
+ *
+ * @author Jooyoung Pyoung
+ *
+ * @since 7.0
+ */
+public class UUIDJsonSerializer extends StdSerializer<UUID> {
```
artembilan: M-m-m. Why do we need this class to be `public` if we are going to remove it with the next Jackson 3 RC?
Consider to move it into the `JacksonJsonObjectMapper` as an inner `private static class` .

spring-integration-core/src/test/java/org/springframework/integration/support/json/JacksonJsonObjectMapperTests.java
```diff
@@ -82,10 +82,10 @@ void compareAutoDiscoveryVsManualModules() {
 		Set<String> collectedModuleNames = getModuleNames(collectWellKnownModulesIfAvailable());
 
 		Set<String> autoModuleNames = getModuleNames(mapper.getObjectMapper().getRegisteredModules());
-		assertThat(autoModuleNames).isEqualTo(collectedModuleNames);
+		assertThat(autoModuleNames).containsAll(collectedModuleNames);
 
 		Set<String> manualModuleNames = getModuleNames(manualMapper.getRegisteredModules());
-		assertThat(manualModuleNames).isEqualTo(collectedModuleNames);
+		assertThat(manualModuleNames).containsAll(collectedModuleNames);
```
artembilan: Any justification for this change, please?

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
```diff
@@ -445,11 +445,11 @@ public enum Mark {
...
 		}
 
-		private final String filePath;
+		private String filePath;
 
-		private final Mark mark;
+		private Mark mark;
 
-		private final long lineCount;
+		private long lineCount;
```
artembilan: I guess you did this to satisfy Jackson 3 on deserialization.
Isn't there a way to make it to use our specific ctor?
For example, via `@JsonCreator`?
See, `DelayedMessageWrapper` and `MessageHistory`.
With that we might need to deprecate that default `FileMarker` ctor.

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/inbound/WebSocketInboundChannelAdapter.java
```diff
@@ -82,7 +83,14 @@ public class WebSocketInboundChannelAdapter extends MessageProducerSupport
...
 		this.defaultConverters.add(new ByteArrayMessageConverter());
-		if (JacksonPresent.isJackson2Present()) {
+		if (JacksonPresent.isJackson3Present()) {
+			DefaultContentTypeResolver resolver = new DefaultContentTypeResolver();
+			resolver.setDefaultMimeType(MimeTypeUtils.APPLICATION_JSON);
+			JacksonJsonMessageConverter converter = new JacksonJsonMessageConverter();
+			converter.setContentTypeResolver(resolver);
+			this.defaultConverters.add(converter);
+		}
+		else if (JacksonPresent.isJackson2Present()) {
```
artembilan: We talk here about defaults.
And our choice by default is only Jackson 3.
You still can use here Jackson 2 by injecting it itself.
See respective `setMessageConverters()`.
This in unlike in other place for a node converter, when there is no end-user hook for custom injection.

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/outbound/WebSocketOutboundMessageHandler.java
```diff
@@ -56,7 +57,14 @@ public class WebSocketOutboundMessageHandler extends AbstractMessageHandler {
...
 		this.defaultConverters.add(new ByteArrayMessageConverter());
-		if (JacksonPresent.isJackson2Present()) {
+		if (JacksonPresent.isJackson3Present()) {
+			DefaultContentTypeResolver resolver = new DefaultContentTypeResolver();
+			resolver.setDefaultMimeType(MimeTypeUtils.APPLICATION_JSON);
+			JacksonJsonMessageConverter converter = new JacksonJsonMessageConverter();
+			converter.setContentTypeResolver(resolver);
+			this.defaultConverters.add(converter);
+		}
+		else if (JacksonPresent.isJackson2Present()) {
```
artembilan: DITTO

src/reference/antora/modules/ROOT/pages/overview.adoc
```diff
@@ -429,7 +429,7 @@ See xref:configuration/annotations.adoc[Annotation Support] for more information
 
 Most classes in Spring Integration (unless stated otherwise) must be declared as beans in the application context and as singletons.
 That means instances of these classes are thread-safe and their lifecycle and connection with other components are managed by Spring dependency injection container.
-The utility and builder classes (`JacksonJsonUtils, `MessageBuilder`, `ExpressionEvalMap`, `IntegrationReactiveUtils` etc.) can be used directly in the Java code.
+The utility and builder classes (`JacksonMessagingUtils, `MessageBuilder`, `ExpressionEvalMap`, `IntegrationReactiveUtils` etc.) can be used directly in the Java code.
```
artembilan: I see that it is an existing bug, but let's fix it since we are already here.
There is missed back tick after `JacksonMessagingUtils` 😄

src/reference/antora/modules/ROOT/pages/redis.adoc
```diff
@@ -348,20 +348,20 @@ The preceding example is a bean configuration, and it expects a `RedisConnection
 By default, the `RedisMessageStore` uses Java serialization to serialize the message.
 However, if you want to use a different serialization technique (such as JSON), you can provide your own serializer by setting the `valueSerializer` property of the `RedisMessageStore`.
 
-Starting with version 4.3.10, the Framework provides Jackson serializer and deserializer implementations for `Message` instances and `MessageHeaders` instances -- `MessageJacksonDeserializer` and `MessageHeadersJacksonSerializer`, respectively.
+Starting with version 4.3.10, the Framework provides Jackson serializer and deserializer implementations for `Message` instances and `MessageHeaders` instances -- `MessageJsonDeserializer` and `MessageHeadersJsonSerializer`, respectively.
```
artembilan: Let's remove the version mentioning in this sentence since the reality does not reflect its relevance any more!

src/reference/antora/modules/ROOT/pages/spel.adoc
```diff
@@ -17,10 +17,10 @@ This is how you can access the `Message` headers and payload properties.
 == SpEL Evaluation Context Customization
 
 Starting with Spring Integration 3.0, you can add additional `PropertyAccessor` instances to the SpEL evaluation contexts used by the framework.
-The framework provides the (read-only) `JsonPropertyAccessor`, which you can use to access fields from a `JsonNode` or JSON in a `String`.
+The framework provides the (read-only) `JacksonPropertyAccessor`, which you can use to access fields from a `JsonNode` or JSON in a `String`.
 You can also create your own `PropertyAccessor` if you have specific needs.
 
-Starting with version 6.4, the `JsonIndexAccessor` implementation is provided that knows how to read indexes from JSON arrays, using Jackson's `ArrayNode` API.
+Starting with version 6.4, the `JacksonIndexAccessor` implementation is provided that knows how to read indexes from JSON arrays, using Jackson's `ArrayNode` API.
```
artembilan: Same here, please.

artembilan (changes requested):
And since you have touched docs as well, please, consider to add a new paragraph in the `whats-new.adoc` as a part of the `== New Components` section.

In general, this is great and look like the last step for that issue to be closed.

Thank you for much for very cool contribution!

spring-integration-core/src/test/java/org/springframework/integration/support/json/JacksonJsonObjectMapperTests.java
anthologia: Because of UUID Serializer, `autoModuleNames` includes one extra module. So, `containsAll()` verifies expected modules are present.

spring-integration-file/src/main/java/org/springframework/integration/file/splitter/FileSplitter.java
anthologia: To use `@JsonCreator`, file module need to add Jackson dependency..
I thought it feels like too much for a simple serialization fix.
But I'm fine. Up to you ! 😁

artembilan: That is just an annotation , and it is OK to not have a dependency in the end: the target project won’t fail.
Just make the Jackson one in this module optional!

spring-integration-core/src/test/java/org/springframework/integration/support/json/JacksonJsonObjectMapperTests.java
artembilan: I just saw that Jackson 3 rc6 is out. Perhaps we can revert all the UUID changes away?
The automatic update should happen this evening or tomorrow  if you won’t do that earlier.

artembilan:
Here is an update to Jackson you have promised: https://github.com/spring-projects/spring-integration/pull/10231.
Although I don’t see the UUID bug you have mentioned…

artembilan:
Looks like something is off with that latest Jackson: 
```

JacksonJsonObjectMapperTests > testJodaWithJodaModule() FAILED
    org.opentest4j.AssertionFailedError at JacksonJsonObjectMapperTests.java:227

JacksonJsonObjectMapperTests > compareAutoDiscoveryVsManualModules() FAILED
    org.opentest4j.AssertionFailedError at JacksonJsonObjectMapperTests.java:85

JacksonMessagingUtilsTests > shouldIncludeWellKnownModules() FAILED
    java.lang.AssertionError at JacksonMessagingUtilsTests.java:65
```
See failure for respective PR from @dependabot.

anthologia:
Here's the UUID bug I mentioned: https://github.com/FasterXML/jackson-databind/issues/5225
I tested with rc6 locally and it seems to be fixed.

I'll check the test failures and fix them !
Quick look suggests `ServiceLoader` isn't picking up the Joda module in `.findAndAddModules(JacksonJsonObjectMapper.class.getClassLoader())`.

anthologia:
The rc6 failures might be related to this issue:
https://github.com/FasterXML/jackson-datatype-joda/issues/175

It appears that `JodaModule`'s `ServiceLoader` metadata was removed in rc6, which could explain why `findAndAddModules()` is no longer discovering it.

I've reported this to the Jackson team, but in the meantime, I'm gonna implement a workaround with the `*Provider` used in our Jackson 2 mapper.

But anyway, UUID Serializer can be removed. 😁

spring-integration-core/src/main/java/org/springframework/integration/aot/CoreRuntimeHints.java
```diff
@@ -166,4 +162,25 @@ private static void registerSpringJdkProxy(ProxyHints proxyHints, Class<?>... pr
 		proxyHints.registerJdkProxy(AopProxyUtils.completeJdkProxyInterfaces(proxiedInterfaces));
 	}
 
+	@SuppressWarnings("removal")
```
anthologia: > ... So, please, bring old entries back with respective `removal` suppression, and add new entries.

_Originally posted by @artembilan in https://github.com/spring-projects/spring-integration/pull/10212#discussion_r2216850652_

I'm not sure this is what you meant.

anthologia:
Sorry for the rebase. Had to resolve conflicts with the Jackson version updates. I've addressed all your review feedback.

For the Redis test failures, I don't currently have a Redis environment set up locally, but I'll configure one and check it soon.

artembilan:
> It appears that JodaModule's ServiceLoader metadata was removed in rc6

I think we need to remove that optional Joda module dependency and respective tests.
We don’t do anything with Joda in the project. That was just to prove the module is OK with the infra. These days, with Java time API, I don’t think anyone uses Joda. 
If you have to use it, just add that dependency into your project and customize ObjectMapper. 
this has to be a separate fix. Then upgrade to rc6 has to be merged. And then this your PR rebased. And we will move on from there.

spring-integration-core/src/main/java/org/springframework/integration/aot/CoreRuntimeHints.java
artembilan: I see what you mean.
Right, let’s mark this method as deprecated to not forget to remove it eventually. Yes , it does not warn because we don’t use API directly. It also looks like these new methods could be static.
On the other hand: Jackson 3 is going to be with us forever, so no need to extract method for it.

build.gradle
```diff
@@ -70,7 +70,7 @@ ext {
     hsqldbVersion = '2.7.4'
     h2Version = '2.3.232'
     jacksonVersion = '2.19.2'
-    jackson3Version = '3.0.0-rc5'
+    jackson3Version = '3.0.0-rc6'
```
artembilan: Please, hold of any rc6 changes.
See my comment about Joda.
I am from the phone , so cannot be very helpful on weekend 😆

artembilan (changes requested):

artembilan:
BTW, I'm not sure what you mean with "real Redis" since we use Testcontainers.
See `RedisContainerTest`. 
And that is enough for debugging.
If you mean an access to the data stored in the Redis, then we would need to think more.
Might be there is a way to not do that, but still get data in debug on the client side before it is sent to Redis.

artembilan:
The Jackson 3 rc6 PR has been just merged: https://github.com/spring-projects/spring-integration/pull/10231.
Please, consider to raise your one and remove that UUID workaround.
Thanks

anthologia:
Found the issue with redis module test failures!
It was a problem when deserializing `MessageGroupMetadata`.
I've fixed it and will commit the changes after work. Think we can get this release today ! 😄

spring-integration-core/src/main/java/org/springframework/integration/aot/CoreRuntimeHints.java
```diff
@@ -103,12 +104,14 @@ public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader)
 			reflectionHints.registerType(JsonPathUtils.class, MemberCategory.INVOKE_PUBLIC_METHODS);
 		}
 
+		registerJackson2Hints(reflectionHints);
```
artembilan, line 108: I think we need to add `@SuppressWarnings("removal")` to this line or the whole method since we call this deprecated API.

spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
artembilan, line 256: Since this class is deprecated, we need to add a `@SuppressWarnings("removal")` to this line or the whole method.

spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupMetadata.java
```diff
@@ -55,9 +57,16 @@ public class MessageGroupMetadata implements Serializable {
 	private volatile String condition;
 
 	public MessageGroupMetadata() {
+		this.messageIds = new LinkedList<>();
+	}
+
+	@JsonCreator
+	public MessageGroupMetadata(List<UUID> messageIds) {
```
artembilan: I wonder if this new ctor could not be `public`.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -49,4 +49,10 @@ Previously deprecated classes in the `spring-integation-hazelcast` module, such
...
 The `AbstractMqttMessageDrivenChannelAdapter` and `ClientManager` implementations now expose a `quiescentTimeout` option which is propagated in their `stop()` method down to the `disconnectForcibly()` API of the MQTT Paho clients.
-See xref:mqtt.adoc[] for more information.
\ No newline at end of file
+See xref:mqtt.adoc[] for more information.
+
+[[x7.0-jackson-3-Support]]
+=== Jackson 3 Support
+
+The Jackson 2 has been deprecated for removal in whole Spring portfolio.
+Spring Integration 7.0 introduces Jackson 3 as the default with new components: `JacksonJsonObjectMapper`, `JacksonPropertyAccessor`, `JacksonIndexAccessor`, and `JacksonMessagingUtils`.
```
artembilan: We don't need to mention our own version since the whole file is really about that version and it says "what's new in this version" in the beginning.

artembilan (changes requested):
Just a couple minor concerns.
I can address them on merge in an hour or so if you won't make it earlier 😄 

Thank you !

anthologia:
Thanks! I can commit the review fixes right now.

But I just found some remaining Jackson2 in amqp / jms modules test code though. Should we handle now or next release?
(Anyway we left Kafka for later.)

artembilan:
I see many other new `removal` warnings.
But that's fine so far.
Merging your PR and address whatever I can spot locally.

Thank you again!

artembilan:
Here is follow up clean up where we have just introduced new compilation warnings: https://github.com/spring-projects/spring-integration/commit/0bc432a884614a642ad2723763be387ac933e194.

I think there is no Jackson-related warnings any more.

PR #10213, opened 2025-07-19
Apply Nullability to MongoDb module

mjd507:
Related to: #10083

### DIFF
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/inbound/MongoDbMessageSource.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/inbound/MongoDbMessageSource.java
@@ -19,6 +19,8 @@ package org.springframework.integration.mongodb.inbound;
 import java.util.Collection;
 import java.util.List;
 
+import org.jspecify.annotations.Nullable;
+
 import org.springframework.context.ApplicationContext;
 import org.springframework.data.mongodb.MongoDatabaseFactory;
 import org.springframework.data.mongodb.core.MongoOperations;
@@ -31,7 +33,6 @@ import org.springframework.expression.Expression;
 import org.springframework.integration.mongodb.support.MongoHeaders;
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.integration.transaction.IntegrationResourceHolder;
-import org.springframework.lang.Nullable;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
@@ -64,6 +65,7 @@ public class MongoDbMessageSource extends AbstractMongoDbMessageSource<Object> {
 	@Nullable
 	private final MongoDatabaseFactory mongoDbFactory;
 
+	@SuppressWarnings("NullAway.Init")
 	private MongoOperations mongoTemplate;
 
 	/**
@@ -105,9 +107,7 @@ public class MongoDbMessageSource extends AbstractMongoDbMessageSource<Object> {
 		if (this.mongoDbFactory != null) {
 			MongoTemplate template = new MongoTemplate(this.mongoDbFactory, getMongoConverter());
 			ApplicationContext applicationContext = getApplicationContext();
-			if (applicationContext != null) {
-				template.setApplicationContext(applicationContext);
-			}
+			template.setApplicationContext(applicationContext);
 			this.mongoTemplate = template;
 		}
 		setMongoConverter(this.mongoTemplate.getConverter());
@@ -124,7 +124,7 @@ public class MongoDbMessageSource extends AbstractMongoDbMessageSource<Object> {
 	 * query will be provided in the {@link MongoHeaders#COLLECTION_NAME} header.
 	 */
 	@Override
-	protected Object doReceive() {
+	protected @Nullable Object doReceive() {
 		Assert.isTrue(isInitialized(), "This class is not yet initialized. Invoke its afterPropertiesSet() method");
 		AbstractIntegrationMessageBuilder<Object> messageBuilder = null;
 
--- a/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
+++ b/spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
@@ -16,7 +16,10 @@
 
 package org.springframework.integration.mongodb.outbound;
 
+import java.util.Objects;
+
 import org.bson.Document;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.data.mongodb.MongoDatabaseFactory;
 import org.springframework.data.mongodb.core.MongoOperations;
@@ -46,22 +49,25 @@ import org.springframework.util.Assert;
  */
 public class MongoDbOutboundGateway extends AbstractReplyProducingMessageHandler {
 
-	private MongoDatabaseFactory mongoDbFactory;
+	private @Nullable MongoDatabaseFactory mongoDbFactory;
 
-	private MongoConverter mongoConverter;
+	private @Nullable MongoConverter mongoConverter;
 
+	@SuppressWarnings("NullAway.Init")
 	private MongoOperations mongoTemplate;
 
+	@SuppressWarnings("NullAway.Init")
 	private EvaluationContext evaluationContext;
 
-	private Expression queryExpression;
+	private @Nullable Expression queryExpression;
 
-	private MessageCollectionCallback<?> collectionCallback;
+	private @Nullable MessageCollectionCallback<?> collectionCallback;
 
 	private boolean expectSingleResult = false;
 
 	private Class<?> entityClass = Document.class;
 
+	@SuppressWarnings("NullAway.Init")
 	private Expression collectionNameExpression;
 
 	public MongoDbOutboundGateway(MongoDatabaseFactory mongoDbFactory) {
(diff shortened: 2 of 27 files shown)


spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/inbound/MongoDbMessageSource.java
```diff
@@ -105,9 +107,7 @@ protected void onInit() {
 		if (this.mongoDbFactory != null) {
 			MongoTemplate template = new MongoTemplate(this.mongoDbFactory, getMongoConverter());
 			ApplicationContext applicationContext = getApplicationContext();
-			if (applicationContext != null) {
```
mjd507, line 108: I think `MessageSource` always be used as a bean ?

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MessageDocument.java
```diff
@@ -42,32 +44,32 @@ public class MessageDocument {
...
 
-	private Integer priority;
+	private @Nullable Integer priority;
 
 	private Long createdTime = 0L;
 
 	private Long groupCreatedTime = 0L;
 
-	private Object groupId;
+	private @Nullable Object groupId;
```
mjd507, line 59: looks like if we used as a MessageStore, `groupId` would be null in this document.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
```diff
@@ -197,9 +200,7 @@ public void setCustomConverters(Object... customConverters) {
 
 	@Override
 	public void afterPropertiesSet() {
-		if (this.applicationContext != null) {
```
mjd507, line 200: `MessageStore` always a bean ?

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
```diff
@@ -611,49 +615,47 @@ public <S> S read(Class<S> clazz, Bson source) {
 		}
 
 		private MessageWrapper readAsMessageWrapper(Bson source) {
```
mjd507, line 615: I can not make this method return nullable, the underlying `spring-data-mongodb` does not allow to return null.
so I assert the `source` param never null?

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/inbound/MongoDbMessageSource.java
artembilan: Has to.
We accept this change and let everything what is not a bean fail. We will take a responsibility to explain in case of complains

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGateway.java
```diff
@@ -147,20 +153,21 @@ protected void doInit() {
...
 	@Override
-	protected Object handleRequestMessage(Message<?> requestMessage) {
+	protected @Nullable Object handleRequestMessage(Message<?> requestMessage) {
 		String collectionName =
 				this.collectionNameExpression.getValue(this.evaluationContext, requestMessage, String.class);
 		Assert.notNull(collectionName, "'collectionNameExpression' cannot evaluate to null");
 		Object result;
 
 		if (this.collectionCallback != null) {
+			var collectionCallbackToUse = this.collectionCallback;
```
artembilan: This var has to be extracted before `if` and used in that `if`.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/outbound/ReactiveMongoDbStoringMessageHandler.java
```diff
@@ -105,6 +110,7 @@ protected void onInit() {
 		super.onInit();
 		this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
 		if (this.mongoTemplate == null) {
+			Objects.requireNonNull(this.mongoDbFactory);
```
artembilan: Place this inside of the template ctor

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
```diff
@@ -144,6 +148,7 @@ protected MessageBuilderFactory getMessageBuilderFactory() {
 	@Override
 	public void afterPropertiesSet() {
 		if (this.mongoTemplate == null) {
+			Objects.requireNonNull(this.mongoDbFactory);
```
artembilan, line 151: DITTO

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MessageDocument.java
artembilan: Ok

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
artembilan: True

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
```diff
@@ -511,18 +512,21 @@ private void updateGroup(Object groupId, Update update) {
...
-		return ((Number) this.template.findAndModify(query,
-						new Update().inc(SEQUENCE, 1L),
-						FindAndModifyOptions.options().returnNew(true).upsert(true),
-						Map.class, this.collectionName)
-				.get(SEQUENCE))  // NOSONAR - never returns null
+		Map<?, ?> result = this.template.findAndModify(query,
+				new Update().inc(SEQUENCE, 1L),
+				FindAndModifyOptions.options().returnNew(true).upsert(true),
+				Map.class, this.collectionName);
+		Objects.requireNonNull(result);
```
artembilan: This looks redundant since you use another one in the next operator.
We need to think to get rid off of this at all if we are sure it is never null.
If this is a critical execution path, it would be better to avoid as much as possible extra calls.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
```diff
@@ -511,18 +512,21 @@ private void updateGroup(Object groupId, Update update) {
...
+		return ((Number) Objects.requireNonNull(result
+				.get(SEQUENCE)))
 				.longValue();
 	}
 
 	@SuppressWarnings(UNCHECKED)
 	private static void enhanceHeaders(MessageHeaders messageHeaders, Map<String, Object> headers) {
 		Map<String, Object> innerMap =
 				(Map<String, Object>) new DirectFieldAccessor(messageHeaders).getPropertyValue(HEADERS);
+		Objects.requireNonNull(innerMap);
```
artembilan: Same here.
Probably `NullAway` suppression is the way to go for these methods .

artembilan (changes requested):

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
mjd507: below MappingMongoConverter and MongoTemplate both use this, so I leave it here.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
mjd507: I suppress it with NullAway.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/AbstractConfigurableMongoDbMessageStore.java
artembilan: Ok. Then `NullAway.Init` on this connectionFactory property ?

artembilan: Or `Assert.notNull()`.
This method is called only once, so having a check is OK.
We use `Objects.requireNonNull` only if we deliberately know that it is not null.
Probably one day I’ll revise them all into a `NullAway` suppressions since I’m very picky for performance 😁

mjd507: `mongoDbFactory` can be null, when we give a `mongoTemplate` via ctor

mjd507: > Or `Assert.notNull()`. This method is called only once, so having a check is OK. We use `Objects.requireNonNull` only if we deliberately know that it is not null. Probably one day I’ll revise them all into a `NullAway` suppressions since I’m very picky for performance 😁

ok

artembilan (approved):
Pulling this locally for final review and possible clean up & merge.

Thank you again for so many great contributions!

artembilan:
Merged as: https://github.com/spring-projects/spring-integration/commit/fb12c0540f400f1879b9169ec8794b54ce4f6f55

artembilan:
Here is some clean up after your change: https://github.com/spring-projects/spring-integration/commit/a3476d25f9a308fa918e25423f870eacd137aa87.

Pay attention to the fix in all those components where we have a choice between template and connection factory.

PR #10236, opened 2025-07-20
GH-10235: Remove Joda Time support from Jackson 3 ObjectMapper

### LINKED ISSUE #10235: Remove Joda Time support from Jackson 3 ObjectMapper
We don’t do anything with Joda in the project. That was just to prove the module is OK with the infra. These days, with Java time API, I don’t think anyone uses Joda. ...

Originally posted by @artembilan in https://github.com/spring-projects/spring-integration/issues/10212#issuecomment-3094515080

## Proposed Changes
This issue will remove all Joda Time dependency and related code from the Jackson 3 ObjectMapper.

anthologia:
Fixes: #10235

* Remove `JodaModule` dependency and related code

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -495,7 +495,6 @@ project('spring-integration-core') {
         api 'io.micrometer:micrometer-observation'
 
         optionalApi 'tools.jackson.core:jackson-databind'
-        optionalApi 'tools.jackson.datatype:jackson-datatype-joda'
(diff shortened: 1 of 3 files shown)


artembilan:
Just a note of convenience.
If you have something ready to PR, no need to open issue.
GH treats both of the equally and we can schedule PR into milestone and so on.
Thanks

PR #10259, opened 2025-07-23
Use `RabbitTestContainer` for AMQP integration tests

mjd507:


artembilan:
I have merged this, but bear in mind that sometimes we have to have a real broker to be able to debug properly what is going on.
Perhaps that is more relevant to Spring AMQP project where we have more low-lever API for RabbitMQ and here in Spring Integration we should more concentrate on the channel adapters behavior.
So, we give it a shot and with time we'll see how inconvenient it is not deal with real broker any more.
Thanks

PR #10260, opened 2025-07-23
GH-10083: Apply Nullability to GraphQL module

artembilan:

* Add `@SuppressWarnings("NullAway")` to the function where we get a message id, which technically is never null in Spring Integration


cppwfs:
LGTM !   Thank you !

PR #10262, opened 2025-07-24
Support listening AMQP 1.0 based messages in `AmqpInboundChannelAdapter`

mjd507:
related to: #10090

I feel this one not easy for me, and not sure if I am implementing it in the right way.

so I just try `AmqpInboundChannelAdapter` first, and leave two queries in the code comment, 

looking for the suggestions. thank you

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
@@ -24,6 +24,7 @@ import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.amqp.Consumer;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.amqp.core.AcknowledgeMode;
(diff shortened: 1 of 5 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
```diff
@@ -254,6 +257,10 @@ protected void onInit() {
 		}
 		Listener messageListener;
 		if (this.messageListenerContainer.isConsumerBatchEnabled()) {
+			// here we use the contract from `MessageListenerContainer`
+			// but for amqp 1.0 - RabbitAmqpListenerContainer, we use `batchSize` (>1) to distinguish batch or not.
+			// should we implement isConsumerBatchEnabled in 1.0 - RabbitAmqpListenerContainer ?
+			// because batchSize are not exposed from the container, how we know it is batch or not for amqp 1.0/
```
artembilan, line 264: Thank you for the effort, but that's not.
The `spring-rabbitmq-client` is for AMQP 1.0 and it is a separate module form reason.
So, it would be natural evolution here in Spring Integration to have fully new channel adapters for that module.
It might not justify a new module here, because eventually we are going to migrate away from AMQP 0.9.
So, this existing channel adapters would be deprecated and new would stay.

I haven't gone with engineering for this new channel adapters yet, so cannot suggest class names to start with 😄

mjd507:
@artembilan should we expose few methods in `spring-amqp` project first?
like 
`RabbitAmqpListenerContainer.isConsumerBatchEnabled()`
`RabbitAmqpListenerContainer.getAcknowledgeMode()`
`RabbitAmqpListenerContainer.getConnectionFactory()`
`RabbitAmqpTemplate.getMessageConverter()`

artembilan (changes requested):
Sorry, looks like I didn't sent my comment yesterday.
I regret to say that, but your current code change is not what we anticipate.

mjd507:
Okay, thanks for clarifying, close this.

PR #10263, opened 2025-07-24
GH-10261: Don't use Jackson 3 API in places with Jackson 2

### LINKED ISSUE #10261: Don't use Jackson 3 specific API in places where we deal with Jackson 2
An example is this code snippet from `KafkaMessageDrivenChannelAdapter`:
```
		if (JacksonPresent.isJackson2Present()) {
			MessagingMessageConverter messageConverter = new MessagingMessageConverter();
			// For consistency with the rest of Spring Integration channel adapters
			messageConverter.setGenerateMessageId(true);
			messageConverter.setGenerateTimestamp(true);
			DefaultKafkaHeaderMapper headerMapper = new DefaultKafkaHeaderMapper();
			headerMapper.addTrustedPackages(JacksonMessagingUtils.DEFAULT_TRUSTED_PACKAGES.toArray(new String[0]));
			this.recordListener.setMessageConverter(messageConverter);
			this.batchListener.setMessageConverter(messageConverter);
		}
```

So, we are supposed to be in this `if` only when `JacksonPresent.isJackson2Present()`. 
However that `JacksonMessagingUtils.DEFAULT_TRUSTED_PACKAGES` is from Jackson 3 specific API.
And that leads to the:
```
Caused by: java.lang.NoClassDefFoundError: tools/jackson/databind/ValueSerializer
    at org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter.<init>(KafkaMessageDrivenChannelAdapter.java:141)
```

Use deprecated `JacksonJsonUtils` instead in those places.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10261

The problem is like this:
```
Caused by: java.lang.NoClassDefFoundError: tools/jackson/databind/ValueSerializer
    at org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter.<init>(KafkaMessageDrivenChannelAdapter.java:141)
```

Not clear why accessing static constant without any references to Jackson API causes imports of that class to be accessed, but that is what we have so far.

* Use already deprecated `JacksonJsonUtils` in those places when `JacksonPresent.isJackson2Present()`

### DIFF
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
@@ -25,7 +25,6 @@ import org.jspecify.annotations.Nullable;
 import org.springframework.integration.dispatcher.MessageDispatcher;
 import org.springframework.integration.dispatcher.RoundRobinLoadBalancingStrategy;
 import org.springframework.integration.dispatcher.UnicastingDispatcher;
-import org.springframework.integration.support.json.JacksonMessagingUtils;
 import org.springframework.integration.support.management.ManageableSmartLifecycle;
 import org.springframework.kafka.config.KafkaListenerContainerFactory;
 import org.springframework.kafka.core.KafkaOperations;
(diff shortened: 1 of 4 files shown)


cppwfs:
LGTM!

Thanks for fixing this!

anthologia:
I guess... It happens because `DEFAULT_TRUSTED_PACKAGES` is not a compile-time constant.
Accessing it triggers class initialization, which resolves all symbolic references in that class (including Jackson 3 types). That’s why a `NoClassDefFoundError` occurs even without explicit access to Jackson classes.

```java
class Constants {
    static final String VALUE = "CONSTANT"; // compile-time constant - inlined
    static final String COMPUTED = "prefix" + System.getProperty("suffix"); // not a compile-time constant - triggers initialization
    static {
        initializeDependencies(); // dependency initialization executed
    }
}

```

PR #10265, opened 2025-07-24
GH-10083: Apply Nullability to Core `expression` package

artembilan:

* Make `SupplierExpression` as to not return null for result of the provided `Supplier`. And fix Nullability from the super contract
* Make `ValueExpression` compatible with super contract, but still ensure not null for return value
* Make `DynamicExpression` compatible with super contract. Provide usability and code style refactoring
* Make `FunctionExpression` to be compatible with super contract. At the same time ensure that input for the function is not null, but result of function could be null. Some other code style and usability refactoring
* Remove redundant logic in the `ExpressionUtils` which is deliberately not reachable due to Nullability
* Fix Nullability for the `ReloadableResourceBundleExpressionSource` and some simple refactoring according to IDE report.
* Make `ExpressionEvalMap` to be aware of null values after expressions evaluation.
* Fix usage of the `ExpressionEvalMap` in the project.
* Mark `CassandraMessageHandler.setQuery()` with `NullAway` since `ExpressionEvalMap` may produce nulls, and Cassandra driver can deal with nulls, but that is not expressed on the API with Nullability
* Fix `ThreadStatePropagationChannelInterceptor` for wrong import for the `@Nullable`
* Slight change in the `GraphQlMessageHandler` according to the mentioned above `expression` package fixes

### DIFF
--- a/spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
+++ b/spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
@@ -145,6 +145,7 @@ public class CassandraMessageHandler extends AbstractReplyProducingMessageHandle
 		setStatementProcessor((ExpressionEvaluatingMessageProcessor<Statement<?>>) expressionEvaluatingMessageProcessor);
 	}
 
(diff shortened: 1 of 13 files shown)


cppwfs:
LGTM 
Great Work!

PR #10272, opened 2025-07-28
GH-10083: Apply Nullability to core dispatcher package

cppwfs:
PartititonDispatcher.partition will always return at least 1 partition. So return value is wrapped with Objects.requireNonNull, for nullaway

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
@@ -20,7 +20,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
@@ -165,11 +164,12 @@ public class PartitionedDispatcher extends AbstractDispatcher {
 	}
 
 	@Override
+	@SuppressWarnings("NullAway")
 	public boolean dispatch(Message<?> message) {
 		populatedPartitions();
 		int partition = Math.abs(this.partitionKeyFunction.apply(message).hashCode()) % this.partitionCount;
 		UnicastingDispatcher partitionDispatcher = this.partitions.get(partition);
-		return Objects.requireNonNull(partitionDispatcher).dispatch(message);
+		return partitionDispatcher.dispatch(message);
 	}
 
 	private void populatedPartitions() {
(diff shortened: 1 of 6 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
```diff
@@ -168,7 +169,7 @@ public boolean dispatch(Message<?> message) {
 		populatedPartitions();
 		int partition = Math.abs(this.partitionKeyFunction.apply(message).hashCode()) % this.partitionCount;
 		UnicastingDispatcher partitionDispatcher = this.partitions.get(partition);
-		return partitionDispatcher.dispatch(message);
+		return Objects.requireNonNull(partitionDispatcher).dispatch(message);
```
artembilan: This is not OK.
The `dispatch()` is a critical runtime execution path, therefore anything extra here would cause a performance degradation.
Right, it could be very minor with this `requireNonNull()`, but I don't want to take my chances.
Let's bite a bullet and mark this method as `@SuppressWarnings("NullAway") // Dataflow analysis limitation`!

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
cppwfs: Understood.   Thank you for catching this!   I'll get an update out later this afternoon!

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
```diff
@@ -165,11 +164,12 @@ public void shutdown() {
 	}
 
 	@Override
+	@SuppressWarnings("NullAway")
```
artembilan: Can we state in the comment to this annotation the reason?
Something similar to what I suggested in the review, but might be more specific to this situation.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
```diff
@@ -164,7 +164,7 @@ public void shutdown() {
 	}
 
 	@Override
-	@SuppressWarnings("NullAway")
+	@SuppressWarnings("NullAway") // Dataflow analysis limitation!
```
artembilan: Yeah... That's not what I meant, especially not with a `!` 😉 
How about this?
```suggestion
	@SuppressWarnings("NullAway") // The partitions map never returns null according to partition hash
```

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dispatcher/PartitionedDispatcher.java
cppwfs: Thank you for the clarification.   I thought that message was a bit terse.   will update.

PR #10274, opened 2025-07-31
Apply NullAbility to MQTT module

mjd507:
Related to: #10083


spring-integration-core/src/main/java/org/springframework/integration/mapping/BytesMessageMapper.java
```diff
@@ -34,14 +34,16 @@
...
 		return toMessage(object, null);
 	}
 
 	@Override
 	@NonNull
-		// override
 	Message<?> toMessage(byte[] bytes, @Nullable Map<String, Object> headers);
 
+	@Override
+	byte @NonNull [] fromMessage(Message<?> message);
```
artembilan: We don’t need `@NonNull` at all.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/mapping/BytesMessageMapper.java
mjd507: yes, yes. removed.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/event/MqttMessageNotDeliveredEvent.java
```diff
@@ -34,7 +36,7 @@ public class MqttMessageNotDeliveredEvent extends MqttMessageDeliveryEvent {
 
 	private final Throwable exception;
 
-	public MqttMessageNotDeliveredEvent(Object source, int messageId, String clientId,
+	public MqttMessageNotDeliveredEvent(Object source, int messageId, @Nullable String clientId,
```
artembilan: I think we need to revise `clientId`. It is not null this or other way.
See `AbstractMqttClientManager.getClientId()`.
We should utilize that in the `AbstractMqttMessageHandler.getClientId()` as an `if..else` to return provided `clientId` or the one supplied by the `clientManager`

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
```diff
@@ -78,9 +79,10 @@ public abstract class AbstractMqttMessageDrivenChannelAdapter<T, C> extends Mess
 
 	private boolean manualAcks;
 
+	@SuppressWarnings("NullAway.Init")
 	private ApplicationEventPublisher applicationEventPublisher;
 
-	private MqttMessageConverter converter;
+	private @Nullable MqttMessageConverter converter;
```
artembilan, line 85: This is configured in the `MqttPahoMessageDrivenChannelAdapter.onInit()`.
Therefore, `@SuppressWarnings("NullAway.Init")`, too.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -192,13 +192,14 @@ private void connect() throws MqttException {
...
 						"If no 'url' provided, connectionOptions.getServerURIs() must not be null");
+				Assert.state(getClientId() != null, "'clientId' must not be null");
 				this.client = this.clientFactory.getAsyncClientInstance(getUrl(), getClientId());
 				this.client.setCallback(this);
 				this.client.connect(connectionOptions).waitForCompletion(getCompletionTimeout());
 				this.client.setManualAcks(isManualAcks());
 			}
 			else {
-				this.client = clientManager.getClient();
+				this.client = Objects.requireNonNull(clientManager.getClient());
```
artembilan: I think `Assert.state()` here to indicate that `clientManager` has to be started.

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/event/MqttMessageNotDeliveredEvent.java
mjd507: indeed. i extract `getClientId()` to the top `ClientManager` interface. also `getUrl()`.
for inbound/outbound adapters, in their ctors, if url/client not provided, will be assigned from the clientManager.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/AbstractMqttMessageDrivenChannelAdapter.java
mjd507: by checking, for outbound adapter, the converter does configured via onInit(). both v3 & v5
but for inbound adapters, the converter looks only init in v3,  v5 not use this converter at all, v5 defines its own `SmartMessageConverter`, so l leave it here.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -381,11 +375,11 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
 		}
 	}
 
-	private AbstractIntegrationMessageBuilder<?> toMessageBuilder(String topic, MqttMessage mqttMessage) {
+	private @Nullable AbstractIntegrationMessageBuilder<?> toMessageBuilder(String topic, MqttMessage mqttMessage) {
```
artembilan: I don't think this method is supposed to return null at all.
Plus, I'm not fun of that ` Objects.requireNonNull()` in this critical path execution.
Let's mark this method with a `@SuppressWarnings("NullAway")` instead!

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -276,7 +276,9 @@ private void connect() throws MqttException {
 				this.mqttClient.connect(this.connectionOptions).waitForCompletion(getCompletionTimeout());
 			}
 			else {
-				this.mqttClient = clientManager.getClient();
+				IMqttAsyncClient client = clientManager.getClient();
+				Assert.state(client != null, "'client' must not be null");
```
artembilan: Let's improve this message!
"The 'client' must not be null. Consider to start 'clientManager'."
And apply it everywhere we deal with `clientManager.getClient()`

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -473,7 +469,7 @@ public void connectComplete(boolean reconnect, String serverURI) {
 	private void subscribe() {
 		var clientManager = getClientManager();
 		if (clientManager != null && this.mqttClient == null) {
-			this.mqttClient = clientManager.getClient();
+			this.mqttClient = Objects.requireNonNull(clientManager.getClient());
```
artembilan: Same `Assert.state(client != null,` instead.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/AbstractMqttMessageHandler.java
```diff
@@ -401,31 +401,30 @@ protected void handleMessageInternal(Message<?> message) {
 		}
 
 		Assert.state(topic != null, "No topic could be determined from the message and no default topic defined");
+		Assert.state(mqttMessage != null,
```
artembilan: This is good, but in the wrong place.
Please, move the whole command just after that `Object mqttMessage` is evaluated!

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -199,7 +198,7 @@ private void connect() throws MqttException {
 			}
 			else {
 				IMqttAsyncClient theClient = clientManager.getClient();
-				Assert.state(theClient != null, "'client' must not be null");
+				Assert.state(theClient != null, "The 'client' must not be null, Consider to start 'clientManager'.");
```
artembilan: Let's fix the sentence!
```suggestion
				Assert.state(theClient != null, "The 'client' must not be null; consider to start the 'clientManager'.");
```

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/MqttPahoMessageDrivenChannelAdapter.java
```diff
@@ -375,11 +374,12 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
 		}
 	}
 
-	private @Nullable AbstractIntegrationMessageBuilder<?> toMessageBuilder(String topic, MqttMessage mqttMessage) {
+	@SuppressWarnings("NullAway") // MessageBuilder is not supposed to be null
```
artembilan: Well, actually, we also have just removed that ` Objects.requireNonNull`.
So, @SuppressWarnings("NullAway") is not only the reason for a null builder.
Let's just use a more generic comment: `// Dataflow analysis limitation` !

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
```diff
@@ -50,7 +52,17 @@ public interface ClientManager<T, C> extends SmartLifecycle, MqttComponent<C> {
...
+
+	/**
+	 * @return the managed client url, never null
+	 */
+	String getUrl();
+
+	/**
+	 * @return the managed clientId, never null
+	 */
+	String getClientId();
```
artembilan, line 70: Please, add, to these new public API `@since 7.0` and some short general sentence.
Could be just copy of that `@return` tag content.

artembilan (changes requested):

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/ClientManager.java
mjd507: make sense. updated

PR #10275, opened 2025-07-31
Apply Nullability to Debezium module

ngocnhan-tran1996:

### DIFF
--- a/spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
+++ b/spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
@@ -18,6 +18,7 @@ package org.springframework.integration.debezium.inbound;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
@@ -55,12 +56,12 @@ public class DebeziumMessageProducer extends MessageProducerSupport {
 
 	private final DebeziumEngine.Builder<ChangeEvent<byte[], byte[]>> debeziumEngineBuilder;
 
-	private DebeziumEngine<ChangeEvent<byte[], byte[]>> debeziumEngine;
+	private @Nullable DebeziumEngine<ChangeEvent<byte[], byte[]>> debeziumEngine;
 
 	/**
 	 * Debezium Engine is designed to be submitted to an {@link Executor}.
 	 */
-	private TaskExecutor taskExecutor;
+	private @Nullable TaskExecutor taskExecutor;
 
 	private String contentType = "application/json";
 
@@ -76,7 +77,6 @@ public class DebeziumMessageProducer extends MessageProducerSupport {
 	 * Create new Debezium message producer inbound channel adapter.
 	 * @param debeziumBuilder - pre-configured Debezium Engine Builder instance.
 	 */
-	@SuppressWarnings("NullAway.Init")
 	public DebeziumMessageProducer(Builder<ChangeEvent<byte[], byte[]>> debeziumBuilder) {
 		Assert.notNull(debeziumBuilder, "'debeziumBuilder' must not be null");
 		this.debeziumEngineBuilder = debeziumBuilder;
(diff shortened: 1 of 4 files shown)


spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
```diff
@@ -76,6 +76,7 @@ public class DebeziumMessageProducer extends MessageProducerSupport {
 	 * Create new Debezium message producer inbound channel adapter.
 	 * @param debeziumBuilder - pre-configured Debezium Engine Builder instance.
 	 */
+	@SuppressWarnings("NullAway.Init")
```
artembilan: Why this annotation here, but not on the property itself?
And why do we need it at all?

artembilan (changes requested):

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
ngocnhan-tran1996: When I ran gradlew `:spring-integration-debezium:check` on my local machine, the following exception occurred:

```
DebeziumMessageProducer.java:79: error: [NullAway] initializer method does not guarantee @NonNull fields debeziumEngine (line 58), taskExecutor (line 63) are initialized along all control-flow paths (remember to check for exceptions or early returns).
        public DebeziumMessageProducer(Builder<ChangeEvent<byte[], byte[]>> debeziumBuilder) {
               ^
  1 error
```

After reviewing the related PRs, I’ve updated my PR accordingly.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
```diff
@@ -55,12 +56,12 @@ public class DebeziumMessageProducer extends MessageProducerSupport {
 
 	private final DebeziumEngine.Builder<ChangeEvent<byte[], byte[]>> debeziumEngineBuilder;
 
-	private DebeziumEngine<ChangeEvent<byte[], byte[]>> debeziumEngine;
+	private @Nullable DebeziumEngine<ChangeEvent<byte[], byte[]>> debeziumEngine;
```
artembilan: This cannot be null according to ctor logic.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
```diff
@@ -55,12 +56,12 @@ public class DebeziumMessageProducer extends MessageProducerSupport {
...
 
 	/**
 	 * Debezium Engine is designed to be submitted to an {@link Executor}.
 	 */
-	private TaskExecutor taskExecutor;
+	private @Nullable TaskExecutor taskExecutor;
```
artembilan: This has to be `@SuppressWarnings("NullAway.Init")` instead since it is initialized in the `onInit()`.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
```diff
@@ -167,7 +168,7 @@ protected void doStart() {
 			return;
 		}
 		this.lifecycleLatch = new CountDownLatch(1);
-		this.taskExecutor.execute(() -> {
+		Objects.requireNonNull(this.taskExecutor).execute(() -> {
```
artembilan: Please, remove all the `Objects.requireNonNull`.

spring-integration-debezium/src/main/java/org/springframework/integration/debezium/inbound/DebeziumMessageProducer.java
```diff
@@ -177,7 +178,7 @@ protected void doStart() {
 				// The batch size, polling frequency, and other parameters are controlled via connector's configuration
 				// settings. This continues until this connector is stopped.
 				// This method can be called repeatedly as needed.
-				this.debeziumEngine.run();
+				Objects.requireNonNull(this.debeziumEngine).run();
```
artembilan: This property also has to be marked with `@SuppressWarnings("NullAway.Init")` instead.

artembilan (changes requested):

PR #10277, opened 2025-07-31
Apply Nullability to feed module

ngocnhan-tran1996:

### DIFF
--- a/spring-integration-feed/src/main/java/org/springframework/integration/feed/inbound/FeedEntryMessageSource.java
+++ b/spring-integration-feed/src/main/java/org/springframework/integration/feed/inbound/FeedEntryMessageSource.java
@@ -26,7 +26,6 @@ import java.net.http.HttpResponse;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
-import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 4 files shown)


spring-integration-feed/src/main/java/org/springframework/integration/feed/inbound/FeedEntryMessageSource.java
```diff
@@ -206,7 +210,7 @@ private SyndEntry getNextEntry() {
 		else {
 			this.lastTime += 1; //NOSONAR - single poller thread
 		}
-		this.metadataStore.put(this.metadataKey, this.lastTime + "");
+		Objects.requireNonNull(this.metadataStore).put(this.metadataKey, this.lastTime + "");
```
artembilan: We prefer to avoid `Objects.requireNonNull` as much as possible, especially in critical path executions.
So, use `@SuppressWarnings("NullAway.Init")` on the `metadataStore` property instead.
In fact, it is really initialized from the `onInit()`.

spring-integration-feed/src/main/java/org/springframework/integration/feed/inbound/FeedEntryMessageSource.java
```diff
@@ -258,7 +262,7 @@ private SyndFeed buildSyndFeed() throws IOException, URISyntaxException, Interru
 			HttpRequest request =
 					HttpRequest.newBuilder()
 							.GET()
-							.uri(this.feedUrl.toURI())
+							.uri(Objects.requireNonNull(this.feedUrl).toURI())
```
artembilan: Well, this one is indeed can be `null` according to the constructors logic.
However, since we have here `if..else`, there is no chance that we come at this point and it is still null.
So, this method has to be marked with a `@SuppressWarnings("NullAway") // Dataflow analysis limitation` instead.

artembilan (changes requested):

PR #10278, opened 2025-08-01
Apply Nullability to syslog module

ngocnhan-tran1996:

### DIFF
--- a/spring-integration-syslog/src/main/java/org/springframework/integration/syslog/RFC5424SyslogParser.java
+++ b/spring-integration-syslog/src/main/java/org/springframework/integration/syslog/RFC5424SyslogParser.java
@@ -133,8 +133,7 @@ public class RFC5424SyslogParser {
 	 * @param r the reader.
 	 * @return the timestamp.
 	 */
-	@Nullable
-	protected Object getTimestamp(Reader r) {
+	protected @Nullable Object getTimestamp(Reader r) {
 
 		int c = r.getc();
 
--- a/spring-integration-syslog/src/main/java/org/springframework/integration/syslog/RFC5424SyslogParser.java
+++ b/spring-integration-syslog/src/main/java/org/springframework/integration/syslog/RFC5424SyslogParser.java
@@ -21,6 +21,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.jspecify.annotations.Nullable;
+
 import org.springframework.integration.JavaUtils;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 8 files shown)


spring-integration-syslog/src/main/java/org/springframework/integration/syslog/inbound/TcpSyslogReceivingChannelAdapter.java
```diff
@@ -36,9 +38,10 @@
 public class TcpSyslogReceivingChannelAdapter extends SyslogReceivingChannelAdapterSupport
 		implements TcpListener, ApplicationEventPublisherAware {
 
+	@SuppressWarnings("NullAway.Init")
 	private volatile AbstractServerConnectionFactory connectionFactory;
 
-	private volatile ApplicationEventPublisher applicationEventPublisher;
+	private volatile @Nullable ApplicationEventPublisher applicationEventPublisher;
```
artembilan: This one also `@SuppressWarnings("NullAway.Init")`.
Because it is initialized from the Application context callbacks.

spring-integration-syslog/src/main/java/org/springframework/integration/syslog/RFC5424SyslogParser.java
```diff
@@ -152,6 +155,7 @@ protected Object getTimestamp(Reader r) {
 		return dateBuilder.toString();
 	}
 
+	@Nullable
 	private Object getStructuredData(Reader r) {
```
artembilan: Please, put `@Nullable` on the return type.
This is a contract exactly of the return type, not a method.

artembilan (changes requested):

spring-integration-syslog/src/main/java/org/springframework/integration/syslog/RFC5424SyslogParser.java
```diff
@@ -131,6 +133,7 @@ public RFC5424SyslogParser(boolean retainOriginal) {
 	 * @param r the reader.
 	 * @return the timestamp.
 	 */
+	@Nullable
 	protected Object getTimestamp(Reader r) {
```
artembilan: Thanks, but how have you missed to fix this one as well?

artembilan (changes requested):

spring-integration-syslog/src/main/java/org/springframework/integration/syslog/RFC5424SyslogParser.java
ngocnhan-tran1996: My bad, I missed it.

PR #10285, opened 2025-08-03
Apply Nullability to stream module

ngocnhan-tran1996:

### DIFF
--- a/spring-integration-stream/src/main/java/org/springframework/integration/stream/ByteStreamReadingMessageSource.java
+++ b/spring-integration-stream/src/main/java/org/springframework/integration/stream/ByteStreamReadingMessageSource.java
@@ -22,6 +22,8 @@ import java.io.InputStream;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import org.jspecify.annotations.Nullable;
+
 import org.springframework.integration.endpoint.AbstractMessageSource;
 import org.springframework.messaging.MessagingException;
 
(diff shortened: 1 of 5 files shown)


spring-integration-stream/src/main/java/org/springframework/integration/stream/ByteStreamReadingMessageSource.java
```diff
@@ -72,6 +72,7 @@ public String getComponentType() {
 		return "stream:stdin-channel-adapter(byte)";
 	}
 
+	@SuppressWarnings("NullAway")
```
artembilan: That's not.
This method indeed can return null.
Please, try to avoid `@SuppressWarnings("NullAway")` as much as possible.
It really hides a lot and we have to be vigilant what we are doing around our code.

The fix should be like this for this method signature:
```
	@Override
	protected byte @Nullable [] doReceive() {
```

artembilan (changes requested):
@ngocnhan-tran1996 ,

Since you are contributing a lot and I'd like to make our cooperation a bit smooth and productive, please, review this article about an importance of the commit message: https://cbea.ms/git-commit/.

What you shown so far is OK, but we could be a bit more convenient adding something like `Related to: https://github.com/spring-projects/spring-integration/issues/10083` to fixes for that issue.
Other one issue - one PR could utilized `Fixes: ` pattern.

I am also not fun of commit messages like `Addressing PR review`.
What PR is from the commit perspective?
Why should I search for PR to determine the reason of the change?
It is really better to express the change in the commit message.

Thanks for understanding!

PR #10286, opened 2025-08-03
Apply Nullability to groovy module

ngocnhan-tran1996:


artembilan (changes requested):
Looks like you have missed `org.springframework.integration.groovy.dsl` package.
Thanks.

PR #10288, opened 2025-08-05
GH-10083: Apply Nullability to `file` module

artembilan:

* The change in the `file.dsl` package has affected the `ComponentsRegistration` contract in the core module. And that one has led to respective fixes in the `AMQP`, `JPA`, `Kafka` modules for DSL packages
* Fix `SimpleJsonSerializer` contract to return an empty JSON instead of `null`. Something similar to what Jackson 3 provides for us by default.
* Rework a `WatchServiceDirectoryScanner` logic to initialize its internals in the ctor to avoid null-away errors
* Fix `FileWritingMessageHandler`, `HeadDirectoryScanner`, and `RecursiveDirectoryScanner` code style and logic according to IDE suggestions and new Nullability functionality
* Fix classes in the `file.config` package according to new Nullability logic
* Implement Nullability for `FileListFilter` abstraction and its implementations
* Fix `FileChannelCache` for a proper constant name.
* Respectively, fix the `NioFileLockerTests` to use a new `FileChannelCache.CHANNEL_CACHE` constant
* The remote directory could be null for some protocol operations. Therefore implement that respectively in the `FileInfo`, `Session` and `RemoteFileOperations`

The change might look big, but it is inevitable and pretty straightforward


spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
```diff
@@ -370,7 +372,7 @@ protected void onInit() {
 	}
 
 	@Override
-	protected AbstractIntegrationMessageBuilder<File> doReceive() {
+	protected @org.jspecify.annotations.Nullable AbstractIntegrationMessageBuilder<File> doReceive() {
```
ngocnhan-tran1996: I think `@Nullable` is enough, no need for the fully qualified class name.

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
```diff
@@ -175,7 +175,7 @@ protected SmbFile enhanceNameWithSubDirectory(SmbFile file, String directory) {
 	}
 
 	@Override
-	protected String getFullFileName(String remoteDirectory, SmbFile smbFile) {
+	protected String getFullFileName(@Nullable String remoteDirectory, SmbFile smbFile) {
```
cppwfs, line 178: Should we have a javadoc here saying that remoteDirectory is ignored in this implementation?
Just curious.

spring-integration-file/src/main/java/org/springframework/integration/file/tail/OSDelegatingFileTailingMessageProducer.java
```diff
@@ -93,11 +95,12 @@ protected void doStart() {
 		super.doStart();
 		destroyProcess();
 		String[] tailOptions = this.options.split("\\s+");
-		this.tailCommand = new String[tailOptions.length + 2];
-		this.tailCommand[0] = "tail";
-		this.tailCommand[this.tailCommand.length - 1] = getFile().getAbsolutePath();
-		System.arraycopy(tailOptions, 0, this.tailCommand, 1, tailOptions.length);
-		this.command = String.join(" ", this.tailCommand);
+		String[] newTailCommand = new String[tailOptions.length + 2];
```
cppwfs, line 98: Why do we need a newTailCommand here?   Readability?

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
```diff
@@ -169,20 +166,16 @@ protected void doStart() {
 
 	@Override
 	protected void doStop() {
-		if (this.idleEventScheduledFuture != null) {
-			this.idleEventScheduledFuture.cancel(true);
+		ScheduledFuture<?> idleEventScheduledFutureToCancel = this.idleEventScheduledFuture;
+		if (idleEventScheduledFutureToCancel != null) {
```
cppwfs, line 170: Why do we need to put `this.idleEventScheduledFuture` into a local variable to test it?  Readability?

spring-integration-file/src/main/java/org/springframework/integration/file/tail/ApacheCommonsFileTailingMessageProducer.java
```diff
@@ -102,7 +103,10 @@ protected void doStart() {
 	@Override
 	protected void doStop() {
 		super.doStop();
-		this.tailer.close();
+		Tailer tailerToClose = this.tailer;
```
cppwfs, line 106: Was this change made for readability?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
```diff
@@ -256,11 +259,11 @@ public boolean isRunning() {
 	 * Polls from the file source. If the result is not null, it will be returned.
 	 * If the result is null, it attempts to sync up with the remote directory to populate the file source.
 	 * At most, maxFetchSize files will be fetched.
-	 * Then, it polls the file source again and returns the result, whether or not it is null.
+	 * Then, it polls the file source again and returns the result, whether it is null.
```
cppwfs: We could simplify this to say, "Then, it polls the file source again and returns the result".

spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotationPolicy.java
```diff
@@ -54,33 +53,29 @@ public interface RotationPolicy {
...
+		/**
+		 * @return the directory
+		 * @deprecated if favor of {@link #directory()} ()}
+		 */
+		@Deprecated(forRemoval = true, since = "7.0")
 		public String getDirectory() {
 			return this.directory;
 		}
 
-		@Override
```
cppwfs, line 79: Just curious.  Why are we removing the `toString`.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/RemoteFileOperations.java
```diff
@@ -106,14 +102,12 @@ public interface RemoteFileOperations<F> {
...
 	/**
 	 * Remove a remote file.
-	 *
 	 * @param path The full path to the file.
 	 * @return true when successful.
 	 */
 	boolean remove(String path);
 
 	/**
-	 * Rename a remote file, creating directories if needed.
```
cppwfs, line 116: Why was the description removed?
Possibly wrong line was removed?

spring-integration-file/src/main/java/org/springframework/integration/file/remote/FileInfo.java
```diff
@@ -55,12 +57,14 @@ public interface FileInfo<F> {
...
 	 * @return the remote directory in which the file resides
 	 */
+	@Nullable
 	String getRemoteDirectory();
 
 	/**
 	 * @return a string representing the permissions of the remote
 	 * file (e.g. -rw-r--r--).
 	 */
+	@Nullable
```
cppwfs, line 67: Should the `@Nullable` be on the same line as definition?

spring-integration-file/src/main/java/org/springframework/integration/file/HeadDirectoryScanner.java
```diff
@@ -49,25 +53,21 @@ public final void setFilter(FileListFilter<File> filter) {
...
-			return Arrays.asList(files).subList(0, Math.min(files.length, this.maxNumberOfFiles));
-		}
+			public List<File> filterFiles(File @Nullable [] files) {
+				return files == null
+						? List.of()
+						: Arrays.asList(files).subList(0, Math.min(files.length, this.maxNumberOfFiles));
+			}
 
-	}
+		}
```
cppwfs, line 69: This tab can be removed.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -514,7 +514,7 @@ private void validateDestinationDirectory(File destinationDirectory, boolean aut
 	}
 
 	@Override // NOSONAR
-	protected Object handleRequestMessage(Message<?> requestMessage) {
+	protected @Nullable Object handleRequestMessage(Message<?> requestMessage) {
```
cppwfs, line 517: Do we want to remove the `// NO SONAR`?

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
```diff
@@ -967,11 +969,13 @@ public void flushIfNeeded(FlushPredicate flushPredicate) {
...
 	 */
-	public void flushIfNeeded(MessageFlushPredicate flushPredicate, Message<?> filterMessage) {
+	public void flushIfNeeded(MessageFlushPredicate flushPredicate, @Nullable Message<?> filterMessage) {
 		doFlush(findFilesToFlush(flushPredicate, filterMessage));
 	}
 
-	private Map<String, FileState> findFilesToFlush(MessageFlushPredicate flushPredicate, Message<?> filterMessage) {
+	private Map<String, FileState> findFilesToFlush(MessageFlushPredicate flushPredicate,
+			@Nullable Message<?> filterMessage) {
+
```
cppwfs, line 978: Remove the extra line.

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileReadingMessageSourceFactoryBean.java
```diff
@@ -128,20 +132,19 @@ protected FileReadingMessageSource createInstance() {
...
 			this.source = new FileReadingMessageSource(this.comparator);
 		}
-		else if (queueSizeSet) {
+		else if (this.queueSize != null) {
 			this.source = new FileReadingMessageSource(this.queueSize);
 		}
 		else {
 			this.source = new FileReadingMessageSource();
 		}
+		Assert.notNull(this.directory, "The 'directory' must be provided.");
```
cppwfs, line 147: This can be removed because the setDirectory already checks for null.

spring-integration-file/src/main/java/org/springframework/integration/file/DefaultDirectoryScanner.java
```diff
@@ -82,17 +84,19 @@ protected FileLocker getLocker() {
 	 */
 	@Override
 	public boolean tryClaim(File file) {
-		return (this.locker == null) || this.locker.lock(file);
+		FileLocker lockerToUse = this.locker;
```
cppwfs, line 87: Was this change made for readability?

cppwfs (changes requested):
Looks good.   Just some nitpicks

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbOutboundGateway.java
artembilan: Not sure why would one be interested in that since this method is `protected` and is not end-user API to call.

spring-integration-file/src/main/java/org/springframework/integration/file/tail/OSDelegatingFileTailingMessageProducer.java
artembilan: To avoid NullAway error.
It complains that `this.tailCommand` could be `null` when we do the logic against it.
With a local variable we got modification only once: on the assignment after this local variable manipulation.
This is similar to what we have in the `destroyProcess()` a little bellow: we need a local variable to avoid possible race condition.

spring-integration-file/src/main/java/org/springframework/integration/file/tail/FileTailingMessageProducerSupport.java
artembilan: Same NPE race condition as in many other places.
If we don't change it, then IDE complains that `this.idleEventScheduledFuture.cancel(true);` could fail with NPE.
And I am kinda agreed with that: there is no guarantee at runtime that `this.idleEventScheduledFuture` does not become `null` after we enter that `if` block.

spring-integration-file/src/main/java/org/springframework/integration/file/tail/ApacheCommonsFileTailingMessageProducer.java
artembilan: DITTO

spring-integration-file/src/main/java/org/springframework/integration/file/remote/aop/RotationPolicy.java
artembilan: True. Because Java `record` does that for us automatically. From Google: 
```
Records use a streamlined syntax that automatically generates common methods, such as:

    A canonical constructor that takes all record components as arguments.
    Accessor methods for each component (e.g., name() for a name component).
    equals(), hashCode(), and toString() methods based on the record's components.
```
More info in: https://www.baeldung.com/java-record-keyword

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
artembilan: Sure!
But if this is only concern in this review, I wouldn't bother 😉

spring-integration-file/src/main/java/org/springframework/integration/file/remote/FileInfo.java
artembilan: Not on methods in the interfaces.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
artembilan: Not yet. The method body is too long and complex.
One day we may come back to it and review its logic.

spring-integration-file/src/main/java/org/springframework/integration/file/FileWritingMessageHandler.java
artembilan: No. That's my preference: if method head is multi-line, then we need extra blank line before method body.
It makes the code much easier to read.

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileReadingMessageSourceFactoryBean.java
artembilan: Right. And what if `setDirectory()` was not called?

spring-integration-file/src/main/java/org/springframework/integration/file/DefaultDirectoryScanner.java
artembilan: Same NPE race condition when we deal with nullable property.

cppwfs (approved):
LGTM

cppwfs:
Squashed Merged

PR #10289, opened 2025-08-05
Make pathFor protected to allow overriding

zUniQueX:
This PR changes the visibility of the `pathFor` method to `protected`.

Intention:
While developing and debugging an application, it can be very helpful to know which locks are currently locked. The UUIDs cannot be reversed to the original lock keys and therefore it can get hard to check which locks are currently obtained.

When overriding the `pathFor` method for this, one can write the lock key directly to the db and view human-readable strings in the db table.


artembilan:
The point is that `LOCK_KEY` is a part of primary key in the table.
Therefore, an arbitrary string is not OK for a proper indexing.

I'm not sure how this `pathFor()` helps you to mitigate human comprehension, but your fix does not hurt.
Therefore, merging.

Thank you for contribution!

PR #10290, opened 2025-08-05
Fix `PartitionedDispatcher` to use `List` for `partitions`

artembilan:
The `partitions` map in the `PartitionedDispatcher` is based on the index for `key`, which is really an overkill and regular list with indexed access should be enough.

* Fix `PartitionedDispatcher` to use a `List` for `partitions`
* Remove `@SuppressWarnings("NullAway")` from `dispatch()` method since there is not going to be a null from `List.get()`


cppwfs:
The change is a good idea.
LGTM

PR #10291, opened 2025-08-07
GH-10083: Apply Nullability into `dsl` package

artembilan:

* Fix all the reported issues and IDE suggestions
* Also, improve Nullability in the `TailAdapterSpec` and couple DSL Specs in JDBC module according to a new Nullability rules in the core `dsl` package

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java
@@ -16,8 +16,6 @@
 
 package org.springframework.integration.dsl;
 
-import org.jspecify.annotations.Nullable;
-
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.support.ErrorMessageStrategy;
 import org.springframework.integration.support.management.observation.MessageReceiverObservationConvention;
@@ -49,10 +47,8 @@ public abstract class MessageProducerSpec<S extends MessageProducerSpec<S, P>, P
 	 * Configure the message producer's bean name.
 	 */
 	@Override
-	public S id(@Nullable String id) {
-		if (id != null) {
-			this.target.setBeanName(id);
-		}
+	public S id(String id) {
+		this.target.setBeanName(id);
 		return super.id(id);
 	}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java
@@ -37,7 +37,10 @@ import org.springframework.messaging.MessageChannel;
 public abstract class MessageProducerSpec<S extends MessageProducerSpec<S, P>, P extends MessageProducerSupport>
 		extends IntegrationComponentSpec<S, P> {
 
-	public MessageProducerSpec(@Nullable P producer) {
+	public MessageProducerSpec() {
+	}
+
+	public MessageProducerSpec(P producer) {
 		this.target = producer;
 	}
 
@@ -47,7 +50,9 @@ public abstract class MessageProducerSpec<S extends MessageProducerSpec<S, P>, P
 	 */
 	@Override
 	public S id(@Nullable String id) {
-		this.target.setBeanName(id);
+		if (id != null) {
+			this.target.setBeanName(id);
+		}
 		return super.id(id);
 	}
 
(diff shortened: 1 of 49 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/QueueChannelSpec.java
```diff
@@ -31,9 +33,9 @@
  */
 public class QueueChannelSpec extends MessageChannelSpec<QueueChannelSpec, QueueChannel> {
 
-	protected Queue<Message<?>> queue; // NOSONAR
+	protected @Nullable Queue<Message<?>> queue; // NOSONAR
```
cppwfs, line 36: Curious.  Why do we need to keep the `//NOSONAR`?

spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java
```diff
@@ -47,7 +50,9 @@ public MessageProducerSpec(@Nullable P producer) {
 	 */
 	@Override
 	public S id(@Nullable String id) {
-		this.target.setBeanName(id);
+		if (id != null) {
+			this.target.setBeanName(id);
```
cppwfs: Just curious, why are we not throwing an error when they tryin to set the id to null, as it pertains to the bean name?

cppwfs (changes requested):
Looks good.   Just a couple of questions.

spring-integration-core/src/main/java/org/springframework/integration/dsl/QueueChannelSpec.java
artembilan: Because this property is set from extension classes.
Yes, we may remove it since does not look like anything currently in our CI/CD complains 😄

spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java
artembilan: Yeah... The logic is a bit convoluted and indeed there is a path when this method is called with `null`.
That is still not a problem because if an id is null, the real bean name is going to be generated in the `IntegrationFlowBeanPostProcessor` on `this.target` bean registration.

cppwfs: That makes sense.   So a bean name does not have to match the id in that case.

artembilan: It does match if `id` is provided.

cppwfs: Could there be a case where:

```
mySpec.id("myID");
//beanName is now "myId"
//id is now "myId"
mySpec.id(null); //user calls id again and sets it to null
// beanName is "myId"
// id is null

Is this an issue?

Comment (artembilan) [reply spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java]: True. This is a problem.

Let me play with that if you are fine having more files to review in this already big PR 😄 
For example:
```
	public static AmqpMessageChannelSpec<?, ?> channel(ConnectionFactory connectionFactory) {
		return channel(null, connectionFactory);
	}

	public static AmqpMessageChannelSpec<?, ?> channel(@Nullable String id, ConnectionFactory connectionFactory) {
		return new AmqpMessageChannelSpec<>(connectionFactory)
				.id(id);
	}
```

So, the first method would not call the second just to have a nice delegation and code flow.
This way an `id` in the second method would not be null.

cppwfs: Code review and coffee are a beautiful thing.  😆

artembilan: Turned out that wasn't too bad.
Just only much more classes to review yet 😄

cppwfs: WOOHOO!   Thanks for looking into this and resolving it!

cppwfs:
LGTM Thanks for the hard work!

spring-integration-core/src/main/java/org/springframework/integration/dsl/MessageProducerSpec.java
artembilan: Thank you! For reviewing this, spotting concern and raising your doubts.
I’m glad that I decided to not save you from boring code pushing the original change directly upstream 🤗

PR #10292, opened 2025-08-08
Apply NullAbility to sftp module

ngocnhan-tran1996:
Related to: #10083

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpInboundFileSynchronizer.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpInboundFileSynchronizer.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.sftp.inbound;
 
 import org.apache.sshd.sftp.client.SftpClient;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.integration.file.remote.session.SessionFactory;
 import org.springframework.integration.file.remote.synchronizer.AbstractInboundFileSynchronizer;
@@ -46,13 +47,13 @@ public class SftpInboundFileSynchronizer extends AbstractInboundFileSynchronizer
 	}
 
 	@Override
-	protected boolean isFile(SftpClient.DirEntry file) {
+	protected boolean isFile(SftpClient.@Nullable DirEntry file) {
 		return file != null && file.getAttributes().isRegularFile();
 	}
 
 	@Override
 	protected String getFilename(SftpClient.DirEntry file) {
-		return file != null ? file.getFilename() : null;
+		return file.getFilename();
 	}
 
 	@Override
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpInboundFileSynchronizer.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpInboundFileSynchronizer.java
@@ -17,7 +17,6 @@
 package org.springframework.integration.sftp.inbound;
 
 import org.apache.sshd.sftp.client.SftpClient;
-import org.jspecify.annotations.Nullable;
 
 import org.springframework.integration.file.remote.session.SessionFactory;
 import org.springframework.integration.file.remote.synchronizer.AbstractInboundFileSynchronizer;
@@ -47,8 +46,8 @@ public class SftpInboundFileSynchronizer extends AbstractInboundFileSynchronizer
 	}
 
 	@Override
-	protected boolean isFile(SftpClient.@Nullable DirEntry file) {
-		return file != null && file.getAttributes().isRegularFile();
+	protected boolean isFile(SftpClient.DirEntry file) {
+		return file.getAttributes().isRegularFile();
 	}
 
 	@Override
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpStreamingMessageSource.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpStreamingMessageSource.java
@@ -22,6 +22,7 @@ import java.util.Comparator;
 import java.util.List;
 
 import org.apache.sshd.sftp.client.SftpClient;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.integration.file.remote.AbstractFileInfo;
 import org.springframework.integration.file.remote.AbstractRemoteFileStreamingMessageSource;
@@ -58,7 +59,7 @@ public class SftpStreamingMessageSource extends AbstractRemoteFileStreamingMessa
 	 */
 	@SuppressWarnings("this-escape")
 	public SftpStreamingMessageSource(RemoteFileTemplate<SftpClient.DirEntry> template,
-			Comparator<SftpClient.DirEntry> comparator) {
+			@Nullable Comparator<SftpClient.DirEntry> comparator) {
 
 		super(template, comparator);
 		doSetFilter(new SftpPersistentAcceptOnceFileListFilter(new SimpleMetadataStore(), "sftpStreamingMessageSource"));
@@ -79,7 +80,7 @@ public class SftpStreamingMessageSource extends AbstractRemoteFileStreamingMessa
 	}
 
 	@Override
-	protected boolean isDirectory(SftpClient.DirEntry file) {
+	protected boolean isDirectory(SftpClient.@Nullable DirEntry file) {
 		return file != null && file.getAttributes().isDirectory();
 	}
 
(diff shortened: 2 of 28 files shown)


spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpInboundFileSynchronizer.java
```diff
@@ -46,13 +47,13 @@ public SftpInboundFileSynchronizer(SessionFactory<SftpClient.DirEntry> sessionFa
 	}
 
 	@Override
-	protected boolean isFile(SftpClient.DirEntry file) {
+	protected boolean isFile(SftpClient.@Nullable DirEntry file) {
```
artembilan: Let's remove this one and respective condition in the body like you did for other methods here.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/inbound/SftpStreamingMessageSource.java
```diff
@@ -79,7 +80,7 @@ protected List<AbstractFileInfo<SftpClient.DirEntry>> asFileInfoList(Collection<
 	}
 
 	@Override
-	protected boolean isDirectory(SftpClient.DirEntry file) {
+	protected boolean isDirectory(SftpClient.@Nullable DirEntry file) {
```
artembilan: Same here, please.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/server/ApacheMinaSftpEvent.java
```diff
@@ -35,7 +36,8 @@ public ApacheMinaSftpEvent(Object source) {
 		super(source);
 	}
 
-	public ApacheMinaSftpEvent(Object source, Throwable cause) {
+	@SuppressWarnings("NullAway") // Overridden method does not define nullability
+	public ApacheMinaSftpEvent(Object source, @Nullable Throwable cause) {
```
artembilan, line 39: Let's then do that!
If that is fine with you in this PR.
I mean the super.super `IntegrationEvent` does declare `@Nullable`, so it is logical that `FileServerEvent` should follow.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/server/DirectoryCreatedEvent.java
```diff
@@ -43,8 +43,7 @@ public DirectoryCreatedEvent(Object source, Path path, Map<String, ?> attrs) {
 		this.attrs = attrs;
 	}
 
-	@Nullable
-	public Path getPath() {
+	public @Nullable Path getPath() {
```
artembilan: Let's remove this annotation!
According to the `DirectoryCreatedEvent`, the `path` is never null.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/server/FileWrittenEvent.java
```diff
@@ -48,8 +48,7 @@ public String getRemoteHandle() {
 		return this.remoteHandle;
 	}
 
-	@Nullable
-	public Path getFile() {
+	public @Nullable Path getFile() {
```
artembilan: DITTO

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/server/PathMovedEvent.java
```diff
@@ -41,13 +41,11 @@ public PathMovedEvent(Object source, Path srcPath, Path dstPath, @Nullable Throw
...
 
-	@Nullable
-	public Path getSrcPath() {
+	public @Nullable Path getSrcPath() {
 		return this.srcPath;
 	}
 
-	@Nullable
-	public Path getDstPath() {
+	public @Nullable Path getDstPath() {
```
artembilan: DITTO on both props.
See ctor contract.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/server/PathRemovedEvent.java
```diff
@@ -41,8 +41,7 @@ public PathRemovedEvent(Object source, Path path, boolean isDirectory, @Nullable
 		this.isDirectory = isDirectory;
 	}
 
-	@Nullable
-	public Path getPath() {
+	public @Nullable Path getPath() {
```
artembilan: And here.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpRemoteFileTemplate.java
```diff
@@ -58,13 +58,13 @@ public SftpRemoteFileTemplate(SessionFactory<SftpClient.DirEntry> sessionFactory
 		super(sessionFactory);
 	}
 
-	@SuppressWarnings("unchecked")
+	@SuppressWarnings({"unchecked", "NullAway"}) // Overridden method does not define nullability
```
artembilan: Let's make super `@Nullable` instead!

ngocnhan-tran1996:
@artembilan 

Has anything been updated regarding the PR build?
Because the log shows the following error:

```
* Where:
Build file '/home/runner/work/spring-integration/spring-integration/build.gradle' line: 532

* What went wrong:
A problem occurred evaluating root project 'spring-integration'.
> org/gradle/util/ConfigureUtil
```

artembilan:
> Has anything been updated regarding the PR build?

Yeah... Looks like `gradle-setup` GH action no install Gradle `9.0.0` which we are not ready yet.
We have to use a wrapper in CI/CD instead.
Give me a couple mins to review your PR and I'll push the fix for GHA workflows I have tested in other projects.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
```diff
@@ -92,19 +93,24 @@ public boolean remove(String path) throws IOException {
...
+	public String[] listNames(@Nullable String path) throws IOException {
 		return doList(path)
 				.map(SftpClient.DirEntry::getFilename)
 				.toArray(String[]::new);
 	}
 
-	public Stream<SftpClient.DirEntry> doList(String path) throws IOException {
+	public Stream<SftpClient.DirEntry> doList(@Nullable String path) throws IOException {
+
+		if (path == null) {
```
artembilan: I don't think this is correct.
Let's see if we can mitigate it as a `.` path internally!

artembilan (changes requested):
Great work!
Let's see if my review valid!
Thank you!

artembilan:
The latest `main` is green again: https://github.com/spring-projects/spring-integration/actions/runs/16838982796.

Please, consider to rebase your branch respectively.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
ngocnhan-tran1996: Could we check not null in here?

`Assert.notNull(path, "The 'path' can not be null");`

artembilan: No. If it is `null`. we set it into `.`
Not here, though, but on the next line and so on.

artembilan (approved):
LGTM.
as long as PR build is green, I'll merge it.
Thank you!
Was nice way to finish the week 😄

artembilan (changes requested):
Looks like SMB module has been affected by your change as well:
```
 SmbTests > testSmbInboundStreamFlow() FAILED
    java.lang.AssertionError at SmbTests.java:148
```

ngocnhan-tran1996:
The test is green on my local machine

image


However, I will revert `RemoteFileTemplate` to avoid affecting other modules.

artembilan:
> However, I will revert RemoteFileTemplate to avoid affecting other modules.

Well, I think it is probably better to fix those modules as well.
That is a natural evolution of the change we are doing so far.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpRemoteFileTemplate.java
artembilan: No. Bring back, please, the fix we have just had in super class.
And fix SMB one respectively.

artembilan (changes requested):

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpRemoteFileTemplate.java
ngocnhan-tran1996: I haven't done anything and the test's green 👀

PR #10293, opened 2025-08-09
Apply NullAbility to scripting module

ngocnhan-tran1996:
Related to: #10083

### DIFF
--- a/spring-integration-groovy/src/main/java/org/springframework/integration/groovy/GroovyScriptExecutingMessageProcessor.java
+++ b/spring-integration-groovy/src/main/java/org/springframework/integration/groovy/GroovyScriptExecutingMessageProcessor.java
@@ -137,7 +137,7 @@ public class GroovyScriptExecutingMessageProcessor extends AbstractScriptExecuti
 	}
 
 	@Override
-	protected ScriptSource getScriptSource(Message<?> message) {
+	protected ScriptSource getScriptSource(@Nullable Message<?> message) {
 		return this.scriptSource;
 	}
 
--- a/spring-integration-groovy/src/main/java/org/springframework/integration/groovy/GroovyScriptExecutingMessageProcessor.java
+++ b/spring-integration-groovy/src/main/java/org/springframework/integration/groovy/GroovyScriptExecutingMessageProcessor.java
@@ -137,7 +137,7 @@ public class GroovyScriptExecutingMessageProcessor extends AbstractScriptExecuti
 	}
 
 	@Override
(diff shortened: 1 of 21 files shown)


spring-integration-groovy/src/main/java/org/springframework/integration/groovy/GroovyScriptExecutingMessageProcessor.java
```diff
@@ -137,7 +137,7 @@ public void setCompilerConfiguration(CompilerConfiguration compilerConfiguration
 	}
 
 	@Override
-	protected ScriptSource getScriptSource(Message<?> message) {
+	protected ScriptSource getScriptSource(@Nullable Message<?> message) {
```
artembilan: Let's fix `ScriptExecutingMessageSource` instead providing a fake message instance!

The `MessageProcessor` cotnract is to not accept null for the `message` argument.

That could be a `static final` property in the `ScriptExecutingMessageSource` with `byte[0]` as payload.

spring-integration-scripting/src/main/java/org/springframework/integration/scripting/AbstractScriptExecutingMessageProcessor.java
```diff
@@ -44,9 +45,9 @@ public abstract class AbstractScriptExecutingMessageProcessor<T>
 
 	private final ScriptVariableGenerator scriptVariableGenerator;
 
-	private ClassLoader beanClassLoader;
+	private @Nullable ClassLoader beanClassLoader;
 
-	private BeanFactory beanFactory;
+	private @Nullable BeanFactory beanFactory;
```
artembilan: Both of these should be `NullAway.Init`

spring-integration-scripting/src/main/java/org/springframework/integration/scripting/DefaultScriptVariableGenerator.java
```diff
@@ -44,8 +46,8 @@ public DefaultScriptVariableGenerator(Map<String, Object> variableMap) {
 		this.variableMap = variableMap;
 	}
 
-	public Map<String, Object> generateScriptVariables(Message<?> message) {
-		Map<String, Object> scriptVariables = new HashMap<String, Object>();
+	public Map<String, Object> generateScriptVariables(@Nullable Message<?> message) {
```
artembilan: After fixing `ScriptExecutingMessageSource` many of these `@Nullable` on `message` would go away.

artembilan (changes requested):

PR #10301, opened 2025-08-10
GH-10083: Apply Nullability into `webflux` module

anthologia:
Related to: #10083


spring-integration-webflux/src/main/java/org/springframework/integration/webflux/dsl/WebFluxMessageHandlerSpec.java
```diff
@@ -48,7 +49,7 @@ public class WebFluxMessageHandlerSpec
 		extends BaseHttpMessageHandlerSpec<WebFluxMessageHandlerSpec, WebFluxRequestExecutingMessageHandler> {
 
 	@Nullable
-	protected final WebClient webClient; // NOSONAR - final
+	protected final WebClient webClient;
```
artembilan: If we already here, let's, please, move `@Nullable` next to the type, not method modifiers.

artembilan (changes requested):

PR #10302, opened 2025-08-10
GH-10083: Apply Nullability into `test` module

anthologia:
Related to: #10083

### DIFF
--- a/spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
+++ b/spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
@@ -20,6 +20,8 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.jspecify.annotations.Nullable;
+
 import org.springframework.context.ApplicationContext;
 import org.springframework.integration.endpoint.AbstractEndpoint;
 import org.springframework.test.context.TestContext;
@@ -51,8 +53,7 @@ class SpringIntegrationTestExecutionListener implements TestExecutionListener {
 		MockIntegrationContext mockIntegrationContext = applicationContext.getBean(MockIntegrationContext.class);
 		this.autoStartupCandidates = mockIntegrationContext.getAutoStartupCandidates();
 		this.autoStartupCandidates.stream()
-				.filter(endpoint -> endpoint.getBeanName() != null
-						&& !match(endpoint.getBeanName(), patterns))
+				.filter(endpoint -> !match(endpoint.getBeanName(), patterns))
 				.peek(endpoint -> endpoint.setAutoStartup(true))
 				.forEach(AbstractEndpoint::start);
 	}
@@ -64,7 +65,7 @@ class SpringIntegrationTestExecutionListener implements TestExecutionListener {
 				.forEach(AbstractEndpoint::stop);
 	}
 
-	private static boolean match(String name, String[] patterns) {
+	private static boolean match(@Nullable String name, String[] patterns) {
 		return patterns.length > 0 &&
 				Arrays.stream(patterns)
 						.anyMatch(pattern -> PatternMatchUtils.simpleMatch(pattern, name));
--- a/spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
+++ b/spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
@@ -17,7 +17,10 @@
 package org.springframework.integration.test.context;
 
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Set;
+
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.context.ApplicationContext;
 import org.springframework.integration.endpoint.AbstractEndpoint;
@@ -37,7 +40,7 @@ import org.springframework.util.PatternMatchUtils;
  */
 class SpringIntegrationTestExecutionListener implements TestExecutionListener {
 
-	private Set<AbstractEndpoint> autoStartupCandidates;
+	private Set<AbstractEndpoint> autoStartupCandidates = new HashSet<>();
 
 	@Override
 	public void prepareTestInstance(TestContext testContext) {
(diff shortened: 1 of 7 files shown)


spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
```diff
@@ -62,7 +65,7 @@ public void afterTestClass(TestContext testContext) {
 				.forEach(AbstractEndpoint::stop);
 	}
 
-	private static boolean match(String name, String[] patterns) {
+	private static boolean match(@Nullable String name, String[] patterns) {
```
artembilan, line 68: How has that happened that we expect null here? Maybe that part has to be fixed instead ? The logic does not make sense then : why would one try to match against a pattern if the source to match is null?

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockIntegration.java
```diff
@@ -72,9 +73,8 @@ public static <T> MessageSource<T> mockMessageSource(T payload, T... payloads) {
 		}
 
 		return (MessageSource<T>) mockMessageSource(new GenericMessage<>(payload),
-				(messages != null
-						? messages.toArray(new Message<?>[messages.size()])
-						: null));
+				(messages != null ?
```
artembilan: I prefer existing style. Please, revert .

artembilan (changes requested):

spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
anthologia: I thought it would be reasonable to use the existing nullable handling in `PatternMatchUtils.simpleMatch` without changing our logic. But I understand your point 😄

spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
```diff
@@ -50,7 +51,8 @@ public void prepareTestInstance(TestContext testContext) {
 		MockIntegrationContext mockIntegrationContext = applicationContext.getBean(MockIntegrationContext.class);
 		this.autoStartupCandidates = mockIntegrationContext.getAutoStartupCandidates();
 		this.autoStartupCandidates.stream()
-				.filter(endpoint -> !match(endpoint.getBeanName(), patterns))
+				.filter(endpoint -> endpoint.getBeanName() != null
+						&& !match(endpoint.getBeanName(), patterns))
```
artembilan: Oh! I see.
Plus it looks like `PatternMatchUtils.simpleMatch` is null-aware.
Let's bring your original change back to this our `match()` method!

Sorry for a noise: was reviewing on weekend from the phone.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -85,6 +87,7 @@ protected MockMessageHandler(ArgumentCaptor<Message<?>> messageArgumentCaptor) {
 	 * @param nextMessageConsumer the Consumer to handle the next incoming message.
 	 * @return this
 	 */
+	@SuppressWarnings("NullAway") // See github.com/uber/NullAway/issues/1075
```
artembilan, line 90: I'm not sure why we need this here.
That `Consumer<Message<?>>` cannot accept a null.
The message is never null in the integration flow.

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
```diff
@@ -127,6 +130,7 @@ protected void handleMessageInternal(Message<?> message) {
 			this.lock.unlock();
 		}
 
+		Assert.notNull(function, "function must not be null");
```
artembilan: Good. Let's also add something like: 
"At least one of  the 'handleNext' or 'handleNextAndReply' has to be attached to this MockMessageHandler."

artembilan (changes requested):

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
anthologia: `@SuppressWarnings` is not for the Consumer param, but for the `lastFunction`'s return value. Lambda returns null which causes NullAway to complain about returning `@Nullable` from `@NonNull` context.

```java
/Users/jooyoung/spring-integration/spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java:94: error: [NullAway] returning @Nullable expression from method with @NonNull return type
                        return null;
                        ^
```

Can't apply `@Nullable` to wildcard `?`, and don't want to change it to `Object`.
(`protected @Nullable Function<Message<?>, @Nullable ?> lastFunction;`)

Any better suggestions?

spring-integration-test/src/main/java/org/springframework/integration/test/context/SpringIntegrationTestExecutionListener.java
anthologia: No worries at all. I always appreciate your feedback.
Hope you had a good weekend despite the review interruption. 😊

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
artembilan: Oh! Yes, I see.
And yes, that function definitely can return null according to the `AbstractReplyProducingMessageHandler.handleRequestMessage()` contract.

Let's see if this helps:
```
protected Function<Message<?>, ? extends @Nullable Object> lastFunction;
```
And respectively, that `handleNextAndReply()`  contract.
Plus, probably `messageFunctions` signature.

artembilan (changes requested):

spring-integration-test/src/main/java/org/springframework/integration/test/mock/MockMessageHandler.java
anthologia: It looks really such an elegant approach !!
But still getting the same error.. Could be a limitation of the analyzer.

I apply your type signature improvement and keep the `@SuppressWarnings`.

BTW, the comment referencing issue looks like this is the exact same issue we're hitting.

PR #10303, opened 2025-08-12
Apply NullAbility to rsocket module

ngocnhan-tran1996:
Related to: #10083

### DIFF
--- a/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
+++ b/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
@@ -18,8 +18,6 @@ package org.springframework.integration.rsocket;
 
 import java.util.Map;
 
-import org.jspecify.annotations.Nullable;
-
 import org.springframework.core.io.buffer.DataBuffer;
 import org.springframework.integration.events.IntegrationEvent;
 import org.springframework.messaging.rsocket.RSocketRequester;
@@ -44,10 +42,10 @@ public class RSocketConnectedEvent extends IntegrationEvent {
 
 	private final DataBuffer data;
 
-	private final @Nullable RSocketRequester requester;
+	private final RSocketRequester requester;
 
 	public RSocketConnectedEvent(Object source, Map<String, Object> headers, DataBuffer data,
-			@Nullable RSocketRequester requester) {
+			RSocketRequester requester) {
 
 		super(source);
 		this.headers = headers;
@@ -63,7 +61,7 @@ public class RSocketConnectedEvent extends IntegrationEvent {
 		return this.data;
 	}
 
-	public @Nullable RSocketRequester getRequester() {
+	public RSocketRequester getRequester() {
 		return this.requester;
 	}
 
--- a/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
+++ b/spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
@@ -18,6 +18,8 @@ package org.springframework.integration.rsocket;
 
 import java.util.Map;
 
+import org.jspecify.annotations.Nullable;
+
 import org.springframework.core.io.buffer.DataBuffer;
 import org.springframework.integration.events.IntegrationEvent;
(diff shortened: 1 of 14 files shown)


spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/RSocketConnectedEvent.java
```diff
@@ -42,10 +44,10 @@ public class RSocketConnectedEvent extends IntegrationEvent {
 
 	private final DataBuffer data;
 
-	private final RSocketRequester requester;
+	private final @Nullable RSocketRequester requester;
 
 	public RSocketConnectedEvent(Object source, Map<String, Object> headers, DataBuffer data,
-			RSocketRequester requester) {
+			@Nullable RSocketRequester requester) {
```
artembilan: I don't think this is correct.
I think we should fix `ServerRSocketMessageHandler.handleConnectionSetup()` to assert that `rsocketRequester` is not null from those headers.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/ServerRSocketMessageHandler.java
```diff
@@ -60,15 +61,16 @@
 public class ServerRSocketMessageHandler extends IntegrationRSocketMessageHandler
 		implements ApplicationEventPublisherAware {
 
+	@SuppressWarnings("NullAway") // Reflection
 	private static final Method HANDLE_CONNECTION_SETUP_METHOD =
 			ReflectionUtils.findMethod(ServerRSocketMessageHandler.class, "handleConnectionSetup", Message.class);
 
-	private final Map<Object, RSocketRequester> clientRSocketRequesters = new HashMap<>();
+	private final Map<Object, @Nullable RSocketRequester> clientRSocketRequesters = new HashMap<>();
```
artembilan: After adding an assert on the `rsocketRequester` from headers, we won't need this one here.

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGateway.java
```diff
@@ -72,33 +72,33 @@
  */
 public class RSocketOutboundGateway extends AbstractReplyProducingMessageHandler {
 
+	@SuppressWarnings("NullAway.Init")
```
artembilan: I don't believe in this.
We do always set this from ctor.
What do I miss, please?

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGateway.java
```diff
@@ -72,33 +72,33 @@
...
 	private final Expression routeExpression;
 
-	private Object[] routeVars;
+	private Object[] routeVars = new Object[0];
```
artembilan: I think this can be `final` and initialized respectively from the ctor.

artembilan (changes requested):

spring-integration-rsocket/src/main/java/org/springframework/integration/rsocket/outbound/RSocketOutboundGateway.java
```diff
@@ -112,10 +108,25 @@ public RSocketOutboundGateway(String route, @Nullable Object... routeVariables)
...
+	/**
+	 * Instantiate based on the provided SpEL expression to evaluate an RSocket endpoint {@code route}
+	 * at runtime against a request message.
+	 * If route is a template and variables expansion is required, it is recommended to do that
+	 * in this expression evaluation, for example using some bean with an appropriate logic.
+	 * @param routeExpression the SpEL expression to use.
+	 * @param routeVariables the variables to expand route template.
+	 */
+	@SuppressWarnings("this-escape")
+	public RSocketOutboundGateway(Expression routeExpression, Object @Nullable [] routeVariables) {
```
artembilan: Please, revise this to be a `private` ctor.
I don't want to encourage users to follow this pattern.
It is really better to do that how it is explained in JavaDocs: fully via expression.
The expression is going to be evaluated against request message at runtime, and there you can build the route whatever way you need.

artembilan (changes requested):

PR #10305, opened 2025-08-12
polish `receive()` aop advice and its triggers

mjd507:
* remove redundant interface for class `CompoundTriggerAdvice`
* fix potential NPE in `DynamicPeriodicTrigger#nextExecution`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/util/DynamicPeriodicTrigger.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/util/DynamicPeriodicTrigger.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.util;
 
+import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/DynamicPeriodicTrigger.java
```diff
@@ -130,14 +131,15 @@ public boolean isFixedRate() {
...
-			return Instant.now().plus(this.initialDuration);
+			return clock.instant().plus(this.initialDuration);
 		}
 		else if (this.fixedRate) {
 			return lastScheduled.plus(this.duration);
 		}
-		return
-				triggerContext.lastCompletion().plus(this.duration); // NOSONAR never null here
+		Instant lastCompletion = triggerContext.lastCompletion();
+		return Objects.requireNonNullElse(lastCompletion, clock.instant()).plus(this.duration);
```
artembilan: I prefer `if..else` instead in critical path execution.

artembilan (changes requested):
Thank you for such a great refinement!

mjd507:
```txt
FileTests > testFileSplitterFlow() FAILED
    java.lang.AssertionError at FileTests.java:262
```
looks the error isn't related to the new changes, local file module is green,
can help trigger build again

PR #10306, opened 2025-08-12
GH-10083: Apply Nullability to TCP/IP module

artembilan:

* Add `@NullMarked` to every package
* Assume `BeanFactory` and `ApplicationEventPublisher` are not null in TCP/IP components. Therefore, fix all the failing tests to ensure those properties and their propagation via `afterPropertiesSet()`


spring-integration-ip/src/main/java/org/springframework/integration/ip/config/TcpConnectionFactoryFactoryBean.java
```diff
@@ -203,16 +210,16 @@ protected AbstractConnectionFactory createInstance() {
...
+		this.connectionFactory.setApplicationContext(this.applicationContext);
 		this.connectionFactory.afterPropertiesSet();
 		return this.connectionFactory;
 	}
 
 	private void setCommonAttributes(AbstractConnectionFactory factory) {
 		factory.setDeserializer(this.deserializer);
-		factory.setInterceptorFactoryChain(this.interceptorFactoryChain);
+		if (this.interceptorFactoryChain != null) {
+			factory.setInterceptorFactoryChain(this.interceptorFactoryChain);
```
cppwfs, line 221: Why is `interceptorFactoryChain` for the factory not allowed to be set to null?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
```diff
@@ -222,21 +220,20 @@ protected void doStart() {
 		if (this.clientConnectionFactory != null) {
 			this.clientConnectionFactory.start();
 		}
-		if (this.isClientMode) {
+		if (this.isClientMode && this.clientConnectionFactory != null) {
```
cppwfs, line 225: With this change we are allowing a `TCPInboundGatway` to be created without a `clientConnectionFactory`.   Before it would throw an `IllegalArgumentException`.   Is this a "What's new-able" entry?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
```diff
@@ -208,7 +211,6 @@ protected void doInit() {
 
 	@Override
 	protected Object handleRequestMessage(Message<?> requestMessage) {
-		Assert.notNull(this.connectionFactory, () -> getClass().getName() + " requires a client connection factory");
```
cppwfs, line 211: Nice!

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractConnectionFactory.java
```diff
@@ -593,23 +594,25 @@ public void stop() {
...
-						executorService.shutdownNow();
+				if (executorService != null) {
+					executorService.shutdown();
+					try {
 						if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) { // NOSONAR magic number
-							logger.debug("Executor failed to shutdown");
+							logger.debug("Forcing executor shutdown");
+							executorService.shutdownNow();
+							if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) { // NOSONAR magic number
+								logger.debug("Executor failed to shutdown");
```
cppwfs, line 604: Should we change this debug message to say something like "Executor failed to shutdown on retry"

spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UnicastSendingMessageHandler.java
```diff
@@ -566,7 +568,9 @@ public void run() {
 	 * (bind) error occurred, without bouncing the JVM.
 	 */
 	public void restartAckThread() {
-		this.taskExecutor.execute(this);
+		if (this.taskExecutor != null) {
+			this.taskExecutor.execute(this);
```
cppwfs: Should we have a debug statement stating the case when the `taskExecutor` is not present and the execute wasn't tried?

cppwfs (changes requested):
The usual annoying questions ;-)

spring-integration-ip/src/main/java/org/springframework/integration/ip/config/TcpConnectionFactoryFactoryBean.java
artembilan: I prefer to not allow to set null and to avoid runtime mutation.
But that is clearly just my preference.
We can revise this when someone complains with a good reason 😄

spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/UnicastSendingMessageHandler.java
artembilan: I'll take a look into this.
Good catch!
Looks like this is a public API which has to be protected this or other way.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
artembilan: I'll add an assert into `onInit()` to ensure that one of the connection factories is provided.
Then this condition would be just to satisfy nullability 😄

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractConnectionFactory.java
artembilan: I don't think such a detail makes sense in this context.
We just failed to shutdown it in time. That's it.
Otherwise it would raise a question like: what is "retry" ?

cppwfs (approved):
LGTM!   Thank you!

Will merge after successful build!

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
```diff
@@ -128,35 +128,33 @@ private boolean doOnMessage(Message<?> message) {
...
-		try {
-			connection.send(reply);
-		}
-		catch (Exception ex) {
-			logger.error(ex, "Failed to send reply");
+			TcpConnection connection = this.connections.get(connectionId);
+			if (connection == null) {
+				publishNoConnectionEvent(message, connectionId);
+				logger.error(() -> "Connection not found when processing reply " + reply + " for " + message);
+				return false;
```
mrpiggi, line 136: Just curious: This method always returns false, but I didn't look into detail. If this is desired, I wondered if for the sake of readability, this one should be changed to void.

artembilan: Good catch, @mrpiggi !

The contract is this:
```
@FunctionalInterface
public interface TcpListener {

	/**
	 * Called by a TCPConnection when a new message arrives.
	 * @param message The message.
	 * @return true if the message was intercepted
	 */
	boolean onMessage(Message<?> message);

}
```
So, we need to revise and return `true` at some point.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractConnectionFactory.java
```diff
@@ -796,12 +799,14 @@ private void keyReadable(final Selector selector, final long now, final Selectio
...
 	}
 
-	private void keyAcceptable(final Selector selector, ServerSocketChannel server, final long now) {
-		try {
-			doAccept(selector, server, now);
-		}
-		catch (Exception ex) {
-			logger.error(ex, "Exception accepting new connection(s)");
+	private void keyAcceptable(final Selector selector, @Nullable ServerSocketChannel server, final long now) {
+		if (server != null) {
```
mrpiggi, line 803: Just nitpicking: I tend to prefer `if (server == null) { return; }` as this reduces the level of indentation

artembilan: I think we had something like that before, but SonarQube has complained about unnecessary `return;` in the method. 🤷 
The `if (server != null) {` also makes less code lines.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpInboundGateway.java
artembilan: FYI: https://github.com/spring-projects/spring-integration/issues/10314

artembilan: FYI: https://github.com/spring-projects/spring-integration/pull/10315

PR #10307, opened 2025-08-13
Fix Kotlin DSL according to IDEA errors and suggestions

artembilan:
It is not clear why it does not fail with Gradle, but right now it is not possible to build the project using IDEA compiler.

Some of the suggestions are good and have led to the Nullability fixes in related Java DSL classes. The `Any` extension requirement for the generic argument of reified types is not clear, but works. The requirement to refine nullable Kotlin types is good.

Somewhat related to: https://github.com/spring-projects/spring-integration/issues/10083

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/TransformerEndpointSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/TransformerEndpointSpec.java
@@ -140,7 +140,7 @@ public class TransformerEndpointSpec extends ConsumerEndpointSpec<TransformerEnd
 	 * @param <P> the type ot expect.
 	 * @return the spec.
 	 */
-	public <P> TransformerEndpointSpec expectedType(@Nullable Class<P> expectedType) {
+	public <P> TransformerEndpointSpec expectedType(Class<P> expectedType) {
 		this.expectedType = expectedType;
 		return this;
 	}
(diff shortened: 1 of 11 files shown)


spring-integration-core/src/main/java/org/springframework/integration/dsl/TransformerEndpointSpec.java
```diff
@@ -140,7 +140,7 @@ public <P, T> TransformerEndpointSpec transformer(GenericTransformer<P, T> trans
 	 * @param <P> the type ot expect.
 	 * @return the spec.
 	 */
-	public <P> TransformerEndpointSpec expectedType(@Nullable Class<P> expectedType) {
+	public <P> TransformerEndpointSpec expectedType(Class<P> expectedType) {
```
cppwfs, line 143: Question:
Is this a breaking change if we state that `expectedType` can not be reset to null.  If so should we have a note about it?

cppwfs (changes requested):
Looks great, I was a bit rusty on my Kotlin...
Just one simple question.

spring-integration-core/src/main/java/org/springframework/integration/dsl/TransformerEndpointSpec.java
artembilan: Reset?
This is part of Java DSL.
So, or you call this method in builder chain or not.
But if you do, it has to come with the value.
Why would one do:
```
.expectedType(String.class)
.expectedType(null)
```

?

cppwfs: I agree it doesn't make sense.   But someone can do it.   
Its not a show stopper in any shape.  I just like to be annoying. 😆

cppwfs (approved):
LGTM!   Thanks for investigating and fixing this.

PR #10308, opened 2025-08-13
GH-10083: Apply Nullability to FTP module

artembilan:

### DIFF
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/config/FtpOutboundChannelAdapterParser.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/config/FtpOutboundChannelAdapterParser.java
@@ -16,6 +16,8 @@
 
 package org.springframework.integration.ftp.config;
 
+import java.util.Objects;
+
 import org.w3c.dom.Element;
 
 import org.springframework.beans.factory.config.BeanDefinition;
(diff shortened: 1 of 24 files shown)


spring-integration-ftp/src/main/java/org/springframework/integration/ftp/config/FtpOutboundChannelAdapterParser.java
```diff
@@ -55,7 +57,8 @@ protected void postProcessBuilder(BeanDefinitionBuilder builder, Element element
 				.iterator()
 				.next()
 				.getValue();
-		templateDefinition.getPropertyValues() // NOSONAR never null
+
+		Objects.requireNonNull(templateDefinition).getPropertyValues()
```
cppwfs, line 61: Would it be correct to say that we don't expect the templateDefinition returned from the builder should never be null?

cppwfs (changes requested):
LGTM 

Just a quick question.

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/config/FtpOutboundChannelAdapterParser.java
artembilan: What you say is correct , but contract enforces us.
I wouldn’t spend too much time on XML parsers: I anticipate their deprecation and removal at some point.
Plus , this is not a critical execution path, so it is OK to say nothing and just follow the contract.

cppwfs (approved):
LGTM
Thank you for doing this!

PR #10311, opened 2025-08-13
GH-10309: Don't register bean for internal `GatewayMH.gatewayProxyFac…

### LINKED ISSUE #10309: ConcurrentModificationException When IntegrationFlow uses gateway

6.5.1

**Describe the bug**

It appears that `MockIntegrationContext.autoStartupCandidates` is updated when `SpringIntegrationTestExecutionListener#prepareTestInstance` is called causing a `ConcurrentModificationException`. This happens because the method's iteration over `AbstractEndpoint`s triggers `GatewayMessageHandler#initialize` for any `IntegrationFlows` with a gateway, which in turn calls `MockIntegrationContext#postProcessBeforeInitialization` to add a new candidate.

Note: in https://github.com/spring-projects/spring-integration/issues/10018 the `GatewayMessageHandler#start` method was updated which indirectly results in `MockIntegrationContext#postProcessBeforeInitialization` to be called so I did not see this issue in Spring Integration 6.4.x.

**To Reproduce**

1. Create an IntegrationFlow where it uses a gateway.
2. Create a Test with annotations `@SpringBootTest` and `@SpringIntegrationTest`.
3. Run the test and you will see `java.util.ConcurrentModificationException` from `org.springframework.integration.test.context.SpringIntegrationTestExecutionListener#prepareTestInstance`

**Expected behavior**

The test should not run into a `ConcurrentModificationException`. Prior to https://github.com/spring-projects/spring-integration/issues/10018 we did not call `MockIntegrationContext#postProcessBeforeInitialization` when a gateway was found on an `IntegrationFlow`.

**Sample**

[Sample Project](https://github.com/venkatg-v/spring-integration-test-gateway-issue)

artembilan:
…toryBean`

Fixes: https://github.com/spring-projects/spring-integration/issues/10309

The call of the `BeanFactory.initializeBean()` on the internal `GatewayProxyFactoryBean` instance of the `GatewayMessageHandler` causes an extra bean to be processed everywhere. Since such a registration happens from the `GatewayMessageHandler.start()`, the `SpringIntegrationTestExecutionListener#prepareTestInstance` may suffer from a `ConcurrentModificationException` on the `autoStartupCandidates`.
Just because its logic is to call `start()` from that `autoStartupCandidates` collection iteration.

Essentially, we don't need that since a `GatewayMessageHandler` is a bean itself. The `GatewayProxyFactoryBean` is used internally to avoid logic duplication of the proxy creating on the provided interface.

* Rework the logic of the `GatewayMessageHandler` to call all the respective `BeanFactory` call-backs manually.
* Add an `IntegrationFlow` bean with a `gateway()` into the `MockMessageHandlerTests` to ensure that `SpringIntegrationTestExecutionListener` does not suffer from a `ConcurrentModificationException` anymore.

**Auto-cherry-pick to `6.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/gateway/GatewayMessageHandler.java
@@ -25,7 +25,6 @@ import java.util.concurrent.locks.ReentrantLock;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.factory.BeanCreationException;
-import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.core.task.SimpleAsyncTaskExecutor;
 import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
 import org.springframework.integration.support.management.ManageableLifecycle;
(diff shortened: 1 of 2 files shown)


cppwfs (approved):
LGTM
Good idea!

PR #10312, opened 2025-08-13
GH-10083: Fix Nullability in some missed packages

artembilan:

* Turn `EntryEventMessagePayload` into `record`. Fix access from tests, respectively
* Extract `ParameterExpressionEvaluator` into a top-level class as it is expected by the `ExpressionEvaluatingParameterSource`

### DIFF
--- a/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/support/parametersource/ExpressionEvaluatingParameterSourceFactory.java
+++ b/spring-integration-jpa/src/main/java/org/springframework/integration/jpa/support/parametersource/ExpressionEvaluatingParameterSourceFactory.java
@@ -141,7 +141,9 @@ public class ExpressionEvaluatingParameterSourceFactory implements ParameterSour
 
 		private @Nullable Object obtainParameterValue(JpaParameter jpaParameter) {
 			Object value = jpaParameter.getValue();
-			if (value == null && jpaParameter.getExpression() != null) {
+			if (value == null) {
+				Assert.notNull(jpaParameter.getExpression(),
+						() -> "One of the 'value' or 'expression' must be provided for 'JpaParameter': " + jpaParameter);
 				Expression expression;
 				if (this.input instanceof Collection<?>) {
 					expression = jpaParameter.getProjectionExpression();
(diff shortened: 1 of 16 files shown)


spring-integration-jpa/src/main/java/org/springframework/integration/jpa/support/parametersource/ExpressionEvaluatingParameterSourceFactory.java
```diff
@@ -139,23 +139,21 @@ public Object getValue(String paramName) {
...
-		@Nullable
-		private Object obtainParameterValue(JpaParameter jpaParameter) {
-			Object value = null;
-			if (jpaParameter.getValue() != null) {
-				value = jpaParameter.getValue();
-			}
-			if (jpaParameter.getExpression() != null) {
+		private @Nullable Object obtainParameterValue(JpaParameter jpaParameter) {
+			Object value = jpaParameter.getValue();
+			if (value == null && jpaParameter.getExpression() != null) {
```
cppwfs: I agree with change.    Should we add this as a note, since it is a change in behavior?

cppwfs (changes requested):
LGTM
Just one silly question.

spring-integration-jpa/src/main/java/org/springframework/integration/jpa/support/parametersource/ExpressionEvaluatingParameterSourceFactory.java
artembilan: I don’t think so. 
I tried to keep outcome of the method the same. Semantically logic is not changed. Just perspective how the value is handled .
What do you think I have broken since it gives an impression that behavior has changed?

cppwfs: The original behavior would allow the expression to be evaluated regardless if `value` was `null` or not (and expression was not `null`).   Now it is only evaluated if `value`  is `null` and `expression` is not `null`.   

I'm wondering if people built their code around the fact that they would prefer to get the  result from a `expression` evaluation even if the `value` is not `null`.

artembilan: Oh! I see. I thought that is `else if`.
Let me look into this again .
Thank you for spotting!

artembilan: So, the contract is this:
```
	/**
	 * Instantiates a new Jpa Parameter without a name. This is useful for specifying
	 * positional Jpa parameters.
	 * @param value If null, the expression property must be set
	 * @param expression If null, the value property must be set
	 */
	public JpaParameter(@Nullable Object value, @Nullable String expression) {
```
Therefore an existing logic is wrong: we really should not go and check for `getExression()` if `getValue()` is not null.
It might rely on assumption that expression is not set, but I don't think we can rely on that.
With that I think it is better to honor that contract even more.

PR #10315, opened 2025-08-14
GH-10314: Revise `TcpListener.onMessage()` contract for `void`

### LINKED ISSUE #10314: Revise `TcpListener.onMessage()` contract
Looks like the `boolean` return from this method is ignored anywhere in the project.
Many places just always returns `false`.
Plus in some places this method called as like it is `void`.

The suggestion is to bite a bullet and change this contract to `void`.
Even if this is a breaking change, it should be OK for the current major version.

CC: @garyrussell

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10314

Looks like a `boolean` return for the `TcpListener.onMessage()` is a leftover of some earlier design or some idea which didn't make it into the project. On the other hand, all the logic in the project around this `onMessage()` usage is properly handled by the delegation or exceptions.

* Change `TcpListener.onMessage()` to have a `void` as return type
* Fix all the respective usages and simplify some implementations with removing those bogus `return false;` and using proper `if..else` logic if necessary
* Fix some code style in the affected classes, like proper `assertThat` or diamond operators

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
@@ -323,35 +323,32 @@ public class TcpOutboundGateway extends AbstractReplyProducingMessageHandler
 	}
 
 	@Override
-	public boolean onMessage(Message<?> message) {
+	public void onMessage(Message<?> message) {
 		String connectionId = message.getHeaders().get(IpHeaders.CONNECTION_ID, String.class);
 		if (connectionId == null) {
 			if (unsolicitedSupported(message)) {
-				return false;
+				return;
 			}
 			logger.error("Cannot correlate response - no connection id");
 			publishNoConnectionEvent(message, null, "Cannot correlate response - no connection id");
-			return false;
+			return;
 		}
 		logger.trace(() -> "onMessage: " + connectionId + "(" + message + ")");
 		AsyncReply reply = this.pendingReplies.get(connectionId);
 		if (reply == null) {
-			if (message instanceof ErrorMessage) {
-				/*
-				 * Socket errors are sent here, so they can be conveyed to any waiting thread.
-				 * If there's not one, simply ignore.
-				 */
-				return false;
-			}
-			else {
+			/*
+			 * Socket errors are sent here, so they can be conveyed to any waiting thread.
+			 * If there's not one, simply ignore.
+			 */
+			if (!(message instanceof ErrorMessage)) {
 				if (unsolicitedSupported(message)) {
-					return false;
+					return;
 				}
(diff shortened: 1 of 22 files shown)


spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
```diff
@@ -489,7 +485,8 @@ boolean isHaveSemaphore() {
 		 * Sender blocks here until the reply is received, or we time out.
 		 * @return The return message or null if we time out
 		 */
-		@Nullable Message<?> getReply() {
+		@Nullable
+		Message<?> getReply() {
```
cppwfs, line 489: Why is the `@Nullable` move to the line above the method?

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
```diff
@@ -250,12 +248,16 @@ public void removeDeadConnection(TcpConnection connection) {
...
+					&& !this.theConnection.equals(this)) {
+
 				tcpConnectionInterceptorSupport.removeDeadConnection(this);
 			}
 			TcpSender sender = getSender();
-			if (sender != null && this.interceptedSenders != null && !(sender instanceof TcpConnectionInterceptorSupport)) {
-				this.interceptedSenders.forEach(snder -> snder.removeDeadConnection(connection));
+			if (sender != null && this.interceptedSenders != null
+					&& !(sender instanceof TcpConnectionInterceptorSupport)) {
+
```
cppwfs, line 259: Blank line can probably be removed.

cppwfs (changes requested):
LGTM!   
Just a couple of formatting nitpicks.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/TcpOutboundGateway.java
artembilan: Because this method does not have modifiers and annotation is treated by formatter as for the whole method.
I agree on consistency discrepancy here, but I prefer to keep formatter happy to avoid unnecessary changes in the future .

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/TcpConnectionInterceptorSupport.java
artembilan: Well, my preference for better readability is to have a blank line after multi-line `if`.
Same reason as with a blank line before method body after multi-line header.

cppwfs (approved):
I agree with your response.

Things look great!

PR #10317, opened 2025-08-15
GH-10083: Apply Nullability to STMOP module

anthologia:

### DIFF
--- a/spring-integration-stomp/src/main/java/org/springframework/integration/stomp/WebSocketStompSessionManager.java
+++ b/spring-integration-stomp/src/main/java/org/springframework/integration/stomp/WebSocketStompSessionManager.java
@@ -39,8 +39,6 @@ import org.springframework.web.socket.messaging.WebSocketStompClient;
  */
 public class WebSocketStompSessionManager extends AbstractStompSessionManager {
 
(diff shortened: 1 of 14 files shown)


spring-integration-stomp/src/main/java/org/springframework/integration/stomp/WebSocketStompSessionManager.java
```diff
@@ -37,17 +39,20 @@
...
-	private volatile WebSocketHttpHeaders handshakeHeaders;
+	private volatile @Nullable WebSocketHttpHeaders handshakeHeaders;
 
 	public WebSocketStompSessionManager(WebSocketStompClient webSocketStompClient, String url, Object... uriVariables) {
 		super(webSocketStompClient);
 		Assert.hasText(url, "'url' must not be empty.");
 		this.url = url;
-		this.uriVariables = uriVariables != null ? Arrays.copyOf(uriVariables, uriVariables.length) : null;
+		this.uriVariables = uriVariables != null ?
+				Arrays.copyOf(uriVariables, uriVariables.length) : EMPTY_URI_VARIABLES;
```
artembilan: I don't think that we need a dedicated constant for empty array.

artembilan (changes requested):

PR #10318, opened 2025-08-15
GH-10083: Apply Nullability to core `util` package

anthologia:

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -561,6 +561,7 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		this.logger.debug(() -> "Handling message with correlationKey [" + correlationKey + "]: " + message);
 
 		UUID groupIdUuid = UUIDConverter.getUUID(correlationKey);
+		Assert.notNull(groupIdUuid, "groupIdUuid must not be null.");
 		Lock lock = this.lockRegistry.obtain(groupIdUuid.toString());
 
 		boolean noOutput = true;
@@ -780,8 +781,10 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	protected void forceComplete(MessageGroup group) { // NOSONAR Complexity
 		Object correlationKey = group.getGroupId();
+		Assert.notNull(correlationKey, "'correlationKey' must not be null");
 		// UUIDConverter is no-op if already converted
 		UUID groupId = UUIDConverter.getUUID(correlationKey);
+		Assert.notNull(groupId, "groupId is must not be null");
 		Lock lock = this.lockRegistry.obtain(groupId.toString());
 		boolean removeGroup = true;
 		boolean noOutput = true;
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -781,7 +781,6 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 
 	protected void forceComplete(MessageGroup group) { // NOSONAR Complexity
 		Object correlationKey = group.getGroupId();
-		Assert.notNull(correlationKey, "'correlationKey' must not be null");
 		// UUIDConverter is no-op if already converted
 		UUID groupId = UUIDConverter.getUUID(correlationKey);
 		Assert.notNull(groupId, "groupId is must not be null");
--- a/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
@@ -561,7 +561,6 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		this.logger.debug(() -> "Handling message with correlationKey [" + correlationKey + "]: " + message);
 
 		UUID groupIdUuid = UUIDConverter.getUUID(correlationKey);
-		Assert.notNull(groupIdUuid, "groupIdUuid must not be null.");
 		Lock lock = this.lockRegistry.obtain(groupIdUuid.toString());
 
 		boolean noOutput = true;
@@ -783,7 +782,6 @@ public abstract class AbstractCorrelatingMessageHandler extends AbstractMessageP
 		Object correlationKey = group.getGroupId();
 		// UUIDConverter is no-op if already converted
 		UUID groupId = UUIDConverter.getUUID(correlationKey);
-		Assert.notNull(groupId, "groupId is must not be null");
 		Lock lock = this.lockRegistry.obtain(groupId.toString());
 		boolean removeGroup = true;
 		boolean noOutput = true;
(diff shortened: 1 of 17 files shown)


spring-integration-core/src/main/java/org/springframework/integration/util/ClassUtils.java
```diff
@@ -180,6 +188,7 @@ public static Class<?> findClosestMatch(Class<?> type, Set<Class<?>> candidates,
 				closestMatch = candidate;
 			}
 			else if (failOnTie && typeDiffWeight < Integer.MAX_VALUE && (typeDiffWeight == minTypeDiffWeight)) {
+				Objects.requireNonNull(closestMatch, "closestMatch should not be null at this point");
```
artembilan: Instead of this this do a ternary operator in the place of `closestMatch.getName()`.
There is a case when `if (typeDiffWeight < minTypeDiffWeight) {` is not true.

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -780,8 +781,10 @@ protected void afterRelease(MessageGroup group, Collection<Message<?>> completed
 
 	protected void forceComplete(MessageGroup group) { // NOSONAR Complexity
 		Object correlationKey = group.getGroupId();
+		Assert.notNull(correlationKey, "'correlationKey' must not be null");
 		// UUIDConverter is no-op if already converted
 		UUID groupId = UUIDConverter.getUUID(correlationKey);
+		Assert.notNull(groupId, "groupId is must not be null");
```
artembilan: It won't be null after `Assert.notNull(correlationKey)`.
Let's consider to mark those `UUIDConverter` methods with `@Contract("null -> null; !null -> !null")`.
See `StringUtils.quote()`, for example.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresChannelMessageTableSubscriber.java
```diff
@@ -298,7 +299,9 @@ public boolean isRunning() {
 	}
 
 	private static String getKey(Object input) {
-		return UUIDConverter.getUUID(input).toString();
+		UUID uuid = UUIDConverter.getUUID(input);
+		Assert.notNull(uuid, "UUID must not be null");
+		return uuid.toString();
```
artembilan: I think after `@Contract("null -> null; !null -> !null")` in the `UUIDConverter`, we won't need these changes here.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -152,7 +153,9 @@ public DistributedLock obtain(Object lockKey) {
 	}
 
 	protected String pathFor(String input) {
-		return UUIDConverter.getUUID(input).toString();
+		UUID uuid = UUIDConverter.getUUID(input);
+		Assert.notNull(uuid, "UUID must not be null");
+		return uuid.toString();
```
artembilan: DITTO

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcChannelMessageStore.java
```diff
@@ -481,7 +481,9 @@ public MessageGroup addMessageToGroup(Object groupId, final Message<?> message)
 	 * @return Returns the UUID as String.
 	 */
 	private String getKey(Object input) {
-		return UUIDConverter.getUUID(input).toString();
+		UUID uuid = UUIDConverter.getUUID(input);
+		Assert.notNull(uuid, "UUID must not be null");
+		return uuid.toString();
```
artembilan: DITTO

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/JdbcMessageStore.java
```diff
@@ -772,7 +772,9 @@ private void updateMessageGroup(String groupId) {
 	}
 
 	private String getKey(Object input) {
-		return UUIDConverter.getUUID(input).toString();
+		UUID uuid = UUIDConverter.getUUID(input);
+		Assert.notNull(uuid, "UUID must not be null");
+		return uuid.toString();
```
artembilan: DITTO

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/ClassUtils.java
anthologia: I still believe `closestMatch` should not be null at this point, but I might be missing something.

To reach this `else if`, we need `typeDiffWeight == minTypeDiffWeight && typeDiffWeight < Integer.MAX_VALUE`, which would mean `minTypeDiffWeight` was already updated from its initial `Integer.MAX_VALUE` in a previous iteration where `closestMatch` was assigned.

Flow is a bit delicate though... 😅

artembilan: OK.
But it is better to avoid extra method call for nothing.
I believe the ternary operator is much cheaper than this `Objects.requireNonNull`

spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
```diff
@@ -89,7 +94,7 @@ private static UUID fromStringBytes(String input) {
 		return UUID.nameUUIDFromBytes(input.getBytes(StandardCharsets.UTF_8));
 	}
 
-	private static byte[] serialize(Object object) {
+	private static byte @Nullable [] serialize(@Nullable Object object) {
```
artembilan, line 98: Don't we need to add `@Contract("null -> null; !null -> !null")` on this method as well?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
artembilan: And why do we still have this here?

spring-integration-core/src/main/java/org/springframework/integration/aggregator/AbstractCorrelatingMessageHandler.java
```diff
@@ -561,6 +561,7 @@ protected void handleMessageInternal(Message<?> message) {
 		this.logger.debug(() -> "Handling message with correlationKey [" + correlationKey + "]: " + message);
 
 		UUID groupIdUuid = UUIDConverter.getUUID(correlationKey);
+		Assert.notNull(groupIdUuid, "groupIdUuid must not be null.");
```
artembilan: Same question here: if `correlationKey` is not null (according to the `Assert.state(correlationKey != null)`), then we don't need this extra check here.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/util/UUIDConverter.java
anthologia: Sorry for low quality.. I think it's too late for me.. 😴

Anyway, I've tried to apply `@Contract` to `MessagingAnnotationUtils` as well.
Please let me know if there's anything to fix !

I'll review this and Kafka Jackson issue over the weekend.

artembilan: No problem!
When I'm done with HTTP module, I'll pull your PR locally for final review, clean up and possible merge.
Thank you and have a good weekend!

artembilan:
Merged as: https://github.com/spring-projects/spring-integration/commit/1bf58100a165fdd54f3d10d79a2e3b625ef4e621.

Nothing critical I have polished in the final commit.

Thank you!

PR #10319, opened 2025-08-15
GH-10083: Apply Nullability to HTTP module

artembilan:

The change affects some core classes, since `getBeanName()` cannot be null after a proper bean initialization

* Add slight code flow improvement in the `BaseHttpInboundEndpoint` implementations regarding `validate()` method
* Fix `DefaultHttpHeaderMapper` tests where a `ConversionService` is required

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/config/ControlBusControllerConfiguration.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/config/ControlBusControllerConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.integration.http.config;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.config.BeanDefinition;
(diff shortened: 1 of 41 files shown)


spring-integration-http/src/main/java/org/springframework/integration/http/config/ControlBusControllerConfiguration.java
```diff
@@ -56,7 +58,7 @@ ControlBusController controlBusController(ControlBusCommandRegistry controlBusCo
 
 		controlBusCommandRegistry.setEagerInitialization(true);
 
-		return new ControlBusController(controlBusCommandRegistry, conversionService.getIfUnique());
+		return new ControlBusController(controlBusCommandRegistry, conversionService.getObject());
```
cppwfs, line 61: Just curious.   Swapping `getIfUnique` for `getObject` is because if there is no object it will throw an Exception instead returning null?

cppwfs (approved):
LGTM.   Just had a simple question.

spring-integration-http/src/main/java/org/springframework/integration/http/config/ControlBusControllerConfiguration.java
artembilan: I see your point.
I understand that it is a behavior change, but it should be a good one.
The `ControlBusController` cannot live without `conversionService`, so `getIfUnique()` is not OK for us since it may end up with a `null`.
Therefore I decide to go a hard failure with just `getObject()` to let the target application know that there is something off with a configuration.
In 99% it should not be a problem since most projects rely on the MVC auto-configuration and that's it.
They don't introduce a custom `FormattingConversionService`.
When they do, and this fails, they should mark one of the  as `@Primary` to avoid injection collision.

With that I think just `FormattingConversionService conversionService` as injection point in this `ControlBusControllerConfiguration` should be enough.
The `ObjectProvider.getObject()` is total equivalent to regular by type injection.

Let me know if you have any other concerns before I go with clean up!

PR #10320, opened 2025-08-16
GH-10083: Apply Nullability to core `events`, `leader`, `message` and `resource` packages

ngocnhan-tran1996:
Related to: #10083


spring-integration-core/src/main/java/org/springframework/integration/resource/ResourceRetrievingMessageSource.java
```diff
@@ -47,11 +49,13 @@ public class ResourceRetrievingMessageSource extends AbstractMessageSource<Resou
...
 	private final String pattern;
 
+	@SuppressWarnings("NullAway.Init")
 	private volatile ApplicationContext applicationContext;
 
+	@SuppressWarnings("NullAway.Init")
 	private volatile ResourcePatternResolver patternResolver;
 
-	private volatile CollectionFilter<Resource> filter;
+	private volatile @Nullable CollectionFilter<Resource> filter;
```
artembilan: ... and remove `volatile` on all the properties.
It is really an anti-pattern to change state of the component at runtime.
Therefore, we treat all of them as state fixed after initialization phase.
Therefore, no need in extra ticks over `volatile` barrier.

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
```diff
@@ -837,6 +838,9 @@ public AdviceMessage<?> convert(Document source) {
 					throw new IllegalStateException("failed to load class: " + inputMessageType, e);
 				}
 			}
+			else {
+				inputMessage = MessageBuilder.withPayload(new byte[0]).build();
```
artembilan: I think it is better to throw an exception instead.
How that happened that `AdviceMessage` was serialized without an `inputMessage`?
So, the state is illegal.
WDYT?

artembilan (changes requested):

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
ngocnhan-tran1996: That makes sense. I wasn’t sure about the behavior in this case, so I decided to initialize the value.

ngocnhan-tran1996:
Can we re-run the build? I haven’t touched the related context package.

artembilan:
> I haven’t touched the related context package.

I know. 
That looks like Gradle build cache hallucination.
We made a change in the `core` module, and therefore all others have to be rebuild.
But apparently we got an `ip` one from cache before, so it didn't fail for a couple days.
Now, it looks like something else has triggered cache invalidation, so we are failing.
The fix for `core` is in, therefore I have just merged your PR.

Thank you!

PR #10340, opened 2025-08-17
GH-10264: Add Jackson 3 support to Kafka module

### LINKED ISSUE #10264: Add Jackson 3 support to Kafka module
Could be possible when related issue is fixed in Spring for Apache Kafka: https://github.com/spring-projects/spring-kafka/issues/3930

anthologia:
Fixes: https://github.com/spring-projects/spring-integration/issues/10264

* Update all Kafka components to conditionally use Jackson 3 classes
* Migrate test configurations and XML beans to use `JsonKafkaHeaderMapper`
* Suppress removal warning in `KafkaProducerMessageHandler`

### DIFF
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
@@ -34,6 +34,7 @@ import org.springframework.kafka.listener.adapter.RecordMessagingMessageListener
 import org.springframework.kafka.support.Acknowledgment;
 import org.springframework.kafka.support.DefaultKafkaHeaderMapper;
 import org.springframework.kafka.support.JacksonPresent;
+import org.springframework.kafka.support.JsonKafkaHeaderMapper;
 import org.springframework.kafka.support.converter.MessagingMessageConverter;
 import org.springframework.kafka.support.converter.RecordMessageConverter;
 import org.springframework.messaging.MessageHandler;
@@ -45,6 +46,7 @@ import org.springframework.util.Assert;
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Jooyoung Pyoung
  *
  * @since 5.4
  *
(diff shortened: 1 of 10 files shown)


spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
```diff
@@ -84,7 +86,16 @@ public SubscribableKafkaChannel(KafkaOperations<?, ?> template, KafkaListenerCon
 		Assert.notNull(factory, "'factory' cannot be null");
 		this.factory = factory;
 
-		if (JacksonPresent.isJackson2Present()) {
+		if (JacksonPresent.isJackson3Present()) {
+			var messageConverter = new MessagingMessageConverter();
+			var headerMapper = new JsonKafkaHeaderMapper();
+			headerMapper.addTrustedPackages(
+					org.springframework.integration.support.json.JacksonMessagingUtils.DEFAULT_TRUSTED_PACKAGES
```
artembilan: Why do we need fully qualified class name here?
If that is deprecated one, then consider to use a new API on the matter.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -108,7 +110,16 @@ public KafkaInboundGateway(AbstractMessageListenerContainer<K, V> messageListene
 		this.messageListenerContainer.setAutoStartup(false);
 		this.kafkaTemplate = kafkaTemplate;
 		setErrorMessageStrategy(new RawRecordHeaderErrorMessageStrategy());
-		if (JacksonPresent.isJackson2Present()) {
+		if (JacksonPresent.isJackson3Present()) {
+			MessagingMessageConverter messageConverter = new MessagingMessageConverter();
+			JsonKafkaHeaderMapper headerMapper = new JsonKafkaHeaderMapper();
+			headerMapper.addTrustedPackages(
+					org.springframework.integration.support.json.JacksonMessagingUtils.DEFAULT_TRUSTED_PACKAGES
```
artembilan: DITTO

artembilan (changes requested):

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/channel/SubscribableKafkaChannel.java
anthologia: I’ll fix the FQCN !

As for the deprecated constant — I already left a comment. https://github.com/spring-projects/spring-integration/pull/10263#issuecomment-3197106446.

I'm not sure... but because of that issue, we can’t switch to the new API here for now..

artembilan: I'm not sure what you are talking about.
The `JacksonMessagingUtils` are for Jackson 3 and used in this Jackson 3 specific code block.
And this is indeed new.
So, all good but we don't need an FQCN here.
😄

anthologia: Ah, got it 😅 I misunderstood — I thought the "deprecated one ..." meant replacing `JacksonJsonUtils` with `JacksonMessagingUtils`.

Thanks !

PR #10341, opened 2025-08-19
GH-10083: Apply Nullability to `core` `store` package

artembilan:

* Update to `io.spring.nullability:0.0.4`
* Also apply Nullability to `core` `support.locks` package
* Add `@Nullable` to generic argument of the `CheckedCallable` and `CheckedFunction`
* To not use delegation in the `LockRegistry` to avoid unnecessary nullability carry-on
* Remove redundant ctors from the `MessageHolder` and `MessageMetadata`. Use `@JsonCreator` instead on the parameterized one
* Remove unnecessary now `Assert.notNull()` in the `MessageStore` methods

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
@@ -212,7 +212,7 @@ public abstract class AbstractKeyValueMessageStore extends AbstractMessageGroupS
 			doAddMessage(message, groupId);
 			if (metadata != null) {
 				UUID id = message.getHeaders().getId();
-				Assert.notNull(id, "'id' must not be null");
+				Assert.state(id != null, () -> "Message 'id' must not be null: " + message);
 				metadata.add(id);
 			}
 			else {
@@ -243,7 +243,7 @@ public abstract class AbstractKeyValueMessageStore extends AbstractMessageGroupS
 			List<UUID> ids = new ArrayList<>();
 			for (Message<?> messageToRemove : messages) {
 				UUID id = messageToRemove.getHeaders().getId();
-				Assert.notNull(id, "Message 'id' must not be null");
+				Assert.state(id != null, () -> "Message 'id' must not be null: " + messageToRemove);
 				ids.add(id);
 			}
 
@@ -359,7 +359,6 @@ public abstract class AbstractKeyValueMessageStore extends AbstractMessageGroupS
 	}
 
 	private @Nullable Message<?> removeMessageFromGroup(UUID id, Object groupId) {
-		Assert.notNull(id, "'id' must not be null");
 		Object object = doRemove(this.messagePrefix + groupId + '_' + id);
 		if (object != null) {
 			return extractMessage(object);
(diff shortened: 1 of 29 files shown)


spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
```diff
@@ -135,27 +133,24 @@ public MessageMetadata getMessageMetadata(UUID messageId) {
...
 
 	protected void doAddMessage(Message<?> message) {
 		doAddMessage(message, null);
 	}
 
 	protected void doAddMessage(Message<?> message, @Nullable Object groupId) {
-		Assert.notNull(message, "'message' must not be null");
 		UUID messageId = message.getHeaders().getId();
 		Assert.notNull(messageId, "Cannot store messages without an ID header");
 		String messageKey = this.messagePrefix + (groupId != null ? groupId.toString() + '_' : "") + messageId;
```
cppwfs: This maybe a bit nitpicky.  But can we use a `StringBuilder` instead of `+`, since this can be called quiet a bit?

spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupStoreReaper.java
```diff
@@ -37,10 +37,11 @@
  */
 public class MessageGroupStoreReaper implements Runnable, DisposableBean, InitializingBean, SmartLifecycle {
 
-	private static Log logger = LogFactory.getLog(MessageGroupStoreReaper.class);
+	private static final Log logger = LogFactory.getLog(MessageGroupStoreReaper.class);
 
 	private final ReentrantLock lifecycleLock = new ReentrantLock();
 
+	@SuppressWarnings("NullAway.Init")
```
cppwfs, line 44: This one is an interesting one.   i.e. the `afterPropertiesSet` does the assert if the user has not set it.  Vs. the usual pattern of it setting the value (in this case it can't).   
I thought about requesting a comment here.   But let's chat .

cppwfs (approved):

spring-integration-core/src/main/java/org/springframework/integration/store/MessageGroupStoreReaper.java
artembilan: Right. Since we have a check in the `afterPropertiesSet()`, this property won't be null at runtime, after initialization.
If we fail in the `afterPropertiesSet()`, then no whole object in the application.

We can revise the logic of this class to rely only on a ctor initialization.
I'm really not a fun of setters for required properties.
But that's different story and it is like that in many places in SF...

cppwfs: Agreed.   This is a low priority, and can be handled in a different PR.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractKeyValueMessageStore.java
artembilan: OK. I did this, but keep in mind that Java is smart enough to turn a simple string concatenation into a `StringBuilder` at compile time.

PR #10344, opened 2025-08-21
GH-10137: Cache `session.list` result in the `AbstractInboundFileSynchronizer`

### LINKED ISSUE #10137: Shouldn't list remote file server each time max-fetch-size is reached

all versions up to 6.5.x

**Describe the bug**

AbstractInboundFileSynchronizingMessageSource shoud list the remote server only once per poll.
However what happens here is each time you download max-fetch-size files, and max-message-per-poll isn't reached, then instead of working on the initial list until, it will list the remote server again, which is unnecessary and costs a lot of time.

Currently if you need a jdbc transaction per file downloaded and keep it to the end of the message flow, you need to use max-fetch-size=1, but then you end up having the remote server listed before downloading each single file, which slow down things considerably.

Chatgpt told me that AbstractRemoteFileStreamingMessageSource could be a solution to this, because it was maintaining a queue that was filled from remote file server list only when queue is empty.
But here I see the queue is cleared when maxFetchSize is reach, therefore it will trigger a remote file server list right after that

```
		if (maxFetchSize > 0 && this.fetched.get() >= maxFetchSize) {
			this.toBeReceived.clear();
			this.fetched.set(0);
		}
```
(https://github.com/spring-projects/spring-integration/blob/23154231f0c7a9fb5e3ebbcd4c8e7207d6e3df76/spring-integration-file/src/main/java/org/springframework/integration/file/remote/AbstractRemoteFileStreamingMessageSource.java#L219)

Therefore I see no solution to have a transaction per downloaded file and it's emitted message without having bad performances.

**To Reproduce**

Use a max-fetch-size=1

**Expected behavior**

Work on the remote file list until exhausted, before polling the server for a file list again.

**Sample**

Uncessary as code analysis shows the issue.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10137

When `maxFetchSize > 0`, the `AbstractInboundFileSynchronizer` performs unnecessary `Session.list()` every time.
Then we go over all those files to filter.
And only after filtering we apply the `maxFetchSize`.

* Add a cache for remote directory entries in the `AbstractInboundFileSynchronizer`
* Guard cache interaction with a `DefaultLockRegistry` per directory
* Rework `AbstractInboundFileSynchronizer.transferFilesFromRemoteToLocal()` to filter all the entries of the `Session.list()` on a first call. If `maxFetchSize > 0`, cache the remaining remote files

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
@@ -24,10 +24,14 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
 import java.util.regex.Matcher;
-import java.util.stream.Collectors;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -53,8 +57,10 @@ import org.springframework.integration.file.remote.session.SessionFactory;
 import org.springframework.integration.file.support.FileUtils;
 import org.springframework.integration.metadata.MetadataStore;
 import org.springframework.integration.metadata.SimpleMetadataStore;
+import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.messaging.MessagingException;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
@@ -85,6 +91,19 @@ public abstract class AbstractInboundFileSynchronizer<F>
 	protected final Log logger = LogFactory.getLog(this.getClass()); // NOSONAR
 
 	private final RemoteFileTemplate<F> remoteFileTemplate;
+
+	private final DefaultLockRegistry lockRegistry = new DefaultLockRegistry();
+
+	@SuppressWarnings("serial")
+	private final Map<String, List<F>> fetchCache =
+			Collections.synchronizedMap(new LinkedHashMap<>(100, 0.75f, true) {
+
+				@Override
+				protected boolean removeEldestEntry(Map.Entry<String, List<F>> eldest) {
+					return size() > 100;
+				}
+
+			});
 
 	@SuppressWarnings("NullAway.Init")
 	private EvaluationContext evaluationContext;
@@ -331,9 +350,6 @@ public abstract class AbstractInboundFileSynchronizer<F>
 			return;
 		}
 		String remoteDirectory = this.remoteDirectoryExpression.getValue(this.evaluationContext, String.class);
-		if (this.logger.isTraceEnabled()) {
-			this.logger.trace("Synchronizing " + remoteDirectory + " to " + localDirectory);
-		}
 		try {
 			Integer transferred = this.remoteFileTemplate.execute(session ->
 					transferFilesFromRemoteToLocal(remoteDirectory, localDirectory, maxFetchSize, session));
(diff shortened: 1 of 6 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
```diff
@@ -86,6 +92,19 @@ public abstract class AbstractInboundFileSynchronizer<F>
...
 
+	private final DefaultLockRegistry lockRegistry = new DefaultLockRegistry();
+
+	@SuppressWarnings("serial")
+	private final Map<String, List<F>> fetchCache =
+			Collections.synchronizedMap(new LinkedHashMap<>(100, 0.75f, true) {
+
+				@Override
+				protected boolean removeEldestEntry(Map.Entry<String, List<F>> eldest) {
+					return size() > 100;
```
cppwfs, line 103: Do we want the 100 to be configurable?

spring-integration-file/src/test/java/org/springframework/integration/file/remote/synchronizer/AbstractRemoteFileSynchronizerTests.java
```diff
@@ -114,12 +118,31 @@ public void testMaxFetchSizeSynchronizer() throws Exception {
...
-		sync.synchronizeToLocalDirectory(mock(File.class), 1);
+		File localDirectory = FileUtils.getTempDirectory();
+		sync.synchronizeToLocalDirectory(localDirectory, 1);
 		assertThat(count.get()).isEqualTo(1);
-		sync.synchronizeToLocalDirectory(mock(File.class), 1);
+
+		@SuppressWarnings("unchecked")
+		Map<String, List<String>> fetchCache = TestUtils.getPropertyValue(sync, "fetchCache", Map.class);
+		List<String> cachedFiles = fetchCache.get("testRemoteDirectory");
+		assertThat(cachedFiles).containsExactly("bar", "baz");
```
cppwfs: Since we are replacing  `foo`.  We may go ahead and replace `bar` and `baz`.

src/reference/antora/modules/ROOT/pages/sftp/max-fetch.adoc
```diff
@@ -26,6 +26,19 @@ Starting with version 5.1, the synchronizer can be provided with a `Comparator<?
 This is useful when restricting the number of files fetched with `maxFetchSize`.
 
 Starting with version 6.4, the `AbstractRemoteFileStreamingMessageSource` has now a convenient `clearFetchedCache()` API to remove references from cache for not processed remote files.
-The references stay in cache because polling configuration does not allow to process all of them in one cycle, and the target `SessionFactory` might be changed between polling cycles, e.g. via `RotatingServerAdvice`.
+The references stay in cache because polling configuration does not allow processing all of them in one cycle, and the target `SessionFactory` might be changed between polling cycles, e.g., via `RotatingServerAdvice`.
+
+Starting with version 7.0, the `AbstractInboundFileSynchronizer` caches a filtered `Session.list(remoteDirectory)` after applying a `maxFetchSize` slicing.
+The logic of the `AbstractInboundFileSynchronizer.transferFilesFromRemoteToLocal()` method is the following:
+
+- If `maxFetchSize > 0`, the lock is acquired against `remoteDirecy` to avoid race condition from different threads, when work is done around cache.
```
cppwfs: `remoteDirectory` is misspelled.

src/reference/antora/modules/ROOT/pages/sftp/max-fetch.adoc
```diff
@@ -26,6 +26,19 @@ Starting with version 5.1, the synchronizer can be provided with a `Comparator<?
...
+The performance degradation is minimal since all the later synchronizations deal only with in-memory cached leftover;
+- If no cache entry for the `remoteDirecy`, the `Session.list(remoteDirectory)` is called and all returned remote files are filtered;
```
cppwfs: same here

src/reference/antora/modules/ROOT/pages/sftp/max-fetch.adoc
```diff
@@ -26,6 +26,19 @@ Starting with version 5.1, the synchronizer can be provided with a `Comparator<?
...
+- the filtered result then sliced by the `maxFetchSize`;
+- then these file entries are being transferred to the local directory;
+- the rest of filtered remote files are cached for later synchronizations;
+- if there is a cache entry for the `remoteDirecy`, such a list is sliced `maxFetchSize` and iterated for the transfer to the local directory;
```
cppwfs: same here.

"such a list is sliced `maxFetchSize`" could be rewritten as " such a list is sliced to the `maxFetchSize`".

src/reference/antora/modules/ROOT/pages/sftp/max-fetch.adoc
```diff
@@ -26,6 +26,19 @@ Starting with version 5.1, the synchronizer can be provided with a `Comparator<?
...
+- if one of the transfers fails, the `filter` is reset from this remote file to the end of the filtered list.
```
cppwfs: "if one of the transfers fails, the `filter` is reset from this remote file "
can be rewritten as 
"if one of the transfers fails, the `filter` is reset from from the failed remote file".

cppwfs (changes requested):
Great work.

Just the usual nitpicks.

spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizer.java
artembilan: I don't think so.
The less choice for user, the better for this class instance state.
Why would one try to poll in parallel more than 100 remote directories?

Issue #10345, opened 2025-08-22
Replace the use of Spring Retry with Core Retry


artembilan:
Similar effort to the one in Spring for Apache Kafka, for example: https://github.com/spring-projects/spring-kafka/issues/4051

anthologia:
Can I give the kafka module a try?

artembilan:
We cannot do that, @anthologia , until upstream issue in Spring Kafka is fixed.

artembilan:
FYI, Spring for Apache Kafka has been migrated to Core Retry.
So, we can start working on respective retry part over here in the `spring-integration-kafka` module.

NOTE: I'll observe a CI build for SNAPSHOT and might jump myself if that breaks seriously.

artembilan:
OK. So, latest SNAPSHOT build is OK.
Probably there is nothing to worry about and you are free to go ahead and fix it respectively.
Take an inspiration from AMQP or JMS module for similar functionality.

artembilan:
Looks like Spring Cloud Stream is failing now since it uses `spring-integration-kafka`, but this one has not been migrated to Core Retry properly yet.
So, I'm treating this as urgent and going to take care myself ASAP.

anthologia:
No problem! Thanks for letting me know. 😄

PR #10346, opened 2025-08-22
GH-10345: Replace Spring Retry in the JDBC module

artembilan:
Related to: https://github.com/spring-projects/spring-integration/issues/10345

Migrate the only class in the ` jdbc ` module - `PostgresSubscribableChannel`, - to the Core Retry

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
@@ -16,12 +16,17 @@
 
 package org.springframework.integration.jdbc.channel;
 
+import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.Executor;
 
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.core.log.LogAccessor;
+import org.springframework.core.retry.RetryException;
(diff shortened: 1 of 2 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -69,7 +73,8 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 
 	private @Nullable TransactionTemplate transactionTemplate;
 
-	private RetryTemplate retryTemplate = RetryTemplate.builder().maxAttempts(1).build();
+	private RetryTemplate retryTemplate =
+			new RetryTemplate(RetryPolicy.builder().maxAttempts(1).delay(Duration.ZERO).build());
```
cppwfs, line 77: Nitpick should the 1 be a constant?

cppwfs (changes requested):
LGTM

Just my usual nitpick.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
artembilan: I don't think so.
That's just "one" and it is not that important for the whole functionality of this class instance.

PR #10359, opened 2025-08-24
GH-10083: Apply NullAbility to `core` `endpoint` package

mjd507:
related to: #10083

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -22,7 +22,6 @@ import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscription;
 import reactor.core.publisher.Flux;
 
-import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.core.AttributeAccessor;
 import org.springframework.integration.IntegrationPattern;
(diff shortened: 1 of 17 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -68,17 +67,18 @@ public abstract class MessageProducerSupport extends AbstractEndpoint
 	@Nullable
 	private MessageReceiverObservationConvention observationConvention;
 
+	@SuppressWarnings("NullAway.Init")
```
artembilan: I believe this one is `@Nullable` as well.
See `getOutputChannel()` contact.
And then `getRequiredOutputChannel()`

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
mjd507: updated. 
I did this due to I found below lifecycle method in this class.  
```java
	@Override
	public void afterSingletonsInstantiated() {
		Assert.state(this.outputChannel != null || StringUtils.hasText(this.outputChannelName),
				"'outputChannel' or 'outputChannelName' is required");
	}
```
will it not be called in some cases?

artembilan: It will be, but since that checks against one or another , both of them has to be treated equal. I guess `getOutputChannel` should do the trick.

PR #10360, opened 2025-08-26
GH-10083: Apply Nullability to some core support packages

ngocnhan-tran1996:
Related to: #10083

This commit applies Nullability to `support.channel`, `support.context` and `support.converter` packages

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/DefaultHeaderChannelRegistry.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/DefaultHeaderChannelRegistry.java
@@ -261,7 +261,7 @@ public class DefaultHeaderChannelRegistry extends IntegrationObjectSupport
 
 	@Override
 	public String getComponentType() {
-		return "channel";
+		return "header-channel-registry";
 	}
 
 	protected record MessageChannelWrapper(MessageChannel channel, long expireAt) {
--- a/spring-integration-core/src/main/java/org/springframework/integration/channel/DefaultHeaderChannelRegistry.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/channel/DefaultHeaderChannelRegistry.java
@@ -259,6 +259,11 @@ public class DefaultHeaderChannelRegistry extends IntegrationObjectSupport
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "channel";
+	}
+
 	protected record MessageChannelWrapper(MessageChannel channel, long expireAt) {
 
 	}
--- a/spring-integration-core/src/main/java/org/springframework/integration/codec/CodecMessageConverter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/codec/CodecMessageConverter.java
@@ -77,4 +77,9 @@ public class CodecMessageConverter extends IntegrationObjectSupport implements M
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "converter";
+	}
+
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
@@ -142,9 +142,8 @@ public abstract class IntegrationObjectSupport implements ComponentSourceAware, 
 	 * Subclasses may implement this method to provide component type information.
 	 */
 	@Override
-	@Nullable
 	public String getComponentType() {
-		return null;
+		return "integration-object-support";
 	}
 
 	@Override
--- a/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
@@ -138,14 +138,6 @@ public abstract class IntegrationObjectSupport implements ComponentSourceAware, 
 		this.componentName = componentName;
 	}
 
-	/**
-	 * Subclasses may implement this method to provide component type information.
-	 */
-	@Override
-	public String getComponentType() {
-		return "integration-object-support";
-	}
-
 	@Override
 	public void setComponentSource(Object source) {
 		this.beanSource = source;
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
@@ -201,6 +201,11 @@ public abstract class AbstractEndpoint extends IntegrationObjectSupport
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "endpoint";
+	}
+
 	/**
 	 * Stop the component and invoke callback.
 	 * @param callback the Runnable to invoke.
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
@@ -201,11 +201,6 @@ public abstract class AbstractEndpoint extends IntegrationObjectSupport
 		}
 	}
 
-	@Override
-	public String getComponentType() {
-		return "endpoint";
-	}
-
 	/**
 	 * Stop the component and invoke callback.
 	 * @param callback the Runnable to invoke.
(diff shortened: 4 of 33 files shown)


spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
```diff
@@ -290,7 +290,7 @@ private void setupRecoveryCallbackIfAny() {
 						context -> {
 							Message messageToRecover =
 									(Message) context.getAttribute(AmqpMessageHeaderErrorMessageStrategy.AMQP_RAW_MESSAGE);
-							if (messageToRecover != null) {
+							if (messageToRecover != null && context.getLastThrowable() != null) {
```
artembilan: Please, hold off these changes.
I'm aware of the failure over there after Spring AMQP migration to Spring Core Retry.
Stay tuned: will PR soon enough.
Thanks

artembilan (changes requested):

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpInboundChannelAdapter.java
ngocnhan-tran1996: I think my changes may have caused the Spring AMQP failure. I’ve reverted the commit related to `amqp` module.

spring-integration-core/src/main/java/org/springframework/integration/support/channel/BeanFactoryChannelResolver.java
```diff
@@ -52,9 +53,9 @@ public class BeanFactoryChannelResolver implements DestinationResolver<MessageCh
 
 	private final Lock lock = new ReentrantLock();
 
-	private BeanFactory beanFactory;
+	private @Nullable BeanFactory beanFactory;
```
artembilan: `NullAway.Init` and remove that `Assert.state(this.beanFactory != null, "BeanFactory is required");` from critical path execution.

spring-integration-core/src/main/java/org/springframework/integration/support/context/NamedComponent.java
```diff
@@ -26,6 +28,7 @@ public interface NamedComponent {
 
 	String getComponentName();
 
+	@Nullable
```
artembilan: It must not be.
Please, consider to fix the guilty one instead.

artembilan (changes requested):

artembilan:
Th fix for AMQP is in.
Please, consider to rebase your branch.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/context/IntegrationObjectSupport.java
```diff
@@ -142,9 +142,8 @@ public void setComponentName(String componentName) {
 	 * Subclasses may implement this method to provide component type information.
 	 */
 	@Override
-	@Nullable
 	public String getComponentType() {
-		return null;
+		return "integration-object-support";
```
artembilan: Not very good.
I believe we are in a situation where we should remove this method implementation here in the `IntegrationObjectSupport`.
And ensure that it is implemented properly everywhere else.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/channel/DefaultHeaderChannelRegistry.java
```diff
@@ -259,6 +259,11 @@ public void run() {
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "channel";
```
artembilan: `header-channel-registry`.

I can argue that this class is not supposed to be an `IntegrationObjectSupport`.
But that's different story.

spring-integration-core/src/main/java/org/springframework/integration/codec/CodecMessageConverter.java
```diff
@@ -77,4 +77,9 @@ public Message<?> toMessage(Object payload, @Nullable MessageHeaders headers) {
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "converter";
```
artembilan, line 82: Same here: this is not an `IntegrationObjectSupport`.
But separate story.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractEndpoint.java
```diff
@@ -201,6 +201,11 @@ public final void stop(Runnable callback) {
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "endpoint";
```
artembilan: I think this is not correct: every each `AbstractEndpoint` must provide their proper implementation.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/connection/AbstractConnectionFactory.java
```diff
@@ -165,6 +165,11 @@ public ApplicationEventPublisher getApplicationEventPublisher() {
 		return this.applicationEventPublisher;
 	}
 
+	@Override
+	public String getComponentType() {
+		return "connection";
```
artembilan: `connection-factory` ?

spring-integration-core/src/main/java/org/springframework/integration/transaction/ExpressionEvaluatingTransactionSynchronizationProcessor.java
```diff
@@ -131,6 +131,11 @@ public void processAfterRollback(IntegrationResourceHolder holder) {
 		doProcess(holder, this.afterRollbackExpression, this.afterRollbackChannel, "afterRollback");
 	}
 
+	@Override
+	public String getComponentType() {
+		return "processor";
```
artembilan, line 136: This one also must  not be an `IntegrationObjectSupport`.

artembilan (changes requested):
Thank you for an update!

You made me think a bit on the state of `IntegrationObjectSupport`.
Some of my comments are not items for action, but some of them are really ask for you to change.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/AbstractPollingEndpoint.java
```diff
@@ -183,6 +183,11 @@ public void setTransactionSynchronizationFactory(
 		this.transactionSynchronizationFactory = transactionSynchronizationFactory;
 	}
 
+	@Override
+	public String getComponentType() {
+		return "polling-endpoint";
```
artembilan, line 188: The `SourcePollingChannelAdapter` does provide an impl of this method.
So, another one - `PollingConsumer` - has to do that as well: `polling-consumer`.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/EventDrivenConsumer.java
```diff
@@ -123,4 +123,9 @@ private void logComponentSubscriptionEvent(boolean add) {
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "event-driven";
```
artembilan, line 128: Well, `event-driven-consumer` to be precise.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -111,6 +111,11 @@ public void setOutputChannelName(String outputChannelName) {
 		return this.outputChannel;
 	}
 
+	@Override
+	public String getComponentType() {
+		return "message-producer";
```
artembilan, line 116: I think neither of abstract class should provide an impl.
That is really responsibility of the specific one.
Without an implementation here, we may spot a bug somewhere down the hierarchy, e.g. `ReactiveMessageSourceProducer`.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/ReactiveStreamsConsumer.java
```diff
@@ -208,6 +208,11 @@ protected void doStop() {
 		}
 	}
 
+	@Override
+	public String getComponentType() {
+		return "reactive-streams";
```
artembilan, line 213: `reactive-streams-consumer` ?

artembilan (changes requested):

artembilan (approved):
OK. I'm taking your PR locally for the final review and clean up before merge.
Thank you!

artembilan:
Merged as: https://github.com/spring-projects/spring-integration/commit/17e6a3165eb0b5f4973a95658a2c8ed0f41f8dd8

PR #10363, opened 2025-08-29
GH-10362: Block on PostgresSubscribableChannel.unsubscribe

### LINKED ISSUE #10362: Unsubscribe of message handler while processing PostgresSubscribableChannel.notifyUpdate()
**Describe the bug**

We encountered a `MessageDeliveryException` during application shutdown.

The stack trace suggests that `PostgresSubscribableChannel` was still processing messages in its `pollAndDispatchMessage` loop while Spring had already unregistered the corresponding message handlers. This led to a `org.springframework.integration.MessageDispatchingException: Dispatcher has no subscribers` error.

From what I can tell, Spring shuts components down in the expected order, but any in-flight messages may no longer be dispatchable at that point.

Would it make sense for `PostgresSubscribableChannel` to block on `unsubscribe` calls while a message is being dispatched, and then exit the polling loop once the message handling has completed? Since this situation likely only occurs asynchronously during shutdown, relying on error handling seems fragile, as at that point Spring may have already stopped other required beans.

A test reproducing the situation can be found at https://github.com/bertschneider/spring-integration-notifyUpdate/blob/main/src/test/java/com/example/springintegration/SpringIntegrationITest.java

Stack trace:

```
org.springframework.integration.MessageDispatchingException: Dispatcher has no subscribers
	at org.springframework.integration.dispatcher.UnicastingDispatcher.doDispatch(UnicastingDispatcher.java:154) ~[spring-integration-core-6.5.1.jar:6.5.1]
	at org.springframework.integration.dispatcher.UnicastingDispatcher.dispatch(UnicastingDispatcher.java:121) ~[spring-integration-core-6.5.1.jar:6.5.1]
	at org.springframework.integration.jdbc.channel.PostgresSubscribableChannel.dispatch(PostgresSubscribableChannel.java:231) ~[spring-integration-jdbc-6.5.1.jar:6.5.1]
	at org.springframework.integration.jdbc.channel.PostgresSubscribableChannel.lambda$doPollAndDispatchMessage$4(PostgresSubscribableChannel.java:220) ~[spring-integration-jdbc-6.5.1.jar:6.5.1]
	at org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:357) ~[spring-retry-2.0.12.jar:na]
	at org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:230) ~[spring-retry-2.0.12.jar:na]
	at org.springframework.integration.jdbc.channel.PostgresSubscribableChannel.lambda$doPollAndDispatchMessage$5(PostgresSubscribableChannel.java:220) ~[spring-integration-jdbc-6.5.1.jar:6.5.1]
	at java.base/java.util.Optional.map(Optional.java:260) ~[na:na]
	at org.springframework.integration.jdbc.channel.PostgresSubscribableChannel.doPollAndDispatchMessage(PostgresSubscribableChannel.java:220) ~[spring-integration-jdbc-6.5.1.jar:6.5.1]
	at org.springframework.integration.jdbc.channel.PostgresSubscribableChannel.pollAndDispatchMessage(PostgresSubscribableChannel.java:190) ~[spring-integration-jdbc-6.5.1.jar:6.5.1]
	at org.springframework.integration.jdbc.channel.PostgresSubscribableChannel.lambda$notifyUpdate$0(PostgresSubscribableChannel.java:183) ~[spring-integration-jdbc-6.5.1.jar:6.5.1]
	at java.base/java.lang.Thread.run(Thread.java:1583) ~[na:na]
```

**Expected behavior**

Processing of in-flight messages is finished.

6.5.1

bertschneider:
Unsubscribing message handlers while processing an in-flight message may lead to a MessageDeliveryException.

Fixes: GH-10362

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
@@ -19,6 +19,7 @@ package org.springframework.integration.jdbc.channel;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.Executor;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.jspecify.annotations.Nullable;
 
@@ -53,6 +54,7 @@ import org.springframework.util.ReflectionUtils;
  * @author Rafael Winterhalter
  * @author Artem Bilan
  * @author Igor Lovich
+ * @author Norbert Schneider
  *
  * @since 6.0
  */
@@ -70,6 +72,8 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 	private final PostgresChannelMessageTableSubscriber messageTableSubscriber;
 
 	private final UnicastingDispatcher dispatcher = new UnicastingDispatcher();
+
+	private final ReentrantReadWriteLock hasHandlersLock = new ReentrantReadWriteLock();
 
 	private @Nullable TransactionTemplate transactionTemplate;
 
@@ -164,12 +168,18 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 
 	@Override
 	public boolean unsubscribe(MessageHandler handle) {
-		boolean unsubscribed = super.unsubscribe(handle);
-		if (this.dispatcher.getHandlerCount() == 0) {
-			this.messageTableSubscriber.unsubscribe(this);
-			this.hasHandlers = false;
+		this.hasHandlersLock.writeLock().lock();
+		try {
+			boolean unsubscribed = super.unsubscribe(handle);
+			if (this.dispatcher.getHandlerCount() == 0) {
+				this.messageTableSubscriber.unsubscribe(this);
+				this.hasHandlers = false;
+			}
+			return unsubscribed;
 		}
-		return unsubscribed;
+		finally {
+			this.hasHandlersLock.writeLock().unlock();
+		}
 	}
 
 	@Override
@@ -209,25 +219,23 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 	}
 
 	private Optional<?> doPollAndDispatchMessage() {
-		if (this.hasHandlers) {
-			TransactionTemplate transactionTemplateToUse = this.transactionTemplate;
-			if (transactionTemplateToUse != null) {
-				return executeWithRetry(() ->
-						transactionTemplateToUse.execute(status ->
-								pollMessage()
-										.filter(message -> {
-											if (!this.hasHandlers) {
-												status.setRollbackOnly();
-												return false;
-											}
-											return true;
-										})
-										.map(this::dispatch)));
+		this.hasHandlersLock.readLock().lock();
+		try {
+			if (this.hasHandlers) {
+				TransactionTemplate transactionTemplateToUse = this.transactionTemplate;
+				if (transactionTemplateToUse != null) {
+					return executeWithRetry(() ->
+							transactionTemplateToUse.execute(status ->
+									pollMessage().map(this::dispatch)));
+				}
+				else {
+					return pollMessage()
+							.map(message -> executeWithRetry(() -> dispatch(message)));
+				}
 			}
-			else {
-				return pollMessage()
-						.map(message -> executeWithRetry(() -> dispatch(message)));
-			}
+		}
+		finally {
+			this.hasHandlersLock.readLock().unlock();
 		}
 		return Optional.empty();
 	}
--- a/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
+++ b/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
@@ -27,6 +27,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
+import java.util.stream.IntStream;
 
 import javax.sql.DataSource;
 
@@ -72,6 +73,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Igor Lovich
  * @author Adama Sorho
  * @author Johannes Edmeier
+ * @author Norbert Schneider
  *
  * @since 6.0
  */
@@ -308,6 +310,49 @@ public class PostgresChannelMessageTableSubscriberTests implements PostgresConta
 		messageStore.addMessageToGroup(groupId, new GenericMessage<>("2"));
 		assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
 		assertThat(payloads).containsExactlyInAnyOrder("1", "2");
+	}
+
+	@Test
+	public void testUnsubscribeHandlerDuringDispatch() throws InterruptedException {
+		int numberOfMessages = 200;
+		CountDownLatch messagesUntilUnsubscribe = new CountDownLatch(10);
+		AtomicInteger receivedMessages = new AtomicInteger();
+		AtomicReference<Throwable> receivedException = new AtomicReference<>();
+
+		postgresChannelMessageTableSubscriber.start();
+		postgresSubscribableChannel.setErrorHandler(receivedException::set);
+
+		MessageHandler messageHandlerOne = message -> {
+			receivedMessages.getAndIncrement();
+			messagesUntilUnsubscribe.countDown();
+		};
+		postgresSubscribableChannel.subscribe(messageHandlerOne);
+
+		MessageHandler messageHandlerTwo = message -> {
+			receivedMessages.getAndIncrement();
+			messagesUntilUnsubscribe.countDown();
+		};
+		postgresSubscribableChannel.subscribe(messageHandlerTwo);
+
+		IntStream.range(0, numberOfMessages).forEach(i -> {
+			taskExecutor.execute(() ->
+				messageStore.addMessageToGroup(groupId, new GenericMessage<>(i)));
+		});
+
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+
+		assertThat(messagesUntilUnsubscribe.await(10, TimeUnit.SECONDS)).isTrue();
+		postgresSubscribableChannel.unsubscribe(messageHandlerOne);
+		postgresSubscribableChannel.unsubscribe(messageHandlerTwo);
+
+		int undeliveredMessages = messageStore.messageGroupSize(groupId);
+		int processedMessages = undeliveredMessages + receivedMessages.get();
+		assertThat(processedMessages).isEqualTo(numberOfMessages);
+		assertThat(receivedException).hasNullValue();
 	}
 
 	@Configuration


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -188,8 +202,13 @@ public void notifyUpdate() {
 		this.executor.execute(() -> {
 			Optional<?> dispatchedMessage;
 			do {
-				dispatchedMessage = pollAndDispatchMessage();
-			} while (dispatchedMessage.isPresent());
+				this.lock.lock();
```
artembilan: This makes the whole component, essentially, single-threaded.
I don't think that is a goal here at all.
Let's see if we can come up with a solution based on on the `Semaphore`!

artembilan (changes requested):
Please, run `./gradlew :spring-integration-jdbc:check` locally before pushing.

It also would be great if you bring some test to cover this logic.

Thanks

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
artembilan: Yeah... Looks like `Semaphore` is not going to work as well.
As long as we got a parallel poll call, that might suffer from the problem when it enters the dispatch part, but we have been just unsubscribed by the first thread releasing semaphore for us.

How about to look into a `ReentrantReadWriteLock`?
So, all those `pollAndDispatchMessage()` would hold a `ReadLock` and will be blocked until `WriteLock` is released from the `unsubscribe()` call.

There is also a `Phaser` abstraction in Java, but I don't think our dynamic `pollAndDispatchMessage()` would guarantee a fair logic around `hasHandlers` state.

bertschneider: I'm only somewhat familiar with the codebase, and I originally assumed that `notifyUpdate` would be called just once, or at most in sequence.

You're right that `ReentrantReadWriteLock` is a good fit here. We only have a single writer (during unsubscribe calls), but potentially many readers through `notifyUpdate`.

We probably can't narrow the lock scope, since the entire flow (from polling a new message from the store to dispatching it via the dispatcher) needs to be protected against concurrent unsubscribes.

I've updated the code to use `ReentrantReadWriteLock`. I had missed `PostgresMetadataStoreTests` the first time around, but added a test there now.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -71,6 +73,8 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 
 	private final UnicastingDispatcher dispatcher = new UnicastingDispatcher();
 
+	private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
```
artembilan: I don't think that we need a `fair` behavior.
According to Javadocs it is a bit slower mode.
And we still would like to have our "readers" as fast as possible.
So, `non-fair` mode is enough for us.
Whoever wins the race - doesn't matter for our logic to finish the work properly.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -153,23 +157,35 @@ protected void onInit() {
 
 	@Override
 	public boolean subscribe(MessageHandler handler) {
-		boolean subscribed = super.subscribe(handler);
-		if (this.dispatcher.getHandlerCount() == 1) {
-			this.messageTableSubscriber.subscribe(this);
-			this.hasHandlers = true;
-			notifyUpdate();
+		this.lock.writeLock().lock();
```
artembilan: I don't think we need a lock in the `subscribe()`.
As long as `hasHandlers` is not set to true we are good with the logic.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -153,23 +157,35 @@ protected void onInit() {
...
 	public boolean unsubscribe(MessageHandler handle) {
-		boolean unsubscribed = super.unsubscribe(handle);
-		if (this.dispatcher.getHandlerCount() == 0) {
-			this.messageTableSubscriber.unsubscribe(this);
-			this.hasHandlers = false;
+		this.lock.writeLock().lock();
+		try {
+			boolean unsubscribed = super.unsubscribe(handle);
+			if (this.dispatcher.getHandlerCount() == 0) {
+				this.messageTableSubscriber.unsubscribe(this);
```
artembilan, line 175: We need to hold lock only around these two lines.
The problem with readers arises only when we got `!hasHandlers`.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -188,8 +204,14 @@ public void notifyUpdate() {
...
-				dispatchedMessage = pollAndDispatchMessage();
-			} while (dispatchedMessage.isPresent());
+				this.lock.readLock().lock();
+				try {
+					dispatchedMessage = pollAndDispatchMessage();
+				}
+				finally {
+					this.lock.readLock().unlock();
+				}
+			} while (this.hasHandlers && dispatchedMessage.isPresent());
```
artembilan: I still don't believe that we need `this.hasHandlers` condition here.
That is exactly what we have as a first condition in the `doPollAndDispatchMessage()`.
I also think that lock could be narrowed down to the `doPollAndDispatchMessage()` around `else` body.
Pay attention that transactional part is covered by rollback.
We got the problem only in that non-transactional part.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -310,6 +311,36 @@ public void testRenewConnection() throws Exception {
...
+		});
+
+		Thread.sleep(50);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		Thread.sleep(50);
+		postgresSubscribableChannel.unsubscribe(messageHandler);
+
+		assertThat(receivedMessageCount.await(2, TimeUnit.SECONDS)).isFalse();
```
artembilan: Having `300` messages there is indeed a chance that it is not going to fulfill in two seconds in any conditions.
So, I don't think the approach in the test is robust.
Plus we need take into account this 2 seconds blocking the whole test suite execution for nothing.
So, we need to revise the logic of the test.

We may aim for something like ensure in the end of test that no lost messages and no exceptions:
1. Have `100` messages to be stored as a reasonable number 
2. The error handler you have so far is OK
3. Have a `CountDownLatch` for the `MessageHandler` (subscriber) as reasonable `50`
4. Also, store received messages to some list to verify in the end (or just `AtomicInteger`)
5. No sleeps in between.
6. Execute 5 parallel `notifyUpdate()`
7. Check for the mentioned latch (`true` this time) (not sure if 10 seconds would be enough).
8. Unsubscribe
9. Calculate the group size in the store + the number of processed messages.
10. The result has to be `100`.

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
bertschneider: The docs state that `a nonfair lock that is continuously contended may indefinitely postpone one or more reader or writer threads`. IMHO, it's unclear which kind of lock could be postponed, and the `unsubscribe` should be processed as fast as possible, hence the `fair` behavior. Regardless, changed.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
bertschneider: True, I don't expect any performance implications and opted for consistency/similarity. Removed.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
bertschneider: True, `this.hasHandlers` is not required in the condition, removed.

The lock could be pulled down into `doPollAndDispatchMessage` to execute the error handling in `pollAndDispatchMessage` outside the locked scope. But I don't think that only the `else` branch should be guarded. 
IMHO, `this.hasHandlers` should only be read in a locked scope, as it could be changed between its check and a subsequent `pollMessage,` reading a message without a `dispatcher` being available. This could happen when only the `else` branch is guarded, couldn't it?
The transaction rollback in the `if` branch would prevent a message loss, but why even let it come to that, as a rollback seems more costly compared to the lock? Using the lock around both branches would also remove the need for the `filter` step.
I pushed a version to see what that would look like, but happy to change it like you suggested.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: Indeed, the fixed assertion time is quite ugly, and the proposed approach is way more elegant. Changed.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
bertschneider: IMHO, `super.unsubscribe()`, as it removed the dispatcher, and `this.hasHandlers = false` need to be locked together. As this only happens once, I don't suspect any performance implications in guarding the whole method body, which seems easier to maintain. 
Or I'm getting something completely wrong here 😬

artembilan: We can have several subscribers, but only the last unsubscribed changes the state.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
artembilan: I agree about guarding `hasHandlers` check by the lock . So, your idea to remove the filter is great.
I think when I designed this class I wanted to stay away from locks as much as possible . Therefore rollback and recommendation to use transaction to not lose data.
Since we cannot reach stability with the current logic, I’m ok to make changes with lock algorithm.

artembilan:
Thank you for an update!
Commenting from the phone.
Will review properly on Tuesday after holidays .

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
```diff
@@ -71,6 +73,8 @@ public class PostgresSubscribableChannel extends AbstractSubscribableChannel
 
 	private final UnicastingDispatcher dispatcher = new UnicastingDispatcher();
 
+	private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
```
artembilan: Let's make its name more specific to its usage!
How about `hasHandlersLock`?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
artembilan: Doesn't look like you have this addressed.
I still believe that we don't need to hold a lock for all the subscribers.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -72,7 +73,7 @@
  * @author Igor Lovich
  * @author Adama Sorho
  * @author Johannes Edmeier
- *
+ * @author Norbert Schneider
  * @since 6.0
```
artembilan, line 78: I prefer a blank line after the `@author` section to emphasize on them

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -310,6 +311,41 @@ public void testRenewConnection() throws Exception {
 		assertThat(payloads).containsExactlyInAnyOrder("1", "2");
 	}
 
+	@Test
+	public void testUnsubscribeHandlerDuringDispatch() throws InterruptedException {
+		int numberOfMessages = 150;
+		CountDownLatch messagesUntilUnsubscribe = new CountDownLatch(10);
```
artembilan, line 318: Why wait for just 10 if we produce so many?
Maybe then decrease a production amount?

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -310,6 +311,41 @@ public void testRenewConnection() throws Exception {
...
+		AtomicInteger receivedMessages = new AtomicInteger();
+		AtomicReference<Throwable> receivedException = new AtomicReference<>();
+
+		postgresChannelMessageTableSubscriber.start();
+		postgresSubscribableChannel.setErrorHandler(receivedException::set);
+		MessageHandler messageHandler = message -> {
+			receivedMessages.getAndIncrement();
+			messagesUntilUnsubscribe.countDown();
+		};
+		postgresSubscribableChannel.subscribe(messageHandler);
```
artembilan: Can we consider to stress it more with another subscriber?
It cannot be the same instance, though, since `AbstractDispatcher` logic is based on a `Set`.
But I don't think it is a big deal to copy/paste lambda for another `MessageHandler` instance.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -310,6 +311,41 @@ public void testRenewConnection() throws Exception {
...
+
+		IntStream.range(0, numberOfMessages).forEach(i -> {
+			taskExecutor.execute(() -> {
+				messageStore.addMessageToGroup(groupId, new GenericMessage<>(String.valueOf(i)));
```
artembilan: I don't think an extra conversion to `String` brings any value for this test.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
```diff
@@ -310,6 +311,41 @@ public void testRenewConnection() throws Exception {
...
+			});
+		});
+
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+		taskExecutor.execute(postgresSubscribableChannel::notifyUpdate);
+
+		assertThat(messagesUntilUnsubscribe.await(30, TimeUnit.SECONDS)).isTrue();
```
artembilan: So, do we really need to wait those `30 seconds`  to have this test passing?
Maybe indeed then it is better to decrease an amount of messages we publish?
Imaging if many of the 2500 tests in the project are that long...

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
bertschneider: Renamed.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: Changed.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: Well, the first `10` should ensure that processing started. If I reduce the overall number of messages, the test is not stable on my system, as processing is just too fast. Adding a second subscriber even increases that number to 200, as the `unsubscribe` of both `MessageHandler`s take a bit longer now.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: Added a second `MessageHandler`.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: Ah, my IDE showed that the `GenericMessage` constructor would require a `String`, but it's actually a generic type. So, yeah, the `String` conversion is not required. Removed.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: It will only wait until the `CountDownLatch` reaches `0`. The timeout is just a safeguard to not block indefinitely if there's an unrelated error. The test takes about 400ms on my system.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/channel/PostgresSubscribableChannel.java
bertschneider: As mentioned above, we need to guard `super.unsubscribe` and `this.hasHandlers` in one lock scope, as otherwise the handler would already be unsubscribed before/independently of the `this.hasHandlers` change. Or do I get something wrong?

artembilan: Yes. I see now my mistake.
I had `this.messageTableSubscriber.unsubscribe(this)` confused with the `super.unsubscribe(handle)`  😢 .

Yes, you conclusion is correct.
The `hasHandlers` has really to reflect the state of the super dispatcher.
When that `if (this.dispatcher.getHandlerCount() == 0) {` is true, we must not proceed with message polling.
So, `hasHandlers` is really a shortcut for that `getHandlerCount()`.
Therefore we need to block until message is processed or until we are done with unsubscription modifying that `getHandlerCount()`.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
artembilan: > take a bit longer now

Does your `fair` comment before make any difference?
I mean it really would be fair to give an `unsubscribe` (`WriteLock`) a precedence to shutdown quicker, rather than let all those reading threads to poll more messages.

I think if you decrease the number of reading threads , that would make some difference on the test execution.

However, I'm leaving this comment just for a curiosity.
Looks like we have reached a consensus when our solution works as expected and test does not take too much time.

Let me know if you are going to look into that or we can merge your solution already now!

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
artembilan: Correct. That's an expectation to have test to be performed as fast as possible.
I usually do only `10 seconds` for a sake of slow environment.
If test takes more than 10 seconds, then it's time to revise it.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: I didn't check it before, but a rough measurement (simply executing the test a few times) shows no observable difference between a normal and a `fair` `ReentrantReadWriteLock`. So, I would leave it as is.

The overall number of messages does not really matter and could even be higher, as processing is interrupted by unsubscribing the handlers anyway. The runtime of "a normal test execution" is defined by the time the `unsubscribe`s take. As the test seems stable, I would leave it this way.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/postgres/PostgresChannelMessageTableSubscriberTests.java
bertschneider: True, 30 seconds seems too long. Reduced it to the suggested 10s.

artembilan (approved):
LGTM.
Will merge when the PR build is green.

bertschneider:
Great! Thanks for your time and support with this PR.

artembilan:
Merged and cherry-picked to `6.5.x` & `6.4.x`.

PR #10364, opened 2025-08-29
GH-10083: Apply Nullability to `core` `handler` package

mjd507:
Related to: #10083

`MessageProcessorMessageSource` now provides a fake message to the underlying messageProcessor since the contract requires.

add `@Nullable` on `CheckedCallable.call()`, since in `LockRequestHandlerAdvice.doInvoke()`, there is a lambda `ExecutionCallback::execute` which will be running inside the `call()`, and its Nullable.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/core/ErrorMessagePublisher.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/core/ErrorMessagePublisher.java
@@ -70,7 +70,7 @@ public class ErrorMessagePublisher implements BeanFactoryAware {
 		this.errorMessageStrategy = errorMessageStrategy;
 	}
 
-	public final void setChannel(MessageChannel channel) {
+	public final void setChannel(@Nullable MessageChannel channel) {
 		this.channel = channel;
 	}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/core/RecoveryCallback.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/core/RecoveryCallback.java
@@ -35,6 +35,6 @@ public interface RecoveryCallback<T extends @Nullable Object> {
 	 * @param cause the cause of the failure
 	 * @return an Object that can be used to replace the callback result that failed
 	 */
-	T recover(AttributeAccessor context, Throwable cause);
+	T recover(AttributeAccessor context, @Nullable Throwable cause);
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/core/RecoveryCallback.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/core/RecoveryCallback.java
@@ -35,6 +35,6 @@ public interface RecoveryCallback<T extends @Nullable Object> {
 	 * @param cause the cause of the failure
 	 * @return an Object that can be used to replace the callback result that failed
 	 */
-	T recover(AttributeAccessor context, @Nullable Throwable cause);
+	T recover(AttributeAccessor context, Throwable cause);
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProcessorMessageSource.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProcessorMessageSource.java
@@ -19,6 +19,8 @@ package org.springframework.integration.endpoint;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.integration.handler.MessageProcessor;
+import org.springframework.integration.support.MutableMessageBuilder;
+import org.springframework.messaging.Message;
 
 /**
  * The {@link org.springframework.integration.core.MessageSource} strategy implementation
@@ -45,7 +47,9 @@ public class MessageProcessorMessageSource extends AbstractMessageSource<Object>
 
 	@Override
 	protected @Nullable Object doReceive() {
-		return this.messageProcessor.processMessage(null);
+		Message<Object> fakeMessage = MutableMessageBuilder.withPayload(new Object(), false).build();
+		// provide a fake message here since the contract of processMessage requires a NonNull Message.
+		return this.messageProcessor.processMessage(fakeMessage);
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProcessorMessageSource.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProcessorMessageSource.java
@@ -29,10 +29,14 @@ import org.springframework.messaging.Message;
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Jiandong Ma
  *
  * @since 5.0
  */
 public class MessageProcessorMessageSource extends AbstractMessageSource<Object> {
+
+	// provide a fake message since the contract of processMessage requires a NonNull Message.
+	static final Message<Object> FAKE_MESSAGE = MutableMessageBuilder.withPayload(new Object(), false).build();
 
 	private final MessageProcessor<?> messageProcessor;
 
@@ -47,9 +51,7 @@ public class MessageProcessorMessageSource extends AbstractMessageSource<Object>
 
 	@Override
 	protected @Nullable Object doReceive() {
-		Message<Object> fakeMessage = MutableMessageBuilder.withPayload(new Object(), false).build();
-		// provide a fake message here since the contract of processMessage requires a NonNull Message.
-		return this.messageProcessor.processMessage(fakeMessage);
+		return this.messageProcessor.processMessage(FAKE_MESSAGE);
 	}
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
@@ -76,7 +76,7 @@ import org.springframework.util.StringUtils;
 public abstract class AbstractMessageProducingHandler extends AbstractMessageHandler
 		implements MessageProducer, HeaderPropagationAware {
 
-	protected final MessagingTemplate messagingTemplate = new MessagingTemplate(); // NOSONAR final
+	protected final MessagingTemplate messagingTemplate = new MessagingTemplate();
 
 	private boolean async;
 
@@ -88,7 +88,7 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	@Nullable
 	private MessageChannel outputChannel;
 
-	private String[] notPropagatedHeaders;
+	private String @Nullable [] notPropagatedHeaders;
 
 	private boolean selectiveHeaderPropagation;
 
@@ -113,7 +113,7 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	@Override
 	public void setOutputChannelName(String outputChannelName) {
 		Assert.hasText(outputChannelName, "'outputChannelName' must not be empty");
-		this.outputChannelName = outputChannelName; //NOSONAR (inconsistent sync)
+		this.outputChannelName = outputChannelName;
 	}
 
 	/**
@@ -213,12 +213,10 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	@Override
 	protected void onInit() {
 		super.onInit();
-		Assert.state(!(this.outputChannelName != null && this.outputChannel != null), //NOSONAR (inconsistent sync)
+		Assert.state(!(this.outputChannelName != null && this.outputChannel != null),
 				"'outputChannelName' and 'outputChannel' are mutually exclusive.");
 		BeanFactory beanFactory = getBeanFactory();
-		if (beanFactory != null) {
-			this.messagingTemplate.setBeanFactory(beanFactory);
-		}
+		this.messagingTemplate.setBeanFactory(beanFactory);
 		this.messagingTemplate.setDestinationResolver(getChannelResolver());
 		setAsyncIfCan();
 		if (!this.sendTimeoutSet) {
@@ -244,9 +242,9 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 		return this.outputChannel;
 	}
 
-	protected void sendOutputs(Object result, Message<?> requestMessage) {
-		if (result instanceof Iterable<?> && shouldSplitOutput((Iterable<?>) result)) {
-			for (Object o : (Iterable<?>) result) {
+	protected void sendOutputs(@Nullable Object result, Message<?> requestMessage) {
+		if (result instanceof Iterable<?> iterableResult && shouldSplitOutput(iterableResult)) {
+			for (Object o : iterableResult) {
 				produceOutput(o, requestMessage);
 			}
 		}
@@ -294,13 +292,12 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	private Map<?, ?> obtainRoutingSlipHeader(MessageHeaders requestHeaders, Object reply) {
 		Map<?, ?> routingSlipHeader = requestHeaders.get(IntegrationMessageHeaderAccessor.ROUTING_SLIP, Map.class);
 		if (routingSlipHeader == null) {
-			if (reply instanceof Message) {
-				routingSlipHeader = ((Message<?>) reply).getHeaders()
+			if (reply instanceof Message<?> replyMessage) {
+				routingSlipHeader = replyMessage.getHeaders()
 						.get(IntegrationMessageHeaderAccessor.ROUTING_SLIP, Map.class);
 			}
-			else if (reply instanceof AbstractIntegrationMessageBuilder<?>) {
-				routingSlipHeader = ((AbstractIntegrationMessageBuilder<?>) reply)
-						.getHeader(IntegrationMessageHeaderAccessor.ROUTING_SLIP, Map.class);
+			else if (reply instanceof AbstractIntegrationMessageBuilder<?> messageBuilder) {
+				routingSlipHeader = messageBuilder.getHeader(IntegrationMessageHeaderAccessor.ROUTING_SLIP, Map.class);
 			}
 		}
 		return routingSlipHeader;
@@ -310,12 +307,11 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	private Object obtainReplyChannel(MessageHeaders requestHeaders, Object reply) {
 		Object replyChannel = requestHeaders.getReplyChannel();
 		if (replyChannel == null) {
-			if (reply instanceof Message) {
-				replyChannel = ((Message<?>) reply).getHeaders().getReplyChannel();
+			if (reply instanceof Message<?> replyMessage) {
+				replyChannel = replyMessage.getHeaders().getReplyChannel();
 			}
-			else if (reply instanceof AbstractIntegrationMessageBuilder<?>) {
-				replyChannel = ((AbstractIntegrationMessageBuilder<?>) reply)
-						.getHeader(MessageHeaders.REPLY_CHANNEL, Object.class);
+			else if (reply instanceof AbstractIntegrationMessageBuilder<?> messageBuilder) {
+				replyChannel = messageBuilder.getHeader(MessageHeaders.REPLY_CHANNEL, Object.class);
 			}
 		}
 		return replyChannel;
@@ -438,7 +434,7 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 		return builder;
 	}
 
-	private Object getOutputChannelFromRoutingSlip(Object reply, Message<?> requestMessage, List<?> routingSlip,
+	private @Nullable Object getOutputChannelFromRoutingSlip(Object reply, Message<?> requestMessage, List<?> routingSlip,
 			AtomicInteger routingSlipIndex) {
 
 		if (routingSlipIndex.get() >= routingSlip.size()) {
@@ -446,7 +442,7 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 		}
 
 		Object path = routingSlip.get(routingSlipIndex.get());
-		Object routingSlipPathValue = null;
+		Object routingSlipPathValue;
 
 		if (path instanceof String string) {
 			routingSlipPathValue = getBeanFactory().getBean(string);
@@ -555,11 +551,13 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 		return true;
 	}
 
-	protected void sendErrorMessage(Message<?> requestMessage, Throwable ex) {
+	protected void sendErrorMessage(Message<?> requestMessage, @Nullable Throwable ex) {
 		Object errorChannel = resolveErrorChannel(requestMessage.getHeaders());
 		Throwable result = ex;
-		if (!(ex instanceof MessagingException)) {
-			result = new MessageHandlingException(requestMessage, ex);
+		if (!(result instanceof MessagingException)) {
+			result = ex == null
+					? new MessageHandlingException(requestMessage)
+					: new MessageHandlingException(requestMessage, ex);
 		}
 		if (errorChannel == null) {
 			logger.error(result,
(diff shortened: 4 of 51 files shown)


spring-integration-core/src/main/java/org/springframework/integration/core/RecoveryCallback.java
```diff
@@ -35,6 +35,6 @@ public interface RecoveryCallback<T extends @Nullable Object> {
 	 * @param cause the cause of the failure
 	 * @return an Object that can be used to replace the callback result that failed
 	 */
-	T recover(AttributeAccessor context, Throwable cause);
+	T recover(AttributeAccessor context, @Nullable Throwable cause);
```
artembilan: This is not correct.
That was really intentional when I have introduced this contract.
See its Javadocs.
It is like error handler in the `catch` block.
So, what kind of error handler it is if it can be used without an error?

This has caused a lot of problems in the project.
Try to build it locally first before pushing changes.

spring-integration-core/src/main/java/org/springframework/integration/core/ErrorMessagePublisher.java
```diff
@@ -70,7 +70,7 @@ public final void setErrorMessageStrategy(ErrorMessageStrategy errorMessageStrat
 		this.errorMessageStrategy = errorMessageStrategy;
 	}
 
-	public final void setChannel(MessageChannel channel) {
+	public final void setChannel(@Nullable MessageChannel channel) {
```
artembilan, line 73: Honestly, I am not a fun of setters accepting nulls.
Let's revise the place where it is used with null instead!

Plus, this doesn't look like a scope of your task for the `handler` package.

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProcessorMessageSource.java
```diff
@@ -45,7 +47,9 @@ public String getComponentType() {
 
 	@Override
 	protected @Nullable Object doReceive() {
-		return this.messageProcessor.processMessage(null);
+		Message<Object> fakeMessage = MutableMessageBuilder.withPayload(new Object(), false).build();
```
artembilan: This has to go to `static final` property of this class.
The message is fake, so it really can be reused every single time.
Then add your name to the `@author` list.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -555,11 +551,13 @@ protected boolean shouldCopyRequestHeaders() {
 		return true;
 	}
 
-	protected void sendErrorMessage(Message<?> requestMessage, Throwable ex) {
+	protected void sendErrorMessage(Message<?> requestMessage, @Nullable Throwable ex) {
```
artembilan: I don't believe that this is correct.
I just cannot find where we pass null to this method.
Please, revise.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -652,7 +648,7 @@ else if (exception != null) {
 			}
 		}
 
-		private void onFailure(Throwable ex) {
+		private void onFailure(@Nullable Throwable ex) {
```
artembilan: I see where you get it.
This is not OK, please, revise that `onFailure(exception instanceof CompletionException ? exception.getCause() : exception);`
And if `exception.getCause() == null` , propagate an `exception` itself.

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractReplyProducingMessageHandler.java
```diff
@@ -160,6 +160,7 @@ else if (!isAsync() && isLoggingEnabled()) {
 		}
 	}
 
+	@SuppressWarnings("NullAway") // dataflow analysis limitation
 	@Nullable
```
artembilan: If we already here, please, move `@Nullable` just before the return type.

spring-integration-core/src/main/java/org/springframework/integration/handler/ExpressionEvaluatingMessageHandler.java
```diff
@@ -35,6 +35,7 @@ public class ExpressionEvaluatingMessageHandler extends AbstractMessageHandler {
 
 	private final ExpressionEvaluatingMessageProcessor<Void> processor;
 
+	@SuppressWarnings("NullAway.Init")
```
artembilan: Well, with a recently raised concern for the `getComponentType()`, we should revise this as well.
Looks like we just don't need this property here and respective setter at all.
I believe this should be like:
```
@Override
	public String getComponentType() {
		return "expression-evaluating-handler";
	}
```

spring-integration-core/src/main/java/org/springframework/integration/handler/LambdaMessageProcessor.java
```diff
@@ -66,6 +66,7 @@ public class LambdaMessageProcessor implements MessageProcessor<Object>, BeanFac
 
 	private final Class<?>[] parameterTypes;
 
+	@SuppressWarnings("NullAway.Init")
```
artembilan, line 68: I cannot comment on a specific line, but since we are in this class already, please, move the `@Nullable` next to the type of the `expectedType` property.

spring-integration-core/src/main/java/org/springframework/integration/handler/MethodInvokingMessageHandler.java
```diff
@@ -36,6 +36,7 @@ public class MethodInvokingMessageHandler extends AbstractMessageHandler impleme
 
 	private final MethodInvokingMessageProcessor<Object> processor;
 
+	@SuppressWarnings("NullAway.Init")
```
artembilan: Same here: remove this property and just do return `"method-invoking-handler";`

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/AbstractRequestHandlerAdvice.java
```diff
@@ -113,6 +115,7 @@ protected Throwable unwrapThrowableIfNecessary(Exception e) {
 		Throwable actualThrowable = e;
 		if (e instanceof ThrowableHolderException) {
 			actualThrowable = e.getCause();
+			Assert.state(actualThrowable != null, "'actualThrowable' must not be null");
```
artembilan: What if we override `getCause()` in the `ThrowableHolderException` with non-null contract?
Or we can have a different method which would return the cause of us as non-null...

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ErrorMessageSendingRecoverer.java
```diff
@@ -72,7 +74,7 @@ public ErrorMessageSendingRecoverer(MessageChannel channel) {
 	 * @since 4.3.10
 	 */
 	@SuppressWarnings("this-escape")
-	public ErrorMessageSendingRecoverer(MessageChannel channel, ErrorMessageStrategy errorMessageStrategy) {
+	public ErrorMessageSendingRecoverer(@Nullable MessageChannel channel, @Nullable ErrorMessageStrategy errorMessageStrategy) {
```
artembilan: Let's consider to remove all the `@Nullable` from this ctors!
The first ctor would be just:
```
	public ErrorMessageSendingRecoverer() {
		this(null);
	}
```
Second just:
```
public ErrorMessageSendingRecoverer(@Nullable MessageChannel channel) {
		setChannel(channel);
	}
```
and third without any checks.
The `DefaultErrorMessageStrategy` is there in super class by default anyway.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ErrorMessageSendingRecoverer.java
```diff
@@ -81,7 +83,7 @@ public ErrorMessageSendingRecoverer(MessageChannel channel, ErrorMessageStrategy
 	}
 
 	@Override
-	public Object recover(AttributeAccessor context, Throwable cause) {
+	public @Nullable Object recover(AttributeAccessor context, @Nullable Throwable cause) {
```
artembilan: Remove `@Nullable` on the `cause` as we discussed before.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ExpressionEvaluatingRequestHandlerAdvice.java
```diff
@@ -237,6 +235,7 @@ protected Object doInvoke(ExecutionCallback callback, Object target, Message<?>
 		}
 	}
 
+	@SuppressWarnings("NullAway") // dataflow analysis limitation
```
artembilan: Can we revise this, please?
I'm not sure where exactly this method is breaking Nullability contract....

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ExpressionEvaluatingRequestHandlerAdvice.java
```diff
@@ -258,7 +257,8 @@ private void evaluateSuccessExpression(Message<?> message) {
 		}
 	}
 
-	private Object evaluateFailureExpression(Message<?> message, Exception exception) {
+	@SuppressWarnings("NullAway") // dataflow analysis limitation
```
artembilan: Same here, please.

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -1086,6 +1095,8 @@ public Object invoke(ParametersWrapper parameters) {
 				message = new MutableMessage<>(parameters.getMessages(), parameters.getHeaders());
 			}
 			try {
+				Assert.state(message != null, "'message' must not be null");
+				Assert.state(this.invocableHandlerMethod != null, "'invocableHandlerMethod' must not be null");
```
artembilan: Let's mark this method with a `NullAway` since this is a critical execution path.
And we are here, neither of these variables is null.

spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java
```diff
@@ -1382,18 +1393,19 @@ public Collection<Message<?>> getMessages() {
...
-		public Message<?> getMessage() {
+		public @Nullable Message<?> getMessage() {
 			return this.message;
 		}
 
 		public Class<?> getFirstParameterType() {
 			if (this.payload != null) {
 				return this.payload.getClass();
 			}
+			Assert.state(this.messages != null, "'messages' must not be null");
```
artembilan: Use `NullAway` instead.
When we are here, the `this.messages` is not null.
I understand that different classes could be used for single message execution and a list of them respectively.
But that is a different story not related to your already huge PR.

spring-integration-core/src/main/java/org/springframework/integration/handler/support/PayloadExpressionArgumentResolver.java
```diff
@@ -50,11 +53,12 @@ public boolean supportsParameter(MethodParameter parameter) {
 	}
 
 	@Override
-	public Object resolveArgument(MethodParameter parameter, Message<?> message) {
+	public @Nullable Object resolveArgument(MethodParameter parameter, Message<?> message) {
 		Expression expression = this.expressionCache.get(parameter);
 		if (expression == null) {
 			Payload ann = parameter.getParameterAnnotation(Payload.class);
-			expression = EXPRESSION_PARSER.parseExpression(ann.expression()); // NOSONAR never null - supportsParameter()
+			Assert.state(ann != null, "'Payload' annotation must not be null");
```
artembilan: `NullAway` instead, please.
You have just remove that `NOSONAR` comment, but didn't pay attention what it says.
So, when we here, the `@Payload` is there on the param.
Therefore never null.
And this extra `state()` call is abit of burden for a critical path execution.
Well, I know that result is going to be cached, but still why do extra call.
Can we also consider to use `computeIfAbsent()`, too, instead of `if` block?

spring-integration-core/src/main/java/org/springframework/integration/handler/support/PayloadsArgumentResolver.java
```diff
@@ -56,15 +58,16 @@ public boolean supportsParameter(MethodParameter parameter) {
...
+	public @Nullable Object resolveArgument(MethodParameter parameter, Message<?> message) {
 		Object payload = message.getPayload();
 		Assert.state(payload instanceof Collection,
 				"This Argument Resolver support only messages with payload as Collection<Message<?>>");
 		Collection<Message<?>> messages = (Collection<Message<?>>) payload;
 
 		if (!this.expressionCache.containsKey(parameter)) {
 			Payloads payloads = parameter.getParameterAnnotation(Payloads.class);
-			String expression = payloads.value(); // NOSONAR never null - supportsParameter()
+			Assert.state(payloads != null, "'Payload' annotation must not be null");
```
artembilan: DITTO

spring-integration-core/src/main/java/org/springframework/integration/util/CheckedCallable.java
```diff
@@ -34,6 +34,7 @@
 @FunctionalInterface
 public interface CheckedCallable<T extends @Nullable Object, E extends Throwable> {
 
+	@Nullable
```
artembilan: That's not.
Please, bring it back.
The `T extends @Nullable Object` on the interface itself is enough.
The rest of complains are IDEA flaws.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/LockRegistry.java
```diff
@@ -126,7 +128,7 @@ default <E extends Throwable> void executeLocked(Object lockKey, Duration waitLo
 	 * @throws TimeoutException when {@link Lock#tryLock(long, TimeUnit)} has elapsed
 	 * @since 6.2
 	 */
-	default <T, E extends Throwable> T executeLocked(Object lockKey, Duration waitLockDuration,
+	default <T, E extends Throwable> @Nullable T executeLocked(Object lockKey, Duration waitLockDuration,
```
artembilan: Please, revert.
Is not related to your `handler` package.
And the way it is works well with NullAway.
IDEA concerns are out of our scope.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
```diff
@@ -280,7 +281,7 @@ public void setGroupCondition(Object groupId, String condition) {
 
 	protected abstract @Nullable Message<?> doPollMessageFromGroup(Object groupId);
 
-	protected <T, E extends RuntimeException> T executeLocked(Object groupId, CheckedCallable<T, E> runnable) {
+	protected <T, E extends RuntimeException> @Nullable T executeLocked(Object groupId, CheckedCallable<T, E> runnable) {
```
artembilan: Please, revert.

spring-integration-core/src/main/java/org/springframework/integration/store/AbstractMessageGroupStore.java
```diff
@@ -245,7 +245,8 @@ public void removeMessageGroup(Object groupId) {
 
 	@Override
 	public boolean removeMessageFromGroupById(Object groupId, UUID messageId) {
-		return executeLocked(groupId, () -> doRemoveMessageFromGroupById(groupId, messageId));
+		Boolean removed = executeLocked(groupId, () -> doRemoveMessageFromGroupById(groupId, messageId));
+		return removed != null && removed;
```
artembilan: DITTO.

artembilan (changes requested):
That's a lot, but great.
Let's see if we can manage such a big review as well!

spring-integration-core/src/main/java/org/springframework/integration/core/RecoveryCallback.java
mjd507: yes, I am doubtful when adding Nullable on Throwable.

we have one place in `RequestHandlerRetryAdvice` can possibly use it without providing an error?
```java
RequestHandlerRetryAdvice.java:69: error: [NullAway] passing @Nullable parameter 'context.getLastThrowable()' where @NonNull is required
                                recoveryCallback.recover(context, context.getLastThrowable());
```

artembilan: Let's ignore that with `NullAway` for now!
That retry component is going to be reworked soon enough: https://github.com/spring-projects/spring-integration/issues/10345.
After respective Spring Kafka issue is fixed before: https://github.com/spring-projects/spring-kafka/issues/4051.

spring-integration-core/src/main/java/org/springframework/integration/support/locks/LockRegistry.java
mjd507: oh, because I add Nullable on `ExecutionCallback.execute()`,  the implementation is `MethodInvocation.proceed()`, which could return null.

And the caller `LockRequestHandlerAdvice.doInvoke()` call this `lockRegistry.executeLocked(lockKey, callback::execute)`, then build with following error, that's I add `Nullable` on `CheckedCallable.call() `

```java
LockRequestHandlerAdvice.java:165: error: [NullAway] referenced method returns @Nullable, but functional interface method org.springframework.integration.util.CheckedCallable.call() returns @NonNull
                                        return this.lockRegistry.executeLocked(lockKey, waitLockDuration, callback::execute);
                                                                                                          ^
```

artembilan: See `AbstractMessageGroupStore` for the solution:
```
	@Override
	public @Nullable Message<?> pollMessageFromGroup(Object groupId) {
		return this.<@Nullable Message<?>, RuntimeException>executeLocked(groupId, () -> doPollMessageFromGroup(groupId));
	}
```

So, this your `LockRequestHandlerAdvice` case should be like as well:
```
					return this.lockRegistry.<@Nullable Object, RuntimeException>executeLocked(lockKey, callback::execute);
```

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -644,11 +642,12 @@ public void accept(@Nullable Object result, @Nullable Throwable exception) {
 				}
 			}
 			else if (exception != null) {
-				onFailure(exception instanceof CompletionException ? exception.getCause() : exception);
+				onFailure(exception instanceof CompletionException && exception.getCause() != null
+						? exception.getCause() : exception);
```
artembilan, line 646: Can we extra a local variable instead of calling `getCause()` several times?
`if..else` would be great before calling this `onFailure()`.

spring-integration-core/src/main/java/org/springframework/integration/handler/ExpressionEvaluatingMessageHandler.java
```diff
@@ -35,8 +35,7 @@ public class ExpressionEvaluatingMessageHandler extends AbstractMessageHandler {
 
 	private final ExpressionEvaluatingMessageProcessor<Void> processor;
 
-	@SuppressWarnings("NullAway.Init")
-	private String componentType;
+	private String componentType = "expression-evaluating-handler";
```
artembilan, line 38: No, I mean remove `setComponentType()` at all.
This contract is not meant to be overridden by end-user.

spring-integration-core/src/main/java/org/springframework/integration/handler/MethodInvokingMessageHandler.java
```diff
@@ -36,8 +36,7 @@ public class MethodInvokingMessageHandler extends AbstractMessageHandler impleme
 
 	private final MethodInvokingMessageProcessor<Object> processor;
 
-	@SuppressWarnings("NullAway.Init")
-	private String componentType;
+	private String componentType = "method-invoking-handler";
```
artembilan, line 39: DITTO

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ErrorMessageSendingRecoverer.java
```diff
@@ -59,8 +59,9 @@ public ErrorMessageSendingRecoverer() {
 	 * The {@link DefaultErrorMessageStrategy} is used for building an error message to publish.
 	 * @param channel the message channel to publish error messages on recovery action.
 	 */
+	@SuppressWarnings("this-escape")
 	public ErrorMessageSendingRecoverer(@Nullable MessageChannel channel) {
```
artembilan, line 63: This must not be `null`.
Looks like you have missed some parts of my previous explanation.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/ErrorMessageSendingRecoverer.java
```diff
@@ -74,7 +75,7 @@ public ErrorMessageSendingRecoverer(@Nullable MessageChannel channel) {
 	 * @since 4.3.10
 	 */
 	@SuppressWarnings("this-escape")
-	public ErrorMessageSendingRecoverer(@Nullable MessageChannel channel, @Nullable ErrorMessageStrategy errorMessageStrategy) {
+	public ErrorMessageSendingRecoverer(MessageChannel channel, ErrorMessageStrategy errorMessageStrategy) {
```
artembilan, line 78: According to this our contract that ternary operator in the `setErrorMessageStrategy()` call is redundant.

mjd507:
ok, I have addressed most review comments, two of them are not revised due to it may take me more time.
1. `LockRequestHandlerAdvice#doInvoke `, I add NullAway for now, `this.lockRegistry.<@Nullable Object, RuntimeException>executeLocked` seems not working.
2. I did not remove `componentType` and setter method in `ExpressionEvaluatingMessageHandler` & `MethodInvokingMessageHandler `,  unit test looks covers that. 

@artembilan  happy Friday,  if any new reviews I will solve on next week. 
if hurry, you always can take from where it is, and help me fix that.

spring-integration-core/src/main/java/org/springframework/integration/core/ErrorMessagePublisher.java
mjd507: I leave nullable here, one of ctors will pass null to this method.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/LockRequestHandlerAdvice.java
```diff
@@ -152,6 +152,7 @@ protected void onInit() {
 		this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
 	}
 
+	@SuppressWarnings("NullAway") // CheckedCallable.call() is nullable
```
artembilan, line 155: I still have doubts in this.
Will take a look one more time when the rest of concerns are addressed.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/handler/ExpressionEvaluatingMessageHandler.java
mjd507: yeah, I tried that, unit test failed a lot, there it will call `setComponentType()`

artembilan:
> if hurry, you always can take from where it is, and help me fix that.

Sounds good!
So, I'm taking it from here since looks like we have some problems which better to solve sooner.

Pulling your branch locally for clean up and merge.

Thank you!

artembilan:
Fixed via: https://github.com/spring-projects/spring-integration/commit/a167170aa870d84a79ece1b061d23a8ef17cf1b4.
And subsequent cleanup and improvements: https://github.com/spring-projects/spring-integration/commit/31a7dd496db35de82f1d060754f768ed6554e6c2

Thank you again!

PR #10373, opened 2025-09-02
GH-10372: Fix the ignoring logic for tmp files

### LINKED ISSUE #10372: Local temporary file suffix filter is combined in a CompositeFileListFilter instead of a ChainFileListFilter, therefore a OR instead of a AND

6.5.0

**Describe the bug**

Local temporary file suffix filter is combined in a CompositeFileListFilter instead of a ChainFileListFilter, therefore a ( localFileListfilter || temporaryFileSuffixFilter ) instead of ( localFileListfilter && temporaryFileSuffixFilter ).

If the file is beeing written as .writting suffix indicates, it shouldn't be available for message creation, yet, even if other custom filters meet the conditions.

The temporaryFileSuffixFilter should be first in the chain.

I got around that by creating a localfilelistfilter chain that includes first the same filesuffix regex filter

```
	private FileListFilter<File> buildFilter() {
		Pattern completePattern = Pattern.compile("^.*(?<!" + this.synchronizer.getTemporaryFileSuffix() + ")$");
		return new CompositeFileListFilter<>(
				Arrays.asList(this.localFileListFilter, new RegexPatternFileListFilter(completePattern)));
	}
```

https://github.com/spring-projects/spring-integration/blob/cf6ede6569ac583a5049957adebe32010b109199/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java#L286

**To Reproduce**

Issue occurs if you add custom localFileListFilter.

**Expected behavior**

.writting files should not be accepted no matter what

**Sample**

Sorry I can't push code from the corporation I am writting this repport.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10372

The `AbstractInboundFileSynchronizingMessageSource` uses a `CompositeFileListFilter` for a combination of the user-provided filter and then the one to ignore temporary files. However, the `CompositeFileListFilter` logic is `OR`, which means the temporary file might be accepted by the user-provided.

* Fix the `AbstractInboundFileSynchronizingMessageSource.buildFilter()` to use a `ChainFileListFilter` instead (with an `AND` logic), and put `ignoreTemporaryFiles` filter as the first one.
* Fix (S)FTP docs for an actual logic regarding local files filtering.

**Cherry-pick to `6.5.x`**

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/remote/synchronizer/AbstractInboundFileSynchronizingMessageSource.java
@@ -20,8 +20,8 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
-import java.util.Arrays;
 import java.util.Comparator;
+import java.util.List;
 import java.util.regex.Pattern;
 
 import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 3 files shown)


artembilan:
NOTE: It won't cherry-pick clean since there is no proper Nullability.
So, some manual conflicts fixing would be necessary.

Thanks

cppwfs (approved):
LGTM.  Thanks for fixing this!

PR #10375, opened 2025-09-03
Rework `camel.adoc` to reflect reality

artembilan:
* It is better to have links to external docs independent of the version
* Rework and change the section of the `camel.adoc` talking about Camel component for Spring Integration. This component was removed from Apache Camel since `4.0`


cppwfs (approved):
LGTM.   Thanks for updating the docs to reflect the updates in the community.

PR #10380, opened 2025-09-03
GH-10379: Optimize Observability API usage

### LINKED ISSUE #10379: Bypass observation in MessageProducerSupport when NOOP registry is configured

6.5.1

**Describe the bug**

We've been looking into ways to improve our app's messaging throughput in high throughput messaging scenarios.

One thing we've tested out and noticed was that `MessageProducerSupport.sendMessage()` always creates observability overhead (calling `IntegrationObservation.HANDLER`) even when `ObservationRegistry.NOOP` is configured. This seems to result in a performance impact of roughly 15-25% or so in very high throughput messaging scenarios.

**To Reproduce**

Just run messages as fast as possible through `MessageProducerSupport` with NOOP observability configured (the default), and watch the throughput.

**Expected behavior**

When `ObservationRegistry.NOOP` is configured, observability overhead should be eliminated entirely. The method should detect NOOP registry and bypass observability infrastructure for optimal performance.

One potential solution to this is to add a simple check in `MessageProducerSupport.sendMessage()`:
```java
if (this.observationRegistry.isNoop()) {
    this.messagingTemplate.send(getRequiredOutputChannel(), trackMessageIfAny(message));  // Direct send
} else {
    // existing observability path
}
```

See this commit for an actual example fix: https://github.com/Nephery/spring-integration/commit/fbb541a6d6758412c78d1a9576bfc1f551839648

**Sample**

1. Clone repository: https://github.com/Nephery/spring-integration-noop-observability-perf
2. Follow [the setup instructions](https://github.com/Nephery/spring-integration-noop-observability-perf?tab=readme-ov-file#setup) to install the spring integration fork.
3. Follow [these instructions](https://github.com/Nephery/spring-integration-noop-observability-perf?tab=readme-ov-file#running-the-test) to run the test twice; once with the current spring integration version, and once more with the bypass solution.
4. Compare the test results between the regular and optimized runs, and notice that the optimized run is a bit better.
    * From running this on my personal computer, [these are the results I got](https://github.com/Nephery/spring-integration-noop-observability-perf?tab=readme-ov-file#performance-results) with roughly a 18% performance increase.

**Note:** When using this NOOP observability bypass in one of our actual apps, we see similar relative performance gains in these very high throughput scenarios.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10379

The `ObservationRegistry.NOOP` still brings some performance overhead.

* Fix `MessageProducerSupport` & `SourcePollingChannelAdapter` to check for `isObserved()` before using the `IntegrationObservation`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
@@ -255,18 +255,27 @@ public abstract class MessageProducerSupport extends AbstractEndpoint
 		}
 
 		try {
-			IntegrationObservation.HANDLER.observation(
-							this.observationConvention,
-							DefaultMessageReceiverObservationConvention.INSTANCE,
-							() -> new MessageReceiverContext(message, getComponentName(), "message-producer"),
-							this.observationRegistry)
-					.observe(() -> this.messagingTemplate.send(getRequiredOutputChannel(), trackMessageIfAny(message)));
+			if (isObserved()) {
+				IntegrationObservation.HANDLER.observation(
+								this.observationConvention,
+								DefaultMessageReceiverObservationConvention.INSTANCE,
+								() -> new MessageReceiverContext(message, getComponentName(), "message-producer"),
+								this.observationRegistry)
+						.observe(() -> sendMessageWithTracking(message));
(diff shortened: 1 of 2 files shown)


cppwfs (approved):
LGTM.   Great optimization!

PR #10382, opened 2025-09-04
GH-10083 Add Nullability to the mapping package

cppwfs:
Related to: #10083
* AbstractHeaderMapper.java - Enhanced null safety by adding @Nullable annotations to fields and methods.
* Updated method signatures to match those in AbstractHeaderMapper for the following classes:
   - DefaultAmqpHeaderMapper.java
   - DefaultSoapHeaderMapper.java
   - DefaultXmppHeaderMapper.java

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/mapping/AbstractHeaderMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/mapping/AbstractHeaderMapper.java
@@ -310,9 +310,7 @@ public abstract class AbstractHeaderMapper<T> implements RequestReplyHeaderMappe
 			@Nullable Object keyId) {
 
 		try {
-			var classLoader = getClassLoader();
-			Assert.state(classLoader != null, "No ClassLoader available");
-			return JsonHeaders.buildResolvableType(classLoader, typeId, contentId, keyId);
+			return JsonHeaders.buildResolvableType(getClassLoader(), typeId, contentId, keyId);
 		}
 		catch (Exception e) {
 			this.logger.debug("Cannot build a ResolvableType from 'json__TypeId__' header", e);
--- a/spring-integration-core/src/main/java/org/springframework/integration/mapping/AbstractHeaderMapper.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/mapping/AbstractHeaderMapper.java
@@ -51,6 +51,7 @@ import org.springframework.util.StringUtils;
  * @author Stephane Nicoll
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Glenn Renfro
  *
  * @since 2.1
  */
@@ -87,7 +88,7 @@ public abstract class AbstractHeaderMapper<T> implements RequestReplyHeaderMappe
 
 	private HeaderMatcher replyHeaderMatcher;
 
-	private ClassLoader classLoader = ClassUtils.getDefaultClassLoader();
+	private @Nullable ClassLoader classLoader = ClassUtils.getDefaultClassLoader();
 
 	/**
 	 * Create a new instance.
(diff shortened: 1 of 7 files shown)


spring-integration-core/src/main/java/org/springframework/integration/mapping/AbstractHeaderMapper.java
```diff
@@ -308,7 +310,9 @@ private ResolvableType createJsonResolvableTypHeaderInAny(Object typeId, @Nullab
 			@Nullable Object keyId) {
 
 		try {
-			return JsonHeaders.buildResolvableType(getClassLoader(), typeId, contentId, keyId);
+			var classLoader = getClassLoader();
+			Assert.state(classLoader != null, "No ClassLoader available");
```
artembilan: Looks like Spring deals with null `ClassLoader` everywhere.
We should revise that `JsonHeaders.buildResolvableType()` to accept `null`, respectively.
In the end it uses `ClassUtils.forName()` which accepts null well.

spring-integration-core/src/main/java/org/springframework/integration/mapping/support/JsonHeaders.java
```diff
@@ -103,9 +105,9 @@ else if (classValue != null) {
 	 * @return the {@link ResolvableType} based on provided class components
 	 * @since 5.2.4
 	 */
-	public static ResolvableType buildResolvableType(Class<?> targetClass, @Nullable Class<?> contentClass,
+	public static ResolvableType buildResolvableType(@Nullable Class<?> targetClass, @Nullable Class<?> contentClass,
 			@Nullable Class<?> keyClass) {
-
+		Assert.state(targetClass != null, "The targetClass must not be null");
```
artembilan: This is not correct.
The `targetClass` must not be null.

Consider to mark `getClassForValue()` with a `@Contract` we talked yesterday in other place:
```
@Contract("_, null -> null; _, !null -> !null")
```

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/mapping/support/JsonHeaders.java
```diff
@@ -105,9 +106,9 @@ else if (classValue != null) {
 	 * @return the {@link ResolvableType} based on provided class components
 	 * @since 5.2.4
 	 */
-	public static ResolvableType buildResolvableType(@Nullable Class<?> targetClass, @Nullable Class<?> contentClass,
+	@Contract("_, !null, _ -> !null; _, _, !null -> !null")
```
artembilan: This is wrong.
The method does not return `null` at all.
See `getClassForValue()`.
That's where the `@Contract` has to be.

spring-integration-core/src/main/java/org/springframework/integration/mapping/support/JsonHeaders.java
```diff
@@ -69,18 +70,18 @@ private JsonHeaders() {
...
 	 */
-	public static ResolvableType buildResolvableType(ClassLoader classLoader, Object targetClassValue,
+	public static ResolvableType buildResolvableType(@Nullable ClassLoader classLoader, Object targetClassValue,
 			@Nullable Object contentClassValue, @Nullable Object keyClassValue) {
 
 		Class<?> targetClass = getClassForValue(classLoader, targetClassValue);
 		Class<?> keyClass = getClassForValue(classLoader, keyClassValue);
 		Class<?> contentClass = getClassForValue(classLoader, contentClassValue);
-
+		Assert.notNull(targetClass, "targetClass must not be null");
```
artembilan: When you fix `getClassForValue()` with the `@Contract`, this is not necessary.

spring-integration-core/src/main/java/org/springframework/integration/mapping/support/JsonHeaders.java
```diff
@@ -69,18 +70,18 @@ private JsonHeaders() {
...
 		return buildResolvableType(targetClass, contentClass, keyClass);
 	}
 
 	@Nullable
```
artembilan: Please, move this next to the return type.

artembilan (changes requested):

artembilan (approved):
LGMT.
Will merge when build is green.

PR #10384, opened 2025-09-05
GH-10083: Add nullability to core metadata

cppwfs:
* ConcurrentMetadataStore.java: Added `@Nullable` annotation to putIfAbsent method return type for better null safety.
* MetadataStore.java: Added @Nullable annotations to get and remove method return types to indicate they can return null.
* PropertiesPersistingMetadataStore.java: Removed explicit null checks with Assert.notNull calls and added `@Nullable` annotations to method signatures while simplifying saveMetadata logic.
* SimpleMetadataStore.java: Added `@Nullable` annotations to get, remove, and putIfAbsent method return types for consistency with interface.
* MetadataStoreSelector.java: Added Assert.state calls to validate that keyStrategy and valueStrategy do not return null values.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/metadata/ConcurrentMetadataStore.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/metadata/ConcurrentMetadataStore.java
@@ -16,10 +16,13 @@
 
 package org.springframework.integration.metadata;
 
+import org.jspecify.annotations.Nullable;
+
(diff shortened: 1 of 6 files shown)


artembilan (approved):
Nice, smooth and looks easy.

I guess since the build is green, all the other `MetadataStore` implementations in the project have Nullability already addressed 😄 .

Thank you!

PR #10387, opened 2025-09-05
GH-10345: Migrate Kafka module to Core Retry

artembilan:
Related to: https://github.com/spring-projects/spring-integration/issues/10345

* Replace `RetryTemplate` from Spring Retry to `RetryTemplate` from Spring Core since Spring Retry is at its sunset.
There is no deprecation for Spring Retry API usage since the idea is to remove Spring Retry dependency altogether for the upcoming major for all projects in the Spring portfolio

* Introduce a `KafkaInboundEndpoint.RetryContext` as an `AttributeAccessorSupport` extension to track the number of retries for functional compatibility

### DIFF
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
@@ -41,19 +41,19 @@ import org.springframework.kafka.support.Acknowledgment;
 public interface KafkaInboundEndpoint {
 
 	/**
-	 * {@link org.springframework.retry.RetryContext} attribute key for an acknowledgment
+	 * The {@link RetryContext} attribute key for an acknowledgment
 	 * if the listener is capable of acknowledging.
 	 */
 	String CONTEXT_ACKNOWLEDGMENT = "acknowledgment";
 
 	/**
-	 * {@link org.springframework.retry.RetryContext} attribute key for the consumer if
+	 * The {@link RetryContext} attribute key for the consumer if
 	 * the listener is consumer-aware.
 	 */
 	String CONTEXT_CONSUMER = "consumer";
 
 	/**
-	 * {@link org.springframework.retry.RetryContext} attribute key for the record.
+	 * The {@link RetryContext} attribute key for the record.
 	 */
 	String CONTEXT_RECORD = "record";
 
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
@@ -21,16 +21,20 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.core.AttributeAccessor;
+import org.springframework.core.AttributeAccessorSupport;
+import org.springframework.core.retry.RetryException;
+import org.springframework.core.retry.RetryTemplate;
+import org.springframework.integration.core.RecoveryCallback;
 import org.springframework.kafka.KafkaException;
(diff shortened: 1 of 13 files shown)


spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
```diff
@@ -67,24 +71,46 @@ public interface KafkaInboundEndpoint {
 	default void doWithRetry(RetryTemplate template, @Nullable RecoveryCallback<?> callback, ConsumerRecord<?, ?> record,
```
cppwfs, line 71: Shouldn't doWithRetry be `@Nullable` since it can return a null?

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -368,19 +357,18 @@ private void doSendAndReceive(Message<?> message) {
...
-				if (KafkaInboundGateway.this.retryTemplate == null) {
-					ATTRIBUTES_HOLDER.remove();
-				}
+				ATTRIBUTES_HOLDER.remove();
 			}
 		}
 
 		private Message<?> enhanceHeadersAndSaveAttributes(Message<?> message, ConsumerRecord<K, V> record) {
 			Message<?> messageToReturn = message;
+			AttributeAccessor retryContext = ATTRIBUTES_HOLDER.get();
```
cppwfs, line 366: We use the type `RetryContext` here instead of casting in other sections of the code. WDYT?

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -510,9 +501,10 @@ private Message<?> enhanceHeadersAndSaveAttributes(Message<?> message, ConsumerR
 				messageSupplier = builder::build;
 			}
 
-			if (KafkaMessageDrivenChannelAdapter.this.retryTemplate != null) {
+			AttributeAccessor retryContext = ATTRIBUTES_HOLDER.get();
```
cppwfs, line 495: Same here we can use the type `RetryContext` and remove the casts below.

cppwfs (changes requested):
LGTM just some nitpicks as usual ;-)

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundEndpoint.java
artembilan: The return does not apply to the logic of Kafka Inbound Endpoints.
Therefore `void` here and just swallowing `null` return from the `RetryTemplate.execute()`.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
artembilan: We cannot do this because `ATTRIBUTES_HOLDER` not always holds exactly `RetryContext`.
Sometimes it is an `ErrorMessageUtils.getAttributeAccessor()`

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
artembilan: DITTO.
Therefore `Reolved`

cppwfs (approved):
LGTM.

snicoll:
Good stuff 👍👍

artembilan:
Thank you for feedback, @snicoll !
FYI: https://github.com/spring-projects/spring-integration/pull/10401

PR #10395, opened 2025-09-08
GH-10083: Add Nullability to core support package

cppwfs:
Related to:  #10083
* Add Nullability support to support.management package
* Add Nullability support to support.management.micrometer package
* Add Nullability support to  support.management.observation
* Add Nullability support to  support.json

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
@@ -234,7 +234,7 @@ public abstract class AbstractMessageSplitter extends AbstractReplyProducingMess
 		Map<String, Object> headers = messageHeaders;
 		Object correlationId = message.getHeaders().getId();
 		AtomicInteger sequenceNumber = new AtomicInteger(1);
-
+		Assert.state(correlationId != null, "Correlation ID must not be null.");
 		return object -> createBuilder(object, headers, correlationId, sequenceNumber.getAndIncrement(), sequenceSize);
 	}
 
@@ -282,6 +282,7 @@ public abstract class AbstractMessageSplitter extends AbstractReplyProducingMess
 				.cloneMessageHistoryIfAny();
 
 		if (this.applySequence) {
+			Assert.state(correlationId != null, "Correlation ID must not be null.");
 			builder.pushSequenceDetails(correlationId, sequenceNumber, sequenceSize);
 		}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
@@ -234,7 +234,6 @@ public abstract class AbstractMessageSplitter extends AbstractReplyProducingMess
 		Map<String, Object> headers = messageHeaders;
 		Object correlationId = message.getHeaders().getId();
 		AtomicInteger sequenceNumber = new AtomicInteger(1);
-		Assert.state(correlationId != null, "Correlation ID must not be null.");
 		return object -> createBuilder(object, headers, correlationId, sequenceNumber.getAndIncrement(), sequenceSize);
 	}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
@@ -48,6 +48,7 @@ import org.springframework.util.ObjectUtils;
  * @param <B> the target builder class type.
  *
  * @author Artem Bilan
+ * @author Glenn Renfro
  *
  * @since 6.4
  *
@@ -68,6 +69,7 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 
 	private volatile boolean modified;
 
+	@SuppressWarnings("NullAway.Init")
 	private String[] readOnlyHeaders;
 
 	protected BaseMessageBuilder(T payload, @Nullable Message<T> originalMessage) {
@@ -285,6 +287,7 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 	 * @return the current {@link BaseMessageBuilder}
 	 * @see IntegrationMessageHeaderAccessor#isReadOnly(String)
 	 */
+	@SuppressWarnings("NullAway") // Dataflow analysis limitation
 	public B readOnlyHeaders(@Nullable String... readOnlyHeaders) {
 		this.readOnlyHeaders = readOnlyHeaders != null ? Arrays.copyOf(readOnlyHeaders, readOnlyHeaders.length) : null;
 		if (readOnlyHeaders != null) {
@@ -319,7 +322,7 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 	private boolean containsReadOnly(MessageHeaders headers) {
 		if (!ObjectUtils.isEmpty(this.readOnlyHeaders)) {
 			for (String readOnly : this.readOnlyHeaders) {
-				if (headers.containsKey(readOnly)) {
+				if (readOnly != null && headers.containsKey(readOnly)) {
 					return true;
 				}
 			}
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
@@ -32,7 +32,6 @@ import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.support.ErrorMessage;
 import org.springframework.messaging.support.GenericMessage;
 import org.springframework.util.Assert;
-import org.springframework.util.ObjectUtils;
 
 /**
  * The {@link AbstractIntegrationMessageBuilder} extension for the default logic to build message.
@@ -64,12 +63,10 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 
 	private final IntegrationMessageHeaderAccessor headerAccessor;
 
-	@Nullable
-	private final Message<T> originalMessage;
+	private final @Nullable Message<T> originalMessage;
 
 	private volatile boolean modified;
 
-	@SuppressWarnings("NullAway.Init")
 	private String @Nullable [] readOnlyHeaders;
 
 	protected BaseMessageBuilder(T payload, @Nullable Message<T> originalMessage) {
@@ -287,7 +284,7 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 	 * @return the current {@link BaseMessageBuilder}
 	 * @see IntegrationMessageHeaderAccessor#isReadOnly(String)
 	 */
-	public B readOnlyHeaders(String... readOnlyHeaders) {
+	public B readOnlyHeaders(String @Nullable ... readOnlyHeaders) {
 		this.readOnlyHeaders = readOnlyHeaders != null ? Arrays.copyOf(readOnlyHeaders, readOnlyHeaders.length) : null;
 		if (readOnlyHeaders != null) {
 			this.headerAccessor.setReadOnlyHeaders(readOnlyHeaders);
@@ -319,9 +316,9 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 	}
 
 	private boolean containsReadOnly(MessageHeaders headers) {
-		if (!ObjectUtils.isEmpty(this.readOnlyHeaders)) {
+		if (this.readOnlyHeaders != null) {
 			for (String readOnly : this.readOnlyHeaders) {
-				if (readOnly != null && headers.containsKey(readOnly)) {
+				if (headers.containsKey(readOnly)) {
 					return true;
 				}
 			}
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
@@ -70,7 +70,7 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 	private volatile boolean modified;
 
 	@SuppressWarnings("NullAway.Init")
-	private String[] readOnlyHeaders;
+	private String @Nullable [] readOnlyHeaders;
 
 	protected BaseMessageBuilder(T payload, @Nullable Message<T> originalMessage) {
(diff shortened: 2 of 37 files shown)


spring-integration-core/src/main/java/org/springframework/integration/splitter/AbstractMessageSplitter.java
```diff
@@ -234,7 +234,7 @@ else if (result instanceof Publisher<?> publisher) {
 		Map<String, Object> headers = messageHeaders;
 		Object correlationId = message.getHeaders().getId();
 		AtomicInteger sequenceNumber = new AtomicInteger(1);
-
+		Assert.state(correlationId != null, "Correlation ID must not be null.");
```
artembilan: This is redundant here.
The `createBuilder()` can accept it as null.
And then e have a check at the `applySequence` condition.

spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
```diff
@@ -285,6 +287,7 @@ public B setPriority(Integer priority) {
 	 * @return the current {@link BaseMessageBuilder}
 	 * @see IntegrationMessageHeaderAccessor#isReadOnly(String)
 	 */
+	@SuppressWarnings("NullAway") // Dataflow analysis limitation
 	public B readOnlyHeaders(@Nullable String... readOnlyHeaders) {
```
artembilan: This `@Nullable` is at wrong place.
Has to be next to ellipsis.
Therefore `private String @Nullable [] readOnlyHeaders;`
Then everything else in this class should be fixed well.

spring-integration-core/src/main/java/org/springframework/integration/support/DefaultMessageBuilderFactory.java
```diff
@@ -41,6 +43,7 @@ public class DefaultMessageBuilderFactory implements MessageBuilderFactory {
 	 * and {@link org.springframework.messaging.MessageHeaders#TIMESTAMP}.
 	 * @since 4.3.2
 	 */
+	@SuppressWarnings("NullAway") // Dataflow analysis limitation
 	public void setReadOnlyHeaders(@Nullable String... readOnlyHeaders) {
```
artembilan: DITTO

spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
```diff
@@ -169,7 +172,7 @@ public void addLifecyclesToRole(String role, List<String> lifecycleBeanNames) {
 	 * Start all registered {@link SmartLifecycle}s in the role.
 	 * @param role the role.
 	 */
-	public void startLifecyclesInRole(String role) {
+	public void startLifecyclesInRole(@Nullable String role) {
```
artembilan: I think this is wrong.
This is not possible by the logic of this API.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonMessageParser.java
```diff
@@ -87,7 +87,7 @@ protected Message<?> parseWithHeaders(JsonParser parser, String jsonMessage,
 				}
 			}
 			Assert.notNull(headers, error);
-
+			Assert.state(payload != null, () -> "Payload must not be null");
```
artembilan: This class is marked for removal.
So, it is OK to not spend any time on it and just mark the whole class with `NullAway`.

spring-integration-core/src/main/java/org/springframework/integration/support/json/Jackson2JsonObjectMapper.java
```diff
@@ -140,7 +141,7 @@ else if (json instanceof Reader) {
 	}
 
 	@Override
-	protected <T> T fromJson(Object json, JavaType type) throws IOException {
+	protected <T> T fromJson(@Nullable Object json, JavaType type) throws IOException {
```
artembilan: Please, ignore all the deprecated classes with `NullAway`

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonUtils.java
```diff
@@ -132,6 +132,7 @@ private static final class AllowListTypeResolverBuilder extends ObjectMapper.Def
 
 		private final String[] trustedPackages;
 
+		@SuppressWarnings("NullAway") // Dataflow analysis limitation
 		AllowListTypeResolverBuilder(String... trustedPackages) {
```
artembilan: `String @Nullable ...` instead.

spring-integration-core/src/main/java/org/springframework/integration/support/json/JsonObjectMapper.java
```diff
@@ -46,11 +50,13 @@ default void toJson(Object value, Writer writer) throws IOException {
...
 	}
 
+	@Nullable
 	default N toJsonNode(Object value) throws IOException {
 		return null;
 	}
 
-	default <T> T fromJson(Object json, Class<T> valueType) throws IOException {
+	@Nullable
+	default <T> T fromJson(@Nullable Object json, Class<T> valueType) throws IOException {
```
artembilan: How that happened that we allow `null` over here?
Doesn't looks like Jackson accepts such a behavior:
```
    public <T> T readValue(String content, JavaType valueType) throws JacksonException
    {
        _assertNotNull("content", content);
```

Please, revise all the places which put a `null` for this API.

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/AbstractHttpRequestExecutingMessageHandler.java
```diff
@@ -346,9 +346,10 @@ protected Object getReply(ResponseEntity<?> httpResponse) {
 		MessageBuilderFactory messageBuilderFactory = getMessageBuilderFactory();
 		if (httpResponse.hasBody() && this.extractResponseBody) {
 			Object responseBody = httpResponse.getBody();
-			replyBuilder = (responseBody instanceof Message<?>)
-					? messageBuilderFactory.fromMessage((Message<?>) responseBody)
-					: messageBuilderFactory.withPayload(responseBody); // NOSONAR - hasBody()
+			Assert.state(responseBody != null, "The response body must not be null");
```
artembilan: Not correct. The `httpResponse.hasBody()` does cover this.
Please, revise. Perhaps in favor of `NullAway`

spring-integration-core/src/main/java/org/springframework/integration/support/json/JacksonJsonObjectMapper.java
```diff
@@ -132,29 +133,29 @@ else if (json instanceof Reader) {
 	}
 
 	@Override
-	protected <T> T fromJson(Object json, JavaType type) throws IOException {
+	protected <T> T fromJson(@Nullable Object json, JavaType type) throws IOException {
 		try {
-			if (json instanceof String) {
-				return this.jsonMapper.readValue((String) json, type);
+			if (json instanceof String jsonString) {
```
artembilan: Please, rebase.
Such a change is already in.

spring-integration-core/src/main/java/org/springframework/integration/support/json/AbstractJacksonJsonObjectMapper.java
```diff
@@ -96,7 +99,7 @@ protected J createJavaType(Map<String, Object> javaTypes, String javaTypeKey) {
 		}
 	}
 
-	protected abstract <T> T fromJson(Object json, J type) throws IOException;
+	protected abstract <T> T fromJson(@Nullable Object json, J type) throws IOException;
```
artembilan: Please, revise these `from` methods.
No null is allowed.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
cppwfs: On line 321 startLifecyclesInRole is used and a nullable value can be passed in.   Do we add a `Assert.state` prior to line 321?

artembilan: Let's revise that `AbstractLeaderEvent` instead!
Remove `AbstractLeaderEvent(Object source)` ctor as it is out of use.
Looks like the other ctor is never used with `null`.

spring-integration-core/src/main/java/org/springframework/integration/support/DefaultMessageBuilderFactory.java
artembilan: Missed to fix properly: `setReadOnlyHeaders(String @Nullable ... readOnlyHeaders)`

spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
```diff
@@ -315,6 +318,7 @@ private void doAddLifecyclesToRole(String role, List<String> lifecycleBeanNames)
 	@Override
 	public void onApplicationEvent(AbstractLeaderEvent event) {
 		if (event instanceof OnGrantedEvent) {
+			Assert.state(event.getRole() != null, "A role is required");
```
artembilan: That's not what I meant.
That `AbstractLeaderEvent` has to be fixed to not accept (and produce) nulls at all.
Thanks

spring-integration-core/src/main/java/org/springframework/integration/support/SmartLifecycleRoleController.java
```diff
@@ -201,7 +204,7 @@ public void startLifecyclesInRole(String role) {
 	 * Stop all registered {@link SmartLifecycle}s in the role.
 	 * @param role the role.
 	 */
-	public void stopLifecyclesInRole(String role) {
+	public void stopLifecyclesInRole(@Nullable String role) {
```
artembilan: Therefore, same here: no null allowed.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/support/DefaultMessageBuilderFactory.java
```diff
@@ -25,13 +25,15 @@
...
  * @author Artem Bilan
+ * @author Glenn Renfro
  *
  * @since 4.0
  *
  */
 public class DefaultMessageBuilderFactory implements MessageBuilderFactory {
 
-	private String[] readOnlyHeaders;
+	@SuppressWarnings("NullAway.Init")
```
artembilan: If the property is nullable, then this suppression is redundant.

spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
```diff
@@ -68,7 +69,8 @@ public abstract class BaseMessageBuilder<T, B extends BaseMessageBuilder<T, B>>
 
 	private volatile boolean modified;
 
-	private String[] readOnlyHeaders;
+	@SuppressWarnings("NullAway.Init")
```
artembilan: Remove this, please.

spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
```diff
@@ -285,7 +287,7 @@ public B setPriority(Integer priority) {
 	 * @return the current {@link BaseMessageBuilder}
 	 * @see IntegrationMessageHeaderAccessor#isReadOnly(String)
 	 */
-	public B readOnlyHeaders(@Nullable String... readOnlyHeaders) {
+	public B readOnlyHeaders(String... readOnlyHeaders) {
```
artembilan: Put `@Nullable` next to ellipsis.

spring-integration-core/src/main/java/org/springframework/integration/support/DefaultMessageBuilderFactory.java
```diff
@@ -57,21 +59,31 @@ public void addReadOnlyHeaders(String... readOnlyHeaders) {
...
+			System.arraycopy(readOnlyHeaders, 0, headers, (this.readOnlyHeaders != null) ? this.readOnlyHeaders.length : 0, readOnlyHeaders.length);
 		}
 		this.readOnlyHeaders = headers;
 	}
 
 	@Override
 	public <T> MessageBuilder<T> fromMessage(Message<T> message) {
-		return MessageBuilder.fromMessage(message)
-				.readOnlyHeaders(this.readOnlyHeaders);
+		if (this.readOnlyHeaders != null) {
```
artembilan: When you fix `BaseMessageBuilder`, then this change has to be reverted.

spring-integration-core/src/main/java/org/springframework/integration/support/DefaultMessageBuilderFactory.java
```diff
@@ -57,21 +59,31 @@ public void addReadOnlyHeaders(String... readOnlyHeaders) {
...
+		else {
+			return MessageBuilder.fromMessage(message);
+		}
 	}
 
 	@Override
 	public <T> MessageBuilder<T> withPayload(T payload) {
-		return MessageBuilder.withPayload(payload)
-				.readOnlyHeaders(this.readOnlyHeaders);
+		if (this.readOnlyHeaders != null) {
```
artembilan: DITTO

spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
```diff
@@ -319,7 +321,7 @@ public Message<T> build() {
 	private boolean containsReadOnly(MessageHeaders headers) {
 		if (!ObjectUtils.isEmpty(this.readOnlyHeaders)) {
 			for (String readOnly : this.readOnlyHeaders) {
-				if (headers.containsKey(readOnly)) {
+				if (readOnly != null && headers.containsKey(readOnly)) {
```
artembilan: This does not reflect reality.
The `String @Nullable []` is about the whole array being null.
To reflect that individual items could be null, too, we have to make it like:
```
@Nullable String @Nullable []
```
But that is not our case.
Therefore this `!= null` is not true.

spring-integration-core/src/main/java/org/springframework/integration/support/BaseMessageBuilder.java
```diff
@@ -319,7 +321,7 @@ public Message<T> build() {
 	private boolean containsReadOnly(MessageHeaders headers) {
 		if (!ObjectUtils.isEmpty(this.readOnlyHeaders)) {
```
artembilan: Let's change this to `!= null` to remove that `@SuppressWarnings("NullAway.Init")` on the property.

artembilan (changes requested):

PR #10396, opened 2025-09-08
GH-10083: Add Nullability to transaction package

cppwfs:


spring-integration-core/src/main/java/org/springframework/integration/transaction/ExpressionEvaluatingTransactionSynchronizationProcessor.java
```diff
@@ -216,7 +218,7 @@ private EvaluationContext prepareEvaluationContextToUse(Object resource) {
 			return evaluationContextWithVariables;
 		}
 		else {
-			return this.evaluationContext;
+			return this.evaluationContext != null ? this.evaluationContext : createEvaluationContext();
```
artembilan: I don't believe in this since the property is `"NullAway.Init`.
If some test fails, then that one has to be fixed calling `afterPropertiesSet()` or so.
However, the `resource` arg of this method has to be marked with the `@Nullable`.

artembilan (changes requested):

PR #10397, opened 2025-09-09
GH-10083: Add NullAway core package

cppwfs:
- Add explicit Consumer cast in AggregatorFactoryBean for lockRegistry method reference to resolve type parameter nullability mismatch
- Complete nullability annotation migration for core integration package

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationMessageHeaderAccessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationMessageHeaderAccessor.java
@@ -114,13 +114,11 @@ public class IntegrationMessageHeaderAccessor extends MessageHeaderAccessor {
 		}
 	}
 
-	@Nullable
-	public Long getExpirationDate() {
+	public @Nullable Long getExpirationDate() {
 		return getHeader(EXPIRATION_DATE, Long.class);
 	}
 
-	@Nullable
-	public Object getCorrelationId() {
+	public @Nullable Object getCorrelationId() {
 		return getHeader(CORRELATION_ID);
 	}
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/IntegrationMessageHeaderAccessor.java
```diff
@@ -182,8 +176,7 @@ public AtomicInteger getDeliveryAttempt() {
 	 * @since 5.1.6
 	 */
 	@SuppressWarnings("unchecked")
-	@Nullable
-	public <T> T getSourceData() {
+	public @Nullable <T> T getSourceData() {
```
artembilan: I believe this has to be next to the `T`, the exact return type, not generic definition

spring-integration-core/src/main/java/org/springframework/integration/IntegrationMessageHeaderAccessor.java
```diff
@@ -192,14 +185,12 @@ public <T> T getSourceData() {
...
-	@Nullable
-	public ContextView getReactorContext() {
+	public @Nullable ContextView getReactorContext() {
 		return getHeader(REACTOR_CONTEXT, ContextView.class);
 	}
 
 	@SuppressWarnings("unchecked")
-	@Nullable
-	public <T> T getHeader(String key, Class<T> type) {
+	public @Nullable <T> T getHeader(String key, Class<T> type) {
```
artembilan: DITTO

artembilan (changes requested):
Glad to see we are close ! 👌

PR #10399, opened 2025-09-10
GH-10083: Add Nullability to support.management

cppwfs:

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
@@ -255,6 +255,7 @@ public class RouterSpec<K extends @Nullable Object, R extends AbstractMappingMes
 		}
 
 		@Override
+		@SuppressWarnings("NullAway") // Performance
 		protected void onInit() {
 			super.onInit();
 			ConversionService conversionService = getConversionService();
@@ -274,7 +275,6 @@ public class RouterSpec<K extends @Nullable Object, R extends AbstractMappingMes
 					throw new MessagingException("Unsupported channel mapping type for router ["
 							+ key.getClass() + "]");
 				}
-				Assert.notNull(channelKey, "'channelKey' must not be null");
 				this.router.setChannelMapping(channelKey, entry.getValue().getComponentName());
 			}
 		}
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
@@ -40,6 +40,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Glenn Renfro
  *
  * @since 5.0
  */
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/management/ControlBusCommandRegistry.java
```diff
@@ -300,7 +302,7 @@ public static final class CommandMethod {
 
 		private final Class<?>[] parameterTypes;
 
-		private String description;
+		private @Nullable String description;
```
artembilan: I would prefer `NullAway.Init`.
See the logic in the `populateDescriptionIntoCommand()`.
So, whenever this `CommandMethod` is initialized after `populateCommandMethod()`, the `description` is not null.

spring-integration-http/src/main/java/org/springframework/integration/http/management/ControlBusController.java
```diff
@@ -132,7 +133,7 @@ public record ControlBusBean(String beanName, List<ControlBusCommand> commands)
 
 	}
 
-	public record ControlBusCommand(String command, String description, List<Class<?>> parameterTypes) {
+	public record ControlBusCommand(String command, @Nullable String description, List<Class<?>> parameterTypes) {
```
artembilan: Therefore the change in this class will go away.

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
```diff
@@ -273,7 +274,7 @@ else if (conversionService.canConvert(key.getClass(), String.class)) {
 					throw new MessagingException("Unsupported channel mapping type for router ["
 							+ key.getClass() + "]");
 				}
-
+				Assert.notNull(channelKey, "'channelKey' must not be null");
```
artembilan: I think this is a false impression.
See the logic:
```
				else if (conversionService.canConvert(key.getClass(), String.class)) {
					channelKey = conversionService.convert(key, String.class);
				}
```

I believe, according to that `canConvert()` Javadocs, the `convert` result would never be `null`.
So, we probably need `@SuppressWarnings("NullAway")` instead.

spring-integration-http/src/main/java/org/springframework/integration/http/management/ControlBusController.java
```diff
@@ -43,6 +43,7 @@
  * The REST Controller to provide the management API for Control Bus pattern.
  *
  * @author Artem Bilan
+ * @author Glenn Renfro
```
artembilan: Sorry, my friend, but looks like this is only change in this class 🥶

artembilan (changes requested):

spring-integration-http/src/main/java/org/springframework/integration/http/management/ControlBusController.java
cppwfs: I debated which direction on both instances.   There was a remote path where they could be null.   So I chose the code path vs the `SuppressWarning`.  But if it has no chance at null, then we are good to `Suppress`.   And for performance for sure.

spring-integration-core/src/main/java/org/springframework/integration/dsl/RouterSpec.java
artembilan: Yeah... This is a critical path execution for the router logic, so I'd rather bite a bullet and suppress than extra code.

spring-integration-http/src/main/java/org/springframework/integration/http/management/ControlBusController.java
cppwfs: Doh!  But I wanted my name here 😄

PR #10401, opened 2025-09-11
GH-10345: Migrate to Spring Core Retry

### LINKED ISSUE #10345: Replace the use of Spring Retry with Core Retry

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10345

* Remove `spring-retry` dependency since all its usage have been replaced with Spring Core Retry and some custom internal implementations
* Rework the `RequestHandlerRetryAdvice` to be based on the Spring Core Retry:
 - accept only `RetryPolicy` and use `RetryTemplate` internally just for stateless behavior
 - Expose `stateKeyFunction`, `newMessagePredicate` and `stateCacheSize` for stateful behavior.
 - Since there is no any stateful utility in Spring Core, implement its algorithm internally in the `RequestHandlerRetryAdvice`
* Remove `RetryStateGenerator` and `SpelExpressionRetryStateGenerator` since they are based on `RetryState` from Spring Retry. And now they are replaced by the mentioned functions above as options on the `RequestHandlerRetryAdvice`
* Rework `RetryAdviceParser` according to a new state of the `RequestHandlerRetryAdvice`. In addition, expose an `exponential-back-off/jitter` XML attribute
* Fix `ErrorMessageStrategy` Javadoc to point to a general `AttributeAccessor` interface from Spring Core
* Remove `ErrorMessageSendingRecoverer.RetryExceptionNotAvailableException` and the logic around it since this is not a case anymore with a new `RequestHandlerRetryAdvice` logic
* Rework failing tests according to the new API and logic of the target classes
* Document breaking change in the `whats-new.adoc`
* Rework `retry` section in the `handler-advice/classes.adoc` to reflect new reality for the `RequestHandlerRetryAdvice`
* Rework the configuration samples in the `retry` section to Java


spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
```diff
@@ -45,94 +65,244 @@
...
+						interruptedException);
+				retryState.exceptions.forEach(retryException::addSuppressed);
+				this.retryListener.onRetryPolicyInterruption(this.retryPolicy, retryCallback, retryException);
+				throw retryException;
+			}
+
+			try {
+				this.retryListener.beforeRetry(this.retryPolicy, retryCallback);
+				Object result = retryCallback.execute();
+				this.retryStateCache.remove(retryState);
```
cppwfs: should this be retryStateKey?

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
```diff
@@ -16,27 +16,47 @@
...
+import org.springframework.util.backoff.BackOffExecution;
 
 /**
- * Uses spring-retry to perform stateless or stateful retry.
- * Stateless retry means the retries are performed internally
- * by the {@link RetryTemplate}; stateful retry means the
- * exception is thrown but state is maintained to support
- * the retry policies. Stateful retry requires a
- * {@link RetryStateGenerator}.
+ * The {@link AbstractRequestHandlerAdvice} implementation for retrying the target
```
cppwfs, line 45: Replace `The` with `Is an`

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
```diff
@@ -16,27 +16,47 @@
...
+ * {@link org.springframework.messaging.MessageHandler} execution.
+ * <p>
+ * By default, this advice performs 3 attempts (plus initial execution) with no delay in between.
+ * <p>
+ * If {@link #setStateKeyFunction(Function)} is provided, the retry logic is turned into a stateful algorithm.
+ * In this case, a {@link BackOffExecution} is cached by the mentioned retry state key if exception is retryable.
+ * Then this exception is thrown back to the caller, e.g.,
```
cppwfs, line 52: Replace `Then this`  with `Then an`

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
```diff
@@ -45,94 +65,244 @@
...
+	 * @param stateCacheSize the size of the cache.
+	 */
+	public void setStateCacheSize(int stateCacheSize) {
+		this.stateCacheSize = stateCacheSize;
+	}
+
+	/**
+	 * Set a {@link Predicate} to determine if a fresh stateful retry state should be started for this request message.
+	 * {@code false} by default to not refresh existing stateful retry states.
+	 * Makes sense only when {@link #setStateKeyFunction(Function)} is provided.
```
cppwfs: Replace `Makes sense` with `Use`

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
```diff
@@ -45,94 +65,244 @@
...
+				if (this.retryPolicy.shouldRetry(ex)) {
+					retryState.exceptions.add(ex);
+					// Short-circuit at the end of the current attempt to avoid extra external future calls
+					if (BackOffExecution.STOP != retryState.nextBackOff()) {
+						throw ex;
+					}
+				}
+				else {
+					this.retryStateCache.remove(retryStateKey);
+					throw ex;
```
cppwfs, line 239: Do we want to mention that retry attempts were exhausted?

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -21,6 +21,42 @@ Junit 4-Based Support Components are deprecated.
...
 The project now leverages https://jspecify.dev/docs/start-here/[JSpecify] annotations to expose null-safe APIs and to check the consistency of those nullability declarations with https://github.com/uber/NullAway[NullAway] as part of its build.
 
+The `spring-retry` dependency, together with all its API usage have been replaced by retry API in Spring Framework Core module.
+This breaking change is the natural evolution of the whole Spring portfolio.
+In general, the following references have been migrated:
+
+- `org.springframework.retry.support.RetryTemplate` -> `org.springframework.core.retry.RetryTemplate`;
+- `org.springframework.retry.RetryPolicy` -> `org.springframework.core.retry.RetryPolicy`;
+- `org.springframework.retry.RecoveryCallback` -> `org.springframework.integration.core.RecoveryCallback`.
+There is no `RecoveryCallback` abstraction in the Spring Framework, since regular `try..catch` on the `RetryException` is enough.
```
cppwfs, line 31: Lines 31 and 32 can be consolidated to:
"There is no `RecoveryCallback` abstraction in the Spring Framework, however for Spring Integration, the `RecoveryCallback` makes sense as a dead-letter publisher to an error channel;"

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -21,6 +21,42 @@ Junit 4-Based Support Components are deprecated.
...
+- `KafkaInboundGatewaySpec`
+- `KafkaMessageDrivenChannelAdapterSpec`
+- `KafkaInboundEndpoint`
+- `KafkaInboundGateway`
+- `KafkaMessageDrivenChannelAdapter`
+
+The `RequestHandlerRetryAdvice` was rebuilt to avoid external API as much as possible.
+For the stateless retry logic, there is just enough to provide a `org.springframework.core.retry.RetryPolicy`.
+The stateful retry logic is activated by the `Function<Message<?>, Object> stateKeyFunction`.
+The used before `RetryStateGenerator` abstraction and its `SpelExpressionRetryStateGenerator` implementation have been removed due to dependency on the `spring-retry` API.
```
cppwfs: Replace `The used before` with `The`

cppwfs (changes requested):
Just my usual nitpicks :-)

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
artembilan: Sorry, Glenn, the sentence like:

```
Is an {@link AbstractRequestHandlerAdvice} implementation for retrying the target...
```

Doesn't sound in my head. 
I'm not English speaker, but even `An {@link AbstractRequestHandlerAdvice} implementation` doesn't sound right for me as a beginning of this text.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
artembilan: Well, technically I wanted to connect this sentence with the previous one where we mention some exception.
Perhaps `Then that exception` would be better?
Although `this` is something what we use in our language to make such a connection.

spring-integration-core/src/main/java/org/springframework/integration/handler/advice/RequestHandlerRetryAdvice.java
artembilan: No, this is not the place.
If we are in this `else` branch, then this exception is not retryable, therefore the retry exhausting does not make sense and we simply re-throw this exception.

The retry is exhausted when `delay` is `BackOffExecution.STOP`.
And the crucial part of the algorithm in that `if (this.retryPolicy.shouldRetry(ex)) {` block.
The first `if (BackOffExecution.STOP != retryState.currentDelay) {` is just for a first retry attempt after an original call.
The next time we check for `BackOffExecution.STOP` after each service failure.
That is what I meant with that `Short-circuit`: no need to wait for another external call to this retry advice to figure out that we cannot retry any more.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
artembilan: How about the ``try..catch` power I've tried to emphasize in the first sentence?

artembilan:
Thank you @cppwfs , for review!

You know, I want you to pay attention to the cache and its impact when no next redelivery for a message.
With a stateful retry we have a hard dependency on the external system to bring the message back to us.
While I made the cache as remove old entries when overcharged it still may lead to some unnecessary memory consumption for all the stale entries in the cache.
I could not figure out a smooth path how to manage that cache.
The cache key is based on a message, and if we would like to remove entry because of external system decision, we might not have enough info about the message to extract the key.
Perhaps we could enrich the `RetryState` with a message as well and just expose the whole cache as getter to let end-user to manipulate with it whatever way they want.
An original message in that state object would have all the required info to determine how old, what content and sometimes even from where it was originated.
We may consider this as a further stateful retry algorithm evolution, but it bothers me a bit right now, so, I want to share a pain with you as some food to think about before merging.

cppwfs:
Here are a couple of options:
Following up on your idea: Giving the user access to the cache for them to manage does give them the freedom to decide how to make their implementation more performant.      It would require documentation to let them know it is at their own risk, because we are giving them a mutable object that can now change the state.    

Another option is that we can allow them to give us a function that extracts properties from the message and we store the properties in the RetryState.   Then offer a cleanup method that they pass a function that returns true or false on whether entries should be removed and the function uses the properties in the RetryState to determine if a retry needs to be removed.

cppwfs (approved):

cppwfs:
The PR looks good.   Do you want to continue the discussion before I merge?

artembilan:
Thanks, Glenn, for feedback.
Let's merge this and create an issue for the future.
Perhaps someone would share more ideas.
Spring Retry provides a `RetryContextCache` abstraction which is also not very useful with that `RetryContext` info exposed...

cppwfs:
LGTM.  Thank you for the great work!

PR #10419, opened 2025-09-15
Update Javadoc references deprecated javadocs

cppwfs:
Replace deprecated `org.junit.Assert#assertThat` references with `org.hamcrest.MatcherAssert#assertThat` in test matcher documentation.

This change updates documentation in HeaderMatcher and PayloadMatcher classes to reference the correct assertThat method location after JUnit Jupiter migration removed Assert.assertThat in favor of the Hamcrest implementation.

This is an optional PR.  Wanted to remove the last of the deprecated markers in SI.

### DIFF
--- a/spring-integration-test-support/src/main/java/org/springframework/integration/test/matcher/HeaderMatcher.java
+++ b/spring-integration-test-support/src/main/java/org/springframework/integration/test/matcher/HeaderMatcher.java
@@ -31,7 +31,7 @@ import org.springframework.messaging.MessageHeaders;
  * Are the {@link MessageHeaders} of a {@link Message} containing any entry
  * or multiple that match?
  * <p>
- * For example using {@link org.junit.Assert#assertThat(Object, Matcher)} for a single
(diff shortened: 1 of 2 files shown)


spring-integration-test-support/src/main/java/org/springframework/integration/test/matcher/HeaderMatcher.java
```diff
@@ -31,7 +31,7 @@
  * Are the {@link MessageHeaders} of a {@link Message} containing any entry
  * or multiple that match?
  * <p>
- * For example using {@link org.junit.Assert#assertThat(Object, Matcher)} for a single
+ * For example using {@link org.hamcrest.MatcherAssert#assertThat(Object, Matcher)} for a single
```
artembilan, line 34: Good catch!

I wonder if that is worth to mention in these Javadocs a `HamcrestCondition`, too...

artembilan (changes requested):

PR #10420, opened 2025-09-15
GH-10162: The full file path for `PersistentAcceptOnceFileListFilter`

### LINKED ISSUE #10162: The `AbstractPersistentAcceptOnceFileListFilter` does not deal with full file name

6.5.0

**Describe the bug**

SftpClient.DirEntry do not hold the full path of the file.
ScanDirEntry does.

**To Reproduce**

N/A

**Expected behavior**

N/A

**Sample**

N/A

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10162

The `AbstractPersistentAcceptOnceFileListFilter` may cause a problem with entries in the `MetadataStore` when the same file name is used from different directories. Let's imagine a scenario when remote directories are created every day with respective timestamp, e.g. today `20250828`, tomorrow `20250829`.
 The files in those directories are based on customer names: `customer1.xml`, `customer2.xml` etc.
Since most of the `AbstractPersistentAcceptOnceFileListFilter` implementations do just something like this: `return file.getName();` that would lead to the same key for the `MetadataStore` operations.

* Fix `SmbPersistentAcceptOnceFileListFilter` to use `SmbFile.getUncPath()` as a notation of the full path file
* Fix `SftpPersistentAcceptOnceFileListFilter` to use `SftpClient.DirEntry.getLongFilename()` if it is not empty.
* Fix `SftpSession.list()` to populated the mentioned `longFileName` into a new `SftpClient.DirEntry` instance based on just fetched
* Introduce an `EnhancedFTPFile` adapter to delegate to the original `FTPFile` and populate an additional `longFileName` property
* Check for this new class in the `FtpPersistentAcceptOnceFileListFilter` to extract its `longFileName` property

### DIFF
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
@@ -19,8 +19,6 @@ package org.springframework.integration.ftp.session;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.logging.Log;
@@ -75,16 +73,16 @@ public class FtpSession implements Session<FTPFile> {
 	@Override
 	public FTPFile[] list(@Nullable String path) throws IOException {
 		FTPFile[] ftpFiles = this.client.listFiles(path);
-		if (StringUtils.hasText(path)) {
-			List<EnhancedFTPFile> enhancedFtpFiles = new LinkedList<>();
-			for (FTPFile ftpFile : ftpFiles) {
-				EnhancedFTPFile enhancedFTPFile = new EnhancedFTPFile(ftpFile);
-				enhancedFTPFile.setLongFileName(path + '/' + ftpFile.getName());
-				enhancedFtpFiles.add(enhancedFTPFile);
-			}
-			return enhancedFtpFiles.toArray(new FTPFile[0]);
+		String remoteDir = StringUtils.hasText(path) ? path : ".";
+		EnhancedFTPFile[] enhancedFtpFiles = new EnhancedFTPFile[ftpFiles.length];
+		for (int i = 0; i < ftpFiles.length; i++) {
+			FTPFile ftpFile = ftpFiles[i];
+			EnhancedFTPFile enhancedFTPFile = new EnhancedFTPFile(ftpFile);
+			enhancedFTPFile.setLongFileName(remoteDir + '/' + ftpFile.getName());
+			enhancedFtpFiles[i] = enhancedFTPFile;
 		}
-		return ftpFiles;
+
+		return enhancedFtpFiles;
 	}
 
 	@Override
--- a/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
+++ b/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
@@ -19,6 +19,8 @@ package org.springframework.integration.ftp.session;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.LinkedList;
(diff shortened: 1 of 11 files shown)


spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
```diff
@@ -71,7 +74,17 @@ public boolean remove(String path) throws IOException {
 
 	@Override
 	public FTPFile[] list(@Nullable String path) throws IOException {
-		return this.client.listFiles(path);
+		FTPFile[] ftpFiles = this.client.listFiles(path);
+		if (StringUtils.hasText(path)) {
+			List<EnhancedFTPFile> enhancedFtpFiles = new LinkedList<>();
+			for (FTPFile ftpFile : ftpFiles) {
+				EnhancedFTPFile enhancedFTPFile = new EnhancedFTPFile(ftpFile);
+				enhancedFTPFile.setLongFileName(path + '/' + ftpFile.getName());
```
cppwfs: If path is the Current Working Directory of "" wouldn't we still want to go through the flow of adding the file name?

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
```diff
@@ -71,7 +74,17 @@ public boolean remove(String path) throws IOException {
 
 	@Override
 	public FTPFile[] list(@Nullable String path) throws IOException {
-		return this.client.listFiles(path);
+		FTPFile[] ftpFiles = this.client.listFiles(path);
+		if (StringUtils.hasText(path)) {
+			List<EnhancedFTPFile> enhancedFtpFiles = new LinkedList<>();
```
cppwfs: Optimization: Since we know the length of the array from line 77 couldn't we allocate an array here and remove the use of the linked list.   Nitpick.

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/inbound/FtpInboundRemoteFileSystemSynchronizerTests.java
```diff
@@ -97,7 +98,7 @@ public void testCopyFileToLocalDir() throws Exception {
 		store.setBaseDirectory("test");
 		store.afterPropertiesSet();
 		FtpPersistentAcceptOnceFileListFilter persistFilter =
-				new FtpPersistentAcceptOnceFileListFilter(store, "foo");
+				new FtpPersistentAcceptOnceFileListFilter(store, "someKey/");
```
cppwfs, line 101: Bye bye `foo`. 😄

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -100,6 +100,11 @@ The `AbstractInboundFileSynchronizer` now caches a filtered result of the `Sessi
 So, later synchronizations deal with the cache only by the `maxFetchSize` until the cache is exhausted.
 See xref:sftp/max-fetch.adoc[] for more information.
 
+All the `AbstractPersistentAcceptOnceFileListFilter` implementations now use a "long file name" for the metadata entry key.
+Perviously, just a file name may cause the metadata overriding problem when the same filter is used for different directories with same file names.
```
cppwfs: Replace `Perviously` with `Previously`

cppwfs (changes requested):
Looks great!  Just my usual nitpicks!

spring-integration-ftp/src/main/java/org/springframework/integration/ftp/session/FtpSession.java
artembilan: Do you mean still set `longName` but without a dir prefix?

cppwfs: Yes.   
I think it would be consistent.   The path still exists, its just... well... empty.

cppwfs:
LGTM.   Thanks for working on this.  It's awesome!

PR #10430, opened 2025-09-16
Update dokka version

cppwfs:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -177,7 +177,7 @@ allprojects {
 
     }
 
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
This is really cool!
Yes, I probably had a problem with `V2` plugin configuration.
However, might your's `noJdkLink` is the trick anyway.

Can you take a look, please, into removing that ` // TODO until https://github.com/Kotlin/dokka/issues/3472` part since we are already here and definitely use a newer plugin version?
Might be there is no problem with Jackson anymore?

Thank you!

PR #10445, opened 2025-09-24
GH-10090: Add channel adapters for AMQP 1.0

artembilan:
Related to: https://github.com/spring-projects/spring-integration/issues/10090

Spring AMQP now provides the `org.springframework.amqp:spring-rabbitmq-client` library to communicate to RabbitMQ with AMQP 1.0 protocol.

* Replace `spring-rabbit` with the `spring-rabbitmq-client` since the latter brings the AMQP 0.9 protocol library as transitive dependency
* Introduce an `AmqpClientMessageHandler` based on the `AsyncAmqpTemplate` where its implementation should be for AMQP 1.0, e.g. `org.springframework.amqp.rabbitmq.client.RabbitAmqpTemplate`
* This outbound channel adapter can behave as a gateway when `setRequiresReply(true)`
* Cover basic `AmqpClientMessageHandler` with integration tests
* Document this new feature

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
@@ -1,0 +1,317 @@
+/*
+ * Copyright 2025-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.amqp.outbound;
+
+import java.util.concurrent.CompletableFuture;
+
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.amqp.core.AsyncAmqpTemplate;
+import org.springframework.amqp.core.Message;
+import org.springframework.amqp.core.MessageProperties;
+import org.springframework.amqp.support.converter.MessageConverter;
+import org.springframework.amqp.support.converter.SimpleMessageConverter;
+import org.springframework.amqp.support.converter.SmartMessageConverter;
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.core.ResolvableType;
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.support.StandardEvaluationContext;
+import org.springframework.integration.amqp.support.AmqpHeaderMapper;
+import org.springframework.integration.amqp.support.DefaultAmqpHeaderMapper;
+import org.springframework.integration.expression.ExpressionUtils;
+import org.springframework.integration.expression.ValueExpression;
+import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.integration.mapping.AbstractHeaderMapper;
+import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * An {@link AbstractReplyProducingMessageHandler} implementation for AMQP 1.0 client.
+ * <p>
+ * With the {@link #setRequiresReply(boolean)} configured as {@code true}, this message handler
+ * behaves as a gateway - the RPC over AMQP.
+ * In this case, the {@link AsyncAmqpTemplate} must be able to convert
+ * a reply into a provided {@link #replyPayloadTypeExpression}.
+ * <p>
+ * This handler is {@code async} by default.
+ * <p>
+ * In async mode, the error is sent to the error channel even if not in a gateway mode.
+ * <p>
+ * The {@link #exchangeExpression}, {@link #routingKeyExpression} and {@link #queueExpression}
+ * are optional.
+ * Then they have to be supplied by the {@link AsyncAmqpTemplate}.
+ *
+ * @author Artem Bilan
+ *
+ * @since 7.0
+ */
+public class AmqpClientMessageHandler extends AbstractReplyProducingMessageHandler {
+
+	private final AsyncAmqpTemplate amqpTemplate;
+
+	private AmqpHeaderMapper headerMapper = DefaultAmqpHeaderMapper.outboundMapper();
+
+	private MessageConverter messageConverter = new SimpleMessageConverter();
+
+	private @Nullable Expression exchangeExpression;
+
+	private @Nullable Expression routingKeyExpression;
+
+	private @Nullable Expression queueExpression;
+
+	private @Nullable Expression replyPayloadTypeExpression;
+
+	private boolean returnMessage;
+
+	@SuppressWarnings("NullAway.Init")
+	private StandardEvaluationContext evaluationContext;
+
+	/**
+	 * Construct an instance with the provided {@link AsyncAmqpTemplate}.
+	 * The {@link AsyncAmqpTemplate} must be an implementation for AMQP 1.0 protocol,
+	 * e.g. {@link org.springframework.amqp.rabbitmq.client.RabbitAmqpTemplate}.
+	 * @param amqpTemplate the {@link AsyncAmqpTemplate} to use.
+	 */
+	@SuppressWarnings("this-escape")
+	public AmqpClientMessageHandler(AsyncAmqpTemplate amqpTemplate) {
+		this.amqpTemplate = amqpTemplate;
+		setAsync(true);
+	}
+
+	public void setHeaderMapper(AmqpHeaderMapper headerMapper) {
+		this.headerMapper = headerMapper;
+	}
+
+	public void setMessageConverter(MessageConverter messageConverter) {
+		this.messageConverter = messageConverter;
+	}
+
+	public void setExchange(String exchange) {
+		setExchangeExpression(new ValueExpression<>(exchange));
+	}
+
+	public void setExchangeExpressionString(String exchangeExpression) {
+		setExchangeExpression(EXPRESSION_PARSER.parseExpression(exchangeExpression));
+	}
+
+	public void setExchangeExpression(Expression exchangeExpression) {
+		this.exchangeExpression = exchangeExpression;
+	}
+
+	public void setRoutingKey(String routingKey) {
+		setRoutingKeyExpression(new ValueExpression<>(routingKey));
+	}
+
+	public void setRoutingKeyExpressionString(String routingKeyExpression) {
+		setRoutingKeyExpression(EXPRESSION_PARSER.parseExpression(routingKeyExpression));
+	}
+
+	public void setRoutingKeyExpression(Expression routingKeyExpression) {
+		this.routingKeyExpression = routingKeyExpression;
+	}
+
+	public void setQueue(String queue) {
+		setQueueExpression(new ValueExpression<>(queue));
+	}
+
+	public void setQueueExpressionString(String queueExpression) {
+		setQueueExpression(EXPRESSION_PARSER.parseExpression(queueExpression));
+	}
+
+	public void setQueueExpression(Expression queueExpression) {
+		this.queueExpression = queueExpression;
+	}
+
+	/**
+	 * Set the reply payload type.
+	 * Used only if {@link #setRequiresReply(boolean)} is {@code true}.
+	 * @param replyPayloadType the reply payload type.
+	 */
+	public void setReplyPayloadType(Class<?> replyPayloadType) {
+		setReplyPayloadType(ResolvableType.forClass(replyPayloadType));
+	}
+
+	/**
+	 * Set the reply payload type.
+	 * Used only if {@link #setRequiresReply(boolean)} is {@code true}.
+	 * @param replyPayloadType the reply payload type.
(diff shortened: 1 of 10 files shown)


artembilan:
This just an initial PoC and provides only an outbound channel adapter implementation.
Can be merged as is, or I can go ahead and push more for other features after review.

I anticipate these features yet:

* `AmqpClientMessageProducer` - an event-driven channel adapter for receiving AMQP 1.0 messages. Going to be based on the `RabbitAmqpListenerContainer`
* `AmqpClientInboundGateway` - a `MessagingGatewaySupport` implementation for receiving AMQP 1.0 messages and replying, respectively. Going to be based on the `RabbitAmqpListenerContainer`
* `AmqpClientMessageSource` - a source polling channel adapter for polling AMQP 1.0 messages from a queue. Going to be based on the `AsyncAmqpTemplate.receive()` API
* Java DSL for all new components
* More docs
* What's New entry.

Thanks

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
```diff
@@ -0,0 +1,317 @@
...
+ * In this case, the {@link AsyncAmqpTemplate} must be able to convert
+ * a reply into a provided {@link #replyPayloadTypeExpression}.
+ * <p>
+ * This handler is {@code async} by default.
+ * <p>
+ * In async mode, the error is sent to the error channel even if not in a gateway mode.
+ * <p>
+ * The {@link #exchangeExpression}, {@link #routingKeyExpression} and {@link #queueExpression}
+ * are optional.
+ * Then they have to be supplied by the {@link AsyncAmqpTemplate}.
```
cppwfs: Replace "Then they` with `In this case they have to be supplied..."

src/reference/antora/modules/ROOT/pages/amqp/amqp-1.0.adoc
```diff
@@ -0,0 +1,55 @@
+[[amqp-1.0]]
+= AMQP 1.0 Support
+
+Starting with version 7.0, Spring Integration provides channel adapters for RabbitMQ AMQP 1.0 support.
+These channel adapters are based on the `org.springframework.amqp:spring-rabbitmq-client` library.
+
+The Spring AMQP documentation talks about https://docs.spring.io/spring-amqp/reference/4.0/rabbitmq-amqp-client.html[RabbitMQ AMQP 1.0 support] in more detail.
```
cppwfs: Rephrase line 7 with "For more details on RabbitMQ AMQP 1.0, see the https://docs.spring.io/spring-amqp/reference/4.0/rabbitmq-amqp-client.html[Spring AMQP documentation]."

src/reference/antora/modules/ROOT/pages/amqp/amqp-1.0.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+[[amqp-1.0-outbound]]
+== AMQP 1.0 Outbound Channel Adapters
+
+The `AmqpClientMessageHandler` is an `AbstractReplyProducingMessageHandler` implementation and can act as a one-way channel adapter or as an outbound gateway depending on the `setRequiresReply()` configuration.
+The instance of this channel adapter requires an `AsyncAmqpTemplate` implementation for AMQP 1.0 protocol, e.g. `RabbitAmqpTemplate` from the mentioned above `spring-rabbitmq-client` library.
+This message handler is asynchronous by default; therefore, publication errors should be handled via `errorChannel` header in the request message or global default `errorChannel` in the application context.
+
+The `exchange` to publish message (together with optional `routingKey`) is mutually exclusive with a `queue` to publish.
+If neith is provided, then `AsyncAmqpTemplate` implementation must ensure some defaults for those destination parts; otherwise the message is going to be rejected as not delivered.
```
cppwfs: `neith` -> `neither`

src/reference/antora/modules/ROOT/pages/amqp/amqp-1.0.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+A default `MessageConverter` is a `org.springframework.amqp.support.converter.SimpleMessageConverter` that can work with `String`, `Serializable` instances, or byte arrays.
```
cppwfs: Maybe this can be reworded as "By default, the `MessageConverter` is a `org.springframework.amqp.support.converter.SimpleMessageConverter` that handles `String`, `Serializable` instances, and byte arrays."

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
```diff
@@ -0,0 +1,317 @@
...
+
+	public void setRoutingKey(String routingKey) {
+		setRoutingKeyExpression(new ValueExpression<>(routingKey));
+	}
+
+	public void setRoutingKeyExpressionString(String routingKeyExpression) {
+		setRoutingKeyExpression(EXPRESSION_PARSER.parseExpression(routingKeyExpression));
+	}
+
+	public void setRoutingKeyExpression(Expression routingKeyExpression) {
```
cppwfs, line 124: `routingExpression`  and `queueExpression` are  nullable.   Can a user set it to null?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
```diff
@@ -0,0 +1,317 @@
...
+		this.returnMessage = returnMessage;
+	}
+
+	@Override
+	public String getComponentType() {
+		return getRequiresReply() ? "amqp:outbound-gateway" : "amqp:outbound-channel-adapter";
+	}
+
+	@Override
+	protected void doInit() {
```
cppwfs, line 193: Do you think we may want to have a builder, since we have these type of rules?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
```diff
@@ -0,0 +1,317 @@
...
+		super.doInit();
+		if (this.headerMapper instanceof AbstractHeaderMapper<?> abstractHeaderMapper) {
+			abstractHeaderMapper.setBeanClassLoader(getBeanClassLoader());
+		}
+		this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
+
+		Assert.state(this.exchangeExpression == null || this.queueExpression == null,
```
cppwfs, line 200: Should we use `hasText` here instead of checking for null?

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
```diff
@@ -0,0 +1,317 @@
...
+		else {
+			sendResultFuture.join();
+		}
+	}
+
+	private Object doSendAndReceive(org.springframework.messaging.Message<?> requestMessage, Message amqpMessage,
+			@Nullable String queue, @Nullable String exchange, @Nullable String routingKey) {
+
+		ParameterizedTypeReference<?> replyType = null;
+		if (this.replyPayloadTypeExpression != null) {
```
cppwfs, line 271: Should we use `hasText` instead of null?

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandlerTests.java
```diff
@@ -0,0 +1,217 @@
...
+	MessageChannel amqpClientSendChannel;
+
+	@Autowired
+	MessageChannel amqpClientSendAndReceiveChannel;
+
+	@Autowired
+	AmqpClientMessageHandler amqpClientGateway;
+
+	@Test
+	void noExchangeOrQueue() {
```
cppwfs: should this be renamed `bothExchangeAndQueue`?

cppwfs (changes requested):
Looks great!   
My usual nitpicks :-)

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
artembilan: No, if you chose to call these methods, then no null allowed.

cppwfs: Understood.   Thank you!

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
artembilan: That's interesting idea, but too hard to implement. 
Those different choice would lead to different types and that is really hard to follow.
Much easier to express rules with exceptions like we do in any other places.
Plus docs should be good.

I'll think about your suggestion in DSL, but no promises.
What stops me is really a support of many classes which would lead from one builder choice or another to avoid paradox of choice.
More thinking, more codding, and eventually more support on our side. 😢 

But thanks for an idea anyway!

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
artembilan: Not sure what you mean.
The `this.exchangeExpression` is an `Expression`, not `String`

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
artembilan: DITTO

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandlerTests.java
artembilan: Sorry, but no. This test is about failing when neither is provided: nor on the channel adapter, neither on `RabbitAmqpTemplate`.

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
cppwfs: No worries at all...   
Just was thinking through my fingers :-)

spring-integration-amqp/src/main/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandler.java
cppwfs: DOH!    I was thinking of a string .  Apologies.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpClientMessageHandlerTests.java
cppwfs: I should say...  `neitherExchangeAndQueue`

src/reference/antora/modules/ROOT/pages/amqp/amqp-1.0.adoc
artembilan: Right. I'm not the best guy to judge English sentences, but what I know that technical, official documentation must be impersonated.
Therefore saying something like a command `see` is not good.
I know that we have a plethora of those `see`, `you`, `they`, `we` etc. throughout the doc, but that does not mean that we should introduce more.
WDYT?

cppwfs: Sounds good.   We can hold off this change.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/inbound/AmqpClientMessageProducerTests.java
```diff
@@ -0,0 +1,180 @@
...
+				.extracting(Message::getPayload)
+				.isEqualTo("test data #2");
+
+		SimpleAcknowledgment acknowledgment = StaticMessageHeaderAccessor.getAcknowledgment(receive);
+		assertThat(acknowledgment).isNotNull();
+		acknowledgment.acknowledge();
+	}
+
+	@Test
+	void receiveBatch() {
```
cppwfs, line 100: Do we want to have a receiveBatchAndAck test as well?

cppwfs (approved):
Changes to previous commit look great!
Addition of the AmqpClientMessageProducer look good as well.  Just one nitpick for an additional test.   

This some cool stuff!   Thanks for working on it!

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/inbound/AmqpClientMessageProducerTests.java
artembilan: Sure!
Will look into that.
I think more about errors.
But let me fix `main` first!

cppwfs: Sweet!   Thanks!

artembilan:
@cppwfs ,

let me know if you are going to merge this after your next review, or is that OK for me to continue piling related features into this PR.
Thanks

cppwfs (approved):
LGTM 
This is great stuff!
Thanks for adding the batch test!

PR #10459, opened 2025-10-02
GH-6375: Harmonize package structures

### LINKED ISSUE #6375: Standardize Package Structure across Modules [INT-2389]
**[Gunnar Hillert](https://jira.spring.io/secure/ViewProfile.jspa?name=hillert)** opened **[INT-2389](https://jira.spring.io/browse/INT-2389?redirect=false)** and commented

Going through the package structures, it seems that SI Modules generally follow a defined package structure, e.g. org.springframework.integration.<module_name>.config etc.

It seems, though, we have certain deviations e.g. putting stuff into the module root package versus having a "core" package.

This might also be something to think about, when we create a common guideline/documentation for writing custom components.

Some common package structures:

* "inbound"
* "outbound"
* "core" (for common)
* "support" if even core needs something
* "mapping"
* "config"
* The "root" package should contain only exceptions and possibly base interfaces

---

**Affects:** 2.1 RC2

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/6375

* Move respective classes to the `inbound`, `outbound` or `support` packages to make the whole picture for the API more concise and clean

The firs commit is for `stream` module


artembilan:
This is not too hard task moving classes from one package to another with respective `@Deprecated` routine, but this might be hard to review.
Just because the code has been copied/pasted and looks like a new now.
So, let me know if you still would like this PRs approach or that wouldbe OK if I just do the stuff and push to `main`.

I have changed the comment in the issue to checkboxes, so we may even divide those packages between us.

I decided to go with a  `@Deprecated` instead of plain move for easier migration even if we are in a major right now.

Any feedback welcome!

Thanks

artembilan:
I think that's enough for now.
It might be easier to review commit by commit, since each of them belongs to the specific module.
However, it is still hard because the number of changes getting bigger.

Please, let me know if this is OK with you and I'll continue to nail the issue.

Thanks

artembilan:
So, the PR is getting bigger and bigger.
Just because it is not only move to new packages to come to harmony in the end, but this also causes an entropy to preserve old classes with deprecation mark.
I also move test classes into new packages as well.
Therefore, I'm not sure what you'd like to review here.
Just let me know how you feel about this overhaul!

I'm going to push more since I got this spare time and it makes sense for the current major we are about to finish.

Thank you for your time since I feel guilty to waste it on this huge PR review 😄

spring-integration-stream/src/main/java/org/springframework/integration/stream/event/StreamClosedEvent.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.springframework.integration.stream.event;
+
+import org.springframework.integration.events.IntegrationEvent;
+
+/**
+ * Application event published when EOF is detected on a stream.
+ *
+ * @author Gary Russell
+ *
```
cppwfs, line 25: Nitpick.  
I'm not sure we need it, but do you think we need an @author tag?  Or is the commit history satisfactory for documentation?

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
```diff
@@ -103,59 +99,10 @@ public MarshallingWebServiceOutboundGateway(@Nullable String uri, WebServiceTemp
...
-
-		PassThroughRequestMessageCallback(@Nullable WebServiceMessageCallback requestCallback, Message<?> requestMessage) {
-			super(requestCallback, requestMessage);
-		}
-
-		@Override
-		public void doWithMessageInternal(WebServiceMessage message, Object payload) throws IOException {
-		}
 
+		super(destinationProvider, webServiceTemplate);
```
cppwfs, line 105: I'm debating...    Do we need tests on the constructors in the deprecated classes?    
My hyper nitpicky voice in the back of my head said we should at least mention it.

spring-integration-jmx/src/main/resources/org/springframework/integration/jmx/config/spring-integration-jmx.xsd
```diff
@@ -301,7 +301,7 @@
 						</xsd:documentation>
```
cppwfs, line 301: Line 131 needs to be updated to the new package name.

spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
```diff
@@ -94,48 +63,16 @@
...
-
-	private Predicate<Path> watchDirPredicate = path -> true;
+@Deprecated(forRemoval = true, since = "7.0")
+public class FileReadingMessageSource extends org.springframework.integration.file.inbound.FileReadingMessageSource {
 
 	/**
 	 * Create a FileReadingMessageSource with a naturally ordered queue of unbounded capacity.
 	 */
 	public FileReadingMessageSource() {
-		this(null);
```
cppwfs, line 138: Wouldn't this be super(null)?

cppwfs (changes requested):
What a PR!   😆 

Great work!   Breaking up the inbound and outbound classes into their own packages makes perfect sense. 

Just my usual nitpicks.

mrpiggi:
> This is not too hard task moving classes from one package to another with respective `@Deprecated` routine, but this might be hard to review. Just because the code has been copied/pasted and looks like a new now. So, let me know if you still would like this PRs approach or that wouldbe OK if I just do the stuff and push to `main`.
> I decided to go with a `@Deprecated` instead of plain move for easier migration even if we are in a major right now.

Looks like a very tedious task. Just be aware, that if you proceed in this manner, you will actually lose the git history of all the moved files. If you want to retain the history you would have to do the moving in one commit and re-introduce the now deprecated classes as new ones in a subsequent commit. Not sure, if it's worth the effort but I just wanted to mention it.

cppwfs:
Thank you for your comment it is very much appreciated.
Even after the removal of the deprecated versions from the main branch, the original classes are still be accessible in all previous tags and branches with their complete history. However, it is an additional step to do a blame.

spring-integration-stream/src/main/java/org/springframework/integration/stream/event/StreamClosedEvent.java
artembilan: This was just a class move.
I add (and request) `@author` tag when a reasonable change is made to the class.

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
artembilan: I could not find yet a decent reason in my guts to cover deprecated classes with tests.
I treat deprecated API from my dependencies as obsolete, therefore it might be broken.
So, I switch to a new API as soon as possible.

But I may consider to have coverage for deprecated as well if you give me more arguments.
Plus, I have you as extra help to do that supporting work 😄

spring-integration-file/src/main/java/org/springframework/integration/file/FileReadingMessageSource.java
artembilan: Why?
If we don't specify `super()`, that `super()` is compiled-in.
And that `super()` does exactly what you are asking.

artembilan:
> do the moving in one commit and re-introduce the now deprecated classes as new ones

That is great idea!
I don't think I'll revert anything what I have done so far, but I'll try to follow your recommendations for two remaining modules.

Thank you for review!

> in all previous tags and branches with their complete history

Yeah... Unfortunately, after removing deprecated classes in the next version we won't have their mentioning anywhere to follow the history.

spring-integration-ws/src/main/java/org/springframework/integration/ws/MarshallingWebServiceOutboundGateway.java
cppwfs: In the code review, I really didn't see an appreciable benefit to adding them.  
I think the rule is, if you add anything outside `super(...)` then a test would be required.
But as with anything, it was worth mentioning.

cppwfs (approved):
LGTM 
Do you want me to merge and we start a new PR for the rest of the work, or continue with this PR?

artembilan:
>  or continue with this PR?

There are just only two modules left.
And I'm at the moment on JDBC.
So, I don't see a reason for a new PR.

Just state tuned!
I'll finish the work for this task today.
Thank you!

cppwfs:
Sounds great!

artembilan:
Just pushed two commits for JDBC module.
I think that is exactly what you mean about preserving history and blame.

artembilan:
OK. I think this is all whatever has needed to be done for the issue.

spring-integration-ip/src/main/java/org/springframework/integration/ip/tcp/package-info.java
```diff
@@ -1,5 +0,0 @@
-/**
- * Base package for TCP Support.
- */
-@org.jspecify.annotations.NullMarked
```
cppwfs, line 4: Should this be readded for the deprecated classes?

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -57,6 +57,13 @@ The `RetryStateGenerator` abstraction and its `SpelExpressionRetryStateGenerator
 
 See xref:handler-advice/classes.adoc#retry-advice[Retry Advice] for more information.
 
+All the modules now follow the standard package structure.
+The inbound and outbound components are declared in the `inbound` and `outbound` packages, respectively.
+The `MessageChannel` implementations in the `channel` package.
+Most of the modules have followed this rule from day first.
```
cppwfs: Update to say "from the first day"

cppwfs (changes requested):
Looks good.
Just a question and a quick doc fix and its ready to go.

cppwfs (approved):
LGTM.   

Want me to merge?

PR #10460, opened 2025-10-03
Drop switch from Redis unlink to delete

michaldo:
Drop support for Redis 3, which is not maintained since 2018

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
@@ -37,6 +37,7 @@ import org.springframework.util.Assert;
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Michal Domagala
  *
  * @since 2.1
  */
@@ -47,8 +48,6 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 	private final RedisTemplate<Object, Object> redisTemplate;
 
 	private boolean valueSerializerSet;
-
-	private volatile boolean unlinkAvailable = true;
 
 	/**
 	 * Construct {@link RedisMessageStore} based on the provided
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
After discussion with Spring Data team, we agreed that fully dropping `DEL` support is fair game from now on.
Please, consider to refine the logic just deal with `UNLINK`.
Also, take a look into `RedisQueueInboundGateway`, `RedisQueueMessageDrivenEndpoint` and `RedisMessageStore` where we still use `DEL`.

Thanks

michaldo:
I decided to remove switch from unlink to dev
About places where del is used, I will take a look later

artembilan:
> I will take a look later

Thank you!

Let's us know, please, if you are going to that as a part of this PR or so!
In the end we would need to mention in the `whats-new.adoc` such a change in Redis module.
Also, the `redis.adoc` need to be revised if we still mention `DEL` somewhere there and probably an algorithm you've just dropped.

artembilan (changes requested):
Also, just noticed:
```
Signed-off-by: michaldo <michaldo@github.com>
```

Please, use your legal name instead according to DCO requirements.
You can configure your Git client to produce a proper name for you into those commits.

michaldo:
This PR is functionally closed. I encountered problem of not careful switch, analyzed it and fixed. I am sure this change is safe because no modern application should be switched to delete mode.
Regarding other places where delete is used, I'm not familiar with context. I'm going to take a look.

artembilan (changes requested):
Please, add your name into the `@author` list of the affected class.
Also, give us your thought about other classes in Redis module which still use `DEL`.

Thank you again!

michaldo:
I add myself as one of `@author`'s
Regarding other classes:

- RedisMessageStore: is it same case switch from unlink -> delete and I removed it also

- RedisQueueInboundGateway, RedisQueueMessageDrivenEndpoint: they use delete command within JMX `@ManagedOperation` command. I feel that JMX action looks like admin action and I think If were someone who executes admin action I prefer to be sure that when action is completed everything is done. It mean I prefer synchronous delete over asynchronous unlink

artembilan:
>  I prefer to be sure that when action is completed everything is done

Sounds reasonable.
That was oversight on my side when I performed search for `RedisTemplate.delete()` 😄 

Thank you for looking over there!

PR #10467, opened 2025-10-06
RedisLockRegistry stopRenew not thread safe #10446

### LINKED ISSUE #10446: RedisLockRegistry stopRenew not thread safe

spring-integration-redis 6.5.0

**Describe the bug**

Basically stopRenew() can be called from multiple threads and results in: 
`java.lang.NullPointerException: Cannot invoke "java.util.concurrent.ScheduledFuture.cancel(boolean)" because "this.renewFuture" is null
	at org.springframework.integration.redis.util.RedisLockRegistry$RedisLock.stopRenew(RedisLockRegistry.java:601) ~[spring-integration-redis-6.5.0.jar:6.5.0]
	at org.springframework.integration.redis.util.RedisLockRegistry$RedisLock.removeLockKey(RedisLockRegistry.java:572) ~[spring-integration-redis-6.5.0.jar:6.5.0]
	at org.springframework.integration.redis.util.RedisLockRegistry$RedisLock.unlock(RedisLockRegistry.java:536) ~[spring-integration-redis-6.5.0.jar:6.5.0]`

**To Reproduce**

Isn't easy to reproduce, I am not an expert of Redis. BUT by static code analysis I can tell that there is one opportunity for this error to occur. 
1. we call unlock(), which calls stopRenew()
2. There is also a taskExecutor that periodically runs renew(). So at the same time renew() could be running in separate thread, and it could happen (?) that 
`Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
					RENEW_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
					RedisLockRegistry.this.clientId, String.valueOf(RedisLockRegistry.this.expireAfter)));`
returns false, thus we go into stopRenew() and we have a race condition. 
I am not sure why would we get "FALSE" there, but maybe someone here can answer this question?

**Expected behavior**

I am expecting unlock() to be a safe method, doing it job regardless the internal state of a lock. 

**Sample**

Just happened on production, and as it is a race condition exception I think static analysis should be sufficient. 

I will create a PR with proposition to add "synchronised" keyword on stopRenew() method. 

Renewable logic is very helpful and was introduced in commit: Commit 4d08e11
Tagging @artembilan  , seems like you will be able to tell whether adding "synchronised" is sufficient :)

PiotrDuz:
Closes #10446

Makes stopRenew() method thread safe.

Signed-off-by: PiotrDuz <39244523+PiotrDuz@users.noreply.github.com>


artembilan (changes requested):
```
Signed-off-by: PiotrDuz <39244523+PiotrDuz@users.noreply.github.com>
```

Please, consider to use your legal name and real email.
For DCO requirements.

PiotrDuz:
Hey, So I wouldn't like to post my personal data open in the web. 
The name is a shortcut of my real name, so there is some association if it is really required :) 

And all the details match ym account , also I sign it from the same account that authored the commit - then there is little room to claim that author has not agreed on the terms. Would it be ok ? :>

artembilan:
OK, @PiotrDuz ,

Got an answer from our management.
So, as long as the email is valid `@users.noreply.github.com`, we even don't need your real name.
If legal concern arises, we would reach your through GitHub support. 😄 

Now I can review and merge your PR.

Thank you for your patience!

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -598,9 +598,11 @@ protected final boolean renew(long expireAfter) {
 			return res;
 		}
 
+		/* This method can be called by more than 1 thread, thus we need to make sure that it is safe*/
```
artembilan: We don't need this comment here.
Because it is obvious from the method and property scope that it is a bug with concurrent access.
Therefore, we follow "local variable extraction" pattern everywhere in similar situations.

Please, remove this comment altogether.

artembilan (changes requested):
And feel free to add your name to the `@author` list of the affected class.

Thank you!

artembilan (changes requested):
```

/home/runner/work/spring-integration/spring-integration/spring-integration-redis/src/test/java/org/springframework/integration/redis/outbound/RedisQueueOutboundChannelAdapterTests.java:134: error: cannot find symbol
		handler.setSerializer(new Jackson3JsonRedisSerializer<Object>(Object.class));
		                          ^
  symbol:   class Jackson3JsonRedisSerializer
  location: class RedisQueueOutboundChannelAdapterTests
```
Please, consider to rebase your branch to the latest `upstream/main`.
We have the fix for that Spring Data change in already.

artembilan:
Any updates, please?
We have release on next Tuesday, so would be great to make this in.
If you cannot update, no worries: we fix it at last moment before release.
Thanks

PiotrDuz:
Hey @artembilan  , sorry for the delay. 
I have removed the comment and rebased on the newest upstream main branch. 
Regards!

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -99,6 +99,7 @@
  * @author Youbin Wu
  * @author Michal Domagala
  * @author Severin Kistler
+ * @author PiotrDuz
```
artembilan: No, only real name can go here.
This part is for the credability: when some one reads our code, they would see who did changes to the class.
When it is nick-name like this there is not too much to acknowledge the person.
Therefore, if you don't want to share your real name, then you might not interested in the credit, thus no need in this `@author` in the class.

Thanks for understanding!

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
PiotrDuz: Understood, this has been changes. No worries ;)

PR #10469, opened 2025-10-07
GH-3189: Properly handle `async` in `ScatterGatherHandler`

### LINKED ISSUE #3189: Scatter Gather Blocking threads when used along with WebFlux & Netty
**Affects Version(s):** 5.3

---
Scatter Gather is blocking on thread waiting for response at `gatherResultChannel.receive()` as this is blocking call.

Is there any alternative option to avoid waiting for response like this, instead subscribe to the response from some subscribable channel or any other alternative(not able to think of a solution) so this will work perfectly fine without holding up on threads.

I have been looking out for solutions but couldn't find any perfect one.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/3189

Despite supporting `async = true`, the `ScatterGatherHandler` does blocking in its `handleRequestMessage()`
on the `gatherResultChannel.receive()` call.

* Fix `ScatterGatherHandler` to handle an `async` mode via internal `Mono` for the reply object
* Use pattern variable expressions for `ifs` in the `ScatterGatherHandler.doInit()` for the better readability
* Extract `ScatterGatherHandler.replyFromGatherResult()` method to avoid code duplication
* Document a new (fixed) functionality

**Auto-cherry-pick to `6.5.x`**

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
@@ -16,7 +16,11 @@
 
 package org.springframework.integration.scattergather;
 
+import java.time.Duration;
+
 import org.jspecify.annotations.Nullable;
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.Sinks;
 
 import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.factory.BeanFactory;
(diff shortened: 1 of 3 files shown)


spring-integration-core/src/main/java/org/springframework/integration/scattergather/ScatterGatherHandler.java
```diff
@@ -191,7 +196,18 @@ private Message<?> enhanceScatterReplyMessage(Message<?> message) {
 	@Override
 	protected @Nullable Object handleRequestMessage(Message<?> requestMessage) {
```
cppwfs, line 200: Do we need to enumerate that a mono is returned when `async` is `true` else a `AbstractIntegrationMessageBuilder` is returned?

cppwfs (approved):
LGTM!

Just one question.

cppwfs:
Thank you for the update!

PR #10470, opened 2025-10-07
GH-6786: Demonstrate `#xpath()` with a `recipient-list-router`

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/6786

The original request was about an `xpath-router`, however using the `#xpath()` SpEL function we don't need anything else

* Show-case the `#xpath()` with a `<int:recipient-list-router>`
* Mention such a feature in the `xpath-routing.adoc`


cppwfs (approved):
LGTM!

Thanks for the test and  documentation!

Issue #10471, opened 2025-10-08
RedisLockRegistry using PUB_SUB_LOCK does not work on AWS Elasticache Valkey Serverless


kistlers:
**In what version(s) of Spring Integration are you seeing this issue?**

6.5.2

**Describe the bug**

RedisLockRegistry using PUB_SUB_LOCK does not work on AWS Elasticache Valkey Serverless. Internally, Elasticache uses `SPUBLISH` instead of `PUBLISH` which is used in this https://github.com/spring-projects/spring-integration/blob/381124451cd9dd74754bba0f5f4d26e073e8f414/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L667 script. Valkey cluster requires all sharded operations in a Lua script to hash to the same cluster slot cache. In the case of `PUBLISH`, this does not matter, but it does with `SPUBLISH` (used in Elasticache Valkey, as described).
In the script, there are three Redis operations:
- `redis.call('GET', KEYS[1])`
- `redis.call('UNLINK', KEYS[1])`
- `redis.call('PUBLISH', ARGV[2], KEYS[1])`
And I, as a developer, cannot control the cluster slot cache of the third operation for each separate call of the locking operation. `ARGV[2]` is the `unlockChannel` (see https://github.com/spring-projects/spring-integration/blob/381124451cd9dd74754bba0f5f4d26e073e8f414/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L690) set globally for this `RedisLockRegistry` (see https://github.com/spring-projects/spring-integration/blob/381124451cd9dd74754bba0f5f4d26e073e8f414/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L206).
I can easily control the cluster slot cache of all other operations by including a `{...}` in the cache, but I cannot do the same for the `unlockChannelKey`

**To Reproduce**

- Create a `RedisLockRegistry` with a `PUB_SUB_LOCK`
- Attach it to an AWS Elasticache Valkey Serverless
  - Alternatively (per my reasoning, that should work) change the `PUBLISH` to an `SPUBLISH` in the `UNLINK_UNLOCK_SCRIPT` and attach it to any valkey/redis.
- Run a `lockRegistry.executeLocked(..)` operation
[...]

PR #10473, opened 2025-10-09
Support for PubSub Lock on AWS Elasticache Valkey

kistlers:
This is an attempt to make issue #10471 work without support for `SPUBLISH`.
I append the `lockKey` to the `channelUnlockKey`. This allows us to control the cluster slot cache this operation is hashed into, by including a hash tag (`{...}`) in the `lockKey`. Refer to the issue description for an explanation of why this matters.

As a consequence, I had to change the `ChannelTopic` to a `PatternTopic` to keep receiving subscription events for the PubSub lock.

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.ConcurrentModificationException;
 import java.util.Date;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.UUID;
@@ -668,7 +669,7 @@ public final class RedisLockRegistry
 		private static final String UNLINK_UNLOCK_SCRIPT = """
 				local lockClientId = redis.call('GET', KEYS[1])
 				if (lockClientId == ARGV[1] and redis.call('UNLINK', KEYS[1]) == 1) then
-					redis.call('PUBLISH', ARGV[2], KEYS[1])
+					redis.call('PUBLISH', KEYS[2], KEYS[1])
 					return true
 				end
 				return false
@@ -689,9 +690,10 @@ public final class RedisLockRegistry
 
 		@Override
 		protected boolean removeLockKeyInnerUnlink() {
+			final String unLockChannelKey = RedisLockRegistry.this.unLockChannelKey + ":" + this.lockKey;
 			return Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
-					UNLINK_UNLOCK_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
-					RedisLockRegistry.this.clientId, RedisLockRegistry.this.unLockChannelKey + ":" + this.lockKey));
+					UNLINK_UNLOCK_REDIS_SCRIPT, List.of(this.lockKey, unLockChannelKey),
+					RedisLockRegistry.this.clientId));
 		}
 
 		private boolean subscribeLock(long time, long expireAfter) throws ExecutionException, InterruptedException {
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -53,7 +53,7 @@ import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.core.StringRedisTemplate;
 import org.springframework.data.redis.core.script.DefaultRedisScript;
 import org.springframework.data.redis.core.script.RedisScript;
-import org.springframework.data.redis.listener.ChannelTopic;
+import org.springframework.data.redis.listener.PatternTopic;
 import org.springframework.data.redis.listener.RedisMessageListenerContainer;
 import org.springframework.data.redis.listener.Topic;
 import org.springframework.integration.support.locks.DistributedLock;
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Please, add DCO into your commit message before we start reviewing: https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring.
Thanks

kistlers:
> Please, add DCO into your commit message before we start reviewing: https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring. Thanks

Done!

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -690,7 +691,7 @@ protected boolean tryRedisLockInner(long time, long expireAfter)
 		protected boolean removeLockKeyInnerUnlink() {
 			return Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
 					UNLINK_UNLOCK_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
-					RedisLockRegistry.this.clientId, RedisLockRegistry.this.unLockChannelKey));
+					RedisLockRegistry.this.clientId, RedisLockRegistry.this.unLockChannelKey + ":" + this.lockKey));
```
artembilan: I understand how this work from the pattern perspective.
But is it really OK to have so many channels for every single lock we are going to use in our application?

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
kistlers: Maybe not. However, I do not see any other way to ensure that all three redis calls in the `UNLINK_UNLOCK_SCRIPT` hash to the same shard if we are running on a system that requires it, like Valkey on AWS.

If the call to `PUBLISH` would not be in that script, that would make all calls in that script use the same `KEYS[1]` as the first argument to a redis call. But, I suppose is has to be in there.
If that were not necessary, the call to Publish could be in a separate script, but that might not be okay, for reasons I do not yet know.

kistlers: I did some research in the middle of typing out the last message.
As far as I do understand it, lots of channels are not an issue. This, as channels do not consume memory, only subscribers do, irrespective of pattern or exact match.

kistlers: On the other hand, extracting the `PUBLISH` call to a separate script might be okay, but it could happen that the application dies between the two script calls and that would make other subscribers wait for the remaining TTL of the lock.

On another note, my understanding is currently that any values used as a key of an operation should be included in the `keys`, rather than the `args`. Meaning either the current or my new `unlockChannelKey` should be passed as `KEYS[2]` rather than `ARGV[2]`. See https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/WorkingWithRedis.html, bullet point "Use sharded pub/sub" or https://redis.io/docs/latest/develop/pubsub/#sharded-pubsub

artembilan (changes requested):
This looks more closer to what I won't worry about 😄 .

If you confirm that this works with Valkey, please, add your name to the `@author` list of the changed class.

Please, also rebase your branch (and re-push to PR) to the latest `main` as I've just pushed the fix Redis tests according to the latest Spring Data Redis.

Thanks

kistlers:
According to my tests against valkey this is working.
My name is in the authors list of the file and I have rebased the branch. While at it, I prepended the issue id to the commit messages.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
artembilan: Yeah... What you say is not obvious from all those docs you are pointing, but your conclusion makes huge sense.
Therefore, I treat an issue as a bug and will back-port it into all the supported versions.
We really had to think about `channel` argument in the `PUBLISH` command as a key 🤦‍♂️

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -688,9 +690,10 @@ protected boolean tryRedisLockInner(long time, long expireAfter)
 
 		@Override
 		protected boolean removeLockKeyInnerUnlink() {
+			final String unLockChannelKey = RedisLockRegistry.this.unLockChannelKey + ":" + this.lockKey;
```
artembilan, line 693: Hold on. Why the value still has to be a part of the channel name?
According to your conclusion, the channel as a key would be hashed properly to the shard.

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
kistlers: The channel key has to be set once on creation of the `RedisLockRegistry` and cannot be changed after that. If I have to include the hash tag `{...}` in the channel key, then all my lock keys would have to hash to the exact same shard, of which there are 16384 that are distributed among the nodes in a cluster setup. Hence, every redis operation would have to hit only one single node and shard in a cluster.

And, the other operations in the same script use only the lock key, which I would also have to force into the same shard as the unlock key.

artembilan: OK. I still don't understand that how that Redis hashing to the shard works, but since you confirm that this is only the and it works well with Valkey, then I'll merge this as is.
Thank you!

artembilan:
@kistlers ,

PR #10485, opened 2025-10-12
Use `assertThatException`

ngocnhan-tran1996:

### DIFF
--- a/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/config/InnerPollerParserTests.java
+++ b/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/config/InnerPollerParserTests.java
@@ -83,15 +83,15 @@ public class InnerPollerParserTests {
 						xsi:schemaLocation="http://www.springframework.org/schema/integration https://www.springframework.org/schema/integration/spring-integration.xsd
 							http://www.springframework.org/schema/integration/jdbc https://www.springframework.org/schema/integration/jdbc/spring-integration-jdbc.xsd
 							http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd">
-				
+
 					<int:poller id="outer" fixed-rate="5000"/>
-				
+
 					<int:channel id="someChannel"/>
-				
+
 					<int-jdbc:inbound-channel-adapter channel="someChannel" jdbc-operations="ops" query="select 1">
 						<int:poller ref="outer" default="true"/>
 					</int-jdbc:inbound-channel-adapter>
-				
+
 					<bean id="ops" class="org.mockito.Mockito" factory-method="mock">
 						<constructor-arg value="org.springframework.jdbc.core.JdbcOperations"/>
 					</bean>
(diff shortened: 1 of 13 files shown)


spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/config/InnerPollerParserTests.java
```diff
@@ -42,120 +41,101 @@ public void testRefGood() {
...
-							"		<int:poller ref=\"outer\" fixed-rate=\"1000\"/>" + // <<<<< fixed-rate not allowed here
-							"	</int-jdbc:inbound-channel-adapter>" +
-							"" +
-							"	<bean id=\"ops\" class=\"org.mockito.Mockito\" factory-method=\"mock\">" +
-							"		<constructor-arg value=\"org.springframework.jdbc.core.JdbcOperations\"/>" +
-							"	</bean>" +
-							"</beans>";
+		// Load context from a String to avoid IDEs reporting the invalid configuration
+		String badContext =
+				"<?xml version=\"1.0\" encoding=\"UTF-8\"?>" +
```
artembilan: Well, since we are already here, please, consider to use Java text block instead of concatenations.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/config/InnerPollerParserTests.java
```diff
@@ -42,120 +41,101 @@ public void testRefGood() {
...
-		}
-		catch (BeanDefinitionParsingException bdpe) {
-			assertThat(bdpe.getMessage()
-					.startsWith("Configuration problem: A 'poller' element that provides a 'ref' must have no other " +
-							"attributes."))
-					.isTrue();
-		}
+		Resource resource = new ByteArrayResource(badContext.getBytes());
+		assertThatExceptionOfType(BeanDefinitionParsingException.class)
+				.isThrownBy(() -> new GenericXmlApplicationContext(resource).close())
```
artembilan: We don't need `close()` since we are going to fail.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/config/InnerPollerParserTests.java
```diff
@@ -42,120 +41,101 @@ public void testRefGood() {
...
-			fail("Expected Failure to load ApplicationContext");
-		}
-		catch (BeanDefinitionParsingException bdpe) {
-			assertThat(bdpe.getMessage()
-					.startsWith("Configuration problem: A 'poller' element that provides a 'ref' must have no other attributes."))
-					.isTrue();
-		}
+		Resource resource = new ByteArrayResource(badContext.getBytes());
+		assertThatExceptionOfType(BeanDefinitionParsingException.class)
+				.isThrownBy(() -> new GenericXmlApplicationContext(resource).close())
```
artembilan: DITTO

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/config/InnerPollerParserTests.java
```diff
@@ -42,120 +41,101 @@ public void testRefGood() {
...
```
artembilan: DITTO

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/config/StoredProcInvalidConfigsTests.java
```diff
@@ -39,54 +38,32 @@
...
-		catch (BeanDefinitionParsingException e) {
-			assertThat(e.getMessage()
-					.contains("Exactly one of 'stored-procedure-name' or 'stored-procedure-name-expression' is " +
-							"required"))
-					.isTrue();
-		}
+	public void testProcedureNameAndExpressionExclusivity() {
+
+		assertThatExceptionOfType(BeanDefinitionParsingException.class)
+				.isThrownBy(() -> this.bootStrap("nameAndExpressionExclusivity"))
```
artembilan: Now we prefer to not use `this.` prefix for method calls.

spring-integration-mongodb/src/test/java/org/springframework/integration/mongodb/outbound/MongoDbOutboundGatewayTests.java
```diff
@@ -102,28 +102,23 @@ void testNoFactorySpecified() {
...
-			gateway.handleRequestMessage(message);
-			fail("Expected the test case to throw an IllegalArgumentException");
-		}
-		catch (IllegalStateException e) {
-			assertThat(e.getMessage()).isEqualTo("no query or collectionCallback is specified");
-		}
+		assertThatIllegalStateException()
+				.isThrownBy(() -> {
+					gateway.afterPropertiesSet();
+					gateway.handleRequestMessage(message);
```
artembilan: We fail here just after `gateway.afterPropertiesSet();`, therefore second call is not reachable, redundant and misleading.
Plus, this second call is not asserted anyhow.
Well, it just has to go.

artembilan (changes requested):

PR #10486, opened 2025-10-13
GH-7971: Add `FileReadingMessageSource.directoryExpression`

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/7971

Extend a `FileReadingMessageSource.directory` logic to the SpEL expression to evaluate on each scan

* Introduce an internal `record DirFile` to keep the `File` and scanned directory per file
* Propagate a new directory down to the `WatchServiceDirectoryScanner` and restart it
* Calculate a root directory from the failed `Message` we offer back to the queue
* Fix tests for a new `FileReadingMessageSource.directoryExpression` property
* Expose a `Supplier<File>` configuration for Java DSL
* Modify `FileTests.MyService.pollDirectories()` for a `Supplier<File>` configuration
* Document this new feature

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/inbound/FileReadingMessageSource.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/inbound/FileReadingMessageSource.java
@@ -46,7 +46,9 @@ import java.util.function.Predicate;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.context.Lifecycle;
+import org.springframework.expression.Expression;
 import org.springframework.integration.endpoint.AbstractMessageSource;
+import org.springframework.integration.expression.ValueExpression;
 import org.springframework.integration.file.DefaultDirectoryScanner;
 import org.springframework.integration.file.DirectoryScanner;
 import org.springframework.integration.file.FileHeaders;
@@ -78,7 +80,7 @@ import org.springframework.util.Assert;
  * {@link org.springframework.integration.file.filters.AcceptOnceFileListFilter}
  * would allow for this.
  * <p>
- * If a external {@link DirectoryScanner} is used, then the {@link FileLocker}
+ * If an external {@link DirectoryScanner} is used, then the {@link FileLocker}
  * and {@link FileListFilter} objects should be set on the external
  * {@link DirectoryScanner}, not the instance of FileReadingMessageSource. An
  * {@link IllegalStateException} will result otherwise.
@@ -108,15 +110,10 @@ public class FileReadingMessageSource extends AbstractMessageSource<File> implem
 
 	private final AtomicBoolean running = new AtomicBoolean();
 
-	/*
-	 * {@link PriorityBlockingQueue#iterator()} throws
-	 * {@link java.util.ConcurrentModificationException} in Java 5.
-	 * There is no locking around the queue, so there is also no iteration.
-	 */
-	private final Queue<File> toBeReceived;
+	private final Queue<DirFile> toBeReceived;
 
 	@SuppressWarnings("NullAway.Init")
-	private File directory;
+	private Expression directoryExpression;
 
 	private DirectoryScanner scanner = new DefaultDirectoryScanner();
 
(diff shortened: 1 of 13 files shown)


spring-integration-file/src/main/java/org/springframework/integration/file/inbound/FileReadingMessageSource.java
```diff
@@ -183,7 +184,18 @@ public FileReadingMessageSource(@Nullable Comparator<File> receptionOrderCompara
 	 */
 	public void setDirectory(File directory) {
 		Assert.notNull(directory, "directory must not be null");
-		this.directory = directory;
+		setDirectoryExpression(new ValueExpression<>(directory));
+	}
+
+	/**
+	 * Specify a SpEL expression for an input directory.
```
cppwfs, line 191: Nitpick.   But isn't it an  implementation of `Expression` instead of a single implementation of `SpELExpression`?

spring-integration-file/src/test/java/org/springframework/integration/file/dsl/FileTests.java
```diff
@@ -436,27 +435,24 @@ public PollableChannel dynamicAdaptersResult() {
...
-						.get();
-				this.beanFactory.initializeBean(integrationFlow, directory.getName());
-				this.beanFactory.getBean(directory.getName() + ".transformer", Lifecycle.class).start();
-				this.beanFactory.getBean(directory.getName() + ".adapter", Lifecycle.class).start();
-			}
+			AtomicInteger index = new AtomicInteger(0);
+			StandardIntegrationFlow integrationFlow = IntegrationFlow
+					.from(Files.inboundAdapter(() -> directories[index.getAndIncrement() % directories.length])
+									.recursive(true),
+							e -> e.poller(p -> p.fixedDelay(100))
```
cppwfs, line 449: Just curious, Was the drop to a delay from 1000 milli to 100 milli on purpose?

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -103,6 +103,12 @@ Previously deprecated classes in the `spring-integation-hazelcast` module, such
 The `AbstractMqttMessageDrivenChannelAdapter` and `ClientManager` implementations now expose a `quiescentTimeout` option which is propagated in their `stop()` method down to the `disconnectForcibly()` API of the MQTT Paho clients.
 See xref:mqtt.adoc[] for more information.
 
+[[x7.0-files-changes]]
+=== Files Changes
+
+The `FileReadingMessageSource` now can be configured with a SpEL `Expression` for its `directory` property.
```
cppwfs, line 109: Same question as above on the use of Spel implementation of `Expression`.

cppwfs (changes requested):
Looks great!

My usual annoying nitpicks ;-)

spring-integration-file/src/main/java/org/springframework/integration/file/inbound/FileReadingMessageSource.java
artembilan: The framework is called `SpEL` and because of many `Expression` classes in the wild it would be better to be specific in our JavaDocs of what exactly framework an `Expression` instance is expected for this setter.
We also use `SpEL` term everywhere else, including docs, so it should not be confusing what is expected.

spring-integration-file/src/test/java/org/springframework/integration/file/dsl/FileTests.java
artembilan: True. I'm very suspicious when test is slower than it should be.
Like in this case, we've got before a delay of one second in between reading from those directories we iterate in the supplier.
Such a delay is not justified for this test goal.

spring-integration-file/src/main/java/org/springframework/integration/file/inbound/FileReadingMessageSource.java
cppwfs: Thanks!    That makes sense.   Just wanted to make sure.

spring-integration-file/src/test/java/org/springframework/integration/file/dsl/FileTests.java
cppwfs: Cool.   Just wanted to make sure!

cppwfs (approved):
LGTM .  
I love the addition!

PR #10507, opened 2025-10-17
Upgrade to Dokka 2.1.0

cppwfs:
- Update `org.jetbrains.dokka` plugin from 2.0.0 to 2.1.0
- Remove Jackson 2.15.3 version resolution strategy workaround that was needed to avoid conflicts in Dokka 2.0.0
- Migrate from deprecated V1 tasks (`dokkaHtmlPartial`, `dokkaHtmlMultiModule`) to V2 API (`dokkaGeneratePublicationHtml`, `dokkaGenerate`)
- Replace `dokkaSourceSets.main` configuration block with `dokkaSourceSets.configureEach`
- Update `docsZip` task to depend on new `dokkaGeneratePublicationHtml` task and add explicit dependency on `:spring-integration-core:dokkaGenerate`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -115,6 +115,11 @@ ext {
     ztZipVersion = '1.17'
 
     javaProjects = subprojects - project(':spring-integration-bom')
+
+    // Modules that have Kotlin sources to document using dokka
+    dokkaProjects = [
+            'spring-integration-core'
+    ]
 }
 
 allprojects {
@@ -188,8 +193,6 @@ configure(javaProjects) { subproject ->
     apply plugin: 'kotlin'
     apply plugin: 'kotlin-spring'
     apply plugin: 'io.spring.nullability'
-    apply plugin: 'org.jetbrains.dokka'
-
 
     apply from: "${rootDir}/gradle/publish-maven.gradle"
 
@@ -395,27 +398,6 @@ configure(javaProjects) { subproject ->
 
     check.dependsOn checkClasspathForConflicts, javadoc
 
-
-    dokka {
-        dokkaPublications.html {
-            outputDirectory.set(file(rootProject.layout.buildDirectory.file('kdoc')))
-        }
-        dokkaSourceSets.configureEach {
-            sourceRoots.setFrom(file('src/main/kotlin'))
-            classpath.from(sourceSets['main'].runtimeClasspath)
-            externalDocumentationLinks.register("spring-integration-api") {
-                url.set(uri("https://docs.spring.io/spring-integration/docs/$version/api/"))
-                packageListUrl.set(file('build/docs/javadoc/element-list').toURI())
-            }
-            externalDocumentationLinks.register("reactor-core") {
-                url.set(uri('https://projectreactor.io/docs/core/release/api/'))
-            }
-            externalDocumentationLinks.register("reactor-streams") {
-                url.set(uri('https://www.reactive-streams.org/reactive-streams-1.0.3-javadoc/'))
-            }
-        }
-    }
-
     publishing {
         publications {
             mavenJava(MavenPublication) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -520,24 +511,14 @@ project('spring-integration-core') {
...
-                    packageListUrl.set(file('build/docs/javadoc/element-list').toURI().toURL())
-                }
-                externalDocumentationLink {
-                    url.set(new URL('https://projectreactor.io/docs/core/release/api/'))
-                }
-                externalDocumentationLink {
-                    url.set(new URL('https://www.reactive-streams.org/reactive-streams-1.0.3-javadoc/'))
-                }
-            }
+dokka {
```
artembilan: Something off with indents?

build.gradle
```diff
@@ -520,24 +511,14 @@ project('spring-integration-core') {
...
+        dokkaPublications.html {
+            outputDirectory.set(file('../build/kdoc'))
```
artembilan: I think we have to use `rootProject.layout.buildDirectory.file('kdoc')` instead.
I don't think it can be resolved properly on all the operation systems.

build.gradle
```diff
@@ -520,24 +511,14 @@ project('spring-integration-core') {
...
+        }
+        dokkaSourceSets.configureEach {
+            noJdkLink.set(false)
+            sourceRoots.setFrom(file('src/main/kotlin'))
+            classpath.from(sourceSets['main'].runtimeClasspath)
```
artembilan: What happened with all those `externalDocumentationLink` from our previous version?

build.gradle
```diff
@@ -1151,7 +1137,8 @@ tasks.register('schemaZip', Zip) {
 }
 
 tasks.register('docsZip', Zip) {
-    dependsOn 'dokkaHtmlMultiModule'
+    dependsOn 'dokkaGeneratePublicationHtml'
+    dependsOn ':spring-integration-core:dokkaGenerate'
```
artembilan: Why is this?
Doesn't sound right place for such a dependency.
What happened to multi in Dokka?
Why it cannot collect all the sub-projects as before any more?
Not directly a question to you, but we need to figure out how to get rid of this explicit dependency.
Imaging now that would have Kotlin API in other modules.

artembilan (changes requested):
Thank you for a swift turn around!

Not too much, just couple nit-picks and some concerns.
Can you share, please, with me a screenshot how the final doc looks?

After this we would be able to migrate to Gradle `9.1` finally 😄

cppwfs:
image
image

build.gradle
cppwfs: We only need to run the documentation in this module.

build.gradle
```diff
@@ -402,6 +395,27 @@ configure(javaProjects) { subproject ->
 
     check.dependsOn checkClasspathForConflicts, javadoc
 
+
+    dokka {
+        dokkaPublications.html {
+            outputDirectory.set(file(rootProject.layout.buildDirectory.file('kdoc')))
```
artembilan: Why do we need call `file()` twice?

build.gradle
```diff
@@ -197,6 +188,8 @@ configure(javaProjects) { subproject ->
     apply plugin: 'kotlin'
     apply plugin: 'kotlin-spring'
     apply plugin: 'io.spring.nullability'
+    apply plugin: 'org.jetbrains.dokka'
```
artembilan: This is not correct: it is waste of resource to apply the plugin in the sub-project where is nothing with Kotlin.

artembilan (changes requested):
This is what Spring Framework does:

https://github.com/spring-projects/spring-framework/blob/main/framework-api/framework-api.gradle

And this line is strange for me: https://github.com/spring-projects/spring-framework/blob/main/buildSrc/src/main/java/org/springframework/build/KotlinConventions.java#L40

artembilan (changes requested):
Looks like there is no way to make it transparent: https://kotlinlang.org/docs/dokka-migration.html#update-documentation-aggregation-in-multi-module-projects.
We have to be explicit with that our module:
```
dependencies {
    dokka(project(':spring-integration-core'))
}
```
in the root project 😢

cppwfs:
I smoothed it out a bit in this last commit.   But yeah, they dropped that support.

build.gradle
artembilan: Why do we still call `file(file())`?

build.gradle
```diff
@@ -115,6 +115,11 @@ ext {
     ztZipVersion = '1.17'
 
     javaProjects = subprojects - project(':spring-integration-bom')
+
+    // Modules that have Kotlin sources to document using dokka
+    dokkaProjects = [
```
artembilan: Can we make it more smarted without such a variable?

build.gradle
```diff
@@ -418,6 +414,30 @@ configure(javaProjects) { subproject ->
     }
 }
 
+// Process only the modules that require kotlin docs
+configure(subprojects.findAll { dokkaProjects.contains(it.name) }) { subproject ->
```
artembilan: We have already a `configure` above.
Such a Dokka configuration could be done there.
Can we go similar to SF way? Kinda check if there is `sourceSets.main.kotlin`, then apply Dokka config in that `subproject`?

build.gradle
```diff
@@ -1151,7 +1143,7 @@ tasks.register('schemaZip', Zip) {
 }
 
 tasks.register('docsZip', Zip) {
-    dependsOn 'dokkaHtmlMultiModule'
+    dependsOn dokkaProjects.collect { ":${it}:dokkaGenerate" }
```
artembilan: Probably with `dokka()` as dependency like in SF, we still would need to have `dokka` in the root project and here that would be just ` dependsOn 'dokkaGenerate'

artembilan (changes requested):
Strange. I thought I sent my review.
Sorry about that.

PR #10508, opened 2025-10-18
Add Jackson JSON serialization for JdbcChannelMessageStore

### LINKED ISSUE #9312: JSON serialization for JDBC channel message store
### Discussed in https://github.com/spring-projects/spring-integration/discussions/9309

<sup>Originally posted by **wimdeblauwe** July  8, 2024</sup>
According to [the documentation](https://docs.spring.io/spring-integration/reference/jdbc/message-store.html#custom-message-insertion), you can use JSON as the serialization format for your JDBC channel message store by overriding the `ChannelMessageStorePreparedStatementSetter`. However, I found it was not that easy. If you only do that, the `MessageRowMapper` throws an exception that it cannot read what was written.

These are all the things I had to do to make it work:

1) Inject a custom queryProvider, preparedStatementSetter and messageRowMapper into the configuration of the `JdbcChannelMessageStore`:

```
  @Bean
  JdbcChannelMessageStore jdbcChannelMessageStore(
      DataSource dataSource,
      ChannelMessageStoreQueryProvider queryProvider,
      ChannelMessageStorePreparedStatementSetter preparedStatementSetter,
      MessageRowMapper messageRowMapper) {
    JdbcChannelMessageStore jdbcChannelMessageStore = new JdbcChannelMessageStore(dataSource);
    jdbcChannelMessageStore.setTablePrefix(CONCURRENT_METADATA_STORE_PREFIX);
    jdbcChannelMessageStore.setChannelMessageStoreQueryProvider(
        new PostgresChannelMessageStoreQueryProvider());
    jdbcChannelMessageStore.setChannelMessageStoreQueryProvider(queryProvider);
    jdbcChannelMessageStore.setPreparedStatementSetter(preparedStatementSetter);
    jdbcChannelMessageStore.setMessageRowMapper(messageRowMapper);
    return jdbcChannelMessageStore;
  }
```

2) Custom `ChannelMessageStorePreparedStatementSetter` that writes 2 colums: one with the type of the payload, and one with the serialized JSON of the payload.

```
  @Bean
  ChannelMessageStorePreparedStatementSetter channelMessageStorePreparedStatementSetter(ObjectMapper objectMapper) {
    return new ChannelMessageStorePreparedStatementSetter() {
      @Override
      public void setValues(PreparedStatement preparedStatement, Message<?> requestMessage, Object groupId, String region,
          boolean priorityEnabled) throws SQLException {
        super.setValues(preparedStatement, requestMessage, groupId, region, priorityEnabled);
        try {
          preparedStatement.setString(6, requestMessage.getPayload().getClass().getName());
          String json = objectMapper.writeValueAsString(requestMessage.getPayload());
          preparedStatement.setObject(7, json, java.sql.Types.OTHER);
        } catch (JsonProcessingException e) {
          throw new RuntimeException("Unable to store message", e);
        }
      }
    };
  }
```

3) Custom `MessageRowMapper`. Unfortunately, the setter at `JdbcChannelMessageStore` only accepts the `MessageRowMapper` class, not the `RowMapper` interface, but overriden the `mapRow` is luckily possible.

```
  @Bean
  MessageRowMapper messageRowMapper(ObjectMapper objectMapper) {
    return new MessageRowMapper(null, null) {
      @Override
      public Message<?> mapRow(ResultSet rs, int rowNum) throws SQLException {
        try {
          String payloadType = rs.getString(rs.findColumn("MESSAGE_PAYLOAD_TYPE"));
          String s = rs.getString(rs.findColumn("MESSAGE_BYTES"));
          Object o = objectMapper.readValue(s, Class.forName(payloadType));
          return new MutableMessage<>(o);
        } catch (JsonProcessingException | ClassNotFoundException e) {
          throw new RuntimeException("Unable to read message", e);
        }
      }
    };
  }
```

4) To support the extra column for the payload type, I needed to provide a custom query provider:

```
@Bean
  ChannelMessageStoreQueryProvider queryProvider() {
    return new PostgresChannelMessageStoreQueryProvider(){
      @Override
      public String getCreateMessageQuery() {
        return """
				INSERT into %PREFIX%CHANNEL_MESSAGE(
					MESSAGE_ID,
					GROUP_KEY,
					REGION,
					CREATED_DATE,
					MESSAGE_PRIORITY,
					MESSAGE_PAYLOAD_TYPE,
					MESSAGE_BYTES)
				values (?, ?, ?, ?, ?, ?, ?)
				""";
      }

      @Override
      public String getPollFromGroupQuery() {
        return """
				delete
				from %PREFIX%CHANNEL_MESSAGE
				where CTID = (select CTID
								from %PREFIX%CHANNEL_MESSAGE
								where %PREFIX%CHANNEL_MESSAGE.GROUP_KEY = :group_key
								and %PREFIX%CHANNEL_MESSAGE.REGION = :region
							order by CREATED_DATE, MESSAGE_SEQUENCE
							limit 1 for update skip locked)
				returning MESSAGE_ID, MESSAGE_PAYLOAD_TYPE, MESSAGE_BYTES;
				""";
      }

    };
  }
```

I also updated my Flyway script that creates the tables to add the `MESSAGE_PAYLOAD_TYPE` column:

```
CREATE TABLE _spring_integration_CHANNEL_MESSAGE
(
    MESSAGE_ID           CHAR(36)     NOT NULL,
    GROUP_KEY            CHAR(36)     NOT NULL,
    CREATED_DATE         BIGINT       NOT NULL,
    MESSAGE_PRIORITY     BIGINT,
    MESSAGE_SEQUENCE     BIGINT       NOT NULL DEFAULT NEXTVAL('_spring_integration_MESSAGE_SEQ'),
    MESSAGE_PAYLOAD_TYPE VARCHAR      NOT NULL,
    MESSAGE_BYTES        JSON,
    REGION               VARCHAR(100) NOT NULL,
    CONSTRAINT _spring_integration_CHANNEL_MESSAGE_PK PRIMARY KEY (REGION, GROUP_KEY, CREATED_DATE, MESSAGE_SEQUENCE)
);
```

I wonder if this is the good approach to have (readable) JSON in the database instead of (unreadable) Java serialization in there? If so, maybe the documentation should be updated to provide a complete solution?</div>

yybmion:
Fixes : #9312 

**JSON serialization** as an alternative to Java serialization, enabling text-based message storage with type information preservation.

- Add `JacksonChannelMessageStorePreparedStatementSetter` for serialization
- Add `JacksonMessageRowMapper` for deserialization
- Include JSON-compatible schemas for **PostgreSQL (JSONB)**, **MySQL (JSON)**, and **H2 (CLOB)**
- Add comprehensive test coverage with database-specific tests
- Add documentation for JSON serialization support


yybmion:
Hi, I'd like your input on two decisions

### 1. Column naming: `MESSAGE_BYTES` or `MESSAGE_CONTENT`.

I kept `MESSAGE_BYTES` to avoid creating custom QueryProviders for each database. While semantically it should be `MESSAGE_CONTENT` (since it's now JSON, not binary), renaming would require:
- Custom `PostgresJsonChannelMessageStoreQueryProvider`
- Custom `MySqlJsonChannelMessageStoreQueryProvider`  
- Custom `H2JsonChannelMessageStoreQueryProvider`
- Maintaining separate query implementations

The current approach works as a drop-in replacement - users just change the schema file. Is this acceptable, or should I implement the custom QueryProviders for better naming?

### 2. Database coverage: PostgreSQL, MySQL, H2 only

I've implemented and tested these three databases. Should I:
- Add untested schemas for Oracle/SQL Server/Derby/DB2 based on documentation?
- Keep only tested databases and add others in follow-up PRs?

I lean toward starting with tested databases only. 

What do you think?

build.gradle
```diff
@@ -76,6 +76,7 @@ ext {
     jmsApiVersion = '3.1.0'
     jpaApiVersion = '3.2.0'
     jrubyVersion = '10.0.2.0'
+    jsonassertVersion = '1.5.1'
```
artembilan: Any real justification to manage an extra dependency?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,183 @@
...
+	 * <b>Package Matching:</b>
+	 * <ul>
+	 *   <li>{@code "com.example"} – trusts {@code com.example} and all subpackages
+	 *       (e.g., {@code com.example.MyClass}, {@code com.example.sub.AnotherClass})</li>
+	 *   <li>{@code "com.example.MyClass"} – trusts only the specific class</li>
+	 *   <li>{@code "*"} – trusts all packages (not recommended for production)</li>
+	 * </ul>
+	 * <p>
+	 * <b>Note:</b> Subpackages are automatically included without wildcards.
+	 *
```
artembilan: No blank lines in method Javadocs, please.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,183 @@
...
+	 * @param trustedPackages the additional packages to trust for deserialization
+	 */
+	public JacksonChannelMessageStorePreparedStatementSetter(String... trustedPackages) {
```
artembilan: The security is not consulted on the serialization, therefore this option is redundant here.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,183 @@
...
+ *     <li>MySQL: {@code JSON} </li>
+ *     <li>H2: {@code CLOB} </li>
+ * </ul>
+ * <p>
+ * <b>Note:</b> The standard Spring Integration schemas use {@code BLOB}/{@code BYTEA} columns
+ * which are not suitable for human-readable JSON storage. Use the provided JSON-specific
+ * schemas from {@code schema-*-json.sql} files.
+ *
+ * @author Yoobin Yoon
+ * @since 7.0
```
artembilan: Blank line after `@author` list.
Just to emphasize who exactly did the change.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
artembilan: This is something what is not used on serialization. Therefore this option is redundant.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonMessageRowMapper.java
```diff
@@ -0,0 +1,160 @@
...
+	 * @param rs the {@link ResultSet} to map
+	 * @param rowNum the current row number
+	 * @return the deserialized {@link Message}
+	 * @throws SQLException if an error occurs reading from the result set or
+	 * if JSON deserialization fails
+	 */
+	@Override
+	public Message<?> mapRow(ResultSet rs, int rowNum) throws SQLException {
+		try {
+			String json = rs.getString("MESSAGE_BYTES");
```
artembilan: Yes, I agree that it can be renamed to the `MESSAGE_CONTENT`, but do we really always going to have it as `String`?
Are you anticipating it as a `CLOB` in standard DBs and some custom type in specific, but that still would not be `byte[]`?

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,183 @@
...
+			if ("PostgreSQL".equalsIgnoreCase(dbProduct)) {
+				preparedStatement.setObject(6, json, Types.OTHER); // NOSONAR magic number
+			}
+			else {
+				preparedStatement.setString(6, json); // NOSONAR magic number
+			}
+		}
+		catch (JacksonException ex) {
+			throw new SQLException("Failed to serialize message to JSON. Message ID: " +
+					requestMessage.getHeaders().getId(), ex);
```
artembilan: Why just `id`?
Why the whole message is wrong to report here?
The `id` does not give us too much what exactly was the cause.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/MessageRowMapper.java
```diff
@@ -51,6 +54,19 @@ public MessageRowMapper(AllowListDeserializingConverter deserializer) {
...
+	/**
+	 * The default constructor for subclasses who are not interested in the message
+	 * deserialization from {@code byte[]} using {@link AllowListDeserializingConverter}.
+	 * The {@link #deserializer} is {@code null} from this constructor,
+	 * therefore any deserialization isn't happened in the default {@link #mapRow} implementation.
+	 * A target implementor must ensure the proper custom logic for reading message from the database.
+	 * @since 7.0
+	 */
+	protected MessageRowMapper() {
+		this.deserializer = null;
```
artembilan: I think this is architecturally is wrong decision.
That is really not this `MessageRowMapper` responsibility to expose a `null deserializer`.
It is really feels like a JSON on should be fully separated class as it is by its internals.
From here, we leave this class as is but change `JdbcChannelMessageStore` to accept an `RowMapper<Message<?>>` interface instead.

spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-h2-json.sql
```diff
@@ -0,0 +1,61 @@
+CREATE TABLE INT_MESSAGE  (
+    MESSAGE_ID    CHAR(36)     NOT NULL,
+    REGION        VARCHAR(100) NOT NULL,
+    CREATED_DATE  TIMESTAMP    NOT NULL,
+    MESSAGE_BYTES CLOB,
```
artembilan: I think we should drop all of these extra schema scripts and just document how to change the target RDBMS to support JSON serialization instead.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/H2JacksonChannelMessageStoreTests-context.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
```
artembilan: Do we really need XML configs to test the feature?
Why cannot we go with plain Java config instead?

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/PostgresJacksonChannelMessageStoreTests.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.transaction.PlatformTransactionManager;
+
+/**
+ * @author Yoobin Yoon
+ * @since 7.0
+ */
+@ContextConfiguration
+public class PostgresJacksonChannelMessageStoreTests extends AbstractJacksonChannelMessageStoreTests {
+
+	private static final PostgreSQLContainer<?> POSTGRES_CONTAINER =
```
artembilan: Why doesn't  `PostgresContainerTest` API work for this test scope?
Therefore, might be the case that this class has to go that `postgres` package in our tests.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/TestMailMessage.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Test payload class for Jackson JSON serialization tests.
+ * Represents a simple mail message.
+ *
+ * @author Yoobin Yoon
+ */
+public class TestMailMessage implements Serializable {
```
artembilan: If we talk about JSON, then `Serializable`  is something redundant.
I also wonder why this class is not a Java `record`.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -86,6 +86,10 @@ The xref:jdbc/dsl.adoc[] chapter provides more details.
 The `JdbcLock` now supports the feature of customized time-to-live for the lock status data.
 See xref:jdbc/lock-registry.adoc[] for more information.
 
+The `JdbcChannelMessageStore` now supports built-in Jackson JSON serialization as an alternative to Java serialization, enabling human-readable message storage.
+New components `JacksonChannelMessageStorePreparedStatementSetter` and `JacksonMessageRowMapper` provide this functionality with JSON-specific database schemas for PostgreSQL, MySQL, and H2.
```
artembilan: I don't think it has any specific value to mention those DBs. 
I believe that any CLOB-supporting DB is OK to have a JSON as its value.
So, your statement about `JSON-specific database schemas` does not reflect reality somehow...
Maybe you meant those custom schema files?
Then, OK: since we talk about dropping them in favor of docs, then this sentence would be changed as well.

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,213 @@
+[[jdbc-channel-message-store-json]]
+= JDBC Channel Message Store JSON Serialization
+
+Version 7.0 introduced Jackson JSON serialization support for `JdbcChannelMessageStore`.
+By default, Spring Integration uses Java serialization to store messages in the database as binary BLOB data, which is not human-readable.
```
artembilan: The "human-readable" statement makes me think twice.
Is it really a goal to let DBA to read the content of messages?
In most cases the `payload` and `headers` contain sensitive business data.
So, expose such an info DB directly might not be very secure.
May we at least don't emphasize somehow that it is better than `byte[]`?
From my point of view it is even dangerous.
I know that we can encode data in DBs, but let's at least don't mention that it is `human-readable`!

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,213 @@
...
+The new JSON serialization support allows messages to be stored as readable JSON, making debugging and troubleshooting significantly easier.
```
artembilan: Yes, and this sentence also drops any data protection rules away.

artembilan (changes requested):
Please, add your legal name to the `sign-off` of the commit.
You can change your Git client config on the matter.

yybmion:
Thanks for the feedback @artembilan!  Two questions:

> Yes, I agree that it can be renamed to the `MESSAGE_CONTENT`, but do we really always going to have it as `String`? Are you anticipating it as a `CLOB` in standard DBs and some custom type in specific, but that still would not be `byte[]`?

### 1. Column Naming

You mentioned `MESSAGE_CONTENT` is more semantic. Should I rename(**schema-(db).sql** file) it?

Is it okay with a breaking change for better naming? Or keep `MESSAGE_BYTES`?

### 2. Type-Safe Reading

Current `getString()` works for PostgreSQL/MySQL/H2. Do you mean we should add defensive handling for cases like:

- `java.sql.Clob` (Oracle, large data)
- `PGobject` (PostgreSQL specifics)
- `byte[]` fallback

artembilan:
I wonder if that could be kept as `MESSAGE_BYTES` and suggest any other name instead for JSON handling .
Might be the case that we would need to adjust our queries to avoid hard-coded name.
I believe with modern RDBMS drivers the `getString()` works well with CLOB.
We do have a Testcontainet for Oracle to check how JSON could be handled over there .

it is unfortunate that you are contributing this too late in the current 7.0 generation.
We do RC1 today which means we don’t have enough time to make your contribution as great. After that it is a code freeze and we won’t be able to rename column until version 8.0.
Either way I think `MESSAGE_BYTES` still makes sense and for JSON it could be anything else . We just need to double check the code for that naming usage.

artembilan (changes requested):
OK. I think we are on the right track, but still need dedicate more thinking on this.
I will aim this for the next `7.1`.
Meanwhile I'll make a small (well, it is going to be big by the amount of code affected 😉 ) breaking change for the today's `7.0.0-RC1` - the last chance where we can do that.
The change would be from `MESSAGE_BYTES` to `MESSAGE_CONTENT`.
This will be a good foundation for your work over here.

Thank you!

artembilan:
Here you are: https://github.com/spring-projects/spring-integration/pull/10524!

yybmion:
@artembilan Thanks for the detailed explanation and for taking on the naming change PR.

I’m updating the PR per your review.

> I think we should drop all of these extra schema scripts and just document how to change the target RDBMS to support JSON serialization instead.

I understand your point about simplifying the schemas and documenting JSON serialization setup instead.
However, for testing purposes, I still think having a dedicated table or script for verifying JSON results would be helpful.

What do you think about keeping a lightweight schema just for the tests (under `src/test/resources`)?

artembilan:
> What do you think about keeping a lightweight schema just for the tests (under src/test/resources)?

That totally makes sense.
And exactly this one could go to the doc as an example what exactly has to be done in the target RDBMS to support JSON.

yybmion:
Thanks for feedback @artembilan! 

I’ve updated the code based on the review for c1d0b32.

artembilan:
> I’ve updated the code based on the review

Thank you @yybmion !

But as I said before: we cannot accept a new code into the current `7.0.0` milestone since at this moment it is called "code freeze" .
The we would have one more month of `7.0.1` patch release fixing bugs.
And only after that we are going to switch to `7.1.0` where we can accept new features.

Does that make sense to you?

yybmion:
Thank you for the explanation @artembilan !

Yes, I completely understand the release schedule.

Please let me know if there are any updates or changes needed before the merge.

Thanks again for your guidance throughout the review process!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,133 @@
...
+	 * Create a new {@link JacksonChannelMessageStorePreparedStatementSetter} with the
+	 * default trusted packages from {@link JacksonMessagingUtils#DEFAULT_TRUSTED_PACKAGES}.
+	 * <p>
+	 * This constructor is suitable when you only need to serialize standard Spring Integration
+	 * and Java classes. Custom payload types will require their package to be added to the
+	 * corresponding {@link JacksonMessageRowMapper}.
+	 */
+	public JacksonChannelMessageStorePreparedStatementSetter() {
+		super();
+		this.objectMapper = JacksonMessagingUtils.messagingAwareMapper();
```
artembilan: I prefer calling other constructor instead.
This way we would have less code smells and more sure that we are on the right track.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonMessageRowMapper.java
```diff
@@ -0,0 +1,122 @@
...
+	 * Create a new {@link JacksonMessageRowMapper} with additional trusted packages
+	 * for deserialization.
+	 * <p>
+	 * The provided packages are appended to the default trusted packages from
+	 * {@link JacksonMessagingUtils#DEFAULT_TRUSTED_PACKAGES}, enabling deserialization
+	 * of custom payload types while maintaining security.
+	 * @param trustedPackages the additional packages to trust for deserialization
+	 */
+	public JacksonMessageRowMapper(String... trustedPackages) {
+		this.objectMapper = JacksonMessagingUtils.messagingAwareMapper(trustedPackages);
```
artembilan: DITTO

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,133 @@
...
+ * }
+ * }</pre>
+ *
+ * @author Yoobin Yoon
+ *
+ * @since 7.0
+ */
+public class JacksonChannelMessageStorePreparedStatementSetter extends ChannelMessageStorePreparedStatementSetter {
+
+	private final ObjectMapper objectMapper;
```
artembilan: Can we go our `JsonObjectMapper` API instead?
No one said that we are always going to work just with Jackson.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JacksonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * Create a new {@link JacksonChannelMessageStorePreparedStatementSetter} with the
+	 * default trusted packages from {@link JacksonMessagingUtils#DEFAULT_TRUSTED_PACKAGES}.
+	 * <p>
+	 * This constructor is suitable when you only need to serialize standard Spring Integration
+	 * and Java classes. Custom payload types will require their package to be added to the
+	 * corresponding {@link JacksonMessageRowMapper}.
+	 */
+	public JacksonChannelMessageStorePreparedStatementSetter() {
```
artembilan: If we go `JsonObjectMapper`, then the name of the class has to be changed to `JSON` as well instead of `Jackson`.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/H2JacksonChannelMessageStoreTests.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+import org.springframework.transaction.PlatformTransactionManager;
+
+/**
+ * @author Yoobin Yoon
+ *
+ * @since 7.0
+ */
+@SpringJUnitConfig
+@DirtiesContext
```
artembilan: You already have `@DirtiesContext` on the super class.
And I believe the first one also can go over there, too.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -90,6 +90,13 @@ The message stores now use a `MESSAGE_CONTENT` column name for serialized messag
 All the out-of-the-box SQL schemas have beed changed, too, to rely on the `MESSAGE_CONTENT` name for the respective column in the `INT_MESSAGE` and `INT_CHANNEL_MESSAGE` tables.
 See xref:jdbc/message-store.adoc[] for more information.
 
+The `JdbcChannelMessageStore` now supports Jackson JSON serialization as an alternative
+to Java serialization. New components `JacksonChannelMessageStorePreparedStatementSetter`
+and `JacksonMessageRowMapper` enable storing messages in JSON format.
+This requires modifying the database schema to use text-based column types
+(such as JSONB, JSON, TEXT, or CLOB) instead of binary types.
```
artembilan: Please, revise all your new documentation addition according to the `one sentence per line` rule: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line.
Right now it is hard to review.
Thanks

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,257 @@
+[[jdbc-channel-message-store-json]]
```
artembilan, line 1: Please, see `nav.adoc` file and add the link to this one just after `** xref:jdbc/message-store.adoc[]` line.

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,257 @@
...
+====
+**Security Consideration:** JSON serialization stores message content as text in the database,
+which may expose sensitive data. Ensure proper database access controls, encryption at rest,
+and consider your organization's data protection requirements before using JSON serialization
+in production environments.
+====
+
+== Configuration
+
+To enable JSON serialization, you need two components:
```
artembilan: Please, ensure that documentation is impersonated.
The human being pointers have to be avoided.
For, example this sentence could sound like this:
```
Two components are available for JSON (de)serialization:
```

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,257 @@
...
+        new JacksonChannelMessageStorePreparedStatementSetter());
+    store.setMessageRowMapper(
+        new JacksonMessageRowMapper("com.example"));
+
+    return store;
+}
+----
+
+The string parameter (`"com.example"`) specifies trusted packages for deserialization.
+Only classes from trusted packages can be deserialized for security.
```
artembilan, line 41: I think we need to mention what default packages are already there.
Which looks like you do below, so, please point to that section from here.

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,257 @@
...
+== Trusted Packages
+
+The `JacksonMessagingUtils.messagingAwareMapper()` validates all deserialized classes
+against a trusted package list to prevent security vulnerabilities.
+
+Default trusted packages include:
+- `java.util`
+- `java.lang`
+- `org.springframework.messaging.support`
+- `org.springframework.integration.support`
```
artembilan, line 220: Doesn't look like the full list.
Please, copy it from the `JacksonMessagingUtils.DEFAULT_TRUSTED_PACKAGES`

artembilan (changes requested):

yybmion:
Thanks for feedback @artembilan. 

Improved as requested in 1e0dfcb1dc5f5c21936851d9318c393156a131b0

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JsonMessageRowMapper.java
```diff
@@ -0,0 +1,129 @@
...
+
+	/**
+	 * Create a new {@link JsonMessageRowMapper} with the default {@link JsonObjectMapper}
+	 * configured for Spring Integration message deserialization.
+	 * <p>
+	 * This constructor is suitable when deserializing standard Spring Integration
+	 * and Java classes. Custom payload types will require their package to be
+	 * specified using {@link #JsonMessageRowMapper(String...)}.
+	 */
+	public JsonMessageRowMapper() {
```
artembilan: I think we don't need this ctor.
And another one could be like `String @Nullable ... trustedPackages`

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JsonChannelMessageStorePreparedStatementSetter.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.integration.support.json.JacksonMessagingUtils;
+import org.springframework.integration.support.json.JsonObjectMapper;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link ChannelMessageStorePreparedStatementSetter} implementation that uses JSON
+ * serialization for {@link Message} objects instead of Java serialization.
+ * <p>
+ * This implementation stores the entire message (including headers and payload) as JSON,
```
artembilan: We have to say `By default,` since there is no guarantee that provided `JsonObjectMapper` is always for Jackson.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/store/channel/JsonMessageRowMapper.java
```diff
@@ -0,0 +1,129 @@
...
+				throw new SQLException("MESSAGE_CONTENT column is null at row " + rowNum);
+			}
+
+			return this.jsonObjectMapper.fromJson(json, Message.class);
+		}
+		catch (IOException ex) {
+			throw new SQLException(
+					"Failed to deserialize message from JSON at row " + rowNum + ". "
+							+ "Ensure the JSON was created by JsonChannelMessageStorePreparedStatementSetter "
+							+ "and contains proper @class type information.",
```
artembilan: I don't think this is only a reason of an error.
Plus some other JSON serializers might not provide that `@class` information and can (de)serialize some other way.

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,262 @@
...
+   MESSAGE_CONTENT CLOB, -- JSON message content
+   REGION VARCHAR(100) NOT NULL,
+   CONSTRAINT JSON_CHANNEL_MESSAGE_PK
+      PRIMARY KEY (REGION, GROUP_KEY, CREATED_DATE, MESSAGE_SEQUENCE)
+);
+----
+
+== JSON Structure
+
+Messages are stored with the following JSON structure:
```
artembilan: I think we need to emphasize here that this is a result of Jackson polymorphic serialization.

src/reference/antora/modules/ROOT/pages/jdbc/message-store-json.adoc
```diff
@@ -0,0 +1,262 @@
...
+
+[IMPORTANT]
+====
+JSON serialization **requires modifying the database schema**.
+The default schema with BLOB/BYTEA column types cannot be used for JSON serialization.
+====
+
+The `MESSAGE_CONTENT` column must be changed to a text-based type that can store JSON.
+
+=== PostgreSQL
```
artembilan: Can we consider to make these 3 samples as tabs?
As well as the following for the whole table schema.

artembilan (changes requested):
Mostly this is great.
Just couple nit-picks and desired clean up.
Let me know if you can make until this end of the day when we are going to perform release.

Thanks

yybmion:
Sure @artembilan, I’ll address the nits today before release.

yybmion:
Thanks @artembilan , Improved as requested in a432da5f505de07d055cb8b31a421921dd922787

build.gradle
```diff
@@ -679,6 +679,7 @@ project('spring-integration-jdbc') {
     dependencies {
         api 'org.springframework:spring-jdbc'
         optionalApi "org.postgresql:postgresql:$postgresVersion"
+        optionalApi 'tools.jackson.core:jackson-databind'
```
artembilan, line 682: We need to move this to the `testImplementation` as with a `JsonObjectMapper` abstraction there is no strong Jackson requirement anymore.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/store/channel/TestMailMessage.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.springframework.integration.jdbc.store.channel;
+
+/**
+ * Test payload class for Jackson JSON serialization tests.
+ *
+ * @author Yoobin Yoon
+ */
+public record TestMailMessage(
```
artembilan, line 24: `@since 7.0` and not `public`.
I don't want test data classes to leak from one package to another.

artembilan (changes requested):

artembilan (approved):
OK. I'm taking this locally for final review, clean up and merge.

Thank you again!

yybmion:
>OK. I'm taking this locally for final review, clean up and merge.

Thanks @artembilan! 
Please let me know if there’s anything else I can help with.

artembilan:
Merged as https://github.com/spring-projects/spring-integration/commit/e2cf27c2e67754d68f476fbb7fc40a2c75580a9a after some fixes and clean up.
Pay attention, please, what I did with dependencies and your new classes to avoid classpath problems when Jackson is not available and we may chose some other option or different Jackson version.

Thank you for such a great contribution; looking forward for more!

PR #10524, opened 2025-10-21
JDBC message stores: Rename column to `MESSAGE_CONTENT`

artembilan:
The reason for this change is that content in the target DB might not always be a byte array. And we already have an outstanding work for providing JSON format for a message content.

Related to: https://github.com/spring-projects/spring-integration/issues/9312

This is mostly an automatic find-n-replace change, therefore not too much to review.


artembilan:
After merging this I also will add a Migration Guide note: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-6.x-to-7.0-Migration-Guide

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -86,6 +86,10 @@ The xref:jdbc/dsl.adoc[] chapter provides more details.
 The `JdbcLock` now supports the feature of customized time-to-live for the lock status data.
 See xref:jdbc/lock-registry.adoc[] for more information.
 
+The message stores now use a `MESSAGE_CONTENT` column name for serialized messages instead of `MESSAGE_BYTES` since the content might not always be stored as a byte array.
+All the out-of-the-box SQL schemas have beed changed, too, to rely on the `MESSAGE_CONTENT` name for the respective column in the `INT_MESSAGE` and `INT_CHANNEL_MESSAGE` tables.
+See xref:jdbc/message-store.adoc[] for more information.
```
cppwfs, line 91: Do we need to mention that they need to update existing schemas?

cppwfs (changes requested):
LGTM, just a quick question.

This does simplify things a quite a bit!

src/reference/antora/modules/ROOT/pages/whats-new.adoc
artembilan: Right. That is what a Migration Guide for.
I don't think the scope of the doc for specific version is a responsibility of such a guidance.

cppwfs (approved):
LGTM!

PR #10525, opened 2025-10-22
Fix Files DSL for auto-create input directory

artembilan:
Related to: https://github.com/spring-projects/spring-integration/issues/7971

The `FileReadingMessageSource` has always had a feature to create a directory to scan on startup. After introducing an expression variant, such a logic is limited only for the `ValueExpression` as only this one comes with a static value.
The Java DSL has been changed to support a `Supplier` variant for an input directory. The `Supplier` does not always give us the same value; therefore, we cannot assume directory auto-creation on startup.

* Fix `FileInboundChannelAdapterSpec` for the `directory(Expression directoryExpression)` instead of `Supplier`
* Fix `Files` to use a `ValueExpression` for statically provided `File directory`


cppwfs (approved):
LGTM.

Thanks for fixing this!

PR #10537, opened 2025-10-30
Update gradle to 9.2.0

cppwfs:
Attempted to resolve TODO on line 1099, however some of the API is still not recognized.  Added a 9.2+ to the TODO


artembilan (changes requested):
Thanks, but I would like to know what is that new problem. The SF has gotten the fix though: https://github.com/spring-projects/spring-framework/blob/main/framework-api/framework-api.gradle

PR #10573, opened 2025-11-13
Optimize `new ArrayList()` allocations whenever possible

artembilan:
To avoid resizing at runtime, it is better to use `ArrayList` with a predefined size. In some cases it is possible from the context of the list we are creating

* Fix typos in Javadocs of the affected classes according to IDE suggestions
* Use diamond operators for modern Java in the affected classes
* pattern matching expressions for `if` where IDE suggested
* Use `switch` in one place instead of `if..else`. Again: good suggestion from IDE
* Fix Nullability for the router hierarchy
* Have to add "redundant" cast in the `XPathRouter` to make NullAway happy

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/support/IntegrationMessageHandlerMethodFactory.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/support/IntegrationMessageHandlerMethodFactory.java
@@ -93,7 +93,7 @@ public class IntegrationMessageHandlerMethodFactory extends DefaultMessageHandle
 	}
 
 	private List<HandlerMethodArgumentResolver> buildArgumentResolvers(boolean listCapable) {
-		List<HandlerMethodArgumentResolver> resolvers = new ArrayList<>();
+		List<HandlerMethodArgumentResolver> resolvers = new ArrayList<>(6);
 		resolvers.add(new PayloadExpressionArgumentResolver());
 		resolvers.add(new NullAwarePayloadArgumentResolver(this.messageConverter));
 		resolvers.add(new PayloadsArgumentResolver());
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/AbstractIntegrationMessageBuilder.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/AbstractIntegrationMessageBuilder.java
@@ -70,7 +70,7 @@ public abstract class AbstractIntegrationMessageBuilder<T> {
 		List<List<Object>> incomingSequenceDetails = getSequenceDetails();
 		if (incomingCorrelationId != null) {
 			if (incomingSequenceDetails == null) {
-				incomingSequenceDetails = new ArrayList<>();
+				incomingSequenceDetails = new ArrayList<>(1);
 			}
 			else {
 				incomingSequenceDetails = new ArrayList<>(incomingSequenceDetails);
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileListFilterFactoryBean.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/config/FileListFilterFactoryBean.java
@@ -130,9 +130,9 @@ public class FileListFilterFactoryBean implements FactoryBean<FileListFilter<Fil
 	private FileListFilter<File> initializeFileListFilter() {
 		validate();
 
-		final List<FileListFilter<File>> filtersNeeded = new ArrayList<>();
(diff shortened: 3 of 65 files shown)


artembilan:
Let me know if this disturbanly long for review!
Thanks

spring-integration-core/src/main/java/org/springframework/integration/handler/support/IntegrationMessageHandlerMethodFactory.java
```diff
@@ -93,7 +93,7 @@ public InvocableHandlerMethod createInvocableHandlerMethod(Object bean, Method m
 	}
 
 	private List<HandlerMethodArgumentResolver> buildArgumentResolvers(boolean listCapable) {
-		List<HandlerMethodArgumentResolver> resolvers = new ArrayList<>();
+		List<HandlerMethodArgumentResolver> resolvers = new ArrayList<>(6);
```
cppwfs, line 96: I know this is being a bit picky but I think a constant would be better here than the number 6.

spring-integration-core/src/main/java/org/springframework/integration/support/AbstractIntegrationMessageBuilder.java
```diff
@@ -70,7 +70,7 @@ public AbstractIntegrationMessageBuilder<T> pushSequenceDetails(Object correlati
 		List<List<Object>> incomingSequenceDetails = getSequenceDetails();
 		if (incomingCorrelationId != null) {
 			if (incomingSequenceDetails == null) {
-				incomingSequenceDetails = new ArrayList<>();
+				incomingSequenceDetails = new ArrayList<>(1);
```
cppwfs, line 73: Can we use a constant here?

spring-integration-file/src/main/java/org/springframework/integration/file/config/FileListFilterFactoryBean.java
```diff
@@ -130,9 +130,9 @@ public boolean isSingleton() {
 	private FileListFilter<File> initializeFileListFilter() {
 		validate();
 
-		final List<FileListFilter<File>> filtersNeeded = new ArrayList<>();
+		final List<FileListFilter<File>> filtersNeeded = new ArrayList<>(3);
```
cppwfs, line 133: here as well :-)

spring-integration-http/src/main/java/org/springframework/integration/http/inbound/HttpRequestHandlingEndpointSupport.java
```diff
@@ -113,9 +113,9 @@
  */
 public abstract class HttpRequestHandlingEndpointSupport extends BaseHttpInboundEndpoint {
 
-	private final List<HttpMessageConverter<?>> defaultMessageConverters = new ArrayList<>();
+	private final List<HttpMessageConverter<?>> defaultMessageConverters = new ArrayList<>(10);
```
cppwfs, line 116: ditto

spring-integration-mongodb/src/main/java/org/springframework/integration/mongodb/store/MongoDbMessageStore.java
```diff
@@ -552,7 +552,8 @@ void setCustomConverters(Object @Nullable ... customConverters) {
 
 		@Override
 		public void afterPropertiesSet() {
-			List<Object> converters = new ArrayList<>();
+			List<Object> converters =
+					new ArrayList<>(7 + (this.customConverters != null ? this.customConverters.length : 0));
```
cppwfs, line 556: Use constant instead of 7.
I know Picky picky :-)

cppwfs (changes requested):
Fantastic!
Thank you for doing this.

Just some nitpicks.

artembilan:
Any reasons asking for those constants?
Logically they don’t make sense because most of those lists are local variable. And technically they won’t make difference at runtime.

cppwfs:
Why 6?  I read the code and I understand why.   But is there a constant we could use to add clarity?

artembilan:
Again: those lists are local variables.
The top level contract would make it more confusing.
If I feel a bit lost understanding the reason behind constant on the matter, then other code readers may suffer as well.
There are a lot of similar `new ArrayList(number)` in SF, and looks like no one complains.

Sorry, to say that but I wish you just bite this bullet and let the "magic number" a shot!
We may come back to constant subject when we got some real problem from community.

Any other reasons not just clarity?

cppwfs (approved):
LGTM

Thank you for the optimizations.   This is great!

PR #10575, opened 2025-11-15
Fix incorrect null check in CompositeExecutor constructor

yybmion:
Error in the `CompositeExecutor` constructor where `primaryTaskExecutor` was being validated twice instead of validating `secondaryTaskExecutor`.

```java
Assert.notNull(primaryTaskExecutor, "'primaryTaskExecutor' cannot be null");
Assert.notNull(primaryTaskExecutor, "'secondaryTaskExecutor' cannot be null");  // validated twice
```


artembilan:
@yybmion ,

thank you for the contribution ; looking forward for more!

PR #10600, opened 2025-11-19
Add nullChannel DSL element to Groovy DSL

cppwfs:
The Groovy DSL was missing support for the nullChannel, which is a special channel that discards messages (acts like /dev/null). This prevents users from using nullChannel in their Groovy integration flows.

Add unit tests to verify:
- nullChannel can be used at the end of a flow
- nullChannel can be used after transform operations
- Messages are properly discarded without exceptions

### DIFF
--- a/spring-integration-groovy/src/main/groovy/org/springframework/integration/groovy/dsl/GroovyIntegrationFlowDefinition.groovy
+++ b/spring-integration-groovy/src/main/groovy/org/springframework/integration/groovy/dsl/GroovyIntegrationFlowDefinition.groovy
@@ -108,6 +108,7 @@ class GroovyIntegrationFlowDefinition {
 	 * at the current {@link IntegrationFlow} chain position.
 	 * The nullChannel acts like "/dev/null".
 	 * @see org.springframework.integration.channel.NullChannel
+	 * @since 7.0.1
 	 */
 	GroovyIntegrationFlowDefinition nullChannel() {
 		this.delegate.nullChannel()
--- a/spring-integration-groovy/src/main/groovy/org/springframework/integration/groovy/dsl/GroovyIntegrationFlowDefinition.groovy
+++ b/spring-integration-groovy/src/main/groovy/org/springframework/integration/groovy/dsl/GroovyIntegrationFlowDefinition.groovy
@@ -107,8 +107,8 @@ class GroovyIntegrationFlowDefinition {
 	 * Populate a {@link org.springframework.integration.channel.NullChannel} instance
 	 * at the current {@link IntegrationFlow} chain position.
 	 * The nullChannel acts like "/dev/null".
-	 * @see org.springframework.integration.channel.NullChannel
 	 * @since 7.0.1
+	 * @see org.springframework.integration.channel.NullChannel
 	 */
 	GroovyIntegrationFlowDefinition nullChannel() {
 		this.delegate.nullChannel()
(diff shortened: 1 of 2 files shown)


spring-integration-groovy/src/main/groovy/org/springframework/integration/groovy/dsl/GroovyIntegrationFlowDefinition.groovy
```diff
@@ -103,6 +103,17 @@ class GroovyIntegrationFlowDefinition {
 		this
 	}
 
+	/**
+	 * Populate a {@link org.springframework.integration.channel.NullChannel} instance
+	 * at the current {@link IntegrationFlow} chain position.
+	 * The nullChannel acts like "/dev/null".
```
artembilan, line 109: `@since 7.0.1`

spring-integration-groovy/src/test/groovy/org/springframework/integration/groovy/dsl/test/GroovyDslTests.groovy
```diff
@@ -361,6 +405,27 @@ class GroovyDslTests {
 
 		}
 
+		@Bean
+		nullChannelFlow() {
+			integrationFlow {
+				{
+					log LoggingHandler.Level.WARN, 'test.category'
+				}
```
artembilan: I think this wrapping is redundant.
The `log` and `nullChannel` should nicely be aligned on the same level.
See that `transform` and `nullChannel` below.

spring-integration-groovy/src/test/groovy/org/springframework/integration/groovy/dsl/test/GroovyDslTests.groovy
```diff
@@ -220,6 +220,50 @@ class GroovyDslTests {
 		assert groovyTestService.result.get() == 'TEST'
 	}
 
+	@Autowired
+	@Qualifier('nullChannelFlow.input')
+	private MessageChannel nullChannelFlowInput
```
artembilan, line 225: Looks like this is out of use.
I wonder if your dynamic flow could just go into that bean definition instead and be used via this channel.

spring-integration-groovy/src/test/groovy/org/springframework/integration/groovy/dsl/test/GroovyDslTests.groovy
```diff
@@ -220,6 +220,50 @@ class GroovyDslTests {
...
+		def tappedMessage = wireTapChannel.receive(1000)
+		registration.destroy()
+
+		assert tappedMessage?.payload == 'test' //verify that a message was sent and assume nullChannel discarded it.
+
+	}
+
+	@Autowired
+	@Qualifier('nullChannelWithTransformFlow.input')
+	private MessageChannel nullChannelWithTransformFlowInput
```
artembilan, line 242: DITTO

artembilan (changes requested):
Almost there! 😄

spring-integration-groovy/src/main/groovy/org/springframework/integration/groovy/dsl/GroovyIntegrationFlowDefinition.groovy
```diff
@@ -103,6 +103,18 @@ class GroovyIntegrationFlowDefinition {
...
+	 * @see org.springframework.integration.channel.NullChannel
+	 * @since 7.0.1
```
artembilan, line 110: Well, I placed request in the specific place.
See an order of the Javadoc tags in our Checkstyle config:
```
<module name="AtclauseOrder">
			<property name="target" value="METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
			<property name="tagOrder" value="@param, @return, @throws, @since, @deprecated, @see"/>
		</module>
```

artembilan (changes requested):
On last nit-pick

PR #10601, opened 2025-11-21
Add `nullChannel()` to Kotlin DSL

cppwfs:
This commit adds support for `nullChannel()` in the Kotlin DSL, mirroring the functionality recently added to the Groovy DSL. The `nullChannel()` method populates a NullChannel instance at the current IntegrationFlow chain position, acting like "/dev/null" to discard messages.

Two test cases verify the functionality:
- Messages sent to a flow ending with `nullChannel()` are discarded
- Transform operations can precede `nullChannel()` in a flow

This enhancement provides feature parity between the Kotlin and Groovy DSLs for null channel support.

### DIFF
--- a/spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
+++ b/spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
@@ -368,7 +368,7 @@ class KotlinDslTests : TestApplicationContextAware {
 				transform<String> { it.uppercase() }
 			}
 
-		
+
 		/*
 		A Java variant for the flow below
 		@Bean
--- a/spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
+++ b/spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
@@ -367,6 +367,8 @@ class KotlinDslTests : TestApplicationContextAware {
 				}
 				transform<String> { it.uppercase() }
 			}
+
+		
 		/*
 		A Java variant for the flow below
 		@Bean
@@ -386,7 +388,6 @@ class KotlinDslTests : TestApplicationContextAware {
 							scatterGather -> scatterGather
 								.gatherTimeout(10_000));
 		}*/
-
 		@Bean
 		fun scatterGatherFlow() =
 			integrationFlow {
--- a/spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
+++ b/spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
@@ -56,6 +56,7 @@ import java.util.function.Function
 
 /**
  * @author Artem Bilan
+ * @author Glenn Renfro
  */
 @SpringJUnitConfig
 @DirtiesContext
@@ -239,6 +240,35 @@ class KotlinDslTests : TestApplicationContextAware {
 		registration.destroy()
 	}
 
+	@Autowired
+	@Qualifier("nullChannelFlow.input")
+	private lateinit var nullChannelFlowInput: MessageChannel
+
+	@Autowired
+	private lateinit var nullCheckWireTapChannel: QueueChannel
+
+
+	@Test
+	fun `nullChannelFlow discards messages`() {
+		nullChannelFlowInput.send(MessageBuilder.withPayload("test").build())
+		val tappedMessage = nullCheckWireTapChannel.receive(1000)
+
+		//verify that a message was sent and assume nullChannel discarded it.
+		assertThat(tappedMessage?.payload).isNotNull().isEqualTo("test")
+	}
+
+	@Autowired
+	@Qualifier("nullChannelWithTransformFlow.input")
+	private lateinit var nullChannelWithTransformFlowInput: MessageChannel
+
+	@Test
+	fun `nullChannel can be used after transform in a flow`() {
+		nullChannelWithTransformFlowInput.send(MessageBuilder.withPayload("test").build())
+		val tappedMessage = nullCheckWireTapChannel.receive(1000)
+		//verify that a message was sent, transformed and assume nullChannel discarded it.
+		assertThat(tappedMessage?.payload).isNotNull().isEqualTo("TEST")
+	}
+
 	@Configuration
 	@EnableIntegration
 	class Config {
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -386,6 +416,27 @@ class KotlinDslTests : TestApplicationContextAware {
...
 			}
 
+		@Bean
+		fun nullCheckWireTapChannel() = QueueChannel()
+
+		@Bean
+		fun nullChannelFlow(@Qualifier("nullCheckWireTapChannel") nullCheckWireTapChannel: MessageChannel) =
+			integrationFlow {
+				wireTap(nullCheckWireTapChannel)
+				nullChannel()
```
artembilan: I think we don't need this bean and respective test.
Looks like the another one covers this functionality pretty well.

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -386,6 +416,27 @@ class KotlinDslTests : TestApplicationContextAware {
...
+			}
+
+		@Bean
+		fun nullChannelWithTransformFlow(
+			@Qualifier("nullCheckWireTapChannel") nullCheckWireTapChannel: MessageChannel) =
+			integrationFlow {
+				transformWith {
```
artembilan: Looks like simple `inline fun <reified P : Any> transform(crossinline function: (P) -> Any) {` would do exactly what we have here so far.
That `id` is out of use anyway.

artembilan (changes requested):
Would you mind fixing also that sample in the `kotlin-dsl.adoc` removing `MessageProcessorMessageSource`?

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
artembilan: OK. 
I mean just this instead:
```
transform<String>{ it.uppercase() }
```
In other words:
```
transform<String>{ it.uppercase() }
wireTap(nullCheckWireTapChannel)
nullChannel()
```

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -376,26 +366,6 @@ class KotlinDslTests : TestApplicationContextAware {
 				transform<String> { it.uppercase() }
 			}
 
-
-		/*
```
artembilan, line 350: Please, revert.
This is important to keep for comparison.

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -171,7 +177,6 @@ class KotlinDslTests : TestApplicationContextAware {
 			integrationFlow(publisher) {
 				transformWith {
 					transformer<Message<Int>> { it.payload * 2 }
-					id("foo")
```
artembilan, line 173: I know `foo` is bad but that is not what we want to do in this test configuration removing `id`.
The point is to have coverage for API.
Removing it everywhere was not a request.
If you worry about coverage more, then we might modify respective test to check for presence of bean with that name.
Thanks

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -340,7 +331,6 @@ class KotlinDslTests : TestApplicationContextAware {
 				log<Any>(LoggingHandler.Level.WARN) { it.payload }
 				transformWith {
 					expression("payload")
-					id("spelTransformer")
```
artembilan, line 313: DITTO here: no removal.
My request was to use just `transform<String>{ it.uppercase() }` in your new bean, not `id` removal.

artembilan (changes requested):
Any chances for the doc fix we have discussed before?

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -345,26 +367,24 @@ class KotlinDslTests : TestApplicationContextAware {
...
+			.recipientFlow(m -> true, sf -> sf.handle((p, h) -> Math.random() * 10))
+			.recipientFlow(m -> true, sf -> sf.handle((p, h) -> Math.random() * 10)),
+			gatherer -> gatherer
+			.releaseStrategy(group ->
+			group.size() == 3 ||
+					group.getMessages()
+						.stream()
+						.anyMatch(m -> (Double) m.getPayload() > 5)),
+			scatterGather -> scatterGather
+			.gatherTimeout(10_000));
```
artembilan: Something is off with reformatting this code snippet: it became unreadable.

artembilan (changes requested):
Fix for the `kotlin-dsl.adoc`, please?

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -345,8 +367,6 @@ class KotlinDslTests : TestApplicationContextAware {
 				}
 				transform<String> { it.uppercase() }
 			}
-
-
```
artembilan, line 349: Sorry to bother you too much for this comment, but it must not stick to the previous code.

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -366,6 +386,7 @@ class KotlinDslTests : TestApplicationContextAware {
 							scatterGather -> scatterGather
 								.gatherTimeout(10_000));
 		}*/
+
```
artembilan: However, it must stick to the bean it is related to.

artembilan (changes requested):

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
```diff
@@ -346,7 +368,7 @@ class KotlinDslTests : TestApplicationContextAware {
 				transform<String> { it.uppercase() }
 			}
 
-
+
```
artembilan: We probably don't have Checkstyle running against Kotlin code but, please, confirm that there are no trailing spaces in this line.
You can configure your IDEA to remove them on file save.

artembilan (changes requested):

spring-integration-core/src/test/kotlin/org/springframework/integration/dsl/KotlinDslTests.kt
cppwfs: Intellij updated

PR #10617, opened 2025-11-30
Update H2 DDL for JDBC Lock Registry lock table

agebhar1:


cppwfs:
Thank you for your contribution!   Keep up the great work!

PR #10625, opened 2025-12-12
Fix `transformer` for `async` mode

artembilan:
The `AbstractMessageProducingHandler` has `async` mode, which is handled by the reply payload type.
However, the `AbstractTransformer` produces the whole `Message` with that async reply payload.

* Fix `AbstractMessageProducingHandler` to check for a `Message` type on a reply object. Extract its payload for the async logic.
And recreate the final message for reply after async container fulfilment
* Disable `WebFluxDslTests.testHttpReactivePostWithError()` as failing against the latest SF


cppwfs (approved):
LGTM

Nice Catch!

PR #10647, opened 2025-12-16
GH-10624: PartitionedChannel: customize worker queue size

### LINKED ISSUE #10624: PartitioningChannel: Allo to customize the Executor used for each Partition
**Expected Behavior**

I want to block on work being dispatched while the thread for the partition is busy.

**Current Behavior**

The executors used for each partition have an unbounded queue, and thus, it doesn't block.

**Context**

- Possible OOM
- Don't want to pull elements from the persisted message Queue into memory too early.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10624

* Expose `workerQueueSize` from the `PartitionedDispatcher` and respective delegates from the `PartitionedChannel` and `PartitionedChannelSpec`


cppwfs (approved):
LGTM. 

Nice feature add!

Do we need to update any docs for setting queue size

PR #10667, opened 2025-12-17
GH-10623: Add `GrpcInboundGateway`

### LINKED ISSUE #10623: Introduce `GrpcInboundGateway`

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10623

* Add `spring-integration-grpc` module
* Manage respective new `io.grpc` dependencies
* Add Checkstyle suppression for new Proto-generated classes
* Add `GrpcHeaders` for convenient API
* Implement `GrpcInboundGateway` with a proxy for gRPC service methods
* Add `TestInProcessConfiguration` for a common in-process infrastructure
* Document new feature

### DIFF
--- a/spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
+++ b/spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
@@ -22,6 +22,8 @@ import java.util.Arrays;
 import io.grpc.BindableService;
 import io.grpc.MethodDescriptor;
 import io.grpc.ServerServiceDefinition;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import io.grpc.stub.StreamObserver;
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
@@ -152,7 +154,9 @@ public class GrpcInboundGateway extends MessagingGatewaySupport implements Binda
 			StreamObserver<Object> responseObserver) {
 
 		sendRequestAndProduceReply(methodDescriptor, requestPayload)
-				.subscribe(responseObserver::onNext, responseObserver::onError, responseObserver::onCompleted);
+				.subscribe(responseObserver::onNext,
+						t -> responseObserver.onError(toGrpcStatusException(t)),
+						responseObserver::onCompleted);
 	}
 
 	private void serverStreaming(MethodDescriptor<?, ?> methodDescriptor, Object requestPayload,
@@ -160,7 +164,9 @@ public class GrpcInboundGateway extends MessagingGatewaySupport implements Binda
 
 		sendRequestAndProduceReply(methodDescriptor, requestPayload)
 				.flatMapMany(payload -> payload instanceof Flux<?> flux ? flux : Flux.just(payload))
-				.subscribe(responseObserver::onNext, responseObserver::onError, responseObserver::onCompleted);
+				.subscribe(responseObserver::onNext,
+						t -> responseObserver.onError(toGrpcStatusException(t)),
+						responseObserver::onCompleted);
 	}
 
 	private StreamObserver<?> clientStreaming(MethodDescriptor<?, ?> methodDescriptor,
@@ -177,15 +183,15 @@ public class GrpcInboundGateway extends MessagingGatewaySupport implements Binda
 
 			@Override
 			public void onError(Throwable t) {
-				throw new IllegalStateException(
-						"gRPC request [" + methodDescriptor.getFullMethodName() + "] has failed", t);
+				throw toGrpcStatusException(t, "gRPC request [" + methodDescriptor.getFullMethodName() + "] has failed");
 			}
 
 			@Override
 			public void onCompleted() {
 				requestPayload.tryEmitComplete();
 				sendRequestAndProduceReply(methodDescriptor, requestPayload.asFlux())
-						.subscribe(responseObserver::onNext, responseObserver::onError,
+						.subscribe(responseObserver::onNext,
+								t -> responseObserver.onError(toGrpcStatusException(t)),
 								responseObserver::onCompleted);
 			}
 
@@ -200,13 +206,13 @@ public class GrpcInboundGateway extends MessagingGatewaySupport implements Binda
 			@Override
 			public void onNext(Object value) {
 				sendRequestAndProduceReply(methodDescriptor, value)
-						.subscribe(responseObserver::onNext, responseObserver::onError);
+						.subscribe(responseObserver::onNext,
+								t -> responseObserver.onError(toGrpcStatusException(t)));
 			}
 
 			@Override
 			public void onError(Throwable t) {
-				throw new IllegalStateException(
-						"gRPC request [" + methodDescriptor.getFullMethodName() + "] has failed", t);
+				throw toGrpcStatusException(t, "gRPC request [" + methodDescriptor.getFullMethodName() + "] has failed");
 			}
 
 			@Override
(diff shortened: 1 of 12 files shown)


spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
```diff
@@ -0,0 +1,234 @@
...
+		return new StreamObserver<>() {
+
+			@Override
+			public void onNext(Object value) {
+				requestPayload.tryEmitNext(value);
+			}
+
+			@Override
+			public void onError(Throwable t) {
+				throw new IllegalStateException(
```
cppwfs: Let's dig into this.  OnError is called to indicate there is a problem, and thus is the last thing called.  At least my understanding of the java docs https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html

I don't think we want to throw an exception here, but rather log the error using the Status class.  
```
Status status = Status.fromThrowable(t);
    
// 1. Log the error instead of throwing
logger.error("Stream failed with status: " + status.getCode(), t);

```

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/TestInProcessConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	@EventListener(ContextRefreshedEvent.class)
+	void startServer() throws IOException {
+		this.server = this.serverBuilder.build().start();
+	}
+
+	@Override
+	public void destroy() {
+		this.server.shutdown();
```
cppwfs: The `shutdown` method  signals the server to shutdown but it is still running till it finishes all its tasks.   We may want to have a `this.server.awaitTermination(5, TimeUnit.SECONDS);` or `this.server.awaitTermination()` to make sure the server is down before proceeding.   

You can also use `shutdownNow()` but you still may want to use `awaitTermination` after.

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/inbound/GrpcInboundGatewayTests.java
```diff
@@ -0,0 +1,250 @@
...
+		return new StreamObserver<>() {
+
+			@Override
+			public void onNext(R value) {
+				replyConsumer.accept(value);
+			}
+
+			@Override
+			public void onError(Throwable t) {
+
```
cppwfs, line 180: Should we test an error condition such that the onError is called?

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -0,0 +1,154 @@
...
+NOTE: The `GrpcHeaders.SERVICE_METHOD` header on the inbound gateway has a value of the gRPC service method name exactly as it is declared in the Protobuf (see `.proto` example above) and how it is stored into the `io.grpc.MethodDescriptor` of the service definition.
+
+[[grpc-inbound-gateway]]
+== Inbound Gateway for gRPC
+
+The `GrpcInboundGateway` is a `MessagingGatewaySupport` implementation to receive gRPC requests, send messages downstream flow and produce gRPC responses.
+For initialization, the instance of this gateway requires only an abstract gRPC service class implementing `BindableService`, usually generated from Protobuf and comes with a `*ImplBase` class name.
+
+WARNING: Only standard gRPC services are supported: a generated `AsyncService` contract is what `GrpcInboundGateway` logic based on.
+The Reactor and Kotlin-based service generation don't make sense in Spring Integration logic since those types are not exposed anyhow from the gateway definition.
```
cppwfs: Removing the word `anyhow` will strengthen the sentence.

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -0,0 +1,154 @@
...
+[source, java]
+----
+@Bean
+GrpcInboundGateway helloWorldService() {
+    return new GrpcInboundGateway(TestHelloWorldGrpc.TestHelloWorldImplBase.class);
+}
+----
+
+The `GrpcInboundGateway` implements a `BindableService` and exposes a `ServerServiceDefinition` based on the mentioned proxy for an `AsyncService` contract of the gRPC service.
+Therefore, exactly instance of this gateway has to be registered into a `ServerBuilder` and no need in any other `*ImplBase` implementations in the application.
```
cppwfs: I think this change will make the sentence a bit clearer:
_Therefore, exactly `one` instance._

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+IMPORTANT: With https://spring.io/projects/spring-grpc[Spring gRPC] and its auto-discovery for `BindableService` implementations, the `GrpcInboundGateway` has to be declared as a top-level bean.
+Therefore, Java DSL API like `IntegrationFlow.from(new GrpcInboundGateway(TestHelloWorldGrpc.TestHelloWorldImplBase.class))` is not recommended because such a `BindableService` implementation won't make it visible for respective Spring gRPC infrastructure.
+
+The `GrpcInboundGateway` uses a `sendAndReceiveMessageReactive()` API to interact with the downstream flow and adapts a `Mono` reply to the gRPC `StreamObserver`.
+As mentioned before, the request message payload is exactly a gRPC request message, and it expects a reply in a from of gRPC response message.
+The downstream logic could be type-safe and deal with gRPC messages similar way as if `*ImplBase` would be implemented manually.
```
cppwfs: Rewrite this line as:
`The downstream logic can be type-safe and deal with gRPC messages in a similar way as if *ImplBase were implemented manually.`

WDYT?

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+The `MethodDescriptor.MethodType.UNARY` and `MethodDescriptor.MethodType.BIDI_STREAMING` are same from the downstream handling logic perspective.
+In other words, the `BIDI_STREAMING` is handled as a loop on request items and the gateway produces a response item for each of them.
```
cppwfs: End the sentence as:
`response item for each.`

WDYT

cppwfs (changes requested):
Wow! this looks great! 

Just my usual nitpicks.

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
cppwfs: Changed my mind here...   I dug a little deeper and throwing an exception is ok here.   We just may want to take advantage of the Status class.

src/reference/antora/modules/ROOT/pages/grpc.adoc
artembilan: I think your sentence would lead into a confusion that only one gateway instance could be used in gRPC server.
And that is not correct: we may have as may as possible, just like with regular gRPC application.
Either way, I'll revise this sentence.
Thanks

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/TestInProcessConfiguration.java
artembilan: With this my server implementation it does not matter if we still have running tasks after closing this test suite.
But anyway, I'll use `shutdownNow()`.
Although, I still don't see a reason in the `awaitTermination()`.

src/reference/antora/modules/ROOT/pages/grpc.adoc
cppwfs: Thank you!

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/TestInProcessConfiguration.java
cppwfs: I was reading the javadocs and they mentioned using it.   I don't think its that important, but I wanted to mention it.

cppwfs (approved):
Looks great!

Thank you for all the changes!

PR #10668, opened 2025-12-19
GH-10574: Rely on `WebServiceTemplate.defaultUri`

### LINKED ISSUE #10574: `spring-integration-ws` WS DSL requires setting URI when WebServiceTemplate already has default URI set

7.0.0.RC2

**Describe the bug**

`Ws.marshallingOutboundGateway` will not start without specifying the `.uri()` in the DSL, even though the supplied `WebServiceTemplate` has a `defaultUri` set

**To Reproduce**

1. Create WebService Marshalling Outbound Gateway, specifying only the WebServiceTemplate
```java
return f -> f.handle(WS.marshallingOutboundGateway(webServiceTemplate);
```

Application fails to start, indicating the Ws.marshallingOutboundGateway requires the URI to be set

**Expected behavior**

`Ws.marshallingOutboundGateway()` should be able to infer the uri from the supplied `WebServiceTemplate` if the defaultUri has been set.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10574

* Change the logic of the `AbstractWebServiceOutboundGateway` to take into account a `WebServiceTemplate.defaultUri` option as an alternative to the explicit `uri` or `destinationProvider`
* Expose `WebServiceTemplate`-based ctor in the `SimpleWebServiceOutboundGateway` (The `MarshallingWebServiceOutboundGateway` already has one)
* Adjust tests and XML parser to cover a new logic
* Mention the change in docs

### DIFF
--- a/spring-integration-ws/src/main/java/org/springframework/integration/ws/outbound/AbstractWebServiceOutboundGateway.java
+++ b/spring-integration-ws/src/main/java/org/springframework/integration/ws/outbound/AbstractWebServiceOutboundGateway.java
@@ -69,13 +69,13 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 
 	private final Lock lock = new ReentrantLock();
 
-	protected final DefaultUriBuilderFactory uriFactory = new DefaultUriBuilderFactory(); // NOSONAR - final
-
-	private final @Nullable String uri;
+	protected final DefaultUriBuilderFactory uriFactory = new DefaultUriBuilderFactory();
 
 	private final @Nullable DestinationProvider destinationProvider;
 
 	private final Map<String, Expression> uriVariableExpressions = new HashMap<>();
+
+	private @Nullable String uri;
 
 	@SuppressWarnings("NullAway.Init")
 	private StandardEvaluationContext evaluationContext;
@@ -90,8 +90,9 @@ public abstract class AbstractWebServiceOutboundGateway extends AbstractReplyPro
 
 	private boolean webServiceTemplateExplicitlySet;
 
-	public AbstractWebServiceOutboundGateway(@Nullable final String uri, @Nullable WebServiceMessageFactory messageFactory) {
-		Assert.hasText(uri, "URI must not be empty");
+	public AbstractWebServiceOutboundGateway(@Nullable final String uri,
+			@Nullable WebServiceMessageFactory messageFactory) {
+
 		this.webServiceTemplate = messageFactory != null ?
 				new WebServiceTemplate(messageFactory) : new WebServiceTemplate();
 		this.destinationProvider = null;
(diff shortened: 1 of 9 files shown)


spring-integration-ws/src/main/java/org/springframework/integration/ws/outbound/AbstractWebServiceOutboundGateway.java
```diff
@@ -106,11 +107,16 @@ public AbstractWebServiceOutboundGateway(DestinationProvider destinationProvider
...
 		// we always call WebServiceTemplate methods with an explicit URI argument,
-		// but in case the WebServiceTemplate is accessed directly we'll set this:
+		// but in case the WebServiceTemplate is accessed directly, we'll set this:
 		this.webServiceTemplate.setDestinationProvider(destinationProvider);
 		this.uri = null;
 	}
 
+	public AbstractWebServiceOutboundGateway(WebServiceTemplate webServiceTemplate) {
+		doSetWebServiceTemplate(webServiceTemplate);
+		this.destinationProvider = null;
```
cppwfs, line 117: Curious as to why we are setting the `destinationProvider` to `null`, since it is default is `null`.

src/reference/antora/modules/ROOT/pages/ws.adoc
```diff
@@ -40,6 +40,7 @@ To invoke a web service when you send a message to a channel, you have two optio
 The former accepts either a `String` or `javax.xml.transform.Source` as the message payload.
 The latter supports any implementation of the `Marshaller` and `Unmarshaller` interfaces.
 Both require a Spring Web Services `DestinationProvider`, to determine the URI of the web service to be called.
+Or explicit uri directly, or via `WebServiceTemplate.defaultUri` (since version 7.1).
```
cppwfs: I think we can replace the period with another comma:   For example
Both require a Spring Web Services `DestinationProvider` to determine the URI of the web service to be called, or an explicit URI can be provided directly or via `WebServiceTemplate.defaultUri` (since version 7.1).

cppwfs (approved):
This is a great addition.   Just my annoying questions and nitpicks!

spring-integration-ws/src/main/java/org/springframework/integration/ws/outbound/AbstractWebServiceOutboundGateway.java
artembilan: Because it is `final`.

cppwfs: DOh!   Yep!

cppwfs (approved):
Thank you for the update!

LGTM!

PR #10677, opened 2025-12-28
Fix writingLatch lifecycle in TcpNioConnection

ouharri:
This PR fixes a lost-notify race in TcpNioConnection where writingLatch could be replaced from
ChannelInputStream.write(), causing the assembler to wait on a stale latch under concurrency.

Changes:
- Do not replace writingLatch from the pipe write path
- Clear writingLatch at the end of the doRead() cycle
- Add a regression test proving the latch is not replaced during pipe write and is cleared after read

Fixes #10671


artembilan:
@ouharri ,

PR #10683, opened 2026-01-07
GH-10666: TcpNio: Restore RejectedExecutionException propagation

### LINKED ISSUE #10666: Closed a connection when occurred RejectedExecutionException on TcpNioConnection#checkForAssembler

6.4.9

**Describe the bug**

Closed a connection when occurred the `RejectedExecutionException` on `TcpNioConnection#checkForAssembler`.
Because It catch at `readPacket` method as follow:

```java
	public void readPacket() {
			logger.debug(getConnectionId() + " Reading...");
		}
		try {
			doRead();
		}
		catch (@SuppressWarnings(UNUSED) ClosedChannelException cce) {
				logger.debug(getConnectionId() + " Channel is closed");
			}
			closeConnection(true);
		}
		catch (Exception e) {
			logger.error("Exception on Read " + getConnectionId() + " " + e.getMessage(), e);
			closeConnection(true); // ★★★★★ close connection this code
		}
	}
```

On the other hand, `AbstractConnectionFactory` is implemented assuming that a `RejectedExecutionException` will be thrown from `readPacket` method.

```java
				boolean delayed = false;
				try {
					connection.readPacket();
				}
				catch (@SuppressWarnings(UNUSED) RejectedExecutionException e1) { // ★★★★ Current implementation is dead code
					delayRead(selector, now, key);
					delayed = true;
				}
```

**Expected behavior**

Instead of closing the connection, it will be delayed Read.
How about modifying it as follows?

```diff
	public void readPacket() {
		if (logger.isDebugEnabled()) {
			logger.debug(getConnectionId() + " Reading...");
		}
		try {
			doRead();
		}
		catch (@SuppressWarnings(UNUSED) ClosedChannelException cce) {
			if (logger.isDebugEnabled()) {
				logger.debug(getConnectionId() + " Channel is closed");
			}
			closeConnection(true);
		}
+		catch (RejectedExecutionException ree) {
+			throw ree;
+		}
		catch (Exception e) {
			logger.error("Exception on Read " + getConnectionId() + " " + e.getMessage(), e);
			closeConnection(true);
		}
	}
```

The behavior seems to have changed in the following commit:

https://github.com/spring-projects/spring-integration/commit/19b9944dd89a9cc96c1d0b4baa6a81532a715372#diff-6451fd5c22e08caea6f3005257a2c56b99adfac27424810b8375d8848a73b61b

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10666

After some refactoring in the past, the `TcpNioConnection.readPacket()` does not re-throw a `RejectedExecutionException` when no sufficient resource to handle a new task in the `Executor`.
This leads to connection closing instead of delaying reads as it is suggested by the `AbstractConnectionFactory`.
Note: the `AbstractConnectionFactory.keyReadable()` still handles a thrown from the task `RejectedExecutionException` but this is a dead code since `TcpNioConnection.readPacket()` does not propagate one.

* Fix `TcpNioConnection.readPacket()` to catch and re-throw `RejectedExecutionException` as it was before the mentioned refactoring instead of falling down to a common `catch` block when this connection is closed.
* Add a mock `TcpNioConnectionTests.delayedReadsDueToRejectedExecutionException()` to ensure that `AbstractConnectionFactory.delayRead()` is called when not enough resources in the provided `Executor`.

### DIFF
--- a/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionTests.java
+++ b/spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionTests.java
@@ -44,6 +44,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.ThreadPoolExecutor.AbortPolicy;
@@ -871,13 +872,29 @@ public class TcpNioConnectionTests implements TestApplicationContextAware {
 		};
 
 		ThreadPoolExecutor threadPoolExecutor =
-				new ThreadPoolExecutor(1, 1, 0, TimeUnit.MILLISECONDS, new SynchronousQueue<>());
+				new ThreadPoolExecutor(1, 2, 0, TimeUnit.MILLISECONDS, new SynchronousQueue<>());
 		factory.setTaskExecutor(threadPoolExecutor);
 		factory.start();
 		SocketChannel chan1 = mock();
-		TcpNioConnection conn1 = mock();
-		Map<SocketChannel, TcpNioConnection> connections = Map.of(chan1, conn1);
 		willReturn(true).given(chan1).isOpen();
+		when(chan1.socket()).thenReturn(mock());
+		TcpNioConnection conn1 = new TcpNioConnection(chan1, false, false, null, null);
+		conn1.setTaskExecutor(threadPoolExecutor);
+		conn1 = spy(conn1);
+		AtomicReference<RejectedExecutionException> reeReference = new AtomicReference<>();
+		doAnswer(invocation -> {
+			try {
+				return invocation.callRealMethod();
+			}
+			catch (RejectedExecutionException ree) {
+				reeReference.set(ree);
+				throw ree;
+			}
+		})
+				.when(conn1)
+				.readPacket();
+
+		Map<SocketChannel, TcpNioConnection> connections = Map.of(chan1, conn1);
 		SelectionKey key1 = mock();
 		willReturn(true).given(key1).isReadable();
 		willReturn(true).given(key1).isValid();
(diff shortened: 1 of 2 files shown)


spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionTests.java
```diff
@@ -854,6 +857,40 @@ public void testWritingLatchClearedAfterRead() throws Exception {
 		cleanupCompositeExecutor(compositeExec);
 	}
 
+	@Test
+	public void delayedReadsDueToRejectedExecutionException() throws InterruptedException {
```
cppwfs, line 862: The test does not execute the new added code above.  But I could be wrong, can you verify?

cppwfs (changes requested):
Thanks for fixing the bug.   I had one question, once answered I'll merge.

spring-integration-ip/src/test/java/org/springframework/integration/ip/tcp/connection/TcpNioConnectionTests.java
artembilan: Yeah… I understand your confusion.
The flow is a bit complicated and fully async.
That selection in connection factory schedule a task for the connection to read. That connection uses the same executor to schedule an assembler. Once it cannot be scheduled due to not sufficient resources in the executor, the exception in the subject is thrown. Therefore I verify here a reaction to that exception for deliberately insufficient executor. Before the fix the code was dead. See issue and commit message for details.

cppwfs: I ask because when I debugged the code I expected the exception to be caught and re-thrown .   But it could be a debug glitch, but the read never got called on line: https://github.com/spring-projects/spring-integration/pull/10683/files#diff-6451fd5c22e08caea6f3005257a2c56b99adfac27424810b8375d8848a73b61bR534

artembilan: I think you are right.
My test comes with flaws.
There is another place where `RejectedExecutionException` is thrown.
So, that led me to wrong impression with green execution.

Plus there is a problem that we do `TcpNioConnection conn1 = mock();`, therefore `readPacket()` is not going to be called 😢 

Thank you and stay tuned!

cppwfs (approved):
Beautiful!

Great work here!

PR #10688, opened 2026-01-09
Update Gradle dependencies to use `implementation`

cppwfs:
Replace deprecated `compile` configuration with `implementation` across all Spring Integration reference documentation. The `compile` configuration was deprecated in Gradle 4.0 and removed in Gradle 7.0. This update ensures documentation examples use current Gradle best practices.

### DIFF
--- a/src/reference/antora/modules/ROOT/pages/amqp.adoc
+++ b/src/reference/antora/modules/ROOT/pages/amqp.adoc
@@ -22,7 +22,7 @@ Gradle::
 +
 [source, groovy, subs="normal", role="secondary"]
 ----
-compile "org.springframework.integration:spring-integration-amqp:{project-version}"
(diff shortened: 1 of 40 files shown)


artembilan:
Thanks.
I’m not sure how is that useful , but we should use relevant labels on issues or PRs.

cppwfs:
`compile` is deprecated on older versions of gradle does not work on the latest gradle version.  So why would we keep the syntax `compile` when the recommendation is to use `implementation`?

artembilan:
I’m not sure in your question.
I fully agreed with the change, so it has been merged.
What I meant is to add a `documentation` label to the PR since it is indeed about docs.
Just some issues management.

cppwfs:
I see now.   Thank you for the merge!

PR #10699, opened 2026-01-13
GH-10682: Fix MQTT for shared subscription on `ClientManager`

### LINKED ISSUE #10682: Issue with MQTT v5 Shared Subscriptions in Mqttv5PahoMessageDrivenChannelAdapter
# In what version(s) of Spring Integration are you seeing this issue?
7.0.1 (Spring Boot 4.0.1)
# Describe the bug
When using MQTT v5 shared subscriptions (e.g., $share/group1/my/topic), the Mqttv5PahoMessageDrivenChannelAdapter fails to receive any messages.
Interestingly, if I subscribe to both the shared topic and the direct topic simultaneously (e.g., $share/group1/my/topic and my/topic), I receive the same message twice. However, if I subscribe only to the shared topic, the adapter remains silent and no messages are processed.
This behavior persists even after the fix reported in Issue #8879.
# To Reproduce
Configure a Mqttv5ClientManager with MQTT v5 options.
Initialize a Mqttv5PahoMessageDrivenChannelAdapter using that client manager.
Add a shared topic subscription using adapter.addTopic("$share/group1/your/topic", 1).
Publish a message to your/topic.
Observe that the integration flow does not receive the message.
Add adapter.addTopic("your/topic", 1) to the same adapter, and observe that two messages are received (one from the direct sub and one from the shared sub), which proves the shared subscription is technically active but ignored when standing alone.
# Expected behavior
The adapter should be able to receive and process messages when subscribed only to a shared topic prefix ($share/{group}/).
# Sample
```java
@Bean
public Mqttv5ClientManager clientManager(MqttConnectionOptions options) {
    return new Mqttv5ClientManager(options, "client-id-" + UUID.randomUUID());
}

@Bean
public IntegrationFlow mqttInboundFlow(Mqttv5ClientManager clientManager) {
    // BUG: If only "$share/g1/test/topic" is added, no messages are received.
    Mqttv5PahoMessageDrivenChannelAdapter adapter = 
            new Mqttv5PahoMessageDrivenChannelAdapter(clientManager, "$share/group1/my/topic");
    
    adapter.setPayloadType(byte[].class);

    return IntegrationFlow.from(adapter)
            .handle(message -> System.out.println("Received: " + message.getPayload()))
            .get();
```

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10682

The Paho MQTT misses the logic to check for shared subscriptions when messages are received via listener.
The regular `MqttCallback` is called as a fallback for those unrouted messages.

* Expose `ClientManager.addDefaultMessageHandler()` option to let `AbstractMqttMessageDrivenChannelAdapter` implementations to provide the hook for those unrouted messages

Related to: https://github.com/eclipse-paho/paho.mqtt.java/issues/827

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/core/AbstractMqttClientManager.java
@@ -55,6 +55,8 @@ public abstract class AbstractMqttClientManager<T, C> implements ClientManager<T
 
 	private final Set<ConnectCallback> connectCallbacks = Collections.synchronizedSet(new HashSet<>());
 
+	private final Set<DefaultMessageHandler<?>> defaultMessageHandlers = Collections.synchronizedSet(new HashSet<>());
+
 	private final String clientId;
 
 	private int phase = DEFAULT_MANAGER_PHASE;
@@ -115,6 +117,11 @@ public abstract class AbstractMqttClientManager<T, C> implements ClientManager<T
 
 	protected Set<ConnectCallback> getCallbacks() {
 		return this.connectCallbacks;
+	}
+
+	@SuppressWarnings({"unchecked", "rawtypes"})
+	protected <M> Set<DefaultMessageHandler<M>> getDefaultMessageHandlers() {
+		return (Set<DefaultMessageHandler<M>>) (Set) this.defaultMessageHandlers;
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


cppwfs (changes requested):
No notes.   LGTM.   

Question(only reason I requested changes).   This feels like a feature add, do we want to add this to our docs?

artembilan:
> This feels like a feature add, do we want to add this to our docs?

No, this is a workaround for a bug in the Paho library.
This is fully internal change, so I don't treat it as a feature and no need to document it.
The logic is fully transparent for end-users.
If they want some peculiar use-case for a `ClientManager`, they are feel free to reach to respective Javadocs of that new public API.

cppwfs (approved):
LGTM!

Thanks for answering my question!

Nice resolution for the Paho issue.

PR #10700, opened 2026-01-14
Fix race condition in the `ReactiveMessageSourceProducerTests`

artembilan:
Currently, the `MessageSource` in the `ReactiveMessageSourceProducerTests` sleeps for 200 milliseconds.
The `StepVerifier` uses `.expectNoEvent(Duration.ofMillis(100))` at that point. Since there is no absolute time consolidation between CPUs when they are busy, it is possible that `Thread.sleep(200)` would finish earlier in one thread, than `StepVerifier` finishes its expectations before `expectNoEvent()` in the another.

* Fix `ReactiveMessageSourceProducerTests` to use `CountDownLatch` as a barrier before emitting the last value, and already after `StepVerifier` finishes waiting for the `expectNoEvent()`

Related to: https://ge.spring.io/s/nrwdzemwqzghu/tests/task/:spring-integration-core:test/details/org.springframework.integration.endpoint.ReactiveMessageSourceProducerTests/testReactiveMessageSourceProducing()?expanded-stacktrace=WyIwIl0&top-execution=1

Even if this looks like a bug, IO marked it as a `task` since the problem is exactly with the test, nothing more.

### DIFF
--- a/spring-integration-core/src/test/java/org/springframework/integration/endpoint/ReactiveMessageSourceProducerTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/endpoint/ReactiveMessageSourceProducerTests.java
@@ -17,7 +17,9 @@
 package org.springframework.integration.endpoint;
 
 import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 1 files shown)


cppwfs (approved):
LGTM.

Thank you for stabilizing the build a bit more!

PR #10705, opened 2026-01-14
Fix MQTTv5 reconnection race condition

### LINKED ISSUE #10697: Revise MQTT Message Handlers logic around recconnection
The current logic is like this:

```
	@Override
	protected void publish(String topic, Object mqttMessage, Message<?> message) {
		Assert.isInstanceOf(MqttMessage.class, mqttMessage, "The 'mqttMessage' must be an instance of 'MqttMessage'");
		long completionTimeout = getCompletionTimeout();
		try {
			if (!this.mqttClient.isConnected()) {
				this.lock.lock();
				try {
					if (!this.mqttClient.isConnected()) {
						this.mqttClient.connect(this.connectionOptions).waitForCompletion(completionTimeout);
					}
				}
				finally {
					this.lock.unlock();
				}
			}
```

Even if we double check `this.mqttClient.isConnected()`, the `this.mqttClient.connect()` still may lead to the race condition with other `mqttClient` consumers and fail with:
```
Caused by: org.eclipse.paho.mqttv5.common.MqttException: Connect already in progress	
	at org.eclipse.paho.mqttv5.client.MqttAsyncClient.connect(MqttAsyncClient.java:734)	
	at org.eclipse.paho.mqttv5.client.MqttAsyncClient.connect(MqttAsyncClient.java:715)	
	at org.springframework.integration.mqtt.outbound.Mqttv5PahoMessageHandler.publish(Mqttv5PahoMessageHandler.java:249)
```

That does not mean that we are good to `publish` already, but at least means that we can check for `this.mqttClient.isConnected()` is a `while` loop in the `catch` block.
According to the Paho library:
```
		if (comms.isConnected()) {
			throw ExceptionHelper.createMqttException(MqttClientException.REASON_CODE_CLIENT_CONNECTED);
		}
		if (comms.isConnecting()) {
			throw new MqttException(MqttClientException.REASON_CODE_CONNECT_IN_PROGRESS);
		}
```
It looks like we need to check for both errors in the `catch` block.

cppwfs:
Fixes: #10697

Address race condition in `Mqttv5PahoMessageHandler` when multiple threads attempt to connect simultaneously.

Changes:
- Add `RetryTemplate` with configurable retry count for connection attempts
- Implement `MqttExceptionMatcher` to handle specific MQTT error codes
- Add `setRetryConnectionCount` method to configure retry behavior
- Create test suite validating retry logic

The retry mechanism specifically handles `REASON_CODE_CONNECT_IN_PROGRESS` and `REASON_CODE_CLIENT_CONNECTED` errors, allowing the client to recover gracefully when connection establishment is already in progress. Default retry count is set to 5 with 100ms delay between attempts.

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
@@ -17,8 +17,6 @@
 package org.springframework.integration.mqtt.outbound;
 
 import java.nio.charset.StandardCharsets;
-import java.time.Duration;
-import java.util.function.Predicate;
 
 import org.eclipse.paho.mqttv5.client.IMqttAsyncClient;
 import org.eclipse.paho.mqttv5.client.IMqttToken;
@@ -35,9 +33,6 @@ import org.eclipse.paho.mqttv5.common.packet.MqttProperties;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.factory.BeanCreationException;
-import org.springframework.core.retry.RetryException;
-import org.springframework.core.retry.RetryPolicy;
-import org.springframework.core.retry.RetryTemplate;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.mapping.HeaderMapper;
 import org.springframework.integration.mqtt.core.ClientManager;
@@ -75,12 +70,7 @@ public class Mqttv5PahoMessageHandler extends AbstractMqttMessageHandler<IMqttAs
 	@Nullable
 	private MqttClientPersistence persistence;
 
-	private int retryConnectionCount = 5;
-
 	private HeaderMapper<MqttProperties> headerMapper = new MqttHeaderMapper();
-
-	@SuppressWarnings("NullAway.Init")
-	private RetryTemplate connectionRetryTemplate;
 
 	public Mqttv5PahoMessageHandler(String url, String clientId) {
 		super(url, clientId);
@@ -127,28 +117,6 @@ public class Mqttv5PahoMessageHandler extends AbstractMqttMessageHandler<IMqttAs
 		this.headerMapper = headerMapper;
 	}
 
-	/**
-	 * Set the number of connection retries when {@code MqttClientException} is thrown with reason
-	 * codes of {@code MqttClientException.REASON_CODE_CONNECT_IN_PROGRESS} or
-	 * {@code MqttClientException.REASON_CODE_CLIENT_CONNECTED}.  This feature is disabled if
-	 * the {@code connectionRetryTemplate} has been set.
-	 * @param retryConnectionCount the retry count.
-	 * @since 7.1
-	 */
-	public void setRetryConnectionCount(int retryConnectionCount) {
-		Assert.isTrue(retryConnectionCount >= 0, "'retryConnectionCount' must be >= 0");
-		this.retryConnectionCount = retryConnectionCount;
-	}
-
-	/**
-	 * Set the {@link RetryTemplate} to use for retrying connections to the MQTT broker.
-	 *
-	 * @param connectionRetryTemplate the {@link RetryTemplate} used for retrying failed connection attempts
-	 */
-	public void setConnectionRetryTemplate(RetryTemplate connectionRetryTemplate) {
-		this.connectionRetryTemplate = connectionRetryTemplate;
-	}
-
 	@Override
 	protected void onInit() {
 		super.onInit();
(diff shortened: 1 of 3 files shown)


artembilan:
Good. I’ll take a look into your idea in the morning. Meanwhile , please , check this where we give a retry option already: https://docs.spring.io/spring-integration/reference/handler-advice.html

cppwfs:
Good point.   I thought about allowing the user set the retry policy but thought it was overkill.   But after your comment above, I reconsidered and added that capability.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5ReconnectionRaceConditionTests.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.willAnswer;
+import static org.mockito.BDDMockito.willDoNothing;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for race condition handling in {@link Mqttv5PahoMessageHandler}
+ * when multiple threads attempt to connect simultaneously
```
artembilan: It is indeed a race condition, but not necessary between different threads connecting.
The client might be disconnected by the broker just after `isConnected()`.
So, let's consider to not confuse with this Javadoc, but just talk about a connection problem!

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5ReconnectionRaceConditionTests.java
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * @since 7.1
+ */
+public class Mqttv5ReconnectionRaceConditionTests {
+
+	@Test
+	public void testConnectAfterSingleInProgress() throws Exception {
+		IMqttAsyncClient mockClient = mock(IMqttAsyncClient.class);
+		IMqttToken mockConnectToken = mock(IMqttToken.class);
+		IMqttToken mockPublishToken = mock(IMqttToken.class);
```
artembilan: We can use just simple `mock()` to simplify our code.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5ReconnectionRaceConditionTests.java
```diff
@@ -0,0 +1,161 @@
...
+		willDoNothing().given(mockPublishToken).waitForCompletion(anyLong());
+		given(mockPublishToken.getMessageId()).willReturn(1);
+
+		Mqttv5PahoMessageHandler handler = createHandler(mockClient);
+
+		Message<String> message = MessageBuilder.withPayload("test").build();
+
+		handler.handleMessage(message);
+		verify(mockClient, times(1)).connect(any(MqttConnectionOptions.class));
+		verify(mockClient, times(1)).publish(anyString(), any(MqttMessage.class), any(), any());
```
artembilan: It is `times(1)` by default?
Why go explicitly?

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5ReconnectionRaceConditionTests.java
```diff
@@ -0,0 +1,161 @@
...
+
+		// Simulate the race condition scenario:
+		// 1. Initially disconnected
+		// 2. After first connect attempt (which throws), another thread completes the connection
+		// 3. Subsequent isConnected() calls return true
+		willAnswer(invocation -> connectionCompleted.get()).given(mockClient).isConnected();
+
+		willAnswer(invocation -> {
+			int attempt = connectAttempts.incrementAndGet();
+			if (attempt == 1) {
```
artembilan: Why do we need to extract into a local variable?

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -245,24 +247,40 @@ protected void publish(String topic, Object mqttMessage, Message<?> message) {
...
+								int reasonCode = ex.getReasonCode();
+								if (reasonCode == MqttClientException.REASON_CODE_CONNECT_IN_PROGRESS) {
+										Thread.sleep(100);
+								}
+								else if (reasonCode != MqttClientException.REASON_CODE_CLIENT_CONNECTED) {
+									throw ex;
+								}
+							}
+						}
+
```
artembilan: Nit-pick, but for clear code this empty line is redundant.

artembilan (changes requested):
This is great!
Thank you!

artembilan (approved):
Thank you!
Will merge when build is green

PR #10708, opened 2026-01-16
Remove previously deprecated API

artembilan:
* Leave only Jackson 2 API until Spring portfolio wide decision
* Some removals after turning classes to Java `record`, where getters don't exist
* Fix `AttributePollingChannelAdapterParser` to use an `AttributePollingMessageSource.class` instead of the name now removed class
* Fix typos in the `JpaExecutor` Javadocs


cppwfs (approved):
LGTM

PR #10749, opened 2026-01-21
Disable `requiresReply` for Transformer

artembilan:
The Transformer pattern was never about letting the null result pass down into a reply message. Or perform filter-like logic via transformation.
The service activator was always advertised for a logic to deal with a null result. Plus, after adopting NullAway, the `AbstractMessageProcessingTransformer` now throws `MessageTransformationException` when the processing result is null.

* Fix `MessageTransformingHandler` to disallow changing `requiresReply` to avoid a false impression that a null result could lead to something but an exception.

More info in: https://stackoverflow.com/questions/79871975/best-practices-for-error-handling-in-synchronous-spring-integration-flows

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/transformer/MessageTransformingHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transformer/MessageTransformingHandler.java
@@ -52,7 +52,7 @@ public class MessageTransformingHandler extends AbstractReplyProducingMessageHan
 	 */
 	@SuppressWarnings("this-escape")
 	public MessageTransformingHandler() {
-		setRequiresReply(true);
+		super.setRequiresReply(true);
 	}
 
 	/**
@@ -74,6 +74,16 @@ public class MessageTransformingHandler extends AbstractReplyProducingMessageHan
 	public void setTransformer(Transformer transformer) {
 		Assert.notNull(transformer, "transformer must not be null");
 		this.transformer = transformer;
+	}
+
+	/**
+	 * Overridden to throw an {@link UnsupportedOperationException} since transformer must never return null
+	 * and {@code requiresReply == true} by default.
+	 * @param requiresReply never used.
+	 */
+	@Override
+	public void setRequiresReply(boolean requiresReply) {
+		throw new UnsupportedOperationException("Transformer always requires a non-null reply payload.");
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


artembilan:
NOTE: this is not for back-porting to `7.0.x` since it causes some breaking change and leads to slightly different behavior in the target project.

src/reference/antora/modules/ROOT/pages/transformer.adoc
```diff
@@ -124,7 +124,8 @@ As of Spring Integration 2.0, a message transformer's transformation method can
 Returning `null` results in an exception (a `MessageTransformationException` to be precise), because a message transformer should always be expected to transform each source message into a valid target message.
 In other words, a message transformer should not be used as a message filter, because there is a dedicated `<filter>` option for that.
 However, if you do need this type of behavior (where a component might return `null` and that should not be considered an error), you could use a service activator.
-Its `requires-reply` value is `false` by default, but that can be set to `true` in order to have exceptions thrown for `null` return values, as with the transformer.
+Its `requiresReply` value is `false` by default, but that can be set to `true` in order to have exceptions thrown for `null` return values, as with the transformer.
```
cppwfs: This sentence is confusing.  It states that the value is `false` by default, but can be set to `true`.   Then we say it's always `true`.

spring-integration-core/src/main/java/org/springframework/integration/transformer/MessageTransformingHandler.java
```diff
@@ -76,6 +76,16 @@ public void setTransformer(Transformer transformer) {
 		this.transformer = transformer;
 	}
 
+	/**
+	 * Overridden to throw an {@link UnsupportedOperationException} since transformer must never return null
+	 * and {@code requiresReply == true} by default.
+	 * @param requiresReply never used.
+	 */
+	@Override
```
cppwfs, line 85: Can we put a `@since` 7.1 on an overridde method, to denote that as of 7.1 we are overriding this behavior?

cppwfs (changes requested):
LGTM.

Just a couple of document changes.

src/reference/antora/modules/ROOT/pages/transformer.adoc
artembilan: This sentence has a connection with the previous one which talks about a service activator.
I mean it is better to read the whole paragraph to see what is going on.
Does that still make sense to change this one?

cppwfs: I see my mistake.  When I read the word `Its` I was thinking, the transformer (i.e. the paragraph's topic) not the service activator.   Maybe we should replace `Its` with "The service activator's".

artembilan: I would not so we would lose the power of the language, but OK.
If that sounds better fro end-user perspective, let it be!

cppwfs (approved):
This is a good catch!

Thanks for fixing this!!!!

Issue #2626, opened 2018-11-12
Integrate GRPC


zhongyang0316:
GRPC is very popular now and many projects need to be integrated. Is there any intention to integrate GRPC?

artembilan:
We don't have any objection to support this.
Although having our low knowledge about this subject, we would like to ask about contribution: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc

Otherwise we will come back to this somewhere in the future.

Thank you for the report and understanding!

artembilan:
@zhongyang0316 ,

What made you to close the issue?
The fact that we don't do anything here doesn't mean that we are against such a feature.

So, I'll reopen it for the future consideration and you are free to elaborate more what is in your mind on the matter.

Thanks for understanding

zhongyang0316:
@artembilan 
Because I see spring-integration-rsocket, rsocket and grpc are very similar, so I  think it will not support grpc anymore.Maybe I misunderstood it,sorry.
Thank you

artembilan:
Well, have just bumped into this article from Twitter: https://medium.com/netifi/differences-between-grpc-and-rsocket-e736c954e60

So, quoting the first paragraph:

> We get asked about this all the time. In general, gRPC and RSocket attempt to solve different problems. gRPC is an RPC framework using HTTP/2. RSocket is a message passing network layer that is lower level.

Therefore we can't just replace gRPC with RSocket, especially when another side is already on gRPC. Unless why day RSocket would provide a gRPC transport, but I doubt that since there is a special Protobuf model in between.
Right now there are TCP and WebSocket transport supported in RSocket.

Any other thoughts?

tsathishkumar:
There is already some work done here, to have a spring starter for grpc.
[https://github.com/LogNet/grpc-spring-boot-starter](https://github.com/LogNet/grpc-spring-boot-starter)

Might be good to adopt it into spring project.

artembilan:
@tsathishkumar ,

Thank you for sharing that with us.
Only the point of this issue that we really talk about Channel Adapter implementations for Spring Integration: https://spring.io/projects/spring-integration

meltsufin:
The LogNet gRPC starter doesn't seem to be actively developed anymore.
The [Yidongnan gRPC Spring Boot Starter](https://github.com/yidongnan/grpc-spring-boot-starter) seems like a more active project. However, neither has support for Spring Integration AFAIK.

[...]

PR #10750, opened 2026-01-21
Add Grpc DSL for inbound/outbound gateways

cppwfs:
* Create `Grpc` factory class with static factory methods
* Create `GrpcOutboundGatewaySpec` for fluent configuration
* Support method name configuration via string or SpEL expression
* Support `CallOptions` customization for advanced gRPC features
* Add test coverage for DSL functionality
* Add DSL docs to reference material
* Add `Grpc.inboundGateway()` factory method for DSL flows
* Create `GrpcInboundGatewaySpec` for fluent configuration
* Update `GrpcBackToBackTests` to demonstrate DSL usage

refer to #2626 for information


spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/Grpc.java
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Create a {@link GrpcOutboundGatewaySpec} for an outbound gateway with a specific method name.
+	 * @param channel the gRPC channel to use for communication
+	 * @param grpcServiceClass the gRPC service class
+	 * @param methodName the name of the gRPC method to call
+	 * @return the {@link GrpcOutboundGatewaySpec}
+	 */
+	public static GrpcOutboundGatewaySpec outboundGateway(Channel channel, Class<?> grpcServiceClass,
+			String methodName) {
```
artembilan: This is not correct.
Since `method` is optional, then it has to be an option on the `GrpcOutboundGatewaySpec`.
The factory method comes only with those args which are required for ctor.

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/Grpc.java
```diff
@@ -0,0 +1,72 @@
...
+public final class Grpc {
+
+	/**
+	 * Create a {@link GrpcOutboundGatewaySpec} for an outbound gateway.
+	 * @param channel the gRPC channel to use for communication
+	 * @param grpcServiceClass the gRPC service class
+	 * @return the {@link GrpcOutboundGatewaySpec}
+	 */
+	public static GrpcOutboundGatewaySpec outboundGateway(Channel channel, Class<?> grpcServiceClass) {
+		return new GrpcOutboundGatewaySpec(new GrpcOutboundGateway(channel, grpcServiceClass));
```
artembilan: The gateway instance has to be created internally in the spec.

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/GrpcOutboundGatewaySpec.java
```diff
@@ -0,0 +1,96 @@
...
+public class GrpcOutboundGatewaySpec extends MessageHandlerSpec<GrpcOutboundGatewaySpec, GrpcOutboundGateway> {
+
+	protected GrpcOutboundGatewaySpec(GrpcOutboundGateway gateway) {
+		this.target = gateway;
+	}
+
+	/**
+	 * Set the name of the gRPC method to call.
+	 * If method name is not provided, the default expression checks the
+	 * {@link org.springframework.messaging.MessageHeaders} for {@link GrpcHeaders#SERVICE_METHOD} or, in case a single
```
artembilan: If class is used only in Javadoc, then fully-qualified name has to be here, like you have for `MessageHeaders`

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/GrpcOutboundGatewaySpec.java
```diff
@@ -0,0 +1,96 @@
...
+	 * @return the spec
+	 * @see GrpcOutboundGateway#setMethodName(String)
+	 */
+	public GrpcOutboundGatewaySpec methodName(String methodName) {
+		this.target.setMethodName(methodName);
+		return this;
+	}
+
+	/**
+	 * Set the {@link org.springframework.expression.spel.standard.SpelExpression} to resolve the gRPC method name at
```
artembilan: This is not correct: the method does not set a `SpelExpression`, but asks for a `String`.
So, please, reword a little bit into something like `Set the SpEL expression` and that's it.

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/GrpcOutboundGatewaySpec.java
```diff
@@ -0,0 +1,96 @@
...
+
+	/**
+	 * Set an {@link Expression} to resolve the gRPC method name at runtime.
+	 * <p>
+	 * The expression is evaluated against the request message.
+	 * @param methodNameExpression the expression to resolve the method name
+	 * @return the spec
+	 * @see GrpcOutboundGateway#setMethodNameExpression(Expression)
+	 */
+	public GrpcOutboundGatewaySpec methodNameExpression(Expression methodNameExpression) {
```
artembilan, line 88: Please, consider to add yet `methodNameFucntion(Function<Message<?>> methodNameFucntion)` option.
See, for example, `AmqpClientMessageHandlerSpec.exchangeFunction`

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/dsl/GrpcInboundGatewaySpecTests.java
```diff
@@ -0,0 +1,100 @@
...
+		assertThat(gateway).isNotNull();
+		assertThat(getReplyTimeout(gateway)).isEqualTo(3000L);
+	}
+
+	@Test
+	void testFluentApiChaining() {
+		GrpcInboundGatewaySpec spec = Grpc.inboundGateway(TestHelloWorldGrpc.TestHelloWorldImplBase.class)
+				.requestTimeout(5000L)
+				.replyTimeout(3000L)
+				.autoStartup(false);
```
artembilan: I don't believe in these unit tests since all the option you test in this class are really not a `GrpcInboundGatewaySpec` responsibility, but rather its super class. And that I guess is covered in some other tests in the `core` module.
We do have integration tests now in that `GrpcBackToBackTests`, so that is what enough for coverage.

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/dsl/GrpcInboundGatewaySpecTests.java
```diff
@@ -0,0 +1,100 @@
...
+
+		GrpcInboundGateway gateway = spec.getObject();
+		assertThat(gateway).isNotNull();
+		assertThat(gateway.isAutoStartup()).isFalse();
+		assertThat(getRequestTimeout(gateway)).isEqualTo(5000L);
+		assertThat(getReplyTimeout(gateway)).isEqualTo(3000L);
+	}
+
+	private Class<?> getGrpcServiceClass(GrpcInboundGateway gateway) {
+		Field grpcServiceClassField = ReflectionUtils.findField(GrpcInboundGateway.class, "grpcServiceClass");
```
artembilan: We already discussed with your before `TestUtils.getPropertyValue()`.
Either way, I hope this test class will be remove in the next commit.
Thanks

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/dsl/GrpcOutboundGatewaySpecTests.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	@Test
+	void testFluentApiChaining() {
+		CallOptions customCallOptions = CallOptions.DEFAULT
+				.withDeadlineAfter(10, TimeUnit.SECONDS);
+
+		GrpcOutboundGatewaySpec spec = Grpc.outboundGateway(this.channel, TestHelloWorldGrpc.class)
+				.methodName("SayHello")
+				.callOptions(customCallOptions);
```
artembilan: I think these two options verification can be covered in our integration test.
Therefore, this whole unit tests class becomes redundant.

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -153,6 +153,22 @@ private static HelloReply newHelloReply(String message) {
 The routing is done on the `GrpcHeaders.SERVICE_METHOD` header populated by the `GrpcInboundGateway`.
 All the downstream transformer business methods are type-safe in regard to gRPC messages for the `TestHelloWorldGrpc.TestHelloWorldImplBase` service.
 
+[[configuration-with-dsl]]
+=== Configuration with DSL
+
+Add the `GrpcInboundGateway` to flows using the DSL:
```
artembilan: This is good, but I think for a new reader it would be better to start with mentioning a `Grpc` factory and give a link to the general Java DSL chapter.

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -153,6 +153,22 @@ private static HelloReply newHelloReply(String message) {
...
+
+[source,java]
+----
+@Bean
+IntegrationFlow grpcInboundFlow() {
+    return IntegrationFlow.from(
+                    Grpc.inboundGateway(TestSingleHelloWorldGrpc.TestSingleHelloWorldImplBase.class)
+                        .requestTimeout(3000L))
```
artembilan, line 166: This sample should have a `.transformer()` as a request processing endpoint, or just `.channel()`  to make it really as a working example.

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -239,3 +255,18 @@ If `async` is set to `false`, then the response object is returned into a reply
...
+
+Add the `GrpcOutboundGateway` to flows using the DSL:
+
+[source,java]
+----
+@Bean
+IntegrationFlow grpcOutboundFlow(ManagedChannel channel) {
+    return f -> f
+            .handle(Grpc.outboundGateway(channel, TestSingleHelloWorldGrpc.class)
+            .methodName("SayHello"));
```
artembilan: I think indents are a bit off in this sample.

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -239,3 +255,18 @@ If `async` is set to `false`, then the response object is returned into a reply
 ** Single `RequestType` object
 
 Client streaming methods return a `Mono<ResponseType>`, while bidirectional streaming methods return a `Flux<ResponseType>` containing the response.
+
+[[configuration-with-dsl]]
```
artembilan: I wonder if it is really correct to use the same `id` for sections.

artembilan (changes requested):
> refer to https://github.com/spring-projects/spring-integration/issues/2626 for information

I think this is `Fixes:` would be OK.
I don't see anything else to do for this subject. Anything else could be separate in its own issue.

And thank you!

Let's see if my review makes sense!

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/Grpc.java
cppwfs: Understand. 
Question:  I used Ftp.java as my guide https://github.com/spring-projects/spring-integration/blob/main/spring-integration-ftp/src/main/java/org/springframework/integration/ftp/dsl/Ftp.java#L190.   So do we need to bug fix here?

artembilan: Well, it is not fully a bug, but rather a consistency and level of responsibility.
If you provide an instance of the `target` into a spec, then it is not good to mutate such an object internally.
But if we create that `target` internally in the spec, then we are free to do with that object whatever we need without any fear that something would be mutated externally.
I understand your concern about that mutating pattern in the `FtpOutboundGatewaySpec`, but the cat is out of the bag (for now).
The problem there with immutability that `FtpOutboundGateway` comes with several overloaded constructors, so hiding its creating into a spec would be a bit complicated.
And one more observation.
Originally DSLs specs was not `protected`, so creating its instance from the factory with the provided object was not a problem at all.
But one day we have introduced DSL Extensions: https://docs.spring.io/spring-integration/reference/7.1/dsl/java-extensions.html, and opened those classes.
Although, as you see we have left mutability problem.
(Plus, originally I didn't see that problem at all: was not that smart 😉 )
Saying that I won't mind having a respective issue in the backlog: perhaps someone could jump and fix it somehow.
Thank you for raising this question!

cppwfs: Wow!   Thank you for the detailed explanation.   Thank you so much!

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/GrpcOutboundGatewaySpec.java
```diff
@@ -0,0 +1,108 @@
...
+public class GrpcOutboundGatewaySpec extends MessageHandlerSpec<GrpcOutboundGatewaySpec, GrpcOutboundGateway> {
+
+	protected GrpcOutboundGatewaySpec(Channel channel, Class<?> grpcServiceClass) {
+		this.target = new GrpcOutboundGateway(channel, grpcServiceClass);
+	}
+
+	/**
+	 * Set the name of the gRPC method to call.
+	 * If method name is not provided, the default expression checks the
+	 * {@link org.springframework.messaging.MessageHeaders} for {@link org.springframework.integration.grpc.GrpcHeaders#SERVICE_METHOD} or, in case a single
```
artembilan: Even Javadocs should not exceed `120` symbols per line.

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/GrpcOutboundGatewaySpec.java
```diff
@@ -0,0 +1,108 @@
...
+
+	/**
+	 * Set the Spel expression for resolving the method name.
+	 * @param methodNameSpelExpression the expression string
+	 * @return the spec
+	 * @see GrpcOutboundGateway#setMethodNameExpression(Expression)
+	 */
+	public GrpcOutboundGatewaySpec methodNameExpression(String methodNameSpelExpression) {
+		this.target.setMethodNameExpression(PARSER.parseExpression(methodNameSpelExpression));
+		return this;
```
artembilan: I prefer to use a delegation to other methods:
```
return methodNameExpression(PARSER.parseExpression(methodNameSpelExpression));
```
This way, if we change anything in the delegate method, it would immediately affect this method.
Otherwise we may miss to fix in both places.

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/GrpcBackToBackTests.java
```diff
@@ -54,19 +60,55 @@ public class GrpcBackToBackTests {
...
+				.asInstanceOf(type(HelloReply.class))
+				.extracting(HelloReply::getMessage)
+				.isEqualTo("Hello John");
+	}
+
+	@Test
+	void sendAndReceiveViaGrpcGatewaysWithFunction(
+			@Autowired @Qualifier("grpcOutboundFlowWithFunction.input")
+			MessageChannel inputChannel) {
+		assertThat(requestReply("Jane", inputChannel))
```
artembilan: I prefer to have a blank line between milti-line method header and its body.
That way it is much easier to read the code.

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/GrpcBackToBackTests.java
```diff
@@ -54,19 +60,55 @@ public class GrpcBackToBackTests {
...
+				.extracting(Message::getPayload)
+				.asInstanceOf(type(HelloReply.class))
+				.extracting(HelloReply::getMessage)
+				.isEqualTo("Hello Jane");
+	}
+
+	private Message<?> requestReply(String payload, MessageChannel inputChannel) {
```
artembilan: If we extract this then `static`

src/reference/antora/modules/ROOT/pages/grpc.adoc
artembilan: Was not addressed.

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -153,6 +153,22 @@ private static HelloReply newHelloReply(String message) {
 The routing is done on the `GrpcHeaders.SERVICE_METHOD` header populated by the `GrpcInboundGateway`.
 All the downstream transformer business methods are type-safe in regard to gRPC messages for the `TestHelloWorldGrpc.TestHelloWorldImplBase` service.
 
+[[configuration-grpc-inbound-gateway-with-dsl]]
+=== Configuration with DSL
+
+Use the `Grpc` factory to add the `GrpcInboundGateway` to flows using the xref:changes-5.0-5.1.adoc#x5.1-java-dsl[DSL]:
```
artembilan: The link doesn't look right.
Why it is from the changes history, but not actual for the current doc version: `xref:dsl.adoc[]`?

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -239,3 +255,24 @@ If `async` is set to `false`, then the response object is returned into a reply
 ** Single `RequestType` object
 
 Client streaming methods return a `Mono<ResponseType>`, while bidirectional streaming methods return a `Flux<ResponseType>` containing the response.
+
+[[configuration-grpc-outbound-gateway-with-dsl]]
+=== Configuration with DSL
+
+Add the `GrpcOutboundGateway` to flows using the DSL:
```
artembilan: Missed link to Java DSL chapter.
What is I'm only interested in the client side, so I look definitely exactly into this section and it does not give me enough.

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/GrpcBackToBackTests.java
```diff
@@ -77,13 +119,39 @@ static class TestConfig {
...
+
+		@Bean
+		IntegrationFlow grpcOutboundFlowWithCustomizations(ManagedChannel channel) {
+			SpelExpressionParser parser = new SpelExpressionParser();
+			CallOptions customCallOptions = CallOptions.DEFAULT
+					.withDeadlineAfter(10, TimeUnit.SECONDS);
+
+			return f -> f.handle(Grpc.outboundGateway(channel, TestSingleHelloWorldGrpc.class)
+					.methodNameExpression(parser.parseExpression("'SayHello'"))
+					.callOptions(customCallOptions));
```
artembilan: Doesn't look like we verify these customizations in the tests.
More over, it looks like all the tests do exactly the same, so I guess changing an input channel from one to another would still makes test green.
If we there with a duplication, then we don't need so many test methods and so many gateways.
I believe that would be enough to customize only one and verify that customizations are propagated from the DSL to the target component.
Something similar what you had in unit tests but against the `GrpcOutboundGateway` bean.

artembilan (changes requested):

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/dsl/GrpcOutboundGatewaySpec.java
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	/**
+	 * Set the FunctionExpression for resolving the method name.
+	 * @param methodNameFunction the function for a {@link FunctionExpression}
+	 * @return the spec
+	 * @see GrpcOutboundGateway#setMethodNameExpression(Expression)
+	 */
+	public GrpcOutboundGatewaySpec methodNameFunction(Function<Message<?>, String> methodNameFunction) {
+		this.target.setMethodNameExpression(new FunctionExpression<>(methodNameFunction));
```
artembilan: The delegation could be done here as well.

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -156,7 +156,9 @@ All the downstream transformer business methods are type-safe in regard to gRPC
 [[configuration-grpc-inbound-gateway-with-dsl]]
 === Configuration with DSL
 
-Use the `Grpc` factory to add the `GrpcInboundGateway` to flows using the xref:changes-5.0-5.1.adoc#x5.1-java-dsl[DSL]:
+Use the `Grpc` factory to add the `GrpcInboundGateway` to flows using the xref:dsl.adoc[DSL].
```
artembilan: Why just `[]` doesn’t work?
I thought ASCIDoc is able to render a link from target section title .

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -156,7 +156,9 @@ All the downstream transformer business methods are type-safe in regard to gRPC
...
+The DSL provides a fluent API that simplifies configuration and improves readability.
```
artembilan: Why do we need to say this here if we just have given a link to the whole DSL chapter?

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -156,7 +156,9 @@ All the downstream transformer business methods are type-safe in regard to gRPC
...
+The inbound gateway acts as a server-side endpoint that exposes your integration flow as a gRPC service.
```
artembilan: Why do we need to repeat ourselves here if we really are in a sub-section of the inbound gateway?
My concern previously was about missing parts in the section which is not a part of another

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -259,20 +268,40 @@ Client streaming methods return a `Mono<ResponseType>`, while bidirectional stre
 [[configuration-grpc-outbound-gateway-with-dsl]]
 === Configuration with DSL
 
-Add the `GrpcOutboundGateway` to flows using the DSL:
+Use the `Grpc` factory to add the `GrpcOutboundGateway` to flows using the xref:dsl.adoc[DSL].
+The DSL provides a fluent API that simplifies configuration and improves readability.
```
artembilan: DITTO here for both sentences: link title and repeating text

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -259,20 +268,40 @@ Client streaming methods return a `Mono<ResponseType>`, while bidirectional stre
...
 }
 
-private HelloReply requestReply(HelloRequest helloRequest) {
-    return HelloReply.newBuilder().setMessage("Hello " + helloRequest.getName()).build();
+private HelloReply upperCase(HelloReply helloReply) {
+    return HelloReply.newBuilder().setMessage(helloReply.getMessage().toUpperCase()).build();
+}
+----
+
+When the gRPC service has only one method, it will be auto-detected.
```
artembilan, line 286: I think from here and down it is repeating of the `Method Name Configuration` section. If we want to say here something, then let’s use a link to there!
Not sure, though, that we need the whole flow sample just for method name. It is lost a bit in that big code. Might be better to mention different method name settings via link and explain how that could be configured via DSL

artembilan (changes requested):

artembilan (approved):
👍

PR #10761, opened 2026-01-27
GH-10760: Support topic filters in messages arrived callback

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10760

The previous fix with protection against subscribed topics did not include the logic about wildcard support in MQTT subscriptions

* Fix `MqttPahoMessageDrivenChannelAdapter` extracting `messageArrivedIfMatched()` which would be called as a fallback common callback. Delegate matching logic to the `MqttTopic.isMatched()` utility
* Fix `Mqttv5PahoMessageDrivenChannelAdapter` extracting `messageArrivedIfMatched()` which would be called as a fallback common callback. Delegate matching logic to the `MqttTopicValidator.isMatched()` utility. However, before that check for the `$share/` prefix and strip two subscription parts. Exactly the logic missed in Paho Client
* Modify `Mqttv5BackToBackTests.testSharedTopicMqttv5Interaction()` for wildcard to be sure new matching logic works as expected

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -38,6 +38,7 @@ import org.eclipse.paho.mqttv5.common.MqttMessage;
 import org.eclipse.paho.mqttv5.common.MqttSubscription;
 import org.eclipse.paho.mqttv5.common.packet.MqttProperties;
 import org.eclipse.paho.mqttv5.common.packet.MqttReturnCode;
+import org.eclipse.paho.mqttv5.common.util.MqttTopicValidator;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.factory.BeanCreationException;
@@ -96,7 +97,7 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 	/**
 	 * Used as a fallback option for shared subscriptions unrouted messages
 	 */
-	private final ClientManager.DefaultMessageHandler<MqttMessage> defaultMessageHandler = this::messageArrived;
+	private final ClientManager.DefaultMessageHandler<MqttMessage> defaultMessageHandler = this::messageArrivedIfMatched;
 
 	private final MqttConnectionOptions connectionOptions;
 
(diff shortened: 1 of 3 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -546,6 +530,21 @@ public void authPacketArrived(int reasonCode, MqttProperties properties) {
 
 	}
 
+	private void messageArrivedIfMatched(String topic, MqttMessage mqttMessage) {
+		for (String subscribedTopic : getTopic()) {
```
cppwfs, line 534: Should we make `subscribedTopic` final?

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackTests.java
```diff
@@ -144,7 +144,7 @@ public void testSimpleMqttv5Interaction() {
 
 	@Test
 	public void testSharedTopicMqttv5Interaction() {
-		this.mqttv5MessageDrivenChannelAdapter.addTopic("$share/group/testTopic");
+		this.mqttv5MessageDrivenChannelAdapter.addTopic("$share/group/#");
```
cppwfs, line 147: I debated this, should we still have a negative test as well.  For example:

```
	@Test
	public void testSharedTopicMqttv5InvalidTopic() {
		this.mqttv5MessageDrivenChannelAdapter.addTopic("$share/group/testTopic");

		String testPayload = "shared topic payload";
		this.mqttOutFlowInput.send(
				MessageBuilder.withPayload(testPayload)
						.setHeader(MqttHeaders.TOPIC, "FailTopic")
						.build());

		Message<?> receive = this.fromMqttChannel.receive(10_000);

		assertThat(receive).isNull();
	}
```

cppwfs (changes requested):
Looks good.  

Have a couple of questions.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
artembilan: I've changed my mind: we don't need extra local variable and can just mutate this scoped `subscribedTopic` one.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackTests.java
artembilan: That is good idea, but doesn't look like it can be reached.
If nothing is subscribed to the topic, then broker drops such a message silently.
So, while it is still true about `assertThat(receive).isNull();`, that does not mean it reflects reality and our filtering algorithm from the question is in action.
And looks like if we try something with another subscriber, the library has a logic to short circuit if there is a subscription for that topic, therefore fallback callback is not called.
Sorry, I don't see a way how would we test this negative situation where in reality the logic has to go to the Paho library itself. Exactly into that their matching algorithm.

cppwfs: I agree.  Thanks for the great writeup!

cppwfs (approved):
LGTM!

Thanks for fixing this!

PR #10765, opened 2026-01-29
Add TestUtils.propertyValue() based on generics

cppwfs:
Replace three-parameter `TestUtils.getPropertyValue()` calls with the type-safe generic method signature across all Spring Integration test modules. The deprecated method that required an explicit `Class<T>` parameter has been replaced with the generic method that infers the type parameter.

### DIFF
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/channel/ChannelTests.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/channel/ChannelTests.java
@@ -18,6 +18,7 @@ package org.springframework.integration.amqp.channel;
 
 import java.io.IOException;
 import java.util.Collection;
+import java.util.Iterator;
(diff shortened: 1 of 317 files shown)


artembilan (approved):
LGMT.
Pulling locally for final review and some clean up.
Thank you very much!

After merging I'll create a https://github.com/spring-projects/spring-integration/wiki page about migration from 7.0 to 7.1. And then I'll ask you to add a note about this change. The point is that new `<T> @Nullable T getPropertyValue(Object root, String propertyPath)` signature is really a breaking change as we see with our tests. Not a bit deal, but still better to mention and guide properly.

Thanks again!

artembilan:
Merged as after some minor clean and addition to the docs: https://github.com/spring-projects/spring-integration/commit/54a50baa5044720d159038d39a9b9a7c69eef884

PR #10791, opened 2026-02-09
GH-10783: Use GETDEL in RedisMessageStore.doRemove() 

### LINKED ISSUE #10783: Update RedisMessageStore.doRemove() to use GETDEL command for atomic operations
This is a follow-up issue to the [CAS/CAD issue](https://github.com/spring-projects/spring-integration/issues/10782).

The Redis OSS ecosystem team would like to contribute to this repository by updating the current `RedisMessageStore.doRemove()` method to use the [GETDEL](https://redis.io/docs/latest/commands/getdel/) command (available since Redis 6.2) which would reduce server calls and allow atomicity when calling this method.

In terms of backwards compatibility, what approach would you prefer to handle this?

Dgramada:
Fixes: gh-10783

* Use atomic `getAndDelete()` (GETDEL) for Redis 6.2+
* Fall back to GET + UNLINK for older Redis versions
* Add volatile flag for backwards compatibility detection

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
@@ -18,6 +18,7 @@ package org.springframework.integration.redis.store;
 
 import java.util.Collection;
 
+import org.jspecify.annotations.Nullable;
 import org.springframework.beans.factory.BeanClassLoaderAware;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.core.BoundValueOperations;
@@ -38,6 +39,7 @@ import org.springframework.util.Assert;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Michal Domagala
+ * @author Yordan Tsintsov
  *
  * @since 2.1
  */
@@ -48,6 +50,8 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 	private final RedisTemplate<Object, Object> redisTemplate;
 
 	private boolean valueSerializerSet;
+
+	private volatile boolean supportsGetDel = true;
 
 	/**
 	 * Construct {@link RedisMessageStore} based on the provided
@@ -90,7 +94,7 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 	}
 
 	@Override
-	protected Object doRetrieve(Object id) {
+	protected @Nullable Object doRetrieve(Object id) {
 		Assert.notNull(id, ID_MUST_NOT_BE_NULL);
 		BoundValueOperations<Object, Object> ops = this.redisTemplate.boundValueOps(id);
 		return ops.get();
@@ -128,8 +132,25 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 	}
 
 	@Override
-	protected Object doRemove(Object id) {
+	protected @Nullable Object doRemove(Object id) {
 		Assert.notNull(id, ID_MUST_NOT_BE_NULL);
+		Object removedObject;
+		if (supportsGetDel) {
+			try {
+				removedObject = this.redisTemplate.boundValueOps(id).getAndDelete();
+			} catch (Exception e) {
+				// GETDEL command is not supported before Redis 6.2
+				logger.debug("GETDEL not supported, falling back to GET + UNLINK", e);
+				removedObject = this.doRetrieveAndUnlink(id);
+				supportsGetDel = false;
+			}
+		} else {
+			removedObject = this.doRetrieveAndUnlink(id);
+		}
+		return removedObject;
+	}
+
+	private @Nullable Object doRetrieveAndUnlink(Object id) {
 		Object removedObject = this.doRetrieve(id);
 		if (removedObject != null) {
 			this.redisTemplate.unlink(id);
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
@@ -56,6 +56,8 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 
 	private volatile boolean supportsGetDel = true;
 
+	private volatile boolean useUnlink = false;
+
 	/**
 	 * Construct {@link RedisMessageStore} based on the provided
 	 * {@link RedisConnectionFactory} and default empty prefix.
@@ -94,6 +96,27 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 		Assert.notNull(valueSerializer, "'valueSerializer' must not be null");
 		this.redisTemplate.setValueSerializer(valueSerializer);
 		this.valueSerializerSet = true;
+	}
+
+	/**
+	 * Configure the deletion strategy used when removing messages from the store.
+	 * <p>By default ({@code false}), the store uses the Redis {@code GETDEL} command.
+	 * <p>When set to {@code true}, the store uses {@code GET} + {@code UNLINK}.
+	 * <p>Consider enabling this option when:
+	 * <ul>
+	 *     <li>Stored messages are large (e.g., several MB) and blocking deletes
+	 *         cause noticeable Redis latency</li>
+	 *     <li>Atomicity between get and delete is not required for your use case</li>
+	 * </ul>
+	 * <p>Note: For small values, Redis may perform synchronous deletion even with
+	 * {@code UNLINK}, so the benefit is primarily for larger payloads.
+	 * @param useUnlink param to set deletion strategy.
+	 * @since 7.0.2
+	 * @see <a href="https://redis.io/commands/getdel/">GETDEL command</a>
+	 * @see <a href="https://redis.io/commands/unlink/">UNLINK command</a>
+	 */
+	public void setUseUnlink(boolean useUnlink) {
+		this.useUnlink = useUnlink;
 	}
 
 	@Override
@@ -137,7 +160,7 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 	@Override
 	protected @Nullable Object doRemove(Object id) {
 		Assert.notNull(id, ID_MUST_NOT_BE_NULL);
-		if (this.supportsGetDel) {
+		if (this.supportsGetDel && !this.useUnlink) {
 			try {
 				return this.redisTemplate.boundValueOps(id).getAndDelete();
 			}
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
@@ -21,7 +21,7 @@ import java.util.Collection;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.beans.factory.BeanClassLoaderAware;
-import org.springframework.dao.DataAccessException;
+import org.springframework.data.redis.RedisSystemException;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.core.BoundValueOperations;
 import org.springframework.data.redis.core.RedisTemplate;
@@ -136,22 +136,17 @@ public class RedisMessageStore extends AbstractKeyValueMessageStore implements B
 	@Override
 	protected @Nullable Object doRemove(Object id) {
 		Assert.notNull(id, ID_MUST_NOT_BE_NULL);
-		Object removedObject;
 		if (this.supportsGetDel) {
 			try {
-				removedObject = this.redisTemplate.boundValueOps(id).getAndDelete();
+				return this.redisTemplate.boundValueOps(id).getAndDelete();
 			}
-			catch (DataAccessException e) {
+			catch (RedisSystemException e) {
 				// GETDEL command is not supported before Redis 6.2
 				logger.debug("GETDEL not supported, falling back to GET + UNLINK", e);
-				removedObject = this.doRetrieveAndUnlink(id);
 				this.supportsGetDel = false;
 			}
 		}
-		else {
-			removedObject = this.doRetrieveAndUnlink(id);
-		}
-		return removedObject;
+		return this.doRetrieveAndUnlink(id);
 	}
 
 	private @Nullable Object doRetrieveAndUnlink(Object id) {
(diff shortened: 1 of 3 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -18,6 +18,7 @@
 
 import java.util.Collection;
 
+import org.jspecify.annotations.Nullable;
```
cppwfs, line 21: A space must follow this import.
Run `./gradlew :spring-integration-redis:build` to get the checkstyle report

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -128,8 +132,25 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
 	}
 
 	@Override
-	protected Object doRemove(Object id) {
+	protected @Nullable Object doRemove(Object id) {
 		Assert.notNull(id, ID_MUST_NOT_BE_NULL);
+		Object removedObject;
+		if (supportsGetDel) {
```
cppwfs: Instance variable `supportsGetDel` requires a `this.` prefix

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -128,8 +132,25 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
...
+			try {
+				removedObject = this.redisTemplate.boundValueOps(id).getAndDelete();
+			} catch (Exception e) {
```
cppwfs: `catch` needs be on its own line.

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -128,8 +132,25 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
...
+				// GETDEL command is not supported before Redis 6.2
+				logger.debug("GETDEL not supported, falling back to GET + UNLINK", e);
+				removedObject = this.doRetrieveAndUnlink(id);
+				supportsGetDel = false;
```
cppwfs: Instance variable supportsGetDel requires a this. prefix

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -128,8 +132,25 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
...
+			}
+		} else {
```
cppwfs: `else` needs be on its own line.

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
cppwfs: What are the specific Exceptions that can occur in this `try/catch`?   Can these be caught instead? `Exception` seems to be to broad.

cppwfs (changes requested):
Thank you so much for this contribution! 
Other than the changes requested please add tests for this contribution.  
Emit the exception from a mock test to verify that fallback functionality is called in the end .


Thank you!

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageStoreTests.java
```diff
@@ -47,12 +47,17 @@
...
  * @author Artem Bilan
  * @author Artem Vozhdayenko
+ * @author Yordan Tsintsov
  *
  */
 class RedisMessageStoreTests implements RedisContainerTest {
 
 	private static RedisConnectionFactory redisConnectionFactory;
 
+	private static final String DEFAULT_PERSON_ADDRESS = "1600 Pennsylvania Av, Washington, DC";
```
cppwfs: Please use a fake address for example "1234 Main String, Somewhere, City, Province.

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageStoreTests.java
```diff
@@ -47,12 +47,17 @@
...
+
+	private static final String DEFAULT_PERSON_NAME = "Barak Obama";
```
cppwfs: Please use a generic person's name for example, "Jane Doe"

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageStoreTests.java
```diff
@@ -146,6 +151,61 @@ void testAddAndRemoveStringMessage() {
...
+		Address address = new Address();
+		address.setAddress(DEFAULT_PERSON_ADDRESS);
+		Person person = new Person(address, DEFAULT_PERSON_NAME);
+
+		Message<Person> objectMessage = new GenericMessage<>(person);
+		store.addMessage(objectMessage);
+		Message<Person> retrievedMessage = (Message<Person>) store.removeMessage(objectMessage.getHeaders().getId());
+		assertThat(retrievedMessage).isNotNull();
+		assertThat(retrievedMessage.getPayload().getName()).isEqualTo(DEFAULT_PERSON_NAME);
+		assertThat(retrievedMessage.getPayload().getAddress().getAddress()).isEqualTo(DEFAULT_PERSON_ADDRESS);
```
cppwfs, line 184: Thank you for adding the tests!
Please add a test or tests that will exercise the fallback functionality for the following lines:

1. https://github.com/spring-projects/spring-integration/pull/10791/changes#diff-80448790fa1f35c9a100a71658d33f04c69b616aba7eab995d48547ac5d7fe70R145-R148

1. https://github.com/spring-projects/spring-integration/pull/10791/changes#diff-80448790fa1f35c9a100a71658d33f04c69b616aba7eab995d48547ac5d7fe70R152

You can emit the exception from a mock test to verify that fallback functionality.

cppwfs (changes requested):
Thank you for the quick response!
I have a couple more requests.

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -128,8 +134,27 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
...
 	@Override
-	protected Object doRemove(Object id) {
+	protected @Nullable Object doRemove(Object id) {
 		Assert.notNull(id, ID_MUST_NOT_BE_NULL);
+		Object removedObject;
+		if (this.supportsGetDel) {
+			try {
+				removedObject = this.redisTemplate.boundValueOps(id).getAndDelete();
+			}
+			catch (DataAccessException e) {
```
artembilan: Isn't there some specific exception message for unsupported operation to check in this block instead of falling back for any other not Redis version related errors?

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -128,8 +134,27 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
...
+				// GETDEL command is not supported before Redis 6.2
+				logger.debug("GETDEL not supported, falling back to GET + UNLINK", e);
+				removedObject = this.doRetrieveAndUnlink(id);
+				this.supportsGetDel = false;
+			}
+		}
+		else {
```
artembilan: I think we can avoid `else` in favor of `return`.
This way we would return from `if` and go here only in `false` case of .
Therefore we would not need to duplicate `doRetrieveAndUnlink()` invocation.
Just nit pick making code more readable.

artembilan (changes requested):

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageStoreTests.java
```diff
@@ -206,6 +215,60 @@ void testRemoveMultipleMessagesSequentially() {
...
+		RedisMessageStore store = new RedisMessageStore(redisConnectionFactory);
+		RedisMessageStore spyStore = spy(store);
+
+		RedisTemplate<Object, Object> mockRedisTemplate = mock(RedisTemplate.class);
+		ReflectionTestUtils.setField(spyStore, "redisTemplate", mockRedisTemplate);
+
+		BoundValueOperations<Object, Object> mockValueOps = mock(BoundValueOperations.class);
+		when(mockRedisTemplate.boundValueOps(any())).thenReturn(mockValueOps);
+		when(mockValueOps.getAndDelete())
+				.thenThrow(new RedisSystemException("ERR unknown command 'GETDEL'", null));
```
artembilan: Is this mock just for testing or real Redis error thrown in the case of no command?
If that is really, then can we narrow that `catch (RedisSystemException e) {` down to checking this error message to trigger fallback?
I mean who knows what really all those `RedisSystemException` errors could be about?

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageStoreTests.java
```diff
@@ -206,6 +215,60 @@ void testRemoveMultipleMessagesSequentially() {
...
+
+		assertThat(removed).isEqualTo(expectedMessage);
+		assertThat(ReflectionTestUtils.getField(spyStore, "supportsGetDel")).isEqualTo(false);
+		verify(mockValueOps).getAndDelete();
+		verify(mockValueOps).get();
+		verify(mockRedisTemplate).unlink(prefixedKey);
+	}
+
+	@Test
+	void testDoRemoveSkipsGetDelAfterFallback() {
```
artembilan: I don't think we need a dedicated test for this.
Consider to modify a previous one with one more `spyStore.removeMessage(id)` call and verify in the end of test that `getAndDelete` is called only once, but `get` and `unlink` twice.
Jut because `supportsGetDel` is set to `false` between two `removeMessage()` calls.

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -128,8 +134,22 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
...
+			try {
+				return this.redisTemplate.boundValueOps(id).getAndDelete();
+			}
+			catch (RedisSystemException e) {
+				// GETDEL command is not supported before Redis 6.2
+				logger.debug("GETDEL not supported, falling back to GET + UNLINK", e);
+				this.supportsGetDel = false;
+			}
+		}
+		return this.doRetrieveAndUnlink(id);
```
artembilan: Minor and might be existing code, but now we prefer to use methods without `this.` prefix.

spring-integration-redis/src/test/java/org/springframework/integration/redis/store/RedisMessageStoreTests.java
```diff
@@ -146,6 +160,115 @@ void testAddAndRemoveStringMessage() {
...
+		assertThat(removed3.getPayload()).isEqualTo("Message 3");
+		assertThat(store.getMessageCount()).isZero();
+	}
+
+	@Test
+	void testDoRemoveFallbackWhenGetDelNotSupported() {
+		RedisMessageStore store = new RedisMessageStore(redisConnectionFactory);
+		RedisMessageStore spyStore = spy(store);
+
+		RedisTemplate<Object, Object> mockRedisTemplate = mock(RedisTemplate.class);
```
artembilan: Consider to use just `mock()`, without any arguments.
That way we won't suffer from unchecked warning due to not compatible generic types.

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/store/RedisMessageStore.java
```diff
@@ -140,17 +141,28 @@ protected void doStoreIfAbsent(Object id, Object objectToStore) {
...
+				else {
+					throw e;
+				}
 			}
 		}
-		return this.doRetrieveAndUnlink(id);
+		return doRetrieveAndUnlink(id);
+	}
+
+	private boolean isGetDelNotSupportedError(Exception e) {
```
artembilan: Cool!
Make this method `static` please, and move it to the end of this inner class

artembilan (changes requested):
Just one minor nit-pick in the code.
And general request to mention this new functionality in the `whats-new.adoc`.
Since this is an internal change I don’t see a reason to mention it in the target `redis.adoc`. Unless we deliberately tell the story there about GET+UNLINK 😅
Thank you!

Dgramada:
I just want to mention that `GETDEL` is a blocking command while `UNLINK` is an async non-blocking command. 
This means although `GETDEL` will reduce network requests and is atomic, `UNLINK` could be more efficient for very big messages. If message size isn't huge `GETDEL` is the better choice, but for very big payloads `UNLINK` may be a better approach. Based on this information, is `GETDEL` the optimal approach?

artembilan:
> but for very big payloads `UNLINK` may be a better approach.

I see.
Can we make then a new option on the `RedisMessageStore` to opt-in for `UNLINK`, e.g. `useUnlink(boolean)` (`false` by default) with Javadoc explanation forthe reasoning of this option?

Thanks

Dgramada:
In this case, should we remove the fallback logic and only rely on a boolean flag that tells us which commands to support?
Also should we make the `GET + UNLINK` to be the default approach for backwards compatibility?

`RedisMessageStore` could have two constructors, the original constructor which supports `GET + UNLINK` and a second constructor which sets the `RedisMessageStore` to use GETDEL.
This way users could choose which approach would benefit them more when setting up `RedisMessageStore` and we can avoid `try/catch` logic.

artembilan:
I think the fallback logic still has to be there . Imaging the situation I provide an application with GETDEL flag compiled-in ( you say ctor), but this app is installed in different environment with different Redis versions. So, to make it working everywhere I probably would need different app versions, too.
The opt-in is OK, but we need to be flexible as possible.

artembilan (approved):
Looking good!
Pulling locally for final review, clean up and merge.
Thank you!

Dgramada:
Thank you too!

artembilan:
Merged as: https://github.com/spring-projects/spring-integration/commit/d2c880c393f9fcdc2eac365e99ab9983080a812c.

Thank you!
Looking forward for more!

PR #10792, opened 2026-02-13
Add CloudEvent DSL support for transformers

cppwfs:
Introduce DSL API for creating CloudEvent transformers through the new `CloudEvent` factory class and associated spec classes:

* Add `CloudEvent` factory with static methods for transformer specs
* Add `FromCloudEventTransformerSpec` for deserializing CloudEvents
* Add `ToCloudEventTransformerSpec` for serializing to CloudEvents
* Update `CloudEventBackToBackTests` to use new DSL API

The `FromCloudEventTransformerSpec` provides configuration for `eventFormat` to handle deserialization.

The `ToCloudEventTransformerSpec` provides fluent configuration for:
* Event format and content type
* CloudEvent attributes (id, source, type, dataSchema, subject)
* Header prefix customization
* Extension pattern matching


spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEvent.java
```diff
@@ -0,0 +1,56 @@
...
+package org.springframework.integration.cloudevents.dsl;
+
+/**
+ * Factory class for CloudEvent components.
+ *
+ * @author Glenn Renfro
+ *
+ * @since 7.1
+ */
+public final class CloudEvent {
```
artembilan: I think the better name for this class would be `CloudEvents` to not confuse with `CloudEvent` instance from Cloud Events API.
Plus the `CloudEvents` would really reflect the protocol itself since its name is really `CloudEvents`, like JDBC, MQTT, AMQP etc. with their respective Java DSL factories in our projects.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEvent.java
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.cloudevents.dsl;
+
+/**
+ * Factory class for CloudEvent components.
```
artembilan: I think the `CloudEvents` word would be better here.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,163 @@
...
+	 * when {@link EventFormatProvider} is to be used to determine
+	 * {@link EventFormat}.
+	 * <p>If {@code eventFormat} and the {@code eventFormatContentTypeExpression} are provided,
+	 * the {@code eventFormat} has precedence.
+	 * @param eventFormatContentTypeExpression the expression to create
+	 * content type for the {@link EventFormatProvider#resolveFormat(String)}
+	 * @return the spec
+	 * @see io.cloudevents.core.format.ContentType
+	 */
+	public ToCloudEventTransformerSpec  eventFormatContentTypeExpression(Expression eventFormatContentTypeExpression) {
```
artembilan: Looks like to many spaces after `ToCloudEventTransformerSpec` word.
Please, consider to reformat the code.
Also, look into String-based and `Fucntion`-based overloads for all the expressions.
As an example you can look into `CamelMessageHandlerSpec.endpointUri` overloads.

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/transformer/CloudEventBackToBackTests.java
```diff
@@ -159,24 +159,23 @@ IntegrationFlow messageToCloudEvent() {
 					.setEventFormatContentTypeExpression(new LiteralExpression(JsonFormat.CONTENT_TYPE));
 
 			return IntegrationFlow.from("messageInputChannel")
-					.transform(new FromCloudEventTransformer())
-					.transform(toCloudEventsTransformer)
+					.handle(org.springframework.integration.cloudevents.dsl.CloudEvent.fromCloudEventTransformer())
```
artembilan: You see? The `CloudEvents` name for the factory will solve FQCN problem 😄 

I also think that direction for the DSL is a bit off.
We talk about transformer and that was really `transform()` EIP-method, but now it has been turned to `handle()` which is not fully user-friendly.
We should revise those new Specs to be not `MessageHandlerSpec` at all, but as plain `MessageTransformer` as possible.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/FromCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * A {@link MessageHandlerSpec} for a {@link MessageTransformingHandler} that uses
+ * a {@link FromCloudEventTransformer}.
+ *
+ * @author Glenn Renfro
+ *
+ * @since 7.1
+ */
+public class FromCloudEventTransformerSpec
```
artembilan: I think we don't need this Spec at all.
It has only to possible variants with `eventFormat` or without.
Therefore I believe something like this in `CloudEvents` factory would satisfy end-user expectations:
```
	public static FromCloudEventTransformer fromCloudEventTransformer() {
		return new FromCloudEventTransformer();
	}

	public static FromCloudEventTransformer fromCloudEventTransformer(EventFormat eventFormat) {
		FromCloudEventTransformer transformer = new FromCloudEventTransformer();
        transformer.setEventFormat(eventFormat);
		return transformer;
	}
```

See `org.springframework.integration.dsl.Transformers` factory for similarities.
Or `org.springframework.integration.file.dsl.Files.toByteArrayTransformer`

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * A {@link MessageHandlerSpec} for a {@link MessageTransformingHandler} that uses
+ * a {@link ToCloudEventTransformer}.
+ *
+ * @author Glenn Renfro
+ *
+ * @since 7.1
+ */
+public class ToCloudEventTransformerSpec
```
artembilan: The target `ToCloudEventTransformer` has a lot of options therefore we cannot go with a lot of overloads in the `CloudEvents` factory.
So, we still need such a Spec as builder pattern, however, as I said in other places it cannot be `MessageHandlerSpec`. It looks like it cannot be just as simple `IntegrationComponentSpec` because all those options belongs to the endpoint produced by the `transform()` EIP-operator. 
The result of this Spec would be used in that `transform()`.
We may consider a new Spec abstraction dedicated for plain `Transformer` configurations, but looks like this `ToCloudEventTransformer` is the only candidate for now.
So, I would suggest to go with one-stop-shop `ToCloudEventTransformerSpec` without any extensions, expose all those options you have so far, plus those new overloades I'm asking. And give it a plain `public ToCloudEventTransformer get()`. 
This way end-user would do in their `IntegrationFlow` configuration:
```
.transform(CloudEvents.toCloudEventTransformer("myHeader").type(m -> "myType").get())
```
which feels more aligned with what we have so far in our tests and docs.
And more intuitive with builder pattern exposure.

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -102,7 +102,7 @@ public ToCloudEventTransformer cloudEventTransformer() {
 public IntegrationFlow cloudEventTransformFlow(ToCloudEventTransformer toCloudEventTransformer) {
     return IntegrationFlows
         .from("inputChannel")
-        .transform(toCloudEventTransformer)
+        .handle(CloudEvent.toCloudEventTransformer())
```
artembilan: This has to be fixed as well as we discuss in the Specs and `CloudEvents` factory.
Plus, I believe this section deserves some sentence about that new `CloudEvents` factory.

artembilan (changes requested):
Great stuff!
Let's hope my comments not too confusing.
Thanks

artembilan (changes requested):
And I forgot one more.
We also discussed with you a possibility to map CloudEvent headers.
Something like `CloudEventHeadersBuilder` with high-level API for those standard `CloudEvent` attributes and extensions, too.
The example we discussed could be taken from the `Mail.headers()`.
This might be a different issue/ PR though.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
cppwfs: I'm not sure I understand .  We don't accept anything but expressions (not strings).

artembilan: Right, the string can be parsed to Expression. See those overloads for examples

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/FromCloudEventTransformer.java
```diff
@@ -85,7 +85,7 @@ public String getComponentType() {
 	@Override
 	protected Object doTransform(Message<?> message) {
 		Assert.state(message.getPayload() instanceof byte[] || message.getPayload() instanceof CloudEvent,
-				"Payload did not contain 'CloudEvent' nor could it be deserialized to a 'CloudEvent'");
+				"Payload did not contain 'CloudEvents' nor could it be deserialized to a 'CloudEvents'");
```
artembilan: This change is not correct.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/ToCloudEventTransformer.java
```diff
@@ -66,7 +66,7 @@
  *   (either configured via {@link #setEventFormat(EventFormat)} or resolved via
  *   {@link #setEventFormatContentTypeExpression(Expression)}), the {@link CloudEvent} is
  *   serialized into the message payload using that format (e.g., JSON, XML).
- *   The output message contains the serialized CloudEvent as payload with
+ *   The output message contains the serialized CloudEvents as payload with
```
artembilan: Neither this

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/ToCloudEventTransformer.java
```diff
@@ -209,7 +209,7 @@ protected void onInit() {
 			String appName = applicationContext.getEnvironment().getProperty("spring.application.name");
 			if (!StringUtils.hasText(appName)) {
 				logger.warn("'spring.application.name' is not set. " +
-						"CloudEvent source URIs will use 'null' as the application name. ");
+						"CloudEvents source URIs will use 'null' as the application name. ");
```
artembilan: DITTO

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/CloudEventHeaders.java
```diff
@@ -28,55 +28,55 @@
 public final class CloudEventHeaders {
 
 	/**
-	 * The prefix used for CloudEvent attribute headers.
+	 * The prefix used for CloudEvents attribute headers.
```
artembilan: And not here. 
And so on many of those changes from `CloudEvent` object to `CloudEvents` word.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,241 @@
...
+	 * Set the {@link Function} to produce a cloud event format content type
+	 * when {@link EventFormatProvider} is to be used to determine
+	 * {@link EventFormat}.
+	 * <p>If {@code eventFormat} and the {@code eventFormatContentTypeExpression} are provided,
+	 * the {@code eventFormat} has precedence.
+	 * @param eventFormatContentTypeFunction the {@link Function} to create
+	 *                                       content type for the {@link EventFormatProvider#resolveFormat(String)}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec eventFormatContentTypeExpression(
```
artembilan: The update for `Function` is good, but they are really not expressions.
Consider to name methods properly.
For example, this one has to be `eventFormatContentTypeFunction`

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,241 @@
...
+	 * Set the {@link Expression} to produce a cloud event format content type
+	 * when {@link EventFormatProvider} is to be used to determine
+	 * {@link EventFormat}.
+	 * <p>If {@code eventFormat} and the {@code eventFormatContentTypeExpression} are provided,
+	 * the {@code eventFormat} has precedence.
+	 * @param eventFormatContentTypeExpression the {@link Function} to create
+	 *                                         content type for the {@link EventFormatProvider#resolveFormat(String)}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec eventFormatContentTypeExpression(Expression eventFormatContentTypeExpression) {
```
artembilan: Why do we still miss an overload with String:
```
public ToCloudEventTransformerSpec eventFormatContentTypeExpression(String  eventFormatContentTypeExpression) {
      return eventFormatContentTypeExpression(PARSER.parseExpression(eventFormatContentTypeExpression));
}
```

See the mentioned `CamelMessageHandlerSpec.endpointUriExpression()`.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
artembilan: This comment was not fully addressed.
I still see `extends ConsumerEndpointSpec`.
But that's not what we want here.
We really want a plain builder without any extension to confuse end-user.
We might need copy/paste the `PARSER` from the `IntegrationComponentSpec`, but nothing more.

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/transformer/CloudEventBackToBackTests.java
```diff
@@ -152,31 +153,30 @@ void convertMessageToSerializedCloudEventToMessage(
...
+	@Bean
+	IntegrationFlow messageToCloudEvent() {
+		ToCloudEventTransformer toCloudEventsTransformer = new ToCloudEventTransformer();
+		toCloudEventsTransformer
+				.setEventFormatContentTypeExpression(new LiteralExpression(JsonFormat.CONTENT_TYPE));
+
+		return IntegrationFlow.from("messageInputChannel")
+				.transform(CloudEvents.fromCloudEventTransformer())
+				.transform(CloudEvents.toCloudEventTransformer()
+						.eventFormatContentTypeExpression(new LiteralExpression(JsonFormat.CONTENT_TYPE)).get())
```
artembilan: This leads us to the overload to let users to not pass any expression.
Like a plain `eventFormatContentType(JsonFormat.CONTENT_TYPE)`.
Otherwise it is not fully a DSL.

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/transformer/FromCloudEventTransformerTests.java
```diff
@@ -166,7 +166,7 @@ void invalidPayloadTypeThrowsException() {
 				.cause()
 				.isInstanceOf(IllegalStateException.class)
 				.hasMessageContaining(
-						"Payload did not contain 'CloudEvent' nor could it be deserialized to a 'CloudEvent'");
+						"Payload did not contain 'CloudEvents' nor could it be deserialized to a 'CloudEvents'");
```
artembilan: Please, revise your change from `CloudEvent` to `CloudEvents`.
Renaming the factory must not trigger a change for `CloudEvent` object.

artembilan (changes requested):

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEvents.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @since 7.1
+ */
+public final class CloudEvents {
+
+	public static FromCloudEventTransformer fromCloudEventTransformer() {
+		return new FromCloudEventTransformer();
+	}
+
+	public static FromCloudEventTransformer fromCloudEventTransformer(EventFormat eventFormat) {
```
artembilan, line 46: Missed Javadocs for these two factory methods

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,292 @@
...
+		return eventIdExpression(new FunctionExpression<>(eventIdFunction));
+	}
+
+	/**
+	 * Set the {@link io.cloudevents.CloudEvent} {@code id}.
+	 * <p>Default is to extract the {@code id} from the {@link MessageHeaders} of the message.
+	 * @param eventId the  {@code id} for each {@link io.cloudevents.CloudEvent}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec eventId(String eventId) {
```
artembilan: I don't think that `id` can be static.
Therefore let's not allow to set that!
Well, of course an `Expression` can be set to the `LiteralExpression`, but that's already not what our API suggests.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
artembilan: You still missing overloads like I suggested in the previous commit to this discussion.
No `PARSER`-based variants.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,292 @@
...
+	 * {@link EventFormat}.
+	 * <p>If {@code eventFormat} and the {@code eventFormatContentTypeExpression} are provided,
+	 * the {@code eventFormat} has precedence.
+	 * @param eventFormatContentTypeFunction the {@link Function} to create
+	 *                                       content type for the {@link EventFormatProvider#resolveFormat(String)}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec eventFormatContentTypeFunction(
+			Function<Message<?>, String> eventFormatContentTypeFunction) {
+		return eventFormatContentTypeExpression(new FunctionExpression<>(eventFormatContentTypeFunction));
```
artembilan, line 298: Code style concern: if method header is multi-line, then blank line before method body.
Makes it more readable.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/ToCloudEventTransformer.java
```diff
@@ -184,7 +184,7 @@ public void setEventFormat(EventFormat eventFormat) {
 	 * <p>If {@code eventFormat} and the {@code eventFormatContentTypeExpression} are provided,
 	 * the {@code eventFormat} has precedence.
 	 * @param eventFormatContentTypeExpression the expression to create
-	 * content type for the {@link EventFormatProvider#resolveFormat(String)}
+	 *                                         content type for the {@link EventFormatProvider#resolveFormat(String)}
```
artembilan: Something off for formatting.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/ToCloudEventTransformer.java
```diff
@@ -287,7 +287,7 @@ public String getComponentType() {
 
 	/**
 	 * Extract {@link CloudEvent} extensions from message headers based on pattern matching.
-	 * @param headers the message headers to extract extensions from
+	 * @param headers           the message headers to extract extensions from
```
artembilan: DITTO

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/ToCloudEventTransformer.java
```diff
@@ -364,7 +364,7 @@ public MessageBuilderMessageWriter create(SpecVersion version) {
 		 * Add a {@link CloudEvent} context attribute to the message headers.
 		 * Map the {@link CloudEvent} attribute to a message header by prepending the configured prefix
 		 * to the attribute name (e.g., "id" becomes "ce-id" with the default prefix).
-		 * @param name the {@link CloudEvent} attribute name
+		 * @param name  the {@link CloudEvent} attribute name
```
artembilan: And here

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/transformer/CloudEventBackToBackTests.java
```diff
@@ -106,8 +107,8 @@ void convertCloudEventToMessageToCloudEvent(
 
 		CloudEvent cloudEvent = jsonFormat.deserialize((byte[]) result.getPayload());
 		assertThat(cloudEvent.getDataContentType()).isEqualTo("text/plain");
-		assertThat(cloudEvent.getId()).isNotNull(); //new ID created when new CloudEvent is created
-		assertThat(cloudEvent.getTime()).isNotNull(); //new time created when new CloudEvent is created
+		assertThat(cloudEvent.getId()).isNotNull(); //new ID created when new CloudEvents is created
+		assertThat(cloudEvent.getTime()).isNotNull(); //new time created when new CloudEvents is created
```
artembilan: Was not addressed. We really talk here about a `CloudEvent` instance not protocol, not our DSL factory

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/transformer/CloudEventBackToBackTests.java
```diff
@@ -152,31 +153,30 @@ void convertMessageToSerializedCloudEventToMessage(
...
-					.transform(new FromCloudEventTransformer())
-					.transform(toCloudEventsTransformer)
-					.channel("outputChannel")
-					.get();
-		}
+	@Bean
+	IntegrationFlow messageToCloudEvent() {
+		ToCloudEventTransformer toCloudEventsTransformer = new ToCloudEventTransformer();
+		toCloudEventsTransformer
+				.setEventFormatContentTypeExpression(new LiteralExpression(JsonFormat.CONTENT_TYPE));
```
artembilan: This variable is out of use.
I guess can be simply removed altogether.

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/transformer/CloudEventBackToBackTests.java
```diff
@@ -152,31 +153,30 @@ void convertMessageToSerializedCloudEventToMessage(
...
-					.transform(toCloudEventsTransformer)
-					.transform(new FromCloudEventTransformer())
-					.channel("outputChannel")
-					.get();
-		}
+	@Bean
+	IntegrationFlow jsonCaseFlow() {
+		return IntegrationFlow.from("jsonCaseInputChannel")
+				.transform(CloudEvents.toCloudEventTransformer(
+						TEST_PATTERNS)
```
artembilan: I find this formatting as a bit off.
Doesn't look like it breaks `120` symbols rule.

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -31,7 +31,7 @@ implementation "org.springframework.integration:spring-integration-cloudevents:{
 
 Use the `ToCloudEventTransformer` to convert Spring Integration messages into CloudEvents-compliant messages.
 This transformer supports the CloudEvents specification v1.0 and serializes CloudEvents if an `EventFormat` or `eventFormatContentTypeExpression` is specified.
-When you specify an `EventFormat` or `eventFormatContentTypeExpression`, the transformer uses the `EventFormat` to generate the `CloudEvent` in the payload.
+When you specify an `EventFormat` or `eventFormatContentTypeExpression`, the transformer uses the `EventFormat` to generate the `CloudEvents` in the payload.
```
artembilan: Was not addressed.
We talk about an instance not factory, not protocol.

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -170,12 +170,12 @@ The transformer accepts messages with the following payload types:
...
 ==== CloudEvent Object Type
 
-When the message payload is a `CloudEvent` instance, the transformer:
+When the message payload is a `CloudEvents` instance, the transformer:
 
 1. Extracts the CloudEvent data and uses it as the message payload.
 2. Maps CloudEvent attributes (`id`, `source`, `type`, `time`, `subject`, `datacontenttype`, `dataschema`) to message headers with a `ce-` prefix.
 3. Maps all CloudEvent extensions to message headers with the `ce-` prefix.
-4. Preserves all original message headers unless a header key matches a `CloudEvent` attribute or extension, in which case the original value is overwritten.
+4. Preserves all original message headers unless a header key matches a `CloudEvents` attribute or extension, in which case the original value is overwritten.
```
artembilan: Looks like there are still many not addressed wrong `CloudEvents` changes.

artembilan (changes requested):

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
cppwfs: I don't understand the DSL supports:
1. Expression
2. Function
3. String

So we drop Expression, or do we add a new one?

artembilan: We add a new one `...Expression(String)` variant.
And there we call `PARSER` and delegate to another one based on `Expression`.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,347 @@
...
+	 * @param extensionPatterns patterns to evaluate whether message headers should be added as extensions
+	 * to the {@link CloudEvent}.
+	 */
+	protected ToCloudEventTransformerSpec(String... extensionPatterns) {
+		this.transformer = new ToCloudEventTransformer(extensionPatterns);
+	}
+
+	/**
+	 * Set the {@link EventFormat} to use for {@link io.cloudevents.CloudEvent} serialization.
+	 * <p>If {@code eventFormat} and the {@code eventFormatContentTypeExpression} are provided, the {@code
```
artembilan: For better readability I would break the line just after comma, rather than `{@code }` block.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,347 @@
...
+		return sourceExpression(new FunctionExpression<>(sourceFunction));
+	}
+
+	/**
+	 * Set the {@link io.cloudevents.CloudEvent} {@code source}.
+	 * <p>Default is {@code "/spring/" + appName + "." + getBeanName())}.
+	 * @param source the {@code source} for each {@link io.cloudevents.CloudEvent}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec source(String source) {
```
artembilan: The `source` attribute must be an `URI`.
See `ToCloudEventTransformer` and what types are expected for all those expression resolutions.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,347 @@
...
+	public ToCloudEventTransformerSpec dataSchemaFunction(Function<Message<?>, String> dataSchemaFunction) {
+		return dataSchemaExpression(new FunctionExpression<>(dataSchemaFunction));
+	}
+
+	/**
+	 * Set the {@code dataSchema} for the {@link io.cloudevents.CloudEvent}.
+	 * @param dataSchema the  {@code dataSchema} for each {@link io.cloudevents.CloudEvent}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec dataSchema(String dataSchema) {
```
artembilan: `dataschema` is also `URI`

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,347 @@
...
+		return subjectExpression(new FunctionExpression<>(subjectFunction));
+	}
+
+	/**
+	 * Set the {@code subject} for the {@link io.cloudevents.CloudEvent}.
+	 * @param subject the {@code subject} for each {@link io.cloudevents.CloudEvent}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec subject(String subject) {
+		return subjectExpression(new LiteralExpression(subject));
```
artembilan, line 260: That is not critical, but it is better from code readability to have the delegation method declared below the place it is called.
In this case that `subjectExpression(Expression subjectExpression)` has to be already after all other overloads.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/ToCloudEventTransformer.java
artembilan: Still not addressed.

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -282,7 +282,7 @@ GenericMessage [
 [#cloudevent-attribute-mapping]
 === CloudEvent Attribute Mapping
 
-The transformer maps `CloudEvent` attributes to message headers using the following `CloudEventHeaders` constants:
+The transformer maps `CloudEvents` attributes to message headers using the following `CloudEventHeaders` constants:
```
artembilan: Looks like you have not revised the doc for unnecessary change from the `CloudEvent`  to `CloudEvents`

artembilan (changes requested):

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
cppwfs: What is the proper way to hand a URI for this expression.?  i.e. uri.toASCIIString()?

artembilan:
```
public ToCloudEventTransformerSpec source(URI source) {
		return sourceExpression(new ValueExpression<>(source));
	}
```

artembilan (changes requested):

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/ToCloudEventTransformerSpec.java
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	/**
+	 * Set the {@link Function} to create {@link io.cloudevents.CloudEvent} {@code source}.
+	 * <p>Default is {@code "/spring/" + appName + "." + getBeanName())}.
+	 * @param sourceFunction the {@link Function} to create the {@code source} for each
+	 * {@link io.cloudevents.CloudEvent}
+	 * @return the spec
+	 */
+	public ToCloudEventTransformerSpec sourceFunction(Function<Message<?>, String> sourceFunction) {
```
artembilan: Probably function must return `URI` as well.

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -159,9 +159,9 @@ The transformer throws an `IllegalArgumentException` if the payload is not a byt
 == `FromCloudEventTransformer`
 
 Use the `FromCloudEventTransformer` to convert CloudEvents into Spring Integration messages.
-This transformer supports the CloudEvents specification v1.0 and processes CloudEvents from two payload types: `CloudEvent` objects or serialized `CloudEvent` byte arrays.
+This transformer supports the CloudEvents specification v1.0 and processes CloudEvents from two payload types: `CloudEvents` objects or serialized `CloudEvents` byte arrays.
```
artembilan: Please, double check these docs changes.
May of those red-to-green are invalid from my point of view.
For example, ` serialized CloudEvents` is wrong. Has to be singular because we talk about an object.
Same about `payload is a CloudEvents instance`.

artembilan (changes requested):

artembilan (approved):
LGTM

PR #10794, opened 2026-02-13
Fix `AbstractMessageProducingHandler` to treat `Flux` as single output

cppwfs:
Fixes #10793
- Exclude `Flux` from iterable splitting in `sendOutputs` to prevent incorrect iteration over reactive streams
- Exclude `Flux` from iterable message-history cloning check in `produceOutput`
- Add `AbstractMessageProducingHandlerTests` to verify `Flux`, `Mono`, and queue-backed `Flux` payloads pass through as single outputs

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
@@ -241,7 +241,8 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	}
 
 	protected void sendOutputs(@Nullable Object result, Message<?> requestMessage) {
-		if (result instanceof Iterable<?> iterableResult && shouldSplitOutput(iterableResult)) {
+		if (!(result instanceof Flux<?>) && result instanceof Iterable<?> iterableResult &&
+				shouldSplitOutput(iterableResult)) {
 			for (Object o : iterableResult) {
 				produceOutput(o, requestMessage);
 			}
@@ -510,7 +511,7 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 		else {
 			builder = getMessageBuilderFactory().withPayload(output);
 			// Assuming that a message in the payload collection is a copy of the request message.
-			if (output instanceof Iterable<?> iterable) {
+			if (!(output instanceof Flux<?>) && output instanceof Iterable<?> iterable) {
 				Iterator<?> iterator = iterable.iterator();
 				if (iterator.hasNext() && iterator.next() instanceof Message<?>) {
 					builder = builder.cloneMessageHistoryIfAny();
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
@@ -241,8 +241,10 @@ public abstract class AbstractMessageProducingHandler extends AbstractMessageHan
 	}
 
 	protected void sendOutputs(@Nullable Object result, Message<?> requestMessage) {
-		if (!(result instanceof Flux<?>) && result instanceof Iterable<?> iterableResult &&
-				shouldSplitOutput(iterableResult)) {
+
+		if (!(result instanceof Flux<?>)
+				&& result instanceof Iterable<?> iterableResult
+				&& shouldSplitOutput(iterableResult)) {
 			for (Object o : iterableResult) {
 				produceOutput(o, requestMessage);
 			}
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/test/java/org/springframework/integration/handler/AbstractMessageProducingHandlerTests.java
```diff
@@ -0,0 +1,184 @@
...
+	@BeforeEach
+	void setup() {
+		this.handler = new TestMessageProducingHandler();
+		this.beanFactory = mock(BeanFactory.class);
+		this.handler.setBeanFactory(this.beanFactory);
+		this.handler.afterPropertiesSet();
+	}
+
+	@Test
+	void shouldSendOutputToConfiguredChannel() {
```
artembilan: This test is redundant because plethora of other tests are doing exactly the same.

spring-integration-core/src/test/java/org/springframework/integration/handler/AbstractMessageProducingHandlerTests.java
```diff
@@ -0,0 +1,184 @@
...
+
+		StepVerifier.create(result)
+				.expectNext("Hello")
+				.expectNext("World")
+				.thenCancel()
+				.verify();
+	}
+
+	@Test
+	void shouldSendOutputToReplyChannel() {
```
artembilan: This test is also redundant: we have many other places where we verify `replyChannel` header interaction.

spring-integration-core/src/test/java/org/springframework/integration/handler/AbstractMessageProducingHandlerTests.java
```diff
@@ -0,0 +1,184 @@
...
+		QueueChannel outputChannel = new QueueChannel();
+		this.handler.setOutputChannel(outputChannel);
+		this.handler.setAsync(true);
+		CompletableFuture<String> future = new CompletableFuture<>();
+		this.handler.setResult(future);
+		Message<?> inputMessage = MessageBuilder.withPayload("test").build();
+
+		this.handler.handleMessageInternal(inputMessage);
+
+		assertThat(outputChannel.receive(0)).isNull();
```
artembilan: I'm not fully sure why this is `null` here.
According to your `TestMessageProducingHandler` the payload of the message should be emitted as a reply.
And since it is not an async container, it should be emitted normally to the `outputChannel` ignoring that `async` mode.
Please, elaborate.

spring-integration-core/src/test/java/org/springframework/integration/handler/AbstractMessageProducingHandlerTests.java
```diff
@@ -0,0 +1,184 @@
...
+import org.springframework.messaging.Message;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Tests for {@link AbstractMessageProducingHandler}.
+ *
+ * @author Glenn Renfro
+ * @since 7.1
```
artembilan: We are going to back-port the fix down to `6.5.7`

spring-integration-core/src/main/java/org/springframework/integration/handler/AbstractMessageProducingHandler.java
```diff
@@ -241,7 +241,8 @@ public MessageChannel getOutputChannel() {
 	}
 
 	protected void sendOutputs(@Nullable Object result, Message<?> requestMessage) {
-		if (result instanceof Iterable<?> iterableResult && shouldSplitOutput(iterableResult)) {
+		if (!(result instanceof Flux<?>) && result instanceof Iterable<?> iterableResult &&
+				shouldSplitOutput(iterableResult)) {
```
artembilan: If we already here with review, please, make every condition on its own line and add blank line before body of this `if` block.

artembilan (changes requested):
Looks great, but I'm not sure in that `Async(true)` behavior.
Thanks

spring-integration-core/src/test/java/org/springframework/integration/handler/AbstractMessageProducingHandlerTests.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	@Test
+	void shouldHandleCompletableFutureInAsyncMode() throws Exception {
+		QueueChannel outputChannel = new QueueChannel();
+		this.handler.setOutputChannel(outputChannel);
+		this.handler.setAsync(true);
+		CompletableFuture<String> future = new CompletableFuture<>();
+		this.handler.setResult(future);
+		Message<?> inputMessage = MessageBuilder.withPayload("test").build();
```
artembilan: I think we can just send completed future as a payload and get rid off of that `result` in the test handler.

PR #10828, opened 2026-02-19
Add CloudEventHeadersBuilder DSL support

cppwfs:
Introduce `CloudEventHeadersBuilder` to allow setting CloudEvent headers via the Spring Integration DSL using direct values, SpEL expressions, or Java functions. Expose the builder through a `CloudEvents.headers()` factory method.


spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,250 @@
...
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.integration.cloudevents.CloudEventHeaders;
+import org.springframework.integration.expression.FunctionExpression;
+import org.springframework.integration.support.MapBuilder;
+import org.springframework.messaging.Message;
+
+/**
+ * The CloudEvent specific {@link MapBuilder} implementation.
+ * <p>Provide a fluent API for building CloudEvent headers with support for
```
artembilan: I prefer in class Javadocs to have `<p>` on its own line.
That way the real paragraph is more readable from the core.
See `GrpcInboundGateway` for example.
Jut minor, of course, but still friendly to deal with our code.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,250 @@
...
+ * literal values, SpEL expressions, and function-based values.
+ *
+ * @author Glenn Renfro
+ *
+ * @since 7.1
+ */
+public class CloudEventHeadersBuilder extends MapBuilder<CloudEventHeadersBuilder, String, Object> {
+
+	CloudEventHeadersBuilder() {
```
artembilan: I think it is fine to have this a `public`.

Also, please, consider to have a `prefix` as a ctor arg.
The default one would indeed fall to the `CloudEventHeaders.PREFIX`, but another one for end-user convenience to do respective choice in their project.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,250 @@
...
+	public CloudEventHeadersBuilder source(URI source) {
+		return put(CloudEventHeaders.EVENT_SOURCE, source);
+	}
+
+	/**
+	 * Set the CloudEvent source.
+	 * @param source the event source as string
+	 * @return the builder
+	 */
+	public CloudEventHeadersBuilder source(String source) {
```
artembilan: I think this is wrong since `URI` one is what we want to have only.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEvents.java
```diff
@@ -50,23 +49,31 @@ public static FromCloudEventTransformer fromCloudEventTransformer(EventFormat ev
...
 	 */
 	public static ToCloudEventTransformerSpec toCloudEventTransformer(String... extensionPatterns) {
 		return new ToCloudEventTransformerSpec(extensionPatterns);
 	}
 
+	/**
+	 * Create a {@link CloudEventHeadersBuilder}.
+	 * @return the CloudEventHeadersBuilder instance
+	 */
+	public static CloudEventHeadersBuilder headers() {
```
artembilan, line 74: And, please, overloaded one with the custom `prefix`.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/FromCloudEventTransformer.java
```diff
@@ -125,10 +125,10 @@ protected Object doTransform(Message<?> message) {
 				getMessageBuilderFactory()
 						.withPayload(payload)
 						.copyHeaders(message.getHeaders())
+						.setHeader(MessageHeaders.CONTENT_TYPE, cloudEvent.getDataContentType())
 						.setHeader(CloudEventHeaders.EVENT_SOURCE, cloudEvent.getSource())
 						.setHeader(CloudEventHeaders.EVENT_TYPE, cloudEvent.getType())
 						.setHeader(CloudEventHeaders.EVENT_ID, cloudEvent.getId())
-						.setHeader(MessageHeaders.CONTENT_TYPE, cloudEvent.getDataContentType())
```
artembilan, line 131: Why the change in this class?

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilderIntegrationTests.java
```diff
@@ -0,0 +1,142 @@
...
+							.source(URI.create("https://example.com"))
+							.subject("test-subject"))
+					.channel("outputChannel");
+		}
+
+		@Bean
+		IntegrationFlow functionFlow() {
+			return flow -> flow
+					.enrichHeaders(CloudEvents.headers()
+							.dataSchema(URI.create("https://example.com"))
```
artembilan: Why no function here?

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilderIntegrationTests.java
```diff
@@ -0,0 +1,142 @@
...
+					.channel("outputChannel");
+		}
+
+		@Bean
+		IntegrationFlow expressionFlow() {
+			return flow -> flow
+					.enrichHeaders(CloudEvents.headers()
+							.idExpression("headers['orderId']")
+							.typeExpression("'order.' + headers['type']")
+							.sourceExpression("'https://example.com/' + headers[ 'type']"))
```
artembilan: I'm not sure in space before header name in the expression...
Do you know that we can simply do `headers.type`?
As well as for all those expression for header extractions.
See `MapAccessor` for details.

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilderIntegrationTests.java
```diff
@@ -0,0 +1,142 @@
...
+@DirtiesContext
+class CloudEventHeadersBuilderIntegrationTests {
+
+	@Autowired
+	@Qualifier("outputChannel")
+	private PollableChannel outputChannel;
+
+	@Test
+	void testEnrichHeadersWithDirectValues(
+			@Autowired @Qualifier("directFlow.input") MessageChannel inputChannel) {
```
artembilan: Why from a new line when all other tests are single line method header?

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -342,3 +342,35 @@ public IntegrationFlow fromCloudEventFlow(FromCloudEventTransformer fromCloudEve
...
+[[cloudevent-headers-builder]]
+== CloudEvent Headers Builder
+
+Use the `CloudEventHeadersBuilder` to set CloudEvent headers using Spring Integration's DSL.
+Access the builder through the `CloudEvents.headers()` factory method.
+
+The builder supports three ways to set header values:
+
+1. **Direct values** - Set static values directly
+2. **SpEL expressions** - Use Spring Expression Language for dynamic values
```
artembilan: We need to mention that request `Message` is the root evaluation context object.

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -342,3 +342,35 @@ public IntegrationFlow fromCloudEventFlow(FromCloudEventTransformer fromCloudEve
...
+3. **Functions** - Provide Java functions that evaluate the values for the attributes
```
artembilan: We need to mention that request `Message` is an input for the function.

artembilan (changes requested):
Looks good!

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/transformer/FromCloudEventTransformer.java
cppwfs: Move the `MessageHeader.CONTENT_TYPE` to the top thus having the `CloudEventHeaders.*` as a group for readability.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -38,7 +39,27 @@
...
+		prefix(prefix);
+	}
+
+	/**
+	 * Set the CloudEvent prefix.
+	 * @param prefix the event prefix
+	 * @return the builder
+	 */
+	public CloudEventHeadersBuilder prefix(String prefix) {
+		return put(CloudEventHeaders.PREFIX, prefix);
```
artembilan: I don't think this is correct.
Here you put the prefix as a map entry. 
However we have to add all the CloudEvent attributes with that provided prefix.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,262 @@
...
+	 */
+	public CloudEventHeadersBuilder() {
+	}
+
+	/**
+	 * Create a new {@link CloudEventHeadersBuilder} with the given prefix.
+	 * @param prefix the CloudEvent header prefix
+	 */
+	public CloudEventHeadersBuilder(String prefix) {
+		prefix(prefix);
```
artembilan: The `prefix`  has to go to class property.
And that property has to be used in attributes API instead of hard-coded constants with default `ce-` prefix.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEvents.java
artembilan: I don't see `headers(String prefix)` overload.

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilderIntegrationTests.java
```diff
@@ -0,0 +1,143 @@
...
+							.sourceFunction(msg -> URI.create("https://example.com/" + msg.getHeaders().get("action"))))
+					.channel("outputChannel");
+		}
+
+		@Bean
+		IntegrationFlow expressionFlow() {
+			return flow -> flow
+					.enrichHeaders(CloudEvents.headers()
+							.idExpression("headers['orderId']")
+							.typeExpression("'order.' + headers['type']")
```
artembilan: Why do you use `MapAccessor` style only on one place?
Why these two were not updated?

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -342,3 +342,35 @@ public IntegrationFlow fromCloudEventFlow(FromCloudEventTransformer fromCloudEve
...
+=== Basic Usage
+
+[source,java]
+----
+@Bean
+public IntegrationFlow cloudEventFlow() {
+    return IntegrationFlow
+        .from("inputChannel")
+        .enrichHeaders(CloudEvents.headers()
+            .idExpression("headers['orderId']")
```
artembilan: May we use `MapAccessor` style in the doc as well, please?

artembilan (changes requested):

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+	public CloudEventHeadersBuilder id(String id) {
+		return put(this.eventIdKey, id);
+	}
+
+	/**
+	 * Set the CloudEvent prefix.
+	 * @param prefix the event prefix
+	 * @return the builder
+	 */
+	public CloudEventHeadersBuilder prefix(String prefix) {
```
artembilan: No. I still don't believe in this method.
Why to give a paradox of choice to user?
There is a ctor requesting such a prefix.
So, on any other mutators, please.

It also may cause the problem.
Let's imaging I have started with a default prefix!
Gave it a couple attributes, then decided to change the prefix.
But only the others after this method are going to be modified.
That is an inconsistency, which is possible, but probably with a generic `enrichHeaders` API, not with this builder for its own DSL-specific case.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+	private String eventSubjectKey = CloudEventHeaders.EVENT_SUBJECT;
+
+	private String eventTimeKey = CloudEventHeaders.EVENT_TIME;
+
+	private String eventDataContentTypeKey = CloudEventHeaders.EVENT_DATA_CONTENT_TYPE;
+
+	private String eventDataSchemaKey = CloudEventHeaders.EVENT_DATA_SCHEMA;
+
+	/**
+	 * Create a new {@link CloudEventHeadersBuilder}.
```
artembilan: `... with default prefix` ?

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,292 @@
...
+	public CloudEventHeadersBuilder() {
+	}
+
+	/**
+	 * Create a new {@link CloudEventHeadersBuilder} with the given prefix.
+	 * @param prefix the CloudEvent header prefix
+	 */
+	public CloudEventHeadersBuilder(String prefix) {
+		this.prefix = prefix;
+		updatePrefix();
```
artembilan: I think we don't need this extra method and do all the properties modification directly here.
I also still believe in some compiler optimization when we declare properties as `final`.
We probably won't need `prefix` after modified the rest of attributes here on the fly.

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEvents.java
```diff
@@ -50,24 +50,42 @@ public static FromCloudEventTransformer fromCloudEventTransformer(EventFormat ev
...
-	 *                          to the {@link io.cloudevents.CloudEvent}
+	 * to the {@link io.cloudevents.CloudEvent}
 	 * @return the {@link ToCloudEventTransformerSpec} instance
 	 */
 	public static ToCloudEventTransformerSpec toCloudEventTransformer(String... extensionPatterns) {
 		return new ToCloudEventTransformerSpec(extensionPatterns);
 	}
 
+	/**
+	 * Create a {@link CloudEventHeadersBuilder}.
```
artembilan: `... with default prefix` ?

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilderIntegrationTests.java
```diff
@@ -0,0 +1,171 @@
...
+@DirtiesContext
+class CloudEventHeadersBuilderIntegrationTests {
+
+	@Autowired
+	@Qualifier("outputChannel")
+	private PollableChannel outputChannel;
+
+	@Test
+	void testEnrichHeadersWithDirectValues(@Autowired @Qualifier("directFlow.input") MessageChannel inputChannel) {
+
```
artembilan: So, we don't need this blank line then?
Just minor, of course, but I wonder if we try to pursue consistence and perfect code.

spring-integration-cloudevents/src/test/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilderIntegrationTests.java
```diff
@@ -0,0 +1,171 @@
...
+							.source(URI.create("https://example.com"))
+							.subject("test-subject"))
+					.channel("outputChannel");
+		}
+
+		@Bean
+		IntegrationFlow functionFlow() {
+			return flow -> flow
+					.enrichHeaders(CloudEvents.headers()
+							.dataSchemaFunction(msg -> URI.create("https://example.com/schema/" +
```
artembilan: I would prefer to put the whole lambda body on a new line, rather than break on string concatenation.

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
artembilan: Was not addressed.
Or just comment here, please, why you have not changed it, but approved.

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/cloudevents.adoc
```diff
@@ -342,3 +342,35 @@ public IntegrationFlow fromCloudEventFlow(FromCloudEventTransformer fromCloudEve
         .get();
 }
 ----
+
+[[cloudevent-headers-builder]]
+== CloudEvent Headers Builder
+
+Use the `CloudEventHeadersBuilder` to set CloudEvent headers using Spring Integration's DSL.
+Access the builder through the `CloudEvents.headers()` factory method.
```
artembilan: Not saying that it is critical.
I feel discomfort with these two sentences.
Both of them are kind of commands I have to do.
Even if they don't use word `you`, it is still implied.
And this doc gives me some instructions.
In my understanding it has to explain a behavior of the API, not commands what and how to use.
Makes sense?
Or do I exaggerate?

artembilan (changes requested):

spring-integration-cloudevents/src/main/java/org/springframework/integration/cloudevents/dsl/CloudEventHeadersBuilder.java
```diff
@@ -0,0 +1,274 @@
...
+	public <P> CloudEventHeadersBuilder typeFunction(Function<Message<P>, String> type) {
+		return put(this.eventTypeKey, new FunctionExpression<>(type));
+	}
+
+	/**
+	 * Set the CloudEvent time.
+	 * @param time the event timestamp
+	 * @return the builder
+	 */
+	@SuppressWarnings("NullAway")
```
artembilan: Please, rebase to the latest main and remove these warnings since nullability is there already.
Thanks

artembilan (changes requested):
This is good now.
Just one concern with rebasing and redundant now suppress annotations.
Thanks

PR #10829, opened 2026-02-19
Add null value support to MapBuilder

cppwfs:
- Add null value support to  StaticHeaderValueMessageProcessor

Fixes: https://github.com/spring-projects/spring-integration/issues/10827

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
@@ -42,11 +42,12 @@ import org.springframework.messaging.Message;
  *
  * @author Mark Fisher
  * @author Artem Bilan
+ * @author Glenn Renfro
  *
  * @since 2.0
  */
 @FunctionalInterface
-public interface MessageProcessor<T> {
+public interface MessageProcessor<T extends @Nullable Object> {
 
 	/**
 	 * Process the Message and return a value (or null).
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
@@ -42,7 +42,6 @@ import org.springframework.messaging.Message;
  *
  * @author Mark Fisher
  * @author Artem Bilan
- * @author Glenn Renfro
  *
  * @since 2.0
  */
--- a/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
@@ -47,7 +47,7 @@ import org.springframework.messaging.Message;
  * @since 2.0
  */
 @FunctionalInterface
-public interface MessageProcessor<T extends @Nullable Object> {
+public interface MessageProcessor<T> {
 
 	/**
 	 * Process the Message and return a value (or null).
--- a/spring-integration-core/src/main/java/org/springframework/integration/transformer/support/StaticHeaderValueMessageProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/transformer/support/StaticHeaderValueMessageProcessor.java
@@ -30,16 +30,16 @@ import org.springframework.messaging.Message;
  *
  * @since 3.0
  */
-public class StaticHeaderValueMessageProcessor<T> extends AbstractHeaderValueMessageProcessor<T> {
+public class StaticHeaderValueMessageProcessor<T extends @Nullable Object> extends AbstractHeaderValueMessageProcessor<T> {
 
-	private final @Nullable T value;
+	private final T value;
 
-	public StaticHeaderValueMessageProcessor(@Nullable T value) {
+	public StaticHeaderValueMessageProcessor(T value) {
 		this.value = value;
 	}
 
 	@Override
-	public @Nullable T processMessage(Message<?> message) {
+	public T processMessage(Message<?> message) {
 		return this.value;
 	}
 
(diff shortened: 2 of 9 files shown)


spring-integration-core/src/main/java/org/springframework/integration/transformer/support/StaticHeaderValueMessageProcessor.java
```diff
@@ -16,26 +16,30 @@
...
  *
  * @since 3.0
  */
 public class StaticHeaderValueMessageProcessor<T> extends AbstractHeaderValueMessageProcessor<T> {
 
-	private final T value;
+	private final @Nullable T value;
 
-	public StaticHeaderValueMessageProcessor(T value) {
+	public StaticHeaderValueMessageProcessor(@Nullable T value) {
```
artembilan: Will it work if we move it to the generic argument definition instead:
```
 AbstractHeaderValueMessageProcessor<T extends @Nullable Object>
```
?

See for example `CheckedCallable`

spring-integration-core/src/test/java/org/springframework/integration/transformer/MethodInvokingTransformerTests.java
```diff
@@ -266,6 +270,36 @@ void headerEnricherConfiguredWithMethodName() {
...
+				.containsKey(MessageHeaders.TIMESTAMP);
+
+		result = getResultMessage(new StaticHeaderValueMessageProcessor<>(Collections.singletonMap("test", "replace")));
+		assertThat(result.getPayload()).isEqualTo("test");
+		assertThat(result.getHeaders()).containsEntry("test", "replace")
+				.containsKey(MessageHeaders.ID)
+				.containsKey(MessageHeaders.TIMESTAMP);
+	}
+
+	private Message<?> getResultMessage(StaticHeaderValueMessageProcessor<?> staticHeaderValueMessageProcessor) {
```
artembilan: `static` ?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
```diff
@@ -42,11 +42,12 @@
  *
  * @author Mark Fisher
  * @author Artem Bilan
+ * @author Glenn Renfro
  *
  * @since 2.0
  */
 @FunctionalInterface
-public interface MessageProcessor<T> {
+public interface MessageProcessor<T extends @Nullable Object> {
```
artembilan: Can we then remove `@Nullable` on the `processMessage()` method?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
cppwfs: It is not a simple change.  It has significant changes downstream.  We can chat.

artembilan: Ok. Can we consider then do not do that here, but only in headers enricher API?
I mean just revert your change for this class altogether

cppwfs: Interesting.   
The IDE marks the  extends MessageProcessor<T> in `HeaderValueMessageProcessor` with "Non Null Type argument expected" when I remove the extends in `MessageProcessor`.
However tests and build show no warnings or errors.   A possible issue with the IDE.

spring-integration-core/src/main/java/org/springframework/integration/handler/MessageProcessor.java
```diff
@@ -42,6 +42,7 @@
  *
  * @author Mark Fisher
  * @author Artem Bilan
+ * @author Glenn Renfro
```
artembilan: Sorry, but looks like this is only change in the class 😄

spring-integration-core/src/test/java/org/springframework/integration/transformer/MethodInvokingTransformerTests.java
```diff
@@ -266,6 +270,36 @@ void headerEnricherConfiguredWithMethodName() {
 		assertThat(result.getHeaders().get("prop3")).isEqualTo("baz");
 	}
 
+	@Test
+	void headerEnricherStaticHeaderValueMessageProcessorNullCheck() {
```
artembilan, line 273: I don't think this test is correct from what we would like to cover.
The idea is to test `StaticHeaderValueMessageProcessor` with null for an individual entry, but that `transformer.setMessageProcessor(staticHeaderValueMessageProcessor);` is a general processor which is expected to produce a map.
So, your `new StaticHeaderValueMessageProcessor<>(null)` here, would lead to that `expected a Map result from processor, but received: ` in the end of the `HeaderEnricher.addHeadersFromMessageProcessor()`.
But we really want to cover that:
```
				if (value != null || !this.shouldSkipNulls) {
					messageBuilder.setHeader(key, value);
				}
```
for individual entries.

artembilan (changes requested):
I think the `HeaderEnricherSpec` API is missing from this change.
I believe that its `public HeaderEnricherSpec headers(MapBuilder<?, String, Object> headers) {` should be null-aware according to the expectations of this change.
And so on down to the individual entries:
```
	public <V> HeaderEnricherSpec header(String name, @Nullable V value, @Nullable Boolean overwrite) {
		AbstractHeaderValueMessageProcessor<V> headerValueMessageProcessor =
				new StaticHeaderValueMessageProcessor<>(value);
		headerValueMessageProcessor.setOverwrite(overwrite);
		return header(name, headerValueMessageProcessor);
	}
```
That's where the power of the nullability on the `StaticHeaderValueMessageProcessor` would be triggered.

artembilan (approved):

PR #10836, opened 2026-02-26
Replace TimeUnit with Duration in Redis queue operations

cppwfs:
The Redis queue components were using the deprecated TimeUnit-based API


spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisQueueInboundGateway.java
```diff
@@ -191,7 +191,7 @@ private void handlePopException(Exception e) {
 	private void receiveAndReply() {
 		byte[] value;
 		try {
-			value = this.boundListOperations.rightPop(this.receiveTimeout, TimeUnit.MILLISECONDS);
+			value = this.boundListOperations.rightPop(Duration.ofMillis(this.receiveTimeout));
```
artembilan: As we discussed, let's have an additional `setReceiveTimeout(Duration)`!

artembilan (changes requested):
Thank you!

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisQueueMessageDrivenEndpoint.java
```diff
@@ -241,10 +241,10 @@ private void popMessageAndSend() {
 		byte[] value = null;
 		try {
 			if (this.rightPop) {
-				value = this.boundListOperations.rightPop(this.receiveTimeout, TimeUnit.MILLISECONDS);
+				value = this.boundListOperations.rightPop(Duration.ofMillis(this.receiveTimeout));
```
artembilan: Any clues why this class was not fixed same way as others?

artembilan (changes requested):

artembilan (changes requested):
Please, rebase to the latest `main`.
We got some fixes according to the breaking changes from upstream projects.
Thanks

spring-integration-redis/src/main/java/org/springframework/integration/redis/inbound/RedisQueueInboundGateway.java
```diff
@@ -123,20 +127,62 @@ public void setSerializer(RedisSerializer<?> serializer) {
...
+	 * <p>
+	 * A timeout of zero can be used to block indefinitely. If not set explicitly
+	 * the timeout value will default to {@code 1000 millis}
+	 * <p>
+	 * See also: https://redis.io/commands/brpop
+	 * @param receiveTimeout String containing the receive timeout as a string serialized (millis) or
+	 * ISO-8601 duration format.
+	 * @since 7.1
+	 */
+	public void setReceiveTimeout(String receiveTimeout) {
```
artembilan: This is nice, but not what should to be done in this component.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/DurationUtil.java
```diff
@@ -0,0 +1,70 @@
...
+ * associated time units.
+ *
+ * TODO: There is probably a better place to put this, or is implemented somewhere else.  This code was lift and shifted
+ * from PeriodicTriggerFactoryBean
+ *
+ * @author Glenn Renfro
+ *
+ * @since 7.1
+ */
+public abstract class DurationUtil {
```
artembilan: I think we don't need this utility.
Looks like there is one in SF already: `DurationFormatterUtils`.
I still don't see a `PropertyEditor` for property placeholders, but that's different story.

I guess we can fix that `PeriodicTriggerFactoryBean`, respectively, but we definitely don't need this utility in our project.
And I also believe that we should not provide string-based setters: let end-user to do the parsing before calling our component!

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/DurationUtil.java
cppwfs: The issue that I ran into is that the XML Bean definitions pass the receiveTimeout as a String and they can't be converted to a Duration.   For some reason it would not use the long overload.   Any suggestions?

artembilan: Yeah... That's exactly why we need a `ProperyEditor` for `Duration`.
We can raise an issue for SF.
There is also another problem for XML bean definitions: if we have overloaded setters, there is no guarantee which one would be tried by JavaBeans introspection.
The solution is to not have overloaded setters, but name a new one as `setReceiveDuration(Duration)`.
Makes sense?

artembilan (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/DurationUtil.java
cppwfs: Sure does!

I'll rename it.

PR #10838, opened 2026-02-26
GH-6792: Add headers mapping to JMS channels

### LINKED ISSUE #6792: Add header-mapper to int-jms:channel [INT-2816]
**[William Gorder](https://jira.spring.io/secure/ViewProfile.jspa?name=wgorder)** opened **[INT-2816](https://jira.spring.io/browse/INT-2816?redirect=false)** and commented

We are using a header mapper to get the amount of retries off of an ActiveMqMessage.

```
/**
* A custom JMS header mapper extending the functionality provided by {@link DefaultJmsHeaderMapper}
* to add a header defining the re-delivery count
* 
 */
@Component("activeMQRedeliveryCounterMapper")
public class ActiveMQRedeliveryCounterMapper extends DefaultJmsHeaderMapper
{
    /**
     * For ActiveMQ, the redeliveryCounter is not a message property, it is accessed by the
     * proprietary getRedeliveryCounter() method. This means a custom header mapper is required to
     * gain access to it as an SI header.
     */
    @Override
    public Map<String, Object> toHeaders(Message jmsMessage)
    {
        Map<String, Object> headers = super.toHeaders(jmsMessage);
        if (jmsMessage instanceof ActiveMQMessage)
        {
            ActiveMQMessage amqMessage = (ActiveMQMessage) jmsMessage;
            int redeliveryCounter = amqMessage.getRedeliveryCounter();
            headers.put("amq_redeliveryCounter", redeliveryCounter);
        }
        return headers;
    }
}
```

Our goal was to get away from using a combination of int-jms:output-channel-adapter and the int-jms:message-driven-channel-adapter and just use jms backed channels.  However it does not seem that it is possible to define a header mapper on this component.  Is there a workaround for this or a reason it is not available?

---

**Affects:** 2.1.4

2 votes, 4 watchers

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/6792

Currently, by default `AbstractJmsChannel` implementations use Java serialization for storing `Message<?>` instance into JMS destination. The deserialized message is produced from the channel as is. There might be use-cases when some JMS properties could be useful in downstream flows.

* Add `DefaultJmsHeaderMapper` into the `AbstractJmsChannel` as a conditional logic to map headers to/from JMS when `jmsTemplate.getMessageConverter()` is not a `MessagingMessageConverter`. Otherwise, then conversion of the message and header mapping is done in that converter
* Refactor logic in the `AbstractJmsChannel` to common API
* Fix Nullability formatting in the `PollableJmsChannel`
* Use lambda style for `MessageListener` in the `SubscribableJmsChannel` instead of inner extra class
* Ensure in the `PollableJmsChannelTests` and `SubscribableJmsChannelTests` that header mapper does it job
* Fix `JmsChannelParserTests` to not assert on the `messageBuilderFactory` since now a `MessageListener` in the `SubscribableJmsChannel` is a direct method of the class
* Fix `JmsChannelHistoryTests` verification according to a new internal logic of the `AbstractJmsChannel`
* Document new feature in the `jms.adoc` and `whats-new.adoc`

### DIFF
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
@@ -16,11 +16,18 @@
 
 package org.springframework.integration.jms.channel;
 
+import jakarta.jms.JMSException;
+
 import org.springframework.integration.IntegrationMessageHeaderAccessor;
 import org.springframework.integration.channel.AbstractMessageChannel;
+import org.springframework.integration.jms.DefaultJmsHeaderMapper;
 import org.springframework.integration.jms.DynamicJmsTemplateProperties;
 import org.springframework.jms.core.JmsTemplate;
+import org.springframework.jms.support.converter.MessageConverter;
+import org.springframework.jms.support.converter.MessagingMessageConverter;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.MessagingException;
 import org.springframework.util.Assert;
 
 /**
@@ -28,6 +35,7 @@ import org.springframework.util.Assert;
  *
  * @author Mark Fisher
  * @author Gary Russell
+ * @author Artem Bilan
  *
  * @since 7.0
  *
@@ -36,27 +44,62 @@ import org.springframework.util.Assert;
  */
 public abstract class AbstractJmsChannel extends AbstractMessageChannel {
 
-	private final JmsTemplate jmsTemplate;
+	protected final DefaultJmsHeaderMapper headerMapper = new DefaultJmsHeaderMapper();
+
+	protected final JmsTemplate jmsTemplate;
 
 	public AbstractJmsChannel(JmsTemplate jmsTemplate) {
 		Assert.notNull(jmsTemplate, "jmsTemplate must not be null");
 		this.jmsTemplate = jmsTemplate;
 	}
 
-	JmsTemplate getJmsTemplate() {
-		return this.jmsTemplate;
-	}
-
 	@Override
 	protected boolean doSend(Message<?> message, long timeout) {
 		try {
 			DynamicJmsTemplateProperties.setPriority(new IntegrationMessageHeaderAccessor(message).getPriority());
-			this.jmsTemplate.convertAndSend(message);
+			MessageConverter messageConverter = this.jmsTemplate.getMessageConverter();
+			this.jmsTemplate.send((session) -> {
+				jakarta.jms.Message jmsMessage = messageConverter.toMessage(message, session);
+				if (!(messageConverter instanceof MessagingMessageConverter)) {
+					MessageHeaders headers = message.getHeaders();
+					this.headerMapper.fromHeaders(headers, jmsMessage);
+				}
+				return jmsMessage;
+			});
 		}
 		finally {
 			DynamicJmsTemplateProperties.clearPriority();
 		}
 		return true;
+	}
+
+	protected Message<?> fromJmsMessage(jakarta.jms.Message message) {
+		MessageConverter converter = this.jmsTemplate.getMessageConverter();
+		try {
+			Object converted = converter.fromMessage(message);
+			Message<?> messageToSend;
+			if (converted instanceof Message<?> convertedMessage) {
+				messageToSend = convertedMessage;
+				if (!(converter instanceof MessagingMessageConverter)) {
+					messageToSend = getMessageBuilderFactory()
+							.fromMessage(messageToSend)
+							.copyHeadersIfAbsent(this.headerMapper.toHeaders(message))
+							.build();
+				}
+			}
+			else {
+				messageToSend = getMessageBuilderFactory()
+						.withPayload(converted)
+						.copyHeaders(this.headerMapper.toHeaders(message))
+						.build();
+			}
+
+			return messageToSend;
+		}
+		catch (JMSException ex) {
+			throw new MessagingException("failed to convert incoming JMS Message", ex);
+		}
+
 	}
 
 }
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/PollableJmsChannel.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/PollableJmsChannel.java
@@ -123,9 +123,7 @@ public class PollableJmsChannel extends AbstractJmsChannel
 
 		Message<?> message = fromJmsMessage(jmsMessage);
 
-		if (logger.isDebugEnabled()) {
-			logger.debug("postReceive on channel '" + this + "', message: " + message);
-		}
+		logger.debug(() -> "postReceive on channel '" + this + "', message: " + message);
 
 		return message;
 	}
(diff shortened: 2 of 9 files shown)


spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
```diff
@@ -36,27 +44,62 @@
...
+		MessageConverter converter = this.jmsTemplate.getMessageConverter();
+		try {
+			Object converted = converter.fromMessage(message);
+			Message<?> messageToSend;
+			if (converted instanceof Message<?> convertedMessage) {
+				messageToSend = convertedMessage;
+				if (!(converter instanceof MessagingMessageConverter)) {
+					messageToSend = getMessageBuilderFactory()
+							.fromMessage(messageToSend)
+							.copyHeadersIfAbsent(this.headerMapper.toHeaders(message))
```
cppwfs, line 86: Would a user want the original header overwritten?   Is this an option we want to add?

spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/PollableJmsChannel.java
```diff
@@ -109,24 +107,25 @@ public Message<?> receive() {
...
 		}
-		Message<?> message = object instanceof Message<?> msg
-				? msg
-				: getMessageBuilderFactory().withPayload(object).build();
-		logger.debug(() -> "postReceive on channel '" + this + "', message: " + message);
+
+		Message<?> message = fromJmsMessage(jmsMessage);
+
+		if (logger.isDebugEnabled()) {
+			logger.debug("postReceive on channel '" + this + "', message: " + message);
```
cppwfs: Could we do this?
`logger.debug(() -> "postReceive on channel '" + this + "', message: " + message);`

spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/SubscribableJmsChannel.java
```diff
@@ -140,115 +130,54 @@ private void configureDispatcher(boolean isPubSub) {
...
-
-			this.jmsTemplate = jmsTemplate;
-			this.dispatcher = dispatcher;
-			this.channel = channel;
-			this.isPubSub = isPubSub;
-			this.messageBuilderFactory = messageBuilderFactory;
+	private void receiveAndDispatch(jakarta.jms.Message message) {
+		Message<?> messageToSend = fromJmsMessage(message);
+		try {
+			this.dispatcher.dispatch(messageToSend);
```
cppwfs, line 165: I noticed in the `SubscribableJmsChannel` that we have several unnecessary null checks for `this.dispatcher` that can be removed.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -54,3 +54,9 @@ See xref:testing.adoc[] for more information.
 The `RedisMessageStore.doRemove` now uses `GETDEL` instead of `GET` + `UNLINK` for Redis 6.2+ by default.
 Use `RedisMessageStore.setUseUnlink(true)` to use `GET` + `UNLINK` when atomicity is not required and `GETDEL` causes noticeable Redis latency.
 See xref:redis.adoc[] for more information.
+
+[[x7.1-jms-changes]]
+=== JMS Support Changes
+
+JMS-backed message channels now map headers to JMS message properties and back.
+See xref:jms.adoc[] for more information.
```
cppwfs, line 62: Blank line at the end of the file.

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/PollableJmsChannelTests.java
```diff
@@ -200,13 +203,13 @@ public void qos() throws Exception {
...
+		Destination queue = new ActiveMQQueue("pollableJmsChannelSelectorTestQueue");
 
 		JmsChannelFactoryBean factoryBean = new JmsChannelFactoryBean(false);
 		CachingConnectionFactory ccf = new CachingConnectionFactory(connectionFactory);
 		ccf.setCacheConsumers(false);
 		factoryBean.setConnectionFactory(ccf);
-		factoryBean.setDestination(this.queue);
+		factoryBean.setDestination(queue);
 
 		factoryBean.setMessageSelector("property='value'");
```
cppwfs, line 214: Do we want to check if this is in the headers?

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/SubscribableJmsChannelTests.java
```diff
@@ -100,15 +103,22 @@ public void queueReference() throws Exception {
...
+		assertThat(receivedList1.get(0))
+				.extracting(Message::getPayload)
+				.isEqualTo("test1");
+		assertThat(receivedList1.get(0).getHeaders())
+				.containsEntry(MessageUtil.JMSXDELIVERYCOUNT, 1)
+				.containsEntry(JmsHeaders.DESTINATION, this.queue)
+				.containsEntry("customHeader", "customValue");
+		assertThat(receivedList2).hasSize(1);
+		assertThat(receivedList2.get(0))
+				.extracting(Message::getPayload)
```
cppwfs, line 120: Should we check for `JMSXDELIVERYCOUNT` here as well?

spring-integration-jms/src/test/java/org/springframework/integration/jms/config/JmsChannelParserTests.java
```diff
@@ -176,8 +176,6 @@ public void topicReferenceChannel() {
 				"container");
 		assertThat(jmsTemplate.getDefaultDestination()).isEqualTo(topic);
 		assertThat(container.getDestination()).isEqualTo(topic);
-		assertThat(TestUtils.<Object>getPropertyValue(channel, "container.messageListener.messageBuilderFactory"))
```
cppwfs, line 179: Why was this test removed?
Is it unnecessary?

The property is still present:
```
assertThat(TestUtils.<Object>getPropertyValue(channel, "messageBuilderFactory"))
		.isSameAs(this.messageBuilderFactory);
```

cppwfs (changes requested):
Great work.   Just a couple of nitpicks!

spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
artembilan: Yeah...
First of all I don't want to expose headers manipulation options from these channels at all.
And secondly, I followed exactly same logic as `MessagingMessageConverter`.
This our logic here is to mimic that converter when `Message` is deserialized as is without any JMS headers.
If header is present there in the message, then it is indeed end-user preference.
If something else should be done, the custom `MessageConverter` is welcome.
Just not header mapping from the channel configuration.

spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/PollableJmsChannel.java
artembilan: Yes. Now we can. 
That was leftover from some other my previous solution.
Thanks for spotting!

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/SubscribableJmsChannelTests.java
artembilan: I don't think so.
It is going to be another message (or even the same) from the same channel.
I don't see a way how this message copy could come without a header with our unconditional logic.
On the other hand, I'll add that check anyway. Just to fulfill your request. Not sure what we should do to the logic to make it failing in the future.

cppwfs (approved):
Love the cleanup above and beyond what I requested.   This looks great!

LGTM!

PR #10855, opened 2026-03-12
GH-10854: Fix remote file long name for SFTP

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10854

The logic in the framework is to use a long file name (preferably, the whole path) for metadata key. This was implemented via: https://github.com/spring-projects/spring-integration/issues/10162

Fix `SftpSession.doList()` not consulting with whatever `SftpClient` returns but rather using a remote dir as long name prefix unconditionally. This would make it exactly what we claim in docs and consistent with FTP

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
@@ -126,16 +126,13 @@ public class SftpSession implements Session<SftpClient.DirEntry> {
 			}
 		}
 		remoteDir = normalizePath(remoteDir);
-		String remoteDirToUse = remoteDir;
-		return StreamSupport.stream(this.sftpClient.readDir(remoteDirToUse).spliterator(), false)
+		String longNameDirPrefix = remoteDir.endsWith("/") ? remoteDir : remoteDir + '/';
+		return StreamSupport.stream(this.sftpClient.readDir(remoteDir).spliterator(), false)
 				.filter((entry) -> !isPattern || PatternMatchUtils.simpleMatch(remoteFile, entry.getFilename()))
 				.map((entry) -> {
-					String longFilename = entry.getLongFilename();
-					if (StringUtils.hasText(longFilename)) {
-						return entry;
-					}
+
 					String filename = entry.getFilename();
-					return new SftpClient.DirEntry(filename, remoteDirToUse + '/' + filename, entry.getAttributes());
+					return new SftpClient.DirEntry(filename, longNameDirPrefix + filename, entry.getAttributes());
 				});
 	}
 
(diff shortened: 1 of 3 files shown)


spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
```diff
@@ -126,16 +126,13 @@ public Stream<SftpClient.DirEntry> doList(@Nullable String path) throws IOExcept
...
-		return StreamSupport.stream(this.sftpClient.readDir(remoteDirToUse).spliterator(), false)
+		String longNameDirPrefix = remoteDir.endsWith("/") ? remoteDir : remoteDir + '/';
+		return StreamSupport.stream(this.sftpClient.readDir(remoteDir).spliterator(), false)
 				.filter((entry) -> !isPattern || PatternMatchUtils.simpleMatch(remoteFile, entry.getFilename()))
 				.map((entry) -> {
-					String longFilename = entry.getLongFilename();
-					if (StringUtils.hasText(longFilename)) {
-						return entry;
-					}
+
```
cppwfs: Remove blank line.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpServerTests.java
```diff
@@ -159,13 +163,16 @@ protected void doTest(SshServer server, Session<SftpClient.DirEntry> session) th
 			}
 		}
 
-		session.write(new ByteArrayInputStream("foo".getBytes()), "bar");
+		session.write(new ByteArrayInputStream("test data".getBytes()), "test_file");
 		list = session.list(".");
-		assertThat(list[1].getFilename()).isEqualTo("bar");
+		assertThat(list[1].getFilename()).isEqualTo("test_file");
+		// The root dir (.) becomes '/' after sftpClient.canonicalPath()
+		// Plus with the SftpConstants.SFTP_V3 settings, we ensure here that SftpSession overrides file long name
```
cppwfs, line 170: Should this be "Starting with the ..."?

cppwfs (changes requested):
Looks great.

Just a couple of nitpicks.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/session/SftpServerTests.java
artembilan: I don't think so.
This is just a comment in the test to give more context what exactly we are testing here and why.

spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
artembilan:
```suggestion
```

cppwfs (approved):
LGTM!  

Thanks for the fix!

PR #10856, opened 2026-03-12
GH-10853: Fix MQTTv5 for any shared subscription pattern

### LINKED ISSUE #10853: Shared subscription prefix "$SharedSubscription" not supported

Since version 6.5.6

**Describe the bug**

We use the mqtt topic prefix "$SharedSubscription" instead of the default prefix "$share" to use to mqtt shared subscription feature.

Since version 6.5.6 of spring-integration-mqtt, all MQTTv5 messages are ignored when the "$SharedSubscription" prefix is used. The topics of incoming messages do not match any subscribed topics, so no messages are processed.

The relevant code appears to be:
https://github.com/spring-projects/spring-integration/blob/79f6938b2a6792ea08dc2c27ba96464c3e1f15e1/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java#L444

The topic prefix for shared subscriptions is currently not configurable. This leads to incompatibility with brokers like IBM MessageSight, which require "$SharedSubscription".

**To Reproduce**

1. Use an MQTT broker that requires the "$SharedSubscription" prefix for shared subscriptions (e.g. IBM MessageSight).
1. Configure spring-integration-mqtt to subscribe to "$SharedSubscription/prod/myTopic".
1. Send a message to "myTopic".
1. The callback function is not triggered because the topic is not recognized.

**Expected behavior**

The callback function should be triggered when a message arrives on the topic "myTopic"—regardless of whether the prefix is "$share" or "$SharedSubscription". The topic prefix should be configurable.

**Sample**

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10853

Apparently, not all MQTT brokers come with `$share/` pattern on shared subscription topics.

* Fix `Mqttv5PahoMessageDrivenChannelAdapter` to check `subscribedTopic` for starting `$` symbol and having `/`
* Adjust `ClientManagerBackToBackTests.sharedTopicWithClientManager()` for IBM MessageSight and Eclipse Amlen pattern with `$SharedSubscription/`

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -405,22 +405,22 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 	}
 
 	@Override
-	public void removeTopic(String... topic) {
+	public void removeTopic(String... topics) {
 		this.topicLock.lock();
 		try {
-			super.removeTopic(topic);
-			for (String topicToRemove : topic) {
+			super.removeTopic(topics);
+			for (String topicToRemove : topics) {
 				this.topicsWithoutSharedSubscriptionPrefix.remove(topicToRemove);
 			}
 			if (this.mqttClient != null && this.mqttClient.isConnected()) {
-				unsubscribe(topic);
+				unsubscribe(topics);
 			}
 			if (!CollectionUtils.isEmpty(this.subscriptions)) {
-				this.subscriptions.removeIf((sub) -> ObjectUtils.containsElement(topic, sub.getTopic()));
+				this.subscriptions.removeIf((sub) -> ObjectUtils.containsElement(topics, sub.getTopic()));
 			}
 		}
 		catch (MqttException ex) {
-			throw new MessagingException("Failed to unsubscribe from topic(s) " + Arrays.toString(topic), ex);
+			throw new MessagingException("Failed to unsubscribe from topics(s) " + Arrays.toString(topics), ex);
 		}
 		finally {
 			this.topicLock.unlock();
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -20,10 +20,13 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.ConcurrentModificationException;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.IntStream;
 
 import org.eclipse.paho.mqttv5.client.IMqttAsyncClient;
@@ -92,6 +95,9 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 		extends AbstractMqttMessageDrivenChannelAdapter<IMqttAsyncClient, MqttConnectionOptions>
 		implements MqttCallback, MqttComponent<MqttConnectionOptions> {
 
+	private static final Pattern SHARED_SUBSCRIPTION_PATTERN =
+			Pattern.compile("^\\$(?:share|SharedSubscription)/[^/]*/(.*)$");
+
 	private final Lock lock = new ReentrantLock();
 
 	/**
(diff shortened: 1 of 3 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -532,7 +532,7 @@ public void authPacketArrived(int reasonCode, MqttProperties properties) {
 
 	private void messageArrivedIfMatched(String topic, MqttMessage mqttMessage) {
 		for (String subscribedTopic : getTopic()) {
-			if (subscribedTopic.startsWith("$share/")) {
+			if (subscribedTopic.startsWith("$") && subscribedTopic.contains("/")) {
 				subscribedTopic = subscribedTopic.split("/", 3)[2];
 			}
```
mrpiggi, line 537: To be safe, maybe go with
```java
if (subscribedTopic.startsWith("$") && subscribedTopic.chars().filter(i -> i == '/').count() >= 2) {
	subscribedTopic = subscribedTopic.split("/", 3)[2];
}
```
as otherwise if the topic is malformed, you would get an `ArrayIndexOutOfBoundsException`

mrpiggi: I thought about it again. We are using mosquitto and there is by default a topic `$SYS/broker/#` that would also match currently, which is not desired. So I would really go with a definite set of prefixes considered as shared subscriptions. Maybe something like this:
```java
private static final String[] SHARED_PREFIXES = { "$share", "$SharedSubscription" };

...

	if (subscribedTopic.startsWith("$")) {
		final String[] split = subscribedTopic.split("/", 3);
		if (split.length == 3 && Arrays.asList(SHARED_PREFIXES).contains(split[0])) {
			subscribedTopic = split[2];
		}
	}
```

artembilan: @mrpiggi ,

thank you for feedback!
This is really great contribution to the project.

I'll take your critics and ideas into account.
We definitely need to fix this properly

artembilan:
@mrpiggi ,

I have pushed some flexible fix we all are worried about.
Feel free to review it again!

We are planning to release next week.

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -193,6 +203,18 @@ public Mqttv5PahoMessageDrivenChannelAdapter(ClientManager<IMqttAsyncClient, Mqt
...
+	}
+
+	private void populateTopicsWithoutSharedSubscriptionPrefix(String... topics) {
+		for (String topic : topics) {
+			String topicWithoutPrefix = topic;
+			Matcher matcher = SHARED_SUBSCRIPTION_PATTERN.matcher(topic);
+			if (matcher.find()) {
+				topicWithoutPrefix = matcher.group(1);
+			}
+			this.topicsWithoutSharedSubscriptionPrefix.put(topic, topicWithoutPrefix);
```
mrpiggi, line 216: As something like `$share/sss/` could result in an empty string as well as `String... topics` could potentially hold blank strings and `null`, maybe:
```java
if (StringUtils.hasText(topicWithoutPrefix)) {
	this.topicsWithoutSharedSubscriptionPrefix.put(topic, topicWithoutPrefix);
}
```
Otherwise looks good.

artembilan: WE do have the `validateTopics()` in the super class before this new method is called, so empty/null is not possible.
The pattern `$share/sss/` is stange, I guess that would be rejected by the broker on subscription.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -386,6 +409,9 @@ public void removeTopic(String... topic) {
 		this.topicLock.lock();
 		try {
 			super.removeTopic(topic);
+			for (String topicToRemove : topic) {
```
cppwfs: I know that topic is a vararg, but it feels like it should be plural.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -92,6 +95,9 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 		extends AbstractMqttMessageDrivenChannelAdapter<IMqttAsyncClient, MqttConnectionOptions>
 		implements MqttCallback, MqttComponent<MqttConnectionOptions> {
 
+	private static final Pattern SHARED_SUBSCRIPTION_PATTERN =
+			Pattern.compile("^\\$(?:share|SharedSubscription)/[^/]*/(.*)$");
```
cppwfs, line 99: Should we add a mocked up test to make sure that the usage of this will work if other patterns are found outside IBM?

Not an absolute requirement, just thinking.

cppwfs (changes requested):
Awesome!

A couple of small questions.

Do we want to put this in the "whats-new" since we are supporting both `share` and `SharedSubscription`  ?

cppwfs (approved):
Love it!

Thanks for the changes

LGTM

PR #10858, opened 2026-03-13
Fix reference link in `FTPS and Shared SSLSession`

mjd507:
current it uses a regional suffix `.cz` , which can not be opened in my region,  however,  `.com` works for me.


cppwfs (approved):
LGTM 

PR #10888, opened 2026-03-18
Fix SessionFactory Code example for ftp/sftp.

mjd507:
The `setTestSession()` method is available in `CachingSessionFactory`, but not in the `DefaultFtpSessionFactory` or `DefaultSftpSessionFactory` implementations.

### DIFF
--- a/src/reference/antora/modules/ROOT/pages/ftp/inbound.adoc
+++ b/src/reference/antora/modules/ROOT/pages/ftp/inbound.adoc
@@ -233,7 +233,6 @@ public class FtpJavaApplication {
         sf.setPort(port);
         sf.setUsername("foo");
         sf.setPassword("foo");
-        sf.setTestSession(true);
         return new CachingSessionFactory<FTPFile>(sf);
     }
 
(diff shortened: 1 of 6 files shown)


src/reference/antora/modules/ROOT/pages/ftp/inbound.adoc
```diff
@@ -233,7 +233,6 @@ public class FtpJavaApplication {
         sf.setPort(port);
         sf.setUsername("foo");
         sf.setPassword("foo");
-        sf.setTestSession(true);
         return new CachingSessionFactory<FTPFile>(sf);
```
mjd507, line 236: I am happy to change to below if it is required in doc.
```java
var cachingSessionFactory = new CachingSessionFactory<>(factory);
cachingSessionFactory.setTestSession(true);
return cachingSessionFactory;
```

artembilan: I think that was a copy/paste artifact.
I don't think we need to to anything else in this sample but just the removal you have done.
The `setTestSession()` has nothing to do with the context of this example here in the doc.
So, approving your current change and nothing more is needed to be done here.

PR #10905, opened 2026-03-31
Add dsl for redis inbound/outbound channel adapter.

mjd507:
Related to: #8108

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
@@ -36,6 +36,15 @@ public final class Redis {
 		return new RedisInboundChannelAdapterSpec(connectionFactory);
 	}
 
+	/**
+	 * The factory to produce a {@link RedisOutboundChannelAdapterSpec}.
+	 * @param connectionFactory the {@link RedisConnectionFactory} to build on
+	 * @return the {@link RedisOutboundChannelAdapterSpec} instance
+	 */
+	public static RedisOutboundChannelAdapterSpec outboundChannelAdapter(RedisConnectionFactory connectionFactory) {
+		return new RedisOutboundChannelAdapterSpec(connectionFactory);
+	}
+
 	private Redis() {
 	}
 
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
@@ -22,6 +22,7 @@ import org.springframework.data.redis.connection.RedisConnectionFactory;
  * Factory class for Redis components.
  *
  * @author Jiandong Ma
+ *
  * @since 7.1
  */
 public final class Redis {
@@ -34,6 +35,9 @@ public final class Redis {
 	 */
 	public static RedisInboundChannelAdapterSpec inboundChannelAdapter(RedisConnectionFactory connectionFactory) {
 		return new RedisInboundChannelAdapterSpec(connectionFactory);
+	}
+
+	private Redis() {
 	}
 
 }
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
@@ -29,7 +29,6 @@ public final class Redis {
 
 	/**
 	 * The factory to produce a {@link RedisInboundChannelAdapterSpec}.
-	 *
 	 * @param connectionFactory the {@link RedisConnectionFactory} to build on
 	 * @return the {@link RedisInboundChannelAdapterSpec} instance
 	 */
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.redis.dsl;
+
+import org.springframework.data.redis.connection.RedisConnectionFactory;
+
+/**
+ * Factory class for Redis components.
+ *
+ * @author Jiandong Ma
+ * @since 7.1
+ */
+public final class Redis {
+
+	/**
+	 * The factory to produce a {@link RedisInboundChannelAdapterSpec}.
+	 *
+	 * @param connectionFactory the {@link RedisConnectionFactory} to build on
+	 * @return the {@link RedisInboundChannelAdapterSpec} instance
+	 */
+	public static RedisInboundChannelAdapterSpec inboundChannelAdapter(RedisConnectionFactory connectionFactory) {
+		return new RedisInboundChannelAdapterSpec(connectionFactory);
+	}
+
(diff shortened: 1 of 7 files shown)


spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/Redis.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Jiandong Ma
+ *
+ * @since 7.1
+ */
+public final class Redis {
+
+	/**
+	 * The factory to produce a {@link RedisInboundChannelAdapterSpec}.
+	 *
```
artembilan: No blank lines in method Javadocs.

spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/RedisInboundChannelAdapterSpec.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * A {@link MessageProducerSpec} for a {@link RedisInboundChannelAdapterSpec}.
+ *
+ * @author Jiandong Ma
+ *
+ * @since 7.1
+ */
+public class RedisInboundChannelAdapterSpec extends MessageProducerSpec<RedisInboundChannelAdapterSpec, RedisInboundChannelAdapter> {
+
+	public RedisInboundChannelAdapterSpec(RedisConnectionFactory connectionFactory) {
```
artembilan: Not `public` , but `protected`.
See many other similar classes in the project.
The point is that instances of these specs must be created via respective factory.
It is up to use when you extend such a class.

spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/RedisInboundChannelAdapterSpec.java
```diff
@@ -0,0 +1,90 @@
...
+		this.target.setTopics(topics);
+		return this;
+	}
+
+	/**
+	 * @param topicPatterns the topicPatterns
+	 * @return the spec
+	 * @see RedisInboundChannelAdapter#setTopicPatterns(String...)
+	 */
+	public RedisInboundChannelAdapterSpec topicsPatterns(String... topicPatterns) {
```
artembilan: I think we want be consistent and name this method as `topicPatterns`.
And I believe that is how English language works: only one plural in the sequence.
Essentially the first part could be treated as possessor of the last one.
Well, essentially we still have to say: `item label names` 😄

spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/package-info.java
```diff
@@ -0,0 +1,2 @@
+@org.jspecify.annotations.NullMarked
+package org.springframework.integration.redis.dsl;
```
artembilan: The line feed in the end of file.
And add some Javadoc to this file.

spring-integration-redis/src/test/java/org/springframework/integration/redis/dsl/RedisTests.java
```diff
@@ -0,0 +1,109 @@
...
+			// Then
+			assertThat(message)
+					.isNotNull()
+					.satisfies(msg -> {
+						assertThat(msg.getHeaders()).containsEntry(RedisHeaders.MESSAGE_SOURCE, Config.inboundChannelAdapterTopic);
+						assertThat(msg.getPayload().toString()).startsWith("test-");
+					});
+			counter++;
+		}
+		assertThat(counter).isEqualTo(numToTest);
```
artembilan: This assertion looks redundant.
I'm not sure how the loop can pass if there is a failure on some of the iterations.

spring-integration-redis/src/test/java/org/springframework/integration/redis/dsl/RedisTests.java
```diff
@@ -0,0 +1,109 @@
...
+
+	@Autowired
+	private QueueChannel inboundChannelAdapterQueueChannel;
+
+	@Test
+	void testInboundChannelAdapter() throws Exception {
+		StringRedisTemplate redisTemplate = new StringRedisTemplate(connectionFactory);
+		redisTemplate.afterPropertiesSet();
+		RedisContainerTest.awaitFullySubscribed(TestUtils.getPropertyValue(inboundChannelAdapter, "container"),
+				redisTemplate, Config.inboundChannelAdapterTopic, inboundChannelAdapterQueueChannel, "foo");
```
artembilan: May we don't use `foo/bar` language in our code, please?

spring-integration-redis/src/test/java/org/springframework/integration/redis/dsl/RedisTests.java
```diff
@@ -0,0 +1,109 @@
...
+			counter++;
+		}
+		assertThat(counter).isEqualTo(numToTest);
+	}
+
+	@Configuration
+	@EnableIntegration
+	public static class Config {
+
+		static String inboundChannelAdapterTopic = "testDslInboundChannelAdapterTopic";
```
artembilan: `final`?
And then let's make it as constant!
Plus it looks like scope of this constant is wider than just this config class.
Move it to the top test class for better visibility.

artembilan (changes requested):
Thank you!

Final we got a crack on that issue!

Not too much, but still some review.

spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/RedisInboundChannelAdapterSpec.java
mjd507: nice catch

spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/RedisOutboundChannelAdapterSpec.java
```diff
@@ -0,0 +1,90 @@
...
+		this.target.setTopic(topic);
+		return this;
+	}
+
+	/**
+	 * @param topicExpression the topicExpression
+	 * @return the spec
+	 * @see RedisPublishingMessageHandler#setTopicExpression(Expression)
+	 */
+	public RedisOutboundChannelAdapterSpec topicExpression(Expression topicExpression) {
```
artembilan, line 78: Add also `topicExpression(String)` and `topicFunction(Function)`. See other similar spec impls

spring-integration-redis/src/main/java/org/springframework/integration/redis/dsl/RedisOutboundChannelAdapterSpec.java
```diff
@@ -0,0 +1,90 @@
...
+		this.target.setTopicExpression(topicExpression);
+		return this;
+	}
+
+	/**
+	 * @param evaluationContext the evaluationContext
+	 * @return the spec
+	 * @see RedisPublishingMessageHandler#setIntegrationEvaluationContext(EvaluationContext)
+	 */
+	public RedisOutboundChannelAdapterSpec taskExecutor(EvaluationContext evaluationContext) {
```
artembilan: The name of the method does not reflect reality. Looks like copy/paste artifact.
Either way, we must not expose setting of the evaluation context.
So, just remove this altogether !

spring-integration-redis/src/test/java/org/springframework/integration/redis/dsl/RedisTests.java
```diff
@@ -0,0 +1,178 @@
...
+				.satisfies(msgList -> {
+					msgList.forEach((msg) -> {
+						assertThat(stringRedisSerializer.deserialize(msg.getChannel())).isEqualTo(TOPIC_FOR_OUTBOUND_CHANNEL_ADAPTER);
+						assertThat(stringRedisSerializer.deserialize(msg.getBody())).startsWith("outbound-test-");
+					});
+				});
+		container.stop();
+	}
+
+	@Configuration
```
artembilan: For better performance we try to use `proxyBeanMethods = false`.
Also, for better code readability we try to omit all the `public` and `private` modifiers in tests.

artembilan (changes requested):
Thank for the update !
I see you have gone a bit further than just inbound. That’s OK with me , however we still need reflect this change in docs: like mention with a short sentence in the ‘whats-new.adoc` and respective parts of target Redis chapter .

artembilan (approved):
Honestly, this is great and ready to be merged.
If you'd like to add more DSL for other Redis components, I would suggest to do that in other PR.

Thank you!

PR #10907, opened 2026-03-31
Introduce spring-integration-dependencies BOM

cppwfs:
- Extract all platform BOMs from `allprojects` into a new `spring-integration-dependencies` subproject using the `java-platform` plugin.
- Consolidate dependency constraints for optional dependencies across all submodules into the new platform BOM.
- Replace individual `dependencyManagement` platform declarations in `build.gradle` with a single reference to the new `:spring-integration-dependencies` platform project.
- Exclude the new subproject from `javaProjects` to prevent Java compilation tasks from running against it.


spring-integration-dependencies/build.gradle
```diff
@@ -0,0 +1,96 @@
+description = 'Spring Integration External Dependencies BOM'
+
+apply plugin: 'java-platform'
+apply from: "${rootDir}/gradle/publish-maven.gradle"
```
cppwfs: Remove publish

spring-integration-dependencies/build.gradle
```diff
@@ -0,0 +1,96 @@
...
+
+javaPlatform {
+	allowDependencies()
```
cppwfs: Do we need this?

spring-integration-dependencies/build.gradle
```diff
@@ -0,0 +1,96 @@
...
+}
+
+tasks.withType(GenerateModuleMetadata).configureEach {
+	enabled = false
```
cppwfs: Remove this

spring-integration-dependencies/build.gradle
```diff
@@ -0,0 +1,96 @@
...
+		api "org.graalvm.sdk:graal-sdk:$graalvmVersion"
+		api "org.graalvm.polyglot:js:$graalvmVersion"
+		api "org.graalvm.polyglot:python:$graalvmVersion"
+
+		// spring-integration-ws
+		api "com.sun.xml.bind:jaxb-impl:$jaxbVersion"
+	}
+}
+
+publishing {
```
cppwfs: Remove

spring-integration-dependencies/build.gradle
```diff
@@ -0,0 +1,96 @@
...
+javaPlatform {
+	allowDependencies()
+}
+
+tasks.withType(GenerateModuleMetadata).configureEach {
+	enabled = false
+}
+
+dependencies {
+	api platform("com.fasterxml.jackson:jackson-bom:$jacksonVersion")
```
cppwfs: Move version properties from main build.gradle to spring-integration-dependencies:build.gradle

spring-integration-dependencies/build.gradle
cppwfs: Yes.  This allows us to use the platform declarations below

build.gradle
```diff
@@ -33,80 +33,17 @@ ext {
...
     javaProjects = subprojects - project(':spring-integration-bom') - project(':spring-integration-dependencies')
 
 }
 
+evaluationDependsOn(':spring-integration-dependencies')
+
+// Expose dependency versions defined in spring-integration-dependencies to the root project
+project(':spring-integration-dependencies').ext.properties.findAll { k, v -> v instanceof String }.each { k, v ->
+	ext.set(k, v)
+}
```
artembilan, line 40: I think this is not what we are aiming for.
All the versions have to be there in that module and here all the dependencies must be used without versions.

artembilan (changes requested):

build.gradle
```diff
@@ -33,80 +33,14 @@ ext {
...
-    tomcatVersion = '11.0.20'
-    xmlUnitVersion = '2.11.0'
-    xstreamVersion = '1.4.21'
-    ztZipVersion = '1.17'
-
-    javaProjects = subprojects - project(':spring-integration-bom')
+    javaProjects = subprojects - project(':spring-integration-bom') - project(':spring-integration-dependencies')
 
+	protobufVersion = '4.34.1'
+	grpcVersion = '1.79.0'
```
artembilan: I'm not very happy to have these duplicated here and there, but that is probably going to be next evolution of this effort.
And I believe that TOML should do the trick: https://docs.gradle.org/current/userguide/version_catalogs.html.

Especially I like bundle support, where those our well-known MQTT drivers could be pull as a single line of import.

build.gradle
```diff
@@ -33,80 +33,14 @@ ext {
...
 }
 
+evaluationDependsOn(':spring-integration-dependencies')
```
artembilan: Why do we still need this here?

build.gradle
```diff
@@ -460,7 +375,7 @@ project('spring-integration-cloudevents') {
 
         testImplementation 'io.cloudevents:cloudevents-json-jackson'
 		// Is not a part of the 'cloudevents-bom'
```
artembilan: I believe this comment should move to respective place.
After removing version from this dependency, it gives us impression that it is already a part of that BOM.
Therefore, to not confuse here and really reflect reality, let's move this comment to the place where we explicitly declare this dependency with version!

spring-integration-dependencies/build.gradle
artembilan: This was not resolved.
We agreed to not publish this platform, but just use internally.

artembilan (changes requested):

build.gradle
artembilan: Well, after thinking about this one more time, it feels like both TOML and this internal platform resolve the same problem: central place to manage dependencies and their versions.
I would leave accept this solution if we would not have a duplication for those gRPC and Rpoto version variables.
Plus I still have doubts in duplicating dependency coordinates.

Would you mind take a look into that TOML support and see if we got any benefits instead?
Looks like Micrometer could be a good example to follow: https://github.com/micrometer-metrics/micrometer/blob/main/gradle/libs.versions.toml

Thanks

build.gradle
cppwfs: After we implement TOML, this can be removed as well as the `spring-integration-dependencies` module.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-zeromq-jeromq = { module = "org.zeromq:jeromq", version.ref = "jeroMqVersion" }
+org-zeroturnaround-zt-zip = { module = "org.zeroturnaround:zt-zip", version.ref = "ztZipVersion" }
+spring-amqp-bom = { module = "org.springframework.amqp:spring-amqp-bom", version.ref = "springAmqpVersion" }
+spring-data-bom = { module = "org.springframework.data:spring-data-bom", version.ref = "springDataVersion" }
+spring-framework-bom = { module = "org.springframework:spring-framework-bom", version.ref = "springVersion" }
+spring-graphql = { module = "org.springframework.graphql:spring-graphql", version.ref = "springGraphqlVersion" }
+spring-kafka-bom = { module = "org.springframework.kafka:spring-kafka-bom", version.ref = "springKafkaVersion" }
+spring-security-bom = { module = "org.springframework.security:spring-security-bom", version.ref = "springSecurityVersion" }
+spring-ws-bom = { module = "org.springframework.ws:spring-ws-bom", version.ref = "springWsVersion" }
+tools-jackson-jackson-bom = { module = "tools.jackson:jackson-bom", version.ref = "jackson3Version" }
```
artembilan: Thanks, Glenn, for speedy turnaround!

Can you tell us, please, how do you feel about this solution?
Is it more flexible, clear and straightforward?
Much easier to digest?
I mean looks like more typing, although we are not going to duplicate dependency coordinates here an there and just use these variables.

If this is good with you, can we please, revise those duplicates?
I'm not sure how this was generated, but I think `tools-jackson-jackson-bom` is too much and could be simplified into `tools-jackson-bom`.
Or like that `org-mongodb-mongodb-driver-bom` into just `mongodb-bom`.
I also this that those explicit `org-mongodb-mongodb-driver-reactivestreams` and `org-mongodb-mongodb-driver-sync` are redundant and we can just use them directly in the target module as we do now for the Testcontainers modules.

artembilan (changes requested):

gradle/libs.versions.toml
cppwfs: Can you tell us, please, how do you feel about this solution?  **I like that we don't have versions in the build.gradle, but in a dedicate location.**
Is it more flexible, clear and straightforward?  **To be honest, I did not know this existed till you mentioned it, so I had to read up on it.   I like how it simplifies the `build.gradle` file.**
Much easier to digest?  **Yes**

Overall, I like the simplification of the `build.gradle` file.   `Build.gradle` now focusses on the building of the application and delegates the version control to something that is dedicated to it (TOML).

cppwfs: I will simplify the duplicate naming i.e. `io-debezium-debezium-embedded` to `io-debezium-embedded`.

This was a side-effect of how the artifact is named. i.e. `io.debezium:debezium-embedded`

artembilan: Well, why just not `debezium`?
Who cares that it is `embedded` in our case?
And what value does that `io` give us?
I may understand something like `commons-io`, `commons-net`, but `debezium`, or `mongodb` or `zeromq` are very catchy names.

Let's discuss every single of them after your next iteration!

Thanks

build.gradle
```diff
@@ -782,8 +712,8 @@ project('spring-integration-mongodb') {
     dependencies {
         api 'org.springframework.data:spring-data-mongodb'
 
-        optionalApi 'org.mongodb:mongodb-driver-sync'
-        optionalApi 'org.mongodb:mongodb-driver-reactivestreams'
+        optionalApi libs.org.mongodb.driver.sync
+        optionalApi libs.org.mongodb.driver.reactivestreams
```
artembilan: So, 'tools.jackson.core:jackson-databind' is OK. but we still need explicit variable for MongoDB drivers?
We do use respective bom, so we don't need extra management for these specific deps and just rely on their normal definition without version.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+apache-derbyclient = { module = "org.apache.derby:derbyclient", version.ref = "derbyVersion" }
+apache-derbytools = { module = "org.apache.derby:derbytools", version.ref = "derbyVersion" }
+apache-ftpserver-core = { module = "org.apache.ftpserver:ftpserver-core", version.ref = "ftpServerVersion" }
+apache-groovy-bom = { module = "org.apache.groovy:groovy-bom", version.ref = "groovyVersion" }
+apache-logging-log4j-bom = { module = "org.apache.logging.log4j:log4j-bom", version.ref = "log4jVersion" }
+apache-tomcat-embed-websocket = { module = "org.apache.tomcat.embed:tomcat-embed-websocket", version.ref = "tomcatVersion" }
+com-esotericsoftware-kryo = { module = "com.esotericsoftware:kryo", version.ref = "kryoVersion" }
+com-fasterxml-jackson-bom = { module = "com.fasterxml.jackson:jackson-bom", version.ref = "jacksonVersion" }
+com-google-protobuf-java = { module = "com.google.protobuf:protobuf-java", version.ref = "protobufVersion" }
+com-google-protobuf-java-util = { module = "com.google.protobuf:protobuf-java-util", version.ref = "protobufVersion" }
```
artembilan: Looks like there is BOM for Protobuf: https://github.com/protocolbuffers/protobuf/blob/main/java/bom/pom.xml

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+tomcatVersion = "11.0.21"
+xmlUnitVersion = "2.11.0"
+xstreamVersion = "1.4.21"
+ztZipVersion = "1.17"
+
+[libraries]
+apache-sshd-sftp = { module = "org.apache.sshd:sshd-sftp", version.ref = "apacheSshdVersion" }
+apache-activemq-artemis-jakarta-client = { module = "org.apache.activemq:artemis-jakarta-client", version.ref = "artemisVersion" }
+apache-activemq-artemis-server = { module = "org.apache.activemq:artemis-server", version.ref = "artemisVersion" }
+apache-activemq-artemis-stomp-protocol = { module = "org.apache.activemq:artemis-stomp-protocol", version.ref = "artemisVersion" }
```
artembilan: Looks like there is a BOM for Artemis: https://mvnrepository.com/artifact/org.apache.artemis/artemis-bom/2.53.0

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+apache-logging-log4j-bom = { module = "org.apache.logging.log4j:log4j-bom", version.ref = "log4jVersion" }
+apache-tomcat-embed-websocket = { module = "org.apache.tomcat.embed:tomcat-embed-websocket", version.ref = "tomcatVersion" }
+com-esotericsoftware-kryo = { module = "com.esotericsoftware:kryo", version.ref = "kryoVersion" }
+com-fasterxml-jackson-bom = { module = "com.fasterxml.jackson:jackson-bom", version.ref = "jacksonVersion" }
+com-google-protobuf-java = { module = "com.google.protobuf:protobuf-java", version.ref = "protobufVersion" }
+com-google-protobuf-java-util = { module = "com.google.protobuf:protobuf-java-util", version.ref = "protobufVersion" }
+com-h2database-h2 = { module = "com.h2database:h2", version.ref = "h2Version" }
+com-hazelcast = { module = "com.hazelcast:hazelcast", version.ref = "hazelcastVersion" }
+com-icegreen-greenmail = { module = "com.icegreen:greenmail", version.ref = "greenmailVersion" }
+com-jayway-jsonpath-json-path = { module = "com.jayway.jsonpath:json-path", version.ref = "jsonpathVersion" }
```
artembilan: Can we have just `com-jayway-jsonpath`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+com-mysql-connector-j = { module = "com.mysql:mysql-connector-j", version.ref = "mysqlVersion" }
```
artembilan: Can this go without `j`?
I guess it looks a bit ugly in the consumer of this API:
```
lib.com.mysql.connector.j
```
🤷

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+com-oracle-database-jdbc-ojdbc11 = { module = "com.oracle.database.jdbc:ojdbc11", version.ref = "oracleVersion" }
```
artembilan: Can this be just `com-oracle-ojdbc11`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+com-rometools-rome = { module = "com.rometools:rome", version.ref = "romeToolsVersion" }
+com-sun-xml-bind-jaxb-impl = { module = "com.sun.xml.bind:jaxb-impl", version.ref = "jaxbVersion" }
```
artembilan: Can this be just `com-sun-jaxb`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+com-thoughtworks-xstream = { module = "com.thoughtworks.xstream:xstream", version.ref = "xstreamVersion" }
+com-willowtreeapps-assertk-jvm = { module = "com.willowtreeapps.assertk:assertk-jvm", version.ref = "assertkVersion" }
```
artembilan: Without `jvm`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+io-cloudevents-bom = { module = "io.cloudevents:cloudevents-bom", version.ref = "cloudEventsVersion" }
+io-cloudevents-xml = { module = "io.cloudevents:cloudevents-xml", version.ref = "cloudEventsVersion" }
+io-debezium-connector-mysql = { module = "io.debezium:debezium-connector-mysql", version.ref = "debeziumVersion" }
+io-debezium-embedded = { module = "io.debezium:debezium-embedded", version.ref = "debeziumVersion" }
+io-github-resilience4j-ratelimiter = { module = "io.github.resilience4j:resilience4j-ratelimiter", version.ref = "resilience4jVersion" }
+io-grpc-bom = { module = "io.grpc:grpc-bom", version.ref = "grpcVersion" }
+io-lettuce-core = { module = "io.lettuce:lettuce-core", version.ref = "lettuceVersion" }
+io-micrometer-bom = { module = "io.micrometer:micrometer-bom", version.ref = "micrometerVersion" }
+io-micrometer-tracing-bom = { module = "io.micrometer:micrometer-tracing-bom", version.ref = "micrometerTracingVersion" }
+io-projectreactor-reactor-bom = { module = "io.projectreactor:reactor-bom", version.ref = "reactorVersion" }
```
artembilan: Just `io-projectreactor-bom` ?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-awaitility = { module = "org.awaitility:awaitility", version.ref = "awaitilityVersion" }
+org-codelibs-jcifs = { module = "org.codelibs:jcifs", version.ref = "jcifsVersion" }
+org-eclipse-angus-mail = { module = "org.eclipse.angus:angus-mail", version.ref = "mailVersion" }
+org-eclipse-angus-jakarta-mail = { module = "org.eclipse.angus:jakarta.mail", version.ref = "mailVersion" }
+org-eclipse-paho-client-mqttv3 = { module = "org.eclipse.paho:org.eclipse.paho.client.mqttv3", version.ref = "pahoMqttClientVersion" }
+org-eclipse-paho-mqttv5-client = { module = "org.eclipse.paho:org.eclipse.paho.mqttv5.client", version.ref = "pahoMqttClientVersion" }
+org-graalvm-polyglot-js = { module = "org.graalvm.polyglot:js", version.ref = "graalvmVersion" }
+org-graalvm-polyglot-python = { module = "org.graalvm.polyglot:python", version.ref = "graalvmVersion" }
+org-graalvm-sdk-graal-sdk = { module = "org.graalvm.sdk:graal-sdk", version.ref = "graalvmVersion" }
+org-hamcrest-library = { module = "org.hamcrest:hamcrest-library", version.ref = "hamcrestVersion" }
```
artembilan: Just `org-hamcrest`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-hibernate-orm-hibernate-core = { module = "org.hibernate.orm:hibernate-core", version.ref = "hibernateVersion" }
```
artembilan: `org-hibernate-core`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-hsqldb-hsqldb = { module = "org.hsqldb:hsqldb", version.ref = "hsqldbVersion" }
```
artembilan: `org-hsqldb` ?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-igniterealtime-smack-experimental = { module = "org.igniterealtime.smack:smack-experimental", version.ref = "smackVersion" }
+org-igniterealtime-smack-extensions = { module = "org.igniterealtime.smack:smack-extensions", version.ref = "smackVersion" }
+org-igniterealtime-smack-java8 = { module = "org.igniterealtime.smack:smack-java8", version.ref = "smackVersion" }
+org-igniterealtime-smack-tcp = { module = "org.igniterealtime.smack:smack-tcp", version.ref = "smackVersion" }
+org-jetbrains-kotlin-scripting-jsr223 = { module = "org.jetbrains.kotlin:kotlin-scripting-jsr223", version.ref = "kotlinVersion" }
```
artembilan: This is redundant.
The dep comes already without version.
And `kotlinVersion` is managed by respective Gradle plugin.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-jetbrains-kotlinx-coroutines-bom = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-bom", version.ref = "kotlinCoroutinesVersion" }
+org-jetbrains-kotlinx-coroutines-reactor = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-reactor", version.ref = "kotlinCoroutinesVersion" }
+org-jruby-complete = { module = "org.jruby:jruby-complete", version.ref = "jrubyVersion" }
+org-junit-junit-bom = { module = "org.junit:junit-bom", version.ref = "junitJupiterVersion" }
```
artembilan: `org-junit-bom`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-junit-pioneer-junit-pioneer = { module = "org.junit-pioneer:junit-pioneer", version.ref = "junitPioneerVersion" }
```
artembilan: `org-junit-pioneer`?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+org-mockito-core = { module = "org.mockito:mockito-core", version.ref = "mockitoVersion" }
+org-mongodb-driver-bom = { module = "org.mongodb:mongodb-driver-bom", version.ref = "mongoDriverVersion" }
+org-mongodb-driver-reactivestreams = { module = "org.mongodb:mongodb-driver-reactivestreams", version.ref = "mongoDriverVersion" }
+org-mongodb-driver-sync = { module = "org.mongodb:mongodb-driver-sync", version.ref = "mongoDriverVersion" }
```
artembilan: As we discussed in other place: this is redundant.
We do use BOM already.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,169 @@
...
+spring-graphql = { module = "org.springframework.graphql:spring-graphql", version.ref = "springGraphqlVersion" }
+spring-kafka-bom = { module = "org.springframework.kafka:spring-kafka-bom", version.ref = "springKafkaVersion" }
+spring-security-bom = { module = "org.springframework.security:spring-security-bom", version.ref = "springSecurityVersion" }
+spring-ws-bom = { module = "org.springframework.ws:spring-ws-bom", version.ref = "springWsVersion" }
+tools-jackson-bom = { module = "tools.jackson:jackson-bom", version.ref = "jackson3Version" }
+
+[plugins]
+dokka = { id = "org.jetbrains.dokka", version = "2.1.0" }
+antora = { id = "org.antora", version = "1.0.0" }
+antora-generate-yml = { id = "io.spring.antora.generate-antora-yml", version = "0.0.1" }
```
artembilan: `io-spring-generate-antora-yml` ?

artembilan (changes requested):

gradle/libs.versions.toml
```diff
@@ -0,0 +1,163 @@
...
+jcifsVersion = "3.0.2"
+jeroMqVersion = "0.6.0"
+jmsApiVersion = "3.1.0"
+jpaApiVersion = "3.2.0"
+jrubyVersion = "10.0.4.0"
+jsonpathVersion = "3.0.0"
+junitJupiterVersion = "6.0.3"
+junitPioneerVersion = "2.3.0"
+kotlinCoroutinesVersion = "1.10.2"
+kotlinVersion = "2.3.20"
```
artembilan: Out of use.
As it has to be.
Kinda leftover after previous iteration.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,163 @@
...
+com-mysql-connector = { module = "com.mysql:mysql-connector-j", version.ref = "mysqlVersion" }
+com-oracle-ojdbc11 = { module = "com.oracle.database.jdbc:ojdbc11", version.ref = "oracleVersion" }
+com-rometools-rome = { module = "com.rometools:rome", version.ref = "romeToolsVersion" }
+com-sun-jaxb = { module = "com.sun.xml.bind:jaxb-impl", version.ref = "jaxbVersion" }
+com-thoughtworks-xstream = { module = "com.thoughtworks.xstream:xstream", version.ref = "xstreamVersion" }
+com-willowtreeapps-assertk = { module = "com.willowtreeapps.assertk:assertk-jvm", version.ref = "assertkVersion" }
+commons-io = { module = "commons-io:commons-io", version.ref = "commonsIoVersion" }
+commons-net = { module = "commons-net:commons-net", version.ref = "commonsNetVersion" }
+io-cloudevents-bom = { module = "io.cloudevents:cloudevents-bom", version.ref = "cloudEventsVersion" }
+io-cloudevents-xml = { module = "io.cloudevents:cloudevents-xml", version.ref = "cloudEventsVersion" }
```
artembilan, line 103: FYI: https://github.com/cloudevents/sdk-java/issues/701

gradle/libs.versions.toml
```diff
@@ -0,0 +1,163 @@
...
+org-awaitility = { module = "org.awaitility:awaitility", version.ref = "awaitilityVersion" }
+org-codelibs-jcifs = { module = "org.codelibs:jcifs", version.ref = "jcifsVersion" }
+org-eclipse-angus-mail = { module = "org.eclipse.angus:angus-mail", version.ref = "mailVersion" }
+org-eclipse-angus-jakarta-mail = { module = "org.eclipse.angus:jakarta.mail", version.ref = "mailVersion" }
+org-eclipse-paho-client-mqttv3 = { module = "org.eclipse.paho:org.eclipse.paho.client.mqttv3", version.ref = "pahoMqttClientVersion" }
+org-eclipse-paho-mqttv5-client = { module = "org.eclipse.paho:org.eclipse.paho.mqttv5.client", version.ref = "pahoMqttClientVersion" }
+org-graalvm-polyglot-js = { module = "org.graalvm.polyglot:js", version.ref = "graalvmVersion" }
+org-graalvm-polyglot-python = { module = "org.graalvm.polyglot:python", version.ref = "graalvmVersion" }
+org-graalvm-sdk-graal-sdk = { module = "org.graalvm.sdk:graal-sdk", version.ref = "graalvmVersion" }
+org-hamcrest     = { module = "org.hamcrest:hamcrest-library", version.ref = "hamcrestVersion" }
```
artembilan: Extra spaces for nothing.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,163 @@
...
+org-hibernate-core = { module = "org.hibernate.orm:hibernate-core", version.ref = "hibernateVersion" }
+org-hsqldb = { module = "org.hsqldb:hsqldb", version.ref = "hsqldbVersion" }
+org-igniterealtime-smack-experimental = { module = "org.igniterealtime.smack:smack-experimental", version.ref = "smackVersion" }
+org-igniterealtime-smack-extensions = { module = "org.igniterealtime.smack:smack-extensions", version.ref = "smackVersion" }
+org-igniterealtime-smack-java8 = { module = "org.igniterealtime.smack:smack-java8", version.ref = "smackVersion" }
+org-igniterealtime-smack-tcp = { module = "org.igniterealtime.smack:smack-tcp", version.ref = "smackVersion" }
+org-jetbrains-kotlinx-coroutines-bom = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-bom", version.ref = "kotlinCoroutinesVersion" }
+org-jetbrains-kotlinx-coroutines-reactor = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-reactor", version.ref = "kotlinCoroutinesVersion" }
```
artembilan: This has to be removed and target usage reverted.
We do import respective BOM for `coroutines`.

build.gradle
```diff
@@ -33,78 +33,10 @@ ext {
...
-    testcontainersVersion = '2.0.4'
-    tomcatVersion = '11.0.21'
-    xmlUnitVersion = '2.11.0'
-    xstreamVersion = '1.4.21'
-    ztZipVersion = '1.17'
-
     javaProjects = subprojects - project(':spring-integration-bom')
 
+	protobufVersion = libs.versions.protobufVersion.get()
+	grpcVersion = libs.versions.grpcVersion.get()
```
artembilan, line 39: Do we really need these variables?
Cannot we just use those references directly in the place where they are needed?
I mean we do have `libs.versions.grpcVersion` variable already, so why do we need extra?

build.gradle
```diff
@@ -699,36 +631,36 @@ project('spring-integration-ip') {
...
     description = 'Spring Integration JMS Support'
     dependencies {
         api 'org.springframework:spring-jms'
-        optionalApi "jakarta.jms:jakarta.jms-api:$jmsApiVersion"
+        optionalApi libs.jakarta.jms.api
 
-        testImplementation "org.apache.activemq:artemis-server:$artemisVersion"
-        testImplementation "org.apache.activemq:artemis-jakarta-client:$artemisVersion"
+        testImplementation 'org.apache.activemq:artemis-server'
+        testImplementation 'org.apache.activemq:artemis-jakarta-client'
```
artembilan: I think we should use now `org.apache.artemis` namespace instead of relocations: https://github.com/apache/artemis/blob/main/artemis-bom/pom.xml#L236

artembilan (changes requested):
This is good progress.
Just couple concerns for what could be done better.

Thanks

gradle/libs.versions.toml
cppwfs: We can create an issue to check back in before we release 7.1.0 GA

artembilan: Well, I have been subscribed to that issue.
So, that should be enough.
But yeah... We can create an issue to track it in our backlog.
Not sure that Cloud Events `4.1` will make in time for our `7.1`

build.gradle
cppwfs: We do need these because the `protobuf` plugin does not handle the variables from TOML.

PR #10908, opened 2026-04-01
GH-10906: DefaultLockRepository: Use LocalDateTime directly

### LINKED ISSUE #10906: JDBC lock repository fails after entering daylight savings

6.0.0-M5 and later. The bug was introduced in the code change for https://github.com/spring-projects/spring-integration/issues/3885.

**Describe the bug**

When the default timezone has daylight savings, then locks can fail after entering daylight savings (and likely also after leaving). The window for this bug to occur depends on the offset of the time zone.

**To Reproduce**

This problem cannot be reproduced reliably, but it's cause can be demonstrated using jshell.

The problem is caused by the following code in DefaultLockRepository.java

```
private Timestamp ttlEpochMillis(Duration ttl) {
	return Timestamp.valueOf(currentTime().plus(ttl));
}

private static Timestamp epochMillis() {
	return Timestamp.valueOf(currentTime());
}

private static LocalDateTime currentTime() {
	return LocalDateTime.now(ZoneOffset.UTC);
}
```

The Timestamp.valueOf method uses the system default time zone, which is Europe/Amsterdam in our case. This does not cause problems most of the time, but some LocalDateTime values are invalid because the forward adjustment of the clock when entering daylight savings. This may cause the timestamp stored in the database to jump back and forth. For example :

```
jshell> java.time.ZoneId.systemDefault()
$1 ==> Europe/Amsterdam

jshell> java.sql.Timestamp.valueOf(java.time.LocalDateTime.of(2026,3,29,2,59,59,999999999))
$2 ==> 2026-03-29 03:59:59.999999999

jshell> java.sql.Timestamp.valueOf(java.time.LocalDateTime.of(2026,3,29,3,0,0))
$3 ==> 2026-03-29 03:00:00.0
``` 

The demonstrated jump back in time breaks the lock temporarily because there is a short window where non-lock-owners may see an expired lock. 

**Expected behavior**

The generated timestamps should be correct, even during daylight savings transitions.

**Sample**

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/10906

The `Timestamp.valueOf()` method uses the system default time zone which may lead to discrepancy when daylight saving happens

According to JDBC 4.2+, we don't need to convert `LocalDateTime` to `Timestamp` anymore

* Fix `DefaultLockRepository` to use `LocalDateTime` as values for SQL statements instead of wrapping them into `Timestamp`

There are no any new tests since this internal change does not affect API logic, and existing working tests prove that everything is still right

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/DefaultLockRepository.java
@@ -16,7 +16,6 @@
 
 package org.springframework.integration.jdbc.lock;
 
-import java.sql.Timestamp;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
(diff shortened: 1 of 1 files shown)


cppwfs:
LGTM!

Issue #10782, opened 2026-02-06
Refactor RedisLockRegistry to leverage Redis 8.4 native CAS/CAD commands


Dgramada:
`RedisLockRegistry` should leverage the new native Compare-And-Set (CAS) and Compare-And-Delete (CAD) commands introduced in [Redis 8.4](https://redis.io/docs/latest/develop/whats-new/8-4/), which will be available in Spring Data Redis 4.1.0 through the new RedisTemplate API methods:

- setIfEqual(key, oldValue, newValue)
- setIfNotEqual(key, oldValue, newValue)
- deleteIfEqual(key, value)
- deleteIfNotEqual(key, value)

A proper example would be the `RedisSpinLock` which could use the new `deleteIfEqual(lockKey, clientId)` template method instead of the current Lua script.
Furthermore, the current `OBTAIN_LOCK_SCRIPT` could be simplified by calling for example [SET](https://redis.io/docs/latest/commands/set/#options) with the IFEQ option instead of calling two separate commands - `GET` and `PEXPIRE`.

**Current Behavior**

`RedisLockRegistry` currently uses Lua scripts to implement CAS/CAD patterns for distributed locking by calling multiple commands.

Additionally, `RedisMessageStore.doRemove() ` performs two separate operations which can be replaced with the currently available template command `setGet() ` introduced in Redis 6.2 as [GETDEL](https://redis.io/docs/latest/commands/getdel/).

**Context**

We are from the Redis OSS ecosystem team and would like to contribute this enhancement to spring-integration-redis.

The current Lua script approach works correctly but introduces overhead from script parsing and execution. With Redis 8.4 introducing native CAS/CAD commands, there's an opportunity to simplify the implementation and improve performance.

We aim to:
- Replace applicable Lua scripts with native Redis 8.4 CAS/CAD commands
- Simplify code maintenance by using direct API calls instead of embedded Lua scripts
- Improve debuggability (native commands are easier to trace in Redis logs)
- Optimize RedisMessageStore.doRemove() using GETDEL for atomic get-and-delete

**Questions for the maintainers:**

1. Is this change meaningful to you?
3. Do you have backwards compatibility requirements?
4. Should we maintain support for Redis versions prior to 8.4?
5. If yes, what is your preferred approach:

- Runtime detection with fallback to Lua scripts?
- Configuration option to choose implementation?
- Separate implementation classes?
6. What is the lowest Redis version you would like to support?

artembilan:
[...]

PR #10930, opened 2026-04-09
Use CAS/CAD commands in RedisLockRegistry

Dgramada:
References #10782 

This PR updates `RedisLockRegistry` to prefer Redis CAS (Compare-And-Set) and CAD (Compare-And-Delete) commands over Lua scripts for lock renewal and lock release in `RedisSpinLock`.

CAS/CAD are native Redis commands ([since. 8.4](https://redis.io/docs/latest/develop/whats-new/8-4/#enhanced-string-operations)) that provide the same conditional semantics as the equivalent Lua scripts but without the overhead of script evaluation.


spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -591,13 +603,42 @@ private void removeLockKey() {
...
+							RedisLockRegistry.this.clientId,
+							spec -> spec.ifEquals()
+									.value(RedisLockRegistry.this.clientId)
+									.expire(Duration.ofMillis(expireAfter)));
+				}
+				catch (RedisSystemException | InvalidDataAccessApiUsageException ex) {
+					if (isCasCadNotSupportedError(ex)) {
+						LOGGER.debug("CAS/CAD for value operations not supported, falling back to Lua script", ex);
+						RedisLockRegistry.this.supportsCasCadOperations = false;
+						res = RedisLockRegistry.this.redisTemplate.execute(
```
cppwfs: The LUA script execution can be put into a method so that we can reduce code duplication.   
Meaning we can use that method here at line 620 and 630.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -833,11 +874,29 @@ protected boolean tryRedisLockInner(long time, long expireAfter) throws Interrup
...
-					RedisLockRegistry.this.clientId));
+			if (RedisLockRegistry.this.supportsCasCadOperations) {
+				try {
+					return RedisLockRegistry.this.redisTemplate.delete(this.lockKey, it -> it.ifEquals().value(RedisLockRegistry.this.clientId));
+				}
+				catch (RedisSystemException | InvalidDataAccessApiUsageException ex) {
+					if (isCasCadNotSupportedError(ex)) {
+						LOGGER.debug("CAS/CAD for value operations not supported, falling back to Lua script", ex);
+						RedisLockRegistry.this.supportsCasCadOperations = false;
+						return Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
```
cppwfs: The LUA script execution can be put into a method so that we can reduce code duplication.
Meaning we can use that method here at line 885 and 895.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1001,93 @@ void testInitialiseWithCustomExecutor() {
...
+	}
+
+	@Test
+	void testUnlockFallbackWhenCasCadNotSupported() {
+		RedisLockRegistry registry = new RedisLockRegistry(redisConnectionFactory, this.registryKey);
+		registry.setRedisLockType(testRedisLockType);
+
+		ReflectionTestUtils.setField(registry, "supportsCasCadOperations", false);
+
+		Lock lock = registry.obtain("foo");
```
cppwfs: nitpick:
Can we change the name of the lock from "foo" to "testLock" in the tests.

Thanks!

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1001,93 @@ void testInitialiseWithCustomExecutor() {
...
+
+		Lock lock2 = registry2.obtain("foo");
+		assertThat(lock2.tryLock()).isTrue();
+		try {
+			assertThatThrownBy(lock1::unlock).isInstanceOf(ConcurrentModificationException.class);
+		}
+		finally {
+			lock2.unlock();
+		}
+		registry1.destroy();
```
cppwfs: probably want to move the `registry1.destroy` and `registry2.destroy` in a finally block to make sure they are destroyed in the case that an exception is thrown.

src/reference/antora/modules/ROOT/pages/redis.adoc
```diff
@@ -1073,6 +866,9 @@ Starting with version 7.0, the `RedisLock` implements `DistributedLock` interfac
 A `RedisLock` can now be acquired using the `lock(Duration ttl)` or `tryLock(long time, TimeUnit unit, Duration ttl)` method, with a specified time-to-live (TTL) value.
 The `RedisLockRegistry` now provides new `renewLock(Object lockKey, Duration ttl)` method, allowing to renew the lock with a custom time-to-live value.
 
+Starting with version 7.1, the `RedisLockRegistry` prefers native Redis CAS (Compare-And-Set) and CAD (Compare-And-Delete) commands over Lua scripts for lock renewal and spin-lock release.
```
cppwfs: Reword this as:
Starting with version 7.1, the `RedisLockRegistry` uses native Redis CAS (Compare-And-Set) and CAD (Compare-And-Delete) commands for Redis 8.4 or later. For older Redis versions, the registry automatically falls back to the previous Lua script-based approach.

cppwfs (changes requested):

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -373,6 +378,13 @@ private Function<String, RedisLock> getRedisLockConstructor(RedisLockType redisL
 		};
 	}
 
+	private boolean isCasCadNotSupportedError(Exception ex) {
```
artembilan: `static` ?

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -591,13 +603,42 @@ private void removeLockKey() {
...
 		protected final boolean renew(long expireAfter) {
-			boolean res = Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
-					RENEW_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
-					RedisLockRegistry.this.clientId, String.valueOf(expireAfter)));
-			if (!res) {
+			Boolean res;
+
+			if (RedisLockRegistry.this.supportsCasCadOperations) {
+				try {
+					res = RedisLockRegistry.this.redisTemplate.boundValueOps(this.lockKey).set(
```
artembilan: Can `boundValueOps` be extracted into an instance property?
The `RedisLock` ctor looks like a nice place for that one.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -591,13 +603,42 @@ private void removeLockKey() {
...
+							RedisLockRegistry.this.clientId,
+							spec -> spec.ifEquals()
+									.value(RedisLockRegistry.this.clientId)
+									.expire(Duration.ofMillis(expireAfter)));
+				}
+				catch (RedisSystemException | InvalidDataAccessApiUsageException ex) {
+					if (isCasCadNotSupportedError(ex)) {
+						LOGGER.debug("CAS/CAD for value operations not supported, falling back to Lua script", ex);
```
artembilan: warn() instead?

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -833,11 +874,29 @@ protected boolean tryRedisLockInner(long time, long expireAfter) throws Interrup
...
+					}
+				}
+			}
+			else {
+				return Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
+						UNLINK_UNLOCK_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
+						RedisLockRegistry.this.clientId));
+			}
 		}
-
```
artembilan, line 840: The blank line still must be present.
Every class member should be surrounded with blank line to make code more readable.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1001,93 @@ void testInitialiseWithCustomExecutor() {
 		assertThatNoException().isThrownBy(() -> redisLockRegistry.setExecutor(mock()));
 	}
 
+	@Test
+	void testRenewFallbackWhenCasCadNotSupported() {
+		RedisLockRegistry registry = new RedisLockRegistry(redisConnectionFactory, this.registryKey);
+		registry.setRedisLockType(testRedisLockType);
+
+		ReflectionTestUtils.setField(registry, "supportsCasCadOperations", false);
```
artembilan: Doesn't look like this is about falback.
With this you explicitly say that we even don't try CasCad.
Consider to spy `RedisTemplate` and stub those respective methods to fail with error.
Then we really would see a fallback in action.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1001,93 @@ void testInitialiseWithCustomExecutor() {
...
+	void testRenewActuallyExtendsTtl() throws InterruptedException {
+		long shortExpiry = 200;
+		RedisLockRegistry registry = new RedisLockRegistry(redisConnectionFactory, this.registryKey, shortExpiry);
+		registry.setRedisLockType(testRedisLockType);
+
+		Lock lock = registry.obtain("foo");
+		assertThat(lock.tryLock()).isTrue();
+		try {
+			registry.renewLock("foo", Duration.ofSeconds(10));
+			Thread.sleep(shortExpiry + 100);
```
artembilan: This is not reliable test.
The CPU could be busy with something else much longer than expire period.
It might be better to verify that TTL was extended after `renew` call.
But sleep and then check if it is still there is not OK.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1001,93 @@ void testInitialiseWithCustomExecutor() {
...
+		registry.setRedisLockType(testRedisLockType);
+
+		assertThat(TestUtils.<Boolean>getPropertyValue(registry, "supportsCasCadOperations")).isTrue();
+
+		ReflectionTestUtils.setField(registry, "supportsCasCadOperations", false);
+
+		Lock lock = registry.obtain("foo");
+		lock.lock();
+		assertThatNoException().isThrownBy(lock::unlock);
+		registry.destroy();
```
artembilan: I think this test does not have a any value.
Can we consider to remove it altogether?

src/reference/antora/modules/ROOT/pages/redis.adoc
artembilan: ... and one sentence per line!
https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

artembilan (changes requested):
Looks promising.
Thank you!

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -173,6 +176,8 @@ protected boolean removeEldestEntry(Entry<String, RedisLock> eldest) {
 	 */
 	private volatile @Nullable RedisMessageListenerContainer redisMessageListenerContainer;
 
+	private volatile boolean supportsCasCadOperations = true;
```
cppwfs, line 183: Need to provide a javadoc for this variable.

spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
```diff
@@ -833,11 +874,29 @@ protected boolean tryRedisLockInner(long time, long expireAfter) throws Interrup
 
 		@Override
 		protected boolean removeLockKeyInnerUnlink() {
-			return Boolean.TRUE.equals(RedisLockRegistry.this.redisTemplate.execute(
-					UNLINK_UNLOCK_REDIS_SCRIPT, Collections.singletonList(this.lockKey),
-					RedisLockRegistry.this.clientId));
+			if (RedisLockRegistry.this.supportsCasCadOperations) {
+				try {
+					return RedisLockRegistry.this.redisTemplate.delete(this.lockKey, it -> it.ifEquals().value(RedisLockRegistry.this.clientId));
```
cppwfs, line 888: On line 450 the javadocs states we do an unlink, but here we are doing a delete.   The javadoc needs to be updated to state that for redis 8.4+ we do a delete else its an unlink.

cppwfs (changes requested):
Again thank you for the contribution.  Just a couple more requests.

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1007,68 @@ void testInitialiseWithCustomExecutor() {
 		assertThatNoException().isThrownBy(() -> redisLockRegistry.setExecutor(mock()));
 	}
 
+	@Test
+	void testRenewFallbackWhenCasCadNotSupported() {
+		RedisLockRegistry registry = new RedisLockRegistry(redisConnectionFactory, this.registryKey);
+		registry.setRedisLockType(testRedisLockType);
+
+		Lock lock = registry.obtain("foo");
```
artembilan: Let's don't use foo/bar language in our code!

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1007,68 @@ void testInitialiseWithCustomExecutor() {
...
+		try {
+			BoundValueOperations<String, String> boundValueOps = mock();
+			ReflectionTestUtils.setField(lock, "boundValueOps", boundValueOps);
+
+			given(boundValueOps.set(any(), (Consumer<SetSpec<String, String>>) any()))
+					.willThrow(new RedisSystemException("CAS failed", new RuntimeException("ERR unknown command")));
+
+			assertThat(TestUtils.<Boolean>getPropertyValue(registry, "supportsCasCadOperations")).isTrue();
+
+			registry.renewLock("foo");
```
artembilan: DITTO

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -999,6 +1007,68 @@ void testInitialiseWithCustomExecutor() {
...
+	}
+
+	@Test
+	void testUnlockDoesNotDeleteOtherClientsLock() throws Exception {
+		RedisLockRegistry registry1 = new RedisLockRegistry(redisConnectionFactory, this.registryKey, 100);
+		registry1.setRedisLockType(testRedisLockType);
+		RedisLockRegistry registry2 = new RedisLockRegistry(redisConnectionFactory, this.registryKey, 10000);
+		registry2.setRedisLockType(testRedisLockType);
+
+		Lock lock1 = registry1.obtain("foo");
```
artembilan: DITTO for all `foo` in this test

src/reference/antora/modules/ROOT/pages/redis.adoc
```diff
@@ -1333,6 +1333,8 @@ Starting with version 7.0, the `RedisLock` implements `DistributedLock` interfac
 A `RedisLock` can now be acquired using the `lock(Duration ttl)` or `tryLock(long time, TimeUnit unit, Duration ttl)` method, with a specified time-to-live (TTL) value.
 The `RedisLockRegistry` now provides new `renewLock(Object lockKey, Duration ttl)` method, allowing to renew the lock with a custom time-to-live value.
 
+Starting with version 7.1, the RedisLockRegistry uses native Redis CAS (Compare-And-Set) and CAD (Compare-And-Delete) commands for Redis 8.4 or later. For older Redis versions, the registry automatically falls back to the previous Lua script-based approach.
```
artembilan: One sentence per line: https://asciidoctor.org/docs/asciidoc-recommended-practices/

artembilan (changes requested):
Just couple minor nit-picks.

Thank you!

artembilan:
@Dgramada ,

PR #10932, opened 2026-04-10
Update 3rd party dependencies to next major version

cppwfs:

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -10,7 +10,7 @@ camelVersion = "4.18.1"
 cloudEventsVersion = "4.0.1"
 commonsDbcp2Version = "2.14.0"
 commonsIoVersion = "2.21.0"
-commonsNetVersion = "3.12.0"
+commonsNetVersion = "3.13.0"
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -155,7 +155,7 @@ spring-ws-bom = { module = "org.springframework.ws:spring-ws-bom", version.ref =
 tools-jackson-bom = { module = "tools.jackson:jackson-bom", version.ref = "jackson3Version" }
 
 [plugins]
-dokka = { id = "org.jetbrains.dokka", version = "2.1.0" }
+dokka = { id = "org.jetbrains.dokka", version = "2.2.0" }
```
artembilan, line 158: Can you run Dokka generation locally and make sure the result are OK looking for us?
I mean this is minor version jump, but still.

Thanks

artembilan (changes requested):

gradle/libs.versions.toml
cppwfs: kdocs are being generated and look the same as what is generated in main.

PR #10933, opened 2026-04-10
Add JmsTemplate overloads to JMS DSL channel specs

### LINKED ISSUE #6656: JmsChannelFactoryBean instanciating it's own jmsTemplate prevents mocking for test  [INT-2679]
**[Sylvain Mougenot](https://jira.spring.io/secure/ViewProfile.jspa?name=smougenot)** opened **[INT-2679](https://jira.spring.io/browse/INT-2679?redirect=false)** and commented

Using jms:outbound-channel-adapter it's not possible to mock jmsTemplate (for test purposes) because JmsChannelFactoryBean  instanciate it's own.

```
private final JmsTemplate jmsTemplate = new DynamicJmsTemplate();
```

It would be nice if we could inject the jmsTemplate.

---

**Affects:** 2.1.3

cppwfs:
Fixes: #6656

- Allow `JmsChannelFactoryBean` to accept an externally provided `JmsTemplate`
- Add `Jms.pollableChannel(JmsTemplate)` and `Jms.pollableChannel(String, JmsTemplate)` factory methods
- Add `Jms.channel(JmsTemplate)` factory method
- Add protected constructors to `JmsPollableMessageChannelSpec` and `JmsMessageChannelSpec` accepting `JmsTemplate`
- Add integration test verifying the polling flow with a `JmsTemplate`

### DIFF
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/config/JmsChannelFactoryBean.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/config/JmsChannelFactoryBean.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.jms.config;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Executor;
 
@@ -70,6 +71,10 @@ public class JmsChannelFactoryBean extends AbstractFactoryBean<AbstractJmsChanne
 	private final boolean messageDriven;
 
 	private JmsTemplate jmsTemplate = new DynamicJmsTemplate();
+
+	private boolean jmsTemplateExplicitlySet;
+
+	private final List<String> templateDelegatingOptions = new ArrayList<>();
 
 	private @Nullable Class<? extends AbstractMessageListenerContainer> containerType;
 
@@ -165,30 +170,37 @@ public class JmsChannelFactoryBean extends AbstractFactoryBean<AbstractJmsChanne
 	 */
 
 	public void setDeliveryPersistent(boolean deliveryPersistent) {
+		recordJmsTemplateOption("deliveryPersistent");
 		this.jmsTemplate.setDeliveryPersistent(deliveryPersistent);
 	}
 
 	public void setExplicitQosEnabled(boolean explicitQosEnabled) {
+		recordJmsTemplateOption("explicitQosEnabled");
 		this.jmsTemplate.setExplicitQosEnabled(explicitQosEnabled);
 	}
 
 	public void setMessageConverter(MessageConverter messageConverter) {
+		recordJmsTemplateOption("messageConverter");
 		this.jmsTemplate.setMessageConverter(messageConverter);
 	}
 
 	public void setMessageIdEnabled(boolean messageIdEnabled) {
+		recordJmsTemplateOption("messageIdEnabled");
 		this.jmsTemplate.setMessageIdEnabled(messageIdEnabled);
 	}
 
 	public void setMessageTimestampEnabled(boolean messageTimestampEnabled) {
+		recordJmsTemplateOption("messageTimestampEnabled");
 		this.jmsTemplate.setMessageTimestampEnabled(messageTimestampEnabled);
 	}
 
 	public void setPriority(int priority) {
+		recordJmsTemplateOption("priority");
 		this.jmsTemplate.setPriority(priority);
 	}
 
 	public void setTimeToLive(long timeToLive) {
+		recordJmsTemplateOption("timeToLive");
 		this.jmsTemplate.setTimeToLive(timeToLive);
 	}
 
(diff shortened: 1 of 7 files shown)


spring-integration-jms/src/main/java/org/springframework/integration/jms/config/JmsChannelFactoryBean.java
```diff
@@ -325,6 +326,10 @@ public void setSessionAcknowledgeMode(int sessionAcknowledgeMode) {
 		this.jmsTemplate.setSessionAcknowledgeMode(sessionAcknowledgeMode);
 	}
 
+	public void setJmsTemplate(JmsTemplate jmsTemplate) {
```
artembilan, line 355: `@since` and some other Javadocs.

And now, having this externally injected we need to protect against mutation.
Pay attention to all those setters delegating to the `this.jmsTemplate`.
We need to reject with something like `assertLocalClient()` in the `HttpRequestExecutingMessageHandler`.
The change is not one-to-one since here we don't have ctor injection, so might be the case that validation should go to the `createInstance()` (or `initializeJmsTemplate()`) when we would not worry about an order all those setters are called.

spring-integration-jms/src/main/java/org/springframework/integration/jms/dsl/Jms.java
```diff
@@ -46,6 +47,18 @@ public static JmsPollableMessageChannelSpec<?, PollableJmsChannel> pollableChann
 		return new JmsPollableMessageChannelSpec<>(connectionFactory);
 	}
 
+	/**
+	 * The template to produce a {@link JmsPollableMessageChannelSpec}.
```
artembilan: Maybe factory?
I'm not that "template" is producing anything.
It might be used from something what is producing, like factory 😄 
Anyway, this is more about a proper wording in the Javadoc like we have in other methods of this class.

spring-integration-jms/src/main/java/org/springframework/integration/jms/dsl/JmsPollableMessageChannelSpec.java
```diff
@@ -57,6 +63,15 @@ protected JmsPollableMessageChannelSpec(JmsChannelFactoryBean jmsChannelFactoryB
 		this.jmsChannelFactoryBean.setBeanFactory(new DefaultListableBeanFactory());
 	}
 
+	protected JmsPollableMessageChannelSpec(JmsChannelFactoryBean jmsChannelFactoryBean,
+			JmsTemplate jmsTemplate) {
+
+		this.jmsChannelFactoryBean = jmsChannelFactoryBean;
+		this.jmsChannelFactoryBean.setJmsTemplate(jmsTemplate);
+		this.jmsChannelFactoryBean.setSingleton(false);
+		this.jmsChannelFactoryBean.setBeanFactory(new DefaultListableBeanFactory());
```
artembilan: Let's consider to move two last lines into the `doGet()` before `getObject()`.
This way we might not have duplication in these two ctors.

artembilan (changes requested):
Some mentioning of this new option in the `whats-new.adoc` would be great.

Thanks

spring-integration-jms/src/main/java/org/springframework/integration/jms/config/JmsChannelFactoryBean.java
```diff
@@ -411,6 +443,15 @@ protected AbstractJmsChannel createInstance() {
 	private void initializeJmsTemplate() {
 		Assert.isTrue(this.destination != null ^ this.destinationName != null,
 				"Exactly one of destination or destinationName is required.");
+		Assert.isTrue(!this.jmsTemplateExplicitlySet || this.templateDelegatingOptions.isEmpty(),
+				() -> {
+					String quoted = this.templateDelegatingOptions.stream()
+							.map(opt -> "'" + opt + "'")
+							.reduce((a, b) -> a + ", " + b)
+							.orElse("");
```
artembilan, line 451: This is interesting, but I think we don't need to track options being set.
Let's just say straight in this error message what options must be set on the template instead if this factory bean!
I mean if at least one of those properties is sent, we set respective boolean flag and check it here.
Then if both for those are `true` we raise an exception.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -62,6 +62,10 @@ See xref:redis.adoc[] for more information.
 JMS-backed message channels now map headers to JMS message properties and back.
 See xref:jms.adoc[] for more information.
 
+JMS DSL now allows the user to set the `JmsTemplate`.
+When using an external `JmsTemplate` it must be configured prior to when it is set, because the DSL can only set the
+configurations for the default `JMSTemplate`.
```
artembilan, line 67: First of all typo in the `JmsTemplate` word.
And second: one sentence per line: https://asciidoctor.org/docs/asciidoc-recommended-practices/.
If if it is too long, it is still has to be on one line.
The `120` is just recommendation for code. 
And I believe it might even not apply for XML or Gradle.
Just Java, Kotlin, Groovy source code.

artembilan (changes requested):

PR #10934, opened 2026-04-11
GH-10931: Add setAllowedOriginPatterns support to ServerWebSocketContainer

### LINKED ISSUE #10931: ServerWebSocketContainer.registerWebSocketHandlers() uses setAllowedOrigins() instead of setAllowedOriginPatterns(), causing incompatibility with allowCredentials=true in Spring Framework 7
## Title

`ServerWebSocketContainer.registerWebSocketHandlers()` uses `setAllowedOrigins()` instead of `setAllowedOriginPatterns()`, causing incompatibility with `allowCredentials=true` in Spring Framework 7

---

# Bug Report: ServerWebSocketContainer incompatible with allowCredentials=true in Spring Framework 7

**Repository:** https://github.com/spring-projects/spring-integration  
**Component:** spring-integration-websocket  
**Spring Integration version:** 7.0.4  
**Spring Framework version:** 7.0.6  
**Spring Boot version:** 4.0.5  

---

## Describe the bug

`ServerWebSocketContainer.registerWebSocketHandlers()` internally calls `WebSocketHandlerRegistration.setAllowedOrigins(this.origins)`, which is incompatible with `allowCredentials=true` when the wildcard `"*"` is used. Spring Framework 7 throws:

```
java.lang.IllegalArgumentException: When allowCredentials is true, allowedOrigins
cannot contain the special value "*" since that cannot be set on the
"Access-Control-Allow-Origin" response header. To allow credentials to a set of
origins, list them explicitly or consider using "allowedOriginPatterns" instead.
```

The root cause is in `ServerWebSocketContainer.java` line 168:

```java
public void registerWebSocketHandlers(WebSocketHandlerRegistry registry) {
    ...
    WebSocketHandlerRegistration registration =
        registry.addHandler(webSocketHandler, this.paths)
                .addInterceptors(this.interceptors)
                .setAllowedOrigins(this.origins);  // <-- should support setAllowedOriginPatterns()
    ...
}
```

`WebSocketHandlerRegistration` has had `setAllowedOriginPatterns()` available since Spring Framework 5.3, but `ServerWebSocketContainer` was never updated to expose or use it.

---

## Steps to reproduce

1. Use Spring Boot 4.0.5 with Spring Integration 7.0.4
2. Configure `ServerWebSocketContainer` with SockJS enabled
3. Have a cross-origin client connect (e.g. UI hosted on a different port)
4. The client sends `GET /messages/info` with an `Origin` header

**Result:** `IllegalArgumentException` thrown by `AbstractSockJsService`, returning HTTP 403 to the client.

**Expected:** The connection succeeds, CORS headers are set correctly.

---

## Stack trace

```
org.springframework.web.socket.sockjs.SockJsException: Uncaught failure in SockJS request,
uri=http://localhost:8081/messages/info
    at ...AbstractSockJsService...
Caused by: java.lang.IllegalArgumentException: When allowCredentials is true,
allowedOrigins cannot contain the special value "*" since that cannot be set on the
"Access-Control-Allow-Origin" response header. To allow credentials to a set of origins,
list them explicitly or consider using "allowedOriginPatterns" instead.
```

---

## Current workaround

Override `registerWebSocketHandlers()` via an anonymous subclass to call `setAllowedOriginPatterns()` directly on the registration:

```java
@Bean
ServerWebSocketContainer serverWebSocketContainer() {
    return new ServerWebSocketContainer("/messages") {
        @Override
        public void registerWebSocketHandlers(WebSocketHandlerRegistry registry) {
            registry.addHandler(getWebSocketHandler(), "/messages")
                    .setAllowedOriginPatterns("*")
                    .withSockJS()
                    .setSuppressCors(true);
        }
    };
}
```

---

## Suggested fix

1. Add `setAllowedOriginPatterns(String... patterns)` method to `ServerWebSocketContainer`
2. Store the patterns in a separate `private String[] originPatterns` field
3. Update `registerWebSocketHandlers()` to prefer `setAllowedOriginPatterns()` when patterns are configured:

```java
WebSocketHandlerRegistration registration =
    registry.addHandler(webSocketHandler, this.paths)
            .addInterceptors(this.interceptors);

if (this.originPatterns.length > 0) {
    registration.setAllowedOriginPatterns(this.originPatterns);
} else {
    registration.setAllowedOrigins(this.origins);
}
```

marcin-nowicki-pl:
`ServerWebSocketContainer.registerWebSocketHandlers()` hardcoded the use of
`setAllowedOrigins()`, making it impossible to use wildcard patterns (e.g. `"*"`)
together with `allowCredentials=true` in Spring Framework 7+, which forbids that
combination.

This commit adds a `setAllowedOriginPatterns(String...)` method to
`ServerWebSocketContainer` that, when configured, is preferred over
`setAllowedOrigins()` in `registerWebSocketHandlers()`. The existing
`setAllowedOrigins()` behavior is preserved for backward compatibility.

Fixes: #10931

### DIFF
--- a/spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
+++ b/spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
@@ -51,6 +51,7 @@ import org.springframework.web.socket.sockjs.transport.TransportHandler;
  * @author Gary Russell
  * @author Christian Tzolov
  * @author Jooyoung Pyoung
+ * @author Marcin Nowicki
  *
  * @since 4.1
  */
@@ -68,6 +69,8 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 	private @Nullable SockJsServiceOptions sockJsServiceOptions;
 
 	private String[] origins = {};
+
+	private String[] originPatterns = {};
 
 	private boolean autoStartup = true;
 
@@ -121,6 +124,18 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 		return this;
 	}
 
+	/**
+	 * Specify origin patterns for which cross-origin requests are allowed from a browser.
+	 * @param originPatterns the origin patterns to allow.
+	 * @return the current ServerWebSocketContainer
+	 * @since 7.0.5
+	 * @see WebSocketHandlerRegistration#setAllowedOriginPatterns(String...)
+	 */
+	public ServerWebSocketContainer setAllowedOriginPatterns(String... originPatterns) {
+		this.originPatterns = Arrays.copyOf(originPatterns, originPatterns.length);
+		return this;
+	}
+
 	public ServerWebSocketContainer withSockJs(SockJsServiceOptions... sockJsServiceOptions) {
 		if (ObjectUtils.isEmpty(sockJsServiceOptions)) {
 			setSockJsServiceOptions(new SockJsServiceOptions());
@@ -164,8 +179,14 @@ public class ServerWebSocketContainer extends IntegrationWebSocketContainer
 
 		WebSocketHandlerRegistration registration =
 				registry.addHandler(webSocketHandler, this.paths)
-						.addInterceptors(this.interceptors)
-						.setAllowedOrigins(this.origins);
+						.addInterceptors(this.interceptors);
+
+		if (!ObjectUtils.isEmpty(this.originPatterns)) {
+			registration.setAllowedOriginPatterns(this.originPatterns);
+		}
+		if (!ObjectUtils.isEmpty(this.origins)) {
+			registration.setAllowedOrigins(this.origins);
+		}
 
 		if (this.handshakeHandler != null) {
 			registration.setHandshakeHandler(this.handshakeHandler);
--- a/spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
+++ b/spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.websocket.server;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Answers;
+
+import org.springframework.integration.websocket.ServerWebSocketContainer;
+import org.springframework.web.socket.config.annotation.WebSocketHandlerRegistration;
+import org.springframework.web.socket.config.annotation.WebSocketHandlerRegistry;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for {@link ServerWebSocketContainer}.
+ *
+ * @author Marcin Nowicki
+ *
+ * @since 7.0.5
+ */
+public class ServerWebSocketContainerTests {
+
+	@Test
+	public void setAllowedOriginPatternsIsUsedInRegistration() {
+		WebSocketHandlerRegistry registry = mock();
+		WebSocketHandlerRegistration registration = mock(Answers.RETURNS_SELF);
+		given(registry.addHandler(any(), any(String[].class))).willReturn(registration);
+
+		ServerWebSocketContainer container = new ServerWebSocketContainer("/test")
+			.setAllowedOriginPatterns("https://example.com");
+
+		container.registerWebSocketHandlers(registry);
+
+		verify(registration).setAllowedOriginPatterns("https://example.com");
+		verify(registration, never()).setAllowedOrigins(any());
+	}
+
+	@Test
+	public void setAllowedOriginsIsUsedWhenNoPatternsConfigured() {
+		WebSocketHandlerRegistry registry = mock();
+		WebSocketHandlerRegistration registration = mock(Answers.RETURNS_SELF);
+		given(registry.addHandler(any(), any(String[].class))).willReturn(registration);
+
+		ServerWebSocketContainer container = new ServerWebSocketContainer("/test")
+			.setAllowedOrigins("https://example.com");
+
+		container.registerWebSocketHandlers(registry);
+
+		verify(registration).setAllowedOrigins("https://example.com");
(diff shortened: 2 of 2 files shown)


spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -164,8 +179,14 @@ public void registerWebSocketHandlers(WebSocketHandlerRegistry registry) {
...
 		WebSocketHandlerRegistration registration =
 				registry.addHandler(webSocketHandler, this.paths)
-						.addInterceptors(this.interceptors)
-						.setAllowedOrigins(this.origins);
+						.addInterceptors(this.interceptors);
+
+		if (!ObjectUtils.isEmpty(this.originPatterns)) {
+			registration.setAllowedOriginPatterns(this.originPatterns);
+		}
+		else {
```
artembilan: I don't think this is correct.
See `AbstractWebSocketHandlerRegistration` logic:
```
		if (!this.allowedOrigins.isEmpty()) {
			this.sockJsServiceRegistration.setAllowedOrigins(StringUtils.toStringArray(this.allowedOrigins));
		}
		if (!this.allowedOriginPatterns.isEmpty()) {
			this.sockJsServiceRegistration.setAllowedOriginPatterns(
					StringUtils.toStringArray(this.allowedOriginPatterns));
		}
```
As well as the target `SockJsServiceRegistration`:
```
service.setAllowedOrigins(this.allowedOrigins);
service.setAllowedOriginPatterns(this.allowedOriginPatterns);
```

Or even `CorsConfiguration`:
```
	public @Nullable String checkOrigin(@Nullable String origin) {
		if (!StringUtils.hasText(origin)) {
			return null;
		}
		String originToCheck = trimTrailingSlash(origin);
		if (!ObjectUtils.isEmpty(this.allowedOrigins)) {
			if (this.allowedOrigins.contains(ALL)) {
				validateAllowCredentials();
				validateAllowPrivateNetwork();
				return ALL;
			}
			for (String allowedOrigin : this.allowedOrigins) {
				if (originToCheck.equalsIgnoreCase(allowedOrigin)) {
					return origin;
				}
			}
		}
		if (!ObjectUtils.isEmpty(this.allowedOriginPatterns)) {
			for (OriginPattern p : this.allowedOriginPatterns) {
				if (p.getDeclaredPattern().equals(ALL) || p.getPattern().matcher(originToCheck).matches()) {
					return origin;
				}
			}
		}
		return null;
	}
```

So, I vote for not `else`, but same as it is done in Spring Framework.

spring-integration-websocket/src/main/java/org/springframework/integration/websocket/ServerWebSocketContainer.java
```diff
@@ -121,6 +124,18 @@ public ServerWebSocketContainer setAllowedOrigins(String... origins) {
 		return this;
 	}
 
+	/**
+	 * Specify origin patterns for which cross-origin requests are allowed from a browser.
+	 * @param originPatterns the origin patterns to allow.
+	 * @return the current ServerWebSocketContainer
+	 * @since 7.1
```
artembilan: We are going to back-port this to `7.0.x`.
Therefore: `7.0.5`

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @author Marcin Nowicki
+ *
+ * @since 7.1
+ */
+public class ServerWebSocketContainerTests {
+
+	@Test
+	public void setAllowedOriginPatternsIsUsedInRegistration() {
+		WebSocketHandlerRegistry registry = mock(WebSocketHandlerRegistry.class);
```
artembilan: Please, use  just `mock()` without class arg.

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
```diff
@@ -0,0 +1,71 @@
...
+		WebSocketHandlerRegistration registration = mock(WebSocketHandlerRegistration.class, Answers.RETURNS_SELF);
```
artembilan: DITTO

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
```diff
@@ -0,0 +1,71 @@
...
+		container.registerWebSocketHandlers(registry);
+
+		verify(registration).setAllowedOriginPatterns("https://example.com");
+		verify(registration, never()).setAllowedOrigins(any());
+	}
+
+	@Test
+	public void setAllowedOriginsIsUsedWhenNoPatternsConfigured() {
+		WebSocketHandlerRegistry registry = mock(WebSocketHandlerRegistry.class);
+		WebSocketHandlerRegistration registration = mock(WebSocketHandlerRegistration.class, Answers.RETURNS_SELF);
```
artembilan: And repeat same type removal in this test, too.

artembilan (changes requested):
Looks promising!

Thank you!

marcin-nowicki-pl:
All changes applied. Thank you for your time and the review!

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
```diff
@@ -0,0 +1,71 @@
...
+		WebSocketHandlerRegistration registration = mock(Answers.RETURNS_SELF);
+		given(registry.addHandler(any(), any(String[].class))).willReturn(registration);
+
+		ServerWebSocketContainer container = new ServerWebSocketContainer("/test")
+			.setAllowedOrigins("https://example.com");
+
+		container.registerWebSocketHandlers(registry);
+
+		verify(registration).setAllowedOrigins("https://example.com");
+		verify(registration, never()).setAllowedOriginPatterns(any());
```
artembilan, line 68: ... and one more test when both of them are provided 😄 ?

artembilan (changes requested):

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
marcin-nowicki-pl: right...

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for {@link ServerWebSocketContainer}.
+ *
+ * @author Marcin Nowicki
+ *
+ * @since 7.1
```
artembilan: Missed this to be changed to `7.0.5`

artembilan (changes requested):

spring-integration-websocket/src/test/java/org/springframework/integration/websocket/server/ServerWebSocketContainerTests.java
marcin-nowicki-pl: updated, thanks again for your patience and time, really appreciated!

artembilan:
@marcin-nowicki-pl ,

marcin-nowicki-pl:
@artembilan ,

Thank you very much for your guidance, support, patience, and time. It was a great honor to participate and contribute. Looking forward for more too!

PR #10941, opened 2026-04-13
GH-8108: Add DSL for Redis Outbound Command Gateway.

mjd507:
Related to: https://github.com/spring-projects/spring-integration/issues/8108

### DIFF
--- a/src/reference/antora/modules/ROOT/pages/redis.adoc
+++ b/src/reference/antora/modules/ROOT/pages/redis.adoc
@@ -1067,7 +1067,7 @@ Java DSL::
 IntegrationFlow outboundGatewayFlow(RedisConnectionFactory redisConnectionFactory) {
     return IntegrationFlow.from("requestChannel")
             .handle(Redis.outboundGatewaySpec(redisConnectionFactory)
-                    .commandExpression("'INCR'"))
+                    .command("INCR"))
             .channel(c -> c.queue("replyChannel"))
             .get();
 }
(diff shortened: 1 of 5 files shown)


src/reference/antora/modules/ROOT/pages/redis.adoc
```diff
@@ -1002,13 +1058,46 @@ If neither attribute is provided, the `payload` is used as the command arguments
...
+[tabs]
+======
+Java DSL::
++
+----
+@Bean
+IntegrationFlow outboundGatewayFlow(RedisConnectionFactory redisConnectionFactory) {
+    return IntegrationFlow.from("requestChannel")
+            .handle(Redis.outboundGatewaySpec(redisConnectionFactory)
+                    .commandExpression("'INCR'"))
```
artembilan: Looking into this I feelike simple `command("INCR")` would be nice to have.
Yes, that would go to the `LiteralExpression` delegation, but still good for end-user experience.

artembilan (changes requested):
Looks great!
Just one suggestion.

Thanks

mjd507:
Looks like a flaky failure on the Kafka batch inbound assertion.  I try to amend this commit and force push.

> Task :spring-integration-kafka:test
> 
> MessageDrivenAdapterTests > testInboundBatch(EmbeddedKafkaBroker) FAILED
    java.lang.AssertionError at MessageDrivenAdapterTests.java:488

artembilan:
Thank you for the update, @mjd507 !

that Kafka batch test is indeed unstable .
I just cannot figure out properties combination to get exactly two records. Or we need to change test logic to verify until both records are there, or just drop such a test off 😢

mjd507:
> that Kafka batch test is indeed unstable . I just cannot figure out properties combination to get exactly two records. Or we need to change test logic to verify until both records are there, or just drop such a test off 😢

ok, I'll create an issue for that.

PR #10942, opened 2026-04-14
test: fix unstable batch consumer in `MessageDrivenAdapterTests`.

mjd507:
By starting the adapter after sending messages, ensure a single poll retrieves the full batch messages.

### DIFF
--- a/spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
+++ b/spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
@@ -110,6 +110,7 @@ import static org.mockito.Mockito.verify;
  * @author Cameron Mayfield
  * @author Urs Keller
  * @author Jooyoung Pyoung
+ * @author Jiandong Ma
  *
  * @since 5.4
  *
@@ -432,8 +433,6 @@ class MessageDrivenAdapterTests implements TestApplicationContextAware {
 			}
 
 		});
-		adapter.start();
-		ContainerTestUtils.waitForAssignment(container, 1);
 
 		Map<String, Object> senderProps = KafkaTestUtils.producerProps(embeddedKafka);
 		senderProps.put(ProducerConfig.LINGER_MS_CONFIG, 0);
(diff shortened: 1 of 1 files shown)


spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
```diff
@@ -443,6 +442,9 @@ public Message<?> toMessage(List<ConsumerRecord<?, ?>> records, Acknowledgment a
 		template.sendDefault(0, 1487694048607L, 1, "foo");
 		template.sendDefault(0, 1487694048608L, 1, "bar");
 
+		adapter.start();
+		ContainerTestUtils.waitForAssignment(container, 1);
```
artembilan, line 446: Thanks for looking into this @mjd507 !

Well, I don't think it helps to make sure both records are in the topic because those `max.record` and so Consumer properties should do the trick somehow from Kafka standpoint, but I'm OK accepting your contribution since that is really out of our project scope to verify Kafka behavior.

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
```diff
@@ -480,12 +487,17 @@ public Message<?> toMessage(List<ConsumerRecord<?, ?>> records, Acknowledgment a
...
+
 		Message<?> error = errors.receive(30000);
 		assertThat(error).isNotNull();
 		assertThat(error.getPayload()).isInstanceOf(ConversionException.class);
 		assertThat(((ConversionException) error.getPayload()).getMessage())
 				.contains("Failed to convert to message");
 		assertThat(((ConversionException) error.getPayload()).getRecords()).hasSize(2);
+		assertThat(restartAssignmentLatch.await(10, TimeUnit.SECONDS)).isTrue();
```
artembilan, line 500: This is somehow out of order of the whole test suite expectations.
I mean we probably should verify assignment first before checking for consumed record.
At the same time I don't understand why do we need to check this at all.
Looks like your main point is to ensure consumer is stopped until after we send records to the topic.

artembilan (changes requested):

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/inbound/MessageDrivenAdapterTests.java
mjd507: yes, I have same doubts when writing this, I saw the first existing latch also do the verify after all the assertions passed. 

I also ask myself do we need the second latch, and I think better have that, the batch mode test has two part, one is happy path , the other is error path , I think we can treat both equally? (keep second latch?)

artembilan: Sorry, I missed the part that we already checking for assignment for the first start.
Yes, it feels like we definitely do need it for the second one.

Thanks.
Merging!..

artembilan (approved):

PR #10961, opened 2026-04-20
Resolve nullability errors after ver upgrade

cppwfs:
- Iterator in JdbcMessageStore was ingored because SingleColumnRowMapper.mapRow() is @Nullable but GROUP_KEY is not null


artembilan:
Superseded with: https://github.com/spring-projects/spring-integration/commit/e4348f5510ee6930bdc27a4a9c30b0ff7c2f45fa

PR #10965, opened 2026-04-21
Update major versions 3rd party dependencies

cppwfs:


gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ assertjVersion = "3.27.7"
 assertkVersion = "0.28.1"
 avroVersion = "1.12.1"
 awaitilityVersion = "4.3.0"
-camelVersion = "4.18.1"
+camelVersion = "4.19.0"
```
artembilan, line 9: Thank you for taking care about this!

No need to do anything right now, but I wonder if we should consider for the next our minor to have Dependabot config for `main` to be relaxed from:
```
    ignore:
      - dependency-name: '*'
        update-types:
          - version-update:semver-major
          - version-update:semver-minor
```
This way during our M and RC phases we would have all the deps updated not only to patches but also their latest major/minor.
Of course, we should keep in mind that config and bring it back to just patches when we move our patches support line.
WDYT?

cppwfs: I completely agree.   I'll add that to our checklist.

PR #10978, opened 2026-04-22
Add message_store pattern in `IntegrationPatternType`

mjd507:
Make `MessageStore` extend `IntegrationPattern` with default message_store type.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/IntegrationPatternType.java
@@ -25,6 +25,7 @@ import java.util.stream.Collectors;
  * Used to indicate which pattern a target component implements.
  *
  * @author Artem Bilan
+ * @author Jiandong Ma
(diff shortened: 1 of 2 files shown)


artembilan (approved):
I think I don't see any problem with this.
We have introduced an `IntegrationPattern` contract to satisfy Integration Graph logic: https://docs.spring.io/spring-integration/reference/graph.html#page-title.
The `MessageStore` is not part of the graph for the flow representation.
So, that is probably why it was missed as part of this contact.
But what you have done really fits into the whole EIP picture.

Thank you! 👍 

WDYT, @cppwfs ?

pavankumarreddyboppidi (approved):

cppwfs:
I know that this is a day late and a dollar short.
I agree with the premise and think it is a solid suggestion.   The only side effect, based on my understanding, this will cause some existing systems to have the label appear in their Monitoring systems (prometheus/datadog), which may cause some confusion, but this is an edge case.

PR #10980, opened 2026-04-24
GH-10847: Fix potential infinite loop in `JdbcMetadataStore.putIfAbsent()`

### LINKED ISSUE #10847: JdbcMetadataStore putIfAbsent may stuck in while loop on MySQL database.

6.5.7

**Describe the bug**

I observed that JdbcMetadataStore stuck in while loop on my production environment(2 service instances) in the following method:

https://github.com/spring-projects/spring-integration/blob/main/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java#L227

Issue may happen where two processes(or threads) will access putIfAbsent method at the same time and one  of them was able to insert record to database.

Issue is related to the fact that MySQL database is using REPEATABLE READ isolation level and if one process(or thread) successfully insert record to database the other will not be able to read it thus it will stuck in while loop forever.

**To Reproduce**

Run two services with MySQL database and default transaction isolation level.

**Expected behavior**

Shouldn't work in distributed environment and MySQL database.

**Sample**

I was able to simulate issue with tests containers and unit tests but I am not able to copy this code here due to strict security policy in company. Pseudo code to reproduce problem:

for(int idx = 0; i<100; i++) {
  exec = Executors.newFixedThreadPool(2)
  futures = new ArrayList(2)
  futures.add(CompletableFuture.supplyAsync(() -> metadataStore.putIfAbsent("testKey" + idx, "testValue")),  exec)
  futures.add(CompletableFuture.supplyAsync(() -> metadataStore.putIfAbsent("testKey" + idx, "testValue")), exec)
  CompletableFuture.allOff(futures).join();
  exec.shutdown();
  exec.awaitTermination(1, TimeUnit.SECONDS);
}

mjd507:
Use two separate programmatic transactions for `putIfAbsent()`, one for insert, one for select.

Try Fixes: https://github.com/spring-projects/spring-integration/issues/10847

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
@@ -22,7 +22,12 @@ import javax.sql.DataSource;
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.SmartLifecycle;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.dao.DataIntegrityViolationException;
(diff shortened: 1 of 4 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
```diff
@@ -279,7 +337,6 @@ public boolean replace(String key, String oldValue, String newValue) {
 	}
 
 	@Override
-	@Transactional
```
mjd507, line 282: remove by accident, this put method may have same issue.

artembilan: I don't think the removal of these annotations requiring external transaction to be active is correct.
The main point of this metadata store is to be a part of broader transactional unit of work.
And if that fails, we just don't store anything in DB.
That's why I recommended to revise transactional setting on the channel adapter where this store is used.

artembilan (changes requested):

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
mjd507: oh, I didn't take broader transactional into consideration. 
closing this PR.

PR #11001, opened 2026-05-11
GH-10990: Fix MQTT v5 inbound payload conversion

### LINKED ISSUE #10990: Mqttv5PahoMessageDrivenChannelAdapter ignores payloadType due to incorrect MessageConverter.toMessage usage

7.0.4

## Describe the bug

I ran into an issue with `Mqttv5PahoMessageDrivenChannelAdapter` when receiving messages. When `payloadType` is set to a class other than `byte[]` or `MqttMessage` (e.g., `setPayloadType(String.class)`), it causes two weird behaviors:

### Symptom 1: `setPayloadType(String.class)` is silently ignored

If I configure `setPayloadType(String.class)`, the `Message` payload is still `byte[]`. I receive no error and no warning; it just silently ignores the type hint I set.

### Symptom 2: payload is silently Base64-encoded when `contentType=application/json`

When the inbound MQTT v5 message has `PayloadFormatIndicator=1` *and* `Content-Type=application/json`, the payload doesn't come out as raw bytes or parsed JSON. Instead, it gets **Base64-encoded** and wrapped in JSON quotes.

For example, if the broker sends payload `{"temperature":25}`, my application receives `"eyJ0ZW1wZXJhdHVyZSI6MjV9"` (wrapped in JSON quotes, emitted as bytes). This is pretty confusing because it looks like a normal string at first glance, so you don't realize something is broken until your downstream JSON parser fails on the Base64 text.

### In Spring Boot

```java
Mqttv5PahoMessageDrivenChannelAdapter adapter = new Mqttv5PahoMessageDrivenChannelAdapter(clientManager, "device");
adapter.setPayloadType(String.class);

return IntegrationFlow
    .from(adapter)
    .log(LoggingHandler.Level.WARN, "MQTT-String-Log", m -> "String after conversion: " + m.getPayload() + "\nHeader: " + m.getHeaders())
    .get();
```

Receive a standard text message from the MQTT broker. Observe the console log: `message.getPayload().getClass()` returns `[B` (byte array) rather than `java.lang.String`.

### In JUnit

```java
@SpringJUnitConfig
public class MqttBugReproducerTest {
    @Autowired
    private ApplicationContext applicationContext;

    @Test
    public void testPayloadConversion() {
        // Construct a raw MQTT v5 message
        String originalJson = "{\"temperature\":25}";
        MqttMessage mqttMessage = new MqttMessage(originalJson.getBytes(StandardCharsets.UTF_8));

        MqttProperties properties = new MqttProperties();
        properties.setContentType("application/json");
        properties.setPayloadFormat(true); // PayloadFormatIndicator=1
        mqttMessage.setProperties(properties);

        //  Initialize the Adapter (bypass actual broker connection)
        Mqttv5PahoMessageDrivenChannelAdapter adapter =
                new Mqttv5PahoMessageDrivenChannelAdapter(
                        "tcp://localhost:1883",
                        "testClient",
                        "test/topic");
		
		// Trigger the bug
        adapter.setPayloadType(String.class);

        QueueChannel outputChannel = new QueueChannel();
        adapter.setOutputChannel(outputChannel);
        adapter.setBeanFactory(applicationContext);
        adapter.afterPropertiesSet();

        // Trigger messageArrived to simulate message inbound
        adapter.messageArrived("test/topic", mqttMessage);

        // Retrieve result and inspect the bug symptom
        Message<?> resultMessage = outputChannel.receive(1000);
        assertNotNull(resultMessage, "No message received on outputChannel");

        Object payload = resultMessage.getPayload();
        System.out.println("Expected Payload Type: String.class");
        System.out.println("Actual Payload Type: " + payload.getClass().getName());
        System.out.println("Expected Payload Content: " + originalJson);
        if (payload instanceof byte[] bytes) {
            System.out.println("Actual Payload Content (byte[] -> UTF8): " + new String(bytes, StandardCharsets.UTF_8));
        } else {
            System.out.println("Actual Payload Content: " + payload);
        }
    }

    @Configuration
    static class TestConfig {
    	/**
    	 * Mock SmartMessageConverter chain,
         * including JacksonJsonMessageConverter 
         * to reproduce Base64 encoding symptoms.
         */
        @Bean(name = "integrationArgumentResolverMessageConverter")
        public SmartMessageConverter integrationArgumentResolverMessageConverter() {
            return new CompositeMessageConverter(List.of(
                    new ByteArrayMessageConverter(),
                    new StringMessageConverter(),
                    new JacksonJsonMessageConverter()	// Jackson 3
            ));
        }
    }
}

```

## Cause Analysis

Both symptoms originate in:

`Mqttv5PahoMessageDrivenChannelAdapter#messageArrived`  [(lines ~444–455)](https://github.com/spring-projects/spring-integration/blob/main/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java#L444):

  ```java
Object payload =
        MqttMessage.class.isAssignableFrom(this.payloadType)
                ? mqttMessage
                : mqttMessage.getPayload();

Message<?> message;
if (MqttMessage.class.isAssignableFrom(this.payloadType) || byte[].class.isAssignableFrom(this.payloadType)) {
    message = new GenericMessage<>(payload, headers);
}
else {
    message = this.messageConverter.toMessage(payload, new MessageHeaders(headers), this.payloadType);
}
  ```

  When `this.payloadType` is set to `String.class`, the code executes:

```
this.messageConverter.toMessage( <byte[] instance>, headers, String.class )
```

I think the issue is a mismatch in API semantics. In the Spring Messaging contract:

- `toMessage` is for **outbound serialization**: It converts a Java object into a serialized format (like byte[]).
- `fromMessage` is for **inbound deserialization**: It decodes raw network byte[] into a typed Java object.

Since `Mqttv5PahoMessageDrivenChannelAdapter` is an **inbound** component handling raw `byte[]` from the broker, using `toMessage` here reverses the expected workflow. The default converter chain sees the `byte[]` input, assumes serialization is already done, and simply envelopes it into a `Message<byte[]>`, completely ignoring the `String.class` hint. What's worse, if the  converter sees the `contentType=application/json`, it assumes the `byte[]` input is binary data (e.g., image, audio), then encodes the data in **base64** and wraps it in JSON quotes.

## Possible Solutions

Here are a few ways we could fix this:

### 1. Replace `toMessage` with `fromMessage` 

This cleanly fixes the problem:

```java
// Conceptual approach:
Message<?> rawMessage = new GenericMessage<>(payload, headers);
Object decodedPayload = this.messageConverter.fromMessage(rawMessage, this.payloadType);
message = new GenericMessage<>(decodedPayload, headers);
```

But it is a **behavior change**. It could affect users who rely on the current `byte[]` output (e.g., users who call `setPayloadType(String.class)` but consume `byte[]` downstream).

### 2. Add an opt-in flag

Add a new setter, e.g., `setDecodePayload(boolean)`, defaulting to false. When true, it uses `fromMessage`. This preserves backward compatibility.

### 3. Restrict `setPayloadType` and update docs

Simply document the legacy behavior and restrict `setPayloadType` to only accept `byte[]`/`MqttMessage`.

I am happy to draft a PR if needed.

goutamadwant:
Fixes #10990.

`Mqttv5PahoMessageDrivenChannelAdapter` used the configured `SmartMessageConverter` in the wrong direction when `payloadType` was set. It now builds an inbound message from the MQTT payload and delegates conversion through `fromMessage(...)`, so the requested target payload type is applied consistently.

Custom MQTT v5 converters that depended on the previous reversed call need to provide inbound payload conversion through `fromMessage(...)`; the 7.1 MQTT changes section documents that behavior change.

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -65,7 +65,6 @@ import org.springframework.messaging.MessagingException;
 import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.converter.SmartMessageConverter;
 import org.springframework.util.Assert;
-import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
 
@@ -462,7 +461,7 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 							.copyHeaders(headers)
 							.build();
 			Object convertedPayload = this.messageConverter.fromMessage(messageToConvert, this.payloadType);
-			if (convertedPayload == null || !ClassUtils.isAssignableValue(this.payloadType, convertedPayload)) {
+			if (convertedPayload == null) {
 				throw new MessageConversionException(messageToConvert, "Failed to convert from MQTT Message");
 			}
 			message =
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -59,12 +59,13 @@ import org.springframework.integration.mqtt.support.MqttHeaderMapper;
 import org.springframework.integration.mqtt.support.MqttHeaders;
 import org.springframework.integration.mqtt.support.MqttMessageConverter;
 import org.springframework.integration.mqtt.support.MqttUtils;
+import org.springframework.integration.support.MutableMessageBuilder;
 import org.springframework.messaging.Message;
-import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.MessagingException;
+import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.converter.SmartMessageConverter;
-import org.springframework.messaging.support.GenericMessage;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
 
@@ -87,6 +88,7 @@ import org.springframework.util.ObjectUtils;
  * @author Artem Vozhdayenko
  * @author Matthias Thoma
  * @author Glenn Renfro
+ * @author Goutam Adwant
  *
  * @since 5.5.5
  *
@@ -448,17 +450,44 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 
 		Message<?> message;
 		if (MqttMessage.class.isAssignableFrom(this.payloadType) || byte[].class.isAssignableFrom(this.payloadType)) {
-			message = new GenericMessage<>(payload, headers);
+			message =
+					getMessageBuilderFactory()
+							.withPayload(payload)
+							.copyHeaders(headers)
+							.build();
 		}
 		else {
-			message = this.messageConverter.toMessage(payload, new MessageHeaders(headers), this.payloadType);
+			Message<?> messageToConvert =
+					MutableMessageBuilder.withPayload(payload, false)
+							.copyHeaders(headers)
+							.build();
+			Object convertedPayload = this.messageConverter.fromMessage(messageToConvert, this.payloadType);
+			if (convertedPayload != null && ClassUtils.isAssignableValue(this.payloadType, convertedPayload)) {
+				message =
+						getMessageBuilderFactory()
+								.withPayload(convertedPayload)
+								.copyHeaders(messageToConvert.getHeaders())
+								.build();
+			}
+			else {
+				message = this.messageConverter.toMessage(payload, messageToConvert.getHeaders(), this.payloadType);
+			}
+			if (message == null || !ClassUtils.isAssignableValue(this.payloadType, message.getPayload())) {
+				throw new MessageConversionException(messageToConvert, "Failed to convert from MQTT Message");
+			}
+			message =
+					getMessageBuilderFactory()
+							.fromMessage(message)
+							.copyHeadersIfAbsent(messageToConvert.getHeaders())
+							.build();
 		}
 
+		Message<?> messageToSend = message;
 		try {
-			sendMessage(message);
+			sendMessage(messageToSend);
 		}
 		catch (RuntimeException ex) {
-			logger.error(ex, () -> "Unhandled exception for " + message);
+			logger.error(ex, () -> "Unhandled exception for " + messageToSend);
 			throw ex;
 		}
 	}
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -471,12 +471,11 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 							.build();
 		}
 
-		Message<?> messageToSend = message;
 		try {
-			sendMessage(messageToSend);
+			sendMessage(message);
 		}
 		catch (RuntimeException ex) {
-			logger.error(ex, () -> "Unhandled exception for " + messageToSend);
+			logger.error(ex, () -> "Unhandled exception for " + message);
 			throw ex;
 		}
 	}
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
@@ -462,23 +462,13 @@ public class Mqttv5PahoMessageDrivenChannelAdapter
 							.copyHeaders(headers)
 							.build();
 			Object convertedPayload = this.messageConverter.fromMessage(messageToConvert, this.payloadType);
-			if (convertedPayload != null && ClassUtils.isAssignableValue(this.payloadType, convertedPayload)) {
-				message =
-						getMessageBuilderFactory()
-								.withPayload(convertedPayload)
-								.copyHeaders(messageToConvert.getHeaders())
-								.build();
-			}
-			else {
-				message = this.messageConverter.toMessage(payload, messageToConvert.getHeaders(), this.payloadType);
-			}
-			if (message == null || !ClassUtils.isAssignableValue(this.payloadType, message.getPayload())) {
+			if (convertedPayload == null || !ClassUtils.isAssignableValue(this.payloadType, convertedPayload)) {
 				throw new MessageConversionException(messageToConvert, "Failed to convert from MQTT Message");
 			}
 			message =
 					getMessageBuilderFactory()
-							.fromMessage(message)
-							.copyHeadersIfAbsent(messageToConvert.getHeaders())
+							.withPayload(convertedPayload)
+							.copyHeaders(messageToConvert.getHeaders())
 							.build();
 		}
 
--- a/spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
+++ b/spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.mqtt;
 
+import java.util.List;
 import java.util.Map;
 
 import org.eclipse.paho.mqttv5.client.IMqttAsyncClient;
@@ -23,14 +24,22 @@ import org.eclipse.paho.mqttv5.client.IMqttMessageListener;
 import org.eclipse.paho.mqttv5.client.IMqttToken;
 import org.eclipse.paho.mqttv5.client.MqttConnectionOptions;
 import org.eclipse.paho.mqttv5.common.MqttException;
+import org.eclipse.paho.mqttv5.common.MqttMessage;
 import org.eclipse.paho.mqttv5.common.MqttSubscription;
+import org.eclipse.paho.mqttv5.common.packet.MqttProperties;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.integration.channel.NullChannel;
+import org.springframework.integration.channel.QueueChannel;
 import org.springframework.integration.mqtt.inbound.Mqttv5PahoMessageDrivenChannelAdapter;
 import org.springframework.integration.test.util.TestUtils;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.converter.ByteArrayMessageConverter;
+import org.springframework.messaging.converter.CompositeMessageConverter;
+import org.springframework.messaging.converter.StringMessageConverter;
 import org.springframework.test.util.ReflectionTestUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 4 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -451,7 +451,12 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
 			message = new GenericMessage<>(payload, headers);
 		}
 		else {
-			message = this.messageConverter.toMessage(payload, new MessageHeaders(headers), this.payloadType);
+			Message<?> messageToConvert = new GenericMessage<>(payload, headers);
```
artembilan: All good, but I suggest to revise the logic in favor of `MutableMessageBuilder.withPayload(payload, false)`.
Then use `MutableMessage.getHeaders()` for the final `new GenericMessage`.
Well, even `new GenericMessage` must not be there in the end.
See `getMessageBuilderFactory()` of this channel adapter.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
```diff
@@ -98,9 +108,36 @@ public void sharedSubscriptionsPopulation() throws Exception {
 				.containsEntry("$SYS/broker/#", "$SYS/broker/#");
 	}
 
+	@Test // GH-10990
```
artembilan: Please, don't use issue link.
It is easy to lose its context when some refactoring or other changes are done to the class.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
```diff
@@ -98,9 +108,36 @@ public void sharedSubscriptionsPopulation() throws Exception {
...
+	public void payloadTypeIsUsedForInboundMqttV5PayloadConversion() throws Exception {
```
artembilan: I don't think `MqttV5` in the method name brings too much value.
Looks like the whole class does that already.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
```diff
@@ -98,9 +108,36 @@ public void sharedSubscriptionsPopulation() throws Exception {
...
+		IMqttAsyncClient client = mock();
+		QueueChannel outputChannel = new QueueChannel();
+		Mqttv5PahoMessageDrivenChannelAdapter adapter = buildAdapterIn(client, false, outputChannel);
+		adapter.setPayloadType(String.class);
+		adapter.setMessageConverter(new CompositeMessageConverter(List.of(
+				new ByteArrayMessageConverter(),
+				new StringMessageConverter())));
+
+		MqttMessage mqttMessage = new MqttMessage("bar".getBytes(StandardCharsets.UTF_8));
```
artembilan: Please, don't use non-professional `foo/bar` language in our code.
And we don't need `StandardCharsets.UTF_8` for our tests.

artembilan (changes requested):
Very promising proposal.
Please, add your name to the `@author` list of all the effected classes.

Thanks

artembilan (changes requested):
Well, the PR build demonstrates that you have not run testing locally before pushing to PR:
```

Mqttv5BackToBackTests > testSimpleMqttv5Interaction() FAILED
    org.opentest4j.AssertionFailedError at Mqttv5BackToBackTests.java:115

Mqttv5BackToBackTests > testSharedTopicMqttv5Interaction() FAILED
    org.opentest4j.AssertionFailedError at Mqttv5BackToBackTests.java:160
```

Please, revise the change.

Thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
goutamadwant: Thanks for your review @artembilan 

updated this path to use `MutableMessageBuilder.withPayload(payload, false)` for the conversion message and `getMessageBuilderFactory()` for the final message creation. I also removed the direct `GenericMessage` construction and kept the existing custom converter path intact.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
goutamadwant: Sure, removed the issue reference from the test annotation.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
goutamadwant: Agreed, renamed the test to avoid repeating MQTT v5 in this class.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5AdapterTests.java
goutamadwant: Replaced the test values with clearer names and removed the unnecessary `StandardCharsets.UTF_8` usage.

goutamadwant:
Thanks for pointing this out. @artembilan 

I had verified the focused adapter test locally, but I missed that this change also impacts the existing MQTT v5 back-to-back flow. I have revised the conversion path now so that the existing custom converter behavior is preserved, while still applying the configured `payloadType`.

The two failing tests are passing locally now.

Verified with:

- `./gradlew :spring-integration-mqtt:test --tests org.springframework.integration.mqtt.Mqttv5BackToBackTests.testSimpleMqttv5Interaction --tests org.springframework.integration.mqtt.Mqttv5BackToBackTests.testSharedTopicMqttv5Interaction`
- `./gradlew :spring-integration-mqtt:test --tests org.springframework.integration.mqtt.Mqttv5AdapterTests --tests org.springframework.integration.mqtt.Mqttv5BackToBackTests`
- `./gradlew :spring-integration-mqtt:check`

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -448,17 +450,44 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
...
+			Object convertedPayload = this.messageConverter.fromMessage(messageToConvert, this.payloadType);
+			if (convertedPayload != null && ClassUtils.isAssignableValue(this.payloadType, convertedPayload)) {
+				message =
+						getMessageBuilderFactory()
+								.withPayload(convertedPayload)
+								.copyHeaders(messageToConvert.getHeaders())
+								.build();
+			}
+			else {
+				message = this.messageConverter.toMessage(payload, messageToConvert.getHeaders(), this.payloadType);
```
artembilan: This is not correct.
If we change a behavior, we do that everywhere, for all the possible scenarios.
That's why I'm aiming this fix only for the current minor.
So, please, consider to revert this new logic, and do everything according to the claim for this issue.
If some tests are failing then they have to be fixed.

artembilan (changes requested):
See my comments in the: https://github.com/spring-projects/spring-integration/pull/11012.
1. The change will be applied only to the current `7.1.0` due to breaking change nature.
2. Please, revise all the new public API for a proper `@since 7.1.0`
3. Don't squash commits to let us easier to follow changes via history. 
4. Learn how properly give a commit message: https://cbea.ms/git-commit/
5. Document this change in the `whats-new.adoc` explaining that is is a breaking change but reasonable. Right, both this fix and that for outbound channel adapter will be squashed to a single doc section, but that will be discussed when we understand which one will be merged first.
6. I told in some other place: the `Tests` section in PR description is over-headed and really redundant. First: the whole module testing covers individual tests in that module. Second: according to PR build failures looks like you don't run them locally anyway. If that is generated by some AI tool, please, consider to verify the text and purpose before giving it us for review. It also feels your you, guys, from the same company since both PRs are similar in the style and both claims redundant commands for Gradle.

Thanks

artembilan:
Hi, @adwantg !

The outbound channel adapter fix on the matter has been merged: https://github.com/spring-projects/spring-integration/pull/11012.
So, now you should rebase your branch to the latest `main` and add respective change into a new `[[x7.1-mqtt-changes]]` section of the `whats-new.adoc`.

Or... just let us know and we finish the work for you!
The point is that release is due next week, so would be great to have this in to avoid divergence in the behavior.

Thanks

goutamadwant:
will work on it today @artembilan

goutamadwant:
Thanks @artembilan. I rebased on latest main, aligned the inbound conversion with fromMessage(..), and added the inbound note to the existing MQTT changes section.

I also simplified the PR description and kept the follow-up commit separate.

Locally verified the MQTT module check and Antora docs build after the update.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -448,17 +450,34 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
...
+							.build();
 		}
 		else {
-			message = this.messageConverter.toMessage(payload, new MessageHeaders(headers), this.payloadType);
+			Message<?> messageToConvert =
+					MutableMessageBuilder.withPayload(payload, false)
+							.copyHeaders(headers)
+							.build();
+			Object convertedPayload = this.messageConverter.fromMessage(messageToConvert, this.payloadType);
+			if (convertedPayload == null || !ClassUtils.isAssignableValue(this.payloadType, convertedPayload)) {
```
artembilan: I don't think we need to check for the result type.
It is really not enforced by that `MessageConverter` contract:
```
@Nullable Object fromMessage(Message<?> message, Class<?> targetClass);
```
If it would, we would have a generic for that expected type.
So, we produce from here whatever is returned by the converter if that is not `null`.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackTests.java
```diff
@@ -172,7 +172,7 @@ void mqttEvents(MqttIntegrationEvent event) {
 		}
 
 		@Bean
-		public SmartMessageConverter mqttStringToBytesConverter() {
+		public SmartMessageConverter mqttBytesToStringConverter() {
```
artembilan, line 176: Your name to the `@author` of this changed class.

artembilan (changes requested):
Very good!
Just couple nit-picks.

Thanks

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5BackToBackTests.java
goutamadwant: Addressed

goutamadwant:
Thanks @artembilan , addressed both comments.

I removed the result type check and added my name to the changed test class author list.
Locally verified the MQTT module check and the full root test task after the update.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
```diff
@@ -448,17 +449,34 @@ public void messageArrived(String topic, MqttMessage mqttMessage) {
...
+					getMessageBuilderFactory()
+							.withPayload(convertedPayload)
+							.copyHeaders(messageToConvert.getHeaders())
+							.build();
 		}
 
+		Message<?> messageToSend = message;
 		try {
-			sendMessage(message);
+			sendMessage(messageToSend);
```
cppwfs: `messageToSend` is redundant.  We can just use message.

cppwfs (changes requested):
Thank you for the contribution.   One more nitpick :-)

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/inbound/Mqttv5PahoMessageDrivenChannelAdapter.java
goutamadwant: Thanks for your review @cppwfs fixed it too

PR #11012, opened 2026-05-20
GH-11011: Fix MQTT v5 outbound payload conversion

### LINKED ISSUE #11011: Mqttv5PahoMessageHandler fails due to incorrect MessageConverter.fromMessage usage
### In what version(s) of Spring Integration are you seeing this issue?

7.0.4

## Describe the bug

This is the outbound mirror of [Issue #10990](https://github.com/spring-projects/spring-integration/issues/10990).

`Mqttv5PahoMessageHandler#buildMqttMessage` calls `MessageConverter.fromMessage(message, byte[].class)` at 
[Mqttv5PahoMessageHandler.java:225](https://github.com/spring-projects/spring-integration/blob/main/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java#L225) to serialize a non-byte[] / non-String payload into bytes. According to the Spring Messaging contract, `fromMessage` is for **inbound** deserialization (bytes → object) while `toMessage` is for **outbound** serialization (object → bytes). The outbound publish path is using the wrong direction.

## To Reproduce

Typical outbound gateway setup with `contentType=application/json`:

```java
@MessagingGateway(defaultRequestChannel = "mqttOutboundChannel",
        defaultHeaders = @GatewayHeader(name = MessageHeaders.CONTENT_TYPE,
                value = MediaType.APPLICATION_JSON_VALUE))
public interface MqttGateway {
    void send(@Header(MqttHeaders.TOPIC) String topic, Object payload);
}

@Bean
IntegrationFlow mqttOutFlow(Mqttv5PahoMessageHandler messageHandler) {
    return IntegrationFlow.from("mqttOutboundChannel")
        .handle(messageHandler)
        .get();
}
```

Calling gateway.send("topic", new TestPojo(...)) triggers the issue, resulting in the following exception:

JsonParseException: Unrecognized token '': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')

The submitted PR includes unit tests in Mqttv5HandlerTests that reproduce the bug and verify the fix.

## Expected behavior

The handler should invoke converter.toMessage(payload, headers). With the default Jackson-backed converter and contentType=application/json, this produces valid JSON bytes.

d0v0c:
Fixes [#11011](https://github.com/spring-projects/spring-integration/issues/11011)

`Mqttv5PahoMessageHandler#buildMqttMessage` was calling `MessageConverter#fromMessage(message, byte[].class)` to serialize non-byte[] / non-String payloads,  the wrong direction per the Spring Messaging contract. Switch to `toMessage(payload, headers)`. 

The class Javadoc gains a short note about the default converter's JSON content-type requirement.

## Test plan

New `Mqttv5HandlerTests` covers two scenarios:

- `publishPojoWithDefaultConverter`: round-trips a POJO through `ConfigurableCompositeMessageConverter` and verifies the published bytes are valid JSON.
- `publishPojoWithCustomConverter`: verifies a user-supplied converter's `toMessage` is invoked on the outbound path.
- `./gradlew :spring-integration-mqtt:test` passes locally with no regressions.

### DIFF
--- a/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
+++ b/spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
@@ -49,11 +49,14 @@ import org.springframework.util.Assert;
 
 /**
  * The {@link AbstractMqttMessageHandler} implementation for MQTT v5.
+ * <p>
+ * Payloads other than {@code byte[]} or {@link String} use the configured {@link MessageConverter}.
  *
  * @author Artem Bilan
  * @author Lucas Bowler
  * @author Artem Vozhdayenko
  * @author Glenn Renfro
+ * @author Deng Pan
  *
  * @since 5.5.5
  */
@@ -214,17 +217,28 @@ public class Mqttv5PahoMessageHandler extends AbstractMqttMessageHandler<IMqttAs
 	private MqttMessage buildMqttMessage(Message<?> message) {
 		Object payload = message.getPayload();
 		byte[] body;
-		if (payload instanceof byte[]) {
-			body = (byte[]) payload;
+		if (payload instanceof byte[] bytes) {
+			body = bytes;
 		}
-		else if (payload instanceof String) {
-			body = ((String) payload).getBytes(StandardCharsets.UTF_8);
+		else if (payload instanceof String string) {
+			body = string.getBytes(StandardCharsets.UTF_8);
 		}
 		else {
 			MessageConverter converter = getConverter();
-			body = (byte[]) converter.fromMessage(message, byte[].class);
-			Assert.state(body != null,
+			Message<?> converted = converter.toMessage(payload, message.getHeaders());
+			Assert.state(converted != null,
 					() -> "The MQTT payload cannot be null. The '" + converter + "' returned null for: " + message);
+			Object convertedPayload = converted.getPayload();
+			if (convertedPayload instanceof byte[] bytes) {
+				body = bytes;
+			}
+			else if (convertedPayload instanceof String string) {
+				body = string.getBytes(StandardCharsets.UTF_8);
+			}
+			else {
+				throw new IllegalStateException("The MQTT payload must be a 'byte[]' or 'String'. The '" + converter +
+						"' returned a '" + convertedPayload.getClass().getSimpleName() + "' for: " + message);
+			}
 		}
 
 		MqttMessage mqttMessage = new MqttMessage();
--- a/spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
+++ b/spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.mqtt;
+
+import org.eclipse.paho.mqttv5.client.IMqttAsyncClient;
+import org.eclipse.paho.mqttv5.client.IMqttToken;
+import org.eclipse.paho.mqttv5.client.MqttConnectionOptions;
+import org.eclipse.paho.mqttv5.common.MqttMessage;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import tools.jackson.databind.ObjectMapper;
+
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.integration.mqtt.core.ClientManager;
+import org.springframework.integration.mqtt.outbound.Mqttv5PahoMessageHandler;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.integration.support.converter.ConfigurableCompositeMessageConverter;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.support.GenericMessage;
+import org.springframework.util.MimeTypeUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Unit tests for {@link Mqttv5PahoMessageHandler}, focusing on the outbound payload
+ * conversion path.
+ *
+ * @author Deng Pan
+ *
+ * @since 6.5.9
+ */
+class Mqttv5HandlerTests {
+
+	@Test
+	void publishPojoWithDefaultConverter() throws Exception {
+		IMqttAsyncClient mockClient = mock();
+		IMqttToken token = mock();
+		given(mockClient.isConnected()).willReturn(true);
+		given(mockClient.publish(anyString(), any(MqttMessage.class), any(), any())).willReturn(token);
+
+		Mqttv5PahoMessageHandler handler = createHandler(mockClient, new ConfigurableCompositeMessageConverter());
+
+		TestPojo pojo = new TestPojo("test", 1);
+		handler.handleMessage(MessageBuilder.withPayload(pojo)
+				.setHeader(MessageHeaders.CONTENT_TYPE, MimeTypeUtils.APPLICATION_JSON_VALUE)
+				.build());
+
+		// Capture the MqttMessage actually handed to paho and round-trip its payload bytes
+		// through Jackson to confirm the POJO was serialized to JSON.
+		ArgumentCaptor<MqttMessage> captor = ArgumentCaptor.forClass(MqttMessage.class);
+		verify(mockClient).publish(anyString(), captor.capture(), any(), any());
+		TestPojo roundTripped = new ObjectMapper().readValue(captor.getValue().getPayload(), TestPojo.class);
+		assertThat(roundTripped).isEqualTo(pojo);
+	}
+
+	@Test
+	void publishPojoWithCustomConverter() throws Exception {
+		IMqttAsyncClient mockClient = mock();
+		IMqttToken token = mock();
+		given(mockClient.isConnected()).willReturn(true);
+		given(mockClient.publish(anyString(), any(MqttMessage.class), any(), any())).willReturn(token);
+
+		MessageConverter converter = mock(MessageConverter.class);
+		given(converter.toMessage(any(), any(MessageHeaders.class)))
+				.willAnswer(invocation -> new GenericMessage<>("test"));
+
+		Mqttv5PahoMessageHandler handler = createHandler(mockClient, converter);
+
+		handler.handleMessage(MessageBuilder.withPayload(new TestPojo("test", 2)).build());
+
+		// Custom converter returns a String; verify the handler applies the String -> bytes safeguard.
+		ArgumentCaptor<MqttMessage> captor = ArgumentCaptor.forClass(MqttMessage.class);
+		verify(mockClient).publish(anyString(), captor.capture(), any(), any());
+		assertThat(captor.getValue().getPayload()).isEqualTo("test".getBytes());
+	}
+
+	private static Mqttv5PahoMessageHandler createHandler(IMqttAsyncClient mockClient, MessageConverter converter) {
+		ClientManager<IMqttAsyncClient, MqttConnectionOptions> clientManager = mock();
+		given(clientManager.getUrl()).willReturn("tcp://localhost:1883");
+		given(clientManager.getClientId()).willReturn("testClient");
+		given(clientManager.getConnectionInfo()).willReturn(new MqttConnectionOptions());
+		given(clientManager.getClient()).willReturn(mockClient);
+
+		Mqttv5PahoMessageHandler handler = new Mqttv5PahoMessageHandler(clientManager);
+		handler.setDefaultTopic("test/topic");
+		handler.setConverter(converter);
+		handler.setBeanFactory(mock(BeanFactory.class));
+		handler.afterPropertiesSet();
+		handler.start();
+		return handler;
+	}
+
+	record TestPojo(String name, int value) { }
+}
--- a/spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
+++ b/spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
@@ -47,7 +47,7 @@ import static org.mockito.Mockito.verify;
  *
  * @author Deng Pan
  *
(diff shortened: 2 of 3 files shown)


spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -222,7 +227,8 @@ else if (payload instanceof String) {
 		}
 		else {
 			MessageConverter converter = getConverter();
-			body = (byte[]) converter.fromMessage(message, byte[].class);
+			Message<?> converted = converter.toMessage(payload, message.getHeaders());
+			body = converted != null ? (byte[]) converted.getPayload() : null;
```
artembilan: How are you sure that all the converters are producing `byte[]`?

artembilan (changes requested):
Not critical now, but for the future contributions, please, consider to study this: https://cbea.ms/git-commit/amp/.
The point is that PR doesn’t matter for a long run. Plus in GitHub first commit message becomes a PR description with head line as a title. So, all those stories about test plan would play nice in the commit message.

thanks

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
d0v0c: Thanks for the commit guide—I've added the test plan to the commit message.

As for the cast, the default `ConfigurableCompositeMessageConverter` safely guarantees a `byte[]` output. `JacksonJsonMessageConverter` defaults to `byte[]` payloads, while the other converters reject unmatched types and return `null`.

However, for a custom converter, I agree with your concern. It could return a different type (such as `String`) and trigger a `ClassCastException`. Added an explicit `instanceof byte[]` check.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -49,11 +49,16 @@
 
 /**
  * The {@link AbstractMqttMessageHandler} implementation for MQTT v5.
+ * <p>
+ * Payloads that are not {@code byte[]} or {@link String} are serialized via
+ * the configured {@link MessageConverter}. The default converter requires
+ * a JSON content-type header, e.g. {@code application/json}.
```
artembilan: I don't think this is correct assumption.
The default one is this:
```
		if (getConverter() == null) {
			setConverter(getBeanFactory()
					.getBean(IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME,
							MessageConverter.class));
		}
```
Which is by Spring Integration infrastructure:
```
	private void registerArgumentResolverMessageConverter() {
		if (!this.beanFactory.containsBean(IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME)) {
			this.registry.registerBeanDefinition(IntegrationContextUtils.ARGUMENT_RESOLVER_MESSAGE_CONVERTER_BEAN_NAME,
					new RootBeanDefinition(ConfigurableCompositeMessageConverter.class));
		}
	}
```
And that one is like this:
```
	private static Collection<MessageConverter> initDefaults() {
		List<MessageConverter> converters = new LinkedList<>();

		if (JacksonPresent.isJackson3Present()) {
			JacksonJsonMessageConverter jsonMessageConverter = new JacksonJsonMessageConverter();
			jsonMessageConverter.setStrictContentTypeMatch(true);
			converters.add(jsonMessageConverter);
		}
		else if (JacksonPresent.isJackson2Present()) {
			var mappingJackson2MessageConverter =
					new org.springframework.messaging.converter.MappingJackson2MessageConverter();
			mappingJackson2MessageConverter.setStrictContentTypeMatch(true);
			mappingJackson2MessageConverter.setObjectMapper(
					new org.springframework.integration.support.json.Jackson2JsonObjectMapper().getObjectMapper());
			converters.add(mappingJackson2MessageConverter);
		}
		converters.add(new ByteArrayMessageConverter());
		converters.add(new ObjectStringMessageConverter());

		// TODO do we port it together with MessageConverterUtils ?
		// converters.add(new JavaSerializationMessageConverter());

		return converters;
	}
```

so, if no Jackson process in classpath, there won't be any JSON conversions.
Please, revise this Javadoc to reflect reality.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -222,9 +227,14 @@ else if (payload instanceof String) {
 		}
 		else {
 			MessageConverter converter = getConverter();
-			body = (byte[]) converter.fromMessage(message, byte[].class);
-			Assert.state(body != null,
+			Message<?> converted = converter.toMessage(payload, message.getHeaders());
+			Assert.state(converted != null,
 					() -> "The MQTT payload cannot be null. The '" + converter + "' returned null for: " + message);
+			Object convertedPayload = converted.getPayload();
+			Assert.state(convertedPayload instanceof byte[],
```
artembilan: I think we can perform `body = ((String) payload).getBytes(StandardCharsets.UTF_8);` one more time since since some converters may return a `String`.
Might be the case that this has to be mentioned in Javadoc above, too.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
```diff
@@ -0,0 +1,113 @@
...
+ * Unit tests for {@link Mqttv5PahoMessageHandler}, focusing on the outbound payload
+ * conversion path.
+ *
+ * @author Deng Pan
+ *
+ * @since 7.1
+ */
+public class Mqttv5HandlerTests {
+
+	record TestPojo(String name, int value) { }
```
artembilan: We prefer to put supporting types in the and of test class.
Well, always in the end.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
```diff
@@ -0,0 +1,113 @@
...
+		MqttConnectionOptions opts = new MqttConnectionOptions();
+		opts.setServerURIs(new String[] {"tcp://localhost:1883"});
+
+		Mqttv5PahoMessageHandler handler = new Mqttv5PahoMessageHandler(opts, "testClient");
+		handler.setDefaultTopic("test/topic");
+		handler.setConverter(converter);
+		handler.setBeanFactory(mock(BeanFactory.class));
+		handler.afterPropertiesSet();
+
+		ReflectionTestUtils.setField(handler, "mqttClient", mockClient);
```
artembilan: Cannot we use `ClientManager` abstraction instead of this reflection?

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
```diff
@@ -0,0 +1,113 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Unit tests for {@link Mqttv5PahoMessageHandler}, focusing on the outbound payload
+ * conversion path.
+ *
+ * @author Deng Pan
+ *
+ * @since 7.1
```
artembilan, line 50: I think we are going to back-port this down to `6.5.9`.
Even if it might be some kind of breaking change, it is better to be consistent with an API we rely on as you have pointed out in the issue.

artembilan (changes requested):
Thanks for the update!

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
d0v0c: I've added the String fallback in the else branch as suggested.

Regarding the Javadoc, I'd prefer to treat this fallback as an internal implementation detail rather than explicitly documenting it. My concern is that putting it in the public Javadoc might turn a defensive safeguard into a guaranteed API contract. Keeping it undocumented encourages users to stick to the correct approach (passing byte[]).

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -49,11 +49,16 @@
 
 /**
  * The {@link AbstractMqttMessageHandler} implementation for MQTT v5.
+ * <p>
+ * Payloads other than {@code byte[]} or {@link String} use the configured {@link MessageConverter}.
+ * The default converter only supports JSON (requires Jackson on the classpath and
```
artembilan: We have discussed with you that this is not correct assumption if Jackson is not on classpath.
Please, revise this Javadoc.
I would fully remove this second sentence you suggest.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -222,9 +227,20 @@ else if (payload instanceof String) {
 		}
 		else {
 			MessageConverter converter = getConverter();
-			body = (byte[]) converter.fromMessage(message, byte[].class);
-			Assert.state(body != null,
+			Message<?> converted = converter.toMessage(payload, message.getHeaders());
+			Assert.state(converted != null,
 					() -> "The MQTT payload cannot be null. The '" + converter + "' returned null for: " + message);
+			Object convertedPayload = converted.getPayload();
+			if (convertedPayload instanceof byte[]) {
```
artembilan: We can use pattern expression instead of cast:
```
if (convertedPayload instanceof byte[] bytes) {
    body = bytes;
}
```

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
```diff
@@ -0,0 +1,114 @@
...
+		given(converter.toMessage(any(), any(MessageHeaders.class)))
+				.willAnswer(invocation -> new GenericMessage<>("test".getBytes()));
+
+		Mqttv5PahoMessageHandler handler = createHandler(mockClient, converter);
+
+		handler.handleMessage(MessageBuilder.withPayload(new TestPojo("test", 2)).build());
+
+		// Outbound serialization must go through toMessage(...), never fromMessage(...).
+		verify(converter).toMessage(any(), any(MessageHeaders.class));
+		verify(converter, never()).fromMessage(any(), any());
```
artembilan: I think assertion and that comment above don't make sense.
Please, remove them.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Unit tests for {@link Mqttv5PahoMessageHandler}, focusing on the outbound payload
+ * conversion path.
+ *
+ * @author Deng Pan
+ *
+ * @since 6.5.9
+ */
+public class Mqttv5HandlerTests {
```
artembilan: With JUnit 5 we don't need `public/private` modifier for classes and methods in tests: makes code much easier to read.

artembilan (changes requested):
The build has failed for Checkstyle:
```
Error: eckstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java:29:1: Wrong order for 'org.springframework.integration.mqtt.core.ClientManager' import. [ImportOrder]
> Task :spring-integration-mqtt:checkstyleTest
```
Please, make sure you use `./gradlew :spring-integration-mqtt:check` locally before pushing to the PR.

spring-integration-mqtt/src/main/java/org/springframework/integration/mqtt/outbound/Mqttv5PahoMessageHandler.java
```diff
@@ -214,17 +217,28 @@ private String obtainTopicToPublish(Message<?> message) {
...
 		Object payload = message.getPayload();
 		byte[] body;
-		if (payload instanceof byte[]) {
-			body = (byte[]) payload;
+		if (payload instanceof byte[] bytes) {
+			body = bytes;
 		}
-		else if (payload instanceof String) {
-			body = ((String) payload).getBytes(StandardCharsets.UTF_8);
+		else if (payload instanceof String s) {
```
artembilan: Well, our preference is to never use single character variable names.
Please, revise this into something like `stringValue` if just `string`  doesn't work.

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
```diff
@@ -0,0 +1,114 @@
...
+
+		MessageConverter converter = mock(MessageConverter.class);
+		given(converter.toMessage(any(), any(MessageHeaders.class)))
+				.willAnswer(invocation -> new GenericMessage<>("test"));
+
+		Mqttv5PahoMessageHandler handler = createHandler(mockClient, converter);
+
+		handler.handleMessage(MessageBuilder.withPayload(new TestPojo("test", 2)).build());
+
+		// Custom converter returns a String; verify the handler applies the String → bytes safeguard.
```
artembilan: I'm not sure that your `→` is UTF-8 symbol to be transferred correctly between different OSs.

artembilan (changes requested):
Looking good!
Very close to be merged.

Thanks

spring-integration-mqtt/src/test/java/org/springframework/integration/mqtt/Mqttv5HandlerTests.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Unit tests for {@link Mqttv5PahoMessageHandler}, focusing on the outbound payload
+ * conversion path.
+ *
+ * @author Deng Pan
+ *
+ * @since 6.5.9
```
artembilan: I've changed my mind: even if this is about a bug, it cause a breaking change.
And someone may even argue that this is a bug at all due to the fact that it is documented and works that way for years.
So, I'm accepting the fix this or other way, but this has to be mentioned in the `whats-new.adoc` as a new `[[x7.1-mqtt-changes]]` section.
So, please, revise this back to the `7.1.0`, too.

artembilan (changes requested):

artembilan (changes requested):
BTW, no need to squash commits from your side.
We do that on merge anyway, and that is easier to review PRs: just follow changes history, not the whole change set.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -73,3 +73,8 @@ configurations for the default `JmsTemplate`.
 The `HttpRequestExecutingMessageHandler` (its XML and Java DSL) now can be used with a `RestClient`.
 The `RestTemplate` configuration is deprecated.
 See xref:http.adoc[] for more information.
+
+[[x7.1-mqtt-changes]]
+=== MQTT Changes
+
+For payloads other than `byte[]` or `String`, the `Mqttv5PahoMessageHandler` now delegates the conversion to the configured `MessageConverter` via `toMessage()`, instead of via `fromMessage()`.
```
artembilan, line 80: Great!

Please, also add a link to the target chapter in the end.
Something like:
```
```

artembilan (changes requested):

PR #11015, opened 2026-05-23
GH-11013: Fix ZookeeperMetadataStore replace result

### LINKED ISSUE #11013: `ZookeeperMetadataStore.replace()` breaks the CAS logic
It reports `true` blindly:
```
			try {
				byte[] bytes = this.client.getData().storingStatIn(currentStat).forPath(getPath(key));
				if (oldValue.equals(IntegrationUtils.bytesToString(bytes, this.encoding))) {
					updateNode(key, newValue, currentStat.getVersion());
				}
				return true;
			}
```
So, if node data is not equal to `oldValue`, the answer has to be `false`.
Similar way it is in the `PropertiesPersistingMetadataStore`.

DragonFSKY:
Fixes: https://github.com/spring-projects/spring-integration/issues/11013

`ZookeeperMetadataStore.replace()` is expected to behave as a compare-and-set operation. It was returning `true` after reading the current ZooKeeper value even when that value did not match the supplied `oldValue`.

* Return `false` when the stored value differs from `oldValue` and no replacement is performed.
* Extend `ZookeeperMetadataStoreTests.testReplace()` to verify the failed replacement result and that the stored value is left unchanged.

Tests:

`./gradlew :spring-integration-zookeeper:test`

### DIFF
--- a/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
+++ b/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
@@ -151,8 +151,9 @@ public class ZookeeperMetadataStore implements ListenableMetadataStore, SmartLif
 				byte[] bytes = this.client.getData().storingStatIn(currentStat).forPath(getPath(key));
 				if (oldValue.equals(IntegrationUtils.bytesToString(bytes, this.encoding))) {
 					updateNode(key, newValue, currentStat.getVersion());
+					return true;
 				}
-				return true;
+				return false;
 			}
 			catch (KeeperException.NoNodeException | KeeperException.BadVersionException ex) {
 				// ignore, the node doesn't exist there's nothing to replace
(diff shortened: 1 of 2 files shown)


artembilan:
> Tests:
 ```
./gradlew :spring-integration-zookeeper:test --tests org.springframework.integration.zookeeper.metadata.ZookeeperMetadataStoreTests.testReplace
./gradlew :spring-integration-zookeeper:test
```
Why do we need to run an individual test if it is the part of the whole test suite?

DragonFSKY:
I simplified the PR description and kept only the full module test command.

The focused test was the initial regression check I used while developing the fix, but it is redundant in the PR description since `:spring-integration-zookeeper:test` includes it.

artembilan (approved):
LGTM.
But for the future contributions, please, consider to learn this: https://cbea.ms/git-commit/.

The point is that PR disappears just immediately after merge.
What is matter is a commit in the history. And would be great to have enough "Why?" in that message.

For now I need to go over and modify your commit message respectively.

Thanks

PR #11020, opened 2026-05-25
Fix misleading Javadoc for messaging annotation post-processors

mjd507:
The `MessagingAnnotationPostProcessor` is really not a `BeanPostProcessor`.

Also, update the Javadocs to describe `MessagingAnnotationPostProcessor` and `MessagingAnnotationBeanPostProcessor` as handling `@Bean` methods and regular component methods, respectively.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java
@@ -50,7 +50,7 @@ import org.springframework.util.StringUtils;
 
 /**
  * An infrastructure {@link BeanPostProcessor} implementation that processes method-level
- * messaging annotations such as @Transformer, @Splitter, @Router, and @Filter.
+ * messaging annotations such as @Transformer, @Splitter, @Router, and @Filter, on regular component methods.
  *
  * @author Artem Bilan
  *
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java
@@ -50,7 +50,7 @@ import org.springframework.util.StringUtils;
 
 /**
  * An infrastructure {@link BeanPostProcessor} implementation that processes method-level
- * messaging annotations such as @Transformer, @Splitter, @Router, and @Filter, on regular component methods.
+ * messaging annotations (@Transformer, @Splitter, @Router, @Filter etc.) on bean methods.
  *
  * @author Artem Bilan
  *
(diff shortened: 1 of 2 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationPostProcessor.java
```diff
@@ -51,8 +50,8 @@
 import org.springframework.util.CollectionUtils;
 
 /**
- * A {@link BeanPostProcessor} implementation that processes method-level
- * messaging annotations such as @Transformer, @Splitter, @Router, and @Filter.
+ * A {@link BeanDefinitionRegistryPostProcessor} implementation that processes method-level
+ * messaging annotations such as @Transformer, @Splitter, @Router, and @Filter, on @Bean methods.
```
cppwfs: How about this, 

```
* A {@link BeanDefinitionRegistryPostProcessor} implementation that processes method-level
* messaging annotations (such as @Transformer, @Splitter, @Router, and @Filter) declared on @Bean methods.
```

spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java
```diff
@@ -50,7 +50,7 @@
 
 /**
  * An infrastructure {@link BeanPostProcessor} implementation that processes method-level
```
cppwfs, line 52: How about this modification?
``` 
* An infrastructure {@link BeanPostProcessor} implementation that processes method-level
* messaging annotations (such as @Transformer, @Splitter, @Router, and @Filter) declared on regular component methods.
 ```

**Review by cppwfs — DISMISSED** [review]
Thank you for your contributions, 
I made a couple of small requests for your review.

File: spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java [hunk 3]
```diff
@@ -50,7 +50,7 @@
...
- * messaging annotations such as @Transformer, @Splitter, @Router, and @Filter.
+ * messaging annotations such as @Transformer, @Splitter, @Router, and @Filter, on regular component methods.
```
artembilan: I think we need to revise this more into something like:
```suggestion
 * messaging annotations(@Transformer, @Splitter, @Router, @Filter etc.) on bean methods.
```

spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationPostProcessor.java
artembilan:
```suggestion
 * messaging annotations(@Transformer, @Splitter, @Router, @Filter etc.) on @Bean configuration methods.
```

artembilan (changes requested):

mjd507:
thank you all!

Btw, I have another query: 

should we provide option to skip this two post processors, since nowadays user may use DSL only.

May be it would be a little improvement for performance.

artembilan:
> should we provide option to skip this two post processors, since nowadays user may use DSL only.

that’s interesting idea!
Let’s create an issue to track that down !
Perhaps something like `@EnableIntegration(parseAnnotations = false)` would do the trick 😅

mjd507:
> that’s interesting idea! Let’s create an issue to track that down !

sure. created:  https://github.com/spring-projects/spring-integration/issues/11024

PR #11025, opened 2026-05-27
Invalid

DragonFSKY:


cppwfs:
Thank you for your contribution! I've been revisiting the original issue, and it is a bit vague. For now, I'm closing it so we can better articulate how EIP should handle this scenario.
We value all contributions and appreciate the work you have done on this.

PR #11026, opened 2026-05-27
Introduce `processAnnotations` attribute to `@EnableIntegration`.

mjd507:
this allows disabling messaging annotation processing when applications rely exclusively on the DSL.

- Add `EnableIntegrationTests` to demonstrate startup performance improvement.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
@@ -55,15 +55,16 @@ import org.springframework.context.annotation.Import;
 public @interface EnableIntegration {
 
 	/**
-	 * Indicate whether to parse method-level annotations such as {@code @Filter}, {@code Router},
-	 * {@code @ServiceActivator}, {@code InboundChannelAdapter}, etc.
-	 * <p> Can be set to {@code false} if the application exclusively uses the Java DSL.
-	 * @return true to parse the method-level annotations.
+	 * Indicate whether to process messaging annotations such as {@code @Filter}, {@code Router},
+	 * {@code @ServiceActivator}, {@code InboundChannelAdapter}, {@code MessagingGateway} etc.
+	 * <p> Can be set to {@code false} if the application exclusively uses the {@code IntegrationFlow}
+	 * definitions.
+	 * @return true to process the messaging annotations.
 	 * @see MessagingAnnotationPostProcessor
 	 * @see MessagingAnnotationBeanPostProcessor
 	 * @see GatewayProxyInstantiationPostProcessor
 	 * @since 7.1
 	 */
-	boolean parseAnnotations() default true;
+	boolean processAnnotations() default true;
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
@@ -42,7 +42,10 @@ import org.springframework.context.annotation.Import;
  * </pre>
  *
  * @author Artem Bilan
+ * @author Jiandong Ma
+ *
  * @since 4.0
+ *
  * @see IntegrationRegistrar
  */
 @Target(ElementType.TYPE)
@@ -50,5 +53,17 @@ import org.springframework.context.annotation.Import;
 @Documented
 @Import(IntegrationRegistrar.class)
 public @interface EnableIntegration {
+
+	/**
+	 * Indicate whether to parse method-level annotations such as {@code @Filter}, {@code Router},
+	 * {@code @ServiceActivator}, {@code InboundChannelAdapter}, etc.
+	 * <p> Can be set to {@code false} if the application exclusively uses the Java DSL.
+	 * @return true to parse the method-level annotations.
+	 * @see MessagingAnnotationPostProcessor
+	 * @see MessagingAnnotationBeanPostProcessor
+	 * @see GatewayProxyInstantiationPostProcessor
+	 * @since 7.1
+	 */
+	boolean parseAnnotations() default true;
 
 }
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -25,6 +25,7 @@ import org.springframework.beans.factory.support.BeanDefinitionBuilder;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.context.ApplicationContextException;
 import org.springframework.context.annotation.ImportBeanDefinitionRegistrar;
+import org.springframework.core.annotation.AnnotationAttributes;
 import org.springframework.core.type.AnnotationMetadata;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.util.ClassUtils;
@@ -35,6 +36,7 @@ import org.springframework.util.ClassUtils;
  * @author Artem Bilan
  * @author Gary Russell
  * @author Chris Bono
+ * @author Jiandong Ma
  *
  * @since 4.0
  */
@@ -64,10 +66,22 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
 
 		registerDefaultConfiguringBeanFactoryPostProcessor(registry);
 		registerIntegrationConfigurationBeanFactoryPostProcessor(registry);
+
+		boolean parseAnnotations = true;
 		if (importingClassMetadata != null) {
-			registerMessagingAnnotationPostProcessors(registry);
+			AnnotationAttributes enableIntegration = AnnotationAttributes.fromMap(
+					importingClassMetadata.getAnnotationAttributes(EnableIntegration.class.getName()));
+
+			if (enableIntegration != null && enableIntegration.containsKey("parseAnnotations")) {
+				parseAnnotations = enableIntegration.getBoolean("parseAnnotations");
+			}
+			if (parseAnnotations) {
+				registerMessagingAnnotationPostProcessors(registry);
+			}
 		}
-		registerGatewayProxyInstantiationPostProcessor(registry);
+		if (parseAnnotations) {
+			registerGatewayProxyInstantiationPostProcessor(registry);
+		}
 	}
 
 	/**
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -67,19 +67,19 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
 		registerDefaultConfiguringBeanFactoryPostProcessor(registry);
 		registerIntegrationConfigurationBeanFactoryPostProcessor(registry);
 
-		boolean parseAnnotations = true;
+		boolean processAnnotations = true;
 		if (importingClassMetadata != null) {
 			AnnotationAttributes enableIntegration = AnnotationAttributes.fromMap(
 					importingClassMetadata.getAnnotationAttributes(EnableIntegration.class.getName()));
 
-			if (enableIntegration != null && enableIntegration.containsKey("parseAnnotations")) {
-				parseAnnotations = enableIntegration.getBoolean("parseAnnotations");
+			if (enableIntegration != null) {
+				processAnnotations = enableIntegration.getBoolean("processAnnotations");
 			}
-			if (parseAnnotations) {
+			if (processAnnotations) {
 				registerMessagingAnnotationPostProcessors(registry);
 			}
 		}
-		if (parseAnnotations) {
+		if (processAnnotations) {
 			registerGatewayProxyInstantiationPostProcessor(registry);
 		}
 	}
--- a/spring-integration-core/src/test/java/org/springframework/integration/config/EnableIntegrationTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/config/EnableIntegrationTests.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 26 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
```diff
@@ -51,4 +54,16 @@
 @Import(IntegrationRegistrar.class)
 public @interface EnableIntegration {
 
+	/**
+	 * Indicate whether to parse method-level annotations such as {@code @Filter}, {@code Router},
```
artembilan: Well, this is not just only about methods.
See `@MessageEndpoint` or `@MessagingGateway`.
So, say here, please, more context-related matter:
```
* Indicate whether to process messaging annotations such as {@code @MessagingGateway}, {@code @Filter}, {@code Router},
```

spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
```diff
@@ -51,4 +54,16 @@
...
+	 * {@code @ServiceActivator}, {@code InboundChannelAdapter}, etc.
+	 * <p> Can be set to {@code false} if the application exclusively uses the Java DSL.
+	 * @return true to parse the method-level annotations.
+	 * @see MessagingAnnotationPostProcessor
+	 * @see MessagingAnnotationBeanPostProcessor
+	 * @see GatewayProxyInstantiationPostProcessor
+	 * @since 7.1
+	 */
+	boolean parseAnnotations() default true;
```
artembilan: We discussed this a bit with @cppwfs and decided to name this as `processAnnotations`.
Yes, we keep this as `true` by default in this version but lean to `false` in the next one.
For now we just introduce this option for awareness, and in the next version we switch the plug to make apps starting better 😄 

And I also wonder if need another option for DSL respectively:
```
boolean processIntegrationFlows() default true;
```
WDYT?

spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
```diff
@@ -51,4 +54,16 @@
 @Import(IntegrationRegistrar.class)
 public @interface EnableIntegration {
 
+	/**
+	 * Indicate whether to parse method-level annotations such as {@code @Filter}, {@code Router},
+	 * {@code @ServiceActivator}, {@code InboundChannelAdapter}, etc.
+	 * <p> Can be set to {@code false} if the application exclusively uses the Java DSL.
```
artembilan: This is really not only about Java DSL.
You see: we also have Kotlin and Groovy DSL which do exactly the same but in their own coding style.
So, let's consider to change wording to the `exclusively uses the {@code IntengrationFlow} definitions.`

artembilan (changes requested):
Looks promising!

I believe this deserves a `whats-new.adoc` entry and some wording about new attribute in the `annotations.adoc` & `overview.adoc`, too.

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -64,10 +66,22 @@ public void registerBeanDefinitions(@Nullable AnnotationMetadata importingClassM
...
 		registerDefaultConfiguringBeanFactoryPostProcessor(registry);
 		registerIntegrationConfigurationBeanFactoryPostProcessor(registry);
+
+		boolean parseAnnotations = true;
 		if (importingClassMetadata != null) {
-			registerMessagingAnnotationPostProcessors(registry);
+			AnnotationAttributes enableIntegration = AnnotationAttributes.fromMap(
+					importingClassMetadata.getAnnotationAttributes(EnableIntegration.class.getName()));
+
+			if (enableIntegration != null && enableIntegration.containsKey("parseAnnotations")) {
```
artembilan: According to our `@EnableIntegration` definition, that attribute is primitive and therefore always there, so we don't need this `containsKey` check.

spring-integration-core/src/test/java/org/springframework/integration/config/EnableIntegrationTests.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+class EnableIntegrationTests {
+
+	@Test
+	void parseAnnotationsPerformanceTest() {
+		long startupTimeWhenParseAnnotationsEnabled;
+		long startupTimeWhenParseAnnotationsDisabled;
+		try (GenericApplicationContext context = new GenericApplicationContext()) {
+			AnnotationConfigUtils.registerAnnotationConfigProcessors(context);
+			AnnotatedBeanDefinitionReader reader = new AnnotatedBeanDefinitionReader(context);
```
artembilan: I think we can simply (or let's say shorten) the test code using `new AnnotationConfigApplicationContext(ConfigWithParseAnnotationsEnabled.class)` instead.
And `StopWatch` around the whole `try..catch`.
Plus, take a look if reusing same `StopWatch` would make sense for our logic.
I mean we can start/stop it several times to gather statistics for different scenarios.
See `MethodInvokingMessageProcessorTests.testPerformanceSpelVersusInvocable()`

artembilan (changes requested):

artembilan:
After looking into this deeper I think we need to be careful.
 The problem is so obvious and proposed solution is so simple and elegant that we have gone into the trap.
The `@EnableIntegration` can be declared in the target application in several places.
Or even more: some of them could be declared in the library like Spring Boot auto-configuration.
This is fine for the current state of this annotation: the `IntegrationRegistrar` makes sure we don't override already registered infrastructure beans.
With this new attribute parsing this `@EnableIntegration` from different places may lead to unexpected result.
For example, one of them comes with defaults and the other disabling annotations processing.
So, one would skip beans to be registered, but another would add them again, just because there are no those beans in the application context.
Or even worse: the first registers all of them due to its defaults, and the other assumes that they are not going to be there, but that is not true due to skip logic.

To avoid that we might go with something like "disabled by default and use `enable` to opt-in".
Therefore I'm think about an `@EnableMessagingAnnotations` like an addition to the `@EnableIntegration` and something similar to the `@EnablePublisher`.

It is probably too late for the current `7.1.0` since we are past `RC` phase already.
So, making a value from the performance improvement benefit and introducing those non-ambiguous options but as a breaking change anyway must go to the next minor.

Just some my thoughts.
I'm open for any other ideas and will keep the issue for the current milestone for a couple more days before we start release next week or until we agree what really to do.

Thanks

cppwfs:
A thought if we offer an "opt-in" model we probably want to wait for a major release to do that.   Spring Batch did something similar and it confused many users that required a lot of messaging, to help them along.

mjd507:
> The `@EnableIntegration` can be declared in the target application in several places. 

My thought for this is our final goal is (by default) disabling messaging annotations processing,  so no matter how many `@EnableIntegration` user declares, the final effect is same. 

> With this new attribute parsing this `@EnableIntegration` from different places may lead to unexpected result. For example, one of them comes with defaults and the other disabling annotations processing. So, one would skip beans to be registered, but another would add them again, just because there are no those beans in the application context. Or even worse: the first registers all of them due to its defaults, and the other assumes that they are not going to be there, but that is not true due to skip logic.

True, this like a middle stage that can not avoid, but the attribute only control whether to register the messaging annotations post processors, we can document if user want to explicitly disable this annotations processing, in this middle stage, they have to ensure only one `@EnableIntegration` in their application. (or we do not need to say anything, the default behavior is same).

> Or even more: some of them could be declared in the library like Spring Boot auto-configuration. This is fine for the current state of this annotation: the `IntegrationRegistrar` makes sure we don't override already registered infrastructure beans. 

I see Spring Boot auto-configuration will active this `@EnableIntegration`, I think we need to extract  a boolean field `processAnnotations` in the `IntegrationProperties`, and Boot can leverage this condition.  also have to mention if use Boot auto-configuration, no need `@EnableIntegration`. (but if end-user have, from the final default stage, it's same result, but if they mixed properties config and annotation config, the worst case is same as now?)

> To avoid that we might go with something like "disabled by default and use `enable` to opt-in". Therefore I'm think about an `@EnableMessagingAnnotations` like an addition to the `@EnableIntegration` and something similar to the `@EnablePublisher`.

I will try this approach as well. 

> It is probably too late for the current `7.1.0` since we are past `RC` phase already. So, making a value from the performance improvement benefit and introducing those non-ambiguous options but as a breaking change anyway must go to the next minor.

can't wait to be merged 😊

spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
mjd507: for DSL `IntegrationFlowBeanPostProcessor`, looks only `instanceof` check,  I think its quick, prefer not add.

```java
	public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
		if (bean instanceof StandardIntegrationFlow standardIntegrationFlow) {
			return processStandardIntegrationFlow(standardIntegrationFlow, beanName);
		}
		else if (bean instanceof IntegrationFlow integrationFlow) {
			return processIntegrationFlowImpl(integrationFlow, beanName);
		}
		if (bean instanceof IntegrationComponentSpec<?, ?> integrationComponentSpec) {
			processIntegrationComponentSpec(beanName, integrationComponentSpec);
		}
		return bean;
	}
```

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
mjd507: I didn't add this `containsKey` in the first, but few test failed because it pass a mock `AnnotationMetadata` to this method. 

`new IntegrationRegistrar().registerBeanDefinitions(mock(), context.getDefaultListableBeanFactory())`.

eventually, the returned a object `AnnotationAttributes` is not null and is a empty map, and the `getBoolean()` method will throw `IllegalArgumentException` because it requires the attribute exists

artembilan:
Another thought before I go to sleep. Even if it is crazy and could be discarded and I’ll definitely change my mind in the morning 😜.

We can introduce `@EnableIntegrationFlows` to do exactly what you propose in this PR: integration infra but without annotations processing. And of course the mentioned `@EnableMessagingAnnotations` to fulfill the gap with opposite behavior : integration infra but no DSL parser.

This way no breaking changes but there is an option for optimization. 

The `@EnableIntegration` could be repurposed in the next version.

we still need to think how to be with Spring Boot and unfortunately they don’t accept new features after RC.

we may look into the respective global property to introduce for disabling annotations: https://docs.spring.io/spring-integration/reference/configuration/global-properties.html. Some interim solution before we incorporate new feature into Spring Boot.

I might come with more thoughts tomorrow .

thanks

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
artembilan: So, that is the test problem.
It is not fair to adapt our prod code for test expectations .

I don’t belive in TDD since it leads to traps like this.

spring-integration-core/src/test/java/org/springframework/integration/config/EnableIntegrationTests.java
mjd507: my bad, I found, if i change the order, let say make the enabling messaging processor context start first,  test will failed. looks performance does **not** have a great improvement.  
I think it could be JVM reason, first time load the class takes time, then second time faster.

artembilan: So, does it mean that the whole effort doesn't worse it?

mjd507: hmm.. Let me try to test after the JVM warmup. if not that good, we can close this.

artembilan: I think the test config here does not prove anything. The point of messaging annotations processing is to have beans in the config with them. Now you have none, therefore nothing to measure for stop watch. Or even more closer to the expectation scenario: beans but without those annotations, so the processor would look redundant.

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/config/EnableIntegrationTests.java
mjd507: > even more closer to the expectation scenario: beans but without those annotations, so the processor would look redundant.

yeah, my purpose is exactly like you said. people use DSL may not use those messaging annotations right, I just test how those redundant messaging annotation processors impact our app startup.

update: by a one-time warmup (load both context first),  in my local, with the minimal configuration case, enable messaging annotations take above 10 ms longer that skip this processors. (I run more than 10 times, and no matter the order after warm up) very stable 10ms better.

so I think is still worth. since users app are bigger than this minimal config.

artembilan: > people use DSL may not use those messaging annotations

right that is the goal of this whole PR. However I mean exactly this test . Without any beans we don’t measure a gain. But when we have them in this watch-based test, we can show-case how harmful those annotation processors when there is nothing for them to parse .

mjd507: > Without any beans we don’t measure a gain. 

ha, `@EnableIntegration` will also register a `DefaultConfiguringBeanFactoryPostProcessor`,  
and many infrastructure beans like `errorChannel`, `nullChannel`, `TaskScheduler`, `IntegrationEvaluationContext`, etc  are registered by default if we not defined.

artembilan: Ok. Double check that in debug please, to see that all of those are fed to the annotation processors .
I’ll take a look in the morning when reach laptop.

mjd507:
> The @EnableIntegration can be declared in the target application in several places.

I think it would be good if we can restrict the `@EnableIntegration` declaration .
like we do in `PublisherRegistrar`. (`@EnablePublisher`)
```java
if (registry.containsBeanDefinition(IntegrationContextUtils.PUBLISHER_ANNOTATION_POSTPROCESSOR_NAME)) {
	throw new BeanDefinitionStoreException("Only one enable publisher definition " +
			"(@EnablePublisher or <annotation-config>) can be declared in the application context.");
}
```
so, only one  `@EnableIntegration`  can be exist in application.  
if use spring boot, user no need to declared it in any configuration. 
this way, we can add feature(like `processAnnotations`) easily by `IntegrationProperties`. 
and user can only change the behavior from the yaml properties.

but, it could cause the startup error if user declared `@EnableIntegration`,  
but I think it acceptable, since it's a one-time removal change.

artembilan:
Closed in favor of: https://github.com/spring-projects/spring-integration/pull/11029.

With all those "iffs" it doesn't worse pursinging this solution.
Plus the `@MessagingGateway` must left outside.

The other PR goes slightly different direction which we will revise a bit later.

Thanks

PR #11027, opened 2026-05-28
Update `FileTests` and `JmsTests` to pure DSL approach

mjd507:


cppwfs:
The use of annotations in tests is acceptable and sometimes a desire of the original developer.    We will decline this PR for these reasons.

Thank you for your contribution!   And thank you for all the hard work!

mjd507:
> The use of annotations in tests is acceptable and sometimes a desire of the original developer.    We will decline this PR for these reasons.

I just find both tests are under dsl package, so I think it's good to make it pure dsl, to align with all other modules dsl tests.

artembilan:
> to align with all other modules dsl tests.

well, those are tests and I don’t care too much in what package they are in. What I worry is more coverage with less code. And this kind of integration tests do the trick for us.

PR #11028, opened 2026-05-28
Revise tests using `IntegrationRegistrar` to register infra beans

mjd507:
Instead, using `@EnableIntegration` approach combined with `AnnotationConfigApplicationContext`.


artembilan (approved):
OK. I don't have any objections on this.
Thanks
@cppwfs , WDYT?

cppwfs:
@artembilan 
No objections.  The solution is easier to read and the proper solution for creating the context.

Issue #11024, opened 2026-05-26
Allow skipping messaging annotation post processor registration in pure DSL scenarios


mjd507:
Currently, enabling `@EnableIntegration` will activate 
- DSL post processors (`DslIntegrationConfigurationInitializer`) 
- Messaging annotation post processors(`MessagingAnnotationPostProcessor`, `MessagingAnnotationBeanPostProcessor`, `GatewayProxyInstantiationPostProcessor`). 
 
All beans are processed by above sets of post processors.

For applications that use DSL only, the messaging annotation post processors is redundant. 
So, to reduce unnecessary startup overhead, propose to add an option to not register messaging annotation post processors.

artembilan:
1. The solution like `@EnableIntegration(processAnnotations = false)` is rejected due to the conflict when several `@EnableIntegration` are declared in different places with different configs: https://github.com/spring-projects/spring-integration/pull/11026
2. The solution like `@EnableMessagingAnnotations` is rejected (at least for now) due to the complexity and confusion how really to disable which was a premise of this issue: https://github.com/spring-projects/spring-integration/pull/11029
3. Both solutions don't help currently with existing applications and for the upcoming Spring Boot `4.1` which does not accept new feature in its current phase.

Right now I see a simple solution like [Integration Global Property](https://docs.spring.io/spring-integration/reference/configuration/global-properties.html).
Let's say like this:
```
spring.integration.annotations.enable=true
```
!
No any annotations changes, preserve default and easy to opt-in. Even with no changes for Spring Boot!

This property should be consulted in the `MessagingAnnotationPostProcessor` and `MessagingAnnotationBeanPostProcessor`.
The `GatewayProxyInstantiationPostProcessor` must be left untouched.
It is related only to the `@MessagingGateway` and does not have hard logic.

We may consider in the future to introduce `@EnableMessagingAnnotations` since we are willing to turn off those annotations by default.
This way the user would have an option to enable them back with familiar annotations API.
The property will be promoted to similar in Spring Boot auto-configuration.

For now I don't see any reasons to make something complex with annotations and introduce some breaking changes and new bugs.

We may consider a fine-grained optimization in the `MessagingAnnotationBeanPostProcessor` to skip some set of beans for their methods processing.
But that is a different story for different day.

Any comments welcome!

Thanks

mjd507:
OK, 

for solution1: seems add restriction to let user remove '@EnableIntegration' is a hard thing! 

[...]

PR #11029, opened 2026-05-29
Introduce `@EnableMessagingAnnotations` to activate messaging annotations post processors

mjd507:
this is another option for #11024

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/xml/AnnotationConfigParser.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/xml/AnnotationConfigParser.java
@@ -30,6 +30,7 @@ import org.springframework.beans.factory.xml.ParserContext;
 import org.springframework.core.type.MethodMetadata;
 import org.springframework.integration.config.EnablePublisher;
 import org.springframework.integration.config.IntegrationRegistrar;
+import org.springframework.integration.config.MessagingAnnotationsRegister;
 import org.springframework.integration.config.PublisherRegistrar;
 import org.springframework.integration.config.annotation.AnnotationMetadataAdapter;
 import org.springframework.util.xml.DomUtils;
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/config/xml/AnnotationConfigParser.java
```diff
@@ -52,6 +53,8 @@ public class AnnotationConfigParser implements BeanDefinitionParser {
 		BeanDefinitionRegistry registry = parserContext.getRegistry();
 		new IntegrationRegistrar()
 				.registerBeanDefinitions(importingClassMetadata, registry);
+		new MessagingAnnotationsRegister()
+				.registerBeanDefinitions(importingClassMetadata, registry);
```
mjd507, line 57: activate messaging annotations for `ClassPathXmlApplicationContext` that uses `<integration:annotation-config/>`

spring-integration-core/src/main/java/org/springframework/integration/config/EnableIntegration.java
```diff
@@ -48,7 +48,8 @@
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@Import(IntegrationRegistrar.class)
+@Import({IntegrationRegistrar.class})
+@EnableMessagingAnnotations
```
mjd507, line 52: activate by default, currently not a good time to fix tests.

mjd507:
there is a `@IntegrationComponentScan` used for scanning `MessagingGateway`, SpringBoot by default auto-configured it.  looks it will out of this `@EnableMessagingAnnotations` control.  what we want is not to scan if user not enable messaging annotations.  just raise this concern.

artembilan:
Interesting idea, and it really was proposed on the other PR.
However, this is still not what we would like to go right now.
Need more engineering thinking.
So, let's continue discussion in the issue!

Thanks

PR #11038, opened 2026-06-02
Introduce `enableAnnotations` in global `IntegrationProperties`

mjd507:
the value default to true but can be set to false via:
1. adding one line config `spring.integration.annotations.enable=false` into the classpath file `spring.integration.properties`
2. defining a new `integrationGlobalProperties` bean with `setEnableAnnotations(false)`.

Also add a test to demonstrate the performance improvement when turnoff the annotation processing. see: `MessagingAnnotationsConfigurationTests.testEnableVersusDisablePerformance`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/MessagingAnnotationBeanPostProcessor.java
@@ -25,6 +25,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.jspecify.annotations.Nullable;
+
 import org.springframework.aop.support.AopUtils;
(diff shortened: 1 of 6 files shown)


artembilan:
Hey, @mjd507 !

Thank you very much for all the help and we really appreciate your effort, flexibility and critics.

Sorry to say that, but this is another attempt we have to reject.
Not because you did something wrong, but because I have misled you and everyone around how this should go.

So, after looking into Spring Boot and specifically [integration auto-configuration](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-integration/src/main/java/org/springframework/boot/integration/autoconfigure/IntegrationAutoConfiguration.java#L96), I have realized that such a global property won't have effect because it is not mapped anyhow in there. So, without introducing such a property into Spring Boot, we cannot have a gain just for Spring Integration.
And that is already not counting the problem you have mentioned about bean resolution in those post-processors.

Thank you again and let's continue brain-storming in the issue!

PR #11039, opened 2026-06-03
Support disabling messaging annotations processing

mjd507:
By adding `spring.integration.annotations.enable=false` into the environment property or application.properties.

When disabled, the following post processors are not registered in the application context:

- `MessagingAnnotationPostProcessor`
- `MessagingAnnotationBeanPostProcessor`

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationConfigUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationConfigUtils.java
@@ -37,6 +37,8 @@ public final class IntegrationConfigUtils {
 
 	public static final String HANDLER_ALIAS_SUFFIX = ".handler";
 
+	public static final String ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING = "spring.integration.annotations.enable";
+
 	public static void registerSpelFunctionBean(BeanDefinitionRegistry registry, String functionId, String className,
 			String methodSignature) {
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationConfigUtils.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationConfigUtils.java
@@ -37,8 +37,6 @@ public final class IntegrationConfigUtils {
 
 	public static final String HANDLER_ALIAS_SUFFIX = ".handler";
 
-	public static final String ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING = "spring.integration.annotations.enable";
-
 	public static void registerSpelFunctionBean(BeanDefinitionRegistry registry, String functionId, String className,
 			String methodSignature) {
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -24,7 +24,9 @@ import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.context.ApplicationContextException;
+import org.springframework.context.EnvironmentAware;
 import org.springframework.context.annotation.ImportBeanDefinitionRegistrar;
+import org.springframework.core.env.Environment;
 import org.springframework.core.type.AnnotationMetadata;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.util.ClassUtils;
@@ -35,10 +37,11 @@ import org.springframework.util.ClassUtils;
  * @author Artem Bilan
  * @author Gary Russell
  * @author Chris Bono
+ * @author Jiandong Ma
  *
  * @since 4.0
  */
-public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
+public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar, EnvironmentAware {
 
 	static {
 		if (ClassUtils.isPresent("org.springframework.integration.dsl.support.Function", null)) {
@@ -47,6 +50,14 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
 					+ "the Java DSL has been merged into the core project. "
 					+ "If it is present on the classpath, it will cause class loading conflicts.");
 		}
+	}
+
+	@SuppressWarnings("NullAway.Init")
+	private Environment environment;
+
+	@Override
+	public void setEnvironment(Environment environment) {
+		this.environment = environment;
 	}
 
 	/**
@@ -64,7 +75,11 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
 
 		registerDefaultConfiguringBeanFactoryPostProcessor(registry);
 		registerIntegrationConfigurationBeanFactoryPostProcessor(registry);
-		if (importingClassMetadata != null) {
+
+		boolean enableMessagingAnnotationsProcessing = this.environment
+				.getProperty(IntegrationConfigUtils.ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING, Boolean.class, true);
+
+		if (importingClassMetadata != null && enableMessagingAnnotationsProcessing) {
 			registerMessagingAnnotationPostProcessors(registry);
 		}
 		registerGatewayProxyInstantiationPostProcessor(registry);
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -52,8 +52,6 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar, Envi
 		}
 	}
 
-	static final String ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING = "spring.integration.annotations.enable";
-
 	@SuppressWarnings("NullAway.Init")
 	private Environment environment;
 
@@ -78,10 +76,10 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar, Envi
 		registerDefaultConfiguringBeanFactoryPostProcessor(registry);
 		registerIntegrationConfigurationBeanFactoryPostProcessor(registry);
 
-		if (importingClassMetadata != null) {
-			if (this.environment.getProperty(ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING, Boolean.class, true)) {
-				registerMessagingAnnotationPostProcessors(registry);
-			}
+		if (importingClassMetadata != null
+				&& this.environment.getProperty("spring.integration.annotations.enable", Boolean.class, true)) {
+
+			registerMessagingAnnotationPostProcessors(registry);
 		}
 		registerGatewayProxyInstantiationPostProcessor(registry);
 	}
--- a/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
@@ -52,6 +52,8 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar, Envi
 		}
 	}
 
+	static final String ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING = "spring.integration.annotations.enable";
+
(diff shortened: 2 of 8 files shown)


artembilan (changes requested):
Please, don't touch `GatewayProxyInstantiationPostProcessor` and `IntegrationComponentScanRegistrar`.
They are for components scanning, for those which represents some entries on classpath.

Their purpose similar to the `@ComponentScan` in Spring Framework.

What we are trying to prevent here is annotations on methods in beans.
So, our goal would be similar to opting-in into something like `@EnableMvc`, but that does not mean that when we don't opt-in, the  `@ComponentScan` should not work, too.

I agree that something like `@EnableIntegrationMethodAnnotations` would sound better for you to not look into the `IntegrationComponentScanRegistrar`, but I don't think we need to go such a fine-grain naming while we can just document this new property and future annotation properly.

Please, make a respective change and we will come back to you shortly for the further review.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationConfigUtils.java
```diff
@@ -37,6 +37,8 @@ public final class IntegrationConfigUtils {
 
 	public static final String HANDLER_ALIAS_SUFFIX = ".handler";
 
+	public static final String ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING = "spring.integration.annotations.enable";
```
artembilan: I think we don’t to have it as a public API which we are going to deprecate in the next version in favor of the annotation to avoid a paradox of choice and extra work on our side.

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -64,7 +75,11 @@ public void registerBeanDefinitions(@Nullable AnnotationMetadata importingClassM
 
 		registerDefaultConfiguringBeanFactoryPostProcessor(registry);
 		registerIntegrationConfigurationBeanFactoryPostProcessor(registry);
-		if (importingClassMetadata != null) {
+
+		boolean enableMessagingAnnotationsProcessing = this.environment
```
artembilan: This local variable and the whole get operation becomes redundant if no `importingClassMetadata` is provided.
Plus, this call will be removed in next version from this class in favor of the logic in new registrar for new annotation

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractIntegrationNamespaceHandler.java
```diff
@@ -50,8 +52,13 @@ public abstract class AbstractIntegrationNamespaceHandler extends NamespaceHandl
...
 	public final @Nullable BeanDefinition parse(Element element, ParserContext parserContext) {
 		if (!this.initialized.getAndSet(true)) {
+			Environment environment = parserContext.getReaderContext().getEnvironment();
+
+			IntegrationRegistrar integrationRegistrar = new IntegrationRegistrar();
+			integrationRegistrar.setEnvironment(environment);
+
 			BeanDefinitionRegistry registry = parserContext.getRegistry();
-			new IntegrationRegistrar().registerBeanDefinitions(null, registry);
+			integrationRegistrar.registerBeanDefinitions(null, registry);
```
artembilan: I think what you did in this class won’t have any affect because this `null` for first arg.

spring-integration-core/src/test/java/org/springframework/integration/config/IntegrationRegistrarTests.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright 2014-present the original author or authors.
```
artembilan: The date must be when it was introduced first

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/UriVariableExpressionTests.java
```diff
@@ -121,6 +122,7 @@ public void testInt3055UriVariablesExpression() {
 
 		AbstractApplicationContext context = new GenericApplicationContext();
 		IntegrationRegistrar registrar = new IntegrationRegistrar();
+		registrar.setEnvironment(context.getEnvironment());
```
artembilan: When you push down the property resolution after imported metadata check , the change in this class won’t be necessary .

artembilan (changes requested):
Please, consider to mention this property in the respective annotations chapter and in the what’s new one, too.

thanks

spring-integration-core/src/main/java/org/springframework/integration/config/xml/AnnotationConfigParser.java
```diff
@@ -43,14 +44,20 @@
...
 	public @Nullable BeanDefinition parse(Element element, ParserContext parserContext) {
+		Environment environment = parserContext.getReaderContext().getEnvironment();
+
+		IntegrationRegistrar integrationRegistrar = new IntegrationRegistrar();
+		integrationRegistrar.setEnvironment(environment);
+
 		ExtendedAnnotationMetadata importingClassMetadata = new ExtendedAnnotationMetadata(element);
 		BeanDefinitionRegistry registry = parserContext.getRegistry();
-		new IntegrationRegistrar()
+		integrationRegistrar
```
artembilan: This looks awkward having a method call on a different line.
This is really not a method chain as it was before with the ctor, and it is not that long to justify separation.
Just nit-pick for the code readability.

Thanks

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -49,6 +52,16 @@ public class IntegrationRegistrar implements ImportBeanDefinitionRegistrar {
 		}
 	}
 
+	static final String ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING = "spring.integration.annotations.enable";
```
artembilan: Why do we need this constant at all?

spring-integration-core/src/main/java/org/springframework/integration/config/IntegrationRegistrar.java
```diff
@@ -64,8 +77,11 @@ public void registerBeanDefinitions(@Nullable AnnotationMetadata importingClassM
 
 		registerDefaultConfiguringBeanFactoryPostProcessor(registry);
 		registerIntegrationConfigurationBeanFactoryPostProcessor(registry);
+
 		if (importingClassMetadata != null) {
-			registerMessagingAnnotationPostProcessors(registry);
+			if (this.environment.getProperty(ENV_ENABLE_MESSAGING_ANNOTATIONS_PROCESSING, Boolean.class, true)) {
```
artembilan: Why these to `ifs` cannot be combined into one `&&`?

src/reference/antora/modules/ROOT/pages/configuration/annotations.adoc
```diff
@@ -20,6 +20,9 @@ public class FooService {
 }
 ----
 
+NOTE: Starting with version 7.1, the `spring.integration.annotations.enable` environment property can be used to control method-level annotation processing.
+By default, this property is `true`. It can be set to `false` for applications that exclusively rely on `IntegrationFlow` definitions.
```
artembilan: One sentence per line, please: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

I think just talking about `IntegrationFlow` is not correct: no one said that `@EnableIntegration` is for annotations and flows.
There are a lot in the infra added for integration, and there are some other ways to implement integration solution, not just with `IntegrationFlow` and/or annotations.
Just saying disable messaging annotations processing should be enough.
That's exactly a purpose of this property, and everything else related to integration is left untouched.

src/reference/antora/modules/ROOT/pages/configuration/annotations.adoc
```diff
@@ -20,6 +20,9 @@ public class FooService {
 }
 ----
 
+NOTE: Starting with version 7.1, the `spring.integration.annotations.enable` environment property can be used to control method-level annotation processing.
```
artembilan: I think the "environment" is a bit confusing here.
It might be messed up with the ENV vars, which will work in Spring of course, but that is not a limitation.

Let's revise this sentence to something more context-specific:
```
the `spring.integration.annotations.enable` property in the Spring environment can be used to control method-level messaging annotations processing.
```

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -35,6 +35,9 @@ See xref:grpc.adoc[] for more information.
 The `MessageTransformingHandler.requiresReply` flag cannot be modified: an `UnsupportedOperationException` is thrown from the overridden `setRequiresReply()` method to indicate the transformer pattern cannot produce nulls for replies.
 See xref:transformer.adoc[] for more information.
 
+A new `spring.integration.annotations.enable` environment property is introduced to control method-level annotation processing. Setting this property to `false` optimizes startup performance for applications that exclusively rely on `IntegrationFlow` definitions.
```
artembilan: According to my comment in the `annotations.adoc` this sentence should be fixed as well.
More over, no need to say too much in this bullet-style `what's new` chapter.
If you want to know more, follow the provided link.

spring-integration-core/src/test/java/org/springframework/integration/config/IntegrationRegistrarTests.java
```diff
@@ -0,0 +1,162 @@
...
+			assertDoesNotContainsBean(context, Introspector.decapitalize(MessagingAnnotationBeanPostProcessor.class.getName()));
+
+			assertDoesNotContainsBean(context, serviceActivator);
+			assertDoesNotContainsBean(context, serviceActivatorHandler);
+
+		}
+	}
+
+	@RetryingTest(10)
+	void testEnableVersusDisablePerformance() {
```
artembilan: I think we don't need this test at all.
It does not verify anything and just prove a premise of the request, which simply could be done outside of the project as a regular issue handling process.

spring-integration-core/src/test/java/org/springframework/integration/config/IntegrationRegistrarTests.java
```diff
@@ -0,0 +1,162 @@
...
+
+			disabledStartupTime = stopWatch.lastTaskInfo().getTimeMillis();
+		}
+
+		Assertions.assertThat(enabledStartupTime).isGreaterThan(disabledStartupTime);
+	}
+
+	@Configuration
+	@EnableIntegration
+	@Import(ServiceActivatorComponent.class)
```
artembilan: This feels like awkward pattern for just regular bean-style class.
In most cases the `@Import` is for `@Configuration` classes from user perspective.
Consider to declare this `ServiceActivatorComponent` as a `@Bean`.

spring-integration-core/src/test/java/org/springframework/integration/config/IntegrationRegistrarTests.java
```diff
@@ -0,0 +1,162 @@
...
+	static class Config {
+
+		@Bean
+		DirectChannel inputChannel() {
```
artembilan: We don't need to declare this bean explicitly: the `@ServiceActivator` does the trick for us to create that bean on the fly.
And that is what you should prove in those tests.
So, if no annotation post-processor, then even no bean for `inputChannel`.

Makes sense?

artembilan (changes requested):

spring-integration-core/src/test/java/org/springframework/integration/config/IntegrationRegistrarTests.java
mjd507: totally make sense !

mjd507 (commented):
The scope is now much more precise and focused,  thanks for the feedback.

artembilan (approved):
LGTM!
Will merge in the morning from laptop.

Thank you!

src/reference/antora/modules/ROOT/pages/configuration/annotations.adoc
```diff
@@ -20,6 +20,8 @@ public class FooService {
 }
 ----
 
+NOTE: Starting with version 7.1, the `spring.integration.annotations.enable` property in the Spring environment can be used to control method-level annotations processing. By default, this property is `true`. It can be set to `false` to disable messaging annotations processing.
```
artembilan: Looks like you still have missed to follow this rule: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

Not a big deal: fixing on merge.

Thanks
```suggestion
NOTE: Starting with version 7.1, the `spring.integration.annotations.enable` property in the Spring environment can be used to control method-level annotations processing. 
By default, this property is `true`. 
It can be set to `false` to disable messaging annotations processing.
```

PR #11043, opened 2026-06-03
GH-11041: Revise cache in the `JdbcLockRegistry`

### LINKED ISSUE #11041: Prevent eviction of held locks from JDBC registry
The logic with:
```
				protected boolean removeEldestEntry(Entry<String, JdbcLock> eldest) {
					return size() > JdbcLockRegistry.this.cacheCapacity;
				}
```
may lead to the case when still held lock is evicted.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/11041

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -185,22 +185,9 @@ public class JdbcLockRegistry implements ExpirableLockRegistry<DistributedLock>,
 
 	@Override
 	public void renewLock(Object lockKey, Duration customTtl) {
-		Assert.isInstanceOf(String.class, lockKey);
-		String path = pathFor((String) lockKey);
-		JdbcLock jdbcLock;
-		this.lock.lock();
-		try {
-			jdbcLock = this.locks.get(path);
-		}
-		finally {
-			this.lock.unlock();
-		}
-
-		if (jdbcLock == null) {
-			throw new IllegalStateException("Could not found mutex at " + path);
-		}
+		JdbcLock jdbcLock = (JdbcLock) obtain(lockKey);
 		if (!jdbcLock.renew(customTtl)) {
-			throw new IllegalStateException("Could not renew mutex at " + path);
+			throw new IllegalStateException("Could not renew lock " + lockKey);
 		}
 	}
 
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -30,6 +30,7 @@ import java.util.concurrent.locks.ReentrantLock;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.dao.DataAccessResourceFailureException;
 import org.springframework.dao.TransientDataAccessException;
+import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.DistributedLock;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.integration.support.locks.RenewableLockRegistry;
@@ -67,7 +68,7 @@ public class JdbcLockRegistry implements ExpirableLockRegistry<DistributedLock>,
 
 	private static final int DEFAULT_IDLE = 100;
 
-	private static final int DEFAULT_CAPACITY = 100_000;
+	private static final int DEFAULT_CAPACITY = 256;
 
 	private final Lock lock = new ReentrantLock();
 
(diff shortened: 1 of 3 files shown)


spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -131,11 +134,15 @@ public void setIdleBetweenTries(Duration idleBetweenTries) {
...
-	 * @param cacheCapacity The capacity of cached lock, (default 100_000).
+	 * @param cacheCapacity The capacity of cached lock, (default 256 locks).
 	 * @since 5.5.6
+	 * @see DefaultLockRegistry
 	 */
 	public void setCacheCapacity(int cacheCapacity) {
 		this.cacheCapacity = cacheCapacity;
+		// Find the highest power of 2 for (n + 1), then subtract 1
+		int mask = Integer.highestOneBit(cacheCapacity + 1) - 1;
+		this.defaultLockRegistry = new DefaultLockRegistry(mask);
```
cppwfs, line 145: Question:
With this change the chances of a collision are higher and if the `ttl` is set to a long period of time, this could cause a large grab of memory, thus negating the original eviction policy.   i.e. a possible OOME.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
cppwfs: This message needs to be updated. It states, "Could not found mutex at ...".    It should state, "Could not find mutex at".

cppwfs: if a held lock is evicted from the `locks` map, a subsequent call to `registry.renewLock("lockKey")` will throw an `IllegalStateException("Could not found mutex")` because it expects the lock to still be in the map.

Should we update renewLock to use obtain(lockKey) instead of locks.get(path)? This allows renewLock to safely reconstruct the lock state and succeed even if the lock was evicted.

cppwfs, line 194: Not used and can be removed.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -131,11 +134,15 @@ public void setIdleBetweenTries(Duration idleBetweenTries) {
 
 	/**
 	 * Set the capacity of cached locks.
-	 * @param cacheCapacity The capacity of cached lock, (default 100_000).
+	 * @param cacheCapacity The capacity of cached lock, (default 256 locks).
```
cppwfs, line 137: Should we update `lock-registry.adoc` around line 47, that, "Starting with version 7.1 the default pool capacity is `256` locks."   ?

cppwfs (changes requested):
I love the idea!   I've added some questions and comments.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: I'm not sure in your statement.
Anything evicted, but not unlocked is still in the memory, and it doesn't matter what size of the cache we have.
The object is there in the memory whether its reference in the cache or not.
More over, with this change we got some memory improvement since `ReentrantLock` instances are reused from the pool.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: I think this is a great catch and it applies even with the current behavior. 
Will fix it respectively with the this change set.
Thank you!

artembilan: This is not correct request here.
We have to back-port the fix down to `6.5.x`.
This API was introduced a bit later and removing it might cause some back-port problems.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: Not correct.
We treat it as a bug and will back-port down to `6.5.x`.
I even mentioned that in the commit message.

But request for the doc is a great catch.

Thank you!

artembilan: Well, apparently we don't need to change that docs.
It is already pretty vague and self-explanatory:
```
Starting with version 5.5.6, the `JdbcLockRegistry` is support automatically clean-up cache for JdbcLock in `JdbcLockRegistry.locks` via `JdbcLockRegistry.setCacheCapacity()`.
See its Javadocs for more information.
```

cppwfs (approved):
LGTM !


Great work!

PR #11045, opened 2026-06-04
GH-11040: Hide exception in gRPC error message

### LINKED ISSUE #11040: Do not propage server error to the client from the `GrpcInboundGateway`
The current code is:
```
responseObserver.onError(toGrpcStatusException(t))
...
private static StatusRuntimeException toGrpcStatusException(Throwable throwable) {
	return toGrpcStatusException(throwable, throwable.getMessage());
}
```
Which sends back to the client an internal server error.

Consider just put something like `Internal Server Error`, and see if original should be logged under `error()`, so it is not get lost in between.

cppwfs:
Prevent internal exceptions from leaking in gRPC error responses.
- Log the actual exception locally instead of sending it over the wire.
- Return a generic internal server error for unhandled exceptions.
- Update GrpcInboundGatewayTests to verify the new error handling.

Fixes: #11040

### DIFF
--- a/spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
+++ b/spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
@@ -64,6 +64,7 @@ import org.springframework.util.StringUtils;
  * Such information can be used, for example, in downstream flow for routing.
  *
  * @author Artem Bilan
+ * @author Glenn Renfro
  *
  * @since 7.1
  */
@@ -237,12 +238,13 @@ public class GrpcInboundGateway extends MessagingGatewaySupport implements Binda
 				.map(Message::getPayload);
 	}
 
-	private static StatusRuntimeException toGrpcStatusException(Throwable throwable) {
-		return toGrpcStatusException(throwable, throwable.getMessage());
+	private StatusRuntimeException toGrpcStatusException(Throwable throwable) {
+		return toGrpcStatusException(throwable, "Internal Server Error");
 	}
 
-	private static StatusRuntimeException toGrpcStatusException(Throwable throwable, @Nullable String description) {
-		return Status.fromThrowable(throwable)
+	private StatusRuntimeException toGrpcStatusException(Throwable throwable, @Nullable String description) {
+		logger.error(throwable, description == null ? "Internal Server Error" : description);
+		return Status.fromCode(Status.Code.INTERNAL)
 				.withDescription(description)
 				.asRuntimeException();
 	}
--- a/spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
+++ b/spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
@@ -239,13 +239,17 @@ public class GrpcInboundGateway extends MessagingGatewaySupport implements Binda
 	}
 
 	private StatusRuntimeException toGrpcStatusException(Throwable throwable) {
-		return toGrpcStatusException(throwable, "Internal Server Error");
+		return toGrpcStatusException(throwable, throwable.getMessage());
 	}
 
 	private StatusRuntimeException toGrpcStatusException(Throwable throwable, @Nullable String description) {
-		logger.error(throwable, description == null ? "Internal Server Error" : description);
-		return Status.fromCode(Status.Code.INTERNAL)
-				.withDescription(description)
+		Status status = Status.fromThrowable(throwable);
+		String statusDescription = (description != null) ? description : status.getDescription();
+		if (status.getCode().equals(Status.UNKNOWN.getCode()) || statusDescription == null) {
+			statusDescription = "Internal Server Error";
+		}
+		logger.error(throwable, statusDescription);
+		return status.withDescription(statusDescription)
 				.asRuntimeException();
 	}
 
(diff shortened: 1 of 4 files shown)


spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
```diff
@@ -237,12 +238,13 @@ private Mono<?> sendRequestAndProduceReply(MethodDescriptor<?, ?> serviceMethod,
...
-		return toGrpcStatusException(throwable, throwable.getMessage());
+	private StatusRuntimeException toGrpcStatusException(Throwable throwable) {
+		return toGrpcStatusException(throwable, "Internal Server Error");
 	}
 
-	private static StatusRuntimeException toGrpcStatusException(Throwable throwable, @Nullable String description) {
-		return Status.fromThrowable(throwable)
+	private StatusRuntimeException toGrpcStatusException(Throwable throwable, @Nullable String description) {
+		logger.error(throwable, description == null ? "Internal Server Error" : description);
+		return Status.fromCode(Status.Code.INTERNAL)
```
artembilan: I think this is not correct.
The application may through those gRPC-specific exceptions, so using always `Status.Code.INTERNAL` is not OK.
I think we need to fallback to the `Internal Server Error` message when `Status` is `UNKNOWN` after ` Status.fromThrowable(throwable)`.
This way we would give end-users opportunity to still emit the whole exception message when they throw it as a `StatusException`.

Like in our test. We deliberately made a choice for that specific status:
```
											.transform(p -> {
												throw Status.UNAVAILABLE.withDescription("intentional")
														.asRuntimeException();
											}))
```

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/inbound/GrpcInboundGatewayTests.java
```diff
@@ -151,14 +161,36 @@ void bidiStreaming() throws InterruptedException {
...
-				.satisfies(e -> {
-					assertThat(e.getStatus().getCode()).isEqualTo(Status.Code.UNAVAILABLE);
-					assertThat(e.getStatus().getDescription())
-							.contains("Failed to transform Message in bean " +
-									"'grpcIntegrationFlow.subFlow#4.method-invoking-transformer#1'");
-				});
+		GrpcInboundGateway gateway = this.applicationContext.getBean(GrpcInboundGateway.class);
+		DirectFieldAccessor accessor = new DirectFieldAccessor(gateway);
+		LogAccessor originalLog = (LogAccessor) accessor.getPropertyValue("logger");
+		LogAccessor spyLog = spy(originalLog);
```
artembilan: I want this test back, and a new one when we throw arbitrary exception and the test verify for the `UNKNOWN` and `Internal Server Error`.
I don't want any spies on logging: doesn't prove anything.

artembilan (changes requested):
I think we also need to look into the docs and explain a bit the logic we expose for those server thrown errors.

Thanks

spring-integration-grpc/src/main/java/org/springframework/integration/grpc/inbound/GrpcInboundGateway.java
```diff
@@ -237,13 +238,18 @@ private Mono<?> sendRequestAndProduceReply(MethodDescriptor<?, ?> serviceMethod,
...
 		return toGrpcStatusException(throwable, throwable.getMessage());
 	}
 
-	private static StatusRuntimeException toGrpcStatusException(Throwable throwable, @Nullable String description) {
-		return Status.fromThrowable(throwable)
-				.withDescription(description)
+	private StatusRuntimeException toGrpcStatusException(Throwable throwable, @Nullable String description) {
+		Status status = Status.fromThrowable(throwable);
+		String statusDescription = (description != null) ? description : status.getDescription();
+		if (status.getCode().equals(Status.UNKNOWN.getCode()) || statusDescription == null) {
```
artembilan: I think we can use `Status.Code.UNKNOWN` instead of extra method calls.

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/grpc.adoc
```diff
@@ -153,6 +153,14 @@ private static HelloReply newHelloReply(String message) {
 The routing is done on the `GrpcHeaders.SERVICE_METHOD` header populated by the `GrpcInboundGateway`.
 All the downstream transformer business methods are type-safe in regard to gRPC messages for the `TestHelloWorldGrpc.TestHelloWorldImplBase` service.
 
+[[grpc-inbound-error-handling]]
+=== Error Handling
+
+If an error occurs during downstream processing, the `GrpcInboundGateway` attempts to map the resulting `Throwable` to a gRPC `StatusRuntimeException`.
```
artembilan: or `StatusException`?
According to the `Status.fromThrowable()` that does not mean that root exception has to be such.
The logic in the app may decide to throw just plain `StatusException` and then wrap it or so.

spring-integration-grpc/src/test/proto/test_hello.proto
```diff
@@ -27,4 +27,8 @@ service TestHelloWorld {
   // Fail with error
   rpc ErrorOnHello(HelloRequest) returns (HelloReply) {}
 
+  // Fail with unknown error
+  rpc UnknownErrorOnHello(HelloRequest) returns (HelloReply) {}
```
artembilan: I wonder if we really need a separate gRPC contract for this our new logic.
Perhaps we can model a logic of existing `transform()` to emit different errors according to the inbound payload?
Like simple `if..else` if `switch` does not yet.

artembilan (changes requested):

spring-integration-grpc/src/test/java/org/springframework/integration/grpc/inbound/GrpcInboundGatewayTests.java
```diff
@@ -267,6 +259,15 @@ private HelloReply requestReply(HelloRequest helloRequest) {
 			return newHelloReply("Hello " + helloRequest.getName());
 		}
 
+		private HelloReply errorReply(HelloRequest helloRequest) {
+			if (helloRequest.getName().equals("Error")) {
+				throw Status.UNAVAILABLE.withDescription("intentional")
+						.asRuntimeException();
+			}
+			throw Status.UNKNOWN.withDescription("")
```
artembilan: Well, I think what we want here is something like `throw new RuntimeException("some service error");`
And in the end we ensure that this `some` does not appear in the client stack trace.

artembilan (changes requested):

PR #11046, opened 2026-06-04
GH-11042: Fix race condition in the ControlBusCommandRegistry

### LINKED ISSUE #11042: The `ControlBusCommandRegistry.controlBusCommands` must be a `ConcurrentHashMap`

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/11042

The `ControlBusCommandRegistry` can be accessed (and populated) concurrently. Therefore, it is better to protect its internal from the `ConcurrentModificationException`

* Use `ConcurrentHashMap` and `synchronized` in the `ControlBusCommandRegistry` whenever concurrent access is possible.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/management/ControlBusCommandRegistry.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/management/ControlBusCommandRegistry.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -71,7 +72,7 @@ public class ControlBusCommandRegistry
 
 	private static final ControlBusMethodFilter CONTROL_BUS_METHOD_FILTER = new ControlBusMethodFilter();
 
-	private final Map<String, Map<CommandMethod, Expression>> controlBusCommands = new HashMap<>();
+	private final Map<String, Map<CommandMethod, Expression>> controlBusCommands = new ConcurrentHashMap<>();
 
 	private boolean eagerInitialization;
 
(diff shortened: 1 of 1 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/management/ControlBusCommandRegistry.java
```diff
@@ -141,14 +142,16 @@ public void postProcessBeforeDestruction(Object bean, String beanName) throws Be
...
 		Map<String, Map<CommandMethod, String>> commands = new HashMap<>(this.controlBusCommands.size());
-		for (Map.Entry<String, Map<CommandMethod, Expression>> beanEntry : this.controlBusCommands.entrySet()) {
-			Map<CommandMethod, String> commandEntries =
-					beanEntry.getValue()
-							.entrySet()
-							.stream()
-							.collect(Collectors.toMap(Map.Entry::getKey,
-									(commandEntry) -> commandEntry.getValue().getExpressionString()));
-			commands.put(beanEntry.getKey(), commandEntries);
+		synchronized (this.controlBusCommands) {
```
cppwfs: Why are we synchronizing on the `ConcurrentHashMap` `controlBusCommands`?  

If it is required why are we not synchronizing on the `controlBusCommands`  on line 201?

cppwfs (changes requested):
Good catch on using ConcurrentHashMap!

Have a question or two on the synchronization.

spring-integration-core/src/main/java/org/springframework/integration/support/management/ControlBusCommandRegistry.java
artembilan: Yep!
You are right: we don't `synchronized (this.controlBusCommands)` since we do not do any modification during iteration and we deal with the `ConcurrentHashMap`, not old `SynchronizedMap` 😄

cppwfs (approved):
Thank you for the changes!

This looks great!

PR #11058, opened 2026-06-08
GH-11049: Block unlock until key is removed

### LINKED ISSUE #11049: The `RedisLockRegistry` unlock race condition on the interrupted thread
The current logic is like this:
```
		@Override
		public final void unlock() {
			if (!this.localLock.isHeldByCurrentThread()) {
				throw new IllegalStateException("You do not own lock at " + this.lockKey);
			}
			if (this.localLock.getHoldCount() > 1) {
				this.localLock.unlock();
				return;
			}
			try {
				if (Thread.currentThread().isInterrupted()) {
					RedisLockRegistry.this.executor.execute(this::removeLockKey);
				}
				else {
					removeLockKey();
				}

				if (LOGGER.isDebugEnabled()) {
					LOGGER.debug("Released lock; " + this);
				}
			}
			catch (Exception e) {
				ReflectionUtils.rethrowRuntimeException(e);
			}
			finally {
				this.localLock.unlock();
			}
		}
```
Where it is very visible that `this.localLock.unlock();` might be called before Redis key is removed in case of interrupted thread.

We cannot get rid of thread shifting since Redis driver does not allow to perform network operations on the interrupted thread. And we also cannot simply rely on the `CountDownLatch` or `Future` since they throw `InterruptedException` immediately in the access to them.

However there is "gotcha" in Java on the matter. 
We can do like this:
```
CountDownLatch latch = new CountDownLatch(1);

this.executor.execute(() -> {
            try {
                removeLockKey();
            } finally {
                latch.countDown();
            }
});
        
boolean interrupted = false;
try {
            while (true) {
                try {
                    latch.await();
                    break; 
                } catch (InterruptedException e) {
                    interrupted = true; 
                }
            }
} finally {
            if (interrupted) {
            }
}
```
The JVM implicitly CLEARS the thread's interrupt flag when throwing an `InterruptedException`. We record it here with the `interrupted` flag, loop back to the `latch.await()` and restore `Thread.currentThread().interrupt()` state.

cppwfs:
- Block the unlock operation until the lock key is actually removed.
- Prevent early return when the current thread is interrupted.
- Propagate any exceptions from the removal task to the caller.
- Add a test to verify blocking behavior on interrupted threads.

Fixes: https://github.com/spring-projects/spring-integration/issues/11049

### DIFF
--- a/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
+++ b/spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
@@ -1121,10 +1121,11 @@ class RedisLockRegistryTests implements RedisContainerTest {
 		CountDownLatch removalStarted = new CountDownLatch(1);
 		CountDownLatch allowRemovalToFinish = new CountDownLatch(1);
 
+		// Create an executor to freeze the removal process
 		registry.setExecutor(runnable -> new Thread(() -> {
 			try {
 				removalStarted.countDown();
-				allowRemovalToFinish.await(); // Hold the Redis delete hostage
+				allowRemovalToFinish.await();
 				runnable.run();
 			}
 			catch (InterruptedException e) {
@@ -1134,6 +1135,7 @@ class RedisLockRegistryTests implements RedisContainerTest {
 
 		Lock lock = registry.obtain("test");
 
+		// Spawn worker thread with error logging
 		Thread testThread = new Thread(() -> {
 			lock.lock();
 			try {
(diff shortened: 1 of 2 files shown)


spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -1114,6 +1114,44 @@ void noSecondLockOnEviction() throws InterruptedException {
...
+				lock.unlock();
+			}
+		});
+
+		testThread.start();
+		testThread.join();
+
+		// UNMODIFIED: FAILS (Expected: true, Actual: false) because unlock() leaked out early.
+		// FIXED: PASSES (true) because the latch forced the thread to wait for the background task.
+		assertThat(backgroundTaskFinished.get()).isTrue();
```
artembilan: I know we don't have too much time before release and we may merge this as is, however I think there is still a flaw in this test and that flag could be true due to race condition between threads and short timeout.
I don't believe in time-based tests and this one does not reflect reality of the fix.
So, according to the report we could unlock local lock, but that delete on the Redis might not happen yet.
The system assumes that lock is unlocked ad let us lock in a fresh instance.

With the fix we ensure that local unlock does not happen until the answer from remove.
Therefore it feels like blocking in the removal thread until some testing barrier (`CountDownLatch`?) and attempt to lock a fresh instance in the main thread is failing.
Just because that interrupted thread is blocked on the unlock operation.
Without the fix it looks like would be possible because we had let a local lock to be unlocked without waiting for removal result.

I won't mind if you have any objections.

Thanks

artembilan (changes requested):

spring-integration-redis/src/test/java/org/springframework/integration/redis/util/RedisLockRegistryTests.java
```diff
@@ -1114,6 +1114,62 @@ void noSecondLockOnEviction() throws InterruptedException {
...
+
+		// Wait until the background thread reaches the removal phase
+		boolean started = removalStarted.await(5, TimeUnit.SECONDS);
+		assertThat(started).isTrue();
+
+		// The unlocking thread is still blocked
+		assertThat(testThread.isAlive()).isTrue();
+
+		// Try to acquire the lock from a fresh registry instance.
+		RedisLockRegistry freshRegistry = new RedisLockRegistry(redisConnectionFactory, this.registryKey);
```
artembilan: OK. The solution in the test is very close to what I have in mind, but still not what we'd like to verify for the problem in the report.
That is really fully OK for another registry instance to be able to lock when we got a removal from Redis, even if unlock in the previous registry instance has not been finished yet.
There is just no way over the network to have that consensus: the removal from shared store is enough indicator.

The story in the report is about the same registry, when we have not finished removal in shared store yet, but already indicating for local users that we have been unlocked.
Therefore, a new registry here is not the direction we would like to verify.

I think the logic of the test should allow the removal to be performed, but must not exit from that thread yet.
The point is that unlocking interrupted thread must be blocked until removal thread signals.

artembilan (changes requested):

artembilan (approved):
LGTM.
Will merge when PR build is green.

Thanks

PR #11112, opened 2026-06-15
IntegrationFlowBuilder falls back to outputChannelName

### LINKED ISSUE #11107: Specifying channel name for inbound JMS components in Java DSL causes startup failure due to no BeanFactory

7.0.5.RELEASE

**Describe the bug**

For inbound JMS adapters and the gateway you can specify a (request) channel directly, or just a channel name. The former works, the latter crashes.

**To Reproduce**

Create an IntegrationFlow using Java DSL and a messageDrivenChannelAdapter (for example) that specifies an outputChannel with a channel name. The application will not start, with the relevant stacktrace being:

```
Caused by: java.lang.IllegalArgumentException: 'beanFactory' must not be null
	at org.springframework.util.Assert.notNull(Assert.java:182) ~[spring-core-7.0.7.jar:7.0.7]
	at org.springframework.integration.support.channel.ChannelResolverUtils.getChannelResolver(ChannelResolverUtils.java:49) ~[spring-integration-core-7.0.5.jar:7.0.5]
	at org.springframework.integration.context.IntegrationObjectSupport.getChannelResolver(IntegrationObjectSupport.java:280) ~[spring-integration-core-7.0.5.jar:7.0.5]
	at org.springframework.integration.endpoint.MessageProducerSupport.getOutputChannel(MessageProducerSupport.java:108) ~[spring-integration-core-7.0.5.jar:7.0.5]
	at org.springframework.integration.dsl.IntegrationFlow.from(IntegrationFlow.java:342) ~[spring-integration-core-7.0.5.jar:7.0.5]
	at org.springframework.integration.dsl.IntegrationFlow.from(IntegrationFlow.java:324) ~[spring-integration-core-7.0.5.jar:7.0.5]
```

**Expected behavior**

The application should start up normally.

cppwfs:
- If outputChannel is not available, check if channelName is available.
- If neither are available create a direct channel.

Fixes: https://github.com/spring-projects/spring-integration/issues/11107

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
@@ -341,13 +341,7 @@ public interface IntegrationFlow {
 			@Nullable IntegrationFlowBuilder integrationFlowBuilderArg) {
 
 		IntegrationFlowBuilder integrationFlowBuilder = integrationFlowBuilderArg;
-		MessageChannel outputChannel;
-		try {
-			outputChannel = messageProducer.getOutputChannel();
-		}
-		catch (IllegalArgumentException argumentException) {
-			outputChannel = null;
-		}
+		MessageChannel outputChannel = messageProducer.getOutputChannel();
 
 		if (outputChannel == null) {
 			String outputChannelName = messageProducer.getOutputChannelName();
@@ -490,13 +484,7 @@ public interface IntegrationFlow {
 			@Nullable IntegrationFlowBuilder integrationFlowBuilderArg) {
 
 		IntegrationFlowBuilder integrationFlowBuilder = integrationFlowBuilderArg;
-		MessageChannel outputChannel;
-		try {
-			outputChannel = inboundGateway.getRequestChannel();
-		}
-		catch (IllegalArgumentException argumentException) {
-			outputChannel = null;
-		}
+		MessageChannel outputChannel = inboundGateway.getRequestChannel();
 
 		if (outputChannel == null) {
 			String requestChannelName = inboundGateway.getRequestChannelName();
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
@@ -41,6 +41,7 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * The main Integration DSL abstraction.
@@ -91,6 +92,7 @@ import org.springframework.util.Assert;
  * @author Gary Russell
  * @author Oleg Zhurakousky
  * @author Artem Vozhdayenko
+ * @author Glenn Renfro
  *
  * @since 5.0
  *
@@ -339,18 +341,24 @@ public interface IntegrationFlow {
 			@Nullable IntegrationFlowBuilder integrationFlowBuilderArg) {
 
 		IntegrationFlowBuilder integrationFlowBuilder = integrationFlowBuilderArg;
-		MessageChannel outputChannel = messageProducer.getOutputChannel();
-		if (outputChannel == null) {
+		MessageChannel outputChannel = null;
+		String outputChannelName = messageProducer.getOutputChannelName();
+		try {
+			outputChannel = messageProducer.getOutputChannel();
+		}
+		catch (IllegalArgumentException argumentException) {
+			if (argumentException.getMessage() != null && !argumentException.getMessage()
+					.contentEquals("'beanFactory' must not be null")) {
+				throw argumentException;
+			}
+		}
+
+		if (!StringUtils.hasText(outputChannelName) && outputChannel == null) {
 			outputChannel = new DirectChannel();
 			messageProducer.setOutputChannel(outputChannel);
 		}
-		if (integrationFlowBuilder == null) {
-			integrationFlowBuilder = from(outputChannel);
-		}
-		else {
-			integrationFlowBuilder.channel(outputChannel);
-		}
-		return integrationFlowBuilder.addComponent(messageProducer);
+		return updateOutputChannelForBuilder(integrationFlowBuilder, outputChannelName, outputChannel)
+				.addComponent(messageProducer);
 	}
 
 	/**
@@ -474,18 +482,49 @@ public interface IntegrationFlow {
 			@Nullable IntegrationFlowBuilder integrationFlowBuilderArg) {
 
 		IntegrationFlowBuilder integrationFlowBuilder = integrationFlowBuilderArg;
-		MessageChannel outputChannel = inboundGateway.getRequestChannel();
-		if (outputChannel == null) {
+		MessageChannel outputChannel = null;
+		String requestChannelName = inboundGateway.getRequestChannelName();
+
+		try {
+			outputChannel = inboundGateway.getRequestChannel();
+		}
+		catch (IllegalArgumentException argumentException) {
+			if (argumentException.getMessage() != null && !argumentException.getMessage()
+					.contentEquals("'beanFactory' must not be null")) {
+				throw argumentException;
+			}
+		}
+
+		if (!StringUtils.hasText(requestChannelName) && outputChannel == null) {
 			outputChannel = new DirectChannel();
 			inboundGateway.setRequestChannel(outputChannel);
 		}
+		return updateOutputChannelForBuilder(integrationFlowBuilder, requestChannelName, outputChannel)
+				.addComponent(inboundGateway);
+	}
+
+	private static IntegrationFlowBuilder updateOutputChannelForBuilder(
+			@Nullable IntegrationFlowBuilder integrationFlowBuilder, @Nullable String channelName,
+			@Nullable MessageChannel outputChannel) {
 		if (integrationFlowBuilder == null) {
-			integrationFlowBuilder = from(outputChannel);
+			if (outputChannel != null) {
+				integrationFlowBuilder = from(outputChannel);
+			}
+			else {
+				Assert.notNull(channelName, "'channelName' must not be null");
+				integrationFlowBuilder = from(channelName);
+			}
 		}
 		else {
-			integrationFlowBuilder.channel(outputChannel);
+			if (outputChannel != null) {
+				integrationFlowBuilder.channel(outputChannel);
+			}
+			else {
+				Assert.notNull(channelName, "'channelName' must not be null");
+				integrationFlowBuilder.channel(channelName);
+			}
 		}
-		return integrationFlowBuilder.addComponent(inboundGateway);
+		return integrationFlowBuilder;
 	}
 
 	@Nullable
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -101,6 +102,16 @@ public void setOutputChannelName(String outputChannelName) {
 		this.outputChannelName = outputChannelName;
 	}
 
+	/**
+	 * Return the output channel name.
```
artembilan, line 106: `... or null if not provided.`

spring-integration-core/src/test/java/org/springframework/integration/dsl/flows/IntegrationFlowTests.java
```diff
@@ -565,6 +569,72 @@ public void cleanUpList() {
...
+
+	@Test
+	public void testMessageProducerSpecWithOutputNullChannel() {
+		IntegrationFlow flow = IntegrationFlow.from(new TestMessageProducerSpec()).get();
+		assertThat(flow).isNotNull();
+	}
+
+	@Test
+	public void testWithChannelName() {
+		IntegrationFlow flow = IntegrationFlow.from("someChannelName").bridge().get();
```
artembilan: This test is redundant: the `from(String messageChannelName)` is covered in many other tests.

spring-integration-core/src/test/java/org/springframework/integration/dsl/flows/IntegrationFlowTests.java
```diff
@@ -565,6 +569,72 @@ public void cleanUpList() {
 		outputStringList.clear();
 	}
 
+	@Test
+	public void testMessageProducerSpecWithOutputChannelName() {
+		IntegrationFlow flow = IntegrationFlow.from(new TestMessageProducerSpec()
+				.outputChannel("someChannelName")).get();
+		assertThat(flow).isNotNull();
```
artembilan: I don't think this is enough.
would be great to verify that `outputChannel` is propagated to the `DirectChannel` in the final `StandardIntegrationFlow` instance.
The `IntegrationFlowBuilder.get()` is to never return null.
It might fail as in case of the issue we are fixing here, but that is a different story.
The goal I guess is to have a flow with that `outputChannel` propagated as it is advertised by that API.

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
```diff
@@ -339,18 +341,24 @@ private static IntegrationFlowBuilder from(MessageProducerSupport messageProduce
...
+		MessageChannel outputChannel = null;
+		String outputChannelName = messageProducer.getOutputChannelName();
+		try {
+			outputChannel = messageProducer.getOutputChannel();
 		}
-		if (integrationFlowBuilder == null) {
-			integrationFlowBuilder = from(outputChannel);
+		catch (IllegalArgumentException argumentException) {
+			if (argumentException.getMessage() != null && !argumentException.getMessage()
+					.contentEquals("'beanFactory' must not be null")) {
```
artembilan: This is a bit fragile to  rely on the exception message in the logic.

I wonder if we can bite a bullet and change the `messageProducer.getOutputChannel()` logic to return `null` when no `BeanFactory` instead of throwing an exception.

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
```diff
@@ -474,18 +482,49 @@ private static IntegrationFlowBuilder from(MessagingGatewaySupport inboundGatewa
...
 			inboundGateway.setRequestChannel(outputChannel);
 		}
+		return updateOutputChannelForBuilder(integrationFlowBuilder, requestChannelName, outputChannel)
+				.addComponent(inboundGateway);
+	}
+
+	private static IntegrationFlowBuilder updateOutputChannelForBuilder(
+			@Nullable IntegrationFlowBuilder integrationFlowBuilder, @Nullable String channelName,
+			@Nullable MessageChannel outputChannel) {
 		if (integrationFlowBuilder == null) {
```
artembilan, line 500: Blank line after multi-line method header.

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
```diff
@@ -474,18 +482,49 @@ private static IntegrationFlowBuilder from(MessagingGatewaySupport inboundGatewa
...
+
+		if (!StringUtils.hasText(requestChannelName) && outputChannel == null) {
 			outputChannel = new DirectChannel();
 			inboundGateway.setRequestChannel(outputChannel);
 		}
+		return updateOutputChannelForBuilder(integrationFlowBuilder, requestChannelName, outputChannel)
+				.addComponent(inboundGateway);
+	}
+
+	private static IntegrationFlowBuilder updateOutputChannelForBuilder(
```
artembilan: According to the logic of this method we do not really update, but add channel to the flow or ... create a new one.

How about `progateProducerOutputChannelToFlow` ?

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -101,6 +102,16 @@ public void setOutputChannelName(String outputChannelName) {
...
+	 * @return the output channel name.
+	 * @since 6.5
```
artembilan: We need to be more specific with the version.
In this case it is `6.5.10`.
The fact that this might be never released as OSS is a different story.

artembilan (changes requested):
Looks cool!

Thank you!

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
cppwfs: So if there is no `outputChannelName` or no `BeanFactory` is found then we return a `null`.  Correct?

Do we want to update the `from` methods so that they can return a `@Nullable` value?

artembilan: I'm not sure what you mean.
Even right now both `outputChannelName` and `outputChannel` could be `null`.
Do you make `from()` as `@Nullable`?
What I want here is to avoid throwing exception and lose performance.
We deliberately know here that there is no `BeanFactory` in the `MessageProducer` when we call its `getOutputChannel()` here.
So, getting null for the `outputChannel` is the trigger for that `new DirectChannel()`.
We check for `outputChannelName` and if it is null we call `getOutputChannel()`.
The logic of that `getOutputChannel()` should be adjusted to deal with null when no `BeanFactory`.

Could you, please, elaborate about `from` and `@Nullable`?

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
```diff
@@ -339,18 +341,24 @@ private static IntegrationFlowBuilder from(MessageProducerSupport messageProduce
...
 		}
-		else {
-			integrationFlowBuilder.channel(outputChannel);
+
+		if (!StringUtils.hasText(outputChannelName) && outputChannel == null) {
+			outputChannel = new DirectChannel();
+			messageProducer.setOutputChannel(outputChannel);
 		}
-		return integrationFlowBuilder.addComponent(messageProducer);
+		return updateOutputChannelForBuilder(integrationFlowBuilder, outputChannelName, outputChannel)
```
artembilan: I think we need to look into `new MessageChannelReference(outputChannelName)` instead.
See that 
```
static IntegrationFlowBuilder from(String messageChannelName) {
		return from(new MessageChannelReference(messageChannelName));
	}
```

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
cppwfs: Understood.   That makes sense!

spring-integration-core/src/main/java/org/springframework/integration/dsl/IntegrationFlow.java
```diff
@@ -339,11 +341,25 @@ private static IntegrationFlowBuilder from(MessageProducerSupport messageProduce
 			@Nullable IntegrationFlowBuilder integrationFlowBuilderArg) {
 
 		IntegrationFlowBuilder integrationFlowBuilder = integrationFlowBuilderArg;
-		MessageChannel outputChannel = messageProducer.getOutputChannel();
+		MessageChannel outputChannel;
+		try {
+			outputChannel = messageProducer.getOutputChannel();
+		}
+		catch (IllegalArgumentException argumentException) {
```
artembilan: This is not what we have discussed.
I said it would be great to avoid an exception code flow control.
It would be better to check for `BeanFactory` in the `getOutputChannel()` and return null instead of throwing an exception.
Why do you still go with an exception, please?

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java
```diff
@@ -101,11 +102,23 @@ public void setOutputChannelName(String outputChannelName) {
...
 	@Override
 	public @Nullable MessageChannel getOutputChannel() {
-		String channelName = this.outputChannelName;
-		if (channelName != null) {
-			this.outputChannel = getChannelResolver().resolveDestination(channelName);
+		if (this.outputChannelName != null) {
+			if (getBeanFactory() == null) {
+				return null;
+			}
+			this.outputChannel = getChannelResolver().resolveDestination(this.outputChannelName);
```
artembilan: Please, bring the local variable back.
There is a potential race condition when we use `this.outputChannelName` concurrently.
So, it might be not null for the `if`, but it could be null at this call.

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -438,6 +449,9 @@ private void initializeIfNecessary() {
 	 */
 	public @Nullable MessageChannel getRequestChannel() {
 		if (this.requestChannel == null && this.requestChannelName != null) {
+			if (getBeanFactory() == null) {
+				return null;
+			}
 			this.requestChannel = getChannelResolver().resolveDestination(this.requestChannelName);
```
artembilan: Well, this one has to be fixed for the local variable as well.
Good to come back to old code for revision 😄

artembilan (changes requested):

spring-integration-core/src/main/java/org/springframework/integration/gateway/MessagingGatewaySupport.java
```diff
@@ -437,8 +448,13 @@ private void initializeIfNecessary() {
 	 * @since 4.2
 	 */
 	public @Nullable MessageChannel getRequestChannel() {
-		if (this.requestChannel == null && this.requestChannelName != null) {
-			this.requestChannel = getChannelResolver().resolveDestination(this.requestChannelName);
+		String requestChannelName = this.requestChannelName;
```
artembilan: We cannot use the local variable name exactly as a property name. It clashes not only on review

artembilan (changes requested):

PR #11115, opened 2026-06-15
Fix multibyte ack-header offset in `DatagramPacketMessageMapper`

uwezkhan:
Round-tripping a UDP message through `DatagramPacketMessageMapper` with a non-ASCII `ip_ackTo` value:

    ackTo "café:11111" -> payload ";ABCD"   (expected "ABCD")

`toMessage` strips the `ip_ackTo`/`id` prefix using `Matcher.end()`, which is a character index into the decoded header string, but applies it as a byte offset into the packet and as the payload length. For a multibyte charset (UTF-8 is the default) the character count is shorter than the byte count, so the copy starts inside the header and the framing shifts.

Before: payload `";ABCD"`, the body is offset by the stray UTF-8 byte and the separator leaks in.
After: the matched prefix is re-encoded with the configured charset to get its byte length, which drives both the source offset and the remaining payload length, so the payload is `"ABCD"`.

Trade-off: one extra `getBytes(charset)` over the matched header region per ack packet, only on the ack path; ASCII ack addresses behave exactly as before.

### DIFF
--- a/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/DatagramPacketMessageMapper.java
+++ b/spring-integration-ip/src/main/java/org/springframework/integration/ip/udp/DatagramPacketMessageMapper.java
@@ -66,6 +66,7 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Dave Syer
  * @author Artem Bilan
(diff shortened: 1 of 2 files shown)


artembilan:
Hey @uwezkhan !

PR #11122, opened 2026-06-16

### LINKED ISSUE #11121: SftpSession.listNames() throws SSH_FX_NO_SUCH_FILE for root directory wildcard paths
## Summary

**Bug introduced in:** 6.0.0

## Root Cause

## Steps to Reproduce

```java
try (Session<SftpClient.DirEntry> session = sessionFactory.getSession()) {
    String[] entries = session.listNames("/*");
}
```

## Actual Behavior

```
java.io.UncheckedIOException: java.lang.RuntimeException: SFTP error (SSH_FX_NO_SUCH_FILE): No such file or directory
    at org.apache.sshd.sftp.client.impl.SftpIterableDirEntry.iterator(SftpIterableDirEntry.java:66)
    at org.springframework.integration.sftp.session.SftpSession.doList(SftpSession.java:102)
    at org.springframework.integration.sftp.session.SftpSession.list(SftpSession.java:68)
    ...
Caused by: SFTP error (SSH_FX_NO_SUCH_FILE): No such file or directory
    at org.apache.sshd.sftp.client.impl.AbstractSftpClient.openDir(AbstractSftpClient.java:829)
```

## Expected Behavior

`listNames("/*")` should correctly parse `remoteDir = ""` (root) and `remoteFile = "*"`, and return the entries in the root directory.

## Proposed Fix

Change `lastIndex > 0` to `lastIndex >= 0` in `SftpSession.doList()` so that root-level paths (where `/` is at index 0) are handled correctly.

yangadam:
## Summary

`SftpSession.listNames("/*")` throws `SftpException: SSH_FX_NO_SUCH_FILE` because the path parsing in `SftpSession.list()` uses `lastIndex > 0` to detect the `/` separator. For root-level paths like `/*`, `lastIndexOf('/')` returns `0`, which is excluded by the `> 0` check. This causes the entire path `/*` to be treated as a literal directory name instead of being split into directory `/` and file pattern `*`.

## Changes

- Changed `lastIndex > 0` to `lastIndex >= 0` in `SftpSession.list()` so root directory paths are correctly parsed
- Added test `lsRootWildcard()` that reproduces the issue and verifies the fix

Fixes: gh-11121

Signed-off-by: Adam Yang <yangadam@users.noreply.github.com>

### DIFF
--- a/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
+++ b/spring-integration-sftp/src/main/java/org/springframework/integration/sftp/session/SftpSession.java
@@ -50,6 +50,7 @@ import org.springframework.util.StringUtils;
  * @author Artem Bilan
  * @author Christian Tzolov
  * @author Darryl Smith
+ * @author Adam Yang
  * @since 2.0
  */
 public class SftpSession implements Session<SftpClient.DirEntry> {
@@ -113,7 +114,10 @@ public class SftpSession implements Session<SftpClient.DirEntry> {
 		if (lastIndex > 0) {
 			remoteDir = remoteDir.substring(0, lastIndex);
 		}
-		String remoteFile = lastIndex > 0 ? remotePath.substring(lastIndex + 1) : null;
+		else if (lastIndex == 0) {
+			remoteDir = "/";
+		}
+		String remoteFile = lastIndex >= 0 ? remotePath.substring(lastIndex + 1) : null;
 		boolean isPattern = remoteFile != null && remoteFile.contains("*");
 
 		if (!isPattern && remoteFile != null) {
--- a/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/outbound/SftpOutboundTests.java
+++ b/spring-integration-sftp/src/test/java/org/springframework/integration/sftp/outbound/SftpOutboundTests.java
@@ -78,6 +78,7 @@ import static org.mockito.Mockito.when;
  * @author Artem Bilan
  * @author Darryl Smith
  * @author Glenn Renfro
+ * @author Adam Yang
  */
 public class SftpOutboundTests implements TestApplicationContextAware {
 
@@ -328,6 +329,10 @@ public class SftpOutboundTests implements TestApplicationContextAware {
 								.map((file) -> new SftpClient.DirEntry(file, file, new SftpClient.Attributes()))
 								.toList();
 				when(sftpClient.readDir("/remote-test-dir")).thenReturn(dirEntries);
+
+				SftpClient.Attributes dirAttributes = new SftpClient.Attributes();
+				dirAttributes.setPermissions(SftpConstants.S_IFDIR);
+				willReturn(dirAttributes).given(sftpClient).stat("/remote-test-dir/");
 
 				return SftpTestSessionFactory.createSftpSession(sftpClient);
 			}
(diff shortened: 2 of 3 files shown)


artembilan (changes requested):
Well, looks like you have not checked the fix locally before opening PR:
```
SftpOutboundTests > testFtpOutboundGatewayInsideChain() FAILED
    org.springframework.messaging.MessageHandlingException at SftpOutboundTests.java:192
        Caused by: org.springframework.messaging.MessagingException at SftpOutboundTests.java:192
            Caused by: java.lang.NullPointerException at SftpOutboundTests.java:192
```
Please, ensure you run `./gradlew "spring-integration-sftp:check` locally before pushing to PR.

Thanks

yangadam:
> Well, looks like you have not checked the fix locally before opening PR:
> 
> ```
> SftpOutboundTests > testFtpOutboundGatewayInsideChain() FAILED
>     org.springframework.messaging.MessageHandlingException at SftpOutboundTests.java:192
>         Caused by: org.springframework.messaging.MessagingException at SftpOutboundTests.java:192
>             Caused by: java.lang.NullPointerException at SftpOutboundTests.java:192
> ```
> 
> Please, ensure you run `./gradlew "spring-integration-sftp:check` locally before pushing to PR.
> 
> Thanks

Thanks for running the build. After debugging, I found that the failure was caused by the lack of mocking for the `stat` method, which previously was not called when the path was `/remote-test-dir/`.

I can add the mocking, but I need to confirm whether it is acceptable to call `stat` whenever the path is `/remote-test-dir/` or `/remote-test-dir/remote-sub-dir/`.

@artembilan

artembilan (changes requested):
> I need to confirm whether it is acceptable to call stat whenever the path is `/remote-test-dir/` or `/remote-test-dir/remote-sub-dir/`.

Correct. So, ask yourself if it is valid to divide `/remote-test-dir` into `/` and `remote-test-dir` as you do in your current fix.
You covered with `boolean isPattern = remoteFile != null && remoteFile.contains("*");`.
But that is not OK for a dir request.

I think we need just trim both leading and trailing `/` in the method you have modified instead.
Something like:
```
String remotePath = StringUtils.trimLeadingCharacter(StringUtils.trimTrailingCharacter(validPath, '/'), '/');
```
With that your `/*` will become `*` as you expect.

yangadam:
Hi @artembilan, thanks for taking the time to review and for the suggestion!

I looked into the trimming approach, but I think the split-then-stat pattern might actually be unavoidable here. A name like `remote-test-dir` could be either a file or a directory — we can never know beforehand without calling `stat`.

In fact, the existing code on main already does this for deeper paths. For example, `/remote-test-dir/remote-sub-dir/` goes through the same flow:

1. `trimTrailingCharacter` → `/remote-test-dir/remote-sub-dir`
2. `lastIndex = 21` (> 0), so `remoteDir = "/remote-test-dir"`, `remoteFile = "remote-sub-dir"`
3. `isPattern = false` → enters the `stat` block → discovers it's a directory → reassigns `remoteDir = remotePath`

My fix for `lastIndex == 0` simply makes `/remote-test-dir` follow the same existing pattern — the only difference is the nesting depth. In both cases, the last path component could be a file or a directory, and `stat` is how the code figures that out.

Would it be OK to keep this approach consistent with the existing behavior for now, and refactor the method later if we'd like to optimize it?

artembilan (changes requested):

Then we'll merge.

Thank you!

PR #11123, opened 2026-06-16

yangadam:
## Summary

## Changes

- Changed `lastIndex > 0` to `lastIndex >= 0` in `SftpSession.doList()` so root directory paths are correctly parsed
- Added test `lsRootWildCard()` that reproduces the issue and verifies the fix

This bug was introduced in 6.0.0, so the fix targets `6.0.x` for forward-merging through newer release branches.

Fixes: gh-11121


yangadam:
Closing — retargeting to 6.5.x since 6.0.x is locked.

PR #11124, opened 2026-06-16

yangadam:
## Summary

Bug introduced in 6.0.0.

## Changes

- Changed `lastIndex > 0` to `lastIndex >= 0` in `SftpSession.doList()` so root directory paths are correctly parsed
- Added test `lsRootWildcard()` that reproduces the issue and verifies the fix

Fixes: gh-11121


yangadam:
@artembilan can you help review this change?

artembilan:
Closing in favor of: https://github.com/spring-projects/spring-integration/pull/11122

I said on issue that the fix might be requested against `main`.
We do backward cherry-picks to all supported branches.

PR #11130, opened 2026-06-16
Use DefaultLockRegistry to support lock pooling

### LINKED ISSUE #11113: Prevent eviction of held locks from ZooKeeper registry
Similar to https://github.com/spring-projects/spring-integration/issues/11041

cppwfs:
- verify that we don't aquire InterProcessMutex more than once per lock
- adjust lock capacity to 256

Fixes: https://github.com/spring-projects/spring-integration/issues/11113

**Auto-cherry-pick to 7.0.x & 6.5.x**

### DIFF
--- a/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
+++ b/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
@@ -32,6 +32,7 @@ import org.apache.curator.framework.recipes.locks.InterProcessMutex;
 
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.core.task.AsyncTaskExecutor;
+import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.messaging.MessagingException;
 import org.springframework.scheduling.concurrent.ExecutorConfigurationSupport;
@@ -59,7 +60,7 @@ public class ZookeeperLockRegistry implements ExpirableLockRegistry<Lock>, Dispo
 
 	private final KeyToPathStrategy keyToPath;
 
-	private static final int DEFAULT_CAPACITY = 30_000;
+	private static final int DEFAULT_CAPACITY = 256;
 
 	private final Lock locksLock = new ReentrantLock();
 
@@ -90,6 +91,8 @@ public class ZookeeperLockRegistry implements ExpirableLockRegistry<Lock>, Dispo
 	private boolean mutexTaskExecutorExplicitlySet;
 
 	private int cacheCapacity = DEFAULT_CAPACITY;
+
+	private volatile DefaultLockRegistry defaultLockRegistry = new DefaultLockRegistry();
 
 	/**
 	 * Construct a lock registry using the default {@link KeyToPathStrategy} which
@@ -141,11 +144,17 @@ public class ZookeeperLockRegistry implements ExpirableLockRegistry<Lock>, Dispo
 
 	/**
 	 * Set the capacity of cached locks.
-	 * @param cacheCapacity The capacity of cached lock, (default 30_000).
+	 * @param cacheCapacity The capacity of cached lock, (default 256 locks).
 	 * @since 5.5.6
+	 *
+	 * @see DefaultLockRegistry
 	 */
 	public void setCacheCapacity(int cacheCapacity) {
 		this.cacheCapacity = cacheCapacity;
+		// Find the highest power of 2 for (n + 1), then subtract 1
+
+		int mask = Integer.highestOneBit(cacheCapacity + 1) - 1;
+		this.defaultLockRegistry = new DefaultLockRegistry(mask);
 	}
 
 	@Override
@@ -155,7 +164,8 @@ public class ZookeeperLockRegistry implements ExpirableLockRegistry<Lock>, Dispo
 		ZkLock lock;
 		this.locksLock.lock();
 		try {
-			lock = this.locks.computeIfAbsent(path, p -> new ZkLock(this.client, this.mutexTaskExecutor, p));
+			lock = this.locks.computeIfAbsent(path, p -> new ZkLock(this.client, this.mutexTaskExecutor, p,
+					(ReentrantLock) ZookeeperLockRegistry.this.defaultLockRegistry.obtain(path)));
 		}
 		finally {
 			this.locksLock.unlock();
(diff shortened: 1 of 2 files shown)


spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -141,11 +144,17 @@ public void setMutexTaskExecutor(AsyncTaskExecutor mutexTaskExecutor) {
 
 	/**
 	 * Set the capacity of cached locks.
-	 * @param cacheCapacity The capacity of cached lock, (default 30_000).
+	 * @param cacheCapacity The capacity of cached lock, (default 256 locks).
 	 * @since 5.5.6
+	 *
```
artembilan, line 149: No blank lines in method Javadocs

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -280,22 +294,41 @@ public void setLastUsed(long lastUsed) {
...
+					boolean locked = false;
+					// this is a bit ugly, but...
+					while (!locked) {
+						locked = doTryLock(1, TimeUnit.SECONDS);
+					}
+				}
+			}
+			catch (InterruptedException ie) {
+				this.delegate.unlock();
+				throw ie;
```
artembilan, line 323: What happened with `Thread.currentThread().interrupt();`?
Remember, that interrupted status is reset we call `Thread.interrupted()`?
See `InterruptedException` Javadocs.
See the same `lockInterruptibly()` implementation in the `JdbcLock`.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -263,11 +273,15 @@ private static final class ZkLock implements Lock {
 
 		private long lastUsed;
 
-		ZkLock(CuratorFramework client, AsyncTaskExecutor mutexTaskExecutor, String path) {
+		private final ReentrantLock delegate;
+
+		ZkLock(CuratorFramework client, AsyncTaskExecutor mutexTaskExecutor, String path,
```
artembilan, line 278: I wonder if it still justifies to keep this class as `static`.
Looks like dropping that would give us a good interaction with outer class.
Including new `defaultLockRegistry` property.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -263,11 +273,15 @@ private static final class ZkLock implements Lock {
...
+				ReentrantLock registryDelegate) {
 			this.client = client;
```
artembilan, line 280: Blank line after multi-line block header where method definition is also a block.
That makes code much easier to read.
Although, after removing `static` from this inner class, the `ZkLock` ctor would be much simpler.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -91,6 +92,8 @@ protected boolean removeEldestEntry(Entry<String, ZkLock> eldest) {
 
 	private int cacheCapacity = DEFAULT_CAPACITY;
 
+	private volatile DefaultLockRegistry defaultLockRegistry = new DefaultLockRegistry();
```
artembilan, line 95: I don't believe in `volatile` on setters.
That is really supposed to happen on the application context initialization, so no concurrent concern.
Therefore no need in hard thread state reset when we use this property.

Interesting though, that there is no `volatile` on same property in the `JdbcLockRegistry`.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -280,22 +294,41 @@ public void setLastUsed(long lastUsed) {
...
+			catch (InterruptedException ie) {
+				this.delegate.unlock();
+				throw ie;
+			}
+			catch (Exception e) {
+				this.delegate.unlock();
+				if (e instanceof RuntimeException runtimeException) {
+					throw runtimeException;
+				}
+				throw new RuntimeException(e);
```
artembilan, line 330: Why `JdbcLockRegistry.JdbcLock.rethrowAsLockException()` doesn't fit into logic here?

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -311,9 +344,37 @@ public boolean tryLock() {
...
+			}
+		}
+
+		private boolean doTryLock(long time, TimeUnit unit) throws InterruptedException {
 			Future<Boolean> future = null;
 			try {
-				long startTime = System.currentTimeMillis();
+				long startTime = System.nanoTime();
```
artembilan, line 377: Why have we started to use nanos?

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -324,39 +385,51 @@ public boolean tryLock(long time, TimeUnit unit) throws InterruptedException {
...
+				}
 				return false;
 			}
 			catch (InterruptedException e) {
 				Thread.currentThread().interrupt();
 				throw e;
 			}
 			catch (Exception e) {
-				throw new MessagingException("Failed to acquire mutex at " + this.path, e);
+				throw new MessagingException("Failed to acquire lock at " + this.path, e);
```
artembilan, line 412: The `MessagingException` is wrong in this contract.
OK. Let's leave for backward compatibility alongside with your `new RuntimeException()` in other places!
And let's raise an issue for new version to bring similar to others over here, too:
```
		private void rethrowAsLockException(Exception e) {
			throw new CannotAcquireLockException("Failed to lock mutex at " + this.lockKey, e);
		}
```

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -280,22 +294,41 @@ public void setLastUsed(long lastUsed) {
...
+					while (!locked) {
+						locked = doTryLock(1, TimeUnit.SECONDS);
+					}
+				}
+			}
+			catch (InterruptedException ie) {
+				this.delegate.unlock();
+				throw ie;
+			}
+			catch (Exception e) {
```
artembilan, line 325: Just nit-pick: it appears more clear when the variable name is not one symbol.
I also know that Checkstyle from Spring Java Format rejects this kind of definitions.
I know we have already enough in this class and in other places, but still would be great to not introduce new.

artembilan (changes requested):
I see now the problem.

Let's take it offline tomorrow!
Or whenever we can!

artembilan:
We will go different solution.

Thanks

PR #11131, opened 2026-06-17
Verify that ZKLock is not engaged on capacity clean

### LINKED ISSUE #11113: Prevent eviction of held locks from ZooKeeper registry

cppwfs:
- Do not remove entry if it still has a zk lock from cache when cacheCapacity limit has been reached.
- If the oldest entry in the cache can't be removed because it is locked.  Find the next oldest entry and remove it
- Add test that validates the fix

Resolves: https://github.com/spring-projects/spring-integration/issues/11113

**Auto-cherry-pick to 7.0.x & 6.5.x**

### DIFF
--- a/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
+++ b/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
@@ -17,6 +17,7 @@
 package org.springframework.integration.zookeeper.lock;
 
 import java.io.Serial;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -47,6 +48,7 @@ import org.springframework.util.Assert;
  * @author Vedran Pavic
  * @author Unseok Kim
  * @author Christian Tzolov
+ * @author Glenn Renfro
  *
  * @since 4.2
  *
@@ -71,7 +73,20 @@ public class ZookeeperLockRegistry implements ExpirableLockRegistry<Lock>, Dispo
 
 				@Override
 				protected boolean removeEldestEntry(Entry<String, ZkLock> eldest) {
-					return size() > ZookeeperLockRegistry.this.cacheCapacity;
+					if (size() > ZookeeperLockRegistry.this.cacheCapacity) {
+						if (!eldest.getValue().isAcquiredInThisProcess()) {
+							return true;
+						}
+						Iterator<Entry<String, ZkLock>> iterator = entrySet().iterator();
+						while (iterator.hasNext()) {
+							Entry<String, ZkLock> entry = iterator.next();
+							if (!entry.getValue().isAcquiredInThisProcess()) {
+								iterator.remove();
+								break;
+							}
+						}
+					}
+					return false;
 				}
 
 			};
--- a/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
+++ b/spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
@@ -17,7 +17,6 @@
 package org.springframework.integration.zookeeper.lock;
 
 import java.io.Serial;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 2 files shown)


spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/lock/ZookeeperLockRegistry.java
```diff
@@ -71,7 +73,20 @@ public class ZookeeperLockRegistry implements ExpirableLockRegistry<Lock>, Dispo
 
 				@Override
 				protected boolean removeEldestEntry(Entry<String, ZkLock> eldest) {
-					return size() > ZookeeperLockRegistry.this.cacheCapacity;
+					if (size() > ZookeeperLockRegistry.this.cacheCapacity) {
+						if (!eldest.getValue().isAcquiredInThisProcess()) {
+							return true;
+						}
+						Iterator<Entry<String, ZkLock>> iterator = entrySet().iterator();
```
artembilan: I just assume that iterator returns entries from oldest first.
Therefore, the first `next()` gives us exactly the eldest entry we’ve just rejected.
And the next one: I’ll just believe you that we can call that `remove()` from the current remove logic.

artembilan (changes requested):
Can we go with `Fixes:` sentence instead ?
Otherwise our Backport bot won’t create required issues: https://github.com/spring-projects/spring-integration/blob/00e93ed89ea0c064cd0da7c8399163a4d77d0203/.github/workflows/backport-issue.yml#L10.

I think silent eviction from the cache will lead to the same dead lock on ZK entry again. I mean we have just obtained but not locked. So, it got evicted. Then we lock a bit later and obtain the same key again and got dead lock.
We may consider to put into cache back on `lock()` call or so.
But feels more like we are fighting ghosts and walking in loops.
I would say let’s just leave cache logic without iterator removal.
And probably let’s change the default strategy for removal flag !

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/lock/ZkLockRegistryTests.java
```diff
@@ -508,6 +508,36 @@ public void setCapacityTest() {
...
+		try {
+			registry.obtain("test2");
+
+			Lock lock1 = registry.obtain("test");
+
+			// Attempt to acquire lock with 10-second timeout
+			boolean acquired = lock1.tryLock(10, TimeUnit.SECONDS);
+
+			// Fail the test if the lock was not acquired within the timeframe
+			assertThat(acquired).isTrue();
```
artembilan: I think this really should go to the `try..finally`.
And probably without any comments.

We also can have an assertion that `lock1` is same as `lock`.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/lock/ZkLockRegistryTests.java
```diff
@@ -508,6 +508,36 @@ public void setCapacityTest() {
...
+	@Test
+	public void testLockEvictionWithAnotherLockOnTheKey() throws InterruptedException {
+		ZookeeperLockRegistry registry = new ZookeeperLockRegistry(this.client);
+		registry.setCacheCapacity(1);
+
+		Lock lock = registry.obtain("test");
+		// cache is full
+		lock.lock();
+		try {
+			registry.obtain("test2");
```
artembilan, line 521: Let's add here comment like:
```
Although cache is full, this does not evict an old entry because it is locked yet
```

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/lock/ZkLockRegistryTests.java
```diff
@@ -517,4 +547,16 @@ private static String toKey(String path) {
...
+		@Override
+		public String pathFor(String key) {
+			return "/SpringIntegration-LockRegistry" + key;
+		}
+
+		@Override
+		public boolean bounded() {
+			return false;
+		}
+	}
```
artembilan, line 561: Every class member should be surrounded with blank lines.
Much easier to ready the code and add new members.

artembilan (changes requested):

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/lock/ZkLockRegistryTests.java
```diff
@@ -508,6 +508,36 @@ public void setCapacityTest() {
...
+			registry.obtain("test2");
+
+			Lock lock1 = registry.obtain("test");
+
+			// Attempt to acquire lock with 10-second timeout
+			boolean acquired = lock1.tryLock(10, TimeUnit.SECONDS);
+
+			try {
+				assertThat(acquired).isTrue();
+				assertThat(lock1).isEqualTo(lock);
```
artembilan: Why `equal` but not `same`?
According to the cache purpose, the get has to return the same object.

artembilan (changes requested):

PR #11137, opened 2026-06-19
Not warn that Redis 7 server does not support DELEX

michaldo:
Actual: 
RedisLockRegistry print warning that Redis server does not support DELEX Expected:
DELEX is added in Redis 8.4. I use Redis 7. I know progress is forward, new features appear in new software, but log level WARN is not appropriate to remind it. 

Basically, I don't want be warned that I use Redis 7 as long as Redis 7 is officially supported


artembilan (approved):
Makes sense.
Thanks
Merging...
Looking forward for more contributions!

PR #11159, opened 2026-07-02
Migrate version mgmt to TOML

cppwfs:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -103,6 +103,7 @@ allprojects {
 		dependencyManagement(platform(libs.spring.security.bom))
 		dependencyManagement(platform(libs.spring.ws.bom))
 		dependencyManagement(platform(libs.org.mongodb.driver.bom))
+		dependencyManagement(platform(libs.apache.artemis.bom))
 	}
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -689,33 +619,33 @@ project('spring-integration-jdbc') {
...
     description = 'Spring Integration JMS Support'
     dependencies {
         api 'org.springframework:spring-jms'
-        provided "jakarta.jms:jakarta.jms-api:$jmsApiVersion"
+        provided libs.jakarta.jms.api
 
-        testImplementation "org.apache.activemq:artemis-server:$artemisVersion"
-        testImplementation "org.apache.activemq:artemis-jakarta-client:$artemisVersion"
+        testImplementation "org.apache.activemq:artemis-server:${libs.versions.artemisVersion.get()}"
+        testImplementation "org.apache.activemq:artemis-jakarta-client:${libs.versions.artemisVersion.get()}"
```
artembilan: It's not like important, but we've agreed on `main` to use Artemis BOM: https://github.com/spring-projects/spring-integration/blob/main/gradle/libs.versions.toml#L74

artembilan (changes requested):

PR #11160, opened 2026-07-03
Preserve original headers in `DefaultErrorMessageStrategy`

mjd507:
When setup `RequestHandlerRetryAdvice` with a recover `ErrorMessageSendingRecoverer`,  after retry exhausted,  the `DefaultErrorMessageStrategy` will build a `ErrorMessage`,  however it does not copy the original message headers.

because I need the header values to do the error routing, so I think it is useful to keep it.

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/support/DefaultErrorMessageStrategy.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/support/DefaultErrorMessageStrategy.java
@@ -24,12 +24,13 @@ import org.springframework.messaging.support.ErrorMessage;
 
 /**
  * A simple {@link ErrorMessageStrategy} implementations which produces
- * a error message with original message if the {@link AttributeAccessor} has
+ * an error message with original message if the {@link AttributeAccessor} has
  * {@link ErrorMessageUtils#INPUT_MESSAGE_CONTEXT_KEY} attribute.
- * Otherwise plain {@link ErrorMessage} with the {@code throwable} as {@code payload}.
+ * Otherwise, plain {@link ErrorMessage} with the {@code throwable} as {@code payload}.
  *
  * @author Gary Russell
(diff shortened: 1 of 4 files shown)


spring-integration-core/src/main/java/org/springframework/integration/support/DefaultErrorMessageStrategy.java
```diff
@@ -41,8 +42,8 @@ public class DefaultErrorMessageStrategy implements ErrorMessageStrategy {
 	public ErrorMessage buildErrorMessage(Throwable throwable, @Nullable AttributeAccessor attributes) {
 		Object inputMessage = attributes == null ? null
 				: attributes.getAttribute(ErrorMessageUtils.INPUT_MESSAGE_CONTEXT_KEY);
-		if (inputMessage instanceof Message) {
-			return new ErrorMessage(throwable, (Message<?>) inputMessage);
+		if (inputMessage instanceof Message<?> message) {
+			return new ErrorMessage(throwable, message.getHeaders(), message);
```
artembilan, line 46: OK. I don't see a harm in this, but still in consistency with our support and compatibility policy, we cannot accept this right now.
This is some kind of breaking change, so we aim it for the next minor `7.2` starting in August.
Please, consider to mention this in the class Javadoc.
Also describe such a logic in the dedicated Error Handling chapter of the doc: https://docs.spring.io/spring-integration/reference/error-handling.html.

When we switch to new version, we will also ask you to add a note about this change in the `whats-new.adoc` dedicated to that new `7.2` version.

Thank you!

artembilan (changes requested):

artembilan (changes requested):
@mjd507 ,

We have switched recently to `7.2`.
Please, consider to add a whats-new note into that `General Changes` sections.

Thanks

PR #11168, opened 2026-07-08
Mark function return type `@Nullable` for lockKey & waitLockDuration

mjd507:


artembilan (changes requested):
What is justification for such a change, please?

mjd507:
> What is justification for such a change, please?

Just want to eliminate the warning in IntelliJ IDEA.

artembilan:
> Just want to eliminate the warning in IntelliJ IDEA.

I see.
You probably mean this:
image

So, yes, agreed and merging.

Thank you!

PR #11170, opened 2026-07-09
GH-11157: Fix caching logic in the `JdbcLockRegistry`

### LINKED ISSUE #11157: JDBC locks not unique

spring-integration-jdbc 6.5.9

**Describe the bug**

JdbcLocks are not unique anymore. With spring-integration-jdbc 6.5.9 DefaultLockRegistry is used in some cases. This automatically results in non-uniqueness of the locks (delegates) which in general even contradicts the documented behavior.

**To Reproduce**

Obtain JdbcLock and tryLock for 2 different keys (e.g. "0" and "641").

**Expected behavior**

As documented in the JdbcLockRegistry class:

"... An {@link ExpirableLockRegistry} using a shared database to co-ordinate the locks. Provides the same semantics as the {@link org.springframework.integration.support.locks.DefaultLockRegistry}, but the locks taken will be global, as long as the underlying database supports the "serializable" isolation level in its transactions."

This change is also backward incompatible. Prior version 6.5.9 locks were not overlapping (e.g. the code below would work).

**Sample**

```java
public static void main(String[] args) throws InterruptedException, ExecutionException {
    JdbcLockRegistry registry = new JdbcLockRegistry(...);

    ExecutorService executor = Executors.newSingleThreadExecutor();

    assert registry.obtain("0").tryLock();

    executor.submit(() -> {
      assert registry.obtain("641").tryLock() : "should be able to lock";
    }).get();
  }
```

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/11157

The previous (https://github.com/spring-projects/spring-integration/issues/11041) solution, based on the shared pool of `ReentrantLock` from the `DefaultLockRegistry`, being stable and smooth with the memory, is proven to be inconvenient with keys collision. So, different keys might stumble on the same shared `ReentrantLock` making the application being blocked for nothing.

* Remove `DefaultLockRegistry` logic from the `JdbcLockRegistry`
* Remove local `JdbcLock.holdCount` as we rely on the `delegate` again
* Revert back default `cacheCapacity` to `100_000` as we don't use shared pool of locks anymore
* Do not evict from the cache still in-held locks
* In the `obtain()`, try to remove unused lock when `cacheCapacity` is reached. If it cannot clean up the room, throw a `CannotAcquireLockException`. This is to prevent an out-of-memory error
* Also, throw a `CannotAcquireLockException` on lock calls for orphaned lock. This is to prevent a race condition when two distinct local locks are locked in different threads against the same key
* Add human-readable `toString()` into `DefaultLockRepository` and `JdbcLockRegistry`
* Use provided lock key for logs and exceptions instead of internal uuid
* Some other reasonable code refactoring in the `JdbcLockRegistry`

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
@@ -30,7 +30,6 @@ import java.util.concurrent.locks.ReentrantLock;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.dao.DataAccessResourceFailureException;
 import org.springframework.dao.TransientDataAccessException;
-import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.DistributedLock;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
 import org.springframework.integration.support.locks.RenewableLockRegistry;
@@ -46,6 +45,13 @@ import org.springframework.util.Assert;
  * {@link org.springframework.integration.support.locks.DefaultLockRegistry}, but the
  * locks taken will be global, as long as the underlying database supports the
  * "serializable" isolation level in its transactions.
+ * <p>
+ * The lock instances are cached by their key for future reuse.
+ * If all the locks in the cache are acquired,
+ * the {@link #obtain(Object)} method throws {@link CannotAcquireLockException}.
+ * Otherwise, the old, unused locks are evicted from the cache.
+ * An attempt to lock such an orphaned {@link JdbcLock} throws another {@link CannotAcquireLockException}
+ * to avoid double locking from different threads on the same key.
  *
  * @author Dave Syer
  * @author Artem Bilan
@@ -68,7 +74,7 @@ public class JdbcLockRegistry implements ExpirableLockRegistry<DistributedLock>,
 
 	private static final int DEFAULT_IDLE = 100;
 
-	private static final int DEFAULT_CAPACITY = 256;
+	private static final int DEFAULT_CAPACITY = 100_000;
 
 	private final Lock lock = new ReentrantLock();
 
@@ -80,7 +86,8 @@ public class JdbcLockRegistry implements ExpirableLockRegistry<DistributedLock>,
 
 				@Override
 				protected boolean removeEldestEntry(Entry<String, JdbcLock> eldest) {
-					return size() > JdbcLockRegistry.this.cacheCapacity;
+					return size() > JdbcLockRegistry.this.cacheCapacity
+							&& !eldest.getValue().isAcquiredInThisProcess();
 				}
 
 			};
@@ -98,8 +105,6 @@ public class JdbcLockRegistry implements ExpirableLockRegistry<DistributedLock>,
 	public static final Duration DEFAULT_TTL = Duration.ofSeconds(10);
 
 	private final Duration ttl;
-
-	private DefaultLockRegistry defaultLockRegistry = new DefaultLockRegistry();
 
 	/**
 	 * Construct an instance based on the provided {@link LockRepository}.
@@ -134,24 +139,36 @@ public class JdbcLockRegistry implements ExpirableLockRegistry<DistributedLock>,
 
 	/**
 	 * Set the capacity of cached locks.
-	 * @param cacheCapacity The capacity of cached lock, (default 256 locks).
+	 * @param cacheCapacity The capacity of cached lock, (default 100_000).
 	 * @since 5.5.6
-	 * @see DefaultLockRegistry
 	 */
 	public void setCacheCapacity(int cacheCapacity) {
 		this.cacheCapacity = cacheCapacity;
-		// Find the highest power of 2 for (n + 1), then subtract 1
-		int mask = Integer.highestOneBit(cacheCapacity + 1) - 1;
-		this.defaultLockRegistry = new DefaultLockRegistry(mask);
 	}
 
 	@Override
 	public DistributedLock obtain(Object lockKey) {
 		Assert.isInstanceOf(String.class, lockKey);
-		String path = pathFor((String) lockKey);
+		String lockKeyToUse = (String) lockKey;
 		this.lock.lock();
 		try {
-			return this.locks.computeIfAbsent(path, key -> new JdbcLock(this.client, this.idleBetweenTries, key));
+			JdbcLock jdbcLock =
+					this.locks.computeIfAbsent(lockKeyToUse, key -> new JdbcLock(lockKeyToUse));
+			if (this.locks.size() > this.cacheCapacity) {
+				long now = System.currentTimeMillis();
+				this.locks.entrySet()
+						.removeIf(entry -> {
+							JdbcLock lock = entry.getValue();
+							return lock != jdbcLock && now - lock.getLastUsed() >= 0 && !lock.isAcquiredInThisProcess();
+						});
+				if (this.locks.size() > this.cacheCapacity) {
+					this.locks.remove(lockKeyToUse);
+					throw new CannotAcquireLockException("There are already " +
+							this.cacheCapacity + " unique JDBC locks acquired in the application " +
+							"from " + this + ". Cannot obtain more at the moment.");
+				}
+			}
+			return jdbcLock;
 		}
 		finally {
 			this.lock.unlock();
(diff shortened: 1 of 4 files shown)


artembilan:
When we are agreed on the solution, I'll go ahead and issue similar fix for `RedisLockRegistry` in the separate PR.

Thanks

src/reference/antora/modules/ROOT/pages/jdbc/lock-registry.adoc
```diff
@@ -44,11 +44,13 @@ So the time to live can be highly reduced, and deployments can retake a lost loc
...
 
-Starting with version 5.5.6, the `JdbcLockRegistry` is support automatically clean-up cache for JdbcLock in `JdbcLockRegistry.locks` via `JdbcLockRegistry.setCacheCapacity()`.
-The default value is `256`.
+Starting with version 5.5.6, the `JdbcLockRegistry` supports automatic cache clean-up for unused `JdbcLock` instances via `JdbcLockRegistry.setCacheCapacity()`.
+The default value is `100_000`.
 See its Javadocs for more information.
-This capacity number is also propagated into the `DefaultLockRegistry` used internally in the `JdbcLockRegistry` for a pool of shared `ReentrantLock` instances.
-So, if higher concurrent access (the number of unique lock keys used in parallel) is expected (or allowed) for the application, the `cacheCapacity` should be increased respectively, with the closest power of 2 in mind (essentially `Integer.highestOneBit(cacheCapacity + 1)`).
+The cache can be also cleaned via `expireUnusedOlderThan(long age)` API.
+When all the cached `JdbcLock` instances are acquired, and neither of them can be evicted, the `obtain(Object lockKey)` method throws a `CannotAcquireLockException`.
```
cppwfs: Replace `and neither of them ` with `and none of them`

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -134,24 +139,36 @@ public void setIdleBetweenTries(Duration idleBetweenTries) {
...
+					this.locks.computeIfAbsent(lockKeyToUse, key -> new JdbcLock(lockKeyToUse));
+			if (this.locks.size() > this.cacheCapacity) {
+				long now = System.currentTimeMillis();
+				this.locks.entrySet()
+						.removeIf(entry -> {
+							JdbcLock lock = entry.getValue();
+							return lock != jdbcLock && now - lock.getLastUsed() >= 0 && !lock.isAcquiredInThisProcess();
+						});
+				if (this.locks.size() > this.cacheCapacity) {
+					this.locks.remove(lockKeyToUse);
```
cppwfs, line 165: Looks like we may have a small issue here:
When at cache at capacity, all entries acquired-in-process  except none evictable, key "test1" already cached and currently held by thread A; thread B calls obtain("test1") — computeIfAbsent returns the same JdbcLock A holds, the cache is still over capacity, and the code throws CannotAcquireLockException while removing "test1" from this.locks.  Thread A, still holding delegate, will now fail its next doLock()/unlock() call with the "evicted from the exhausted cache" IllegalStateException, even though it legitimately owns the lock. 
The removal on the exception path should be conditional on the map not already containing an acquired lock for that key (or should simply never remove a key whose value isAcquiredInThisProcess()).

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -134,24 +139,36 @@ public void setIdleBetweenTries(Duration idleBetweenTries) {
...
 		try {
-			return this.locks.computeIfAbsent(path, key -> new JdbcLock(this.client, this.idleBetweenTries, key));
+			JdbcLock jdbcLock =
+					this.locks.computeIfAbsent(lockKeyToUse, key -> new JdbcLock(lockKeyToUse));
+			if (this.locks.size() > this.cacheCapacity) {
+				long now = System.currentTimeMillis();
+				this.locks.entrySet()
+						.removeIf(entry -> {
+							JdbcLock lock = entry.getValue();
+							return lock != jdbcLock && now - lock.getLastUsed() >= 0 && !lock.isAcquiredInThisProcess();
```
cppwfs, line 162: I think `now - lock.getLastUsed` is always `true`.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
```diff
@@ -313,29 +324,31 @@ public boolean tryLock(Duration waitTime, Duration ttl) throws InterruptedExcept
...
 				}
 			}
 		}
 
 		private boolean doLock(Duration ttl) {
-			boolean acquired = this.mutex.acquire(this.path, ttl);
+			if (!JdbcLockRegistry.this.locks.containsKey(this.key)) {
+				throw new IllegalStateException("The lock '" + this.key + "' was evicted from the exhausted cache " +
+						"due to its unused period for " + (System.currentTimeMillis() - this.lastUsed) +
+						" milliseconds. Obtain a fresh one from " + JdbcLockRegistry.this);
```
cppwfs, line 349: Can we just say "Obtain a fresh one from the JdbcRegistry", instead of "Obtain a fresh one from JdbcLockRegistry@5a3962d2"?

cppwfs (changes requested):
This is great!   Thanks for doing this!

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: Well, there is still some timing race condition when clock is not ticking.
I saw that in tests.
So, `long now = System.currentTimeMillis();` and `private volatile long lastUsed = System.currentTimeMillis();`.
There is a chance that `now - lock.getLastUsed == 0`.
Therefore, just `> 0` is not enough when `obtain()` is called rapidly.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: Hm. How that happened for you since for me this exception message is:
```
"java.lang.IllegalStateException: The lock 'lock1' was evicted from the exhausted cache due to its unused period for 3290 milliseconds. Obtain a fresh one from JdbcLockRegistry{client=DefaultLockRepository{id='e9a92e02-ca3b-46de-ba71-e4f50b9f254b', region='DEFAULT'}}"
```
See a new `toString()` implementation in the `JdbcLockRegistry` as a part of this PR.

spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/lock/JdbcLockRegistry.java
artembilan: Looking into this.
Thanks

artembilan: OK.
I think that is not possible.

1. `registry.setCacheCapacity(1);`
2. Obtain `lock1` and lock.
3. Cache is full and nothing to evict.
4. When we call for `lock2`, it is not going to be locked, therefore we cannot add more entries. And we cannot evict `lock1` since it is locked. So, remove `lock2` and throw an exception.
5. If we call for `lock1` again, we don't go over cache capacity because new entry is not added to the cache. Therefore early exit from that `if (this.locks.size() > this.cacheCapacity) {`

cppwfs: I think this test may show the scenario better.   It will pass and it shouldn't.
```
	@Test
	void obtainDoesNotEvictCurrentlyHeldLockOfRequestedKey() {
		DefaultLockRepository client = new DefaultLockRepository(dataSource);
		client.setApplicationContext(this.context);
		client.afterPropertiesSet();
		client.afterSingletonsInstantiated();
		JdbcLockRegistry registry = new JdbcLockRegistry(client);
		registry.setCacheCapacity(2);

		DistributedLock lock1 = registry.obtain("lock1");
		lock1.lock();
		DistributedLock lock2 = registry.obtain("lock2");
		lock2.lock();
		try {
			// Shrinking the capacity below the number of currently held locks forces
			// the next `obtain` for an already-cached, held key to see `size() > cacheCapacity`.
			registry.setCacheCapacity(1);
			
			//verify that lock1 is still present
			assertThat(getRegistryLocks(registry)).containsKey("lock1");

			assertThatExceptionOfType(CannotAcquireLockException.class)
					.isThrownBy(() -> registry.obtain("lock1"))
					.withMessageStartingWith("There are already 1 unique JDBC locks acquired in the application");

			// The cache entry for 'lock1' was wrongly evicted even though this thread still holds it.
			assertThat(getRegistryLocks(registry)).doesNotContainKey("lock1");

			// The still-owning thread now fails to re-lock (e.g. on any reentrant lock() call)
			// even though it never lost its local `ReentrantLock`.
			assertThatExceptionOfType(CannotAcquireLockException.class)
					.isThrownBy(lock1::lock)
					.havingCause()
					.isInstanceOf(IllegalStateException.class)
					.withMessageStartingWith("The lock 'lock1' was evicted from the exhausted cache due to its unused period for");
		}
		finally {
			lock2.unlock();
		}
	}
```

artembilan: `Shrinking`?
This is not what you have talked originally.
The scenario is made up and is a trap you made yourself .
I think our logic is adamant to fail on re-locking since we have just removed what you asked. 

I prefer to have setters as immutable options, but is not possible with Java.

I can add the guard into that setter to disallow mutation.

Either way: this is a new peculiar scenario you did not describe originally.

Tell me, please, if that is possible without shrinking as you stated before?

cppwfs: You are correct.   
I created a threaded scenario test and the code as it stands prevents the scenario I came up with.   The only way is to purposefully reduce the capacity.  
Apologies for the noise, I should have run the tests proving the scenario before writing up the issue.   What is in the head does not always translate to reality :-)

cppwfs (approved):
Looks good to me.  Great work solving this!
Sorry for the noise earlier.

PR #11172, opened 2026-07-10
Adjust cache capacity in the `JdbcLockRegistry`

artembilan:
When `cacheCapacity` has shrunk, and we obtain already held lock again, it is evicted due to the current `if (this.locks.size() > this.cacheCapacity)` logic. This leads to several side effects for already locked instance.

* Implement the logic for shrinking `cacheCapacity`: evict unsed lock when cache size is higher than new `cacheCapacity`.
* Fix `obtain()` logic to skip eviction attempt if obtained lock is held


artembilan:
Related to: https://github.com/spring-projects/spring-integration/pull/11170

cppwfs (approved):
Looks great!

Thanks for making this change.

PR #11181, opened 2026-07-14
GH-11180: Fix caching logic in the RedisLockRegistry

### LINKED ISSUE #11180: Redis locks are not not unique
Similar to https://github.com/spring-projects/spring-integration/issues/11157 but with `RedisLockRegistry`.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/11180

The previous solution, based on the shared pool of `ReentrantLock` from the `DefaultLockRegistry`, is proven to be inconvenient with keys collision: different keys might stumble on the same shared `ReentrantLock` making the application being blocked for nothing.

* Remove `DefaultLockRegistry` logic from the `RedisLockRegistry`
* Remove local `RedisLock.holdCount` as we rely on the `localLock` again
* Revert back default `cacheCapacity` to `100_000` as we don't use a shared pool of locks anymore
* Do not evict from the cache still in-held locks
* In the `obtain()`, try to remove unused lock when `cacheCapacity` is reached. If it cannot clean up the room, throw a `CannotAcquireLockException`. This is to prevent an out-of-memory error
* Also, throw a `CannotAcquireLockException` on lock calls for an orphaned lock. This is to prevent a race condition when two distinct local locks are locked in different threads against the same key
* Add human-readable `toString()` into the `RedisLockRegistry`

### DIFF
--- a/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
+++ b/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java
@@ -62,7 +62,6 @@ import org.springframework.data.redis.core.script.RedisScript;
 import org.springframework.data.redis.listener.PatternTopic;
 import org.springframework.data.redis.listener.RedisMessageListenerContainer;
 import org.springframework.data.redis.listener.Topic;
-import org.springframework.integration.support.locks.DefaultLockRegistry;
 import org.springframework.integration.support.locks.DistributedLock;
 import org.springframework.integration.support.locks.ExpirableLockRegistry;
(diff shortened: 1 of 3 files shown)


cppwfs (approved):
Looks great.  
I found no issues.

Thanks for doing this!

hyunto:
Could you let me know which Spring Boot version this Pull Request will first be released in? It doesn't appear to be included in 4.0.7 or 4.1.0. Is it scheduled for 4.0.8 and 4.1.1?

cppwfs:
Next Release is 4.0.5 and 4.1.1.   It should be in these releases as well as the 4.2.0-M1.

PR #11184, opened 2026-07-15
GH-11183: Fix batch mode message ID generation in KafkaMessageDrivenChannelAdapter

### LINKED ISSUE #11183: KafkaMessageDrivenChannelAdapter in batch mode doesn't generate message ID, causing splitter to fail
## Description
This issue occurs after upgrading to Spring Boot 4. When using `Kafka.messageDrivenChannelAdapter` with `ListenerMode.batch` followed by a `.split()`, the splitter fails with:

```
java.lang.IllegalStateException: Correlation ID must not be null.
	at org.springframework.util.Assert.state(Assert.java:80)
	at org.springframework.integration.splitter.AbstractMessageSplitter.createBuilder(AbstractMessageSplitter.java:284)
	at org.springframework.integration.splitter.AbstractMessageSplitter.lambda$prepareMessageBuilderFunction$0(AbstractMessageSplitter.java:237)
	at org.springframework.integration.util.FunctionIterator.next(FunctionIterator.java:98)
	at org.springframework.integration.splitter.AbstractMessageSplitter.produceOutput(AbstractMessageSplitter.java:321)
	at org.springframework.integration.handler.AbstractMessageProducingHandler.sendOutputs(AbstractMessageProducingHandler.java:253)
	at org.springframework.integration.handler.AbstractReplyProducingMessageHandler.handleMessageInternal(AbstractReplyProducingMessageHandler.java:152)
	at org.springframework.integration.handler.AbstractMessageHandler.doHandleMessage(AbstractMessageHandler.java:102)
```

## Expected Behavior
The batch message should have a message ID generated, similar to record mode, so that the splitter can properly correlate the split messages.

## Actual Behavior
When using batch mode, the `BatchMessagingMessageConverter` is not configured to generate message IDs, causing the splitter to fail when trying to set correlation headers.

## Example Configuration
```java
IntegrationFlow
    .from(Kafka.messageDrivenChannelAdapter(consumerFactory(),
            KafkaMessageDrivenChannelAdapter.ListenerMode.batch, TEST_TOPIC)
            .configureListenerContainer(c -> c.ackMode(ContainerProperties.AckMode.MANUAL)))
    .split()
    .channel(c -> c.queue("results"))
    .get();
```

## Root Cause
In `KafkaMessageDrivenChannelAdapter` constructor, when in batch mode, the code was setting the same `MessagingMessageConverter` (intended for record mode) on the batch listener instead of creating a proper `BatchMessagingMessageConverter` with message ID generation enabled.

## Proposed Fix
Ensure that when in batch mode, a `BatchMessagingMessageConverter` is created with `setGenerateMessageId(true)` and `setGenerateTimestamp(true)`, similar to how it's done for record mode.

Zernov-A:
Fixes #11183

When using `Kafka.messageDrivenChannelAdapter` with `ListenerMode.batch` followed by a `.split()`, the splitter fails because the batch message doesn't have a message ID generated. This is required by the splitter to properly set correlation headers on split messages.

The root cause was that the constructor was setting the same `MessagingMessageConverter` (intended for record mode) on the batch listener instead of creating a proper `BatchMessagingMessageConverter` with message ID generation enabled.

This fix ensures that when in batch mode, a `BatchMessagingMessageConverter` is created with `setGenerateMessageId(true)` and `setGenerateTimestamp(true)`, similar to how it's done for record mode.

## Changes
- Modified `KafkaMessageDrivenChannelAdapter` constructor to properly configure `BatchMessagingMessageConverter` for batch mode
- Added test case `testBatchListener()` in `KafkaDslTests` to verify the fix
- Added assertions for message ID in existing tests to ensure consistency

### DIFF
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
@@ -78,6 +78,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Urs Keller
  * @author Jooyoung Pyoung
+ * @author Zernov Oleksii
  *
  * @since 5.4
  */
@@ -144,7 +145,6 @@ public class KafkaMessageDrivenChannelAdapter<K, V> extends MessageProducerSuppo
 					JacksonMessagingUtils.DEFAULT_TRUSTED_PACKAGES
 							.toArray(new String[0]));
 			messageConverter.setHeaderMapper(headerMapper);
-			this.recordListener.setMessageConverter(messageConverter);
 		}
 		else if (JacksonPresent.isJackson2Present()) {
 			var headerMapper = new org.springframework.kafka.support.DefaultKafkaHeaderMapper();
@@ -152,8 +152,9 @@ public class KafkaMessageDrivenChannelAdapter<K, V> extends MessageProducerSuppo
 					org.springframework.integration.support.json.JacksonJsonUtils.DEFAULT_TRUSTED_PACKAGES
 							.toArray(new String[0]));
 			messageConverter.setHeaderMapper(headerMapper);
-			this.recordListener.setMessageConverter(messageConverter);
 		}
+
+		this.recordListener.setMessageConverter(messageConverter);
 
 		if (this.mode.equals(ListenerMode.batch)) {
 			var batchMessagingMessageConverter = new BatchMessagingMessageConverter(messageConverter);
--- a/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
+++ b/spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
@@ -57,6 +57,7 @@ import org.springframework.kafka.support.JacksonPresent;
 import org.springframework.kafka.support.JsonKafkaHeaderMapper;
 import org.springframework.kafka.support.KafkaHeaders;
 import org.springframework.kafka.support.converter.BatchMessageConverter;
+import org.springframework.kafka.support.converter.BatchMessagingMessageConverter;
 import org.springframework.kafka.support.converter.ConversionException;
 import org.springframework.kafka.support.converter.KafkaMessageHeaders;
 import org.springframework.kafka.support.converter.MessageConverter;
@@ -144,7 +145,6 @@ public class KafkaMessageDrivenChannelAdapter<K, V> extends MessageProducerSuppo
 							.toArray(new String[0]));
 			messageConverter.setHeaderMapper(headerMapper);
 			this.recordListener.setMessageConverter(messageConverter);
-			this.batchListener.setMessageConverter(messageConverter);
 		}
 		else if (JacksonPresent.isJackson2Present()) {
 			var headerMapper = new org.springframework.kafka.support.DefaultKafkaHeaderMapper();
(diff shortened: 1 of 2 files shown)


spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -153,7 +153,13 @@ else if (JacksonPresent.isJackson2Present()) {
 							.toArray(new String[0]));
 			messageConverter.setHeaderMapper(headerMapper);
 			this.recordListener.setMessageConverter(messageConverter);
```
artembilan: I think the `this.recordListener.setMessageConverter()` should go outside of the `if..else`, too.

artembilan (changes requested):

Looking good!

Thank you!

Zernov-A:
> Please, add your name to the `@author` list of the affected classes.
> 
> Looking good!
> 
> Thank you!

Thank you for the review! I've added the `@author` tag and moved `setMessageConverter()` outside the `if..else` blocks as requested.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -77,6 +78,7 @@
  * @author Artem Bilan
  * @author Urs Keller
  * @author Jooyoung Pyoung
+ * @author Zernov Oleksii
```
artembilan: I think the international naming convention is to express first name first.
See an example of my name in this list which is Ukrainian, too.

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
```diff
@@ -269,6 +280,17 @@ void channels(@Qualifier("topic6Channel") MessageChannel topic6Channel,
 				.isEqualTo("foo");
 	}
 
+	@Test
+	void testBatchListener() {
+		this.kafkaTemplate.send(TEST_TOPIC10, "foo");
```
artembilan: Please, don't use `foo/bar` language in the code.
I know we have a lot of them already there, but we prefer to get rid of it eventually.

Thanks

spring-integration-kafka/src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
```diff
@@ -341,6 +363,20 @@ public IntegrationFlow topic2ListenerFromKafkaFlow() {
...
 
+		@Bean
+		public IntegrationFlow batchListenerFromKafkaFlow() {
+			return IntegrationFlow
+					.from(Kafka
+							.messageDrivenChannelAdapter(consumerFactory(),
+									KafkaMessageDrivenChannelAdapter.ListenerMode.batch, TEST_TOPIC10)
+							.configureListenerContainer(c ->
+									c.ackMode(ContainerProperties.AckMode.MANUAL)
+											.id("topic10ListenerContainer")))
```
artembilan: I don't think that we need any of these customization to proof the concept.
I mean the goal of this specific test to make sure splitter doesn't fail.
So, what container id or what mode, doesn't matter. 
Therefore redundant from the readability perspective.
Less code in the test - less chance of bugs in tests 😄

artembilan (changes requested):
Looks good.
Just couple minors.
Thanks

artembilan (approved):
LGTM.

Will be merged automatically when PR build is green.

For the future: all the info you have included into PR description, could simply go to the first commit - and it becomes that PR description. 
Now I had to add it to the final commit manually.

PR #11205, opened 2026-07-28
GH-11194: Use simple confirms in `AmqpOutboundEndpoint`

### LINKED ISSUE #11194: Options to use simple publisher confirms in AMQP outbound endpoints
### Discussed in https://github.com/spring-projects/spring-integration/discussions/11187

<sup>Originally posted by **rainerfrey-inxmail** July 17, 2026</sup>
Spring AMQP offers a simple publisher confirms mechanism using scoped operations and the `waitForConfirms()` methods on the RabbitTemplate. This is already used with multiSend for payloads that are iterables of messages only.
Would it be possible to add  a configuration option that causes the send method to use this mechanism for single / general messages? (Except for team resources), what would be the problem with that?</div>

xxxxxxjun:
Fixes #11194

`waitForConfirm` is ignored for single messages with `ConfirmType.SIMPLE`, since it only acts when the message carries `CorrelationData`. As you suggested in #11187, the send now runs inside a `RabbitTemplate.invoke()` scope with `waitForConfirmsOrDie()` at the end when `waitForConfirm` is true and there is no correlation data, the same way `multiSend` does.

Two decisions the issue does not cover:

- The branch is also guarded on `isSimplePublisherConfirms()`, so `CORRELATED` and `NONE` factories keep their current behavior. `setMultiSend()` and `BoundRabbitChannelAdvice` already require the same.
- `endpointInit()` warns when `waitForConfirm` is set on a factory with no publisher confirms at all, which was previously ignored without a signal.

Documented on the option: a returned message does not fail the send, since the broker acks an unroutable mandatory publish; and the scoped send cannot be combined with a `BoundRabbitChannelAdvice` on the same `RabbitTemplate`.

### DIFF
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests2.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests2.java
@@ -53,6 +53,7 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 /**
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Jun Cho
  *
  * @since 5.2
  *
@@ -63,14 +64,18 @@ public class AmqpOutboundEndpointTests2 implements RabbitTestContainer {
 
 	static final String QUEUE_TEST_CONFIRM_OK = "testConfirmOk";
 
+	static final String QUEUE_TEST_SIMPLE_CONFIRMS = "testSimpleConfirms";
+
 	@BeforeAll
 	static void initQueue() throws IOException, InterruptedException {
 		RABBITMQ.execInContainer("rabbitmqadmin", "declare", "queue", "name=" + QUEUE_TEST_CONFIRM_OK);
+		RABBITMQ.execInContainer("rabbitmqadmin", "declare", "queue", "name=" + QUEUE_TEST_SIMPLE_CONFIRMS);
 	}
 
 	@AfterAll
 	static void deleteQueue() throws IOException, InterruptedException {
 		RABBITMQ.execInContainer("rabbitmqadmin", "delete", "queue", "name=" + QUEUE_TEST_CONFIRM_OK);
+		RABBITMQ.execInContainer("rabbitmqadmin", "delete", "queue", "name=" + QUEUE_TEST_SIMPLE_CONFIRMS);
 	}
 
 	@Test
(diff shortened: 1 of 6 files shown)


spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests2.java
```diff
@@ -118,6 +123,32 @@ void testWithReject(@Autowired IntegrationFlow flow, @Autowired RabbitAdmin admi
...
+	void simpleConfirmsWithReject(@Autowired IntegrationFlow simpleConfirmsFlow, @Autowired RabbitAdmin admin,
+			@Autowired RabbitTemplate template) {
+
+		Queue queue = QueueBuilder.nonDurable().autoDelete().maxLength(1L).overflow(Overflow.rejectPublish).build();
+		admin.declareQueue(queue);
+		try {
+			simpleConfirmsFlow.getInputChannel()
+					.send(new GenericMessage<>("test", Collections.singletonMap("rk", queue.getName())));
+			assertThatThrownBy(() -> simpleConfirmsFlow.getInputChannel()
+					.send(new GenericMessage<>("test", Collections.singletonMap("rk", queue.getName()))))
```
artembilan: Looks like we can have a `queue.getName()` as a variable to reuse.
As well as the whole message is used twice.

Can you explain also, please, how this test works?
What do we try to prove?
That `template.receive()` looks similar to the one in the previous test.

Thanks

artembilan (changes requested):
Looks promising!
Please, consider to add a section into the `whats-new.adoc`.
You can find example in from the previous one which is now `changes-7.0-7.1.adoc`

Thanks again!

xxxxxxjun:
Hi @artembilan, thanks for the review! Added an AMQP section to `whats-new.adoc`.

spring-integration-amqp/src/test/java/org/springframework/integration/amqp/outbound/AmqpOutboundEndpointTests2.java
xxxxxxjun: Thanks, extracted both. I also dropped the `template.receive()` assertion: the first `send()` returning already proves the message was accepted, so it wasn't adding anything here.

What the test proves is that the confirm is actually awaited. Before this change, `waitForConfirm` with no correlation data skipped the wait, so the broker nack was discarded and both sends succeeded. The test fails without the fix.

It differs from `testWithReject` in where the exception comes from: here the nack surfaces from `waitForConfirmsOrDie()` as an `AmqpIOException` rather than the `AmqpException` the correlated path builds. I narrowed the assertion to that type to make it explicit.

`simpleConfirmsOk` passes without the fix too. It's only there to check that wrapping the send in `invoke()` doesn't break the exchange, routing key and header mapping.

xxxxxxjun:
Hi @artembilan, CI failed due to a GitHub runner issue earlier, so I tried a push. I've reset the branch back to the clean original commit, and it's ready for re-check when you have time. Thanks!

xxxxxxjun:
The build failed on `simpleConfirmsWithReject`: the second send did not throw, so the queue was not at its limit when it was published.

The most likely reason is the test setup rather than the change. The reject queue is declared through the `admin` bean, which is on the correlated connection factory, while the flow publishes through another one, so the first publish can go unrouted.

I couldn't reproduce it locally, so I matched `testWithReject` and declared the queue with a `RabbitAdmin` on the publishing connection factory.

PR #11207, opened 2026-07-28
GH-11196: Fix DI for Dynamically Registered Flow Beans

### LINKED ISSUE #11196: Dynamic IntegrationFlow registration skips dependency injection for additional and internal components

Observed with:

- Spring Integration 7.1.0
- Spring Framework 7.0.8
- Java 21
- Non-Boot application, although the problem does not appear to be Boot-specific

The same application behavior worked with:

- Spring Integration 6.3.3
- Spring Framework 6.1.12

The behavioral change appears to have been introduced in Spring Integration 6.4 by #9442 / #9450, when dynamic flow beans and their components were changed from BeanDefinition-backed registration to direct singleton registration.

**Describe the bug**

Components associated with a dynamically registered IntegrationFlow no longer receive normal Spring dependency injection.

This affects at least:

1. Additional beans registered using:

```
 integrationFlowContext.registration(flow)
    .addBean(beanName, bean)
    .register();
```

2. Additional components created by a custom DSL extension and added using the protected:

`BaseIntegrationFlowDefinition.addComponent(...)`

The objects are constructed using new and are subsequently registered by Spring Integration as beans. With Spring Integration 6.3.3, they were registered using a `BeanDefinition `with an instance supplier and then obtained through` BeanFactory.getBean()`. Consequently, normal dependency population occurred and fields or methods annotated with `@Autowired` were processed.

With Spring Integration 7.1.0, the objects are registered using approximately:

```
beanFactory.registerSingleton(beanName, bean);
beanFactory.initializeBean(bean, beanName);
```

`initializeBean()` performs initialization callbacks and bean post-processing, but it does not perform the dependency-population phase in which `@Autowired` fields and methods are processed.

As a result, the object is present in the bean factory, but its `@Autowired` fields remain null.

There is also a related proxying problem.` initializeBean()` may return a wrapped object, such as an AOP proxy, but the original object has already been registered in the singleton registry. In some registration paths, the return value is ignored completely. Therefore, bean post-processors that replace the original object cannot provide normal bean semantics.

Calling `autowireBean()` manually is not an equivalent durable workaround because it does not restore the complete definition-backed creation lifecycle, particularly replacement by proxies and other bean-definition-dependent processing.

**To Reproduce**

The simplest reproducer uses the public `IntegrationFlowRegistrationBuilder.addBean()` API.

```
package example;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.context.annotation.AnnotationConfigApplicationContext;
import org.springframework.integration.channel.DirectChannel;
import org.springframework.integration.dsl.context.IntegrationFlowContext;

class DynamicFlowAutowiringTests {

    @Test
    void additionalBeanShouldReceiveDependencyInjection() {
        try (AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext(
            TestConfiguration.class
        )) {
            SupportComponent supportComponent = new SupportComponent();
            IntegrationFlow flow = IntegrationFlow.from(new DirectChannel()).bridge().get();
            context.getBean(IntegrationFlowContext.class)
                .registration(flow)
                .id("testFlow")
                .addBean("testSupportComponent", supportComponent)
                .register();

            Assertions.assertTrue(context.containsBean("testSupportComponent"));

            /*
             * Passes with Spring Integration 6.3.3.
             * Fails with Spring Integration 7.1.0 because dependency is null.
             */
            Assertions.assertNotNull(supportComponent.dependency);
        }
    }

    @Configuration
    @EnableIntegration
    static class TestConfiguration {
        @Bean
        Dependency dependency() {
            return new Dependency();
        }
    }

    static final class Dependency {
    }

    static final class SupportComponent {
        @Autowired
        Dependency dependency;
    }
}
```

The same problem occurs for components added internally to a dynamic flow by a custom DSL extension:
```
package example;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.context.annotation.AnnotationConfigApplicationContext;
import org.springframework.integration.channel.DirectChannel;
import org.springframework.integration.dsl.IntegrationFlowExtension;
import org.springframework.integration.dsl.context.IntegrationFlowContext;

class DynamicFlowComponentAutowiringTests {

    @Test
    void componentAddedToDynamicFlowShouldReceiveDependencyInjection() {
        try (AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext(
            TestConfiguration.class
        )) {
            TestFlowExtension extension = new TestFlowExtension();
            IntegrationFlow flow = extension.channel(new DirectChannel()).addSupportComponent().bridge().get();
            context.getBean(IntegrationFlowContext.class)
                .registration(flow)
                .id("testFlow")
                .register();

            /*
             * Passes with Spring Integration 6.3.3.
             * Fails with Spring Integration 7.1.0 because dependency is null.
             */
            Assertions.assertNotNull(extension.supportComponent.dependency);
        }
    }

    static final class TestFlowExtension extends IntegrationFlowExtension<TestFlowExtension> {

        final SupportComponent supportComponent = new SupportComponent();

        TestFlowExtension addSupportComponent() {
            return addComponent(this.supportComponent, "testFlow.supportComponent");
        }
    }

    @Configuration
    @EnableIntegration
    static class TestConfiguration {
        @Bean
        Dependency dependency() {
            return new Dependency();
        }
    }

    static final class Dependency {
    }

    static final class SupportComponent {

        @Autowired
        Dependency dependency;
    }
}
```

**Expected behavior**

Beans and components registered as part of a dynamic IntegrationFlow should receive normal Spring bean management, as they did before Spring Integration 6.4:

- `@Autowired` field and method injection
- property population
- Aware callbacks
- initialization callbacks
- bean post-processing
- use of replacement objects returned by bean post-processors, including AOP proxies
- destruction callbacks when the dynamic flow is removed

The public `IntegrationFlowRegistrationBuilder.addBean()` documentation describes these objects as additional beans registered in the application context, including support components which need an application context. It is therefore surprising that these objects are registered as beans but do not receive dependency injection.

The same expectation applies to components contributed through the supported custom DSL extension mechanism using `IntegrationFlowExtension `and `addComponent()`.

**Additional analysis**

In Spring Integration 6.3.3, `StandardIntegrationFlowContext` registered dynamic flow objects approximately as follows:

```
AbstractBeanDefinition beanDefinition =
        BeanDefinitionBuilder
                .genericBeanDefinition(bean.getClass(), () -> bean)
                .getRawBeanDefinition();

beanDefinitionRegistry.registerBeanDefinition(
        beanName,
        beanDefinition);

beanFactory.getBean(beanName);
```

That caused the normal Spring bean-creation pipeline to run against the supplied instance.

In Spring Integration 7.1.0, registration is approximately:

```
beanFactory.registerSingleton(beanName, bean);
return beanFactory.initializeBean(bean, beanName);
```

Similarly, `IntegrationFlowBeanPostProcessor.registerSingleton()` performs:

```
beanFactory.registerSingleton(beanName, component);
beanFactory.initializeBean(component, beanName);
beanFactory.getBean(beanName);

```
This does not invoke dependency population. In the latter case, the return value from initializeBean() is also discarded.

I understand that singleton registration was introduced to address the merged bean-definition cache retention described in #9442. However, it also removed observable and useful bean-management semantics that existing dynamic-flow implementations relied on.

**Requested outcome**

Could dynamic flow registration preserve full Spring bean-management semantics while also avoiding the bean-definition cache problem?

Possible solutions might include:

- an officially supported definition-backed dynamic registration mode with complete cleanup;
- registration methods accepting a BeanDefinition, bean class, or managed supplier;
- a Spring Integration-managed child bean factory/application context for removable flows;
- another mechanism that performs dependency population and correctly retains replacement objects returned by bean post-processors.

At minimum, it would be useful to clarify the officially supported approach for dynamically created flow components that require `@Autowired` injection and AOP proxying.

The current change is difficult to work around at application level without either manually duplicating Spring’s initialization process or introducing a separate child application context for every dynamic flow.

artembilan:
Fixes: https://github.com/spring-projects/spring-integration/issues/11196

Since dynamic `IntegrationFlow` beans and their support components were
switched from `BeanDefinition`-backed registration to direct singleton
registration (to avoid retaining merged bean definitions), such beans
skip the property-population phase of the bean lifecycle, and any
object replacement performed by a `BeanPostProcessor` (e.g. an AOP
proxy) is silently discarded.
As a result:
- `@Autowired` fields/methods on beans added via
`IntegrationFlowContext.registration().addBean()`, or contributed by a
custom `IntegrationFlowExtension.addComponent()`, were never populated
- any proxy or other replacement object returned from
`initializeBean()` was lost because the raw, pre-processed instance had
already been bound in the singleton registry under that bean name

* Call `DefaultListableBeanFactory.autowireBean()` before
`initializeBean()` in `StandardIntegrationFlowContext.registerBean()`
and `IntegrationFlowBeanPostProcessor.registerSingleton()` to restore
`@Autowired` injection for singleton-registered dynamic flow components
* Register the singleton *after* `initializeBean()` completes, using its
return value, so a proxy (or other post-processor replacement) is what
ends up bound in the bean factory
* Add a clarifying comment on the existing `getBean()` calls explaining
they force early `FactoryBean#getObject()` resolution
* Add `ManualFlowTests.additionalBeanReceivesDependencyInjection()` and
`ManualFlowTests.additionalBeanPostProcessorReplacementIsPreserved()`
to verify the fix

### DIFF
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
@@ -41,7 +41,6 @@ import org.springframework.beans.factory.config.EmbeddedValueResolver;
 import org.springframework.beans.factory.support.AbstractBeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
 import org.springframework.beans.factory.support.BeanDefinitionOverrideException;
-import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.beans.factory.support.DefaultListableBeanFactory;
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.ApplicationContext;
@@ -266,7 +265,7 @@ public class IntegrationFlowBeanPostProcessor
 					}
 					else if (component instanceof FixedSubscriberChannel fixedSubscriberChannel) {
 						String channelBeanName = fixedSubscriberChannel.getComponentName();
-						if ("Unnamed fixed subscriber channel" .equals(channelBeanName)) {
+						if ("Unnamed fixed subscriber channel".equals(channelBeanName)) {
 							channelBeanName = flowNamePrefix + "channel" +
 									BeanFactoryUtils.GENERATED_BEAN_NAME_SEPARATOR + channelNameIndex++;
 						}
@@ -518,7 +517,8 @@ public class IntegrationFlowBeanPostProcessor
 			this.beanFactory.registerDependentBean(parentName, beanName);
 		}
 
-		((BeanDefinitionRegistry) this.beanFactory).registerBeanDefinition(beanName, beanDefinition);
+		this.beanFactory.registerBeanDefinition(beanName, beanDefinition);
+		// Force early FactoryBean#getObject() resolution, e.g. for ConsumerEndpointFactoryBean et al.
 		this.beanFactory.getBean(beanName);
 	}
 
@@ -535,8 +535,11 @@ public class IntegrationFlowBeanPostProcessor
 			this.beanFactory.registerDependentBean(parentName, beanName);
 		}
 
-		this.beanFactory.registerSingleton(beanName, component);
-		this.beanFactory.initializeBean(component, beanName);
+		this.beanFactory.autowireBean(component);
+		Object initializedComponent = this.beanFactory.initializeBean(component, beanName);
+		// Register the post-processed (e.g. possibly proxied) instance, not the raw one.
+		this.beanFactory.registerSingleton(beanName, initializedComponent);
+		// Force early FactoryBean#getObject() resolution, e.g. for ConsumerEndpointFactoryBean et al.
 		this.beanFactory.getBean(beanName);
 	}
 
--- a/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/StandardIntegrationFlowContext.java
+++ b/spring-integration-core/src/main/java/org/springframework/integration/dsl/context/StandardIntegrationFlowContext.java
@@ -164,8 +164,11 @@ public final class StandardIntegrationFlowContext implements IntegrationFlowCont
 				componentSourceAware.setComponentDescription(description);
 			}
 		}
-		this.beanFactory.registerSingleton(beanName, bean);
-		return (B) this.beanFactory.initializeBean(bean, beanName);
+		this.beanFactory.autowireBean(bean);
+		B initializedBean = (B) this.beanFactory.initializeBean(bean, beanName);
+		// Register the post-processed (e.g. possibly proxied) instance, not the raw one.
+		this.beanFactory.registerSingleton(beanName, initializedBean);
+		return initializedBean;
 	}
 
 	/**
@@ -175,8 +178,7 @@ public final class StandardIntegrationFlowContext implements IntegrationFlowCont
 	 * @return the IntegrationFlowRegistration for provided {@code id} or {@code null}
 	 */
 	@Override
-	@Nullable
-	public IntegrationFlowRegistration getRegistrationById(String flowId) {
+	public @Nullable IntegrationFlowRegistration getRegistrationById(String flowId) {
 		return this.registry.get(flowId);
 	}
 
--- a/spring-integration-core/src/test/java/org/springframework/integration/dsl/manualflow/ManualFlowTests.java
+++ b/spring-integration-core/src/test/java/org/springframework/integration/dsl/manualflow/ManualFlowTests.java
@@ -36,10 +36,13 @@ import java.util.function.Supplier;
 import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Flux;
 
+import org.springframework.aop.framework.ProxyFactory;
+import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.factory.BeanCreationNotAllowedException;
 import org.springframework.beans.factory.ListableBeanFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
+import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.beans.factory.support.BeanDefinitionOverrideException;
 import org.springframework.beans.factory.support.DefaultListableBeanFactory;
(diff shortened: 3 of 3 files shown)


cppwfs:
Why are we creating branches on repository and opening PR's here?   I thought we created PR's from our local repo.

spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
```diff
@@ -535,8 +535,11 @@ private void registerSingleton(Object component, String beanName, @Nullable Obje
 			this.beanFactory.registerDependentBean(parentName, beanName);
 		}
 
-		this.beanFactory.registerSingleton(beanName, component);
-		this.beanFactory.initializeBean(component, beanName);
+		this.beanFactory.autowireBean(component);
+		Object initializedComponent = this.beanFactory.initializeBean(component, beanName);
+		// Register the post-processed (e.g. possibly proxied) instance, not the raw one.
+		this.beanFactory.registerSingleton(beanName, initializedComponent);
+		// Force early FactoryBean#getObject() resolution, e.g. for ConsumerEndpointFactoryBean et al.
```
cppwfs, line 542: I think we have an issue where we are not handling all the cases.  `IntegrationFlowBeanPostProcessor` has an issue where at line 179 we get a pointer to the `endpoint`.   The code that is introduced creates a new instance later on but the original pointer to the `endpoint` is being used.    Thus we need to hand back the `registeredEndpoint` from the `registerComponent` method on line 197 and then use it on line 199 instead of the `endPoint` which is the original.

cppwfs (changes requested):
Looks good, but I think I found a bug.   In all honesty it was AI that found it, I had grok it and play with it so that I could explain it.

spring-integration-core/src/main/java/org/springframework/integration/dsl/context/IntegrationFlowBeanPostProcessor.java
artembilan: OK.
Sounds too complicated and probably very academic.
Can you ask your AI to reproduce what it think is the problem?
Thanks

artembilan: Well, the recommendation is to not proxy `MessageHandler` manually, but rather use:
```
	/**
	 * Configure a list of {@link Advice} objects to be applied, in nested order, to the
	 * endpoint's handler. The advice objects are applied only to the handler.
	 * @param advice the advice chain.
	 * @return the endpoint spec.
	 */
	public S advice(Advice... advice) {
```
So, the claim by AI does not fully apply.
There is a corner case when raw object is given to DSL.
But if you follow standard Spring recommendations to wire beans with beans, then you are safe.
More over it is better to reuse as many beans for dynamic integration flows as possible.
It is better to declare `MessageHandler` as bean and have injected into that dynamic integration flow.

cppwfs: I see.   So in this case, we can accept this PR, since the edge case is bad practice.   WDYT?

artembilan: Yes. We may revise for particular request, but that would take some other significant changes .
Probably not related to this change

cppwfs (approved):
Thanks for fixing this.   After our discussion, the pattern identified is not considered good practice and not a part of this fix.   

LGTM!

PR #11214, opened 2026-07-29
Add file configurations to sensitive metadata

### LINKED ISSUE #11199: Document `FileHeaders.ORIGINAL_FILE` as a Sensitive Metadata
See https://docs.spring.io/spring-integration/reference/security.html#sensitive-metadata.
And `FileWritingMessageHandler`, especially its `setDeleteSourceFiles()`.

cppwfs:
- Add FileHeaders and other configurations from FileWritingMessageHandler to sensitive-metadata section where applicable

Fixes: https://github.com/spring-projects/spring-integration/issues/11199

### DIFF
--- a/src/reference/antora/modules/ROOT/pages/security.adoc
+++ b/src/reference/antora/modules/ROOT/pages/security.adoc
@@ -162,6 +162,15 @@ The mentioned above validation and filtering might be applied to them before the
 
 - `spring-integration-file`
  * `file_name`
+ * `delete_source_files`
+ * `destination_directory`
+ * `destination_directory_expression`
+ * `original_file`
+ * `relative_path`
+ * `rename_to`
+ * `remote_diretory`
+ * `remote_file`
+
 
 - `spring-integration-ip`
  * `ip_ackTo`
--- a/src/reference/antora/modules/ROOT/pages/security.adoc
+++ b/src/reference/antora/modules/ROOT/pages/security.adoc
@@ -168,7 +168,7 @@ The mentioned above validation and filtering might be applied to them before the
  * `file_remoteDirectory`
  * `file_remoteFile`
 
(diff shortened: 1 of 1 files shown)


src/reference/antora/modules/ROOT/pages/security.adoc
```diff
@@ -162,6 +162,14 @@ The mentioned above validation and filtering might be applied to them before the
 
 - `spring-integration-file`
  * `file_name`
+ * `file_originalFile`
+ * `file_relativePath`
+ * `file_renameTo`
+ * `file_remoteDirectory`
+ * `file_remoteFile`
+ * `file_setModified`
```
artembilan: Not clear how `timestamp` could be exploited.

src/reference/antora/modules/ROOT/pages/security.adoc
```diff
@@ -162,6 +162,14 @@ The mentioned above validation and filtering might be applied to them before the
...
+
+NOTE: The following settings in `FileWritingMessageHandler` are also sensitive: `deleteSourceFiles`, `destinationDirectory`, `destinationDirectoryExpression`.
```
artembilan: How is `deleteSourceFiles` sensitive?
The `destinationDirectory` is also static property.
How that can be exploited?

artembilan (changes requested):

src/reference/antora/modules/ROOT/pages/security.adoc
cppwfs: I'll remove destinationDirectory.    
But `setDeleteSourceFiles` was explicitly mentioned in the Issue.

artembilan: You cannot say that here in the OSS, but `deleteSourceFiles` is the trigger for around `file_originalFile` header - exactly the one being sensitive according to the input.
Not `deleteSourceFiles` property itself which is still `false` by default.

PR #11240, opened 2026-08-02
GH-11235: Fix form data read in MultipartAware converter

### LINKED ISSUE #11235: Investigate flaky/failing `HttpDslTests.testHttpProxyFlow()`
`testHttpProxyFlow()` currently fails:

```
java.lang.AssertionError: Response content expected:<FOO> but was:<Error>
	at org.springframework.integration.http.dsl.HttpDslTests.testHttpProxyFlow(HttpDslTests.java:139)
```

with this warning in stdout:

```
Resolved [org.springframework.http.converter.HttpMessageNotReadableException: No multipart boundary found in Content-Type: "application/x-www-form-urlencoded"]
```

The test wires a mocked `ClientHttpRequestFactory` into `HttpRequestExecutingMessageHandler` by reflectively pulling it out of `RestTestClient` (`TestUtils.getPropertyValue(restTestClient, "restClient.clientRequestFactory")`), then re-issues the proxied request through `MockMvc`. The multipart-boundary warning suggests the proxied outbound request is no longer built the way the internal gateway expects — possibly due to an internal `RestTestClient`/`RestClient` structural change in a recent Spring Framework version breaking the reflective field path, rather than an actual bug in `HttpRequestExecutingMessageHandler` itself.

Disabled for now (see commit disabling it) to keep the build green; needs a proper investigation into the actual request being proxied and why it now round-trips as `Error` instead of `FOO`.

kdelay:
Fixes: gh-11235

`HttpDslTests.testHttpProxyFlow()` fails because of a regression in `MultipartAwareFormHttpMessageConverter`, not because of the reflective `RestTestClient` field path.

The converter used to delegate to an `AllEncompassingFormHttpMessageConverter`, which reads and writes both `application/x-www-form-urlencoded` and `multipart/form-data`. It now delegates to a `MultipartHttpMessageConverter`, which handles only `multipart/form-data`, so any form-urlencoded request body fails with:

```
org.springframework.http.converter.HttpMessageNotReadableException:
No multipart boundary found in Content-Type: "application/x-www-form-urlencoded"
```

The proxied request in that test is a `POST` to `/service/internal` with a form-urlencoded body, which is why it surfaced there. Any HTTP inbound gateway or channel adapter receiving a form post is affected, since `HttpRequestHandlingEndpointSupport` registers this converter first.

Changes:

* non-multipart reads and writes go to a `FormHttpMessageConverter`; the `MultipartHttpMessageConverter` is kept for multipart, selected by the same `isMultipart()` rule the old delegate used (a `multipart` content type, or a non-`String` value in the map)
* `getSupportedMediaTypes()` reports the media types of both delegates and `canWrite()` accepts both; with the multipart-only delegate, `application/x-www-form-urlencoded` had disappeared from both
* the read delegates against a `MultiValueMap` target type. `HttpRequestHandlingEndpointSupport.extractRequestBody()` falls back to `byte[]` when no request payload type is configured, and `FormHttpMessageConverter.read()` returns a single-value `Map` for a non-`MultiValueMap` type, which then fails the cast in this converter.
* `setCharset()` now applies to both delegates

Verification:

* `MultipartAwareFormHttpMessageConverterTests` is new; with only the test files applied and the converter reverted, all four of its tests fail, as does `testHttpProxyFlow()`
* `./gradlew :spring-integration-http:test` — 235 tests, 0 failures
* `./gradlew :spring-integration-webflux:test` — 35 tests, 0 failures
* `checkstyleMain` and `checkstyleTest` pass for the `http` module

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/converter/MultipartAwareFormHttpMessageConverter.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/converter/MultipartAwareFormHttpMessageConverter.java
@@ -18,14 +18,18 @@ package org.springframework.integration.http.converter;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.core.ResolvableType;
 import org.springframework.http.HttpInputMessage;
 import org.springframework.http.HttpOutputMessage;
 import org.springframework.http.MediaType;
+import org.springframework.http.converter.FormHttpMessageConverter;
 import org.springframework.http.converter.HttpMessageConverter;
 import org.springframework.http.converter.HttpMessageNotReadableException;
 import org.springframework.http.converter.HttpMessageNotWritableException;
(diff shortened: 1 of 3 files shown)


artembilan:
Duplicate of: https://github.com/spring-projects/spring-integration/pull/11236?

give you AI a hint to not fix if it is present on the issue already .

PR #11241, opened 2026-08-03
Modularize Gradle build into per-module build files

cppwfs:
- Add a standalone build.gradle to all 42 spring-integration-* modules, each applying the shared conventions plus its own module-specific dependencies and plugin needs (protobuf for core/grpc, groovy plugin for the groovy module)
- Remove all per-module project('...') {} blocks and the configure(javaProjects) {} block from root build.gradle, cutting it from 1155 to roughly 300 lines
- Keep spring-integration-bom's configuration in root build.gradle since it is a java-platform aggregator, not a java-conventions consumer
- Defer schemaZip and distZip's subproject sourceSets/jar access into gradle.projectsEvaluated to fix task-registration-order failures that surfaced once modules stopped being eagerly configured inline with root
- Extract shared java/kotlin/checkstyle/publishing config from root build.gradle into buildSrc convention plugins (spring-integration.java-conventions and spring-integration.protobuf-conventions)
- Implement convention plugins as Plugin<Project> classes that delegate to resource .gradle scripts, working around a Gradle 9.6.1 bug where the groovy-gradle-plugin precompiled-script adapter throws UnknownServiceException when the same plugin is applied to more than one project


buildSrc/src/main/resources/conventions/spring-integration-java-conventions.gradle
```diff
@@ -0,0 +1,240 @@
...
+    testImplementation 'io.projectreactor:reactor-test'
+    testImplementation 'org.testcontainers:testcontainers-junit-jupiter'
+
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
+
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-core'
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-jcl'
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-slf4j2-impl'
```
artembilan: No, I don't want anything in `buildSrc`.
This is exta code to support and a bit confusing.
But it looks like it is spreading dependency management.
Why `configure(javaProjects)` doesn't work for us?
I don't want any custom Gradle plugins.

spring-integration-cassandra/build.gradle
```diff
@@ -0,0 +1,12 @@
+plugins {
+    id 'spring-integration.java-conventions'
```
artembilan: Doesn't look like this custom plugin helps enough.
We still have to copy/paste it every where.
May we just split root config per modules and leave everything common as is?

That claim about decreasing size is a bit not fair since we are introduce much more complexity into that `buildSrc`

artembilan (changes requested):

cppwfs:
I took a different tack and so I started from scratch.   So there is only one commit.

artembilan (approved):
Very cool!
Merging.
We can come back to any leftover concerns with new Gradle later on.

Thank you!

PR #11242, opened 2026-08-04
Resolve Camel's Nullability errors

### LINKED ISSUE #11221: Upgrade `org.apache.camel:camel-bom` to `4.21.0`

cppwfs:
- Camel has introduced JSpecify to establish nullability policy

Fixes: https://github.com/spring-projects/spring-integration/issues/11221


spring-integration-camel/src/main/java/org/springframework/integration/camel/outbound/CamelMessageHandler.java
```diff
@@ -256,12 +259,18 @@ private Exchange prepareInExchange(Endpoint endpoint, ExchangePattern exchangePa
 	@Nullable
 	private AbstractIntegrationMessageBuilder<?> buildReply(ExchangePattern exchangePattern, Exchange result) {
 		if (result.isFailed()) {
-			throw CamelExecutionException.wrapCamelExecutionException(result, result.getException());
+			Exception exception = result.getException();
+			Assert.state(exception != null, "Exchange has failed and has no exception");
```
artembilan: I think this is an overhead.
See the contract: 
```
    /**
     * Returns the exception associated with this exchange
     *
     * @return the exception (or null if no faults)
     */
    Exception getException();
```

So, `f (result.isFailed()) {`, then `getException()` is not null.
And that is very true by the implementation:
```
    public boolean isFailed() {
        return exception != null;
    }
```

We may add `@SuppressWarnings("NullAway")`, but probably would be more clean to use `getException()` instead of ` isFailed()`.

artembilan (changes requested):

PR #11243, opened 2026-08-05
Deprecate RestTemplate-based Http outbound-gateway config

### LINKED ISSUE #11234: Fix `RestTemplate`-related deprecations in the HTTP module

cppwfs:
- Deprecate RestTemplate Error Handling methods
- Add RestClient.ResponseSpec.ErrorHandler-based errorHandler overloads.
- Refactor HttpAdapterParsingUtils to share validation across rest-template and rest-client reference kinds.
- Migrate OutboundResponseTypeTests and related XML/DSL tests to configure RestClient beans instead of RestTemplate.
- Remove obsolete withUrlExpressionAndTemplate XML test case.
- Deprecate user access of RestClient deprecated functions

Fixes: https://github.com/spring-projects/spring-integration/issues/11234


spring-integration-http/src/main/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandler.java
```diff
@@ -227,10 +233,35 @@ public void setErrorHandler(ResponseErrorHandler errorHandler) {
...
+	}
+
+	/**
+	 * Set the status {@link Predicate} and {@link RestClient.ResponseSpec.ErrorHandler} for the underlying
+	 * {@link RestClient}.
+	 * @param predicate the predicate for the response status to match for this error handler.
+	 * @param errorHandler The error handler.
+	 * @since 7.2
+	 */
+	public void setErrorHandler(Predicate<HttpStatusCode> predicate, RestClient.ResponseSpec.ErrorHandler errorHandler) {
```
artembilan: This cannot be called `set` since it has two arguments.
I guess like that `defaultStatusHandler` should be good.

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandler.java
```diff
@@ -227,10 +233,35 @@ public void setErrorHandler(ResponseErrorHandler errorHandler) {
 	}
 
 	/**
-	 * Set a list of {@link HttpMessageConverter}s to be used by the underlying {@link RestTemplate}.
+	 * Set the {@link RestClient.ResponseSpec.ErrorHandler} for the underlying {@link RestClient}, applied to any
+	 * {@link HttpStatusCode#isError() error} response status.
+	 * @param errorHandler The error handler.
+	 * @since 7.2
+	 */
+	public void setErrorHandler(RestClient.ResponseSpec.ErrorHandler errorHandler) {
```
artembilan: I think this one also should be called `setDeafaultStatusHandler` according to the underlying `RestClient` API.

spring-integration-http/src/test/java/org/springframework/integration/http/config/OutboundResponseTypeTests-context-encoding-mode-fail.xml
```diff
@@ -9,7 +9,7 @@
 		<constructor-arg value="org.springframework.web.client.RestTemplate" type="java.lang.Class"/>
 	</bean>
 
-	<int-http:outbound-gateway url="/fake"
+	<int-http:outbound-gateway url="/testRestTemplate"
```
artembilan: I think we have agreed to remove all the `RestTemplate`-based tests.
Can we do that for this one, too?

spring-integration-http/src/test/java/org/springframework/integration/http/config/OutboundResponseTypeTests.java
```diff
@@ -86,11 +92,52 @@ public class OutboundResponseTypeTests {
...
+							}));
+		}
+
+		@Bean
+		public static MockRestServiceServer mockServerWithConverters(RestClient.Builder restClientWithConvertersBuilder) {
+			return MockRestServiceServer.bindTo(restClientWithConvertersBuilder).build();
+		}
+
+		@Bean
+		public static RestClient restClientWithConverters(RestClient.Builder restClientWithConvertersBuilder) {
```
artembilan: Why all of these beans are `static`?
I'm not talking already about redundant `public`

spring-integration-http/src/test/java/org/springframework/integration/http/dsl/HttpDslTests.java
```diff
@@ -425,13 +430,27 @@ public Validator customValidator() {
...
+			if (response.getStatusCode().is4xxClientError()) {
+				ex = HttpClientErrorException.create("400 error", response.getStatusCode(), response.getStatusText(),
+						response.getHeaders(), body, charset);
+			}
+			else if (response.getStatusCode().is5xxServerError()) {
+				ex = HttpServerErrorException.create("500 error", response.getStatusCode(), response.getStatusText(),
+						response.getHeaders(), body, charset);
+			}
+			else {
+				ex = new UnknownHttpStatusCodeException("unknown error", response.getStatusCode().value(),
```
artembilan: Let's see `MockClientHttpResponse` and `StatusHandler.createException()` instead of this copy/paste logic!

artembilan (changes requested):
Looks great!

We need to mention in the XSD that `rest-template` is deprecated.

Please, revise also documentation on the matter.

I think whats-new entry also would be great edition.

Thanks

artembilan (changes requested):
Missed to ask since that was not mentioned anyhow in the issue and in this change set.

Please, mention such a change in the `whats-new` and we probably also need to revise all those `RestTemplate` wording in the `http` docs.

Thanks

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandler.java
```diff
@@ -227,10 +234,37 @@ public void setErrorHandler(ResponseErrorHandler errorHandler) {
...
+	}
+
+	/**
+	 * Set the status {@link Predicate} and {@link RestClient.ResponseSpec.ErrorHandler} for the underlying
+	 * {@link RestClient}.
+	 * @param predicate the predicate for the response status to match for this error handler.
+	 * @param errorHandler The error handler.
+	 * @since 7.2
+	 */
+	public void setDefaultStatusHandler(Predicate<HttpStatusCode> predicate,
```
artembilan: I think I said before that this method style (two-arguments) cannot be called as setter.
Consider to name it just `defaultStatusHandler()`

spring-integration-http/src/test/java/org/springframework/integration/http/dsl/HttpDslTests.java
artembilan: Was not addressed. 😢

src/reference/antora/modules/ROOT/pages/http/outbound.adoc
```diff
@@ -37,6 +37,8 @@ You can configure those converters as well as the `ClientHttpRequestFactory` ins
 By default, the HTTP request is generated by using an instance of `SimpleClientHttpRequestFactory`, which uses the JDK `HttpURLConnection`.
 Use of the Apache Commons HTTP Client is also supported through `CommonsClientHttpRequestFactory`, which you can inject (as shown earlier).
 
+Starting with version 7.2, you can configure a `RestClient.ResponseSpec.ErrorHandler` (instead of the deprecated `RestTemplate`-based `ResponseErrorHandler`) via the `setDefaultStatusHandler(RestClient.ResponseSpec.ErrorHandler)` or `setDefaultStatusHandler(Predicate<HttpStatusCode>, RestClient.ResponseSpec.ErrorHandler)` methods (or the matching `errorHandler()` overloads on the Java DSL `Http.outboundGateway()`/`Http.outboundChannelAdapter()`), to customize how the underlying `RestClient` reacts to error response statuses.
```
artembilan: When you rename that `setDefaultStatusHandler(Predicate<HttpStatusCode>`, please, don't forget here, too.

artembilan (changes requested):
Very cool!

spring-integration-http/src/test/java/org/springframework/integration/http/dsl/HttpDslTests.java
cppwfs: `RestClient.ResponseSpec` has no nested `StatusHandler` type at all — only `onStatus(...)` methods. So `RestClient.ResponseSpec.StatusHandler` as a named path doesn't exist; `StatusHandler `is a sibling top-level class in the same package, not nested under `ResponseSpec`.

artembilan: I don't think this is what I've asked...

My point was about this `if..else` logic which I think could be simplified to:
```
public void handle(HttpRequest request, ClientHttpResponse response) throws IOException {
    MockClientHttpResponse mockResponse = new MockClientHttpResponse("Error".getBytes(), response.getStatusCode());
    throw StatusHandler.createException(mockResponse, List.of());
}
```

artembilan (changes requested):

spring-integration-http/src/test/java/org/springframework/integration/http/dsl/HttpDslTests.java
cppwfs: Thank you for the clarification.   The issue is that  `StatusHandler` is not a public class. https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/client/StatusHandler.java#L49

artembilan: Doh!

Thank you for being patience!
And sorry for the noise.
I missed that part because of:
```
public static RestClientResponseException createException(
			ClientHttpResponse response, List<HttpMessageConverter<?>> converters) throws IOException {
```
I wouldn't do that in the non-`public` class.

OK. No reason to pursue minimalism then!
Merging 😄

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandler.java
artembilan: M-m-m. Hold on.
I still have concerns to this method name.
Please, advice.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -30,3 +30,10 @@ The `wait-for-confirm` option on the outbound channel adapter now also works wit
 When the message has no correlation data, the send is performed within the scope of a `RabbitTemplate.invoke()` operation and `waitForConfirmsOrDie()` is used.
 A warning is now logged when the option is set but the connection factory supports no publisher confirms at all.
 See xref:amqp/outbound-channel-adapter.adoc[] for more information.
+
+[[x7.2-http-changes]]
+=== HTTP Support Changes
+
+The `RestTemplate`-based configuration for the HTTP outbound gateway and outbound channel adapter (the `rest-template` attribute, and the respective Java DSL and constructor options) is now deprecated in favor of `RestClient`-based configuration.
+The `HttpRequestExecutingMessageHandler` also gains `setDefaultStatusHandler(RestClient.ResponseSpec.ErrorHandler)` and `setDefaultStatusHandler(Predicate<HttpStatusCode>, RestClient.ResponseSpec.ErrorHandler)` methods (with matching Java DSL `errorHandler()` overloads) to configure `RestClient`-based error handling.
+See xref:http/outbound.adoc[] for more information.
```
artembilan: We may have other HTTP module changes in this version.
I don't see a reason to be so fine-grained, or even specify links for each change.
Just have one sentence in the end pointing to the whole HTTP chapter!

artembilan (changes requested):

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
```diff
@@ -123,8 +123,8 @@ public HttpMessageHandlerSpec errorHandler(ResponseErrorHandler errorHandler) {
 	 * @since 7.2
 	 */
 	public HttpMessageHandlerSpec defaultStatusHandler(RestClient.ResponseSpec.ErrorHandler errorHandler) {
-		Assert.isTrue(!isClientSet(), "the 'errorHandler' must be specified on the provided client");
-		this.target.setDefaultStatusHandler(errorHandler);
+		Assert.isTrue(!isClientSet(), "the 'defaultErrorHandler' must be specified on the provided client");
```
artembilan: `defaultStatusHandler` ?

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
```diff
@@ -123,8 +123,8 @@ public HttpMessageHandlerSpec errorHandler(ResponseErrorHandler errorHandler) {
...
+		this.target.defaultStatusHandler(errorHandler);
```
artembilan: I don't think that is what I've asked.
It is indeed a setter if it has a single argument.
My ask was only for two-argument method to be inconsistent with Java beans specification.

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
```diff
@@ -136,11 +136,11 @@ public HttpMessageHandlerSpec defaultStatusHandler(RestClient.ResponseSpec.Error
 	 * @return the spec
 	 * @since 7.2
 	 */
-	public HttpMessageHandlerSpec errorHandler(Predicate<HttpStatusCode> predicate,
+	public HttpMessageHandlerSpec defaultErrorHandler(Predicate<HttpStatusCode> predicate,
 			RestClient.ResponseSpec.ErrorHandler errorHandler) {
 
-		Assert.isTrue(!isClientSet(), "the 'errorHandler' must be specified on the provided client");
-		this.target.setDefaultStatusHandler(predicate, errorHandler);
+		Assert.isTrue(!isClientSet(), "the 'defaultErrorHandler' must be specified on the provided client");
```
artembilan: DITTO

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandler.java
```diff
@@ -239,8 +239,8 @@ public void setErrorHandler(ResponseErrorHandler errorHandler) {
 	 * @param errorHandler The error handler.
 	 * @since 7.2
 	 */
-	public void setDefaultStatusHandler(RestClient.ResponseSpec.ErrorHandler errorHandler) {
-		setDefaultStatusHandler(HttpStatusCode::isError, errorHandler);
+	public void defaultStatusHandler(RestClient.ResponseSpec.ErrorHandler errorHandler) {
```
artembilan: See? Why has this turned to be not a setter on Java bean class?

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -717,6 +718,48 @@ public void failWhenSetRequestFactoryWithExternalRestClient() {
 				.withMessageContaining("externally configured RestClient");
 	}
 
+	@Test
+	public void defaultStatusHandlerAppliesOnlyToMatchingPredicate() throws IOException {
+		ClientHttpRequestFactory requestFactory = mock(ClientHttpRequestFactory.class);
+		ClientHttpRequest clientRequest = mock(ClientHttpRequest.class);
```
artembilan: No arguments for `mock()` API, please.

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -717,6 +718,48 @@ public void failWhenSetRequestFactoryWithExternalRestClient() {
...
+		when(requestFactory.createRequest(any(URI.class), any(HttpMethod.class))).thenReturn(clientRequest);
```
artembilan: Cannot it be just `any(), any()` ?
I mean doesn't like there any other contract in the `ClientHttpRequestFactory` to not be able to infer.

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -717,6 +718,48 @@ public void failWhenSetRequestFactoryWithExternalRestClient() {
...
+		handler.setRequestFactory(requestFactory);
+		handler.setHttpMethod(HttpMethod.GET);
+		handler.setExpectedResponseType(String.class);
+		handler.defaultStatusHandler(HttpStatusCode::is4xxClientError,
+				(request, response) -> handledStatus.set(response.getStatusCode()));
+		setBeanFactory(handler);
+		handler.afterPropertiesSet();
+		handler.setOutputChannel(new QueueChannel());
+
+		ClientHttpResponse response = mock(ClientHttpResponse.class);
```
artembilan: DITTO

src/reference/antora/modules/ROOT/pages/http/outbound.adoc
```diff
@@ -37,7 +37,7 @@ You can configure those converters as well as the `ClientHttpRequestFactory` ins
...
 
-Starting with version 7.2, you can configure a `RestClient.ResponseSpec.ErrorHandler` (instead of the deprecated `RestTemplate`-based `ResponseErrorHandler`) via the `setDefaultStatusHandler(RestClient.ResponseSpec.ErrorHandler)` or `setDefaultStatusHandler(Predicate<HttpStatusCode>, RestClient.ResponseSpec.ErrorHandler)` methods (or the matching `errorHandler()` overloads on the Java DSL `Http.outboundGateway()`/`Http.outboundChannelAdapter()`), to customize how the underlying `RestClient` reacts to error response statuses.
+Starting with version 7.2, you can configure a `RestClient.ResponseSpec.ErrorHandler` (instead of the deprecated `RestTemplate`-based `ResponseErrorHandler`) via the `defaultStatusHandler(RestClient.ResponseSpec.ErrorHandler)` or `defaultStatusHandler(Predicate<HttpStatusCode>, RestClient.ResponseSpec.ErrorHandler)` methods (or the matching  `defaultErrorHandler()` overloads on the Java DSL `Http.outboundGateway()`/`Http.outboundChannelAdapter()`), to customize how the underlying `RestClient` reacts to error response statuses.
```
artembilan: When you bring back setter, please, fix this doc, too.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -35,5 +35,5 @@ See xref:amqp/outbound-channel-adapter.adoc[] for more information.
 === HTTP Support Changes
 
 The `RestTemplate`-based configuration for the HTTP outbound gateway and outbound channel adapter (the `rest-template` attribute, and the respective Java DSL and constructor options) is now deprecated in favor of `RestClient`-based configuration.
-The `HttpRequestExecutingMessageHandler` also gains `setDefaultStatusHandler(RestClient.ResponseSpec.ErrorHandler)` and `setDefaultStatusHandler(Predicate<HttpStatusCode>, RestClient.ResponseSpec.ErrorHandler)` methods (with matching Java DSL `errorHandler()` overloads) to configure `RestClient`-based error handling.
+The `HttpRequestExecutingMessageHandler` also gains `defaultStatusHandler(RestClient.ResponseSpec.ErrorHandler)` and `defaultStatusHandler(Predicate<HttpStatusCode>, RestClient.ResponseSpec.ErrorHandler)` methods (with matching Java DSL `defaultErrorHandler()` overloads) to configure `RestClient`-based error handling.
```
artembilan: DITTO

src/reference/antora/modules/ROOT/pages/whats-new.adoc
artembilan: Was not addressed.

artembilan (changes requested):

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -717,6 +718,48 @@ public void failWhenSetRequestFactoryWithExternalRestClient() {
 				.withMessageContaining("externally configured RestClient");
 	}
 
+	@Test
+	public void defaultStatusHandlerAppliesOnlyToMatchingPredicate() throws IOException {
+		ClientHttpRequestFactory requestFactory = mock(ClientHttpRequestFactory.class);
```
artembilan: Hm.
How this was missed to become just `mock()` as well?

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -717,6 +718,48 @@ public void failWhenSetRequestFactoryWithExternalRestClient() {
...
+		handler.setExpectedResponseType(String.class);
+		handler.defaultStatusHandler(HttpStatusCode::is4xxClientError,
+				(request, response) -> handledStatus.set(response.getStatusCode()));
+		setBeanFactory(handler);
+		handler.afterPropertiesSet();
+		handler.setOutputChannel(new QueueChannel());
+
+		ClientHttpResponse response = mock();
+		when(response.getStatusCode()).thenReturn(HttpStatus.NOT_FOUND);
+		when(response.getHeaders()).thenReturn(new HttpHeaders());
```
artembilan: I don't think we need `mock()`.
Use `new MockClientHttpResponse()` instead.
I believe same could be done for the `ClientHttpRequest` - `new MockClientHttpRequest()`.
We also don't need `ClientHttpRequestFactory` as mock.
It could simply be a lambda impl with request returning.

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -717,6 +718,48 @@ public void failWhenSetRequestFactoryWithExternalRestClient() {
...
+
+		assertThat(handledStatus.get())
+				.isEqualTo(HttpStatus.NOT_FOUND);
+
+		handledStatus.set(null);
+
+		when(response.getStatusCode()).thenReturn(HttpStatus.INTERNAL_SERVER_ERROR);
+		when(clientRequest.execute()).thenReturn(response);
+
+		assertThatException()
```
artembilan: Isn't there a specific exception thrown from the `handleMessage` ?

spring-integration-http/src/test/java/org/springframework/integration/http/dsl/HttpDslTests.java
```diff
@@ -422,13 +427,27 @@ public Validator customValidator() {
...
```
artembilan: Why do we call `response.getStatusCode()` so many times in this method?
I also wonder if we even need such a complex logic here at all.
Looks like what we do in the test is just this:
```
				.andExpect(status().isForbidden())
				.andExpect(content().string("Error"));
```
So, apparently just `HttpClientErrorException.create()` is enough for this test suite logic.
I mean less code in the test classes, less headache to determine what is going on and why 🤷

artembilan (changes requested):

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -749,10 +748,10 @@ public void defaultStatusHandlerAppliesOnlyToMatchingPredicate() throws IOExcept
 
 		handledStatus.set(null);
 
-		when(response.getStatusCode()).thenReturn(HttpStatus.INTERNAL_SERVER_ERROR);
+		response = new MockClientHttpResponse(new byte[0], HttpStatus.INTERNAL_SERVER_ERROR);
 		when(clientRequest.execute()).thenReturn(response);
```
artembilan: I still think the `MockClientHttpRequest` can be used instead of `mock()`.
See its `setResponse(ClientHttpResponse clientHttpResponse)` API.

artembilan (changes requested):
Very close.
Thank you!

PR #11247, opened 2026-08-08
GH-10847: Fix `putIfAbsent()` loop in `JdbcMetadataStore`

### LINKED ISSUE #10847: JdbcMetadataStore putIfAbsent may stuck in while loop on MySQL database.

lsh1215:
Fixes: gh-10847

Follow-up to my [analysis comment](https://github.com/spring-projects/spring-integration/issues/10847#issuecomment-5218258467) on the issue.

## The defect

`putIfAbsent()` mixes a locking statement with a non-locking one inside a single transaction: `INSERT ... SELECT ... HAVING COUNT(*) = 0` to insert, then a plain `SELECT` (`getValueQuery`) to re-read. The MySQL 8.0 manual advises against this under [`REPEATABLE READ`](https://dev.mysql.com/doc/refman/8.0/en/innodb-transaction-isolation-levels.html#isolevel_repeatable-read): a non-locking `SELECT` is served from the transaction's read view, while locking statements use the most recent state of the database.

Whenever the read view was created before the commit that added the row, the insert sees the row and reports 0 affected rows, the re-read cannot see it, and the loop has no branch left that can terminate. Instrumented, every logged pass from 2 to 22,000 of one stuck call reports `INSERT affected=0` with no exception next to `SELECT -> EMPTY`.

`put()` has the same insert-then-re-read shape and is unaffected, because it re-reads with `getValueForUpdateQuery`.

## The change

Re-read with the locking query when, and only when, the non-locking one comes back empty.

Keeping the non-locking query as the first attempt is deliberate. Switching the re-read to `FOR UPDATE` unconditionally was tried in #10980 and breaks `verifyJdbcMetadataStoreConcurrency`: 99 of the 100 threads were never blind to begin with, and making all of them take an exclusive lock serialises them on one key. Applied only on the empty path the lock fires on roughly 5% of re-reads in that same test.

No new SQL is introduced. `getValueForUpdateQuery` is the statement `put()` and `remove()` already issue, so every vendor already executes it in production.

`@Transactional` stays, no isolation level is forced from inside the store, and `tryToPutIfAbsent()` is untouched.

## Tests

- `MySqlMetadataStoreTests.verifyPutIfAbsentSeesConcurrentlyCommittedValue` - the reproducing case. Fails with a `TimeoutException` before the change. It pins `ISOLATION_REPEATABLE_READ` on the `TransactionTemplate` rather than trusting the server default, otherwise it would also pass against unfixed code if that default ever changed.

`./gradlew :spring-integration-jdbc:check` passes; the existing 100-thread test was also run over 50 rounds with no failures.

## Scope and known limitations

Please weigh these; I kept the change narrow rather than deciding them on my own.

**Swallowed deadlocks in `tryToPutIfAbsent()`.** `CannotAcquireLockException` is a `TransientDataAccessException`, so a deadlock is caught there and returned as `0`, which the loop reads as "the row already exists". A deadlock means the whole transaction was rolled back, so the code then keeps working inside a dead transaction and the transaction manager commits it none the wiser. This PR ends the loop; that silent partial rollback stays. It touches the broader transactional unit of work directly, so I left it alone - happy to take it separately.

**The locking re-read can throw.** The old empty path took no locks and could not block. `SELECT ... FOR UPDATE` can, so a lock-wait timeout or deadlock on that read now propagates out of `putIfAbsent()`. I chose propagation over catch-and-retry because `put()` and `remove()` already issue the same locking read without catching, and retrying inside a transaction MySQL has already rolled back is what causes the problem above. Worth confirming this is the behaviour you want.

**Configurations still unbounded.** With `setLockHint("")` the two queries are the same, so the guard skips the second one and MySQL `REPEATABLE READ` loops exactly as before - `PersistentAcceptOnceFileListFilterExternalStoreTests` uses that setting. PostgreSQL under an explicit `REPEATABLE READ` also still loops, since its locking reads stay inside the snapshot. There is no loop bound anywhere; whether a bounded retry that fails with a diagnosable exception would be better than spinning is your call.

**New gap lock.** Under MySQL `REPEATABLE READ` a `SELECT ... FOR UPDATE` matching nothing takes a gap lock held until the caller's transaction commits. Previously this path took no locks. `remove()` already gap-locks a missing key, so it is not a new class of behaviour, but it is new for `putIfAbsent()`.

### DIFF
--- a/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
+++ b/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStore.java
@@ -34,6 +34,7 @@ import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Implementation of {@link ConcurrentMetadataStore} using a relational database via JDBC.
@@ -52,6 +53,7 @@ import org.springframework.util.Assert;
  * @author Bojan Vukasovic
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Sanghun Lee
  *
  * @since 5.0
  */
@@ -117,6 +119,8 @@ public class JdbcMetadataStore implements ConcurrentMetadataStore, InitializingB
 
 	private boolean checkDatabaseOnStart = true;
 
+	private boolean lockHintProvided;
+
 	/**
 	 * Instantiate a {@link JdbcMetadataStore} using provided dataSource {@link DataSource}.
 	 * @param dataSource a {@link DataSource}
@@ -176,6 +180,7 @@ public class JdbcMetadataStore implements ConcurrentMetadataStore, InitializingB
 				this.jdbcTemplate.execute((ConnectionCallback<String>) connection ->
 						connection.getMetaData().getDatabaseProductName());
 		this.getValueQuery = String.format(this.getValueQuery, this.tablePrefix);
+		this.lockHintProvided = StringUtils.hasText(this.lockHint);
 		this.getValueForUpdateQuery = String.format(this.getValueForUpdateQuery, this.tablePrefix, this.lockHint);
 		this.replaceValueQuery = String.format(this.replaceValueQuery, this.tablePrefix);
 		this.replaceValueByKeyQuery = String.format(this.replaceValueByKeyQuery, this.tablePrefix);
@@ -240,7 +245,17 @@ public class JdbcMetadataStore implements ConcurrentMetadataStore, InitializingB
 					return this.jdbcTemplate.queryForObject(this.getValueQuery, String.class, key, this.region);
 				}
 				catch (EmptyResultDataAccessException e) {
-					//somebody deleted it between calls. try to insert again (go to beginning of while loop)
+					//the insert saw the row, so this snapshot is older than its commit. read it under
+					//a lock, which is not served from the snapshot. no point without a lock hint
+					if (this.lockHintProvided) {
+						try {
+							return this.jdbcTemplate.queryForObject(this.getValueForUpdateQuery, String.class, key,
+									this.region);
+						}
+						catch (EmptyResultDataAccessException ex) {
+							//somebody deleted it between calls. try to insert again (go to beginning of while loop)
+						}
+					}
 				}
 			}
 		}
--- a/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlMetadataStoreTests.java
+++ b/spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/mysql/MySqlMetadataStoreTests.java
@@ -19,6 +19,7 @@ package org.springframework.integration.jdbc.mysql;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
 import javax.sql.DataSource;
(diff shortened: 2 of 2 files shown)


spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
```diff
@@ -88,6 +104,65 @@ public void keyAndValuesAreNotOverwrittenOnPutIfAbsent() {
 		assertThat(bar).isEqualTo("bar");
 	}
 
+	@Test
+	@Timeout(10)
```
artembilan: We don’t need mock test since in the place where we are testing there is always Docker.
Please, remove it.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
```diff
@@ -88,6 +104,65 @@ public void keyAndValuesAreNotOverwrittenOnPutIfAbsent() {
...
+
+		assertThat(store.putIfAbsent("someKey", "someValue")).isNull();
+
+		// With an empty lock hint both queries are semantically the same, so the empty re-read is
+		// not retried with a lock; the loop goes straight back to the insert.
+		verify(operations, times(1)).queryForObject(anyString(), eq(String.class), any(), any());
+	}
+
+	@Test
+	public void putIfAbsentReturnsNullForAStoredNullValue() {
```
artembilan: I think this is a separate story.
Let’s revise it in its own  issue/PR!

artembilan (changes requested):

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
lsh1215: Removed. The MySQL test covers it.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
lsh1215: Agreed, removed from this PR. I will open a separate issue for it.

lsh1215: Opened #11248 for it.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
lsh1215: One correction to my earlier reply: the MySQL test covers the locking re-read, but the empty `lockHint` branch of the guard is now untested. It is hard to cover with a container, since the behaviour there is that the loop still does not terminate. Happy to drop the guard if you would rather not carry an untested branch.

artembilan: Right. But isn’t that case useless?
Apparently to make it working (stop looping) such a hint must be provided.
Isn’t that what we do by default?
Why do we need to test no-hint use-case if it is deliberately going to fail?
So, my idea is to not test it at all

lsh1215: Agreed, nothing to test there.

PR #11254, opened 2026-08-10
GH-11253: Fix directory-creation race in `FileWritingMessageHandler`

### LINKED ISSUE #11253: `FileWritingMessageHandler`: concurrent auto-creation of the same destination directory fails with "Destination directory [...] could not be created."

Observed in production with 6.3.1. The affected code is unchanged on `main` (7.1.x); reproduced against 7.1.0 (see sample below).

**Describe the bug**

When `FileWritingMessageHandler` is configured with a `directory-expression` (e.g. via `<int-file:outbound-channel-adapter directory-expression="..."/>`) and `auto-create-directory` is enabled (the default), the destination directory is validated/created per message on the calling thread:

```java
private static void validateDestinationDirectory(File destinationDirectory, boolean autoCreateDirectory) {
	if (!destinationDirectory.exists() && autoCreateDirectory) {
		Assert.isTrue(destinationDirectory.mkdirs(),
				() -> "Destination directory [" + destinationDirectory + "] could not be created.");
	}
	...
```

This check-then-act sequence is not atomic. When two or more threads concurrently deliver messages that resolve to the same, not-yet-existing destination directory, all of them may pass the `!destinationDirectory.exists()` check, but only one `File.mkdirs()` call wins. `File.mkdirs()` returns `false` when the directory already exists by the time it runs, so the losing threads fail with:

```
java.lang.IllegalArgumentException: Destination directory [/path/to/output/subdir] could not be created.
```

even though the directory exists at that point and the write could simply proceed.

Any multi-threaded flow (e.g. a poller with a task executor) that writes to dynamically resolved directories is affected. In our case, a file-processing flow that mirrors input subfolder structures into an output folder fails sporadically whenever several files from the same new subfolder are processed in parallel.

This looks like the actual root cause behind #2239 (same symptom, closed in 2017 as environment-related).

**To Reproduce**

Run the sample below (plain `main`, only `spring-integration-file` required). Eight threads concurrently send one message each to the same freshly resolved destination directory, up to 500 rounds with a new directory per round. Against 7.1.0 it consistently fails for us in the very first rounds:

```
FAILED in round 0: java.lang.IllegalArgumentException: Destination directory [/tmp/si-race-.../sub-0] could not be created.
```

**Expected behavior**

Auto-creation of the destination directory should be idempotent and safe under concurrency. `java.nio.file.Files.createDirectories(Path)` provides exactly these semantics (it does not fail when the directory already exists, including when it is created concurrently) and would also report the actual cause via `IOException` when creation genuinely fails, e.g.:

```java
if (!destinationDirectory.exists() && autoCreateDirectory) {
	try {
		Files.createDirectories(destinationDirectory.toPath());
	}
	catch (IOException ex) {
		throw new IllegalArgumentException(
				"Destination directory [" + destinationDirectory + "] could not be created.", ex);
	}
}
```

With this change applied locally, the reproducer below passes (3 runs x 500 rounds x 8 threads, no failure).

**Sample**

```java
import java.nio.file.Files;
import java.nio.file.Path;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.CyclicBarrier;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.Future;

import org.springframework.context.annotation.AnnotationConfigApplicationContext;
import org.springframework.expression.spel.standard.SpelExpressionParser;
import org.springframework.integration.file.FileHeaders;
import org.springframework.integration.file.outbound.FileWritingMessageHandler;
import org.springframework.integration.support.MessageBuilder;

public class Repro {

	@EnableIntegration
	static class Config {
	}

	public static void main(String[] args) throws Exception {
		Path base = Files.createTempDirectory("si-race-");

		AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext(Config.class);
		FileWritingMessageHandler handler = new FileWritingMessageHandler(
				new SpelExpressionParser().parseExpression("headers['dir']"));
		handler.setBeanFactory(ctx.getBeanFactory());
		handler.setExpectReply(false);
		handler.afterPropertiesSet();

		int threads = 8;
		ExecutorService executor = Executors.newFixedThreadPool(threads);
		try {
			for (int round = 0; round < 500; round++) {
				String dir = base.resolve("sub-" + round).toString();
				CyclicBarrier barrier = new CyclicBarrier(threads);
				List<Future<?>> futures = new ArrayList<>();
				for (int t = 0; t < threads; t++) {
					String fileName = "file-" + t + ".txt";
					futures.add(executor.submit(() -> {
						barrier.await();
						handler.handleMessage(MessageBuilder.withPayload("payload")
								.setHeader("dir", dir)
								.setHeader(FileHeaders.FILENAME, fileName)
								.build());
						return null;
					}));
				}
				for (Future<?> f : futures) {
					try {
						f.get();
					}
					catch (Exception e) {
						Throwable rootCause = e;
						while (rootCause.getCause() != null) {
							rootCause = rootCause.getCause();
						}
						System.out.println("FAILED in round " + round + ": " + rootCause);
						return;
					}
				}
			}
			System.out.println("no failure in 500 rounds");
		}
		finally {
			executor.shutdownNow();
		}
	}

}
```

(Note for `FileWritingMessageHandler` in 6.x: the class resides in `org.springframework.integration.file` instead of `org.springframework.integration.file.outbound`; the affected code is identical.)

tknall:
Fixes GH-11253

`FileWritingMessageHandler.validateDestinationDirectory()` used a non-atomic
`exists()`/`mkdirs()` sequence. With a `directory-expression`, this code runs per
message on the calling thread, so concurrent messages resolving to the same,
not-yet-existing destination directory raced on `File.mkdirs()`: the losing threads
received `false` (the directory had been created concurrently in the meantime) and
failed with `IllegalArgumentException: Destination directory [...] could not be created.`
although the directory existed at that point.

Use the idempotent and concurrency-safe `Files.createDirectories()` instead. The
exception type and message are retained for backward compatibility; a genuine creation
failure now additionally carries the underlying `IOException` as cause.

Verified with the reproducer from GH-11253: fails within the first rounds before the
change, passes 3 x 500 rounds x 8 threads after it.


spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
```diff
@@ -710,6 +719,44 @@ public void checkPathThrowsOnDirectoryTraversal() {
...
+
+		int concurrency = 8;
+		ExecutorService executorService = Executors.newFixedThreadPool(concurrency);
+		try {
+			for (int round = 0; round < 50; round++) {
+				File directory = new File(this.tempDir, "destination-" + round);
+				CyclicBarrier barrier = new CyclicBarrier(concurrency);
+				List<Future<?>> results = new ArrayList<>();
+				for (int i = 0; i < concurrency; i++) {
+					Message<String> message = MessageBuilder.withPayload("foo")
```
artembilan: No `foo/bar` language, please.
We slowly migrate existing code, but definitely don't want more of them in a new one.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
```diff
@@ -710,6 +719,44 @@ public void checkPathThrowsOnDirectoryTraversal() {
...
+							.setHeader("directory", directory.getAbsolutePath())
+							.setHeader(FileHeaders.FILENAME, "file-" + i + ".txt")
+							.build();
+					results.add(executorService.submit(() -> {
+						barrier.await();
```
artembilan: I'm not sure that we need a `barrier`...
Please, elaborate.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
```diff
@@ -710,6 +719,44 @@ public void checkPathThrowsOnDirectoryTraversal() {
...
+						handler.handleMessage(message);
+						return null;
+					}));
+				}
+				for (Future<?> result : results) {
+					result.get(10, TimeUnit.SECONDS);
```
artembilan: I think we don't need such a complexity.
We just need one `CountDownLatch` to countDown in the `Runnable` after that `handler.handleMessage()`.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
```diff
@@ -710,6 +719,44 @@ public void checkPathThrowsOnDirectoryTraversal() {
...
+				}
+				assertThat(new File(directory, "file-0.txt")).exists();
```
artembilan: This doesn't look finished.
I would expect here a loop to check that all requested files are created.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
```diff
@@ -710,6 +719,44 @@ public void checkPathThrowsOnDirectoryTraversal() {
 			.withRootCauseInstanceOf(InvalidPathException.class);
 	}
 
+	@Test
+	public void concurrentWritesToSameNewDestinationDirectory() throws Exception {
+		FileWritingMessageHandler handler = new FileWritingMessageHandler(
+				new SpelExpressionParser().parseExpression("headers['directory']"));
```
artembilan: We don't need SpEL parsing for such an expression style.
The `FunctionExpression` will do the trick for us:
```
new FunctionExpression<Message<?>>(message -> message.getHeaders().get("directory"))
```

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
```diff
@@ -710,6 +719,44 @@ public void checkPathThrowsOnDirectoryTraversal() {
...
+		handler.setBeanFactory(TEST_INTEGRATION_CONTEXT);
+		handler.setApplicationContext(new GenericApplicationContext());
```
artembilan: The `TEST_INTEGRATION_CONTEXT` is already an `ApplicationContext`, so we don't need this new instance.
More over, please, double check if we need it at all.
I just removed same line of code in the `@BeforeEach` method of this class - and all tests have passed.

artembilan (changes requested):
Please, find my review in lines.

Thank you!

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
tknall: Sorry, I was not aware of that.

Renamed the payload to `"test data"`.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
tknall: The barrier was there to line up all threads right **before** `handleMessage()` to maximize the chance of hitting the `exists()`/`mkdirs()` window.

However, I re-verified without it -> the test still failed consistently against the unpatched code in 10 out of 10 runs. So you are right, it is not needed; removed.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
tknall: Simplified as suggested: plain `executorService.execute()` with a single `CountDownLatch` counted down after `handleMessage()`, no futures involved.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
tknall: Added the assertion loop verifying that all requested files exist.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
tknall: Your are right. I wasn't aware of that. Thanks for the hint.
Done. Replaced by FunctionExpression.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
tknall: You are right - removed the `setApplicationContext(new GenericApplicationContext())` line and double-checked that the test still behaves as intended...

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
artembilan: Thank you!
You don't need to comment for every single my review line that you are agreed with me.
There is enough just one common: I'll see that we are on the same page with your next change set.
However you are free to discuss further if you think I'm wrong in my review.
You also don't need to squash all the commits after PR is opened.
I saw that question from my AI, so I believe yours also is trying to follow our CONTRIBUTION.md.
I guess that one has to be fixed respectively 😄

artembilan (changes requested):
```
[ant:checkstyle] [ERROR] /home/runner/work/spring-integration/spring-integration/spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java:54:1: Wrong lexicographical order for 'org.springframework.integration.channel.QueueChannel' import. Should be before 'org.springframework.integration.expression.FunctionExpression'. [ImportOrder]
```

Please, make sure you run `./gradlew check` (or just for that specific module) before pushing changes upstream to the PR.

Thank you!

PR #11259, opened 2026-08-11
GH-11248: Make METADATA_VALUE NOT NULL in JdbcMetadataStore schemas

### LINKED ISSUE #11248: `JdbcMetadataStore.putIfAbsent()` returns `null` for an existing row with a null `METADATA_VALUE`
Split out from #10847 / #11247, where @artembilan asked for this to be its own story.

`ConcurrentMetadataStore.putIfAbsent` documents its return value as:

> @return null if successful, the old value otherwise.

`METADATA_VALUE` is nullable in every shipped schema (`schema-mysql.sql`, `schema-derby.sql`, and the rest), so a row can exist with a null value. For such a row `JdbcMetadataStore.putIfAbsent()` returns `null`: the insert reports 0 affected rows because the row is there, and `queryForObject` returns `null` rather than throwing `EmptyResultDataAccessException`. Nothing was stored, yet the caller is told it was.

`MetadataStoreSelector` reads the return value literally:

```java
return this.metadataStore.putIfAbsent(key, value) == null;
```

and its javadoc says a `null` return "means that the value has been placed in the `MetadataStore`". So an idempotent receiver keyed on a null-valued row accepts every duplicate and never records anything.

Reproducing it needs no concurrency:

```java
new JdbcTemplate(dataSource)
        .update("INSERT INTO INT_METADATA_STORE(METADATA_KEY, METADATA_VALUE, REGION) VALUES (?, ?, ?)",
                "nullValued", null, "DEFAULT");

// returns null, as if the value had just been stored
metadataStore.putIfAbsent("nullValued", "someValue");
```

The obvious fix is not obvious, which is why I did not fold it into #11247. `putIfAbsent()` distinguishes "row absent" from "row present with a null value" only by whether `queryForObject` throws, and the retry loop relies on that distinction to terminate. Collapsing the two cases would leave the loop with no exit. Options I can see:

- Make the column non-nullable, which is a schema change for existing users.
- Reject null values on write, since `put()` and `putIfAbsent()` already `Assert.notNull` the value - a null can only get there out of band.
- Have the store treat a null-valued row as present and return something other than `null`, which the interface's return type does not allow without changing the contract.

Happy to work on whichever direction you prefer, or to leave it as documentation if you consider a null `METADATA_VALUE` unsupported in the first place.

kalayciburak:
Fixes: gh-11248

## Summary

`JdbcMetadataStore.putIfAbsent()` treats a successful read of a SQL `NULL` `METADATA_VALUE` as "insert succeeded" (`return null`), so a null-valued row can make callers (e.g. `MetadataStoreSelector`) believe a value was stored when it was not.

Per maintainer guidance on the issue, this change makes `METADATA_VALUE` **`NOT NULL`** in all shipped `schema-*.sql` scripts for 7.2.0 and documents the non-null contract on `JdbcMetadataStore`. The store API already rejects null values via `Assert.notNull`; the schema now matches that contract.

Existing databases created with older scripts should migrate the column to `NOT NULL` (and remove any null-valued rows) when upgrading.

## Test plan

- [x] `./gradlew :spring-integration-jdbc:test --tests org.springframework.integration.jdbc.metadata.JdbcMetadataStoreTests` (12/12)
- [x] `./gradlew :spring-integration-jdbc:checkstyleMain :spring-integration-jdbc:checkstyleTest`


lsh1215:
Thanks for picking this up — I opened gh-11248.

One case I want to make sure I understand: the `NOT NULL` applies only to tables created from the new scripts. On a table created by an older script the column stays nullable, so a row with a null value can still exist — the insert affects 0 rows, `queryForObject()` returns that `null` instead of throwing, and `putIfAbsent()` reports it as stored. That's the case I filed.

@artembilan, did "react respectively in the `JdbcMetadataStore`" include a guard for that, or are tables from older scripts out of scope here? Either way works for me — just wanted to ask while the PR is open.

Minor: `jdbc/metadata-store.adoc` still shows the DDL with a nullable `METADATA_VALUE`.

kalayciburak:
yeah — older schemas are out of scope for a runtime guard here. Artem's note was basically "don't let null into the DB": NOT NULL on the shipped scripts + the existing Assert.notNull on put/putIfAbsent. Upgrade path is drop null rows and alter the column (same as the class javadoc).

also fixed the DDL sample in `jdbc/metadata-store.adoc` — it still showed a nullable METADATA_VALUE.

artembilan:
> On a table created by an older script the column stays nullable

Right. We will call this out in the Migration Guide.

We probably need to double check the logic in that `putIfAbsent()` what is the reaction for the `null`.

Thanks

kalayciburak:
on putIfAbsent: value is already Assert.notNull, so a null arg fails fast with IAE (covered by the new test). return null only means "we inserted".

if a legacy row still has a null METADATA_VALUE, the insert path sees the key, queryForObject returns null, and that can look like a successful insert — that's the bug this NOT NULL change is aimed at. new schemas can't store it; old ones need the migrate step you'll put in the Migration Guide.

also squashed + fixed DCO sign-off name mismatch on the adoc commit.

artembilan:
Thanks

> also squashed

I guess that is what your AI is doing according to our CONTRIBUTING.md.
Not a first time.
The idea behind that instruction was about the first commit to open PR.
All other should be as they come for a proper change history to easier for us to review.
We squash in the end on merge.

I'll go ahead and rephrase the rule in the CONTRIBUTING.md: not a first time I hear squash confirmation on subsequent pushes 😄

kalayciburak:
got it — won't squash follow-ups. the squash was only to clean a DCO author/sign-off mismatch on the docs commit, not to hide history. next time I'll leave commits as they come and fix sign-off per commit. thanks for clarifying

src/reference/antora/modules/ROOT/pages/jdbc/metadata-store.adoc
```diff
@@ -29,12 +29,15 @@ For example, the following listing shows the H2 DDL for the metadata table:
...
 	METADATA_KEY VARCHAR(255) NOT NULL,
-	METADATA_VALUE VARCHAR(4000),
+	METADATA_VALUE VARCHAR(4000) NOT NULL,
 	REGION VARCHAR(100) NOT NULL,
 	constraint INT_METADATA_STORE_PK primary key (METADATA_KEY, REGION)
 );
 ----
 
+`METADATA_VALUE` is non-null (since 7.2).
+Existing databases created with older scripts should remove any null-valued rows and alter the column to `NOT NULL` when upgrading.
```
artembilan: Let's move this sentence to the `whats-new.adoc`!
Create a new section there like:
```
[[x7.2-jdbc-changes]]
=== JDBC Support Changes
```
Mention the change you've just did and add a link to the target chapter.

spring-integration-jdbc/src/test/java/org/springframework/integration/jdbc/metadata/JdbcMetadataStoreTests.java
```diff
@@ -124,4 +126,27 @@ void noTableThrowsExceptionOnStart() {
...
+	void putIfAbsentRejectsNullValue() {
+		assertThatExceptionOfType(IllegalArgumentException.class)
+				.isThrownBy(() -> this.metadataStore.putIfAbsent("foo", null))
+				.withMessageContaining("value");
+	}
+
+	@Test
+	void putRejectsNullValue() {
+		assertThatExceptionOfType(IllegalArgumentException.class)
+				.isThrownBy(() -> this.metadataStore.put("foo", null))
```
artembilan: No `foo/bar` language, please.
Might be the case that we have a lot in this class already, but that is a different story to be addressed eventually.
Let's just don't introduce more!

artembilan (changes requested):
Add your name to the `@author` list of the `JdbcMetadataStoreTests`.

Thank you!

kalayciburak:
moved the note to whats-new under JDBC Support Changes, dropped foo keys in the new tests, and added myself to the @author list on JdbcMetadataStoreTests.

artembilan:
> moved the note to whats-new...

Thanks. 
You don't need to post comments about what you've done if you agree with my review.
I'll see that we are on the same page in your new commit.

Use code snippets in your comments for annotation and classes/methods.
Right no you have called that Author Software GitHub organization into our discussion 😄 

See GH docs about commenting style: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax

PR #11260, opened 2026-08-11
GH-11229: Upgrade GraalvmVersion to 25.2.4

### LINKED ISSUE #11229: Upgrade `org.graalvm.sdk:graal-sdk` to `25.2.4`

cppwfs:
- Mark referencedScript as @Disabled in Jsr223ServiceActivatorTests.
- Reference the upstream GraalPython issue tracking the underlying
variable-scoping bug.
- This issue was identified graalvmVersion was updated from 25.0.4 to 25.2.4
- Will create a new issue to check future release of GraalVM to verify that their
fix works and remove the `@Disabled` flag from tests.

Resolves: https://github.com/spring-projects/spring-integration/issues/11229


artembilan (changes requested):
You said `Fixes` an update issue, but commit (or PR title) is misleading.
And then you mark this PR as bug and with respective milestone.
Why?
We had that discussion before: no milestone and no labels on PR if it fixes existing issue.

Please, change the commit message to upgrade matter which is the trigger for this change.
The test (pay attention: single) is just a consequence of upgrade.

Thank you for the follow up issue to re-enable test back!

spring-integration-scripting/src/test/java/org/springframework/integration/scripting/config/jsr223/Jsr223ServiceActivatorTests.java
```diff
@@ -61,6 +63,7 @@ public class Jsr223ServiceActivatorTests {
 	@Autowired
 	private MessageChannel withScriptVariableGenerator;
 
+	@Disabled("Remove Disabled when https://github.com/oracle/graalpython/issues/1065 is added to a release ")
```
artembilan:
```suggestion
	@Disabled("Until https://github.com/oracle/graalpython/issues/1065")
```

artembilan (changes requested):

artembilan (approved):

PR #11270, opened 2026-08-14
GH-11268: Suppress RestTemplate removal warnings

### LINKED ISSUE #11268: Suppress RestTemplate Deprecation Warnings
- All deprecated `RestTemplate` utilizations should have a `SuppressWarning` until they are removed in a future release.   
- Replace deprecated `RestTemplate` tests with `RestClient`, if the `RestClient` comparable test is not present.   `RestTemplate` tests that have a comparable RestClient version should be removed.

kalayciburak:
Add `@SuppressWarnings("removal")` on remaining RestTemplate-based HTTP outbound APIs until they are removed.

- Keep the production RestTemplate wrappers as-is, just quiet the removal warnings
- Drop test-side suppressions and move those cases onto RestClient
- Remove RestTemplate parser coverage that already has a RestClient twin

**Verification**
- `./gradlew :spring-integration-http:check`

Fixes: gh-11268

### DIFF
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpOutboundGatewayParser.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpOutboundGatewayParser.java
@@ -113,10 +113,7 @@ public class HttpOutboundGatewayParser extends AbstractConsumerEndpointParser {
 					.addIndexedArgumentValue(1, new RuntimeBeanReference(restClientRef));
 		}
 		else {
-			for (String referenceAttributeName : HttpAdapterParsingUtils.LOCAL_CLIENT_REFERENCE_ATTRIBUTES) {
-				IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element, referenceAttributeName);
-			}
-			IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "encoding-mode");
+			HttpAdapterParsingUtils.configureLocalClientAttributes(builder, element);
 		}
 		return builder;
 	}
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpOutboundGatewayParser.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpOutboundGatewayParser.java
@@ -35,6 +35,7 @@ import org.springframework.util.StringUtils;
  * @author Artem Bilan
  * @author Shiliang Li
  * @author Arun Sethumadhavan
+ * @author Burak Kalayci
  */
 public class HttpOutboundGatewayParser extends AbstractConsumerEndpointParser {
 
--- a/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/Http.java
+++ b/spring-integration-http/src/main/java/org/springframework/integration/http/dsl/Http.java
@@ -37,6 +37,7 @@ import org.springframework.web.client.RestTemplate;
  * @author Artem Bilan
  * @author Shiliang Li
  * @author Arun Sethumadhavan
+ * @author Burak Kalayci
  *
  * @since 5.0
  */
@@ -90,6 +91,7 @@ public final class Http {
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static HttpMessageHandlerSpec outboundChannelAdapter(URI uri, @Nullable RestTemplate restTemplate) {
 		return outboundChannelAdapter(uri, toRestClient(restTemplate));
 	}
@@ -115,6 +117,7 @@ public final class Http {
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static HttpMessageHandlerSpec outboundChannelAdapter(String uri, @Nullable RestTemplate restTemplate) {
 		return outboundChannelAdapter(uri, toRestClient(restTemplate));
 	}
@@ -142,6 +145,7 @@ public final class Http {
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static <P> HttpMessageHandlerSpec outboundChannelAdapter(Function<Message<P>, ?> uriFunction,
 			RestTemplate restTemplate) {
 
@@ -174,6 +178,7 @@ public final class Http {
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static HttpMessageHandlerSpec outboundChannelAdapter(Expression uriExpression,
 			@Nullable RestTemplate restTemplate) {
 
@@ -243,6 +248,7 @@ public final class Http {
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static HttpMessageHandlerSpec outboundGateway(URI uri, @Nullable RestTemplate restTemplate) {
 		return outboundGateway(uri, toRestClient(restTemplate));
 	}
@@ -268,6 +274,7 @@ public final class Http {
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static HttpMessageHandlerSpec outboundGateway(String uri, @Nullable RestTemplate restTemplate) {
 		return outboundGateway(uri, toRestClient(restTemplate));
 	}
@@ -295,6 +302,7 @@ public final class Http {
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static <P> HttpMessageHandlerSpec outboundGateway(Function<Message<P>, ?> uriFunction,
 			RestTemplate restTemplate) {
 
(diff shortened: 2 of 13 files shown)


spring-integration-http/src/test/java/org/springframework/integration/http/HttpProxyScenarioTests.java
```diff
@@ -59,11 +59,13 @@
  * @author Artem Bilan
  * @author Gary Russell
  * @author Arun Sethumadhavan
+ * @author Burak Kalayci
  *
  * @since 3.0
  */
 @SpringJUnitConfig
 @DirtiesContext
+@SuppressWarnings("removal")
```
cppwfs: Per the issue we don't want to suppress the warnings but rather replace RestTemplate with RestClient in the tests.

spring-integration-http/src/test/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandlerTests.java
```diff
@@ -966,6 +970,7 @@ private static void setBeanFactory(HttpRequestExecutingMessageHandler handler) {
 		handler.setBeanFactory(TEST_INTEGRATION_CONTEXT);
 	}
 
+	@SuppressWarnings("removal")
```
cppwfs: Ditto

spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundChannelAdapterParserTests.java
```diff
@@ -63,9 +63,11 @@
  * @author Shiliang Li
  * @author Glenn Renfro
  * @author Arun Sethumadhavan
+ * @author Burak Kalayci
  */
 @SpringJUnitConfig
 @DirtiesContext
+@SuppressWarnings("removal")
```
cppwfs: Per the issue we don't want to suppress the warnings but rather replace RestTemplate with RestClient in the tests.

spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundGatewayParserTests.java
```diff
@@ -60,9 +60,11 @@
  * @author Biju Kunjummen
  * @author Glenn Renfro
  * @author Arun Sethumadhavan
+ * @author Burak Kalayci
  */
 @SpringJUnitConfig
 @DirtiesContext
+@SuppressWarnings("removal")
```
cppwfs: Ditto

cppwfs (changes requested):
Thank you so much for your contributions!   Its off to a great start!

artembilan (changes requested):
Right.

The `RestTemplate` is not our class to test around.
We would like to move to a replacement API as soon as possible.
Therefore, our tests must be fixed right now to prove that new API is good and have coverage for new API instead of the one out of support already.

And teach your AI agent that `./gradlew :spring-integration-http:check` task is for everything what it iterates separately.

Thanks

spring-integration-http/src/test/java/org/springframework/integration/http/HttpProxyScenarioTests.java
kalayciburak: yep, dropped the test suppress and switched those to RestClient

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/Http.java
```diff
@@ -90,6 +91,7 @@ public static HttpMessageHandlerSpec outboundChannelAdapter(Expression uriExpres
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	public static HttpMessageHandlerSpec outboundChannelAdapter(URI uri, @Nullable RestTemplate restTemplate) {
```
artembilan: My IDE complains like this:
image
Therefore, I prefer to use FQCN for deprecated classes instead of imports.

spring-integration-http/src/test/java/org/springframework/integration/http/HttpProxyScenarioTests.java
```diff
@@ -215,4 +198,44 @@ public void testHttpMultipartProxyScenario() throws Exception {
...
 	}
 
+	@SuppressWarnings("unchecked")
+	private static void injectRestClient(HttpRequestExecutingMessageHandler handler,
+			Function<HttpEntity<?>, ResponseEntity<?>> exchange,
+			Consumer<String> uriAsserter) {
+
+		RestClient restClient = mock(RestClient.class);
+		RestClient.RequestBodyUriSpec spec = mock(RestClient.RequestBodyUriSpec.class);
+		RestClient.ResponseSpec responseSpec = mock(RestClient.ResponseSpec.class);
```
artembilan: Use `mock()` API without arguments.

spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundGatewayParserTests.java
```diff
@@ -60,6 +60,7 @@
  * @author Biju Kunjummen
  * @author Glenn Renfro
  * @author Arun Sethumadhavan
+ * @author Burak Kalayci
```
artembilan: This is only the change in this class.
Do you have anything else to fix here?
Otherwise, restore its original state.

artembilan (changes requested):

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/Http.java
kalayciburak: switched RestTemplate to FQCN

spring-integration-http/src/test/java/org/springframework/integration/http/HttpProxyScenarioTests.java
kalayciburak: yep, mock() without args

spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundGatewayParserTests.java
kalayciburak: nothing else in there, restored

artembilan: Thanks.
However, you don't need to comment on every single my review if you agreed with me.
You just fix and push: it will be clear from follow up commits if we are on the same page.
Otherwise it is a bit of spam in my mail box.
Feel free to discuss further, though, if you are not agree with our request! 
😄

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
```diff
@@ -56,6 +57,7 @@ public class HttpMessageHandlerSpec
 	 * @deprecated Since 7.1 in favor of {@link RestClient}-based configuration.
 	 */
 	@Deprecated(since = "7.1", forRemoval = true)
+	@SuppressWarnings("removal")
 	protected HttpMessageHandlerSpec(URI uri, @Nullable RestTemplate restTemplate) {
```
artembilan: Well, doesn't look like you fully understood my request for FQCN...

spring-integration-http/src/main/java/org/springframework/integration/http/dsl/HttpMessageHandlerSpec.java
```diff
@@ -109,6 +113,7 @@ public HttpMessageHandlerSpec requestFactory(ClientHttpRequestFactory requestFac
 	 * @deprecated Since 7.2 in favor of {@link RestClient.ResponseSpec.ErrorHandler}.
 	 */
 	@Deprecated(since = "7.2", forRemoval = true)
+	@SuppressWarnings("removal")
 	public HttpMessageHandlerSpec errorHandler(ResponseErrorHandler errorHandler) {
```
artembilan: DITTO

artembilan (changes requested):

spring-integration-http/src/main/java/org/springframework/integration/http/outbound/HttpRequestExecutingMessageHandler.java
```diff
@@ -67,11 +67,13 @@
...
  * @author Arun Sethumadhavan
  * @author Glenn Renfro
+ * @author Burak Kalayci
  *
  * @since 2.0
  */
 public class HttpRequestExecutingMessageHandler extends AbstractHttpRequestExecutingMessageHandler {
 
+	@SuppressWarnings("removal")
 	private final @Nullable RestTemplate restTemplate;
```
artembilan: Why didn't you fix this class?

artembilan (changes requested):

artembilan (changes requested):
Two deprecation warnings are still missed:
```
/home/runner/work/spring-integration/spring-integration/spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundChannelAdapterParserTests.java:270: warning: [removal] ResponseErrorHandler in org.springframework.web.client has been deprecated and marked for removal
> Task :spring-integration-http:compileTestJava
	public static class StubErrorHandler implements ResponseErrorHandler {
	                                                ^
/home/runner/work/spring-integration/spring-integration/spring-integration-http/src/test/java/org/springframework/integration/http/config/HttpOutboundGatewayParserTests.java:68: warning: [removal] ResponseErrorHandler in org.springframework.web.client has been deprecated and marked for removal
	public static final ResponseErrorHandler mockResponseErrorHandler = mock();
	                    ^
2 warnings
```

Please, make sure you run `./gradlew :spring-integration-http:check` locally before pushing to PR.


For those `ResponseErrorHandler` I don't think that we need to suppress, but rather fix those tests to use a new API instead.
The point is that this deprecated one is going to be removed, so better to have coverage on whatever we rely now on.

Thanks

spring-integration-http/src/main/java/org/springframework/integration/http/config/HttpOutboundGatewayParser.java
```diff
@@ -113,10 +113,7 @@ else if (StringUtils.hasText(restClientRef)) {
 					.addIndexedArgumentValue(1, new RuntimeBeanReference(restClientRef));
 		}
 		else {
-			for (String referenceAttributeName : HttpAdapterParsingUtils.LOCAL_CLIENT_REFERENCE_ATTRIBUTES) {
-				IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element, referenceAttributeName);
-			}
-			IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "encoding-mode");
+			HttpAdapterParsingUtils.configureLocalClientAttributes(builder, element);
```
artembilan, line 117: Please, add your name to all of these new classes you have improved.

artembilan (changes requested):
Just one nit-pick.
Thank you!

artembilan:
@kalayciburak ,

PR #11276, opened 2026-08-16
GH-11275: Replace unnecessary fully-qualified type references with im…

### LINKED ISSUE #11275: FullyQualifiedType usage found by new checkstyle check
Checkstyle `import` package introduced a new check `UnnecessaryFullyQualifiedTypeCheck` currently in review https://github.com/checkstyle/checkstyle/pull/21004, while running this check on `spring-integration` repo it flagged 150 violations in `spring-integration`

Diff report:https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/4bb42ce_2026073849/reports/diff/spring-integration/index.html 

**how the check work**:
The check flags a fully qualified type reference like `java.util.Map` as unnecessary whenever it is the only type using that simple name in the file meaning it could safely be imported and written as just Map. It stays quiet whenever a second, different type with the same simple name also shows up `via import, declaration, another qualified reference, or an on-demand import`, since qualification is genuinely needed there to tell the two apart.

Since the check isn't merged yet, I'm not asking for any code changes  just wondering if there's a specific reason behind using fully-qualified names here instead of importing?.any feedback on the approach how the check itself could be improved, would be great too.

check doc link:https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/4147d0d_20260808220533/checks/imports/unnecessaryfullyqualifiedtype.html#

ZaheerAhmadDev:
Fixes   gh-11275

Replaced fullyqualifiedtype referenced with `import`
It make code more readeable and consistant.
these violation found in:https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/4bb42ce_2026073849/reports/diff/spring-integration/index.html 
Did fix some of the violation, most of the violation is `java.lang.string` we can import it, instead of using fullyqualified.

### DIFF
--- a/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpMessageSource.java
+++ b/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpMessageSource.java
@@ -28,6 +28,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.MessageProperties;
 import org.springframework.amqp.rabbit.batch.BatchingStrategy;
 import org.springframework.amqp.rabbit.batch.SimpleBatchingStrategy;
@@ -195,7 +196,7 @@ public class AmqpMessageSource extends AbstractMessageSource<Object> {
 					resp.getEnvelope(), StandardCharsets.UTF_8.name());
 			messageProperties.setConsumerQueue(this.queue);
 			Map<String, @Nullable Object> headers = this.headerMapper.toHeadersFromRequest(messageProperties);
-			var amqpMessage = new org.springframework.amqp.core.Message(resp.getBody(), messageProperties);
+			var amqpMessage = new Message(resp.getBody(), messageProperties);
 			Object payload;
 			if (this.batchingStrategy.canDebatch(messageProperties)) {
 				List<Object> payloads = new ArrayList<>();
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/config/AmqpOutboundChannelAdapterParserTests.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/config/AmqpOutboundChannelAdapterParserTests.java
@@ -43,6 +43,7 @@ import org.springframework.beans.factory.parsing.BeanDefinitionParsingException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.support.ClassPathXmlApplicationContext;
 import org.springframework.core.log.LogAccessor;
+import org.springframework.expression.Expression;
 import org.springframework.integration.amqp.outbound.AmqpOutboundEndpoint;
 import org.springframework.integration.amqp.support.AmqpHeaderMapper;
 import org.springframework.integration.channel.DirectChannel;
@@ -118,7 +119,7 @@ public class AmqpOutboundChannelAdapterParserTests {
 		AmqpOutboundEndpoint endpoint = TestUtils.getPropertyValue(eventDrivenConsumer, "handler");
 		assertThat(TestUtils.<Object>getPropertyValue(endpoint, "defaultDeliveryMode")).isNotNull();
 		assertThat(TestUtils.<Boolean>getPropertyValue(endpoint, "lazyConnect")).isFalse();
-		assertThat(TestUtils.<org.springframework.expression.Expression>getPropertyValue(endpoint, "delayExpression")
+		assertThat(TestUtils.<Expression>getPropertyValue(endpoint, "delayExpression")
 				.getExpressionString()).isEqualTo("42");
 		assertThat(TestUtils.<Boolean>getPropertyValue(endpoint, "headersMappedLast")).isFalse();
 
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/config/AmqpOutboundGatewayParserTests.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/config/AmqpOutboundGatewayParserTests.java
@@ -32,6 +32,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.parsing.BeanDefinitionParsingException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.support.ClassPathXmlApplicationContext;
+import org.springframework.expression.Expression;
 import org.springframework.integration.amqp.outbound.AmqpOutboundEndpoint;
 import org.springframework.integration.amqp.outbound.AsyncAmqpOutboundGateway;
 import org.springframework.integration.channel.QueueChannel;
@@ -97,7 +98,7 @@ public class AmqpOutboundGatewayParserTests {
 
 		assertThat(sendTimeout).isEqualTo(Long.valueOf(777));
 		assertThat(TestUtils.<Boolean>getPropertyValue(gateway, "lazyConnect")).isTrue();
-		assertThat(TestUtils.<org.springframework.expression.Expression>getPropertyValue(gateway, "delayExpression")
+		assertThat(TestUtils.<Expression>getPropertyValue(gateway, "delayExpression")
 				.getExpressionString()).isEqualTo("42");
 	}
 
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/dsl/AmqpClientTests.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/dsl/AmqpClientTests.java
@@ -16,6 +16,7 @@
 
 package org.springframework.integration.amqp.dsl;
 
+import com.rabbitmq.client.amqp.Environment;
 import com.rabbitmq.client.amqp.impl.AmqpEnvironmentBuilder;
 import org.junit.jupiter.api.Test;
 
@@ -94,7 +95,7 @@ class AmqpClientTests implements RabbitTestContainer {
 	static class ContextConfiguration {
 
 		@Bean
-		com.rabbitmq.client.amqp.Environment environment() {
+		Environment environment() {
 			return new AmqpEnvironmentBuilder()
 					.connectionSettings()
 					.port(RabbitTestContainer.amqpPort())
@@ -103,7 +104,7 @@ class AmqpClientTests implements RabbitTestContainer {
 		}
 
 		@Bean
-		AmqpConnectionFactory connectionFactory(com.rabbitmq.client.amqp.Environment environment) {
+		AmqpConnectionFactory connectionFactory(Environment environment) {
 			return new SingleAmqpConnectionFactory(environment);
 		}
 
--- a/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/inbound/AmqpClientInboundGatewayTests.java
+++ b/spring-integration-amqp/src/test/java/org/springframework/integration/amqp/inbound/AmqpClientInboundGatewayTests.java
@@ -27,6 +27,7 @@ import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.amqp.core.Declarables;
+import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.MessageBuilder;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.rabbitmq.client.AmqpConnectionFactory;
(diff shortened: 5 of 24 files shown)


artembilan (changes requested):
@ZaheerAhmadDev ,

thank you for an update!
May we understand that this PR is going to fix the mentioned issue in full potential?
Or do you anticipate some phases?

ZaheerAhmadDev:
Not likely, this PR is mostly to get real-world feedback and show the check has practical value it only fixes some part of the violations. Once the check is merged upstream, I can open a new PR to address the rest.
if that works for you.?

artembilan:
Right. Why do we need a separate PR when everything can be addressed here?
More over we cannot separate check and its fix: the project is going to fail until we fix all of them  .

ZaheerAhmadDev:
> Right. Why do we need a separate PR when everything can be addressed here? More over we cannot separate check and its fix: the project is going to fail until we fix all of them .

Make sense, lets fix all the violation in this PR.

ZaheerAhmadDev:
Done fixed all the violations.
@artembilan, please take a look when you get a chance.

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
```diff
@@ -28,13 +28,10 @@
...
 import org.springframework.core.convert.support.DefaultConversionService;
+import org.springframework.integration.support.json.Jackson2JsonObjectMapper;
 import org.springframework.integration.support.json.JacksonPresent;
 import org.springframework.integration.support.utils.IntegrationUtils;
-import org.springframework.messaging.converter.ByteArrayMessageConverter;
-import org.springframework.messaging.converter.CompositeMessageConverter;
-import org.springframework.messaging.converter.GenericMessageConverter;
-import org.springframework.messaging.converter.JacksonJsonMessageConverter;
-import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.converter.*;
```
artembilan: No asterisk imports, please.
I think there is a respective Checkstyle rule 😁.

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
```diff
@@ -28,13 +28,10 @@
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
+import org.springframework.integration.support.json.Jackson2JsonObjectMapper;
```
artembilan: I believe that this class is deprecated , therefor FQCN for it in the code, pkease

spring-integration-http/src/main/java/org/springframework/integration/http/inbound/HttpRequestHandlingEndpointSupport.java
```diff
@@ -45,6 +45,7 @@
 import org.springframework.http.converter.feed.AtomFeedHttpMessageConverter;
 import org.springframework.http.converter.feed.RssChannelHttpMessageConverter;
 import org.springframework.http.converter.json.JacksonJsonHttpMessageConverter;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
```
artembilan: DITTO

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -45,10 +45,7 @@
 import org.springframework.kafka.listener.ConsumerSeekAware;
 import org.springframework.kafka.listener.ContainerProperties;
 import org.springframework.kafka.listener.adapter.RecordMessagingMessageListenerAdapter;
-import org.springframework.kafka.support.Acknowledgment;
-import org.springframework.kafka.support.JacksonPresent;
-import org.springframework.kafka.support.JsonKafkaHeaderMapper;
-import org.springframework.kafka.support.KafkaHeaders;
+import org.springframework.kafka.support.*;
```
artembilan: DITTO

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -120,7 +117,7 @@ public KafkaInboundGateway(AbstractMessageListenerContainer<K, V> messageListene
 		}
 		else if (JacksonPresent.isJackson2Present()) {
 			MessagingMessageConverter messageConverter = new MessagingMessageConverter();
-			var headerMapper = new org.springframework.kafka.support.DefaultKafkaHeaderMapper();
+			var headerMapper = new DefaultKafkaHeaderMapper();
```
artembilan: This one is deprecated.
I thought we discussed that with you to skip.
Maybe there is a way for Checkstyle to read a file for such a class usage and spot expected `@Deprecated` at the class name ?

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -52,10 +52,7 @@
 import org.springframework.kafka.listener.adapter.FilteringMessageListenerAdapter;
 import org.springframework.kafka.listener.adapter.RecordFilterStrategy;
 import org.springframework.kafka.listener.adapter.RecordMessagingMessageListenerAdapter;
-import org.springframework.kafka.support.Acknowledgment;
-import org.springframework.kafka.support.JacksonPresent;
-import org.springframework.kafka.support.JsonKafkaHeaderMapper;
-import org.springframework.kafka.support.KafkaHeaders;
+import org.springframework.kafka.support.*;
```
artembilan: DITTO

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -147,7 +144,7 @@ public KafkaMessageDrivenChannelAdapter(AbstractMessageListenerContainer<K, V> m
 			messageConverter.setHeaderMapper(headerMapper);
 		}
 		else if (JacksonPresent.isJackson2Present()) {
-			var headerMapper = new org.springframework.kafka.support.DefaultKafkaHeaderMapper();
+			var headerMapper = new DefaultKafkaHeaderMapper();
```
artembilan: DITTO

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
```diff
@@ -66,13 +66,7 @@
...
 import org.springframework.kafka.listener.ErrorHandlingUtils;
 import org.springframework.kafka.listener.LoggingCommitCallback;
-import org.springframework.kafka.support.Acknowledgment;
-import org.springframework.kafka.support.JacksonPresent;
-import org.springframework.kafka.support.JsonKafkaHeaderMapper;
-import org.springframework.kafka.support.KafkaHeaders;
-import org.springframework.kafka.support.KafkaUtils;
-import org.springframework.kafka.support.LogIfLevelEnabled;
-import org.springframework.kafka.support.TopicPartitionOffset;
+import org.springframework.kafka.support.*;
```
artembilan: DITTO

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
```diff
@@ -273,7 +267,7 @@ public KafkaMessageSource(ConsumerFactory<K, V> consumerFactory,
 			messagingMessageConverter.setHeaderMapper(headerMapper);
 		}
 		else if (JacksonPresent.isJackson2Present()) {
-			var headerMapper = new org.springframework.kafka.support.DefaultKafkaHeaderMapper();
+			var headerMapper = new DefaultKafkaHeaderMapper();
```
artembilan: DITTO

artembilan (changes requested):
Thanks, but I don’t see an addition to the Checkstyle.xml.
Or do you mean to fix classes first and apply check because there are so many of them?

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
ZaheerAhmadDev: Done revert it.

> I think there is a respective Checkstyle rule 😁.

`avoidstarimport` check does that :).

spring-integration-core/src/main/java/org/springframework/integration/support/converter/ConfigurableCompositeMessageConverter.java
ZaheerAhmadDev: Done revert it back to `FQCN`.

spring-integration-http/src/main/java/org/springframework/integration/http/inbound/HttpRequestHandlingEndpointSupport.java
ZaheerAhmadDev: Done, now all deprecate revert back to FQCN.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
ZaheerAhmadDev: Done.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
ZaheerAhmadDev: > I thought we discussed that with you to skip.

i did skip it but some of them have been missed now all deprecated are FQCN.

> Maybe there is a way for Checkstyle to read a file for such a class usage and spot expected @Deprecated at the class name ?

Suppressionxpath can supress it https://checkstyle.sourceforge.io/filters/suppressionxpathfilter.html .

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
ZaheerAhmadDev: Done.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
ZaheerAhmadDev: Done revert it to FQCN.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
ZaheerAhmadDev: Done.

spring-integration-kafka/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
ZaheerAhmadDev: Done revert it to fullyqualified.

ZaheerAhmadDev:
> Thanks, but I don’t see an addition to the Checkstyle.xml. Or do you mean to fix classes first and apply check because there are so many of them?

currently check is not yet released it is in review, when we release it with new release, i will add it to then with followup pr.
if we add it now it will gonna show error of not found. 
i did fix all of the violation excluding `AvroTestClass`  and deprecated classes, as there is suppression for `AvroTestClass` in  `checkstyle-suppression.xml` which will exclude this check automatically also  but we need to add suppression for deprecated classes will gonna added it when we add check in `checkstyle.xml` when Checkstyle releases the check, 
in a follow up pr. 

 
Screenshot_20260817_132546

artembilan:
Thank you for the update!

When the PR build is green, I'll merge this.

Re.:

> but we need to add suppression for deprecated classes

Honestly, I'm not very happy with that `checkstyle-suppression.xml`.
The point that when we remove a deprecated API usage, we would definitely forget that there are some entries in that file.
Also, looking into the changes you provide here, I feel like there was a value in those explicit FQCN due to the fact of naming collision between different libraries.
Let's take for example a `Message` class traveling around so much!
I do see that we don't have real concern, but telling explicitly in the consumer class where we use that `Message` class from might be a bit useful when you read our classes.

And summarize: we might disable this check in the future to avoid braking our habits and not introducing extra maintenance burden with that `checkstyle-suppression.xml`.

Thank you anyway!

PR #11278, opened 2026-08-17
GH-11277: Fix local directory creation race in remote file GET

### LINKED ISSUE #11277: `AbstractRemoteFileOutboundGateway`: concurrent GET into the same new local directory fails with "Failed to make local directory"

Reproduced against `main` (7.2.0-SNAPSHOT). The affected code is identical in `v6.0.0`, `v6.3.0` and `v7.0.0`.

**Describe the bug**

When a GET (or MGET) gateway is configured with a `local-directory-expression` rather than a fixed `local-directory`, the local directory is created per message on the calling thread:

```java
private File generateLocalDirectory(Message<?> message, @Nullable String remoteDirectory) {
	...
	File localDir = ExpressionUtils.expressionToFile(this.localDirectoryExpression, evaluationContext, message,
			"Local Directory");
	if (!localDir.exists()) {
		Assert.isTrue(localDir.mkdirs(), () -> "Failed to make local directory: " + localDir);
	}
	return localDir;
}
```

This check-then-act sequence is not atomic. When two or more threads concurrently deliver messages resolving to the same, not-yet-existing local directory, all of them can pass `!localDir.exists()`, but only one `File.mkdirs()` wins. `File.mkdirs()` returns `false` when the directory already exists by the time it runs, so the losing threads fail with:

```
java.lang.IllegalArgumentException: Failed to make local directory: /path/to/local/subdir
	at org.springframework.util.Assert.isTrue(Assert.java:136)
	at org.springframework.integration.file.remote.gateway.AbstractRemoteFileOutboundGateway.generateLocalDirectory(AbstractRemoteFileOutboundGateway.java:1431)
	at org.springframework.integration.file.remote.gateway.AbstractRemoteFileOutboundGateway.get(AbstractRemoteFileOutboundGateway.java:1175)
```

even though the directory exists at that point and the transfer could simply proceed.

The eager creation in `afterPropertiesSet()` does not cover this case, because it is deliberately restricted to a `ValueExpression`:

```java
Assert.notNull(this.localDirectoryExpression, "localDirectory must not be null");
if (this.localDirectoryExpression instanceof ValueExpression) {
	setupLocalDirectory();
}
```

So with a message-dependent expression the directory is never pre-created, and creation relies entirely on the per-message path above. Any multi-threaded flow (for example a poller with a task executor, or an executor-backed channel in front of the gateway) that fetches several remote files into the same dynamically resolved local directory is affected.

This is the same defect class as GH-11253, fixed for `FileWritingMessageHandler` in GH-11254.

**To Reproduce**

Add the following test to `RemoteFileOutboundGatewayTests` (eight threads, 50 rounds, a new local directory per round). Against current `main` it fails in the first round:

```java
@Test
public void testGetConcurrentCreateSameLocalDirectory(@TempDir File localRoot) throws Exception {
	SessionFactory sessionFactory = mock(SessionFactory.class);
	TestRemoteFileOutboundGateway gw = new TestRemoteFileOutboundGateway(sessionFactory, "get", "payload");
	// A non-ValueExpression skips the eager setupLocalDirectory() in afterPropertiesSet(),
	// so the directory is created per message on the calling thread.
	gw.setLocalDirectoryExpression(
			new FunctionExpression<Message<?>>((message) -> message.getHeaders().get("localDir")));
	gw.afterPropertiesSet();
	when(sessionFactory.getSession()).thenReturn(new TestSession() {

		@Override
		public TestLsEntry[] list(String path) {
			return new TestLsEntry[] {
					new TestLsEntry(path, 1234, false, false, 12345, "-rw-r--r--")
			};
		}

		@Override
		public void read(String source, OutputStream outputStream) throws IOException {
			outputStream.write("testfile".getBytes(StandardCharsets.UTF_8));
		}

	});

	int concurrency = 8;
	ExecutorService executorService = Executors.newFixedThreadPool(concurrency);
	AtomicReference<Throwable> failure = new AtomicReference<>();
	try {
		for (int round = 0; round < 50; round++) {
			File localDirectory = new File(localRoot, "local-" + round);
			CountDownLatch getsDone = new CountDownLatch(concurrency);
			for (int i = 0; i < concurrency; i++) {
				// A distinct remote file per thread; only the destination directory is shared.
				Message<String> message = MessageBuilder.withPayload("f" + i)
						.setHeader("localDir", localDirectory.getAbsolutePath())
						.build();
				executorService.execute(() -> {
					try {
						gw.handleRequestMessage(message);
					}
					catch (Throwable ex) { // NOSONAR - the race surfaces as an unchecked exception
						failure.compareAndSet(null, ex);
					}
					finally {
						getsDone.countDown();
					}
				});
			}
			assertThat(getsDone.await(10, TimeUnit.SECONDS)).isTrue();
			assertThat(failure.get()).isNull();
			for (int i = 0; i < concurrency; i++) {
				assertThat(new File(localDirectory, "f" + i)).exists();
			}
		}
	}
	finally {
		executorService.shutdownNow();
	}
}
```

**Expected behavior**

Creation of the local directory should be idempotent and safe under concurrency. `java.nio.file.Files.createDirectories(Path)` provides exactly these semantics - it does not fail when the directory already exists, including when it was created concurrently - and reports the actual cause through `IOException` when creation genuinely fails:

```java
if (!localDir.exists()) {
	try {
		Files.createDirectories(localDir.toPath());
	}
	catch (IOException ex) {
		throw new IllegalArgumentException("Failed to make local directory: " + localDir, ex);
	}
}
```

This mirrors the fix applied to `FileWritingMessageHandler` in GH-11254 and keeps the exception type and message unchanged. With it applied locally, `:spring-integration-file:check` passes (332 tests, 0 failures) and the test above passes; reverting only the production change makes that single test fail again.

`PropertiesPersistingMetadataStore` already uses the concurrency-tolerant form (`!baseDir.mkdirs() && !baseDir.exists()`), so the idiom is established in the codebase.

**Sample**

The reproducer above is self-contained within the existing `RemoteFileOutboundGatewayTests` harness. I have the fix and the test ready and will open a pull request against this issue.

dlwldn30:
Fixes: #11277

## The defect

`AbstractRemoteFileOutboundGateway.generateLocalDirectory()` creates the local directory with a non-atomic `exists()`/`mkdirs()` sequence, per message, on the calling thread:

```java
if (!localDir.exists()) {
	Assert.isTrue(localDir.mkdirs(), () -> "Failed to make local directory: " + localDir);
}
```

The eager creation in `afterPropertiesSet()` does not cover this, because it is restricted to a `ValueExpression`:

```java
if (this.localDirectoryExpression instanceof ValueExpression) {
	setupLocalDirectory();
}
```

So with a `local-directory-expression` the directory is never pre-created and creation relies entirely on the per-message path. Concurrent messages resolving to the same, not-yet-existing directory all pass `!localDir.exists()`, only one `mkdirs()` wins, and the losing threads get `false` and fail:

```
java.lang.IllegalArgumentException: Failed to make local directory: /private/tmp/junit-.../local-0
	at org.springframework.util.Assert.isTrue(Assert.java:136)
	at org.springframework.integration.file.remote.gateway.AbstractRemoteFileOutboundGateway.generateLocalDirectory(AbstractRemoteFileOutboundGateway.java:1431)
	at org.springframework.integration.file.remote.gateway.AbstractRemoteFileOutboundGateway.get(AbstractRemoteFileOutboundGateway.java:1175)
```

The directory exists at that point and the transfer could have proceeded.

The code is identical in `v6.0.0`, `v6.3.0`, `v7.0.0` and `main`.

## The change

Use `Files.createDirectories()`, which is idempotent and does not report an already-existing directory as a failure, and propagate a genuine creation failure as the cause:

```java
if (!localDir.exists()) {
	try {
		Files.createDirectories(localDir.toPath());
	}
	catch (IOException ex) {
		throw new IllegalArgumentException("Failed to make local directory: " + localDir, ex);
	}
}
```

This mirrors `FileWritingMessageHandler`, which has used the same call for the same class of problem since #11254.

## The two adjusted tests

Attaching the `IOException` as cause moves the *root* cause of the reported `IllegalArgumentException` from itself to a `FileSystemException`:

```
MessageHandlingException
  └ MessagingException: Failed to execute on session
      └ IllegalArgumentException: Failed to make local directory     <- was the root cause
          └ FileSystemException: Read-only file system               <- is the root cause now
```

`FtpServerOutboundTests.testInt2866InvalidLocalDirectoryExpression` and its `SftpServerOutboundTests` counterpart pinned that position with `hasRootCauseInstanceOf(IllegalArgumentException.class)`. That held only because `Assert.isTrue()` left the chain empty, so the exception happened to be its own root - it was not an intentional assertion about the root.

They now assert the same type and message where they occur in the chain rather than at its root:

```java
.hasStackTraceContaining("java.lang.IllegalArgumentException: Failed to make local directory");
```

`hasCauseInstanceOf()` would not work either - the `IllegalArgumentException` sits two levels down, under the `MessagingException` from the session template.

**These assertions pass against the unfixed gateway as well**, so they assert the contract rather than encoding this change. Verified by restoring only the production file to the base commit and re-running both classes - green.

## Tests

`RemoteFileOutboundGatewayTests.testGetConcurrentCreateSameLocalDirectory` - 8 threads, 50 rounds, a fresh local directory per round, one distinct remote file per thread so only the destination directory is shared. It takes an isolated `@TempDir` parameter rather than the shared static one, so it cannot perturb `testMputRecursive`.

The previous revision was verified only against `:spring-integration-file:check`, which is how the `-ftp` and `-sftp` failures were missed. All three modules that build on this class are now in scope:

| module | result |
| --- | --- |
| `:spring-integration-file:check` | **332 tests, 0 failures, 8 skipped** |
| `:spring-integration-ftp:test` | **92 tests, 0 failures, 5 skipped** |
| `:spring-integration-sftp:test` | **92 tests, 0 failures, 1 skipped** |

`checkstyleMain` and `checkstyleTest` pass for all three.

Revert check: restoring only the production file with `git checkout <base> -- <file>` (not `git stash`, which would have been a no-op after the commit) makes exactly the new test fail - 36 tests completed, 1 failed.

## Scope

**`setupLocalDirectory()` has the same shape** (`exists()` then `mkdirs()`, at line 612) but runs once from `afterPropertiesSet()` on a single thread, so it cannot race. I left it alone to keep the diff minimal - happy to align it for consistency if you prefer.

**`AbstractInboundFileSynchronizingMessageSource` and `FileReadingMessageSource`** carry the same pattern. The former I have not traced to a concurrent caller; the latter is guarded by `running.getAndSet(true)` in `start()`. Both look safe today, but they are the same class of code and might be worth a sweep separately.

No open PR touches `AbstractRemoteFileOutboundGateway.java`.

Contributed on behalf of [Goatshave](https://github.com/Goatshave).

### DIFF
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
@@ -23,7 +23,6 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Calendar;
@@ -31,10 +30,6 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.jupiter.api.Test;
@@ -43,7 +38,6 @@ import org.mockito.ArgumentCaptor;
 
 import org.springframework.expression.common.LiteralExpression;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
-import org.springframework.integration.expression.FunctionExpression;
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.file.filters.AbstractSimplePatternFileListFilter;
 import org.springframework.integration.file.remote.AbstractFileInfo;
@@ -78,7 +72,6 @@ import static org.mockito.Mockito.when;
  * @author Liu Jiong
  * @author Artem Bilan
  * @author Jooyoung Pyoung
- * @author Jiwoo Lee
  *
  * @since 2.1
  */
@@ -814,64 +807,6 @@ public class RemoteFileOutboundGatewayTests implements TestApplicationContextAwa
 		File out = new File(this.tmpDir + "/x/f1");
 		assertThat(out.exists()).isTrue();
 		out.delete();
-	}
-
-	@Test
-	public void testGetConcurrentCreateSameLocalDirectory(@TempDir File localRoot) throws Exception {
-		SessionFactory sessionFactory = mock();
-		TestRemoteFileOutboundGateway gw = new TestRemoteFileOutboundGateway(sessionFactory, "get", "payload");
-		gw.setLocalDirectoryExpression(
-				new FunctionExpression<Message<?>>((message) -> message.getHeaders().get("localDir")));
-		gw.afterPropertiesSet();
-		when(sessionFactory.getSession()).thenReturn(new TestSession() {
-
-			@Override
-			public TestLsEntry[] list(String path) {
-				return new TestLsEntry[] {
-						new TestLsEntry(path, 1234, false, false, 12345, "-rw-r--r--")
-				};
-			}
-
-			@Override
-			public void read(String source, OutputStream outputStream) throws IOException {
-				outputStream.write("testfile".getBytes(StandardCharsets.UTF_8));
-			}
-
-		});
-
-		int concurrency = 8;
-		ExecutorService executorService = Executors.newFixedThreadPool(concurrency);
-		AtomicReference<Throwable> failure = new AtomicReference<>();
-		try {
-			for (int round = 0; round < 50; round++) {
-				File localDirectory = new File(localRoot, "local-" + round);
-				CountDownLatch getsDone = new CountDownLatch(concurrency);
-				for (int i = 0; i < concurrency; i++) {
-					Message<String> message = MessageBuilder.withPayload("f" + i)
-							.setHeader("localDir", localDirectory.getAbsolutePath())
-							.build();
-					executorService.execute(() -> {
-						try {
-							gw.handleRequestMessage(message);
-						}
-						catch (Throwable ex) {
-							failure.compareAndSet(null, ex);
-						}
-						finally {
-							getsDone.countDown();
-						}
-					});
-				}
-				assertThat(getsDone.await(10, TimeUnit.SECONDS)).isTrue();
-				assertThat(failure.get()).isNull();
-				for (int i = 0; i < concurrency; i++) {
-					assertThat(new File(localDirectory, "f" + i)).exists();
-				}
-			}
-		}
-		finally {
-			executorService.shutdownNow();
-		}
 	}
 
 	@Test
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
@@ -23,6 +23,7 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Calendar;
@@ -30,6 +31,10 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.jupiter.api.Test;
@@ -38,6 +43,7 @@ import org.mockito.ArgumentCaptor;
 
 import org.springframework.expression.common.LiteralExpression;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
+import org.springframework.integration.expression.FunctionExpression;
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.file.filters.AbstractSimplePatternFileListFilter;
 import org.springframework.integration.file.remote.AbstractFileInfo;
@@ -72,6 +78,7 @@ import static org.mockito.Mockito.when;
  * @author Liu Jiong
  * @author Artem Bilan
  * @author Jooyoung Pyoung
+ * @author Jiwoo Lee
  *
  * @since 2.1
  */
@@ -807,6 +814,67 @@ public class RemoteFileOutboundGatewayTests implements TestApplicationContextAwa
 		File out = new File(this.tmpDir + "/x/f1");
 		assertThat(out.exists()).isTrue();
 		out.delete();
+	}
+
+	@Test
+	public void testGetConcurrentCreateSameLocalDirectory(@TempDir File localRoot) throws Exception {
+		SessionFactory sessionFactory = mock(SessionFactory.class);
+		TestRemoteFileOutboundGateway gw = new TestRemoteFileOutboundGateway(sessionFactory, "get", "payload");
+		// A non-ValueExpression skips the eager setupLocalDirectory() in afterPropertiesSet(),
+		// so the directory is created per message on the calling thread.
+		gw.setLocalDirectoryExpression(
+				new FunctionExpression<Message<?>>((message) -> message.getHeaders().get("localDir")));
+		gw.afterPropertiesSet();
+		when(sessionFactory.getSession()).thenReturn(new TestSession() {
+
+			@Override
+			public TestLsEntry[] list(String path) {
+				return new TestLsEntry[] {
+						new TestLsEntry(path, 1234, false, false, 12345, "-rw-r--r--")
+				};
+			}
+
+			@Override
+			public void read(String source, OutputStream outputStream) throws IOException {
+				outputStream.write("testfile".getBytes(StandardCharsets.UTF_8));
+			}
+
+		});
+
+		int concurrency = 8;
+		ExecutorService executorService = Executors.newFixedThreadPool(concurrency);
+		AtomicReference<Throwable> failure = new AtomicReference<>();
+		try {
+			for (int round = 0; round < 50; round++) {
+				File localDirectory = new File(localRoot, "local-" + round);
+				CountDownLatch getsDone = new CountDownLatch(concurrency);
+				for (int i = 0; i < concurrency; i++) {
+					// A distinct remote file per thread; only the destination directory is shared.
+					Message<String> message = MessageBuilder.withPayload("f" + i)
+							.setHeader("localDir", localDirectory.getAbsolutePath())
+							.build();
+					executorService.execute(() -> {
+						try {
+							gw.handleRequestMessage(message);
+						}
+						catch (Throwable ex) { // NOSONAR - the race surfaces as an unchecked exception
+							failure.compareAndSet(null, ex);
+						}
+						finally {
+							getsDone.countDown();
+						}
+					});
+				}
+				assertThat(getsDone.await(10, TimeUnit.SECONDS)).isTrue();
+				assertThat(failure.get()).isNull();
+				for (int i = 0; i < concurrency; i++) {
+					assertThat(new File(localDirectory, "f" + i)).exists();
+				}
+			}
+		}
+		finally {
+			executorService.shutdownNow();
+		}
 	}
 
 	@Test
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
@@ -818,10 +818,8 @@ public class RemoteFileOutboundGatewayTests implements TestApplicationContextAwa
 
 	@Test
 	public void testGetConcurrentCreateSameLocalDirectory(@TempDir File localRoot) throws Exception {
-		SessionFactory sessionFactory = mock(SessionFactory.class);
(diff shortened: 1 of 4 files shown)


cppwfs (changes requested):

When building the full project we see that there are test failures in `FtpServerOutboundTests` and `SftpServerOutboundTests`.

dlwldn30:
Reproduced both failures locally - `FtpServerOutboundTests.testInt2866InvalidLocalDirectoryExpression` and its `SftpServerOutboundTests` counterpart:

```
Expecting a throwable with root cause being an instance of:
  java.lang.IllegalArgumentException
but was an instance of:
  java.nio.file.FileSystemException
```

Both assert the *root* cause. Attaching the `IOException` as the cause of the `IllegalArgumentException` moved the root cause to `FileSystemException`. My description claimed nothing observable changes - the type and the message were kept, but the added cause is observable, and these two tests observe exactly that. I had only run `:spring-integration-file:check`, which is why I missed it.

Fixed by leaving the exception untouched and making the `mkdirs()` result race-tolerant instead:

```java
if (!localDir.exists()) {
	Assert.isTrue(localDir.mkdirs() || localDir.exists(),
			() -> "Failed to make local directory: " + localDir);
}
```

This is the idiom already used by `PropertiesPersistingMetadataStore`, and it closes the race the same way - `mkdirs()` returning `false` is only a failure when the directory is still absent.

Verification:

- `:spring-integration-file:check` - 332 tests, 0 failures, 8 skipped; `checkstyleMain` and `checkstyleTest` pass
- `FtpServerOutboundTests` - 24 tests, 0 failures
- `SftpServerOutboundTests` - 20 tests, 0 failures
- Revert check: restoring only the production file with `git checkout HEAD~1 -- <file>` makes `testGetConcurrentCreateSameLocalDirectory` fail again (36 tests completed, 1 failed)

One point for you to weigh rather than me: `Files.createDirectories()` is what `FileWritingMessageHandler` has used since #11254, so dropping it here costs that consistency and the underlying `IOException` as a cause. Keeping it would mean relaxing the root-cause assertion in those two tests. I went with the form that leaves the existing tests untouched, since changing them to accommodate my own change looked like the wrong trade - happy to switch if you prefer the `Files.createDirectories()` version.

The PR description is updated accordingly.

artembilan:
I think moving to the `Files.createDirectories()` is the right direction for consistency and less stacktrace.
Consider to adjust those tests instead.
Even if this is a bit of a breaking change due to different exception, it is likely better than more stacktrace or even race condition.

dlwldn30:
Thanks - switched to `Files.createDirectories()` and adjusted the two tests.

One thing worth flagging, since it changed what the adjustment had to look like. The `IllegalArgumentException` is not the direct cause either, so swapping `hasRootCauseInstanceOf` for `hasCauseInstanceOf` does not work - it sits two levels down:

```
MessageHandlingException
  └ MessagingException: Failed to execute on session
      └ IllegalArgumentException: Failed to make local directory     <- was the root cause
          └ FileSystemException: Read-only file system               <- is the root cause now
```

The old assertion held only because `Assert.isTrue()` left the chain empty, so the exception happened to be its own root. Rather than re-pin it to `FileSystemException` - which would pin a position again, and couple the test to whichever `FileSystemException` subtype the platform throws - both tests now assert the type and message wherever they occur in the chain:

```java
.hasStackTraceContaining("java.lang.IllegalArgumentException: Failed to make local directory");
```

**These pass against the unfixed gateway too**, so they are not written around this change. I checked by restoring only the production file to the base commit and re-running both classes - green.

Verification now covers every module that builds on this class, which is what I should have done the first time:

| module | result |
| --- | --- |
| `:spring-integration-file:check` | 332 tests, 0 failures, 8 skipped |
| `:spring-integration-ftp:test` | 92 tests, 0 failures, 5 skipped |
| `:spring-integration-sftp:test` | 92 tests, 0 failures, 1 skipped |

`checkstyleMain` and `checkstyleTest` pass for all three. Revert check still valid - restoring only the production file makes exactly the new concurrency test fail.

The PR description is updated accordingly.

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/outbound/FtpServerOutboundTests.java
```diff
@@ -211,8 +211,7 @@ public void testGetWithRemove() {
 	@Test
 	public void testInt2866InvalidLocalDirectoryExpression() {
 		assertThatCode(() -> this.invalidDirExpression.send(new GenericMessage<Object>("/ftpSource/ ftpSource1.txt")))
-				.hasRootCauseInstanceOf(IllegalArgumentException.class)
```
cppwfs, line 214: Thank you for the update!  
There is a need for us to test the type of exception that was thrown as well.   So let's add that back to the test but checking for `FileSystemException`.

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/outbound/SftpServerOutboundTests.java
```diff
@@ -191,8 +191,7 @@ public void testInt2866InvalidLocalDirectoryExpression() {
 		assertThatExceptionOfType(Exception.class)
 				.isThrownBy(() ->
 						this.invalidDirExpression.send(new GenericMessage<Object>("sftpSource/ sftpSource1.txt")))
-				.withRootCauseInstanceOf(IllegalArgumentException.class)
```
cppwfs, line 194: Ditto

cppwfs (changes requested):
Thank you again for the contribution!

spring-integration-ftp/src/test/java/org/springframework/integration/ftp/outbound/FtpServerOutboundTests.java
dlwldn30: Restored the type assertion as `hasRootCauseInstanceOf(FileSystemException.class)`, alongside the existing message check. Thank you!

spring-integration-sftp/src/test/java/org/springframework/integration/sftp/outbound/SftpServerOutboundTests.java
dlwldn30: Done here as well.

dlwldn30:
@cppwfs friendly ping on this one.

Both tests now assert `hasRootCauseInstanceOf(FileSystemException.class)` alongside the message check, as requested.

The build is still waiting on workflow approval, so for reference, locally:

- `:spring-integration-file:check` - 332 tests, 0 failures, 8 skipped
- `:spring-integration-ftp:test` - 92 tests, 0 failures, 5 skipped
- `:spring-integration-sftp:test` - 92 tests, 0 failures, 1 skipped

`checkstyleMain` and `checkstyleTest` pass for all three.

Happy to adjust anything else.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
```diff
@@ -809,6 +816,67 @@ public void read(String source, OutputStream outputStream) throws IOException {
 		out.delete();
 	}
 
+	@Test
+	public void testGetConcurrentCreateSameLocalDirectory(@TempDir File localRoot) throws Exception {
+		SessionFactory sessionFactory = mock(SessionFactory.class);
```
artembilan: Use `mock()` without arguments.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
```diff
@@ -809,6 +816,67 @@ public void read(String source, OutputStream outputStream) throws IOException {
...
+				for (int i = 0; i < concurrency; i++) {
+					// A distinct remote file per thread; only the destination directory is shared.
+					Message<String> message = MessageBuilder.withPayload("f" + i)
+							.setHeader("localDir", localDirectory.getAbsolutePath())
+							.build();
+					executorService.execute(() -> {
+						try {
+							gw.handleRequestMessage(message);
+						}
+						catch (Throwable ex) { // NOSONAR - the race surfaces as an unchecked exception
```
artembilan: We don't work with SonarQube any more.
Therefore, we don't need all of these extra noisy comment.

artembilan (changes requested):

cppwfs:
Again thank you for the contribution.   Also rebase your PR and that will resolve the build failure.

spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
```diff
@@ -809,6 +816,64 @@ public void read(String source, OutputStream outputStream) throws IOException {
...
+				outputStream.write("testfile".getBytes(StandardCharsets.UTF_8));
+			}
+
+		});
+
+		int concurrency = 8;
+		ExecutorService executorService = Executors.newFixedThreadPool(concurrency);
+		AtomicReference<Throwable> failure = new AtomicReference<>();
+		try {
+			for (int round = 0; round < 50; round++) {
```
artembilan: How slow is this test?
I mean you made the point with this verification that atomic Java native `Files.createDirectories()` feature works well.
But is it really our project responsibility to test that method behavior?
I feel like we don't need it at all since it is out of scope.

artembilan (changes requested):
Just one question.
Thank you!

spring-integration-file/src/test/java/org/springframework/integration/file/remote/gateway/RemoteFileOutboundGatewayTests.java
dlwldn30: 0.541s in a full class run — 62% of `RemoteFileOutboundGatewayTests` test time, against an 8ms median there. Not slow on its own, but disproportionate for what it covers.

Agreed on the scope, so it is removed. The imports it pulled in went with it, so the test class is untouched by this PR now — what is left is the gateway fix plus the two assertion updates.

Locally: `:spring-integration-file:check` 331 tests, `:spring-integration-ftp:test` 92, `:spring-integration-sftp:test` 92, all green, checkstyle clean.

PR #11279, opened 2026-08-17
feat(core): add Embabel gRPC correlating agent

mohamed-zaytoon:


artembilan:
This looks out of scope, mixing concerns and I even see it is harmful since it remove a lot of logic from the `AbstractCorrelatingMessageHandler`.

Closed as `Invalid`.
Consider to make a separate project as an extension to the API we provide here so far.
The AI agent cannot be as a replacement of the logic we provide here.

Thanks for understanding!

PR #11351, opened 2026-08-20
Use latest Spring Framework 7.1.0-SNAPSHOT

cppwfs:
The latest snapshots have introduced some breaking changes.
    Resolve the issues below
  - Treat a null result from `MessageConverter.fromMessage()` as a
  conversion failure across all inbound JMS paths.
  - Throw `MessageConversionException` from `AbstractJmsChannel`,
  `JmsDestinationPollingSource`, `ChannelPublishingJmsMessageListener`,
  and `JmsOutboundGateway` instead of silently discarding the message.
  - Remove the discard branch in `SubscribableJmsChannel `since a null
  conversion is no longer possible without an exception.
  - Add tests asserting `MessageConversionException` for each of the
  four affected components.
  - Updated core to use builder for `SpelParserConfiguration` as the
   constructor is now deprecated.

### DIFF
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
@@ -94,6 +94,11 @@ public abstract class AbstractJmsChannel extends AbstractMessageChannel {
 		return true;
 	}
 
+	/**
+	 * Converts {@link jakarta.jms.Message} to a Spring {@link Message}.
+	 * @return Spring {@link Message}
+	 * @throws MessageConversionException if the converter returns {@code null}.
+	 */
 	protected Message<?> fromJmsMessage(jakarta.jms.Message message) {
 		MessageConverter converter = this.jmsTemplate.getMessageConverter();
 		try {
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
@@ -24,6 +24,7 @@ import org.springframework.integration.jms.DefaultJmsHeaderMapper;
 import org.springframework.integration.jms.DynamicJmsTemplateProperties;
 import org.springframework.integration.jms.JmsHeaderMapper;
 import org.springframework.jms.core.JmsTemplate;
+import org.springframework.jms.support.converter.MessageConversionException;
 import org.springframework.jms.support.converter.MessageConverter;
 import org.springframework.jms.support.converter.MessagingMessageConverter;
 import org.springframework.messaging.Message;
@@ -97,6 +98,9 @@ public abstract class AbstractJmsChannel extends AbstractMessageChannel {
 		MessageConverter converter = this.jmsTemplate.getMessageConverter();
 		try {
 			Object converted = converter.fromMessage(message);
+			if (converted == null) {
+				throw new MessageConversionException("Message converter returned null for: " + message);
+			}
 			Message<?> messageToSend;
 			if (converted instanceof Message<?> convertedMessage) {
 				messageToSend = convertedMessage;
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
@@ -95,7 +95,7 @@ public abstract class AbstractJmsChannel extends AbstractMessageChannel {
 	}
 
 	/**
-	 * Converts {@link jakarta.jms.Message} to a Spring {@link Message}.
+	 * Convert {@link jakarta.jms.Message} to a Spring {@link Message}.
 	 * @return Spring {@link Message}
 	 * @throws MessageConversionException if the converter returns {@code null}.
 	 */
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
@@ -83,6 +83,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Gary Russell
  * @author Ngoc Nhan
+ * @author Glenn Renfro
  *
  * @since 7.0
  */
@@ -400,6 +401,19 @@ public class ChannelPublishingJmsMessageListener
 		this.beanFactory = beanFactory;
 	}
 
+	/**
+	 * Processes an incoming JMS message, transforms it into a Spring Integration message,
+	 * delegates execution to the message gateway, and optionally constructs and sends a reply.
+	 * If inbound message conversion or header mapping fails, the error is routed to the configured
+	 * error channel on the gateway delegate. If no error channel is present, the runtime exception
+	 * is rethrown. When {@code expectReply} is enabled, the resulting reply is converted back to a JMS message,
+	 * populated with headers and correlation identifiers from the original request, and dispatched
+	 * to the resolved JMS destination.
+	 * @param jmsMessage the incoming {@link jakarta.jms.Message} to process
+	 * @param session the active JMS {@link jakarta.jms.Session} associated with the listener
+	 * @throws JMSException if a native JMS error occurs while processing, resolving destinations, or sending replies
+	 * @throws MessageConversionException if payload extraction produces a {@code null} result when required
+	 */
 	@Override
 	public void onMessage(jakarta.jms.Message jmsMessage, Session session) throws JMSException {
 		Message<?> requestMessage;
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
@@ -60,6 +60,7 @@ import org.springframework.integration.support.utils.IntegrationUtils;
 import org.springframework.jms.listener.SessionAwareMessageListener;
 import org.springframework.jms.listener.adapter.ListenerExecutionFailedException;
 import org.springframework.jms.support.JmsUtils;
+import org.springframework.jms.support.converter.MessageConversionException;
 import org.springframework.jms.support.converter.MessageConverter;
 import org.springframework.jms.support.converter.SimpleMessageConverter;
 import org.springframework.jms.support.destination.DestinationResolver;
@@ -408,6 +409,9 @@ public class ChannelPublishingJmsMessageListener
 				result = this.messageConverter.fromMessage(jmsMessage);
 				this.logger.debug(() -> "converted JMS Message [" + jmsMessage + "] to integration Message payload ["
 						+ result + "]");
+				if (result == null) {
+					throw new MessageConversionException("Message converter returned null for: " + jmsMessage);
+				}
 			}
 			else {
 				result = jmsMessage;
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
@@ -402,17 +402,13 @@ public class ChannelPublishingJmsMessageListener
 	}
 
 	/**
-	 * Processes an incoming JMS message, transforms it into a Spring Integration message,
-	 * delegates execution to the message gateway, and optionally constructs and sends a reply.
-	 * If inbound message conversion or header mapping fails, the error is routed to the configured
-	 * error channel on the gateway delegate. If no error channel is present, the runtime exception
-	 * is rethrown. When {@code expectReply} is enabled, the resulting reply is converted back to a JMS message,
-	 * populated with headers and correlation identifiers from the original request, and dispatched
-	 * to the resolved JMS destination.
+	 * Convert the JMS message to a Spring Integration message and send it to the gateway,
+	 * optionally sending a reply back to the JMS destination.
 	 * @param jmsMessage the incoming {@link jakarta.jms.Message} to process
 	 * @param session the active JMS {@link jakarta.jms.Session} associated with the listener
 	 * @throws JMSException if a native JMS error occurs while processing, resolving destinations, or sending replies
-	 * @throws MessageConversionException if payload extraction produces a {@code null} result when required
+	 * @throws MessageConversionException if request payload extraction produces a {@code null} result and no
+	 * error channel is configured, or if the reply message cannot be converted to a JMS message
 	 */
 	@Override
 	public void onMessage(jakarta.jms.Message jmsMessage, Session session) throws JMSException {
@@ -421,10 +417,12 @@ public class ChannelPublishingJmsMessageListener
 			final Object result;
 			if (this.extractRequestPayload) {
 				result = this.messageConverter.fromMessage(jmsMessage);
-				this.logger.debug(() -> "converted JMS Message [" + jmsMessage + "] to integration Message payload ["
-						+ result + "]");
 				if (result == null) {
 					throw new MessageConversionException("Message converter returned null for: " + jmsMessage);
+				}
+				else {
+					this.logger.debug(() -> "converted JMS Message [" + jmsMessage + "] to integration Message payload ["
+							+ result + "]");
 				}
 			}
 			else {
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSource.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSource.java
@@ -27,6 +27,7 @@ import org.springframework.integration.jms.JmsHeaderMapper;
 import org.springframework.integration.jms.util.JmsAdapterUtils;
 import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.jms.core.JmsTemplate;
+import org.springframework.jms.support.converter.MessageConversionException;
 import org.springframework.jms.support.converter.MessageConverter;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
@@ -139,6 +140,9 @@ public class JmsDestinationPollingSource extends AbstractMessageSource<Object> {
 				MessageConverter converter = this.jmsTemplate.getMessageConverter();
 				if (converter != null) {
 					object = converter.fromMessage(jmsMessage);
+					if (object == null) {
+						throw new MessageConversionException("Message converter returned null for: " + jmsMessage);
+					}
 				}
 			}
 			AbstractIntegrationMessageBuilder<?> builder =
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSource.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSource.java
@@ -42,6 +42,7 @@ import org.springframework.util.Assert;
  * @author Mark Fisher
  * @author Oleg Zhurakousky
  * @author Artem Bilan
+ * @author Glenn Renfro
  *
  * @since 7.0
  */
@@ -125,6 +126,7 @@ public class JmsDestinationPollingSource extends AbstractMessageSource<Object> {
 	 * Will receive a JMS {@link jakarta.jms.Message} converting and returning it as
 	 * a Spring Integration {@link Message}. This method will also use the current
 	 * {@link JmsHeaderMapper} instance to map JMS properties to the MessageHeaders.
+	 * @throws MessageConversionException if the converter returns {@code null}.
 	 */
 	@Override
 	protected @Nullable Object doReceive() {
@@ -138,11 +140,9 @@ public class JmsDestinationPollingSource extends AbstractMessageSource<Object> {
 			Object object = jmsMessage;
 			if (this.extractPayload) {
 				MessageConverter converter = this.jmsTemplate.getMessageConverter();
-				if (converter != null) {
-					object = converter.fromMessage(jmsMessage);
-					if (object == null) {
-						throw new MessageConversionException("Message converter returned null for: " + jmsMessage);
-					}
+				object = converter.fromMessage(jmsMessage);
+				if (object == null) {
+					throw new MessageConversionException("Message converter returned null for: " + jmsMessage);
 				}
 			}
 			AbstractIntegrationMessageBuilder<?> builder =
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSource.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSource.java
@@ -123,10 +123,11 @@ public class JmsDestinationPollingSource extends AbstractMessageSource<Object> {
 	}
 
 	/**
-	 * Will receive a JMS {@link jakarta.jms.Message} converting and returning it as
-	 * a Spring Integration {@link Message}. This method will also use the current
+	 * Receive a JMS {@link jakarta.jms.Message}, converting and returning it as
+	 * a Spring Integration {@link Message}. Use the current
 	 * {@link JmsHeaderMapper} instance to map JMS properties to the MessageHeaders.
-	 * @throws MessageConversionException if the converter returns {@code null}.
+	 * @throws MessagingException if header mapping or payload conversion fails, e.g. wrapping a
+	 * {@link MessageConversionException} when the converter returns {@code null} or fails to convert.
 	 */
 	@Override
 	protected @Nullable Object doReceive() {
--- a/spring-integration-jms/src/main/java/org/springframework/integration/jms/outbound/JmsOutboundGateway.java
+++ b/spring-integration-jms/src/main/java/org/springframework/integration/jms/outbound/JmsOutboundGateway.java
@@ -725,9 +725,9 @@ public class JmsOutboundGateway extends AbstractReplyProducingMessageHandler
 	}
 
 	/**
-	 * Handles an incoming request message by dispatching it over JMS and returning the reply.
-	 * Ensures the component is initialized before processing. If a reply container is configured,
-	 * it manages the container lifecycle by auto-starting it and scheduling an idle stopper task
+	 * Handle an incoming request message by dispatching it over JMS and returning the reply.
+	 * Ensure the component is initialized before processing. If a reply container is configured,
+	 * manage the container lifecycle by auto-starting it and scheduling an idle stopper task
 	 * when an idle timeout is defined.
 	 * @param requestMessage the Spring Integration {@link Message} to process and send
 	 * @return the processed reply object or built reply message, or {@code null} if no reply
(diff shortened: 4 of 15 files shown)


spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/SubscribableJmsChannel.java
```diff
@@ -161,6 +161,11 @@ public void destroy() {
 
 	private void receiveAndDispatch(jakarta.jms.Message message) {
 		Message<?> messageToSend = fromJmsMessage(message);
+		if (messageToSend == null) {
+			this.logger.debug(() -> "The JMS Message [" + message
```
artembilan: I think we need to look into details how Spring Framework JMS deals with such a `null` after conversion: https://github.com/spring-projects/spring-framework/blob/main/spring-jms/src/main/java/org/springframework/jms/support/converter/MessageConverter.java#L50

More over, it looks like you are inconsistent in handling it in your proposal.
All the cases are `inbound`.
Why do process them differently in similar components?

artembilan (changes requested):
Please, repurpose this PR just for SF `7.2.0-SNAPSHOT`.
Looks like this is going to take a while until it is in a good shape.
The rest of dependencies should not suffer.

Thanks

cppwfs:
This change includes those that were in PR https://github.com/spring-projects/spring-integration/pull/11367.   So if this one still needs alot of work then merge 11367 and I'll continue work on this PR.

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,80 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.nullable;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+
+/**
+ * @author Glenn Renfro
```
artembilan, line 34: `@since 7.2` ?

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class JmsDestinationPollingSourceTests implements TestApplicationContextAware {
```
artembilan: I think we agreed that `public` is redundant in tests.

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,80 @@
...
+		given(jmsTemplate.receiveSelected(nullable(String.class))).willReturn(jmsMessage);
+
+		JmsDestinationPollingSource source = new JmsDestinationPollingSource(jmsTemplate);
+		source.setBeanFactory(TEST_INTEGRATION_CONTEXT);
+		source.afterPropertiesSet();
+
+		// A JMS message converted to a null payload is a conversion failure, not a message to discard.
+		assertThatExceptionOfType(MessagingException.class)
+				.isThrownBy(source::receive)
+				.withCauseInstanceOf(MessageConversionException.class);
```
artembilan, line 54: Please, confirm that `MessageConversionException` is what SF does now in JMS module.

artembilan (changes requested):
I think this is something what deserves some docs, too.
Including `whats-new`.

Thanks

artembilan:
I mean.
Document JMS part as it is some kind of breaking change.

artembilan:
Also, please, add a comment into this PR with a `Attention required:`  hint explaining how JMS module now fails for null conversion results.
Thanks

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
cppwfs: Spring Framework seems to follow the documentation JMSMessage:

1. Implementations should return null if for example  if  `jakarta.jms.TextMessage` or `jakarta.jms.ObjectMessage` whose body was never set.
2. Throws `jakarta.jms.JMSException` if thrown by JMS API methods.
3. Throws `MessageConversionException` in case of conversion failure.

In our case I went the strict path and stated that a `null` throws an exception.   But I think that is in error.   Rather it should set the payload of the message to `null`.   But wanted to run that past you before I make that change.

artembilan: >Rather it should set the payload of the message to `null`. 

That's the point: we cannot do that since `Message` payload cannot be `null`.

Please, investigate what Spring Framework JMS module does with `null` result from the converted.
For example, look into `AbstractAdaptableMessageListener`, or whatever place there is using the mentioned `fromMessage()` API.

I've seen Javadocs, but that is not obvious what direction is right for consistency.
That's why I asked you to look for inspiration in Spring Framework.

cppwfs: JMS allows messages to have an unset/empty body, meaning MessageConverter.fromMessage(...) legitimately returns null. However, when creating a `org.springframework.messaging.Message` instance it requires a non-null payload.

In order to support JMS Spring Framework allows the `org.springframework.jms.support.MessageConverter` to return a null if the body of the `JmsMessage` was never set when converting the payload.
The change noted here shows this is as the goal of this commit: https://github.com/spring-projects/spring-framework/commit/ba13cae6ccad7cac1f881859efd334b64835abc8
However, we do see that this change honors the fact that a Spring Message can not be created with a `null` payload as denoted in the changes in `org.springframework.jms.support.converter.MessagingMessageConverter`.   
And when the JMS Messages payload has not been set a `MessageConversionException` is thrown when creating a new Spring Message with this `null` payload.

This PR reflects the same behavior as shown in the Spring Framework when handling a `null` payload returned from a `MessageCoverter` for a JMS message when creating a new Spring message with this payload.

spring-integration-jms/src/main/java/org/springframework/integration/jms/channel/AbstractJmsChannel.java
```diff
@@ -93,10 +94,18 @@ protected boolean doSend(Message<?> message, long timeout) {
 		return true;
 	}
 
+	/**
+	 * Converts {@link jakarta.jms.Message} to a Spring {@link Message}.
```
artembilan: Method Javadocs must be imperative, like commands.

spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
```diff
@@ -399,6 +401,19 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
 		this.beanFactory = beanFactory;
 	}
 
+	/**
+	 * Processes an incoming JMS message, transforms it into a Spring Integration message,
```
artembilan: I think this Javadoc was generated by AI.
Please, revise it as it is too much info.
And it is not imperative.

spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
```diff
@@ -399,6 +401,19 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
...
+	 * delegates execution to the message gateway, and optionally constructs and sends a reply.
+	 * If inbound message conversion or header mapping fails, the error is routed to the configured
+	 * error channel on the gateway delegate. If no error channel is present, the runtime exception
+	 * is rethrown. When {@code expectReply} is enabled, the resulting reply is converted back to a JMS message,
+	 * populated with headers and correlation identifiers from the original request, and dispatched
+	 * to the resolved JMS destination.
+	 * @param jmsMessage the incoming {@link jakarta.jms.Message} to process
+	 * @param session the active JMS {@link jakarta.jms.Session} associated with the listener
+	 * @throws JMSException if a native JMS error occurs while processing, resolving destinations, or sending replies
+	 * @throws MessageConversionException if payload extraction produces a {@code null} result when required
```
artembilan: I don't think this is only case.
The message converter may thrown such an exception for other reason.
We probably should rephrase it somehow saying that in case of null such an exception is thrown, too.
Or what is better wording would it be from English perspective 😄

spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/ChannelPublishingJmsMessageListener.java
```diff
@@ -408,6 +423,9 @@ public void onMessage(jakarta.jms.Message jmsMessage, Session session) throws JM
 				result = this.messageConverter.fromMessage(jmsMessage);
 				this.logger.debug(() -> "converted JMS Message [" + jmsMessage + "] to integration Message payload ["
```
artembilan: And no this debug message is suspicious with `null` result.
I think we need to revise logic as `if..else`.

spring-integration-jms/src/main/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSource.java
```diff
@@ -124,6 +126,7 @@ protected void onInit() {
 	 * Will receive a JMS {@link jakarta.jms.Message} converting and returning it as
 	 * a Spring Integration {@link Message}. This method will also use the current
 	 * {@link JmsHeaderMapper} instance to map JMS properties to the MessageHeaders.
+	 * @throws MessageConversionException if the converter returns {@code null}.
```
artembilan: DITTO

spring-integration-jms/src/main/java/org/springframework/integration/jms/outbound/JmsOutboundGateway.java
```diff
@@ -722,6 +724,21 @@ public boolean isRunning() {
 		return this.active;
 	}
 
+	/**
+	 * Handles an incoming request message by dispatching it over JMS and returning the reply.
```
artembilan: DITTO

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/PollableJmsChannelTests.java
```diff
@@ -87,6 +91,46 @@ public void queueReference() throws Exception {
...
+		factoryBean.setConnectionFactory(ccf);
+		factoryBean.setDestinationName("nullPayloadPollableQueue");
+		factoryBean.setPubSubDomain(false);
+		factoryBean.setBeanFactory(mock());
+		factoryBean.afterPropertiesSet();
+		PollableJmsChannel channel = (PollableJmsChannel) factoryBean.getObject();
+		assertThat(channel.send(new GenericMessage<>("test1"))).isTrue();
+
+		// A JMS message converted to a null payload is a conversion failure, not a message to discard.
+		new DirectFieldAccessor(channel).setPropertyValue("jmsTemplate", nullConvertingTemplate(ccf));
```
artembilan: Why do we need to do all of this if we have `JmsChannelFactoryBean.setMessageConverter()`?

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/SubscribableJmsChannelTests.java
```diff
@@ -282,6 +283,26 @@ public void dispatcherHasNoSubscribersQueue() throws Exception {
 				.withMessageContaining("Dispatcher has no subscribers for jms-channel 'noSubscribersChannel'.");
 	}
 
+	@Test
+	public void nullPayloadFromConverterThrowsMessageConversionException() throws Exception {
+		JmsChannelFactoryBean factoryBean = new JmsChannelFactoryBean(true);
+		factoryBean.setConnectionFactory(connectionFactory);
+		factoryBean.setDestinationName("nullPayloadSubscribableQueue");
+		factoryBean.setBeanName("nullPayloadChannel");
+		factoryBean.setMessageConverter(mock());
```
artembilan: I think this `mock` trick could be used in the previous `PollableJmsChannelTests` as well.

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/SubscribableJmsChannelTests.java
```diff
@@ -282,6 +283,26 @@ public void dispatcherHasNoSubscribersQueue() throws Exception {
...
+		factoryBean.setBeanFactory(mock());
+		factoryBean.afterPropertiesSet();
+		SubscribableJmsChannel channel = (SubscribableJmsChannel) factoryBean.getObject();
+		channel.afterPropertiesSet();
+
+		AbstractMessageListenerContainer container = TestUtils.getPropertyValue(channel, "container");
+		MessageListener listener = (MessageListener) container.getMessageListener();
+
+		// A JMS message converted to a null payload is a conversion failure, not a message to discard.
```
artembilan: I don't see value in this comment any more.

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
artembilan: Not fixed.
I see changes in the test methods, but class declaration could also benefit from not having `public`  modifier 😄

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,82 @@
...
+
+		JmsTemplate jmsTemplate = mock();
+		given(jmsTemplate.getMessageConverter()).willReturn(converter);
+		given(jmsTemplate.receiveSelected(nullable(String.class))).willReturn(jmsMessage);
+
+		JmsDestinationPollingSource source = new JmsDestinationPollingSource(jmsTemplate);
+		source.setBeanFactory(TEST_INTEGRATION_CONTEXT);
+		source.afterPropertiesSet();
+
+		// A JMS message converted to a null payload is a conversion failure, not a message to discard.
```
artembilan: DITTO

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,82 @@
...
+		assertThatExceptionOfType(MessagingException.class)
+				.isThrownBy(source::receive)
+				.withCauseInstanceOf(MessageConversionException.class);
+	}
+
+	@Test
+	void payloadFromConverterIsUsed() throws Exception {
```
artembilan: I don't see value in this test.
Or better to say: we do have such a coverage in other tests in the module.

spring-integration-jms/src/test/java/org/springframework/integration/jms/outbound/JmsOutboundGatewayTests.java
```diff
@@ -155,6 +160,36 @@ public void testReplyContainerRecovery() throws Exception {
...
+		when(connectionFactory.createConnection()).thenReturn(connection);
+		Session session = mock();
+		when(connection.createSession(false, Session.AUTO_ACKNOWLEDGE)).thenReturn(session);
+		when(session.createTemporaryQueue()).thenReturn(mock());
+		when(session.createQueue("testDestination")).thenReturn(mock());
+		when(converter.toMessage(any(), eq(session))).thenReturn(mock());
+		when(session.createProducer(any())).thenReturn(mock());
+		MessageConsumer consumer = mock();
+		when(session.createConsumer(any())).thenReturn(consumer);
+		when(consumer.receive(anyLong())).thenReturn(mock());
```
artembilan: See `MockSettings` instead of all of these, e.g. `Answers.RETURNS_MOCKS`.
More over I think this is too much just for an exception in the end.
See if `StubConnection` and friends can help anyhow instead of this mock nightmare.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -43,3 +43,8 @@ See xref:jdbc/metadata-store.adoc[] for more information.
 
 The `RestTemplate`-based configuration for the HTTP outbound gateway and outbound channel adapter (the `rest-template` attribute, and the respective Java DSL and constructor options) is now deprecated in favor of `RestClient`-based configuration, which also gains new options for configuring error handling.
 See the xref:http.adoc[HTTP Support] chapter for more information about these and other changes.
+
+[[x7.2-jms-changes]]
+=== JMS Support Changes
```
artembilan, line 48: Missed blank line after the title.

src/reference/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -43,3 +43,8 @@ See xref:jdbc/metadata-store.adoc[] for more information.
...
+Spring Integration supports Spring Framework's ability to return `null` when converting a JMS `Message` where the payload has not been set.
+This also means that if a Spring `Message` uses the payload of a JMS `Message` that is null, a  `MessageConversionException` will be thrown.
```
artembilan: I think this text needs to be revised.
It is also missing a link to the target chapter.
And I don't see a change in the target chapter.

artembilan (changes requested):

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/PollableJmsChannelTests.java
```diff
@@ -87,6 +93,26 @@ public void queueReference() throws Exception {
...
+		JmsChannelFactoryBean factoryBean = new JmsChannelFactoryBean(false);
+		CachingConnectionFactory ccf = new CachingConnectionFactory(connectionFactory);
+		ccf.setCacheConsumers(false);
+		factoryBean.setConnectionFactory(ccf);
+		factoryBean.setDestinationName("nullPayloadPollableQueue");
+		factoryBean.setPubSubDomain(false);
+		factoryBean.setBeanFactory(mock());
+		MessageConverter messageConverter = spy(new SimpleMessageConverter());
+		given(messageConverter.fromMessage(any())).willReturn(null);
+		factoryBean.setMessageConverter(messageConverter);
```
artembilan: I don't think I asked for a `spy` when said that same `factoryBean.setMessageConverter(mock());` trick could be used here.
Why do we do it different in this class than in the next one?
Why do we need to stub `willReturn(null)` if `mock()` returns `null` by default?
And why is the `spy()` if we don't verify anything in the end.
And I agree that we must not.
Please, elaborate on this code which is not what we've agreed upon in the next class.

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,62 @@
...
+ * @since 7.2
+ */
+class JmsDestinationPollingSourceTests implements TestApplicationContextAware {
+
+	@Test
+	void nullPayloadFromConverterThrowsMessageConversionException() throws Exception {
+		Message jmsMessage = new StubTextMessage("test");
+
+		MessageConverter converter = mock();
+		given(converter.fromMessage(any())).willReturn(null);
```
artembilan: Again: why explicit `.willReturn(null)` if it is assumed by default via `mock()`?

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,62 @@
...
+
+		JmsTemplate jmsTemplate = mock();
+		given(jmsTemplate.getMessageConverter()).willReturn(converter);
+		given(jmsTemplate.receiveSelected(nullable(String.class))).willReturn(jmsMessage);
+
+		JmsDestinationPollingSource source = new JmsDestinationPollingSource(jmsTemplate);
+		source.setBeanFactory(TEST_INTEGRATION_CONTEXT);
```
artembilan: I'm not sure if we need real `ApplicationContext` in this test class at all since everything else are just mocks.
Cannot just `mock()` work here, too?

src/reference/antora/modules/ROOT/pages/jms.adoc
```diff
@@ -858,6 +858,9 @@ NOTE: When you provide your own `MessageConverter` instance, it is still wrapped
 This means that the 'extract-request-payload' and 'extract-reply-payload' properties can affect the actual objects passed to your converter.
 The `HeaderMappingMessageConverter` itself delegates to a target `MessageConverter` while also mapping the Spring Integration `MessageHeaders` to JMS message properties and back again.
 
+IMPORTANT: If `MessageConverter.fromMessage()` returns `null` (for example, because the JMS `Message` has no payload), Spring Integration treats this as a conversion failure and throws a `MessageConversionException`, rather than silently discarding the message.
+This applies to the inbound and outbound channel adapters, the inbound and outbound gateways, and JMS-backed message channels (see xref:jms.adoc#jms-channel[JMS-backed Message Channels]).
```
artembilan: I don't think it is true for the `outbound channel adapters`.

I also think that we don't need full link to that `jms-channel` section exactly in this `jms.adoc` file.

Please, revise.

artembilan (changes requested):
Very close.
Thank you!

spring-integration-jms/src/main/java/org/springframework/integration/jms/outbound/JmsOutboundGateway.java
```diff
@@ -780,6 +798,9 @@ private AbstractIntegrationMessageBuilder<?> buildReply(jakarta.jms.Message jmsR
 			result = this.messageConverter.fromMessage(jmsReply);
 			logger.debug(() ->
 					"converted JMS Message [" + jmsReply + "] to integration Message payload [" + result + "]");
```
artembilan: Why this debug message was not fixed to `if..else` like in other place we have agreed?

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/PollableJmsChannelTests.java
```diff
@@ -87,6 +92,24 @@ public void queueReference() throws Exception {
...
+
+		MessageConverter messageConverter = mock();
+
+		JmsTemplate jmsTemplate = mock();
+		given(jmsTemplate.getMessageConverter()).willReturn(messageConverter);
+		given(jmsTemplate.receive()).willReturn(jmsMessage);
+
+		PollableJmsChannel channel = new PollableJmsChannel(jmsTemplate);
+		channel.setBeanFactory(mock());
+		channel.afterPropertiesSet();
```
artembilan: According to `channel.receive()` code flow for our `null` payload, we don't need these two lines.

spring-integration-jms/src/test/java/org/springframework/integration/jms/channel/SubscribableJmsChannelTests.java
```diff
@@ -282,6 +283,25 @@ public void dispatcherHasNoSubscribersQueue() throws Exception {
...
+	public void nullPayloadFromConverterThrowsMessageConversionException() throws Exception {
+		JmsChannelFactoryBean factoryBean = new JmsChannelFactoryBean(true);
+		factoryBean.setConnectionFactory(connectionFactory);
+		factoryBean.setDestinationName("nullPayloadSubscribableQueue");
+		factoryBean.setBeanName("nullPayloadChannel");
+		factoryBean.setMessageConverter(mock());
+		factoryBean.setBeanFactory(mock());
+		factoryBean.afterPropertiesSet();
+		SubscribableJmsChannel channel = (SubscribableJmsChannel) factoryBean.getObject();
+		channel.afterPropertiesSet();
```
artembilan, line 297: This is suspicious since it has been just called for its `FB`.
Please, revise the setting of this test (and probably all new introduced here) to get rid of those options which are out of use and don't contribute to the test outcome.

spring-integration-jms/src/test/java/org/springframework/integration/jms/inbound/JmsDestinationPollingSourceTests.java
```diff
@@ -0,0 +1,68 @@
...
+
+		BeanFactory beanFactory = mock();
+		String evaluationContextBeanName = IntegrationContextUtils.INTEGRATION_EVALUATION_CONTEXT_BEAN_NAME;
+		given(beanFactory.containsBean(evaluationContextBeanName)).willReturn(true);
+		given(beanFactory.getBean(evaluationContextBeanName, StandardEvaluationContext.class))
+				.willReturn(new StandardEvaluationContext());
+
+		JmsDestinationPollingSource source = new JmsDestinationPollingSource(jmsTemplate);
+		source.setBeanFactory(beanFactory);
+		source.afterPropertiesSet();
```
artembilan: Same here.
If we remove these two lines, we won't need to worry about SpEL at all.

spring-integration-jms/src/test/java/org/springframework/integration/jms/outbound/JmsOutboundGatewayTests.java
```diff
@@ -155,6 +161,40 @@ public void testReplyContainerRecovery() throws Exception {
...
+				return null;
+			}
+
+		});
+		gateway.setBeanFactory(TEST_INTEGRATION_CONTEXT);
+
+		Connection connection = new StubConnection("reply");
+		when(connectionFactory.createConnection()).thenReturn(connection);
+
+		gateway.afterPropertiesSet();
```
artembilan: DITTO

src/reference/antora/modules/ROOT/pages/jms.adoc
```diff
@@ -928,6 +931,8 @@ Use `subscription` to name the subscription.
 Starting with version 7.1, the `AbstractJmsChannel` and its implementations use a `DefaultJmsHeaderMapper` internally to map all headers from the message-to-send to JMS message properties; and on consuming side, map all JMS message properties into message headers.
 The custom header-mapping and payload conversion logic can be done by the `org.springframework.jms.support.converter.MessagingMessageConverter` injection into the `JmsTemplate` used for `AbstractJmsChannel` instances.
 
+Starting with version 7.2, if the `MessageConverter` returns `null` when converting an incoming JMS `Message`, `AbstractJmsChannel` (and, therefore, `SubscribableJmsChannel` and `PollableJmsChannel`) treats it as a conversion failure and throws a `MessageConversionException` rather than discarding the message.
```
artembilan: Why cannot we say here something like instead "the `AbstractJmsChannel` implementations..."?
What if we are going to have more than those you've just iterated?

src/reference/antora/modules/ROOT/pages/jms.adoc
```diff
@@ -858,6 +858,9 @@ NOTE: When you provide your own `MessageConverter` instance, it is still wrapped
 This means that the 'extract-request-payload' and 'extract-reply-payload' properties can affect the actual objects passed to your converter.
 The `HeaderMappingMessageConverter` itself delegates to a target `MessageConverter` while also mapping the Spring Integration `MessageHeaders` to JMS message properties and back again.
 
+IMPORTANT: If `MessageConverter.fromMessage()` returns `null` (for example, because the JMS `Message` has no payload), Spring Integration treats this as a conversion failure and throws a `MessageConversionException`, rather than silently discarding the message.
+This applies to the inbound channel adapters, the inbound and outbound gateways, and JMS-backed message channels.
```
artembilan, line 862: And why did you remove the link?
I asked just for a reference in the same chapter, similar to the one in the end of `grpc.adoc`:
```
See <<grpc-outbound-method-name-configuration>>.
```

artembilan (changes requested):

artembilan:
Attention required: 

Jms `MessageConverter`s  now return `null` when converting an incoming JMS `Message` with no payload. 
The `AbstractJmsChannel` treats it as a conversion failure and throws a `MessageConversionException` rather than discarding the message.

PR #11353, opened 2026-08-22
GH-11352: Propagate a failing close() when writing files

### LINKED ISSUE #11352: `FileWritingMessageHandler`: a failing `close()` is swallowed and the truncated file is still promoted

Reproduced against `main` (7.2.0-SNAPSHOT). `cleanUpFileState` is identical in `v6.3.0`, `v6.5.0`, `v7.0.0` and `v7.1.0`.

**Describe the bug**

`cleanUpFileState()` discards every `IOException`, including the one raised by `closeable.close()`:

```java
private void cleanUpFileState(File fileToWriteTo, @Nullable FileState state, @Nullable Closeable closeable) {
	try {
		if (state == null || this.flushTask == null) {
			if (closeable != null) {
				closeable.close();
			}
			clearState(fileToWriteTo, state);
		}
		else {
			state.lastWrite = System.currentTimeMillis();
		}
	}
	catch (IOException ex) {
	}
}
```

All three write paths call it from a `finally` block - `writeStringToFile()`, `writeBytesToFile()` and `appendStreamToFile()`. The closeable is a `BufferedWriter` or a `BufferedOutputStream`, so `close()` is where the buffer is flushed, and therefore where a full disk or a failing file system surfaces. That exception is dropped, the caller returns normally, and `cleanUpAfterCopy()` renames the temporary file onto the result file. A truncated - or, if the payload fits in the buffer, empty - file is promoted, and the message is reported as written.

Two details that constrain a fix. The `state != null && this.flushTask != null` branch deliberately skips the close for `APPEND_NO_FLUSH` and should stay as it is. And when `close()` throws, `clearState()` is skipped as well, so an `APPEND` entry can be left behind in `fileStates`.

**To Reproduce**

Add the following test to `FileWritingMessageHandlerTests`. The first two assertions pass, which is the bug: no exception reaches the caller and the result file is created. The third fails with `expected: "important payload" but was: ""`.

```java
@Test
void closeFailureIsSwallowedAndTruncatedFileIsPromoted(@TempDir File localRoot) throws Exception {
	File out = new File(localRoot, "out");
	// A writer whose close() fails the way a full disk does: the buffered content
	// never reaches the file and the IOException surfaces from close().
	FileWritingMessageHandler handler = new FileWritingMessageHandler(out) {

		@Override
		protected BufferedWriter createWriter(File fileToWriteTo, boolean append)
				throws FileNotFoundException {

			Writer target;
			try {
				target = new FileWriter(fileToWriteTo, append);
			}
			catch (IOException ex) {
				throw new FileNotFoundException(ex.getMessage());
			}
			return new BufferedWriter(target) {

				@Override
				public void close() throws IOException {
					target.close();
					throw new IOException("No space left on device");
				}

			};
		}

	};
	handler.setFileNameGenerator(message -> "payload.txt");
	handler.setOutputChannel(new NullChannel());
	handler.setBeanFactory(TEST_INTEGRATION_CONTEXT);
	handler.setApplicationContext(new GenericApplicationContext());
	handler.afterPropertiesSet();

	assertThatCode(() -> handler.handleMessage(new GenericMessage<>("important payload")))
			.doesNotThrowAnyException();

	File result = new File(out, "payload.txt");
	assertThat(result).exists();
	assertThat(new String(Files.readAllBytes(result.toPath()), StandardCharsets.UTF_8))
			.isEqualTo("important payload");
}
```

Running the whole class gives 34 tests, 1 failure - only the new one.

**Expected behavior**

A write whose data never reached the disk should not be reported as successful, and the temporary file should not be promoted. `close()` on a buffered writer is part of the write, so a failure there belongs to the caller.

The clean up still has to run on the failure path, and it must not mask the exception that is already on its way out, so the close failure should be attached as a suppressed exception there - the same rule try-with-resources uses. The `state != null && this.flushTask != null` branch stays untouched, and `clearState()` still runs either way.

**Sample**

The reproducer above is self-contained within the existing `FileWritingMessageHandlerTests` harness. I have the fix ready and will open a pull request against this issue.

dlwldn30:
Fixes: https://github.com/spring-projects/spring-integration/issues/11352

`cleanUpFileState()` discarded every `IOException`, including the one raised by closing the buffered writer or stream. All three write paths call it from a `finally` block, and `close()` is where the buffer is flushed, so a full disk was reported as a successful write and the truncated temporary file was still renamed onto the result file.

`cleanUpFileState()` now propagates and the callers invoke it on the success path, where a failure reaches the caller. On the failure path it goes through `cleanUpFileStateSuppressing()`, which attaches the close error to the exception already on its way out instead of replacing it, the way try-with-resources does. `clearState()` still runs in both cases and the `APPEND_NO_FLUSH` branch is unchanged. Keeping a single `finally` and handing it the in-flight exception would work too and would keep the call sites shorter; happy to reshape it that way if you prefer.

Verified with `./gradlew --rerun-tasks :spring-integration-file:check`: 332 tests, 0 failures, 6 skipped, checkstyle clean on main and test. `FileWritingMessageHandler` is referenced only from this module, so that is the whole affected surface. Reverting just the production file makes the new test fail and leaves the other 331 passing.

Contributed on behalf of [Goatshave](https://github.com/Goatshave).

### DIFF
--- a/spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
+++ b/spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
@@ -17,12 +17,16 @@
 package org.springframework.integration.file.outbound;
 
 import java.io.BufferedOutputStream;
+import java.io.BufferedWriter;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.attribute.PosixFilePermission;
@@ -90,6 +94,7 @@ import static org.mockito.Mockito.when;
  * @author Alen Turkovic
  * @author Glenn Renfro
  * @author Thomas Knall
+ * @author Jiwoo Lee
  */
 public class FileWritingMessageHandlerTests implements TestApplicationContextAware {
 
@@ -774,6 +779,49 @@ public class FileWritingMessageHandlerTests implements TestApplicationContextAwa
 		assertThat(result).isNotNull();
 		assertThat(result.getPayload()).isInstanceOf(File.class);
 		return (File) result.getPayload();
+	}
+
+	@Test
+	void closeFailureIsReportedAndTheFileIsNotPromoted(@TempDir File localRoot) throws Exception {
+		File out = new File(localRoot, "out");
+		// A writer whose close() fails the way a full disk does: the buffered content
+		// never reaches the file and the IOException surfaces from close().
+		FileWritingMessageHandler handler = new FileWritingMessageHandler(out) {
+
+			@Override
+			protected BufferedWriter createWriter(File fileToWriteTo, boolean append)
+					throws FileNotFoundException {
+
+				Writer target;
+				try {
+					target = new FileWriter(fileToWriteTo, append);
+				}
+				catch (IOException ex) {
+					throw new FileNotFoundException(ex.getMessage());
+				}
+				return new BufferedWriter(target) {
+
+					@Override
+					public void close() throws IOException {
+						target.close();
+						throw new IOException("No space left on device");
+					}
+
+				};
+			}
+
+		};
+		handler.setFileNameGenerator(message -> "payload.txt");
+		handler.setOutputChannel(new NullChannel());
+		handler.setBeanFactory(TEST_INTEGRATION_CONTEXT);
+		handler.setApplicationContext(new GenericApplicationContext());
+		handler.afterPropertiesSet();
+
+		assertThatExceptionOfType(MessageHandlingException.class)
+				.isThrownBy(() -> handler.handleMessage(new GenericMessage<>("important payload")))
+				.withStackTraceContaining("java.io.IOException: No space left on device");
+
+		assertThat(new File(out, "payload.txt")).doesNotExist();
 	}
 
 }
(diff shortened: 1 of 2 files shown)


artembilan (approved):
Verified locally on the PR branch (isolated worktree): `./gradlew --rerun-tasks :spring-integration-file:check` → BUILD SUCCESSFUL, new test passes, checkstyle clean.

Reviewed the logic against the linked issue:
- `cleanUpFileState()` now propagates `close()` failures via a `try { close() } finally { clearState() }`, so a truncated write is no longer promoted to the result file, and `clearState()` still runs (fixing the "APPEND entry left behind in `fileStates`" concern raised in the issue) even when `close()` throws.
- `cleanUpFileStateSuppressing()` correctly mirrors try-with-resources semantics: on the failure path it attaches any close error as `addSuppressed` on the exception already propagating, rather than replacing it.
- The `state != null && this.flushTask != null` (`APPEND_NO_FLUSH`) branch is untouched, as required — deferred/buffered close failures there are already handled (silently, pre-existing) by `FileState.close()`/the async `Flusher`, out of scope for this fix.
- Traced exception propagation end-to-end: `writeStringToFile`/`writeBytesToFile`/`appendStreamToFile` (`throws IOException`) → `LockRegistry.executeLocked` (generic `throws E`) → `handle*Message` (`throws IOException`) → surfaces as `MessageHandlingException`, matching the new test's assertion. Confirmed the temp file is never renamed onto the result file on this path.

Nice, minimal, well-scoped fix with a solid regression test. No concerns.

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
```diff
@@ -776,4 +781,47 @@ protected File messageToFile(Message<?> result) {
 		return (File) result.getPayload();
 	}
 
+	@Test
+	void closeFailureIsReportedAndTheFileIsNotPromoted(@TempDir File localRoot) throws Exception {
```
artembilan: Why doesn't existing `tempDir` property in this class work here for this new method?
Why do we need a new `@TempDir` ?

artembilan (changes requested):

spring-integration-file/src/test/java/org/springframework/integration/file/outbound/FileWritingMessageHandlerTests.java
dlwldn30: Switched to the existing `tempDir` field — the extra `@TempDir` parameter is gone.

`:spring-integration-file:check` locally: 332 tests, 0 failures, checkstyle clean.
CI is still waiting on workflow approval.

PR #11367, opened 2026-08-24
Update all Spring Dependencies except SF

cppwfs:
- SF Dependency update will be in another PR
- SI MicroGaugeBuilder no longer accepts null for obj
  - This tracks with latest changes to Micrometer
  - Since all known impls in the project pass `this` there is no side effect

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,13 +1,13 @@
 [versions]
 # Spring Portfolio dependencies. Usually managed by release train
-micrometerTracingVersion = "1.8.0-M1"
-micrometerVersion = "1.18.0-M1"
+micrometerTracingVersion = "1.8.0-SNAPSHOT"
+micrometerVersion = "1.18.0-SNAPSHOT"
 reactorVersion = "2026.0.0-SNAPSHOT"
-springAmqpVersion = "4.2.0-M1"
-springDataVersion = "2026.1.0-M1"
+springAmqpVersion = "4.2.0-SNAPSHOT"
+springDataVersion = "2026.1.0-SNAPSHOT"
 springGraphqlVersion = "2.0.5"
-springKafkaVersion = "4.2.0-M1"
-springSecurityVersion = "7.1.1"
+springKafkaVersion = "4.2.0-SNAPSHOT"
+springSecurityVersion = "7.2.0-SNAPSHOT"
 springVersion = "7.1.0-M1"
 springWsVersion = "5.0.2"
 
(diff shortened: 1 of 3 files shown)


gradle/libs.versions.toml
```diff
@@ -1,13 +1,13 @@
...
-micrometerTracingVersion = "1.8.0-M1"
-micrometerVersion = "1.18.0-M1"
+micrometerTracingVersion = "1.8.0-SNAPSHOT"
+micrometerVersion = "1.18.0-SNAPSHOT"
 reactorVersion = "2026.0.0-SNAPSHOT"
-springAmqpVersion = "4.2.0-M1"
-springDataVersion = "2026.1.0-M1"
+springAmqpVersion = "4.2.0-SNAPSHOT"
+springDataVersion = "2026.1.0-SNAPSHOT"
 springGraphqlVersion = "2.0.5"
```
artembilan: `2.1.0-SNAPSHOT`

gradle/libs.versions.toml
```diff
@@ -1,13 +1,13 @@
...
-springKafkaVersion = "4.2.0-M1"
-springSecurityVersion = "7.1.1"
+springKafkaVersion = "4.2.0-SNAPSHOT"
+springSecurityVersion = "7.2.0-SNAPSHOT"
 springVersion = "7.1.0-M1"
 springWsVersion = "5.0.2"
```
artembilan: `5.1.0-SNAPSHOT`

gradle/libs.versions.toml
```diff
@@ -1,13 +1,13 @@
 [versions]
 # Spring Portfolio dependencies. Usually managed by release train
-micrometerTracingVersion = "1.8.0-M1"
-micrometerVersion = "1.18.0-M1"
+micrometerTracingVersion = "1.8.0-SNAPSHOT"
+micrometerVersion = "1.18.0-SNAPSHOT"
 reactorVersion = "2026.0.0-SNAPSHOT"
```
artembilan, line 5: Not question to you, but I'm not sure why Dependabot has updated only Reactor over weekend.
We do have respective snapshots for Spring AMQP and Kafka for a while.
🤷

artembilan (changes requested):

gradle/libs.versions.toml
cppwfs: I was curious about that myself.

artembilan: Just to close the gap:
there is a new feature in Dependabot - `cooldown`: https://github.blog/changelog/2026-07-14-dependabot-version-updates-introduce-default-package-cooldown/
😄

cppwfs: Nice idea...
Explains the behavior.

PR #11381, opened 2026-08-27
Restart Apache Commons `Tailer` if its background thread dies

artembilan:
## Problem

`ApacheCommonsFileTailingMessageProducer` schedules the `commons-io` `Tailer` on
a task executor and never checks on it again:

```java
Tailer theTailer = Tailer.builder()...get();
getTaskExecutor().execute(theTailer);
this.tailer = theTailer;
```

If the `Tailer`'s background thread terminates for *any* reason, the adapter
silently stops producing messages forever - there is no restart logic, unlike
`OSDelegatingFileTailingMessageProducer`, which already reschedules the native
`tail` process via `startProcessMonitor()` if it dies while the adapter is
still running.

## How this is triggered in practice

Reading `Tailer.run()` (`commons-io` 2.22.0) shows a real path to that silent
death, on ordinary log rotation:

```java
if (length < position) {
    // File was rotated
    listener.fileRotated();
    try (RandomAccessResourceBridge save = reader) {
        reader = tailable.getRandomAccess(RAF_READ_ONLY_MODE); // (1)
        ...
        position = 0;
    } catch (final FileNotFoundException e) {
        // in this case we continue to use the previous reader and position values
        listener.fileNotFound();
        ThreadUtils.sleep(delayDuration);
    }
    continue;
}
```

If the reopen at (1) throws `FileNotFoundException` - a normal window during
rotation, when the old file has been moved/removed but the replacement hasn't
been created yet - the comment claims the previous `reader` is still usable.
It isn't: the try-with-resources already closed it as `save` before the reopen
attempt failed, and `reader` still points at that closed handle. Once the
replacement file grows past the stale `position`, the next poll takes the
`length > position` branch and calls `readLines(reader)` on the closed
resource, which throws. `Tailer.run()`'s outer `catch (Exception e)` swallows
it (`listener.handle(e)`) and the method returns - the tailing thread is gone
for good, with only a published event (if the caller happens to be listening
for `FileTailingEvent`s) as a trace.

I could not reproduce this race on Windows locally, because `File.renameTo()`
returns `false` there while the `Tailer` holds an open read handle on the
source file, so the "file briefly absent" window this needs never occurs; the
originally reported failure came from a POSIX CI runner where rename-in-place
is unremarkable.

## Fix

* Wrap the `Tailer` execution so that, once `Tailer.run()` returns while this
  adapter `isActive()`, a brand new `Tailer` is built and its start is
  scheduled after `tailAttemptsDelay`, mirroring
  `OSDelegatingFileTailingMessageProducer`'s existing restart behavior.
* Track and cancel any pending restart in `doStop()`.
* Added `FileTailingMessageProducerTests#apacheTailerRestartsAfterUnderlyingThreadDies`,
  which kills the underlying `Tailer` directly (instead of trying to
  reproduce the OS-timing-dependent rotation race) and asserts that message
  production resumes afterward.

## Test plan

- [x] `./gradlew :spring-integration-file:test --tests "org.springframework.integration.file.tail.*"` - all green, repeated runs stable
- [x] `./gradlew :spring-integration-file:checkstyleMain :spring-integration-file:checkstyleTest`

### DIFF
--- a/spring-integration-file/src/main/java/org/springframework/integration/file/tail/ApacheCommonsFileTailingMessageProducer.java
+++ b/spring-integration-file/src/main/java/org/springframework/integration/file/tail/ApacheCommonsFileTailingMessageProducer.java
@@ -17,6 +17,8 @@
 package org.springframework.integration.file.tail;
 
 import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.ScheduledFuture;
 
 import org.apache.commons.io.input.Tailer;
 import org.apache.commons.io.input.TailerListener;
@@ -43,6 +45,8 @@ public class ApacheCommonsFileTailingMessageProducer extends FileTailingMessageP
 	private boolean reopen = false;
 
 	private volatile @Nullable Tailer tailer;
+
+	private volatile @Nullable ScheduledFuture<?> restartFuture;
 
 	/**
 	 * The delay between checks of the file for new content in milliseconds.
@@ -87,6 +91,18 @@ public class ApacheCommonsFileTailingMessageProducer extends FileTailingMessageP
 	@Override
 	protected void doStart() {
 		super.doStart();
+		runTailer();
+	}
+
+	/**
+	 * Build and run a new {@link Tailer}, arranging for it to be restarted (after
+	 * {@link #getMissingFileDelay()}) should its background thread terminate while this
+	 * adapter is still active. The underlying {@code commons-io} {@link Tailer} can leave
+	 * its reader in a broken state and let the thread die if a file rotation is observed
+	 * concurrently with the replacement file being briefly absent; without a restart, such
+	 * an event would otherwise silently stop message production forever.
+	 */
+	private void runTailer() {
 		Tailer theTailer =
 				Tailer.builder()
 						.setDelayDuration(this.pollingDelay)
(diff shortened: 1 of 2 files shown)


artembilan:
Worth flagging separately: the underlying issue is arguably a bug in `commons-io`'s `Tailer.run()` itself, not something we can fully fix from our side. The comment at the `FileNotFoundException` catch after a detected rotation says:

```java
} catch (final FileNotFoundException e) {
    // in this case we continue to use the previous reader and position values
    listener.fileNotFound();
    ThreadUtils.sleep(delayDuration);
}
```

but by that point the previous `reader` has already been closed by the enclosing `try (RandomAccessResourceBridge save = reader) { reader = tailable.getRandomAccess(...); ... }` (as `save`), since the reopen assignment to `reader` never completed. The comment's stated intent (keep using the old reader/position) is not what the code does - the next successful poll calls `readLines()` on a closed resource, throws, and the outer `catch (Exception e)` in `run()` swallows it and lets the thread die.

This restart wrapper here papers over that (and is a reasonable thing for us to have regardless, mirroring `OSDelegatingFileTailingMessageProducer`), but the real fix - not silently ending up with a dead reader reference after a failed reopen-on-rotation - belongs upstream in `commons-io`. I'll look at reporting this to the `commons-io` project separately.

